PR #13, opened 2017-12-14
connect/exec: Catch throws instead of dumping core.  Fixes #7

### LINKED ISSUE #7: terminate called after throwing an instance of 'std::runtime_error'
I launched a multipass then started something intensive, and now the vm has locked up. I want to get in to debug it, but `multipass connect` won't let me. Now, I don't mind that I can't get in, I'll throw this away and start again, but the error message I get when I try and connect could do with some love:- 

```
alan@hal:~$ multipass connect finer-mite 
terminate called after throwing an instance of 'std::runtime_error'
  what():  ssh: No route to host
Aborted (core dumped)

```

townsend2010:


src/client/cmd/connect.cpp
```diff
@@ -45,8 +45,16 @@ mp::ReturnCode cmd::Connect::run(mp::ArgParser* parser)
...
-        mp::SSHClient ssh_client{host, port, priv_key_blob};
-        ssh_client.connect();
+        try
+        {
+            mp::SSHClient ssh_client{host, port, priv_key_blob};
+            ssh_client.connect();
+        }
+        catch (const std::exception& e)
+        {
+            cerr << "connect failed: " << e.what() << std::endl;
```
Saviq: Think we can give better messages here, at least for some common scenarios?

src/client/cmd/exec.cpp
```diff
@@ -48,8 +48,16 @@ mp::ReturnCode cmd::Exec::run(mp::ArgParser* parser)
...
-        mp::SSHClient ssh_client{host, port, priv_key_blob};
-        return static_cast<mp::ReturnCode>(ssh_client.exec(args));
+        try
+        {
+            mp::SSHClient ssh_client{host, port, priv_key_blob};
+            return static_cast<mp::ReturnCode>(ssh_client.exec(args));
+        }
+        catch (const std::exception& e)
+        {
+            cerr << "exec failed: " << e.what() << std::endl;
```
Saviq: Ditto.

Saviq:
Also, could we have tests for this?

src/client/cmd/connect.cpp
townsend2010: The what() should give a pretty good description about what the error is.  In the case that popey put in the message would be:
connect failed: ssh: No route to host

Other ssh issues will have different messages.  Is that good enough?

townsend2010:
Regarding tests, I think Alberto has been working on a bunch of refactoring to make things more testable, but currently, testing this is not really possible.

src/client/cmd/connect.cpp
townsend2010: It's coming from libssh via ssh_get_error().  The "ssh:" part of this is something we add when throwing the runtime_error.  See include/multipass/ssh/throw_on_error.h for more detail.  I'm really not sure how much more we can do without redoing a bunch of stuff...

townsend2010: I guess I could just put in something like, "Can't connect to the instance" for any errors and be done with it?

PR #20, opened 2017-12-15
sshfs_mount: Release ownership of channel unique_ptr. Fixes #18

### LINKED ISSUE #18: Unmounting a share can sometimes lead to a segfault in multipassd
Seems to be a race condition when unmounting shares.  I've already done some investigation, but seems to have been introduced when we switched over to using the SSHProcess class.

townsend2010:


albaguirre:
Yeah LGTM.

PR #25, opened 2017-12-20
snap: use plain source-type

Saviq:
Since we're in a git checkout already, using `source-type: git` only
really hurts us, because it pulls all the submodules from the network
instead locally, and that's 500MB of git repos at the moment...


Saviq:
Yay. This saves us 4 minutes on CI, likely more on local snap builds.

PR #27, opened 2017-12-20
qemu: Add script to check for KVM support. Fixes #12

### LINKED ISSUE #12: KVM support detection not robust enough
Following the instructions in the README. I got the following issue.

```
$ sudo snap install multipass --beta --classic
multipass (beta) 2017.2.2 from 'saviq' installed
$ multipass launch --name foo
failed to launch: Could not access KVM kernel module: No such file or directory 
failed to initialize KVM: No such file or directory
```

townsend2010:


snap/snapcraft.yaml
```diff
@@ -28,6 +28,11 @@ apps:
       QT_PLUGIN_PATH: $SNAP/usr/lib/x86_64-linux-gnu/qt5/plugins
     command: bin/multipass
     completer: etc/bash_completion.d/snap.multipass
+  check-kvm-support:
+    command: bin/check_kvm_support
```
Saviq: We don't need to expose this in the snap, it's enough that you install it and then invoke it from a known path.

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -261,19 +261,12 @@ void mp::QemuVirtualMachineFactory::configure(const std::string& name, YAML::Nod
...
 void mp::QemuVirtualMachineFactory::check_hypervisor_support()
 {
-    QProcess cmd;
-    cmd.start("bash",
-              QStringList() << "-c"
-                            << "egrep -m1 -w '^flags[[:blank:]]*:' /proc/cpuinfo | egrep -wo '(vmx|svm)'");
-    cmd.waitForFinished();
-    auto virt_type = cmd.readAll().trimmed();
+    QProcess check_kvm;
+    check_kvm.start("check_kvm_support");
```
Saviq: How would this work when running from a local build?

src/platform/backends/qemu/CMakeLists.txt
```diff
@@ -24,3 +24,7 @@ target_link_libraries(qemu_backend
   ip_address_pool
   ssh
   yaml)
+
+install(PROGRAMS check_kvm_support
+  DESTINATION bin
```
Saviq: Best use ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} here.

src/platform/backends/qemu/CMakeLists.txt
```diff
@@ -24,3 +24,7 @@ target_link_libraries(qemu_backend
...
+  COMPONENT multipassd)
```
Saviq: Does this change anything?

snap/snapcraft.yaml
townsend2010: Ok, cool, I'll remove this.

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
townsend2010: It won't work, but do we really care?  Local builds would be for developers, and developers should be able to figure if their machine currently supports kvm, right?

src/platform/backends/qemu/CMakeLists.txt
townsend2010: Not sure I understand why that is necessary...what is the advantage of doing that?

src/platform/backends/qemu/CMakeLists.txt
townsend2010: Well, "COMPONENT" was specified for install of multipassd in src/daemon/CMakeLists.txt, so I think we should specify the same component for check_kvm_support since is should be installed when multipassd is.

src/platform/backends/qemu/CMakeLists.txt
Saviq: Hmm I suppose there is none, wonder what it gives us to set it in the first place, if we're `install()`ing it manually into bin in the first place.

src/platform/backends/qemu/CMakeLists.txt
Saviq: Ack

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
Saviq: Well, it will throw just below, no?

src/platform/backends/qemu/CMakeLists.txt
townsend2010: I thought it was needed for 'make install' when building manually.

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
townsend2010: No, the return value from bash of a missing binary is 127, so it will just basically be a no-op since I'm only checking a return value of 1.

Saviq: d'oh.

snap/snapcraft.yaml
Saviq: Just this, then, and I'll be happy with the code, will test out first thing tomorrow.

Saviq:
The service didn't seem to do its thing:

```
$> systemctl status snap.multipass.qemu-kvm-init.service 
● snap.multipass.qemu-kvm-init.service - Service for snap application multipass.qemu-kvm-init
   Loaded: loaded (/etc/systemd/system/snap.multipass.qemu-kvm-init.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Fri 2018-01-05 21:05:24 CET; 21s ago
 Main PID: 8656 (code=exited, status=0/SUCCESS)

sty 05 21:05:24 michal-laptop systemd[1]: Starting Service for snap application multipass.qemu-kvm-init...
sty 05 21:05:24 michal-laptop systemd[1]: Started Service for snap application multipass.qemu-kvm-init.
$> lsmod | grep kvm
$> 
```

Looks like it assumes `qemu-kvm-*` paths...
```
case "$arch" in
    x86_64 | i686)
        kvm=/usr/bin/qemu-system-x86_64
[...]
# Silently exit if the package isn't installed anymore
if [ -z "$kvm" -o ! -e "$kvm" ]; then
    exit 0
```

Then, this happened:
```
$> multipass start adorable-boar    
start failed: Could not access KVM kernel module: No such file or directory                                            
failed to initialize KVM: No such file or directory        
```
So the `check_kvm_support` script didn't run?

townsend2010:
Hmm, this all works on my system with a locally built snap from this branch.

$ sudo modprobe -r kvm_intel 
$ lsmod | grep kvm
$ sudo snap install multipass_2017.0.0_amd64.snap --dangerous --classic
multipass 2017.2.2-19-g9182784 installed
$ lsmod | grep kvm
kvm_intel             200704  0
kvm                   581632  1 kvm_intel
irqbypass              16384  1 kvm
$ systemctl status snap.multipass.qemu-kvm-init.service
● snap.multipass.qemu-kvm-init.service - Service for snap application multipass.qemu-kvm-init
   Loaded: loaded (/etc/systemd/system/snap.multipass.qemu-kvm-init.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Tue 2018-01-09 12:12:31 EST; 10s ago
  Process: 13905 ExecStart=/usr/bin/snap run multipass.qemu-kvm-init (code=exited, status=0/SUCCESS)
 Main PID: 13905 (code=exited, status=0/SUCCESS)
      CPU: 55ms

Jan 09 12:12:31 Slave1 systemd[1]: Starting Service for snap application multipass.qemu-kvm-init...
Jan 09 12:12:31 Slave1 systemd[1]: Started Service for snap application multipass.qemu-kvm-init.

townsend2010:
Oh, maybe it's because I have qemu-system-common installed locally on the system as well?

Anyways, will try to reproduce why the check-kvm-support script didn't run either.

townsend2010:
Well, check-kvm-support also runs in my locally built snap as well.

$ sudo modprobe -r kvm_intel
$ lsmod | grep kvm
$ multipass launch -n test-snap -c 1 -m 512M xenial
failed to launch: /dev/kvm does not exist. Please try 'sudo modprobe kvm_intel'.

So I really don't know what's going on with your setup, so we'll just have to wait until you are back to try to figure this out.

townsend2010:
Oops, snapcraft.yaml is updated now.

townsend2010:
Thanks! I will wait on an ack from Saviq before merging since he had issues.

Saviq:
You should include msr-tools in the snap, I suppose? I didn't have it installed and it 

```
⟫ ./check_kvm_support
Must be root to check BIOS for KVM extensions support.
/dev/kvm does not exist. Please try 'sudo modprobe kvm_intel'.
⟫ sudo ./check_kvm_support
/dev/kvm does not exist. Please try 'sudo modprobe kvm_intel'.
⟫ sudo modprobe kvm_intel
modprobe: ERROR: could not insert 'kvm_intel': Operation not supported
```

That said, it still doesn't run automagically for me. I built the snap with
`SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft`

When my virtualization is disabled, I just get:
```
⟫ multipass start crucial-iguana
start failed: failed to determine IP address
```

When I just unload the modules (`sudo rmmod kvm_intel kvm`), qemu, understandably, fails to start and the qemu error message is forwarded on the console.

Note I've purged `*qemu*` from my system, so the init script doesn't even try to load the modules, as it checks for qemu using:
```
case "$arch" in
    x86_64 | i686)
        kvm=/usr/bin/qemu-system-x86_64
[...]
# Silently exit if the package isn't installed anymore     
if [ -z "$kvm" -o ! -e "$kvm" ]; then
    exit 0
fi
```

townsend2010:
@Saviq,

Hmm, I'm not really sure what is going on.  Both @albaguirre and I are not having any issues, so what is different with your system???  Based on both the service and the new script not running at all, it really seems something with your Snap related paths is messed up, but I don't know enough about that to know for sure.

I guess at this point, you are going to have to debug this since @albaguirre and I are unable to reproduce.

townsend2010:
Ok, I believe I tracked down the issues @Saviq was seeing and updated the branch.

I updated the branch with the following changes:
- It now also checks for the kvm support during the start command.
- Removed the oneshot qemu-init service as that doesn't always work.
- Instead, added a modprobe to the check_kvm_support script if it detects the module isn't loaded.

PR #33, opened 2018-01-09
qemu: Use dnsmasq for DNS and DHCP of VM's. Fixes #1

### LINKED ISSUE #1: Uses non-host DNS servers
I'm in an office (as it turns out Canonical) where there is no access to 8.8.8.8 which is the default DNS in multipass. Nothing resolves inside my multipass machine, but external host DNS resolution is fine.

This will also affect some of the software vendors we're talking to, who may use multipass as a way to have clean build environments for their software. If the multipass instance can't see internal repositories then it will fail to build. They may also have similar restrictions to us in terms of DNS access.

townsend2010:
- This also supports VM's that have already been created with the old static IP method.

### DIFF
--- a/src/network/ip_address_pool.cpp
+++ b/src/network/ip_address_pool.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Canonical, Ltd.
+ * Copyright (C) 2017-2018 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -110,6 +110,25 @@ mp::IPAddress mp::IPAddressPool::obtain_ip_for(const std::string& name)
         return ip;
     }
     return it->second;
+}
+
+std::experimental::optional<mp::IPAddress> mp::IPAddressPool::check_ip_for(const std::string& name)
+{
+    const auto it = ip_map.find(name);
+    if (it == ip_map.end())
+    {
+        return {};
+    }
+
+    return std::experimental::optional<mp::IPAddress>{it->second};
+}
+
+mp::IPAddress mp::IPAddressPool::first_free_ip()
+{
+    if (ips_in_use.empty())
+        return start_ip;
+
+    return *ips_in_use.crbegin() + 1;
 }
 
 void mp::IPAddressPool::remove_ip_for(const std::string& name)
--- a/src/platform/backends/qemu/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/dnsmasq_server.cpp
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2018 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "dnsmasq_server.h"
+
+#include <QFile>
+
+namespace mp = multipass;
+
+namespace
+{
+auto start_dnsmasq_process(const QDir& data_dir, const mp::IPAddress& start_ip, const mp::IPAddress& end_ip)
+{
+    auto cmd = std::make_unique<QProcess>();
+    cmd->start("dnsmasq",
+               QStringList() << "--keep-in-foreground"
+                             << "--strict-order"
+                             << "--bind-interfaces"
+                             << "--except-interface=lo"
+                             << "--interface=mpbr0"
+                             << "--listen-address=10.122.122.1"
+                             << "--dhcp-no-override"
+                             << "--dhcp-authoritative"
+                             << QString("--dhcp-leasefile=%1").arg(data_dir.filePath("dnsmasq.leases"))
+                             << QString("--dhcp-hostsfile=%1").arg(data_dir.filePath("dnsmasq.hosts")) << "--dhcp-range"
+                             << QString("%1,%2,infinite")
+                                    .arg(QString::fromStdString(start_ip.as_string()))
+                                    .arg(QString::fromStdString(end_ip.as_string())));
+
+    cmd->waitForStarted();
+    return cmd;
+}
+}
+
+mp::DNSMasqServer::DNSMasqServer(const Path& path, const IPAddress& start, const IPAddress& end)
+    : start_ip{start},
+      end_ip{end},
+      data_dir{QDir(path + "/vm-ips")},
+      dnsmasq_cmd{start_dnsmasq_process(data_dir, start_ip, end_ip)}
+{
+}
+
+mp::DNSMasqServer::~DNSMasqServer()
+{
+    dnsmasq_cmd->kill();
+    dnsmasq_cmd->waitForFinished();
+}
+
+std::experimental::optional<mp::IPAddress> mp::DNSMasqServer::get_ip_for(const std::string& hw_addr)
+{
+    QProcess cmd;
+
+    cmd.start("bash",
+              QStringList() << "-c"
+                            << QString("grep \"%1\" %2 | cut -d ' ' -f3")
+                                   .arg(QString::fromStdString(hw_addr))
+                                   .arg(data_dir.filePath("dnsmasq.leases")));
+
+    cmd.waitForFinished();
+
+    auto ip_addr = cmd.readAll().trimmed();
+
+    if (ip_addr.isEmpty())
+        return {};
+    else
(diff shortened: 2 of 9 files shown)


include/multipass/ip_address.h
```diff
@@ -40,9 +40,10 @@ struct IPAddress
     bool operator<=(const IPAddress& other) const;
     bool operator>(const IPAddress& other) const;
     bool operator>=(const IPAddress& other) const;
+    IPAddress operator=(const IPAddress& other);
     IPAddress operator+(int value) const;
 
-    const std::array<uint8_t, 4> octets;
+    std::array<uint8_t, 4> octets;
```
albaguirre: There should be no need to remove const here.

IPAddress was made so that it's invariant after construction.

src/network/ip_address.cpp
```diff
@@ -40,6 +40,10 @@ std::array<uint8_t, 4> parse(const std::string& ip)
     int b = -1;
     int c = -1;
     int d = -1;
+
+    if (ip.empty())
+        return {{static_cast<uint8_t>(a), static_cast<uint8_t>(b), static_cast<uint8_t>(c), static_cast<uint8_t>(d)}};
+
```
albaguirre: The caller should create an IPAddress object with 0xFFFFFFFF instead of giving an empty string.

src/network/ip_address.cpp
```diff
@@ -114,6 +118,12 @@ bool mp::IPAddress::operator>=(const IPAddress& other) const
     return as_uint32() >= other.as_uint32();
 }
 
+mp::IPAddress mp::IPAddress::operator=(const IPAddress& other)
+{
+    octets = other.octets;
+    return other;
+}
+
```
albaguirre: I omitted the copy assignment operator so that an instance of IPAddress is invariant.

src/network/ip_address_pool.cpp
```diff
@@ -112,6 +112,24 @@ mp::IPAddress mp::IPAddressPool::obtain_ip_for(const std::string& name)
...
+    return it->second;
+}
+
+mp::IPAddress mp::IPAddressPool::first_free_ip()
+{
+    if (ips_in_use.empty())
+        return start_ip;
+
+    return *ips_in_use.crbegin() + 1;
+}
```
albaguirre: Based on the use, I don't think this will do the right thing.

Because old VM's could have a hole in their IP Assignments, not necessarily contiguous depending on the order of how other VM's were removed.

Exposing this "first free ip" should not be exposed externally.

src/network/ip_address.cpp
albaguirre: I left out the copy assignment operator because I wanted IPAddress instances to be invariant.

include/multipass/ip_address.h
albaguirre: There should be no need to remove const here.

It's there because instances of IPAddress should be invariant after construction.

src/network/ip_address.cpp
albaguirre: I don't see why we need to special case this here.

The caller should instantiate an IPAddress with 0xFFFFFFFF instead, otherwise it should throw an exception.

src/network/ip_address_pool.cpp
```diff
@@ -112,6 +112,24 @@ mp::IPAddress mp::IPAddressPool::obtain_ip_for(const std::string& name)
     return it->second;
 }
 
+mp::IPAddress mp::IPAddressPool::check_ip_for(const std::string& name)
+{
+    const auto it = ip_map.find(name);
+    if (it == ip_map.end())
+    {
+        return mp::IPAddress{""};
```
albaguirre: This looks like it should return a std::experimental::optional<IPAddress> instead.

src/network/ip_address_pool.cpp
townsend2010: What I'm trying to accomplish here is to figure out which IP for dnsmasq to begin handing out dynamic IP's.  Since ips_in_use is a std::set, then even if there are holes in assignment, the IP's should be in order, so if I take the very last one in the set and add 1, then I'm getting the first IP in the range that dnsmasq can use.

For example, say the set has 10.122.122.2, 10.122.122.4, and 10.122.122.6, then this should return 10.122.122.7 and I can set dnsmasq to use the range of 10.122.122.7 to 10.122.122.254.  I really don't care that there are holes, all I'm trying to get is a range.

src/network/ip_address_pool.cpp
townsend2010: That's interesting and didn't know about that:)  But what does that gain us over what I'm already doing here?

Seems I would still need logic to either return the found IPAddress object or {} if not found and the caller would need to use value_or().

albaguirre: It would cleanly separate concerns. Instead of trying to make IPAddress a mutable object and have a reserved "invalid" value, you can leave IPAddress as an immutable design and have the "invalid/valid" be represented by the optional type instead - this has all the semantics that you use with the proposed code.

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -0,0 +1,81 @@
...
+                            << QString("grep \"%1\" %2 | cut -d ' ' -f3")
+                                   .arg(QString::fromStdString(hw_addr))
+                                   .arg(data_dir.filePath("dnsmasq.leases")));
+
+    cmd.waitForFinished();
+
+    auto ip_addr = cmd.readAll().trimmed();
+
+    if (ip_addr.isEmpty())
+        return mp::IPAddress{0xFFFFFFFF};
```
albaguirre: Looks like this method could also use a std::experimental::optional as well. Would make it clear for the caller that the IPAddress could be unset.

src/platform/backends/qemu/qemu_virtual_machine.h
```diff
@@ -54,9 +57,10 @@ class QemuVirtualMachine final : public VirtualMachine
     void on_error();
     void on_shutdown();
     VirtualMachine::State state;
-    const IPAddress ip;
+    std::unique_ptr<IPAddress> ip;
```
albaguirre: A unique_ptr would do the job, but I think it's the semantics required map well to an std::experimental::optional type

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -194,12 +198,32 @@ void mp::QemuVirtualMachine::on_shutdown()
...
 std::string mp::QemuVirtualMachine::ssh_hostname()
 {
-    return ip.as_string();
+    return ipv4();
 }
 
 std::string mp::QemuVirtualMachine::ipv4()
 {
-    return ip.as_string();
+    if (*ip == mp::IPAddress{0xFFFFFFFF})
```
albaguirre: With ip being a std::experimental::optional, This would just be if (!ip)

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -194,12 +198,32 @@ void mp::QemuVirtualMachine::on_shutdown()
...
+    {
+        using namespace std::literals::chrono_literals;
+
+        auto deadline = std::chrono::steady_clock::now() + std::chrono::minutes(2);
+
+        while (std::chrono::steady_clock::now() < deadline)
+        {
+            QCoreApplication::processEvents();
+            ip = std::make_unique<mp::IPAddress>(dnsmasq_server->get_ip_for(mac_addr));
```
albaguirre: if get_ip_for(...) returns an optional I believe you could directly assign it to ip here instead, i.e. ip = dnsmasq_server->get_ip_for(mac_addr);

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -194,12 +198,32 @@ void mp::QemuVirtualMachine::on_shutdown()
...
+
+            if (*ip != mp::IPAddress{0xFFFFFFFF})
```
albaguirre: With ip as a std::experimental::optional, I believe this can become if(ip) return ip.value().as_string();

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -205,8 +206,9 @@ mp::VirtualMachine::UPtr mp::QemuVirtualMachineFactory::create_virtual_machine(c
     set_ip_forward();
     set_nat_iptables();
 
-    return std::make_unique<mp::QemuVirtualMachine>(desc, ip_pool.obtain_ip_for(desc.vm_name), tap_device_name,
-                                                    generate_mac_address(), monitor);
+    return std::make_unique<mp::QemuVirtualMachine>(
+        desc, ip_pool.check_ip_for(desc.vm_name).value_or(mp::IPAddress{0xFFFFFFFF}), tap_device_name,
```
albaguirre: QemuVirtualMachine should also take a std::experimental::optional<IPAddress> which then would allow you to remove:
.value_or(mp::IPAddress{0xFFFFFFFF})

src/network/ip_address_pool.cpp
albaguirre: OK, based on IRC conversation it should be ok with the caveat that a big gap in IP addresses could exist if the user created many VM's then deleted VM's in between but we'll assume that would be a rare occurrence.

albaguirre:
For the snap, we need to include dnsmasq as well.

PR #42, opened 2018-01-16
client: Support JSON output for the list and info commands.

townsend2010:


src/client/cmd/list.cpp
```diff
@@ -31,25 +35,64 @@ using RpcMethod = mp::Rpc::Stub;
...
+{
+    QJsonObject list_json;
+    QJsonArray instances;
+
+    for (const auto& instance : reply.instances())
+    {
+        QJsonObject instance_obj;
+        instance_obj.insert("Name", QString::fromStdString(instance.name()));
+        instance_obj.insert("State", QString::fromStdString(status_string(instance.status())));
+        instance_obj.insert("IPv4", QString::fromStdString(instance.ipv4()));
```
kalikiana: IPv6 missing here?

src/client/cmd/info.cpp
```diff
@@ -94,6 +101,38 @@ std::ostream& operator<<(std::ostream& out, const multipass::InfoReply::Info& in
...
+
+QString toJson(mp::InfoReply& reply)
+{
+    QJsonObject info_json;
+
+    for (const auto& info : reply.info())
+    {
+        QJsonObject instance_info;
+        instance_info.insert("State", QString::fromStdString(status_string(info.status())));
+        instance_info.insert("IPv4", QString::fromStdString(info.ipv4()));
```
kalikiana: IPv6 missing here?

Saviq:
It feels weird to wrap this in a top-level dictionary... Unless we have a use case for the dictionary, like errors?

And if we are, we should wrap consistently, like `info` doesn't wrap now, but `list` does. Errors are printed in plain text though. Should we try and output errors in JSON as well?

I'd agree with @kalikiana about lowercase (and no whitespace?) keys.

Should we also print out data in more machine-parsable way there? Like the release name, disk and memory usage? A dictionary of `interface: [IP, IP]`?

Basically, I'd like us to be cautious about what we do here to avoid breaking customers later - better leave out some details than change the format then.

src/client/cmd/info.cpp
townsend2010: We don't support IPv6 yet.  It had also been removed from the table output as well.

src/client/cmd/list.cpp
townsend2010: Same as above:)

townsend2010:
Ok, I'll make keys lowercase and with no whitespace.

@Saviq, regarding how we want everything laid out in the json output, could you please add some example output of how you would like to see it in the CLI doc? I made a best guess based on the current table output and how the daemon's instance json database file is.

townsend2010:
Ok, I've updated this to make keys all lowercase and no whitespace.  Also, here is a sample output of what this does:

```
$ multipass info test-vm1 test-vm2 --format json
{
    "errors": [
    ],
    "info": {
        "test-vm1": {
            "image_hash": "20965d51ca17",
            "ipv4": "",
            "mounts": [
                {
                    "target_path": "host_home",
                    "source_path": "/home/townsend"
                }
            ],
            "state": "STOPPED"
        },
        "test-vm2": {
            "image_hash": "9856a634c47e",
            "ipv4": "10.122.122.3",
            "mounts": [
            ],
            "state": "RUNNING"
        }
    }
}
```

```
$ multipass list --format json
{
    "list": [
        {
            "ipv4": "",
            "name": "test-vm1",
            "state": "STOPPED"
        },
        {
            "ipv4": "",
            "name": "test-cloud-init",
            "state": "STOPPED"
        },
        {
            "ipv4": "10.122.122.3",
            "name": "test-vm2",
            "state": "RUNNING"
        },
        {
            "ipv4": "",
            "name": "test-vm4",
            "state": "STOPPED"
        },
        {
            "ipv4": "",
            "name": "test-vm3",
            "state": "STOPPED"
        }
    ]
}
```

kalikiana:
@townsend2010 Nice! Just a nit-pick: I'd rather see the lists in both cases use something like "instances" rather than the names of the commands.

townsend2010:
Yeah, I prefer to use the "instances" keyword in both outputs.  @Saviq, any issues with that?

kalikiana:
@Saviq I might be over-looking something... but I'd expect that targets must be unique, but mutliple sources could be re-used, no?

townsend2010:
Regarding "instances" vs. "info"/"list", I'm for whatever is best for the consumer of the output.

Regarding "mounts" being a dict, yeah, I think that makes sense.  But as @kalikiana pointed out, it's the target path that is unique, so we can key on that. Also, I'll add the "gid_mappings"/"uid_mappings keys, but they will be empty for now since we are not passing that info from the daemon yet. An example snippet output would be like:

```
"mounts": {
    "target/path1": {
        "gid_mappings": [
        ],
        "source_path": "source/path",
        "uid_mappings": [
        ]
    },
    "target/path2": {
        "gid_mappings": [
        ],
        "source_path": "source/path",
        "uid_mappings": [
        ]
    }
}
```

What do ya think?

townsend2010:
What else is needed in order to get this merged?

Saviq:
One thing I can still see is IPs being strings, when it should be an array, no?

I'll get it reviewed first thing tomorrow, @kalikiana can you please comment on whether you're happy with the format above?

townsend2010:
Not sure I understand why IPv4 should be an array.

Saviq:
Because you can have multiple IPs per interface.

townsend2010:
Hmm, I guess it's possible someone would configure the same interface in the VM to have multiple IP addresses, but I'm not sure how we would discover the user defined IP address.  But at any rate, I can make "ipv4" an array and do something like:

```
"ipv4": [
    "10.122.122.3",
    "10.10.2.4"
]
```

townsend2010:
I guess 'list' should also use an array for ipv4 if we are doing that for 'info'...

Saviq:
We would discover it by doing `ip addr` inside the host. And yes, for `info`, too.

townsend2010:
Ack on how to check for additional IP addresses.  That will be done later:)

I've updated the branch to make "ipv4" an array of IP address strings for both 'info' and 'list'.

kalikiana:
Very happy with the updated format!

Saviq:
I'm OK adding it as-is, but agree with @albaguirre that we need to streamline this. Now we have `errors: []` in `info`, but not in list, and anyway the errors still come in as plain text on the console output.

PR #43, opened 2018-01-16
daemon: Poll for cloud-init to complete during launch. Fixes #41

### LINKED ISSUE #41: Need to wait for `cloud-init` to complete
If you try to interact with an instance before `cloud-init` completes, it might fail (say, `apt` still updating), or even worse we can interrupt some `cloud-init` steps and leave the instance in a bad state.

townsend2010:


Saviq:
I was thinking we'd _`grep`_ the console output for `cloud-init`, any reason against that?

townsend2010:
I'm really not a fan of scraping the console for information like this and would rather poll in a more generic way.

albaguirre:
On scrapping log vs checking for the done file, I think the latter is easier and less intrusive overall, specially if in the future we end up using the "phone_home" module instead.

townsend2010:
@Saviq,
@albaguirre and @gerboland are also not in favor of implementing console scraping unless it's a last resort.  I believe we should proceed with this direction:)

Saviq:
Fine by me.

PR #44, opened 2018-01-17
ignore shell aliases when determining instances for autocomplete

jameinel:
This changes the calls to "grep" to use the unaliased form. "\grep"
This helps me because I tend to "alias grep='grep --color -nI'" which ends up putting the line numbers in front of the instance list. By using '\grep' it avoids my aliases.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -19,9 +19,9 @@ _multipass_complete()
         local state=$1
 
         local cmd="multipass list"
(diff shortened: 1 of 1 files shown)


jameinel:
I would think so, given that they all could be potentially aliased, and we
aren't using the /PATH prefixes.

jameinel:
I'm not super familiar with best practices around writing bash completion
scripts, though.

PR #60, opened 2018-01-23
mount: Sanitize target paths based on specific directories. Fixes #17

### LINKED ISSUE #17: `multipass mount` allows mounting over instance rootfs
`multipass mount $PWD instance:/` is accepted (although it did crash on me once), `sshfs` is running inside the instance, but rootfs is fortunately safe.

When it did mount, trying to `multipass umount instance` sent the daemon into a loop somewhere.

Still, we should sanitize the target path somewhat.

townsend2010:


src/utils/utils.cpp
```diff
@@ -51,6 +51,13 @@ bool mp::utils::valid_hostname(const QString& name_string)
     return matcher.exactMatch(name_string);
 }
 
+bool mp::utils::valid_target_path(const QString& target_path)
+{
+    QRegExp matcher("^(/|/dev/*.*|/proc/*.*|/sys/*.*|/home/ubuntu/*)");
```
Saviq: Do we not care that this is easy to cheat (even by accident) by using double slashes?

`multipass mount $PWD //`

Use Qt 5's [`QRegularExpression`](https://doc.qt.io/qt-5.9/qregularexpression.html) instead?

townsend2010:
Ok, updated with better matching and tests:)

townsend2010:
@Saviq, are you good with the recent changes?

PR #66, opened 2018-01-24
qemu: Set a 5GB virtual disk size on launch. Fixes #29

### LINKED ISSUE #29: VM is too small
Within a few minutes of sparking up a default 16.04 VM I'd run out of space. I think 2GB is a little too small to be actually useful. Perhaps 5G should be a default?

townsend2010:


townsend2010:
Yeah, if cloud images grow, so will the default disk size as well.

PR #68, opened 2018-01-24
qemu: Catch if the qemu process is not running during start

### LINKED ISSUE #64: Unexpected qemu process exit is not relayed back to the client during launch & start
I wanted to run a new instance on a VPS.

    # multipass launch --name=foo
    failed to launch: failed to determine IP address

Running qemu by hand gives me this (command copied from `ps -flA | grep qemu` on another machine where an instance of the same name was running fine):

    (qemu) qemu-system-x86_64: cannot set up guest memory 'pc.ram': Cannot allocate memory

Increasing the memory of the system fixed it.

townsend2010:
Catch if the qemu process is not running during IP discovery and wait for ssh. Fixes #64

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -25,6 +25,7 @@
 #include <multipass/name_generator.h>
 #include <multipass/query.h>
 #include <multipass/ssh/ssh_session.h>
+#include <multipass/start_exception.h>
 #include <multipass/utils.h>
 #include <multipass/version.h>
 #include <multipass/virtual_machine.h>
@@ -492,6 +493,18 @@ try // clang-format on
     server->Write(reply);
 
     return grpc::Status::OK;
+}
+catch (const mp::StartException& e)
+{
+    auto name = e.name();
+
+    config->factory->remove_resources_for(name);
+    config->vault->remove(name);
+    vm_instance_specs.erase(name);
+    vm_instances.erase(name);
+    persist_instances();
+
+    return grpc::Status(grpc::StatusCode::ABORTED, e.what(), "");
 }
 catch (const std::exception& e)
 {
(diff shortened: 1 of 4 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -212,6 +213,10 @@ std::string mp::QemuVirtualMachine::ipv4()
         while (std::chrono::steady_clock::now() < deadline)
         {
             QCoreApplication::processEvents();
+
+            if (vm_process->state() == QProcess::NotRunning)
```
Saviq: Can we please have a FAILED machine state? Pretty please? :D

And then we can listen on the `QProcess` status events :)

townsend2010: Yeah, a FAILED state will make sense for ``start``, but for ``launch``, I'm not so sure.  If we get a failure during ``launch``, we should report the issue back to the client and delete any trace of the instance from the daemon's perspective.

Also, in this particular case, ``qemu-system-x86_64`` just exits, so there in no QProcess error per se.  And we already have a connection to the ``finished`` signal, so we would have to determine in that handler if the process exited because ``stop`` was issued or for some other reasons.

I agree that looking for ``NotRunning`` while polling for various stages during ``launch``/``start`` is not the best, but I think we'll have to re-architecture some things to achieve an event based way of catching things like this.

src/daemon/daemon.cpp
```diff
@@ -493,6 +494,11 @@ try // clang-format on
 
     return grpc::Status::OK;
 }
+catch (const mp::StartException& e)
+{
+    vm_instances.erase(e.name());
```
albaguirre: I think this to also do:

config->factory->remove_resources_for(name);
config->vault->remove(name);
vm_instance_specs.erase(name);

PR #72, opened 2018-01-25
rpc: Use a Unix domain socket instead of localhost.

townsend2010:


src/daemon/daemon_rpc.cpp
```diff
@@ -59,6 +62,11 @@ auto make_server(const std::string& server_address, multipass::Rpc::Service* ser
     if (server == nullptr)
         throw std::runtime_error("Failed to start multipass gRPC service at " + server_address);
 
+    auto group = getgrnam("adm");
```
albaguirre: We need our own group here, like "multipass".

src/daemon/daemon_rpc.cpp
```diff
@@ -59,6 +62,11 @@ auto make_server(const std::string& server_address, multipass::Rpc::Service* ser
...
+    auto socket_path = QString::fromStdString(server_address).section("unix:", 1, 1).toStdString();
+    chown(socket_path.c_str(), 0, group->gr_gid);
+    chmod(socket_path.c_str(), S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
```
albaguirre: We need to check the return value here... if we failed to change the permissions of the socket, it's pretty much useless, so I would consider this a fatal condition - it needs to throw a std::runtime_error

src/daemon/daemon_rpc.cpp
townsend2010: I'm not really sure.  I was going by what @Saviq specified in the Trello card.

albaguirre: OK, short term adm is fine.

src/daemon/daemon_rpc.cpp
```diff
@@ -59,6 +62,17 @@ auto make_server(const std::string& server_address, multipass::Rpc::Service* ser
...
+    if (!group)
+        throw std::runtime_error("Could not determine group id for 'adm'.");
+
+    auto socket_path = QString::fromStdString(server_address).section("unix:", 1, 1).toStdString();
+    if (chown(socket_path.c_str(), 0, group->gr_gid) == -1)
+        throw std::runtime_error("Could not set ownership of the multipass socket.");
+
+    if (chmod(socket_path.c_str(), S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP) == -1)
+        throw std::runtime_error("Could not set permissions for the multipass socket.");
+
```
albaguirre: This section can go in daemon_main.cpp, but since it's leaking the fact that the server address is a socket...it belongs somewhere in platform.

townsend2010: Yeah, I was thinking about that last night:)  I'll work on moving this section.

src/daemon/daemon_main.cpp
```diff
@@ -47,6 +49,20 @@ sigset_t make_and_block_signals(std::vector<int> sigs)
     return sigset;
 }
 
+void set_server_permissions(const std::string& server_address)
+{
+    auto group = getgrnam("adm");
```
albaguirre: I think it should check if the server_address starts with "unix:" before attempting the rest.

PR #76, opened 2018-01-26
sshfs_mount: Fix issue with rename in the sftp server. Fixes #71

### LINKED ISSUE #71: some actions on mounted directory fail with permission denied
I created a directory on host and mounted it to VM
`multipass mount remote:/home/marian/VMs/multipass/git Mate-17-10:git`

when I connect to VM, I can create files and directories in mounted dir, but git clone fails

```
Cloning into 'mate-applets'...
error: could not write config file /home/ubuntu/git/mate-applets/.git/config: Operation not permitted
fatal: could not set 'core.filemode' to 'true'

```

When I move outside the mounted directory, git clone works.

When I return back to host and run git clone into that directory, it works.

Going back to VM, I can see cloned dir, I start ./autogen.sh. It starts to work, creates things as expected, but then fails with 
`./autogen.sh: 438: /usr/bin/mate-autogen: ./configure: Permission denied`

Maybe exec flag on mount missing?
It is mounted as 

`:/home/marian/VMs/multipass/git on /home/ubuntu/git type fuse.sshfs (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other)`

townsend2010:
When handling a rename from sshfs and the new path already exists, then remove
the existing file first before calling rename().

### DIFF
--- a/src/sshfs_mount/sshfs_mount.cpp
+++ b/src/sshfs_mount/sshfs_mount.cpp
@@ -632,6 +632,15 @@ private:
 
     void handle_rename(sftp_client_message msg)
     {
+        if (QFile::exists(ssh_string_get_char(msg->data)))
+        {
+            if (!QFile::remove(ssh_string_get_char(msg->data)))
+            {
+                reply_status(msg);
+                return;
+            }
+        }
+
         if (!QFile::rename(sanitize_path_name(QString(msg->filename)), ssh_string_get_char(msg->data)))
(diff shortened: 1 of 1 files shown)


townsend2010:
I looked at how rename() is implemented in POSIX wrt to if the new file already exists and discovered that it will call unlink() first before calling link().

This is really no different than what the proposed code will do.

PR #86, opened 2018-01-31
Avoid std::endl

albaguirre:


gerboland:
Why? Is avoiding a flush really that important? Now I worry if an error message mightn't be printed to user's console when it should be.

PR #89, opened 2018-02-01
sshfs_mount: Add new mount exception instead of throwing a tuple

townsend2010:


albaguirre:
Looks like an unrelated intermittent test failure --^. So merging...

PR #93, opened 2018-02-01
Sshfsmount cleanup

albaguirre:


albaguirre:
It's probably easier to review by looking at each commit diff

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -341,24 +341,39 @@ class SftpServer
 
         attr.size = file_info.size();
 
-        auto map = uid_map.find(file_info.ownerId());
-        if (map != uid_map.end())
+        const auto no_id_info_available = static_cast<uint>(-2);
```
townsend2010: Would making this a constexpr declared in the anonymous namespace be more efficient?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -582,14 +582,11 @@ class SftpServer
             return;
         }
 
-        auto len = msg->len;
-        if (len > (2 << 15))
-        {
-            /* 32000 */
-            len = 2 << 15;
-        }
+        const auto max_packet_size = 65536u;
```
townsend2010: Maybe a constexpr for this too?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -616,27 +616,26 @@ class SftpServer
...
+        {
+            sftp_reply_status(msg, SSH_FX_EOF, nullptr);
+            return;
+        }
 
-        // Send at most 50 directory entries to avoid ssh packet size overflow
-        while (!handle->entry_list.isEmpty() && count < 50)
+        auto& dir_entries = handle->entry_list;
+
+        const auto max_num_entries_per_packet = 50;
```
townsend2010: Another constexpr candidate?

src/sshfs_mount/sshfs_mount.cpp
albaguirre: Using https://godbolt.org/, it doesn't seem like it makes a difference.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -401,30 +391,31 @@ class SftpServer
 
     void handle_close(sftp_client_message msg)
     {
-        auto handle = handle_from_msg(msg);
-        if (!handle)
+        const auto id = sftp_handle(msg->sftp, msg->handle);
+
+        auto erased = open_file_handles.erase(id);
+        erased += open_dir_handles.erase(id);
```
townsend2010: Is there any way to tell what type (file or dir) is being closed instead of just calling erase on both unordered_maps?  Or would two erase() calls be just a quick as figuring out the type and then calling erase() on the particular unordered_map?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -142,10 +155,30 @@ QString longname_from(const QFileInfo& file_info)
...
+        out |= QFileDevice::ExeGroup;
+    if (perms & Permissions::read_other)
+        out |= QFileDevice::ReadOther;
+    if (perms & Permissions::write_other)
+        out |= QFileDevice::WriteOther;
+    if (perms & Permissions::exec_other)
+        out |= QFileDevice::ExeOther;
+
+    return out;
 }
```
townsend2010: On line 401, there is a call to go the other way- from Qt to the POSIX octal format.  We might as well do the explicit mapping for it as well.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Doing an assignment each time we enter that function vs. setting the no_id_info_available at compile is the same efficiency?  It just seemed like a candidate for global definition that is set one time.

albaguirre: Yeah, looking at the assembly output, even without optimization, the compiler just uses a constant value, no storage allocated for it.

albaguirre: Oh actually, I take that back. It's under -O1 and above that it doesn't make a difference, with no optimizations yeah it would be more efficient to change it.

src/sshfs_mount/sshfs_mount.cpp
albaguirre: Yeah you would still have to lookup the id in one map then the other to figure out which type it was. So same amount of lookups.

PR #96, opened 2018-02-04
Point libssh submodule to patched fork.

### LINKED ISSUE #45: `exec` does not exit on EOF
It's possible to pipe stdin into `multipass exec`, but EOF there isn't passed through?

This works:
```
⟫ multipass exec instance tee foo
bar
^D
```

But this hangs:
```
⟫ echo bar | multipass exec instance tee foo
```

albaguirre:
Fixes #45

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,4 +8,4 @@
 	ignore = dirty
 [submodule "3rd-party/libssh/libssh"]
 	path = 3rd-party/libssh/libssh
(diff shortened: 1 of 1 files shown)


albaguirre:
Make sure you issue a sync so the URL is updated on an existing repository:
 ```
 git submodule sync
 git submodule update --init --recursive
```

townsend2010:
Yeah, seems the libssh devs are not very responsive:(

PR #97, opened 2018-02-04
Fix build issues with clang 7.x

albaguirre:


src/client/cmd/list.cpp
```diff
@@ -41,13 +41,13 @@ std::string status_string(const multipass::ListVMInstance_Status& status)
...
-    case mp::InfoReply::RUNNING:
+    case mp::ListVMInstance_Status_RUNNING:
         status_val = "RUNNING";
         break;
-    case mp::InfoReply::STOPPED:
+    case mp::ListVMInstance_Status_STOPPED:
         status_val = "STOPPED";
         break;
-    case mp::InfoReply::TRASHED:
+    case mp::ListVMInstance_Status_TRASHED:
```
townsend2010: I'm not sure I understand why these changes fix clang build issues.

albaguirre: The error spew:

 error: comparison of two values with different enumeration types in switch statement ('multipass::ListVMInstance_Status' and 'const multipass::InfoReply::Status' (aka 'const multipass::InfoReply_Status'))

townsend2010: An, ok, cool:)

PR #102, opened 2018-02-05
info: Add an '--all' option to list info for all instances. Fixes #73

### LINKED ISSUE #73: feature request: add --all option to info
It would be nice if multipass mount without arguments could show a list of all defined mounts. Plus also show if it is an active mount with running instance or an inactive mount with stopped instance.

townsend2010:


townsend2010:
Oh, right.  Added now:)

Issue #105, opened 2018-02-07
mount command hangs when mounting a directory that has double quotes and spaces


albaguirre:
`mkdir /tmp/dquote\"\ spaces`
`bin/multipass mount /tmp/dquotes\"\ spaces foo:dquotes_and_spaces`

The command hangs as does the multipassd daemon.

Issue #104, opened 2018-02-06
mount command does not support directories with single or double quotes


albaguirre:
`mkdir /tmp/dquote\"`
`multipass mount /tmp/dquote\" foo:dquote`
```
mount failed: The following errors occured:
error mounting "dquotes": bash: -c: line 0: unexpected EOF while looking for matching `"'
bash: -c: line 1: syntax error: unexpected end of file
```
`mkdir /tmp/single\'`
`multipass mount /tmp/single\' foo:single`
```
mount failed: The following errors occured:
error mounting "single": bash: -c: line 0: unexpected EOF while looking for matching `''
bash: -c: line 1: syntax error: unexpected end of file
```

PR #106, opened 2018-02-07
Simplify sshfsmount

albaguirre:


albaguirre:
For issues #104 and #105 and I tested using [this script](https://pastebin.ubuntu.com/26533166/), which tests for single/double quotes and space combinations.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -830,7 +830,7 @@ mp::SshfsMount::SshfsMount(std::function<SSHSession()> make_session, const std::
                            const std::string& target, const std::unordered_map<int, int>& gid_map,
                            const std::unordered_map<int, int>& uid_map, std::ostream& cout)
     : ssh_session{make_session()},
-      sshfs_process{create_sshfs_process(ssh_session, source, target)},
+      sshfs_process{create_sshfs_process(ssh_session, mp::utils::escape_char(source, '"'), target)},
```
townsend2010: We should also escape the target since it's possible someone would do:

``bin/multipass mount /tmp/dquotes\"\ spaces foo``

which leads to the same bash failure.

albaguirre: Oh right! done.

PR #112, opened 2018-02-08
Use output interface

townsend2010:


include/multipass/cli/formatted_output.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef MULTIPASS_FORMATTED_OUTPUT_H
+#define MULTIPASS_FORMATTED_OUTPUT_H
+
+#include <multipass/rpc/multipass.grpc.pb.h>
+
+#include <string>
+
+namespace multipass
+{
+class FormattedOutput
```
albaguirre: I tend to avoid -er names but given the interface methods below, this seems more like a "Formatter" than something holding formatted output.

With a name like FormattedOutput I would expect something like

```
class FormattedOutput
{
public:
    virtual void write(const InfoReply& reply) = 0;
    virtual void write(const ListReply& reply) = 0;
    virtual std::string str() const = 0;
};
```
Then with an operator << overload:
```
std::ostream& operator<<(std::ostream& out, const mp::FormattedOutput& f)
{
    out << f.str();
    return out;
}
```

include/multipass/cli/formatted_output.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+    virtual ~FormattedOutput() = default;
+    virtual std::string process_info(InfoReply& reply) const = 0;
+    virtual std::string process_list(ListReply& reply) const = 0;
```
albaguirre: Maybe just overloads? Also const missing on the input arguments. Something like:

```
virtual std::string format(const InfoReply& reply) const = 0;
virtual std::string format(const ListReply& reply) const = 0;
```

include/multipass/cli/json_output.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#ifndef MULTIPASS_JSON_OUTPUT
+#define MULTIPASS_JSON_OUTPUT
+
+#include <multipass/cli/formatted_output.h>
+
+namespace multipass
+{
+class JsonOutput final : public FormattedOutput
```
albaguirre: JsonFormatter?

include/multipass/cli/table_output.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#ifndef MULTIPASS_TABLE_OUTPUT
+#define MULTIPASS_TABLE_OUTPUT
+
+#include <multipass/cli/formatted_output.h>
+
+namespace multipass
+{
+class TableOutput final : public FormattedOutput
```
albaguirre: TableFormatter?

src/client/cmd/info.cpp
```diff
@@ -149,22 +36,20 @@ mp::ReturnCode cmd::Info::run(mp::ArgParser* parser)
     }
 
     auto on_success = [this](mp::InfoReply& reply) {
+        std::string output;
+
         if (format_type == "json")
         {
-            cout << toJson(reply).toStdString();
+            JsonOutput json_output;
+            output = json_output.process_info(reply);
```
albaguirre: Now that there's a proper interface, multipass::Client should have a member variable map that holds all the different Formatters.

`std::map<std::string, std::unique_ptr<mp::Formatter>> formatters;`

`std::map` because we want ordered entries so that the "format" option help can be dynamically generated from this table, where the first entry is the default.

The the mp::Client constructor can instantiate it as:

```
{{"table", std::make_unique<TableFormatter>,
{"json", std::make_unique<JsonFormatter>}}
```

Then Command can hold a reference to the map.
A utility method in Command can validate the format option from the map, which can be called by the concrete command at parsing time, something like:

```
Formatter* Command::formatter_for(std::string format)
{
   auto entry = formatters.find(format);
   if (entry != formatters.end())
       return entry.second;
   return nullptr;
}
```
In cmd::Info::parse_args(...) then:

```
chosen_formatter = formatter_for(format);
if (chosen_formatter == nullptr)
{
    cerr << "Invalid format type given.\n";
    return ParseCode::CommandLineError;
}
```

Finally, this section would just simply do:

`cout << chosen_formatter.format(reply);`

PR #117, opened 2018-02-09
Add sftpserver tests

albaguirre:


albaguirre:
More tests for Sftpserver will come in a different PR.

PR #126, opened 2018-02-13
url_downloader: Set a 10 sec. timeout when downloading files

### LINKED ISSUE #108: Launch stuck after switching access point and subsequently claiming network be disabled
I was using `multipass launch` to spin up an instance. While it was doing its thing the WiFi access point I was on lost Internet connectivity, which prompted me to switch to another one. The on-going launch got stuck. So I cancelled it and tried again. From this point on I'm getting `failed to launch: Network access is disabled.` even though connectivity is fine now.

This is on the beta channel which right now is 2017.2.3 aka revision 111.

townsend2010:
QNetworkAccessManager has no built in QNetworkReply timeout, so need to use a
QTimer to do the dirty work.

Fixes #108

### DIFF
--- a/src/client/cmd/animated_spinner.cpp
+++ b/src/client/cmd/animated_spinner.cpp
@@ -64,8 +64,8 @@ void mp::AnimatedSpinner::stop()
         lock.unlock();
         if (t.joinable())
             t.join();
-        clear_line(cout);
     }
+    clear_line(cout);
 }
 
 void mp::AnimatedSpinner::draw()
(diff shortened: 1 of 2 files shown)


townsend2010:
Not sure why we need to expose the timeout externally.  Needing to set the timeout greater than 10 seconds would imply you have latencies over 10 seconds which, to me, says you have bigger issues than your downloads timing out:)

But if we do need to make this configurable, then yeah, let's tackle that later.

Issue #115, opened 2018-02-09
multipass shell should be a login shell


simos:
You can get a shell in multipass with

```
$ multipass exec myvm bash
ubuntu@myvm:~$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
ubuntu@myvm:~$ snap list
Name         Version  Rev   Developer  Notes
core         16-2.30  3887  canonical  core
hello-world  6.3      27    canonical  -
ubuntu@myvm:~$ hello-world
hello-world: command not found
ubuntu@myvm:~$ /snap/bin/hello-world
Hello World!
ubuntu@myvm:~$ 

```

The issue is that `multipass exec myvm bash` does not give you a login shell. A login shell is required so that /etc/profile.d/apps-bin-path.sh runs, which would set correctly the $PATH and add /snap/bin to it.

# Workaround

The workaround is to run instead the following (add -- and then --login).

```
$ multipass exec myvm -- bash --login
ubuntu@myvm:~$ echo $PATH
/home/ubuntu/bin:/home/ubuntu/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
ubuntu@myvm:~$ 
```

# What should happen? 

Ideally, there should be a dedicated command to get a login shell to the VM.
I suggest

```
$ multipass shell go
ubuntu@myvm:~$ echo $PATH
/home/ubuntu/bin:/home/ubuntu/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
ubuntu@myvm:~$ 

```

Having a `shell` subcommand would make it similar to LXD. In LXD it gives you a root login shell because LXD supports many different distributions that may not have a default non-root account.
If `multipass` is meant to ever be only for Ubuntu images, then `multipass shell myvm` should be OK to give a non-root login shell.

# References

 - [x] https://github.com/lxc/lxd/issues/4233

Saviq:
We do have a `connect` command already:
```
$ multipass help connect
Usage: multipass connect [options] <name>
Open a prompt on the instance.
```
But we will indeed rename it to `multipass shell` to align with `lxd`.

PR #143, opened 2018-02-16
cli: rename `connect` to `shell` to align with lxd (Fixes: #115)

Saviq:
The `lxc` command line has a (non-advertised yet) `shell` command, let's
rename our `connect` to align. It's a better name as well.

I've left `connect` as an alias to not break people's muscle memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,5 +66,5 @@ sudo <multipass>/build/bin/multipassd &
 ## Connect to a running instance
 
 ```
-<multipass>/build/bin/multipass connect foo
+<multipass>/build/bin/multipass shell foo
 ```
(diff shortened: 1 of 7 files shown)


src/client/cmd/connect.cpp
```diff
@@ -52,35 +52,40 @@ mp::ReturnCode cmd::Connect::run(mp::ArgParser* parser)
...
     return dispatch(&RpcMethod::ssh_info, request, on_success, on_failure);
 }
 
-std::string cmd::Connect::name() const { return "connect"; }
+std::string cmd::Connect::name() const { return "shell"; }
+
+std::vector<std::string> cmd::Connect::aliases() const
+{
+    return {name(), "sh", "connect"};
+}
```
gerboland: Strange that Connect's name is "shell" and that "connect" is only an alias. I know it's avoiding code churn, but I'd prefer we stick to naming classes after their primary command name.

townsend2010:
Well, we didn't rename our classes when we switched from 'trash' to 'delete' and from 'empty-trash' to 'purge'.  My opinion is if we want to rename our files and classes to match the actual command, that should come as a separate PR.

PR #148, opened 2018-02-21
Rename classes to match cmd names

albaguirre:
Depends on #147

Use [this diff view instead](https://github.com/albaguirre/multipass/compare/fix-daemon-tests...rename-classes-to-match-cmds) before #147 is merged

### DIFF
--- a/src/client/cmd/delete.h
+++ b/src/client/cmd/delete.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Canonical, Ltd.
+ * Copyright (C) 2017-2018 Canonical, Ltd.
  *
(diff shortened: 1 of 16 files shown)


src/client/cmd/delete.h
```diff
@@ -13,8 +13,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
- *
  */
 
 #ifndef MULTIPASS_TRASH_H
```
townsend2010: Probably should rename to MULITPASS_DELETE_H and fix up the #endif comment too,

PR #149, opened 2018-02-21
[daemon] Make the 5GB default daemon-wide

Saviq:


src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -277,16 +277,10 @@ mp::VMImage mp::QemuVirtualMachineFactory::prepare_source_image(const mp::VMImag
...
-    // Default to a 5GB virtual disk
-    QString disk_size{"5G"};
+    auto disk_size = QString::fromStdString(desc.disk_space);
 
-    if (!desc.disk_space.empty())
-    {
-        disk_size = QString::fromStdString(desc.disk_space);
-
-        if (disk_size.endsWith("B"))
-            disk_size.chop(1);
```
gerboland: This looks important, sure it should go?

Saviq: Look below? desc.disk_space will never be empty with the change above.

gerboland: Oh yes, there it is. Works for me anyway

PR #164, opened 2018-02-22
bash: Fix various bugs in bash completion

### LINKED ISSUE #152: `--format` does not autocomplete
There's no autocompletion for the new `--format` option.

### LINKED ISSUE #153: Bash completion does not include short versions or aliases
The following are examples of what breaks autocompletion:
`multipass delete -p<Tab>`
`multipass sh <Tab>`

Both are valid multipass commands, but don't include full versions of the options / commands, which results in interrupted completion.

### LINKED ISSUE #157: Mount completion finishes on non-empty folders
```
$ multipass mount /<Tab><Tab>
/bin    /dev    /home   /lib32  /lxd    /mnt    /proc   /run    /snap   /sys    /usr
/boot   /etc    /lib    /lib64  /media  /opt    /root   /sbin   /srv    /tmp    /var
$ multipass mount /hom<Tab>
$ multipass mount /home <Tab>
```
In the last step, note there's a space after `/home`, meaning the completion for the source path finished.

I wonder, do we really need the bash completion script to be a script? Couldn't the `multipass` command return valid options and pass it up to bash via a wrapper? It feels like encoding all this logic inside a bash script is bound to result in such issues, getting out of sync etc.

### LINKED ISSUE #163: For some commands, bash completion shows the same instance name multiple times
For example, if one instance is defined and it's name is `foo` and then you do:

`$ multipass start <Tab><Tab>`
`$ mulitpass start foo <Tab>`
`$ multipass start foo foo <Tab>`

and so on.

townsend2010:
Add the `--format` option for list and info. Fixes #152.
Autocomplete the supported types for `--format`/`-f`.
Fix mount source completion on non-empty folders. Fixes #157.
Support option autocompletion for command aliases. Fixes #153.
Show instance name only once when it has been selected in the command line. Fixes #163.
Autocomplete instance names for the `delete` command.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -1,4 +1,4 @@
-# Copyright © 2017 Canonical Ltd.
+# Copyright © 2017-2018 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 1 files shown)


completions/bash/multipass
```diff
@@ -12,6 +12,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+. /etc/bash_completion
+
```
albaguirre: This will cause an infinite recursion when opening a new terminal

townsend2010:
Oops, didn't mean to leave that in there.  Fixed:)

PR #170, opened 2018-02-26
Update cli

albaguirre:


albaguirre:
@Saviq Are you ok with the wording for mount and find?

src/client/cmd/find.cpp
```diff
@@ -118,13 +118,18 @@ QString cmd::Find::short_help() const
...
 
 mp::ParseCode cmd::Find::parse_args(mp::ArgParser* parser)
 {
-    parser->addPositionalArgument("string", "Optionally find images matching this string", "[<remote:>]<string>");
+    parser->addPositionalArgument("string",
+                                  "The value to search for in [<type:>]<string> format, where "
+                                  "<type> can be either ‘release’ or ‘daily’. If <type> is omitted, "
+                                  "it defaults to ‘release’. <string> can be a partial image hash or an "
+                                  "ubuntu release version, codename or alias",
+                                  "[<string>]");
```
Saviq: I think we need to do better to explain that you don't have to provide `<string>`.

albaguirre: @Saviq, what about "An optional value to search for in ...."

Saviq: +1

albaguirre: @Saviq ok fixed.

PR #186, opened 2018-03-08
copy-files: Add support for copying files between instances and host

townsend2010:
I'm going to do a follow up PR for correct permissions on files.


albaguirre:
LGTM

PR #194, opened 2018-03-21
tests: Add some tests for the ssh client code

townsend2010:


src/platform/console/unix_console.cpp
```diff
@@ -70,6 +70,8 @@ void mp::UnixConsole::setup_console()
         saved_terminal = terminal_local;
         cfmakeraw(&terminal_local);
         tcsetattr(fileno(stdin), TCSANOW, &terminal_local);
+
+        setup_complete = true;
```
albaguirre: Looking at the usage of console, i.e.:

```
    if (console->is_interactive())
    {
        console->setup_console();
    }
```
I think that's doing "ask then do", instead of ["Tell, don't ask"](https://pragprog.com/articles/tell-dont-ask)

Does the console really need to be setup after ssh_channel_request_shell/ssh_channel_request_exec? I don't see an immediate reason for that. If not required, then, the Console class on construction should setup the console; that makes the Console class invariant and consequently you don't need the "setup_complete" boolean.

townsend2010:
@albaguirre,

Ahhh, yeah, you're right.  Not sure why I did that before, but probably just some left over cruft from when I was originally working on this and trying to make it work:)

At any rate, I updated the PR to reflect this.

PR #197, opened 2018-03-22
console: Change ssh pty size from within the console class

townsend2010:
This avoids a delay in resizing the output when changing the size of
an interactive terminal.

### DIFF
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Canonical, Ltd.
+ * Copyright (C) 2017-2018 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -17,9 +17,9 @@
 
 #include "unix_console.h"
 
-#include <atomic>
+#include <multipass/auto_join_thread.h>
+#include <multipass/platform_unix.h>
 
-#include <signal.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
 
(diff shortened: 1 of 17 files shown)


src/platform/console/unix_console.cpp
```diff
@@ -27,33 +25,39 @@ namespace mp = multipass;
...
+void change_ssh_pty_size(ssh_channel channel)
 {
-    bool old = value;
-    while (old && !value.compare_exchange_weak(old, false))
-    {
-    }
-    return old;
+    struct winsize win = {0, 0, 0, 0};
+
+    ioctl(fileno(stdout), TIOCGWINSZ, &win);
```
albaguirre: is TIOCGWINSZ signal safe?

src/platform/console/unix_console.cpp
```diff
@@ -27,33 +25,39 @@ namespace mp = multipass;
...
+    ssh_channel_change_pty_size(channel, win.ws_col, win.ws_row);
```
albaguirre: This is not signal safe.
ssh_channel_change_pty_size can allocate.

Unfortunately, signal handlers are very restrained on what they can do without deadlocks or other concurrency issues.

The best you can do is avoid them all together and use a thread, see for example UnixSignalHandler in daemon_main.cpp which uses a combination of sigaddset, sigprocmask and sigwait to avoiding signal handlers entirely.

src/platform/console/unix_console.h
```diff
@@ -36,16 +42,18 @@ class UnixConsole final : public Console
...
+
+    static void setup_environment();
 
 private:
     void setup_console() override;
     void restore_console() override;
 
     bool interactive{false};
     struct termios saved_terminal;
+    ssh_channel channel;
```
albaguirre: Clang complains that channel is unused:

> In file included from ../src/platform/console/unix_console.cpp:18:
../src/platform/console/unix_console.h:54:17: error: private field 'channel' is not used [-Werror,-Wunused-private-field]
    ssh_channel channel;

PR #201, opened 2018-03-29
[info][list]: Don't throw exceptions when image data cannot be retrieved

### LINKED ISSUE #199: Need to cache relevant image info when downloading an image
Recently, many older bionic daily images have disappeared from the cloud-images simplestreams database, so when multipass tries to look up image information for a particular instance via simplestreams, the information for that hash is missing, which causes multipass to throw an error and break `list` and `info`.

We need to do two things here:
1. Cache the relevant image info for the downloaded image and then copy that when an instance is created.
2. Still allow `list` and `info` to work and not bubble up the thrown exception and instead perhaps fill in the missing fields with something like "MISSING" or "NA".

townsend2010:
Partially fixes #199

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -609,7 +609,6 @@ try // clang-format on
         }
 
         auto info = response->add_info();
-        auto vm_image = fetch_image_for(name, config->factory->fetch_type(), *config->vault);
         auto& vm = it->second;
         info->set_name(name);
         if (deleted)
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -636,9 +635,16 @@ try // clang-format on
...
-        info->set_id(vm_image.id);
+        try
+        {
+            auto vm_image = fetch_image_for(name, config->factory->fetch_type(), *config->vault);
+            auto vm_image_info = config->image_host->info_for_full_hash(vm_image.id);
+            info->set_image_release(vm_image_info.release_title.toStdString());
+            info->set_id(vm_image.id);
+        }
+        catch (const std::exception& e)
+        {
```
albaguirre: Can you add an error log here in the catch?

src/daemon/daemon.cpp
```diff
@@ -722,9 +728,15 @@ try // clang-format on
...
-        auto vm_image_info = config->image_host->info_for_full_hash(vm_image.id);
-        entry->set_current_release("Ubuntu " + vm_image_info.release_title.toStdString());
+        try
+        {
+            auto vm_image = fetch_image_for(name, config->factory->fetch_type(), *config->vault);
+            auto vm_image_info = config->image_host->info_for_full_hash(vm_image.id);
+            entry->set_current_release("Ubuntu " + vm_image_info.release_title.toStdString());
+        }
+        catch (const std::exception& e)
+        {
```
albaguirre: Can you add an error log here in the catch?

Issue #130, opened 2018-02-15
Daemon should re/store instance state on startup


townsend2010:
Stopping the daemon should store the instance state and restore it again on startup.

townsend2010:
I think once we move to using the libvirt backend by default and https://forum.snapcraft.io/t/process-lifecycle-on-snap-refresh/140/37 being fully implemented in snapd and snapcraft, we won't need to be concerned about saving instance state since the qemu process will survive through multipassd/libvirtd restarts.

As it stands right now, when the libvirt backend work is merged and using multipassd outside of snap, instance state will remain when multipassd is stopped or restarted.  For example, if an instance is running and multipassd is shut down, the instance will remain running and when multipassd starts again, it will detect via libvirtd that the instance is running and set state accordingly.

Saviq:
Well, `multipassd` restarting is one thing, how about boot?

townsend2010:
@Saviq,

Currently, for the QEMU driver, when multipassd receives `SIGTERM`, it will try to gracefully shut down running instances.  We do this because multipassd owns the QEMU QProcess's, so we try to avoid just yanking the VM away.  That said, if a system reboots, it sends `SIGTERM` to mulipassd, so the instances will be gracefully shutdown and thus, the running state is no longer the case.  Any opinions on how we should handle this for host reboots, etc?

townsend2010:
Oh, I know what we should do.  On multipassd shutdown, we should try to shutdown the instances gracefully, but the saved state shouldn't bu updated, so on the subsequent multipassd start, it will start up those particular instances automatically.

PR #212, opened 2018-04-17
Support libvirt backend

townsend2010:
To test with a local build:
`$ MULTIPASS_VM_BACKEND=LIBVIRT ./multipassd`

For a snap, after installing it, do:
`$ sudo snap set multipass vm-backend=LIBVIRT`

Also, when installed as a snap, `multipass.virsh` also works for administering the VM via virsh.

### DIFF
--- a/snap/bin/launch-multipassd
+++ b/snap/bin/launch-multipassd
@@ -2,5 +2,6 @@
 
 export http_proxy="$(snapctl get proxy.http)"
 export https_proxy="$(snapctl get proxy.https)"
+export MULTIPASS_VM_BACKEND="$(snapctl get vm-backend)"
 
 exec $SNAP/bin/multipassd
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,4 +1,19 @@
-#!/bin/sh -e
+#!/bin/bash -e
 
-# Just a placeholder, however validation of the proxy.{http,https}
-# configuration could be made here
+vm_backend="$(snapctl get vm-backend)"
+
+if [[ -n $vm_backend && ! $vm_backend =~ (LIBVIRT|QEMU) ]]; then
+    echo "\"$vm_backend\" is not a supported VM backend"
+    exit 1
+fi
+
+vm_backend_saved="$(snapctl get vm-backend-saved)"
+
+if [[ $vm_backend == "QEMU" && -z $vm_backend_saved ]]; then
+    exit 0
+fi
+
+if [[ $vm_backend != $vm_backend_saved ]]; then
+    snapctl set vm-backend-saved="$vm_backend"
+    snapctl restart $SNAP_NAME
+fi
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+
+install -D $SNAP/var/snap/multipass/common/libvirt/libvirtd.conf $SNAP_COMMON/libvirt/libvirtd.conf
+sed -i 's/unix_sock_group = "libvirtd"/unix_sock_group = "sudo"/' $SNAP_COMMON/libvirt/libvirtd.conf
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: multipass
-version: '2017.0.0'
+version: 'git'
 version-script: |
   git describe
 summary: Ubuntu at your fingertips
@@ -15,7 +15,7 @@ apps:
   multipassd:
     command: bin/launch-multipassd
     environment:
-      LD_LIBRARY_PATH: $SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu::$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu
+      LD_LIBRARY_PATH: $SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu:$SNAP/lib:$SNAP/usr/lib
       PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
       QT_PLUGIN_PATH: $SNAP/usr/lib/x86_64-linux-gnu/qt5/plugins
       XDG_DATA_HOME: $SNAP_COMMON/data
@@ -23,11 +23,23 @@ apps:
     daemon: simple
   multipass:
     environment:
-      LD_LIBRARY_PATH: $SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu::$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu
+      LD_LIBRARY_PATH: $SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu:$SNAP/lib:$SNAP/usr/lib
       PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
       QT_PLUGIN_PATH: $SNAP/usr/lib/x86_64-linux-gnu/qt5/plugins
     command: bin/multipass
     completer: etc/bash_completion.d/snap.multipass
+  libvirt-bin:
+    command: bin/launch-libvirtd
+    environment:
+      LD_LIBRARY_PATH: $SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu:$SNAP/lib:$SNAP/usr/lib
+      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
+      LC_ALL: C
+    daemon: simple
+  virsh:
+    command: bin/virsh
+    environment:
+      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
+      LC_ALL: C
 
 parts:
   qtbase5-dev:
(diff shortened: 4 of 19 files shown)


snap/bin/launch-multipassd
```diff
@@ -3,4 +3,12 @@
 export http_proxy="$(snapctl get proxy.http)"
 export https_proxy="$(snapctl get proxy.https)"
 
+vm_backend="$(snapctl get vm-backend)"
+
+if [ "$vm_backend" = libvirt ]; then
+    export MULTIPASS_VM_BACKEND=LIBVIRT
+elif [ "$vm_backend" = qemu ]; then
+    export MULTIPASS_VM_BACKEND=QEMU
+fi
```
Saviq: Shouldn't this fall back to qemu _unless_ it's "libvirt"?
What's the value on install?

snap/hooks/configure
```diff
@@ -1,4 +1,10 @@
...
-# Just a placeholder, however validation of the proxy.{http,https}
-# configuration could be made here
+vm_backend="$(snapctl get vm-backend)"
+
+if [[ -n $vm_backend && ! $vm_backend =~ (libvirt|qemu) ]]; then
+    echo "\"$vm_backend\" is not a supported VM backend"
+    exit 1
+fi
+
+snapctl restart $SNAP_NAME
```
Saviq: Restart seems brutal, but I suppose there's nothing better we can do.

snap/hooks/install
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e
+
+mkdir -p $SNAP_COMMON/libvirt
+cp -r $SNAP/var/snap/multipass/common/libvirt/libvirtd.conf $SNAP_COMMON/libvirt
```
Saviq: I wonder if `install` would be better used here? `-r` doesn't seem necessary?

snap/hooks/install
```diff
@@ -0,0 +1,6 @@
...
+sed -i 's/unix_sock_group = \"libvirtd\"/unix_sock_group = \"sudo\"/' $SNAP_COMMON/libvirt/libvirtd.conf
```
Saviq: Should be no need to escape the double quotes?

Will we have to rely on `sudo` membership long-term?

snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: multipass
-version: '2017.0.0'
+version: '2018.0.0'
```
Saviq: If you're touching it, do `git` instead, there's no point having any one version here.

snap/bin/launch-multipassd
townsend2010: On install, $vm_backend is not set and thus MULTIPASS_VM_BACKEND is not set, so in src/platform/platform_linux.cpp, this defaults to use qeum as the backend.

This is only used so we can use `snap set` to change back and forth after install.

snap/hooks/configure
townsend2010: Yep, we have to restart multipassd in order for the changed backend to take effect and this is the only way I know of on how to do that.

snap/hooks/install
townsend2010: Do you mean the `install` stanza in snapcraft.yaml?  If so, that is install when building the snap.  This is for the actual install of the snap where $SNAP_COMMON will be available.

About `-r`, you're right, that is left over cruft from when I was actually recursively copying directories here, which became unnecessary.  I'll fix that.

snap/hooks/install
townsend2010: I thought I ended up having to use the double quotes, but maybe that's not necessary.  It works with it, but I can try it without the escaped double quotes.

Regarding `sudo`, we need to put in some group here that will common on all distros we want to support.  We have to use system groups here since, and correct me if I'm wrong, snaps don't have a concept of separate groups.  That said, I noticed that on xenial, it uses `libvirtd` as the group, but on artful, it uses `libvirt`, so I had to change it to something both would support out of the box.  Also, Solus seems to be an issue wrt to groups, so I thought I would go with the same group we use for the multipass socket.

snap/snapcraft.yaml
townsend2010: I'm not sure how that would be accomplished. Could you please point to an example on how to do that?

snap/hooks/configure
Saviq: Fixing #130 will help with the UX of this.

snap/hooks/install
Saviq: Have a look at how the `lxd` snap solved this? They don't require group membership.

snap/bin/launch-multipassd
townsend2010: Well, yeah, I could do that and either use `snap set multipass vm-backend={LIBVIRT|QEMU}`or use the lower case `libvirt`/`qemu` and that would be the env var.

snap/snapcraft.yaml
townsend2010: Ah, ok, sure, I can do that:)

snap/hooks/install
townsend2010: lol, ok, didn't think of that, but sure:)

snap/hooks/install
townsend2010: This is needed for `multipass.virsh` to connect to the libvirt socket only.

snap/hooks/configure
townsend2010: Are you concerned that when someone wants to switch backends, for example, from qemu to libvirt and they have running VM's, that the VM's would automatically start using the new backend since they were running with the prior backend?

I added a comment to #130 about how that is going to be resolved, but that won't fix switching backends and I kind of think when switching backends, requiring an explicit start to instances should be required since it's a pretty drastic change in daemon and instance management.

snap/bin/launch-multipassd
townsend2010: I think I'm going to require `LIBVIRT`/`QEMU` on `snap set` and get rid of the lower case choices.  Does that work for you?

snap/snapcraft.yaml
townsend2010: Ok, fixed.

snap/hooks/install
townsend2010: Ok, uses `install` now.

snap/bin/launch-multipassd
townsend2010: Ok, uses the direct `snapctl get`\`export` now.

snap/snapcraft.yaml
```diff
@@ -126,21 +137,83 @@ parts:
...
+    - --without-storage-sheepdog
+    - --without-storage-rbd
+    - --without-storage-lvm
+    - --without-selinux
+    - --prefix=/snap/$SNAPCRAFT_PROJECT_NAME/current
+    - --localstatedir=/var/snap/$SNAPCRAFT_PROJECT_NAME/common
+    - --sysconfdir=/var/snap/$SNAPCRAFT_PROJECT_NAME/common
+    - DNSMASQ=/snap/$SNAPCRAFT_PROJECT_NAME/current/usr/sbin/dnsmasq
+    - DMIDECODE=/snap/$SNAPCRAFT_PROJECT_NAME/current/usr/sbin/dmidecode
+    prepare: |
```
Saviq: `snapcraft` from the candidate channel shouts:

```
DEPRECATED: The 'install' keyword has been replaced by 'override-build'
See http://snapcraft.io/docs/deprecation-notices/dn9 for more information.
DEPRECATED: The 'prepare' keyword has been replaced by 'override-build'
See http://snapcraft.io/docs/deprecation-notices/dn7 for more information.
```

townsend2010: Hmm, those links are dead:(  And the official docs don't mention a thing about those being deprecated: https://docs.snapcraft.io/build-snaps/scriptlets

Also, doing a Google search of `snapcraft override-build` does not yield any results.

My opinion is to leave `prepare` and `install` there until `override-build` is official and well documented.

snap/snapcraft.yaml
```diff
@@ -126,21 +137,83 @@ parts:
...
+      apt update
+      apt source libvirt
```
Saviq: Oh, tricky. Are we sure we want "old" libvirt like that?

townsend2010: I figure we want to use the version of libvirt found in Xenial since that's the target distro version for built snaps.  It would be no different than listing `libvirt` in `stage-packages` accept I had to rebuild libvirt to account for the hard coded paths.

Saviq:
There's also this, when using QEMU backend:
```
kwi 18 15:28:00 michal-laptop multipass.multipassd[11262]: E0418 15:28:00.133570102   11262 uri_parser.cc:45]           bad uri.scheme: ''
kwi 18 15:28:00 michal-laptop multipass.multipassd[11262]: E0418 15:28:00.133636438   11262 uri_parser.cc:51]                            ^ here
kwi 18 15:28:00 michal-laptop multipass.multipassd[11262]: E0418 15:28:00.133641863   11262 http_proxy.cc:52]           cannot parse value of 'http_proxy' env var
```

We should also wrap libvirtd in a NOOP when the backend isn't LIBVIRT, wdyt?

townsend2010:
Those warning messages have been there for some time and are from grpc.  It's because the `hhtp_proxy` env var is being set willy nilly when launching multipassd.

As far as wrapping libvirtd in a no-op if the qemu backend is set, I can do that.

snap/snapcraft.yaml
Saviq: Well, snaps are supposed to be target agnostic, not "target xenial". Any case, my question was whether it would be more interesting to use newer libvirt? Fine, if not.

Saviq: Yeah it's heavily patched by our security team:
https://launchpad.net/ubuntu/+source/libvirt/1.3.1-1ubuntu10.19

townsend2010: Yes, Ubuntu patches libvirt quite a bit.  And my thinking is that I tried to just use libvirt in stage-packages, but the only problem was hard coded paths that are set during compile time.  So I'm basically taking what would be given in stage-packages and instead building it from package source with new paths that work in the snap...and maybe a little tweaking on what actually gets built.

albaguirre: OK, in that case, probably better to keep it as is.

PR #216, opened 2018-04-20
Save mac address (Fixes #214)

### LINKED ISSUE #214: qemu: bionic instances lose networking
In bionic, `cloud-init` stores the network MAC address for netplan and new interfaces don't get DHCP.

While this sounds like a possible cloud-init issue, we should keep the MAC stable.

townsend2010:


src/daemon/daemon.cpp
```diff
@@ -436,13 +448,27 @@ try // clang-format on
     auto user_data_cloud_init_config = YAML::Load(request->cloud_init_user_data());
     prepare_user_data(user_data_cloud_init_config, vendor_data_cloud_init_config);
     config->factory->configure(name, meta_data_cloud_init_config, vendor_data_cloud_init_config);
-    auto vm_desc = to_machine_desc(request, name, vm_image, meta_data_cloud_init_config, user_data_cloud_init_config,
-                                   vendor_data_cloud_init_config, *config->ssh_key_provider);
+
+    std::string mac_addr;
+    while (true)
```
Saviq: So we're putting the limit on the number of instances at 16M? ;D

townsend2010: lol, but I seriously doubt we really need to limit the while loop

PR #259, opened 2018-05-30
xz_decoder: Add an xz decoder for extracting compressed images

townsend2010:


src/xz_decoder/xz_image_decoder.cpp
```diff
@@ -0,0 +1,108 @@
...
+            throw std::runtime_error("not a .xz file");
+        case XZ_OPTIONS_ERROR:
+            throw std::runtime_error("unsupported options in the .xz headers");
+        case XZ_DATA_ERROR:
+        case XZ_BUF_ERROR:
+            throw std::runtime_error("xz file is corrupt");
+        default:
+            throw std::runtime_error("xz decoder unhandled error condition");
+        }
+    }
```
albaguirre: Shouldn't this check be done right after xz_dec_run?

Maybe xz_dec_run and this check sould be another function? like bool decode_chunk(...) that does xz_dec_run and returns true if the loop needs to continue, false if not and throws on error?

src/xz_decoder/xz_image_decoder.cpp
```diff
@@ -0,0 +1,108 @@
...
+        if (decode_buf.out_pos == max_size)
+        {
+            decoded_file.write(write_data.data(), decode_buf.out_pos);
+            decode_buf.out_pos = 0;
+        }
+
+        if (ret == XZ_OK)
+            continue;
+
+        decoded_file.write(write_data.data(), decode_buf.out_pos);
```
albaguirre: Couldn't this write and the one above be condensed into one? Assuming the error checking happens right after xz_dec_run like described in the other comment.

src/xz_decoder/xz_image_decoder.cpp
```diff
@@ -0,0 +1,108 @@
...
+        case XZ_OK:
+            break;
+        case XZ_STREAM_END:
+            return;
+        case XZ_MEM_ERROR:
+            throw std::runtime_error("xz decoder memory allocation failed");
+        case XZ_MEMLIMIT_ERROR:
+            throw std::runtime_error("xz decoder memory usage limit reached");
+        case XZ_FORMAT_ERROR:
+            throw std::runtime_error("not a .xz file");
```
albaguirre: I suppose we can remove the "." here no?

src/xz_decoder/xz_image_decoder.cpp
```diff
@@ -0,0 +1,108 @@
...
+        case XZ_OPTIONS_ERROR:
+            throw std::runtime_error("unsupported options in the .xz headers");
```
albaguirre: I suppose we can remove the "." here too no?

PR #263, opened 2018-06-01
daemon: Restore instance state when starting

### LINKED ISSUE #130: Daemon should re/store instance state on startup

townsend2010:
Persist the instance state in the DB and read this when the daemon starts. If instance
was running but is not currently, then start the instance.

Fixes #130

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Canonical, Ltd.
+ * Copyright (C) 2017-2018 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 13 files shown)


Saviq:
IIUC this also stores states like UNKNOWN and STARTING, is this really what we want to do?

Shouldn't instead we store a "desired state" that would get updated on user action (instance shutting down should be considered user action in this case, I suppose)?

townsend2010:
@Saviq,
I kind of think it's a good idea to store the actual state, but I see what you mean about other "running" states.  So instead, maybe I should check if the state is not `off` or `shutdown` instead.  All other states imply that the VM should at least be started.

Saviq:
Yeah maybe that's enough, if it's not stopped, it's started.

PR #277, opened 2018-06-12
image_vault: Verify the sha256 hash of images downloaded via SimpleSt…

### LINKED ISSUE #275: Verify image downloads if checksum is available
Need to verify image downloads if a checksum is available.  For SimpleStreams, the checksum happens to be the image id we use.  For the `--image` option, that will be a separate issue.

townsend2010:
…reams

Fixes #275

### DIFF
--- a/src/client/cmd/launch.cpp
+++ b/src/client/cmd/launch.cpp
@@ -78,7 +78,8 @@ mp::ReturnCode cmd::Launch::run(mp::ArgParser* parser)
             {LaunchProgress_ProgressTypes_IMAGE, "Retrieving image: "},
             {LaunchProgress_ProgressTypes_KERNEL, "Retrieving kernel image: "},
             {LaunchProgress_ProgressTypes_INITRD, "Retrieving initrd image: "},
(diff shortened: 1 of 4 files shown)


src/client/cmd/launch.cpp
```diff
@@ -78,7 +78,8 @@ mp::ReturnCode cmd::Launch::run(mp::ArgParser* parser)
             {LaunchProgress_ProgressTypes_IMAGE, "Retrieving image: "},
             {LaunchProgress_ProgressTypes_KERNEL, "Retrieving kernel image: "},
             {LaunchProgress_ProgressTypes_INITRD, "Retrieving initrd image: "},
-            {LaunchProgress_ProgressTypes_EXTRACT, "Extracting image: "}};
+            {LaunchProgress_ProgressTypes_EXTRACT, "Extracting image: "},
+            {LaunchProgress_ProgressTypes_HASH, "Computing image hash: "}};
```
albaguirre: Could this be "Verifying image" instead?

townsend2010:
@albaguirre, good idea.  I reworded that and also changed the enum value's name.

PR #280, opened 2018-06-14
Add certificate generator

albaguirre:


src/cert/ssl_cert_provider.cpp
```diff
@@ -0,0 +1,147 @@
...
+        throw std::runtime_error("Failed to allocate RSA structure");
+
+    std::unique_ptr<BIGNUM, BIGNUMDeleter> value{BN_new()};
+    if (value == nullptr)
+        throw std::runtime_error("Failed to allocate BIGNUM structure");
+
+    if (!BN_set_word(value.get(), RSA_F4))
+        throw std::runtime_error("Failed to set BIGNUM to RSA_F4");
+
+    if (!RSA_generate_key_ex(rsa.get(), 2048, value.get(), nullptr))
```
Saviq: What happens here when we don't have enough entropy?

tests/test_ssl_cert_provider.cpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include "src/cert/ssl_cert_provider.h"
+
+#include <gmock/gmock.h>
+
+namespace mp = multipass;
+using namespace testing;
+
+TEST(SSLCertProvider, creates_cert)
```
Saviq: Related to above - what if test machine doesn't have enough entropy?

albaguirre: As pointed above, it could block if entropy pool is not initialized. 

In practice, we have not hit this, as test_ssh_key_provider does something similar.

src/cert/ssl_cert_provider.cpp
albaguirre: I traced down to see what the source of entropy was...it essentially does a getrandom syscall, without any flags
http://man7.org/linux/man-pages/man2/getrandom.2.html

So potentially it could block only when the entropy pool is not initialized (i.e. /dev/urandom). 

This is similar to OpenSSHKeyProvider.

PR #289, opened 2018-06-20
snap: Revert libvirtd 'stop-mode: sigterm' daemon mode

townsend2010:
This is causing problems with other processes that are spawned by libvirtd.


Saviq:
Yeah, things can get weird indeed.

PR #324, opened 2018-07-31
metrics_provider: Add support for sending anonymous metrics

townsend2010:
Initial commit for adding anonymous KPI metrics support.  I'd like some feedback on what we are sending, the opt-in prompt message, and what condition sends it.


src/daemon/daemon.cpp
```diff
@@ -432,6 +479,26 @@ grpc::Status mp::Daemon::launch(grpc::ServerContext* context, const LaunchReques
...
+        metrics_opt_in = -1;
+        return grpc::Status(grpc::StatusCode::NOT_FOUND, "", "");
+    }
+    else if (metrics_opt_in == -1)
+    {
+        metrics_opt_in == QString::number(request->opt_in());
+        QFile opt_in_file{QDir(config->data_directory).filePath(metrics_opt_in_name)};
+        opt_in_file.open(QIODevice::WriteOnly);
+        opt_in_file.write(QString::number(request->opt_in()).toUtf8());
+        opt_in_file.close();
```
Saviq: We should also send a one-off "denied" metric here.

src/client/cmd/launch.cpp
```diff
@@ -220,3 +158,84 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+            else if (error == LaunchError::INVALID_HOSTNAME)
+            {
+                cerr << "Invalid instance name supplied: " << request.instance_name() << "\n";
+            }
+        }
+
+        return ReturnCode::CommandFail;
+    };
+
+    auto streaming_callback = [this, &spinner](mp::LaunchReply& reply) {
```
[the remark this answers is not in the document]
townsend2010: I don't think that is really possible with how the grpc streaming implementation is done.  Streaming is basically one way, so the server is sending the data to the client, which we can get the prompt to ask about opting in, but there isn't really a way to send the answer back to the server at this point.

src/client/cmd/launch.cpp
```diff
@@ -220,3 +158,84 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+    auto on_success = [this, &spinner](mp::LaunchReply& reply) {
+        spinner.stop();
+        cout << "Launched: " << reply.vm_instance_name() << "\n";
+        return ReturnCode::Ok;
+    };
+
+    auto on_failure = [this, &spinner](grpc::Status& status) {
+        spinner.stop();
+
+        if (status.error_code() == grpc::StatusCode::NOT_FOUND)
```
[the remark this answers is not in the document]
townsend2010: No, this is triggered by the daemon seeing that the opt-in file has not been created and set and telling the client about it.  I used `NOT_FOUND` here since that was the most relevant grpc status enum I could find:)

src/client/cmd/launch.cpp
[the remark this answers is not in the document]
townsend2010: Also, yes, I need to account for non-interactive shells, so I'll add that logic in.

src/daemon/daemon.cpp
townsend2010: How would a "denied" metric look like?

src/daemon/daemon.cpp
```diff
@@ -307,6 +309,49 @@ auto grpc_status_for(fmt::memory_buffer& errors)
...
+    if (!id_file.exists())
+    {
+        id = mp::MetricsProvider::generate_unique_id();
+        id_file.open(QIODevice::WriteOnly);
+        id_file.write(id.toUtf8());
+    }
+    else
+    {
+        id_file.open(QIODevice::ReadOnly);
+        id = QString(id_file.readAll());
```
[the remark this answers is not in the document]
townsend2010: Not sure if we should be worried about it since it's just an anonymous randomly generated identifier, but I suppose if the value read from the file is not a valid UUID, then we should generate a new one and save and use that.

src/daemon/daemon.cpp
Saviq: Just a `{"denied": 1}` or some such? Just so that we can get a total number.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
Saviq: I suppose that would be true for any of the values we store... Let's get to that when we build the config system.

src/daemon/daemon.cpp
townsend2010: Ok, seems [{"denied": 1}] is accepted by the endpoint, so we'll go with that:)

src/daemon/daemon.cpp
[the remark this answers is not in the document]
townsend2010: Ack

src/daemon/daemon.cpp
Saviq: As long as that will give us a sum of that ;)

townsend2010: > As long as that will give us a sum of that ;)

Not really sure.  @cmars, will that `denied` like that give a total of all `denied`'s sent to the endpoint?

townsend2010:
Ok, I think this is ready for another round of review;)

src/daemon/daemon.cpp
```diff
@@ -307,6 +309,49 @@ auto grpc_status_for(fmt::memory_buffer& errors)
...
+}
+
+auto get_metrics_opt_in(const mp::Path& data_path)
+{
+    auto val{0};
+    QFile opt_in_file{QDir(data_path).filePath(metrics_opt_in_name)};
+
+    if (!opt_in_file.exists())
+    {
+        val = -2;
```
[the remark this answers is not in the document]
townsend2010: For now, I added "later", but it will nag the user on every launch until they commit to yes or no:)  We can add the counter later.

src/client/cmd/launch.cpp
```diff
@@ -220,3 +169,111 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+                        break;
+                    }
+                    else if (answer.empty() || std::regex_match(answer, later))
+                    {
+                        request.set_opt_in_reply(static_cast<int>(mp::OptInStatus::LATER));
+                        break;
+                    }
+                    else
+                    {
+                        cout << "Invalid answer given. Please try again.\n";
```
Saviq: I wonder if this was better:
```
Send usage data (yes/no/Later)? BLAH
Please answer yes/no/Later: BLAH
Please answer yes/no/Later: l
...
```

vs. the current:
```
Send usage data (yes/no/Later)? BLAH
Invalid answer given. Please try again.
Send usage data (yes/no/Later)? BLAH
Invalid answer given. Please try again.
Send usage data (yes/no/Later)? l
...
```

src/rpc/multipass.proto
```diff
@@ -75,6 +76,7 @@ message LaunchReply {
         LaunchProgress launch_progress = 2;
         string create_message = 3;
     }
+    bool opt_in_missing = 4;
```
Saviq: I'd name this something more descriptive... `metrics_pending`?

Saviq: I'd name this something more descriptive... `metrics_pending`?

src/client/cmd/launch.cpp
```diff
@@ -220,3 +169,111 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+                     << "We'd also like to know how long it takes for certain operations to complete.\n\n"
+                     << "Send usage data (yes/no/Later)? ";
+
+                while (true)
+                {
+                    std::string answer;
+                    std::getline(std::cin, answer);
+                    if (std::regex_match(answer, yes))
+                    {
+                        request.set_opt_in_reply(static_cast<int>(mp::OptInStatus::ACCEPTED));
```
albaguirre: The RPC should have an equivalent enum and there should be to and from methods to convert between mp::OptInStatus and the RPC equivalent (therefore, removing all static_casts)

src/rpc/multipass.proto
```diff
@@ -44,6 +44,7 @@ message LaunchRequest {
     string time_zone = 7;
     string cloud_init_user_data = 8;
     string remote_name = 9;
+    int32 opt_in_reply = 10;
```
albaguirre: This should be an enum that mirrors mp::OptInStatus, this way there's no ambiguity which values can be set here.

src/metrics/metrics_provider.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    post_request(body);
+
+    return true;
+}
+
+void mp::MetricsProvider::send_denied()
+{
+    QJsonObject denied;
+    denied.insert("denied", 1);
```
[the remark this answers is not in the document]
townsend2010: Good question...the Omnibus service on the other end accepts this and it's not sending any sort of identifying information to it, which I kind of thought was the purpose of denying metrics- not sending any identifying info.  I know ubuntu-report just sends a "denied" kind of message back to the server with nothing else with it.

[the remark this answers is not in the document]
Saviq: @cmars can you suggest a minimal "this client denied sending metrics" JSON?

PR #341, opened 2018-08-16
metrics_provider: Retry network errors

townsend2010:
Use a separate thread for the actual sending of the metrics.

### DIFF
--- a/src/metrics/metrics_provider.cpp
+++ b/src/metrics/metrics_provider.cpp
@@ -21,33 +21,141 @@
 
 #include <fmt/format.h>
 
-#include <QByteArray>
 #include <QDateTime>
 #include <QEventLoop>
-#include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QNetworkReply>
 #include <QNetworkRequest>
 #include <QRegExp>
-#include <QString>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
+using namespace std::literals::chrono_literals;
+
 namespace
 {
 constexpr auto category = "metrics";
+
+void post_request(const QUrl& metrics_url, const QByteArray& body)
+{
+    QNetworkAccessManager manager;
+    QNetworkRequest request{metrics_url};
+    request.setHeader(QNetworkRequest::ContentTypeHeader, "application/json");
+    request.setHeader(QNetworkRequest::ContentLengthHeader, body.size());
+    request.setHeader(QNetworkRequest::UserAgentHeader, "multipassd/1.0");
+
+    QEventLoop event_loop;
+    std::unique_ptr<QNetworkReply> reply{metrics_url.isLocalFile() ? manager.put(request, body)
+                                                                   : manager.post(request, body)};
+
+    QObject::connect(reply.get(), &QNetworkReply::finished, &event_loop, &QEventLoop::quit);
+
+    event_loop.exec();
+
+    auto buff = reply->readAll();
+
+    if (reply->error() == QNetworkReply::ProtocolInvalidOperationError)
+    {
+        auto error_msg = QJsonDocument::fromJson(buff).object();
+        mpl::log(mpl::Level::error, category,
+                 fmt::format("Metrics error: {} - {}", error_msg["code"].toString().toStdString(),
+                             error_msg["message"].toString().toStdString()));
+    }
+    else if (reply->error() != QNetworkReply::NoError)
+    {
+        QObject::disconnect(reply.get(), 0, 0, 0);
+        throw std::runtime_error(fmt::format("Metrics error: {}", reply->errorString().toStdString()));
+    }
+}
 }
 
-mp::MetricsProvider::MetricsProvider(const QUrl& metrics_url, const QString& unique_id)
-    : metrics_url{metrics_url}, unique_id{unique_id}
+mp::MetricsProvider::MetricsProvider(const QUrl& url, const QString& unique_id)
+    : metrics_url{url}, unique_id{unique_id}, metrics_sender{[this] {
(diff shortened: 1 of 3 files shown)


src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
+              }
+
+              auto body = QJsonDocument(metric_batches).toJson(QJsonDocument::Compact);
+              lck.unlock();
+
+              try
+              {
+                  post_request(metrics_url, body);
+                  lck.lock();
+                  if (metric_batches.first().isNull())
```
albaguirre: Once the metrics are sent why are they being checked again? I think an explicit exit flag would be clearer.

src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
+
+              if (metric_batches.empty())
+              {
+                  metrics_cv.wait(lck);
+              }
+
+              // Null element signals thread to exit
+              if (metric_batches.first().isNull())
+              {
+                  lck.unlock();
```
albaguirre: No need for the manual unlock here.

But in any case, this should be clearer with an explicit exit flag.

src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
+                      lck.unlock();
+                      return;
+                  }
+                  metric_batches = QJsonArray{};
+                  lck.unlock();
+              }
+              catch (const std::exception& e)
+              {
+                  mpl::log(mpl::Level::error, category, fmt::format("{} - Attempting to resend", e.what()));
+                  std::this_thread::sleep_for(60s);
```
albaguirre: I think this should probably be an exponentially backoff timeout.

Also instead of sleep (which cannot be aborted) maybe using the timeout version of the condition variable is best.

So maybe the loop looks something like:
```
std::unique_lock<std::mutex> lock{metrics_mutex};
auto timeout = std::chrono::seconds::max(); //max because we don't want to initially time out until metrics are available
while (running)
{
cv.wait_for(lock, timeout, [&] { return metrics_available || !running;})

if (!running) return;
if (!metrics_available) continue; //timed out and no metrics available yet

<attempt to send metrics>
if (failed)
   timeout = <increase timeout exponentially until some ceiling value?>;
else
  return;
}
```

src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
 
-mp::MetricsProvider::MetricsProvider(const QUrl& metrics_url, const QString& unique_id)
-    : metrics_url{metrics_url}, unique_id{unique_id}
+mp::MetricsProvider::MetricsProvider(const QUrl& url, const QString& unique_id)
+    : metrics_url{url}, unique_id{unique_id}, metrics_sender{[this] {
+          std::unique_lock<std::mutex> lck(metrics_mutex, std::defer_lock);
+          while (true)
+          {
+              while (!lck.try_lock())
+                  std::this_thread::sleep_for(1ms);
```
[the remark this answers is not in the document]
townsend2010: So before, I did a lck.lock() (and even a std::unique_lock<std::mutex> lock{metrics_mutex};) here and that resulted in some very unpredictable behavior in the unit tests where sometimes lock() would throw owner_dead (even though it wasn't) or it would just block forever unless I broke into it with gdb and stepped through the code.

src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
+    request.setHeader(QNetworkRequest::UserAgentHeader, "multipassd/1.0");
+
+    QEventLoop event_loop;
+    std::unique_ptr<QNetworkReply> reply;
+
+    // For unit testing
+    if (metrics_url.isLocalFile())
+        reply = std::unique_ptr<QNetworkReply>(manager.put(request, body));
+    else
+        reply = std::unique_ptr<QNetworkReply>(manager.post(request, body));
```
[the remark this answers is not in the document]
townsend2010: Yeah, that's cleaner:)

src/metrics/metrics_provider.cpp
```diff
@@ -21,27 +21,103 @@
...
+    }
+}
 }
 
-mp::MetricsProvider::MetricsProvider(const QUrl& metrics_url, const QString& unique_id)
-    : metrics_url{metrics_url}, unique_id{unique_id}
+mp::MetricsProvider::MetricsProvider(const QUrl& url, const QString& unique_id)
+    : metrics_url{url}, unique_id{unique_id}, metrics_sender{[this] {
+          std::unique_lock<std::mutex> lck(metrics_mutex, std::defer_lock);
+          while (true)
```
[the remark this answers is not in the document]
townsend2010: Ok, that's probably better than relying on a null QJsonValue.

src/metrics/metrics_provider.cpp
townsend2010: A null QJsonValue is the indicator that the thread needs to exit, so after posting the metrics, the thread should check for the NULLness.  Sure a flag would be fine too, but I was just avoiding cluttering up the class a bit, but it would probably be clearer and also would allow for the explicit thread exit.  I'll just use an atomic bool for this.

src/metrics/metrics_provider.cpp
townsend2010: Ok

src/metrics/metrics_provider.cpp
townsend2010: Yeah, I was planning on adding a backoff timeout for each network failure.  Also, we do want to unlock before sending the metrics to allow any new metrics to be added to the JsonArray independent of the thread trying to send out the metrics.  It allows for queuing of metrics by adding a new JSonArray entry.

PR #345, opened 2018-08-20
[travis] add macOS builds

Saviq:
This ships out the pull request number / commit id to a host defined in
encrypted MAC_{HOST,USER,PORT} variables to build. Expects a
`multipass_build/build-in-snapshot.sh` to be maintained there.


Saviq:
This shouldn't actually build on macOS because we don't have the encrypted data.

PR #354, opened 2018-08-24
change boot wait message

powersj:
Rather than make the user think cloud-init is the only thing running in the background, update the message to instead be more generic while waiting for system to boot and initialize.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -621,7 +621,7 @@ try // clang-format on
     vm->start();
     vm->wait_until_ssh_up(std::chrono::minutes(5));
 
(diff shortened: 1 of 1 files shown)


src/daemon/daemon.cpp
```diff
@@ -621,7 +621,7 @@ try // clang-format on
     vm->start();
     vm->wait_until_ssh_up(std::chrono::minutes(5));
 
-    reply.set_create_message("Waiting for cloud-init to complete");
+    reply.set_create_message("Waiting for boot and initialization to complete");
```
townsend2010: It would probably be better to just say 'initialization' here since the VM is already booted at this point.

PR #355, opened 2018-08-24
cloud-init: use status subcommand

powersj:
The cloud-init command has a status subcommand that can be used to
determine when cloud-init is complete instead of waiting for the
specific file.

There is also a --wait flag that can be used to wait while boot is on-going as well, but to remain consistent with the previous behavior of checking and re-trying I omitted this option flag.


powersj:
Ah shoot I forgot about the trusty version. I'll close this then.

PR #362, opened 2018-09-04
[travis] use ccache in LXD builds

Saviq:


Saviq:
This seems to make it slower rather than faster... ¿?

tests/travis-Debug.patch
```diff
@@ -1,13 +1,13 @@
 --- a/snap/snapcraft.yaml
 +++ b/snap/snapcraft.yaml
-@@ -143,9 +143,8 @@ parts:
-     - golang
```
ricab: Is this somehow related to ccache?

Saviq: Note this is a diff of a diff, so this isn't an actual change of the resulting file, but rather just some diff fuzz.

ricab: Right, just the context lines changing... got to get used to diff-squared. Please don't ever go beyond that :)

Saviq:
Some travis slowness ensued...

bors r=ricab

PR #363, opened 2018-09-05
Collect instance metrics

townsend2010:
Here is the first stab at introducing more instance specific metrics.  This uses the histogram aggregation approach by placing many metrics in 'bins' and counting the number of metrics places in a particular bin.

Also, I'm not 100% sure of the key strings for each metric, so let's scrutinize those some:)

Lastly, any easy way to test the output of this is to change `metrics_url` in `src/daemon/daemon_config.h` to be something like `file:///tmp/metrics.json` and then inspecting that file.  I used something like `cat /tmp/metrics.json | jq` to get a better output.


townsend2010:
@albaguirre,

Yeah, the bin labels I'm not so sure about. I think we are supposed to hash that out in Brussels. However, I'm not sure if using something like "0-512", "512-1024" is good because which bin do we place a value of "512" in?  That is kind of why I made the bins with no overlapping values at all...but again, something to hash out soon:)

townsend2010:
Gonna wait to merge this until we have quorum on how this data should be sent, etc.

Saviq:
bors try

PR #367, opened 2018-09-06
Fix copy files wildcard

townsend2010:


src/ssh/scp_client.cpp
```diff
@@ -52,7 +52,17 @@ mp::SCPClient::SCPClient(SSHSessionUPtr ssh_session) : ssh_session{std::move(ssh
...
-    SCPUPtr scp{make_scp_session(*ssh_session, SSH_SCP_WRITE, destination_path.c_str())};
+    auto full_destination_path{destination_path};
+    if (full_destination_path.empty())
+    {
+        full_destination_path.append(QFileInfo(QString::fromStdString(source_path)).fileName().toStdString());
+    }
+    else if (QFileInfo(QString::fromStdString(destination_path)).isDir())
+    {
+        full_destination_path.append("/" + QFileInfo(QString::fromStdString(source_path)).fileName().toStdString());
+    }
```
albaguirre: Perhaps the above can be broken into:

A couple of utility functions (maybe even include them in utils.h?)
```
auto filename_for(const std::string& path)
{
    return QFileInfo(QString::fromStdString(path)).fileName().toStdString();
}

bool is_dir(const std::string& path)
{
    return QFileInfo(QString::fromStdString(path)).isDir();
}
```

Then you can lambda initialize full_destination_path like:

```
const std::string full_destination_path = [&]{
        if (destination_path.empty())
            return filename_for(source_path);
        if (is_dir(destination_path))
            return destination_path + "/" + filename_for(source_path);
        return destination_path;
    }();

```

Though I see this is the same logic as in pull_file, maybe it should be a full function, like:

```
std::string full_destination(const std::string& source, const std::string& dest)
{
...
}
```

src/ssh/scp_client.cpp
```diff
@@ -90,37 +100,50 @@ void mp::SCPClient::pull_file(const std::string& source_path, const std::string&
...
+        if (r == SSH_ERROR || r == SSH_SCP_REQUEST_WARNING)
+            throw std::runtime_error("Error receiving information for file: " + std::string(ssh_get_error(scp.get())));
+
+        auto size = ssh_scp_request_get_size(scp.get());
+        std::string filename{ssh_scp_request_get_filename(scp.get())};
+
+        auto total{0u};
+        const auto len{65536u};
+        std::vector<char> data;
+        data.reserve(len);
```
albaguirre: Prefer std::array instead:

`std::array<char, 65536u> data; `

and use data.size() instead of len.

src/ssh/scp_client.cpp
```diff
@@ -90,37 +100,50 @@ void mp::SCPClient::pull_file(const std::string& source_path, const std::string&
...
+
+        auto full_destination_path{destination_path};
+        if (full_destination_path.empty())
+        {
+            full_destination_path = filename;
+        }
+        else if (QFileInfo(QString::fromStdString(destination_path)).isDir())
+        {
+            full_destination_path.append("/" + filename);
+        }
```
albaguirre: Since it's the same logic as in push_file, this should be its own function in the anonymous namespace.

townsend2010: It's not quite the same logic in that in `pull_file`, it will use the filename retrieved from the scp server, but in `push_file`, it will the filename provided in the source path.  So then, we would still have to pass in the filename we want to use.  But yeah, probably some way to genericise this some...

src/ssh/scp_client.cpp
albaguirre: The main reason I say to use std::array, is that doing a std::vector::reserve and then using the pointer from std::vector::data() is undefined behavior.

PR #385, opened 2018-09-12
Add shutdown delay option

townsend2010:


src/daemon/daemon.cpp
```diff
@@ -1233,9 +1239,6 @@ try // clang-format on
             continue;
         }
 
-        if (it->second->current_state() == VirtualMachine::State::running)
-            continue;
-
```
albaguirre: Why is this removed? 

It should be essentially a no-op starting an already running VM.

src/platform/backends/backend_utils/backend_utils.cpp
```diff
@@ -107,3 +108,25 @@ std::string mp::backend::image_format_for(const mp::Path& image_path)
 
     return image_record["format"].toString().toStdString();
 }
+
+void mp::backend::shutdown_instance(const std::string& vm_name, VirtualMachine::State& state,
+                                    QTimer& delay_shutdown_timer, std::chrono::minutes delay,
+                                    std::function<void()> const& process_shutdown_request)
+{
```
albaguirre: This should be check if the state is already stopped.

Otherwise, you end up being able to change the state of stopped instances into DELAYED_SHUTDOWN which will never happen as they are stopped already.

src/daemon/daemon.cpp
```diff
@@ -1340,7 +1343,14 @@ try // clang-format on
     for (const auto& name : instances_to_stop)
     {
         auto it = vm_instances.find(name);
-        it->second->shutdown();
+        if (request->cancel_shutdown())
+        {
+            it->second->start();
```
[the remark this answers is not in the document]
townsend2010: I implemented the timer in the VM since I thought it would be cleaner to always keep a timer object with each VM instead of creating some new unordered_map of <vm_name, QTimer> in the daemon.  But if you feel it's more RAII to keep the timer in the daemon itself, then I'm fine with.  Always different ways to accomplish the same thing:)

src/platform/backends/backend_utils/backend_utils.cpp
```diff
@@ -107,3 +108,25 @@ std::string mp::backend::image_format_for(const mp::Path& image_path)
...
+    {
+        mpl::log(mpl::Level::info, vm_name,
+                 fmt::format("Shutdown request delayed for {} minutes",
+                             delay.count() * std::chrono::seconds::period::num / std::chrono::seconds::period::den));
+        delay_shutdown_timer.setSingleShot(true);
+        QObject::connect(&delay_shutdown_timer, &QTimer::timeout, process_shutdown_request);
+
+        state = VirtualMachine::State::delayed_shutdown;
+
+        delay_shutdown_timer.start(std::chrono::duration_cast<std::chrono::milliseconds>(delay));
```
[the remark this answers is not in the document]
townsend2010: I looked at the QTimer::start() prototype and it says std::chrono::milliseconds which is why I do the duration_cast.  I didn't check if the compiler automatically casts std::chrono::minutes to std::chrono::milliseconds here.  I'll try that.

src/daemon/daemon.cpp
townsend2010: I removed it because if there is a start called on a VM that has a delayed shutdown running, then it's supposed to cancel the delayed shutdown.  The way I implemented it in the VM, then the QTimer is never cancelled in this case.  But if I change it as discussed above, then I won't remove this call and instead, will add a call to see if a delayed shutdown is running.

src/platform/backends/backend_utils/backend_utils.cpp
townsend2010: Yep, you're right:)

src/daemon/daemon.cpp
albaguirre: But won't you have a different state then? The delayed shutdown state? Not running state?

src/daemon/daemon.cpp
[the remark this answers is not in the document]
albaguirre: Right, the main thing was making the shutdown explicit instead of calling start. 

But yeah, one can avoid that all together if the daemon (or an entity owned by the daemon) implements this logic, since the implementation seems to apply to any backend.

PR #388, opened 2018-09-13
qemu: ensure dnsmasq releases unused ip addresses

### LINKED ISSUE #383: qemu backend - dnsmasq leases don't expire, ip addresses
Creating many different VM's each with a unique name, eventually makes dnsmasq exhaust its ip address pool.

The addresses don't expire and are not released when a vm instance is stopped or even purged.

it can be reproduced with the following:

```
#!/bin/bash
for i in {1..512}
do
  bin/multipass launch -n test${i}
  bin/multipass delete -p test${i}  
done
```

albaguirre:
Fixes #383

### DIFF
--- a/src/platform/backends/qemu/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/dnsmasq_server.cpp
@@ -17,22 +17,31 @@
 
 #include "dnsmasq_server.h"
 
+#include <multipass/logging/log.h>
 #include <multipass/utils.h>
+
+#include <fmt/format.h>
 
 #include <fstream>
 namespace mp = multipass;
+namespace mpl = multipass::logging;
 
 namespace
 {
-auto start_dnsmasq_process(const QDir& data_dir, const QString& bridge_name, const mp::IPAddress& bridge_addr,
+auto start_dnsmasq_process(const QDir& data_dir, const std::string& bridge_name, const mp::IPAddress& bridge_addr,
                            const mp::IPAddress& start_ip, const mp::IPAddress& end_ip)
 {
     auto cmd = std::make_unique<QProcess>();
+
+    QObject::connect(cmd.get(), &QProcess::readyReadStandardError,
+                     [&cmd]() { mpl::log(mpl::Level::error, "dnsmasq", cmd->readAllStandardError().data()); });
+
     cmd->start("dnsmasq",
                QStringList() << "--keep-in-foreground"
                              << "--strict-order"
                              << "--bind-interfaces"
-                             << "--except-interface=lo" << QString("--interface=%1").arg(bridge_name)
+                             << "--except-interface=lo"
+                             << QString("--interface=%1").arg(QString::fromStdString(bridge_name))
                              << QString("--listen-address=%1").arg(QString::fromStdString(bridge_addr.as_string()))
                              << "--dhcp-no-override"
                              << "--dhcp-authoritative"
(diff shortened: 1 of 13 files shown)


src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -17,22 +17,31 @@
...
+#include <fmt/format.h>
+
 #include <fstream>
 namespace mp = multipass;
+namespace mpl = multipass::logging;
 
 namespace
 {
-auto start_dnsmasq_process(const QDir& data_dir, const QString& bridge_name, const mp::IPAddress& bridge_addr,
+auto start_dnsmasq_process(const QDir& data_dir, const std::string& bridge_name, const mp::IPAddress& bridge_addr,
```
townsend2010: Not sure why you changed `bridge_name` from a `QString` to a `std::string`.  It's used as a `QString` in this function.

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -47,12 +56,11 @@ auto start_dnsmasq_process(const QDir& data_dir, const QString& bridge_name, con
 }
 } // namespace
 
-mp::DNSMasqServer::DNSMasqServer(const Path& path, const QString& bridge_name, const IPAddress& bridge_addr,
+mp::DNSMasqServer::DNSMasqServer(const Path& path, const std::string& bridge_name, const IPAddress& bridge_addr,
```
townsend2010: Again, not sure why `bridge_name` is now a `std::string`.

tests/CMakeLists.txt
```diff
@@ -12,9 +12,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-# Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
 
-list(APPEND BACKEND_TESTS test_qemu_backend.cpp)
+list(APPEND BACKEND_TESTS
```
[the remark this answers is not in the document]
albaguirre: Yeah a proper merge has to occur on the private branch.

src/platform/backends/qemu/dnsmasq_server.cpp
albaguirre: Just the usual class boundary.

Saviq:
This means the instance will get a different IP every time it boots, I'm not sure we want that - let's see if we can rely on the lease being shorter? Like a day?

Also maybe we can use a /16 mask so we have a 256-times less likely of us hitting the issue ;)

townsend2010:
I think instead of doing the dhcp release on shutdown, we do the release when an instance is being purged.

src/platform/backends/qemu/dnsmasq_server.cpp
townsend2010: lol, well, in this case, I prefer to leave them all as QStrings since everything consumes them as QStrings :)

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -260,6 +260,8 @@ void mp::QemuVirtualMachine::on_error()
 void mp::QemuVirtualMachine::on_shutdown()
 {
     state = State::off;
+    ip = nullopt;
+    dnsmasq_server->release_mac(mac_addr);
```
townsend2010: As I mentioned in a comment, but maybe it would be better to do the explicit release when an instance is purged.

tests/CMakeLists.txt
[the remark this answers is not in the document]
townsend2010: Ah, ok:)

PR #391, opened 2018-09-14
daemon config: remove string constant

albaguirre:


albaguirre:
I was just recently changing daemon_config and noticed it. 

I usually put constants like this on the implementation file, just to avoid polluting the namespace of any other files which include daemon_config.h

Issue #387, opened 2018-09-13
provide snapcraft aliases for VMs


sergiusens:
As a proposal:

- snapcraft-core18 -> 18.04
- snapcraft-core16 -> 16.04
- snapcraft-core -> 16.04 (this is a special snowflake as it is not a base, but it is a _core_ base?)

Saviq:
@sergiusens would you object to it being `snapcraft:…` instead of the hyphen? It would make for a better isolation of the snapcraft build VMs IMO.

sergiusens:
El 18/09/18 a las 10:15, Michał Sawicz escribió:
>
> @sergiusens <https://github.com/sergiusens> would you object to it 
> being |snapcraft:…| instead of the hyphen? It would make for a better 
> isolation of the snapcraft build VMs IMO.
>

No objections!

PR #393, opened 2018-09-17
custom_image_host: add snapcraft aliases

sergiusens:
The aliases point to snapcraft-$base using images from cloud images.

Signed-off-by: Sergio Schvezov <sergiusens@gmail.com>

### DIFF
--- a/src/daemon/custom_image_host.cpp
+++ b/src/daemon/custom_image_host.cpp
@@ -37,7 +37,7 @@ auto multipass_default_aliases(mp::URLDownloader* url_downloader)
 
     for (const auto& line : sha256_sums)
     {
-        if (line.contains("ubuntu-core-16-amd64.img.xz"))
+        if (line.endsWith("ubuntu-core-16-amd64.img.xz"))
         {
             hash = QString(line.split(' ').first());
             break;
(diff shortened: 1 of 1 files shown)


townsend2010:
I don't think we want to use `snapcraft:core16` unless we are saying `snapcraft` is a new custom remote.  Is that the case here?  #387 doesn't list is like this.

townsend2010:
Fixes #387

Saviq:
> Is that the case here?

It just felt right... we have a way of "grouping" images/aliases so thought it could be the right thing to do?

Otherwise, which remote does "snapcraft-core16" belong to?

src/daemon/custom_image_host.cpp
```diff
@@ -49,6 +49,56 @@ auto multipass_default_aliases(mp::URLDownloader* url_downloader)
 
     default_aliases.insert({"core", core_image_info});
 
+    // Image information on 16.04
+    const QString url_16{"https://cloud-images.ubuntu.com/releases/16.04/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img"};
+    const auto last_modified = url_downloader_16->last_modified({url_16});
```
townsend2010: `last_modified` has already been declared.  Either rename to be unique or fix the previous declaration to not be `const` and drop `auto` here.

Also in this line, `url_downloader_16` has not been declared anywhere.

src/daemon/custom_image_host.cpp
```diff
@@ -49,6 +49,56 @@ auto multipass_default_aliases(mp::URLDownloader* url_downloader)
...
+        if (line.endsWith("ubuntu-16.04-server-cloudimg-amd64-disk1.img"))
+        {
+            hash = QString(line.split(' ').first());
+            break;
+        }
+    }
+
+    // Image information on 18.04
+    const QString url_18{"https://cloud-images.ubuntu.com/releases/18.04/release/ubuntu-18.04-server-cloudimg-amd64.img"};
+    const auto last_modified = url_downloader_18->last_modified({url_18});
```
townsend2010: Same as above, `last_modified` has already been declared and `url_downloader_18` has not been declared:)

townsend2010:
After a bit of discussion with @Saviq, it probably makes sense to use a special `snapcraft` remote for this so it would be in the form of `snapcraft:core16`, etc.  If this indeed the intent, then I will take over this and add support for the special `snapcraft` remote:)

townsend2010:
I've proposed #398 in place of this PR.

PR #403, opened 2018-09-21
Fix broken aliases

### LINKED ISSUE #401: Launching by aliases from SimpleStreams remotes is broken
The last batch of commits broke launching by SimpleStreams aliases.

### LINKED ISSUE #402: Launching images via aliases that match multiple remotes may use the incorrect image
If an image using alias `foo` from remote `bar` has been prepared and launched and then launching an image using alias `foo` from a different remote `baz`, then the image associated with `bar` will be used.

townsend2010:
This fixes #401 and fixes #402.


albaguirre:
bors +r

albaguirre:
macmini is offline :), merging.

PR #418, opened 2018-09-27
daemon/info: Don't throw on exceptions when running commands in the i…

### LINKED ISSUE #381: multipass info fails for Ubuntu Core images
```
$ multipass launch core
Launched: gallant-moonfish
$ multipass info gallant-moonfish
info failed: failed to run 'lsb_release -ds', exit code:true
```
Ubuntu Core doesn't contain lsb_release.

townsend2010:
…nstance

If info cannot be retrieved in a running instance, it shouldn't throw but instead, an empty string should be used.

Fixes #381


townsend2010:
Since Mac builds are failing, I'm going to merge this manually.

PR #424, opened 2018-10-01
utils: Use throw instead of return when instance polling methods time…

### LINKED ISSUE #414: Timing out on SSH results in `Launched:` even when status is UNKNOWN
Having trouble launching core images (#412) I saw an inconsistent behaviour: the `launch` command exits with 0 and `Launched:` printed, but the instance is in UNKNOWN state.

townsend2010:
… out

Fixes #414


Saviq:
I said:

bors r+

Saviq:
Now I've fixed the Mac:

bors r+

PR #427, opened 2018-10-02
libvirt: Destroy bridge in the LibVirtVirtualMachineFactory dtor

### LINKED ISSUE #327: Removing multipass leaves mpvirtbr0 bridge behind
I installed the multipass snap (revision 334 from beta channel) and configured it to use the libvirt driver. I created a few containers and then removed the multipass snap from the system. After the snap was removed there was still a `mpvirtbr0` network device present. A new installation of the multipass snap caused now a `mpvirtbr1` device to be created which will cause multipass to not find the right network bridge when the system is rebooted. libvirt will fail to start then with:

`Aug 08 10:06:44 saturn libvirtd[3188]: Cannot get interface MTU on 'mpvirtbr1': No such device`

and multipassd with

`Aug 06 09:07:54 saturn multipass.multipassd[5868]: libvirt: XML-RPC error : Failed to connect socket to '/var/snap/multipass/common/run/libvirt/libvirt-sock': Connection refused`

Only way to get out of this is to manually modify the configuration of all present containers to use `mpvirtbr0` instead of `mpvirtbr1`.

### LINKED ISSUE #413: Network bridges leaking
When switching between `QEMU` and `LIBVIRT` backends, the bridges are left behind:
```
$ brctl show               
bridge name     bridge id               STP enabled     interfaces
mpqemubr0       8000.52540055481d       no              mpqemubr0-dummy
mpvirtbr0       8000.52540039b09a       yes             mpvirtbr0-nic
```

townsend2010:
Revert back to using a non-persistent libvirt network and instead, save the subnet
like is done for the Qemu backend.

Fixes #327, fixes #413


Saviq:
I'm unsure about this... I was surprised to see the `mpvirtbr0` bridge when I switched to `driver=QEMU`, but if that's how libvirt behaves with "static" networks, maybe we shouldn't go out of our way to delete it just for the sake of cleaning up. Especially since this would only happen when switching drivers.

townsend2010:
In a way, I think libvirtd should be the one responsible for tearing down networks if it is stopped, but it doesn't.  Even the regular libvirt from the Debian package will not remove bridges when it is stopped.

That said, I'm of the opinion that multipassd is really the one in control of the `mpvirtbr0` bridge, so if multipassd is stopped, it should probably clean up the bridge via libvirtd.

gerboland:
I'm not aware of any harm in removing the bridge, since it is only for multipass' consumption. So I don't object to this

Saviq:
OK then,

bors r+

townsend2010:
bors r=Saviq

Saviq:
One more time:

bors r+

PR #429, opened 2018-10-03
ssh_process: Return any data already read if the channel is closed

### LINKED ISSUE #425: mount on startup sometimes fail due to the instance not being ready yet
While we've increased the timeout for some commands, it seems we've missed some, especially around mounting:

```
$ SNAPCRAFT_BUILD_INFO=1 SNAPCRAFT_BUILD_ENVIRONMENT=multipass snapcraft
Launching a VM.
Removing "/home/multipass/project": stoi
An error occurred when trying to start the instance with 'multipass': returned exit code 2.
```

townsend2010:
Also add more verbose debugging info for catching cases like this in the future.

Fixes #425

Co-authored-by: Chris Townsend <christopher.townsend@canonical.com>

### DIFF
--- a/src/ssh/ssh_process.cpp
+++ b/src/ssh/ssh_process.cpp
@@ -129,7 +129,7 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
     {
         num_bytes = ssh_channel_read_timeout(channel.get(), buffer.data(), buffer.size(), is_std_err, timeout);
         mpl::log(mpl::Level::debug, category,
-                 fmt::format("{}:{} {}(): num_bytes={}", __FILE__, __LINE__, __FUNCTION__, num_bytes));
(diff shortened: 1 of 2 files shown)


src/ssh/ssh_process.cpp
```diff
@@ -116,17 +128,25 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
...
+                return output.str();
+            }
 
             throw std::runtime_error(fmt::format("error while reading ssh channel for remote process '{}'"
                                                  " - error: {}",
                                                  cmd, num_bytes));
         }
         output.write(buffer.data(), num_bytes);
+        mpl::log(mpl::Level::debug, category,
+                 fmt::format("{}:{} {}(): output = {}", __FILE__, __LINE__, __FUNCTION__, output.str()));
```
Saviq: I'm worried this might be too much, imagine a long-running command through this, do we want to log that much?

townsend2010: Yeah, probably too verbose right here.  I'll take that out.

PR #433, opened 2018-10-05
Map mounting user's uid/gid by default (Fixes #331)

### LINKED ISSUE #331: Map mounting user to `multipass` automatically
When mounting into instances, we need to automatically map from the mounting user to the `multipass` user. Otherwise permissions over the mounts will be all over the place.

We already allow `--uid-map`, `--gid-map` set to arbitrary values so there's no more of a security concern here. Giving access to multipass means giving root access to the filesystem.

Saviq:


Saviq:
@townsend2010 done.

PR #436, opened 2018-10-10
[sshfs] fix file metadata passthrough

gerboland:
handle_opendir only put file names on the open_dir_handles so handle_readdir had no directory information. QFileInfo returns default file attributes if the file it points to does not exist. Combination
of these facts meant sometimes incorrect file metadata could be returned.

The test I added fails without the patch, so this is correcting something.

### DIFF
--- a/include/multipass/sshfs_mount/sftp_server.h
+++ b/include/multipass/sshfs_mount/sftp_server.h
@@ -72,7 +72,7 @@ private:
     SSHSession ssh_session;
     const SftpSessionUptr sftp_server_session;
     const std::string source_path;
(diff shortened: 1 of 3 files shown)


Saviq:
```
[ RUN      ] SftpServer.handles_readdir_attributes_preserved
/root/build_multipass/parts/multipass/src/tests/test_sftpserver.cpp:830: Failure
      Expected: (test_file_attrs->permissions & 0700) >> 6
      Which is: 2
To be equal to: (test_file_info.permissions() & 0x700) >> 8
      Which is: 7
[  FAILED  ] SftpServer.handles_readdir_attributes_preserved (0 ms)
```

PR #444, opened 2018-10-17
platform: Split out the no_id_info_available const so it can be used …

townsend2010:
…elsewhere


Saviq:
Should we maybe put it together with `mp::sftp_server::default_id` somewhere?

Saviq:
Thanks,

bors r+

PR #446, opened 2018-10-17
Add vm suspend option

### LINKED ISSUE #374: Provide a way to suspend/resume a machine
This is how we stop https://github.com/sergiusens/snapcraft/blob/924aa5de6e35e69679d83b66f005d767f86cd701/snapcraft/internal/build_providers/_qemu/_qemu_driver.py#L107

And this is how we start https://github.com/sergiusens/snapcraft/blob/924aa5de6e35e69679d83b66f005d767f86cd701/snapcraft/internal/build_providers/_qemu/_qemu_command.py#L110

qemu-img can show you the list of existing tags, given that you only allow to set ram (and probably CPU) only on `multipass launch` a fixed tag name could do, or leave room for the future and tag them the the necessary information to load later.

townsend2010:
Fixes #374

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -12,6 +12,21 @@ driver=${driver:-QEMU}
 driver_saved=${driver_saved:-QEMU}
 
 if [[ $driver != $driver_saved ]]; then
+    /snap/bin/multipass list &>/dev/null || /bin/true
+
+    if [ $? -ne 0 ]; then
+        echo "Cannot determine status of instances. Not switching driver."
+        exit 1
+    fi
+
+    if /snap/bin/multipass list --format=csv | grep -Ev '(\+--|Name)' | grep --quiet -v 'STOPPED'; then
+        echo "All instances need to be stopped."
+        echo "Please save any data and stop the instances before proceeding:"
+        echo
+        echo "multipass stop --all"
+        exit 1
+    fi
+
     echo "$driver" > $SNAP_COMMON/driver
     snapctl restart $SNAP_NAME
 fi
--- a/src/client/cmd/suspend.cpp
+++ b/src/client/cmd/suspend.cpp
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2018 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "suspend.h"
+
+#include "animated_spinner.h"
+
+#include <multipass/cli/argparser.h>
+
+namespace mp = multipass;
+namespace cmd = multipass::cmd;
+using RpcMethod = mp::Rpc::Stub;
+
+mp::ReturnCode cmd::Suspend::run(mp::ArgParser* parser)
+{
+    auto ret = parse_args(parser);
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+
+    auto on_success = [](mp::SuspendReply& reply) { return ReturnCode::Ok; };
+
+    AnimatedSpinner spinner{cout};
+    auto on_failure = [this, &spinner](grpc::Status& status) {
+        spinner.stop();
+        cerr << "suspend failed: " << status.error_message() << "\n";
+        return return_code_for(status.error_code());
+    };
+
+    std::string message{"Suspending "};
+    if (request.instance_name().empty())
+        message.append("all instances");
+    else if (request.instance_name().size() > 1)
+        message.append("requested instances");
+    else
+        message.append(request.instance_name().Get(0));
+    spinner.start(message);
+    request.set_verbosity_level(parser->verbosityLevel());
+    return dispatch(&RpcMethod::suspend, request, on_success, on_failure);
+}
+
+std::string cmd::Suspend::name() const
+{
+    return "suspend";
+}
+
+QString cmd::Suspend::short_help() const
+{
+    return QStringLiteral("Suspend running instances");
+}
+
+QString cmd::Suspend::description() const
+{
+    return QStringLiteral("Suspend the named instances, if running. Exits with\n"
+                          "return code 0 if successful.");
+}
+
+mp::ParseCode cmd::Suspend::parse_args(mp::ArgParser* parser)
+{
+    parser->addPositionalArgument("name", "Names of instances to suspend", "<name> [<name> ...]");
+
+    QCommandLineOption all_option("all", "Suspend all instances");
+    parser->addOptions({all_option});
+
+    auto status = parser->commandParse(this);
+    if (status != ParseCode::Ok)
+        return status;
+
+    auto num_names = parser->positionalArguments().count();
+    if (num_names == 0 && !parser->isSet(all_option))
+    {
+        cerr << "Name argument or --all is required\n";
+        return ParseCode::CommandLineError;
+    }
+
+    if (num_names > 0 && parser->isSet(all_option))
+    {
+        cerr << "Cannot specify name";
+        if (num_names > 1)
+            cerr << "s";
+        cerr << " when --all option set\n";
+        return ParseCode::CommandLineError;
+    }
+
+    for (const auto& arg : parser->positionalArguments())
+    {
+        auto entry = request.add_instance_name();
+        entry->append(arg.toStdString());
+    }
+
+    return status;
+}
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -385,6 +385,7 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
     QObject::connect(&rpc, &mp::DaemonRpc::on_ssh_info, &daemon, &mp::Daemon::ssh_info, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_start, &daemon, &mp::Daemon::start, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_stop, &daemon, &mp::Daemon::stop, Qt::BlockingQueuedConnection);
+    QObject::connect(&rpc, &mp::DaemonRpc::on_suspend, &daemon, &mp::Daemon::suspend, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_delete, &daemon, &mp::Daemon::delet, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_umount, &daemon, &mp::Daemon::umount, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_version, &daemon, &mp::Daemon::version, Qt::BlockingQueuedConnection);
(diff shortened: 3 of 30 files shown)


src/client/cmd/suspend.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    for (const auto& arg : parser->positionalArguments())
+    {
+        auto entry = request.add_instance_name();
+        entry->append(arg.toStdString());
+    }
+
+    return status;
+}
```
Saviq: Man this is a lot of boilerplate, wonder if we could abstract some of that out (in a different PR).

src/daemon/daemon.cpp
```diff
@@ -1439,6 +1444,51 @@ catch (const std::exception& e)
...
+    }
+
+    for (const auto& name : instances_to_suspend)
+    {
+        auto it = vm_instances.find(name);
+        it->second->suspend();
+    }
+
+    return grpc::Status::OK;
+}
```
Saviq: This, too, seems like could be abstracted away, a lot of our commands would do most of the same?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -226,10 +226,13 @@ auto get_domain_definition(virConnectPtr connection, const mp::VirtualMachineDes
     return domain;
 }
 
-auto get_domain_state(virDomainPtr domain)
+auto domain_state_for(virDomainPtr domain)
```
Saviq: Can we move all the renames into a separate PR please?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -240,25 +243,37 @@ mp::LibVirtVirtualMachine::LibVirtVirtualMachine(const mp::VirtualMachineDescrip
...
-    state = get_domain_state(domain.get());
+    state = domain_state_for(domain.get());
+}
+
+mp::LibVirtVirtualMachine::~LibVirtVirtualMachine()
+{
+    update_suspend_status = false;
+
+    if (state == State::running)
+        suspend();
```
Saviq: So we suspend on deletion of `LibVirtVirtualMachine`? But only if it's running? Are we sure that's right? And that it shouldn't shut down instead?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -269,12 +284,33 @@ void mp::LibVirtVirtualMachine::stop()
...
+        if (update_suspend_status)
+        {
+            state = State::suspended;
+            update_state();
+        }
+    }
+
+    monitor->on_suspend();
+}
+
```
Saviq: So we're NOOPing when it's shut down, but user asked to suspend? I wonder if that shouldn't be an error instead?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -158,6 +194,18 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
...
+            {
+                mpl::log(mpl::Level::info, vm_name, "VM suspending");
+            }
+            else if (event.toString() == "RESUME")
+            {
+                mpl::log(mpl::Level::info, vm_name, "VM suspended");
+                if (state == State::suspended || state == State::running)
+                {
+                    vm_process->kill();
+                }
```
Saviq: Is that the right event to kill the vm on? `RESUME`¿?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -190,17 +243,29 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
 
 mp::QemuVirtualMachine::~QemuVirtualMachine()
 {
-    remove_tap_device(QString::fromStdString(tap_device_name));
-
     update_shutdown_status = false;
-    shutdown();
+    if (state == State::running)
+        suspend();
```
Saviq: Suspend/resume is faster than booting it up again?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -224,11 +289,29 @@ void mp::QemuVirtualMachine::shutdown()
     }
     else
     {
-        vm_process->terminate();
+        vm_process->kill();
```
Saviq: Why this change?

src/platform/backends/qemu/qemu_virtual_machine.h
```diff
@@ -22,9 +22,9 @@
 #include <multipass/optional.h>
 #include <multipass/virtual_machine.h>
 
+#include <QStringList>
+
 class QProcess;
-class QFile;
-class QString;
```
Saviq: This some drive-by cleanup?

src/rpc/multipass.proto
```diff
@@ -156,7 +157,8 @@ message InstanceStatus {
         STOPPED = 3;
         DELETED = 4;
         DELAYED_SHUTDOWN = 5;
-        UNKNOWN = 6;
+        SUSPENDED = 6;
+        UNKNOWN = 7;
```
Saviq: Should we maybe avoid breaking compatibility like this?

src/daemon/daemon.cpp
townsend2010: Yep, I think there are many places where we can do refactoring and clean up the code.  Multipass just grew so quickly that some things have gotten out of control;)

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: Sure, I'm a big proponent of focused PR's and I'm not really sure why I let these slip in.  I think I did this early on and never backed them out:)  Anyways, I'll back these out and propose later.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: So, what I'm trying to accomplish here is when the daemon is `SIGTERM`'d with running VM's, like when a `snap refresh` is done, we can at least save the state of the VM and then when the daemon restarts, the VM will be started using the saved image.

If a VM is being deleted, the daemon will explicitly shut it down if it is running before destroy the `LibVirtVirtualMachine` object.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Yes, by far.  On my machine boot takes ~35-40s, whereas resume from a saved memory snapshot takes ~5-10s.  Also, saving the memory snapshot takes less time than cleanly shutting down the machine too.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: I have found in some cases that the qemu process is in some state where it ignore the `SIGTERM`, so we have to wait 30s before Qt will actually just kill the process.  Since we really don't care about the process at this point, a `kill()` just gets the job done and we don't have to wait.

src/platform/backends/qemu/qemu_virtual_machine.h
townsend2010: Yes, as I mentioned earlier, I like focused PR's, but I knowingly did this here since it's quite benign.  But if you prefer, I can also revert this and propose in another PR.

src/rpc/multipass.proto
townsend2010: Yeah, we haven't been too worried about compatibility since the one and only client and daemon have been in lock step, but yes, at some point care should be taken when updating things like this.  It would probably be better to have `UNKNOWN` first in the enum, and then it will look cleaner when adding new states, etc.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: Yep, missed that...but it should be handled in the daemon, so will fix that there.

townsend2010: I fixed this in the recent push by adding some log messages when this situation occurs.

src/platform/backends/qemu/qemu_virtual_machine.h
townsend2010: As mentioned on IRC, it's ok to leave this.

Saviq:
I managed to get it into a state where it thinks it's suspended, but seems the VM wasn't actually killed:
```
$ multipass info snapcraft-multipass                                                                                                                                                                 
Name:           snapcraft-multipass
State:          SUSPENDED
IPv4:           --
Release:        --
Image hash:     072d1fef3479 (Ubuntu 16.04 LTS)
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         /home/michal/dev/canonical/multipass/multipass => /home/multipass/project
$ ps aux | grep qemu
root     10716  9.9  5.0 3068700 410532 ?      Sl   20:34   0:03 qemu-system-x86_64 --enable-kvm -hda /var/snap/multipass/common/data/multipassd/vault/instances/snapcraft-multipass/ubuntu-16.04-server-cloudimg-amd64-disk1.img -drive file=/var/snap/multipass/common/data/multipassd/vault/instances/snapcraft-multipass/cloud-init-config.iso,if=virtio,format=raw,snapshot=off,read-only -smp 2 -m 2G -device virtio-net-pci,netdev=hostnet0,id=net0,mac=52:54:00:d0:2d:d9 -netdev tap,id=hostnet0,ifname=tap-c4218ab4e59,script=no,downscript=no -qmp stdio -chardev null,id=char0 -serial chardev:char0 -nographic -loadvm suspend                                                                                               
```

What I did to get here:
- start the instance
- run screen inside the instance, with a while loop printing `date` every second
- suspend it
- resume it
- suspend again

townsend2010:
@Saviq,

Hmm, I was able to recreate the same log output as you, but I'm not able to make multipassd think the instance is suspended when it's not.  However, I wonder if I should only switch state to suspended only **after** the memory snapshot steps have completed.

I'll look into the multipassd crash when switching drivers when an instance is a suspended state.  I should note that switching drivers will not resume an instance in a suspended state since the methods used to accomplish this are different between the drivers.

townsend2010:
Ok, found the issue and worked around it with the last commit.

snap/hooks/configure
```diff
@@ -12,6 +12,19 @@ driver=${driver:-QEMU}
...
+    if [ $? -ne 0 ]; then
+        echo "Cannot determine status of instances. Not switching driver."
+        exit 1
+    fi
+
+    if /snap/bin/multipass list --format=csv | grep --quiet -E 'STOPPED'; then
+        echo "$driver" > $SNAP_COMMON/driver
+        snapctl restart $SNAP_NAME
+    else
+        echo "All instances need to be in the stopped state."
```
Saviq:
```suggestion
        echo "All instances need to be stopped."
```

snap/hooks/configure
```diff
@@ -12,6 +12,19 @@ driver=${driver:-QEMU}
...
+        echo "Please save any data, and then stop them before proceeding."
```
Saviq:
```suggestion
        echo "Please save any data and stop the instances before proceeding:"
        echo
        echo "multipass stop --all"
```

snap/hooks/configure
```diff
@@ -12,6 +12,19 @@ driver=${driver:-QEMU}
 driver_saved=${driver_saved:-QEMU}
 
 if [[ $driver != $driver_saved ]]; then
-    echo "$driver" > $SNAP_COMMON/driver
-    snapctl restart $SNAP_NAME
+    /snap/bin/multipass list &>/dev/null
```
Saviq:
```suggestion
    /snap/bin/multipass list &>/dev/null || /bin/true
```

snap/hooks/configure
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
```
Saviq: Please undo that.

snap/hooks/configure
```diff
@@ -12,6 +12,19 @@ driver=${driver:-QEMU}
...
-    echo "$driver" > $SNAP_COMMON/driver
-    snapctl restart $SNAP_NAME
+    /snap/bin/multipass list &>/dev/null
+
+    if [ $? -ne 0 ]; then
+        echo "Cannot determine status of instances. Not switching driver."
+        exit 1
+    fi
+
+    if /snap/bin/multipass list --format=csv | grep --quiet -E 'STOPPED'; then
```
Saviq: You have to make sure that all of them are stopped.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -190,17 +244,32 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
...
+    vm_process->waitForFinished();
 }
 
 void mp::QemuVirtualMachine::start()
 {
     if (state == State::running)
         return;
 
+    if (state == State::suspending)
+        throw std::runtime_error("cannot start an instance while suspending");
```
Saviq:
```suggestion
        throw std::runtime_error("cannot start the instance while suspending");
```

townsend2010:
Ok, updated:)

Issue #443, opened 2018-10-16
Incomplete Build Dependencies


pekof:
In README.md there is line under Build Dependencie with missing parameter:
`mk-build-deps -s sudo -i`
It needs package name and/or control file.

Saviq:
It defaults to `debian/control`, so if you're in the source directory it should Just Work™?

pekof:
Thank you.
So line from next paragraph `cd <multipass>` should be executed before `mk-build-deps`.
That's unclear in README.md

Saviq:
Here's a PR: #445, please comment on it if it makes it more clear. Thanks!

pekof:
That's clearer.
Good job 👍

pekof:
FYI Ubuntu 18.04LTS don't have all required build dependencies (missing:  libprocess-cpp-dev).
So I added 16.04LTS to sources.list to be able to install multipass-build-deps_1.0_all.deb

I assume you don't see that issue if you don't do clean install of Ubuntu 18.04LTS

Saviq:
Hmm... It's certainly there:

https://launchpad.net/ubuntu/+source/process-cpp

Maybe you didn't have `universe` enabled?

pekof:
Code name for Ubuntu 18.04 LTS is bionic.
Somehow that package skipped this release, so I'm using  libprocess-cpp-dev from xenial.

pekof:
I'm currently trying to build multipass and there are some required dev  packages still missing libsystemd and  libvirt.

Cmake output:
`-- Checking for module 'libsystemd'
--   No package 'libsystemd' found
`

PR #448, opened 2018-10-18
[readme] replace building instructions with snapcraft (Fixes #447)

### LINKED ISSUE #447: Missing buld requirements libvirt and libsystemd
Cmake returned errors 
```
--   No package 'libvirt' found
--   No package 'libsystemd' found
```

I fixed errors by editing `debian/control` and rerunning `mk-build-deps -s sudo -i`

```diff --git a/debian/control b/debian/control
index 01efd40..a51236d 100644
--- a/debian/control
+++ b/debian/control
@@ -18,3 +18,5 @@ Build-Depends: build-essential,
                npm,
                pkg-config,
                qtbase5-dev,
+               libvirt-dev,
+               libsystemd-dev,
```

Saviq:
Our build setup got complex enough that it's best to just develop the
snap instead of building manually.


townsend2010:
Perhaps we should move the old build instructions to a HACKING.md file for those who still want to build from source directly?

README.md
```diff
@@ -38,34 +38,20 @@ multipass help <command>
...
-mk-build-deps -s sudo -i
+sudo snap install snapcraft
 ```
 
-## Building
+## Build and install the snap
 
 ```
 cd <multipass>
-git submodule update --init --recursive
```
gerboland: submodule pull still needed

README.md
```diff
@@ -38,34 +38,20 @@ multipass help <command>
...
-mkdir build
-cd build
-cmake ../
-make
+snapcraft
+snap install --dangerous --classic multipass*.snap
```
gerboland: "snap try" might be nice to mention as is more convenient for development

Saviq:
Yeah problem is you can't build on either pristine 18.04 or 16.04... The former is missing `process-cpp` (#443) and the latter has old Qt...

README.md
```diff
@@ -38,34 +38,20 @@ multipass help <command>
...
 Here's a set of steps to build and run your own build of multipass:
 
-## Build Dependencies
+## Install snapcraft
 
 ```
-cd <multipass>
-apt install devscripts equivs
-mk-build-deps -s sudo -i
+sudo snap install snapcraft
```
gerboland: --classic needed

gerboland: after I install snapcraft, my $PATH is not updated to add /snap/bin, so the next snapcraft command fails unless I know about that/open new shell

README.md
gerboland: add "git clone https://github.com/CanonicalLtd/multipass" first too?

README.md
```diff
@@ -38,34 +38,20 @@ multipass help <command>
...
+## Build and install the snap
 
 ```
 cd <multipass>
-git submodule update --init --recursive
-mkdir build
-cd build
-cmake ../
-make
+snapcraft
```
gerboland: Failed on Bionic due to wrong qt version. cleanbuild should do  - but then needs lxd instructions added.

pekof: bionic (18.04) is kind of mess, just tried building on clean install and got error
`Failed to fetch stage packages: Error downloading packages for part 'libqt5core5a': The package 'libdouble-conversion1v5' was not found..`
I already know that libprocess-cpp-dev is not available on bionic.

IMHO solution is to add xenial (16.04) to sources.list

townsend2010: Yeah, I definitely would not recommend building the snap directly on the host system.  I would definitely  recommend adding `SNAPCRAFT_BUILD_ENVIRONMENT=multipass` or `SNAPCRAFT_BUILD_ENVIRONMENT=lxd` here.

Saviq: I believe that's taken care of by adding `base:` to snapcraft.yaml (which I now did).

pekof:
So what is correct build environment?
According to https://forum.snapcraft.io/t/debugging-building-snaps/6274 it should be Ubuntu 16.04.
If I have 18.04 I need to snap install multipass and snapcraft, define `SNAPCRAFT_BUILD_ENVIRONMENT=multipass` and run `snapcraft`?

Saviq:
@pekof yeah, if you want to build on 18.04, using a build environment is the right approach.

pekof:
I tried that     `$ snapcraft` got
`SNAPCRAFT_BUILD_ENVIRONMENT must be one of: 'host' or 'lxd'.`

then i tried `snapcraft cleanbuild ` and got 
```
 ... on device: '/home/multipass/multipass/parts/kvm-support/build/src'"), ('/home/multipass/multipass/parts/kvm-support/src/CMakeLists.txt', '/home/multipass/multipass/parts/kvm-support/build/CMakeLists.txt', '[Errno 28] No space left on device')]
An error occurred when trying to execute "sh -c 'cd multipass; /snap/bin/snapcraft snap --output multipass_git_amd64.snap'" with 'multipass': returned exit code 1.
```
I don't know what was size of instance at maximum point, but I assume it was at most 5GB.
I will now try with `SNAPCRAFT_BUILD_ENVIRONMENT=lxd` and `snapcraft` without `cleanbuild`.

BTW I manged to build multipass usng cmake and make on bionic with added xenial sources which make it ok environment for hacking.

Saviq:
@pekof what version of snapcraft do you have (`snapcraft --version`)?

The disk space... how much free space do you have in wherever your LXD instances go? `lxc storage list` should help to find where that is.

Mikerr1111:
@townsend2010 for the old build instructions I was thinking a link to a project Wiki would be nice in the Readme. So there would be the new instructions in the readme and right below there will be a link to a Wiki for those that want the old build instructions.

Saviq:
I'm not sure what to do with this then… It's again impossible to build on `disco` here, we're not testing building anything else than the snap (and I'd say that's on purpose)…

PR #450, opened 2018-10-18
rpc/mount: Use the grpc map type for the uid/gid maps

townsend2010:


townsend2010:
I see a few more places I can refactor with this, so going to hold off on being merged for now.

PR #463, opened 2018-10-25
Restart cmd

### LINKED ISSUE #217: Multipass should have a restart command next to start and stop
At times you really just want to restart a machine multipass manages. At the moment you have to do a `multipass stop my-machine && multipass start my-machine`. Having a short version of this as `multipass restart my-machine` would be pretty handy.

ricab:
This fixes #217. It implements a restart command by delegating on the existing stop and start commands.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -49,20 +49,23 @@
 
 #include <stdexcept>
 #include <unordered_set>
+#include <functional>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
-using namespace std::chrono_literals;
 
 namespace
 {
+
+using namespace std::chrono_literals;
+
 constexpr auto category = "daemon";
 constexpr auto instance_db_name = "multipassd-vm-instances.json";
 constexpr auto uuid_file_name = "multipass-unique-id";
 constexpr auto metrics_opt_in_file = "multipassd-send-metrics.yaml";
 constexpr auto reboot_cmd = "sudo reboot";
-constexpr auto up_timeout = 2min; // TODO: use this in places that wait for ssh to be up
+constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
@@ -957,7 +960,7 @@ try // clang-format on
                 case mp::VirtualMachine::State::starting:
                     return mp::InstanceStatus::STARTING;
                 case mp::VirtualMachine::State::restarting:
-                    return mp::InstanceStatus::RESTARTING; // TODO @ricab confirm this is enough
+                    return mp::InstanceStatus::RESTARTING;
                 case mp::VirtualMachine::State::running:
                     return mp::InstanceStatus::RUNNING;
                 case mp::VirtualMachine::State::delayed_shutdown:
@@ -1086,7 +1089,7 @@ try // clang-format on
         case mp::VirtualMachine::State::starting:
             return mp::InstanceStatus::STARTING;
         case mp::VirtualMachine::State::restarting:
-            return mp::InstanceStatus::RESTARTING; // TODO @ricab confirm this is enough
+            return mp::InstanceStatus::RESTARTING;
         case mp::VirtualMachine::State::running:
             return mp::InstanceStatus::RUNNING;
         case mp::VirtualMachine::State::delayed_shutdown:
@@ -1332,46 +1335,6 @@ catch (const std::exception& e)
     return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), "");
 }
 
-grpc::Status mp::Daemon::restart(grpc::ServerContext* context, const RestartRequest* request,
-                                 grpc::ServerWriter<RestartReply>* server) // clang-format off
-try
-{
-    mpl::ClientLogger<RestartReply> logger{mpl::level_from(request->verbosity_level()), *config->logger, server};
-
-    const auto tgts_and_st = process_targets(request->instance_name(), vm_instances, deleted_instances);
-    const auto& tgts = tgts_and_st.first; // use structured bindings instead in C++17
-    auto& tgts_st = tgts_and_st.second;  // idem
-
-    if(!tgts_st.ok())
-        return tgts_st;
-
-    // reboot all targets
-    // no reuse of status vars to avoid assignment (and enable RVO)
-    auto reboot_st = cmd_vms(tgts, [this](VirtualMachine& vm, const VMSpecs& specs){
-        if(vm.state == VirtualMachine::State::delayed_shutdown)
-            delayed_shutdown_instances.erase(vm.vm_name);
-
-        if(!vm.is_running())
-            return grpc::Status{grpc::StatusCode::INVALID_ARGUMENT,
-                                fmt::format("instance \"{}\" is not running", vm.vm_name), ""};
-
-        return ssh_reboot(vm.ssh_hostname(), vm.ssh_port(), specs.ssh_username, *config->ssh_key_provider);
-    });
-
-    if(!reboot_st.ok())
-        return reboot_st;
-
-    // wait for all the targets to be back up
-    return cmd_vms(tgts, [this](VirtualMachine& vm, const VMSpecs&){
-        vm.wait_until_ssh_up(up_timeout);
-        return grpc::Status::OK;
-    });
-}
-catch(const std::exception& e)
-{
-    return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), "");
-}
-
 grpc::Status mp::Daemon::start(grpc::ServerContext* context, const StartRequest* request,
                                grpc::ServerWriter<StartReply>* server) // clang-format off
 try // clang-format on
@@ -1609,7 +1572,37 @@ catch (const std::exception& e)
     return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), "");
 }
 
-// TODO @ricab move restart here
+grpc::Status mp::Daemon::restart(grpc::ServerContext* context, const RestartRequest* request,
+                                 grpc::ServerWriter<RestartReply>* server) // clang-format off
+try
+{
+    mpl::ClientLogger<RestartReply> logger{mpl::level_from(request->verbosity_level()), *config->logger, server};
+
+    auto tgts_and_st = process_targets(request->instance_name(), vm_instances, deleted_instances);
+    const auto& tgts = tgts_and_st.first; // use structured bindings instead in C++17
+    auto& st = tgts_and_st.second; // idem
+
+    if(st.ok())
+    {
+        using namespace std::placeholders; // for _1, _2
+        st = cmd_vms(tgts, std::bind(&Daemon::reboot_vm, this, _1, _2)); // 1st pass to reboot all targets
+
+        if(st.ok())
+        {
+            st = cmd_vms(tgts, [this](auto& vm, const auto&){
+                vm.wait_until_ssh_up(up_timeout); // 2nd pass waits for them to be up
+                return grpc::Status::OK;
+            });
+        }
+    }
+
+    return st;
+}
+catch(const std::exception& e)
+{
+    return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), "");
+}
+
 
 grpc::Status mp::Daemon::delet(grpc::ServerContext* context, const DeleteRequest* request,
                                grpc::ServerWriter<DeleteReply>* server) // clang-format off
@@ -1898,6 +1891,18 @@ void mp::Daemon::start_mount(const VirtualMachine::UPtr& vm, const std::string& 
     mount_threads[name][target_path] = std::move(sshfs_mount);
 }
 
+grpc::Status mp::Daemon::reboot_vm(VirtualMachine& vm, const VMSpecs& specs)
+{
+    if(vm.state == VirtualMachine::State::delayed_shutdown)
+        delayed_shutdown_instances.erase(vm.vm_name);
+
+    if(!vm.is_running())
+        return grpc::Status{grpc::StatusCode::INVALID_ARGUMENT,
+                            fmt::format("instance \"{}\" is not running", vm.vm_name), ""};
+
+    return ssh_reboot(vm.ssh_hostname(), vm.ssh_port(), specs.ssh_username, *config->ssh_key_provider);
+}
+
 
 grpc::Status
 mp::Daemon::cmd_vms(const std::vector<std::string>& tgts,
@@ -1908,7 +1913,7 @@ mp::Daemon::cmd_vms(const std::vector<std::string>& tgts,
     */
     for(const auto& tgt : tgts)
     {
-        auto st = cmd(*vm_instances.at(tgt), vm_instance_specs.at(tgt));
+        const auto st = cmd(*vm_instances.at(tgt), vm_instance_specs.at(tgt));
         if(!st.ok())
             return st; // Fail early
     }
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -66,6 +66,12 @@ constexpr auto uuid_file_name = "multipass-unique-id";
 constexpr auto metrics_opt_in_file = "multipassd-send-metrics.yaml";
 constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
+constexpr auto term_sshd_retries = 10;
+constexpr auto term_sleep_factor = 10ms;
+constexpr auto term_root_sshd_cmd = "sudo pkill -SIGTERM -P1 -f `which sshd`";
+constexpr auto kill_root_sshd_cmd = "sudo pkill -SIGKILL -P1 -f `which sshd`"; /* Once the root sshd has been killed,
+we want pkill to fail. Since the first orphaned child is adopted by ppid 1, matching by parent is not enough, hence
+the '-f `which sshd`' part. */
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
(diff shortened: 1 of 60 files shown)


ricab:
Hi @Saviq, thank you for the feedback! I see, honestly I did not consider multiple client implementations. Yes, I suppose it makes sense to keep the logic on the daemon then. 

I would normally have discussed plans before jumping to an implementation, but I did not want to pledge, as I am not familiar with the code and might find it required more time than I was able to provide. I regret that now, I should have at least subscribed to the issue, but didn't think of that.

Let me coordinate with @pekof on the issue then.

townsend2010:
Hi @ricab,

I appreciate the contribution too:)  But, yeah, I agree with @Saviq that this should definitely be implemented in the daemon.  As a hint, there is a `RESTARTING` state that is used for when a user issues a `restart` from within an instance, so that may be a good place to look on ideas on how to implement this:)

Also, I saw your comment in the code that some refactoring in the client would be good and I couldn't agree more.  I plan on in the near future refactoring common client code into its own lib so bringing up a new command is not so cumbersome.

ricab:
Hi @townsend2010, thanks for the feedback and hint :)

Yeah, I wanted to do this with minimal intervention and, seeing the issue description, I had hoped that a top-layer stop-start combination would be enough. Admittedly, I was not thinking of multiple clients and I felt the additional logic at the client was less than if there was an additional exchange with the server. True, the required procedures (command line parsing, RPC, error handling, the "spinner", etc.) already exist in identical commands; so, in principle, little *information* has to be added to the client in a server-side-restart approach. In practice, though, the code to do this currently needs to be duplicated, barring a significant refactoring, which is beyond my free time ATM. This was what I managed to bypass with the client-side approach: by delegating on identical commands I was effectively reusing most of their logic.

But, as I said above, I recognize the client is *in principle* dumber with the server-side restart, so I get your point. I will try to have a look at this soon.

ricab:
OK, so I had a look at the daemon code and I find there is an analogous choice to be made: 
 `Daemon::restart` can be implemented by either

1. delegating on the existing sibling slots `Daemon::stop` and `Daemon::start`, or
2. duplicating most of the code in `start`, but relying on a `VirtualMachine::restart` instead (currently missing), or
3. factoring common code out to new (private) methods that the command slots delegated on.

Option 3 would perhaps be ideal, but its scope extends beyond the current issue (such a refactoring would encompass the other commands as well). It can always be done separately, after `restart` is in place (I'd be happy to do it if I had the time :wink: )

Option 2 would allow setting the state as `restarting`, but it would require adding a restart method to `VirtualMachine` and its implementations. I wonder if that is enough to justify the additional intrusion and complexity (after all, a restart is just a stop/start at some level down the stack). Option 1 reduces duplication by maximizing code reuse within exiting structures, but instances state would say "RUNNING" and "STARTING" instead of "RESTARTING".

What's your take on this? Which option would you prefer? Is there a better alternative?

Cheers

ricab:
I thought about this a bit more and I suppose option 2 above could also be done by running `sudo reboot` directly form the daemon, (through ssh, like `Daemon::info`). Not sure what interfaces qemu/libvirt offer for this otherwise. There is a [system_reset](https://en.wikibooks.org/wiki/QEMU/Monitor#system_reset) in qemu, but I don't think it is what we want (it is a hard reset).

@townsend2010 and @Saviq perhaps a real-time chat would be best to discuss this?

townsend2010:
Hi @ricab,

I think probably the best way to accomplish this in the daemon is, as you say, issue a `reboot` inside the instance and set state to `RESTARTING` and then run wait_until_ssh_up() to set status back to `RUNNING`.

ricab:
@townsend2010 and @Saviq, following up on the chat earlier, I did some debugging and `wait_until_ssh_up` is fine after all. Sorry about the noise.

In my tests, I used `pkill`, which exits with 0 when there is a process match. But that does not mean it actually sent the signal! In particular, *`pkill` returns 0 even if it did not have permission to kill a matching process*. I failed to realize that. I could not `ssh` to the vm after the 0 exit code so I thought I had killed the process. But I had actually forgotten `sudo`. During debugging `sudo reboot` must have had enough time to bring `sshd` down, explaining why I could not connect.

Sorry about the noise. I have a working solution that I want to look at with fresh eyes tomorrow before pushing.

src/daemon/daemon.cpp
```diff
@@ -1790,3 +1912,33 @@ void mp::Daemon::start_mount(const VirtualMachine::UPtr& vm, const std::string&
 
     mount_threads[name][target_path] = std::move(sshfs_mount);
 }
+
+grpc::Status mp::Daemon::reboot_vm(VirtualMachine& vm, const VMSpecs& specs)
```
townsend2010: You should be able to drop `const VMSpecs& specs` here since you can also get the vm's ssh_username via `vm.ssh_username()`.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
     QObject::connect(&rpc, &mp::DaemonRpc::on_stop, &daemon, &mp::Daemon::stop, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_suspend, &daemon, &mp::Daemon::suspend, Qt::BlockingQueuedConnection);
+    QObject::connect(&rpc, &mp::DaemonRpc::on_restart, &daemon, &mp::Daemon::restart, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_delete, &daemon, &mp::Daemon::delet, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_umount, &daemon, &mp::Daemon::umount, Qt::BlockingQueuedConnection);
     QObject::connect(&rpc, &mp::DaemonRpc::on_version, &daemon, &mp::Daemon::version, Qt::BlockingQueuedConnection);
 }
+
+template<typename Targets, typename InstanceMap>
+grpc::Status validate_targets(const Targets& targets,
```
townsend2010: This is some good refactoring here.  However, I think I would prefer a different function name since it wasn't clear to me what 'targets' meant.  Maybe something like `validate_instances` and change template name `Targets` to `Instances`.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+                fmt::format_to(errors, "instance \"{}\" does not exist\n", name);
+            else
+                fmt::format_to(errors, "instance \"{}\" is deleted\n", name);
+        }
+
+    return errors.size() ? grpc_status_for(errors) : grpc::Status::OK;
+}
+
+template<typename Targets, typename InstanceMap>
+auto process_targets(const Targets& targets,
```
townsend2010: Same here as above.  Use something like `process_instances`.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
     QObject::connect(&rpc, &mp::DaemonRpc::on_version, &daemon, &mp::Daemon::version, Qt::BlockingQueuedConnection);
 }
+
+template<typename Targets, typename InstanceMap>
+grpc::Status validate_targets(const Targets& targets,
+                              const InstanceMap& vms,
+                              const InstanceMap& deleted)
+{
+    fmt::memory_buffer errors;
+    for(const auto& name : targets)
```
townsend2010: Just a little nitpick, I would prefer the `for` be bracketed as well.  I know it's a single `if`, but it's multi-line and threw me off a bit.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+    return errors.size() ? grpc_status_for(errors) : grpc::Status::OK;
+}
+
+template<typename Targets, typename InstanceMap>
+auto process_targets(const Targets& targets,
+                     const InstanceMap& vms,
+                     const InstanceMap& deleted)
+-> std::pair<std::vector<typename Targets::value_type>, grpc::Status>
+{   // TODO: use this in commands that currently duplicate the same kind of code
+    auto st = validate_targets(targets, vms, deleted);
```
townsend2010: I would prefer `st` to be named `status` here to be a bit more clear what the variable is about.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+    auto tgts = std::vector<typename Targets::value_type>{};
```
townsend2010: Also, maybe use `valid_instances` instead of `tgts` here.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+                      : grpc::Status{grpc::StatusCode::FAILED_PRECONDITION,
+                                     fmt::format("Could not stop sshd. '{}' exited with code {}", stop_ssh_cmd, ecode),
+                                     proc.read_std_error()};
+}
+
+grpc::Status ssh_reboot(const std::string& hostname, int port,
+                        const std::string& username, const mp::SSHKeyProvider& key_provider)
+{
+    mp::SSHSession session{hostname, port, username, key_provider};
+    stop_accepting_ssh_connections(session); /* This allows us to later detect when the machine
```
townsend2010: Could you please put the comment above `stop_accepting_ssh_connections(session);` and use C++ style comments?

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+    has finished restarting by waiting for SSH to be back up. Otherwise, there would be a race
+    condition, and we would be unable to distinguish whether it had ever been down. */
+
+    auto proc = exec_and_log(session, reboot_cmd);
+    try
+    {
+        auto ecode = proc.exit_code();
+        return grpc::Status{grpc::StatusCode::FAILED_PRECONDITION,
+                            fmt::format("Reboot command exited with code {}", ecode),
+                            proc.read_std_error()}; /* we shouldn't get this far: a
```
townsend2010: Same as above for C++ style comments.

src/daemon/daemon.cpp
```diff
@@ -386,10 +395,91 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
...
+                            successful reboot command does not return */
+    }
+    catch(const mp::ExitlessSSHProcessException&) { /* this is the expected path */ }
```
townsend2010: C++ style comments please

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,38 @@ catch (const std::exception& e)
     return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), "");
 }
 
+grpc::Status mp::Daemon::restart(grpc::ServerContext* context, const RestartRequest* request,
+                                 grpc::ServerWriter<RestartReply>* server) // clang-format off
+try
```
townsend2010: Need a `// clang-format on` here.

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,38 @@ catch (const std::exception& e)
...
+{
+    mpl::ClientLogger<RestartReply> logger{mpl::level_from(request->verbosity_level()), *config->logger, server};
+
+    auto tgts_and_st = process_targets(request->instance_name(), vm_instances, deleted_instances);
```
townsend2010: I'd prefer a more descriptive variable name here.  Maybe something like `instances_and_status`.

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,38 @@ catch (const std::exception& e)
...
+    const auto& tgts = tgts_and_st.first; // use structured bindings instead in C++17
```
townsend2010: Same for variable name here.  `instances` or something descriptive.

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,38 @@ catch (const std::exception& e)
...
+    auto& st = tgts_and_st.second; // idem
```
townsend2010: `status` here to be clear please.

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,38 @@ catch (const std::exception& e)
...
+
+    if(st.ok())
+    {
+        using namespace std::placeholders; // for _1, _2
+        st = cmd_vms(tgts, std::bind(&Daemon::reboot_vm, this, _1, _2)); // 1st pass to reboot all targets
+
+        if(st.ok())
+        {
+            st = cmd_vms(tgts, [this](auto& vm, const auto&){
+                vm.wait_until_ssh_up(up_timeout); /* 2nd pass waits for them
```
townsend2010: C++ style comments, please

ricab:
Hi @townsend2010, thank you for the feedback. I have yet to read your review carefully, but just a few points concerning your top comment:

> Generally, I think some of the variable names and new functions need to be named better

Sure, I will try to improve them.

> Also, please use C++ style commenting for your inline comments that are C style comments.

I usually try to make comments stick to the code they refer to, so that unintended cut-offs are signalled at compilation. [Here|https://akrzemi1.wordpress.com/2015/04/22/a-clever-comment-style/] is an argument for it. To stay below a certain line limit (I thought you use 120 by looking at the code, but I am not sure), many comments take more than one line, which is why I used `/**/`. But I can get rid of that by putting them above the code, if you prefer.

> Lastly, please run clang-format on all of your code to get the style more in line with the rest of the code.

Sure. Is there a specific style or config I should use?

I will reply to inline comments later on.

townsend2010:
Hi @ricab,

Hmm, that is an interesting read on comments:)  I definitely see the point in that now, but I think since the rest of our code is using the C++-style comments, we should stick with them here.  In the future, I'm not opposed to a PR that would change all of the comment styles to the C style...

As far as clang-format, multipass has a .clang-format file in the top level that defines what we use.

ricab:
OK sure, that makes sense.

Thanks for pointing out the file.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

ricab: BTW, I was not entirely happy with "process" here either, but I couldn't think of a better name. Please let me know if you have one.

src/daemon/daemon.cpp
ricab: True, especially with the auto. Done

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: In this case I will use a reconsideration plea :)

I thought it was nice to show that the comment "belongs" to the scope of the catch, although perhaps not worth a 3 or 4-line block:
```
catch
{
    // this is the expected path
}
```
And the two-lines alternative looks strange:

```
catch {   // this is the expected path
}
```

In general, do you prefer no c-style comments at all, in any case? For instance, I also find that clarity and brevity are sometimes improved by localized comments, say, in rvalue function arguments: `f(/* use_cache = */ true); ` (a variable could be used here, but that arguably adds to the noise). Or to signal unused parameters that you still want to name: `void f(Abc& /* alphabet */)`.

Let me know what the policy is and I will react accordingly.

src/daemon/daemon.cpp
ricab: Sorry, missed that.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
ricab: Done.

src/daemon/daemon.cpp
townsend2010: Maybe something like 'find_instances_for(...)` would be better.

src/daemon/daemon.cpp
townsend2010: Well, clang-format is going to muck with this any ways, and there other places in the code where we do something similar to:
```
catch
{
    // this is the expected path
}
```
So it's probably better to just change it to be consistent.

As far as c-style comments, I'm not opposed to them, but we should use them in a consistent manner, like for example, the article you pointed out for reasons to use them.  But if we do that, it should be used in all of the code and introduced as a separate PR.  Also, we use c-style comments for our copyright headers;)

ricab: All right, thanks. Will do that.

src/daemon/daemon.cpp
ricab: Deal, done.

ricab:
Hi @townsend2010 ,

So something came up and I got delayed, but I believe I have addressed all review bits. I did not "resolve conversation[s]" because I was not sure whether that should be up to me.

For clang-format, I don't know if you have a simpler mechanism, like a ready shortcut, or some commit hook, but I did:

```
git log --name-only --pretty=format:"" --author=ricab --stat <root_commit>.. | egrep '*.h|*.cpp' | sort -u | xargs clang-format-6.0 -i -style=file
```

Where `<root_commit>` was the first commit I branched off of. Let me know if you find any loose ends.

townsend2010:
@ricab,

For clang-format, I just put the script found at https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format into my users bin/ directory and then run `git clang-format` only on the files you have touched.  Your clang-format was a little too aggressive:)

Also, I just merged a cli refactoring branch, so you may want to update this branch to account for that.

Otherwise, I'll review the code again:)

ricab:
> Your clang-format was a little too aggressive:)

OK sorry for that, first time I used the tool :) IIUC, your suggestion applies to individual commits. I just employed it on my last commit, but I am not sure how to format multiple commits without formatting whole files, so I left the "aggressive" formatting in. I hope that is OK.

> Also, I just merged a cli refactoring branch, so you may want to update this branch to account for that.

Just did so. If you could confirm everything is as expected, that would be great.

ricab:
Right, sorry again about that. I am reverting the completion file as you request. But I think I only formatted files I touched, no? I thought so, as I included `--author=ricab` in the command above.

townsend2010:
No worries.

The files I saw touched by clang-format only are:
tests/test_libvirt_backend.cpp
src/platform/backends/qemu/qemu_virtual_machine.cpp
src/platform/backends/libvirt/libvirt_virtual_machine.h

But really, it's ok and not necessary to revert those files.  I was just mainly pointing that out as an aside:)

ricab:
@townsend2010 you're right. I did change those files at one point, but only to add some TODO comments for myself that I later removed (stuff like `// TODO @ricab decide what to do about this`). I have now reverted the completions file. Cheers :)

src/daemon/daemon.cpp
```diff
@@ -1504,6 +1594,40 @@ catch (const std::exception& e)
...
+    auto& status = instances_and_status.second;         // idem
+
+    if (status.ok())
+    {
+        status = cmd_vms(instances,
+                         std::bind(&Daemon::reboot_vm, this, std::placeholders::_1)); // 1st pass to reboot all targets
+
+        if (status.ok())
+        {
+            status = cmd_vms(instances, [this](auto& vm) {
```
townsend2010: clang on the Mac build fails:
```
../src/daemon/daemon.cpp:1615:42: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
            status = cmd_vms(instances, [this](auto& vm) {
```

ricab: I was having a look just now. It's a leftover capture from an earlier version. Will remove it now.

ricab:
@townsend2010 I just pushed the fix.

townsend2010:
@ricab,

Ok, thanks.

bors r+

PR #464, opened 2018-10-26
[snap] reduce pull load

Saviq:
Use `override-pull` to avoid pulling in any source when not needed.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -156,6 +156,7 @@ parts:
 
   qemu:
     plugin: nil
+    override-pull: ""
     stage-packages:
     - on amd64: [qemu-system-x86]
     - on i386: [qemu-system-x86]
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -142,7 +142,6 @@ parts:
     - git
     - golang
     - libsystemd-dev
-    source: .
```
sergiusens: I would leave `source: .` here to make this future proof (implicit sources are not a thing in snapcraft 3.0 when using bases)

snap/snapcraft.yaml
```diff
@@ -156,6 +155,7 @@ parts:
 
   qemu:
     plugin: nil
+    override-pull: ""
```
sergiusens: alternatively, create `snap/local` and use that as `source` and remove this stanza. But this implementation is future proof and one line removal away when moving to bases.

Saviq:
bors r=townsend2010,sergiusens

Saviq:
bors r=townsend2010,sergiusens

Issue #461, opened 2018-10-25
Delayed shutdown prevents shelling in


Saviq:
Steps:
```
$ multipass stop <instance> --time +10
$ multipass shell <instance>
shell failed: instance "snapcraft-multipass" is not running
```

It should instead cancel the shutdown.

pekof:
I wouldn't expect shutdown to be canceled without someone issuing cancel command.

PR #465, opened 2018-10-26
shell to machine in delayed-shutdown state

pekof:
Fix bug #461 
ssh_info returns connection data if machine is still running

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -57,6 +57,7 @@ public:
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
     virtual void wait_for_cloud_init(std::chrono::milliseconds timeout) = 0;
     virtual void update_state() = 0;
+    virtual bool is_running() { return current_state() == State::running || current_state() == State::delayed_shutdown; }
 
     VirtualMachine::State state;
     const SSHKeyProvider& key_provider;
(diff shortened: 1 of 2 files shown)


pekof:
@Saviq if you think that's necessary I will do it.

My only issue is what's use case for delayed stop if shell or exec would cancel it.
I can imagine that one user might want machine stopped in 30 minutes and connect to it in meantime.

townsend2010:
@pekof, thanks for the contribution too!

The original reason we recommended cancelling the delayed shutdown was a case were snapcraft would do a build in a Multipass instance and then when it's done, it puts the instance in a delayed shutdown.  At this point, a user shell's/exec's into the instance and does stuff and then all of a sudden, the instance is shutdown from underneath them.

However, after thinking about it, a user poking around a snapcraft instance is probably not the best thing to do and if they do, they are on their own.  snapcraft is the true consumer of that instance and not the user.

So my recommendation is that the consumer of the instance should be the one to issue a `multipass stop --cancel` on the instance or the user who is subverting the consumer should be cognizant of what state the instance is in.  Given that, `shell` and `exec` should **not** cancel a delayed shutdown.

pekof:
@Saviq and @townsend2010 if I add commit that cancels delayed shutdown and you decide that that behavior is not desired you can still merge prior commit?
Or should I wait for decision?

townsend2010:
Hmm, let's try this again...
bors r=saviq

PR #468, opened 2018-11-01
Refactor common cli functions

townsend2010:


src/client/cmd/common_cli.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace
+{
+mp::ReturnCode return_code_for(const grpc::StatusCode& code)
+{
+    return code == grpc::StatusCode::UNAVAILABLE ? mp::ReturnCode::DaemonFail : mp::ReturnCode::CommandFail;
+}
+} // namespace
+
+mp::ParseCode cmd::handle_all_option(ArgParser* parser, std::ostream& cerr)
```
gerboland: This handles either instance name or --all being used, function name not great...

src/client/cmd/common_cli.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return return_code_for(status.error_code());
+}
+
+void cmd::install_sshfs_for(const std::string& instance_name, int verbosity_level, grpc::Channel* rpc_channel,
+                            Rpc::Stub* stub, std::ostream& cout, std::ostream& cerr)
+{
+    std::vector<Command::UPtr> command;
+    command.push_back(std::make_unique<Exec>(*rpc_channel, *stub, cout, cerr));
+
+    auto args = QStringList() << ""
```
gerboland: why have an empty string as first member of the QStringList?

src/client/cmd/list.h
```diff
@@ -24,6 +24,8 @@
 
 namespace multipass
 {
+class Formatter;
```
gerboland: Needed?

src/client/cmd/common_cli.cpp
townsend2010: Hmm, yeah, so maybe something like `check_for_name_and_all_option_conflict()`?  Seems kind of verbose, but that is really what it is doing:)

src/client/cmd/common_cli.cpp
townsend2010: Since we are basically overloading the exec command here, that empty string represents the actual binary name in the list of arguments, like `multipass` when using it from the actual command line,  Since we don't care what it is here, I just used the empty string.  It could have been any string, like `foo` or `dummy` or something.  If it's clearer to put something here, I can do that.

src/client/cmd/list.h
townsend2010: We do since we have a `Formatter*` member and we didn't include the actual header for the `Formatter` class.

src/client/cmd/common_cli.cpp
gerboland: Yep, is verbose but not that bad. I'd vote for it

src/client/cmd/common_cli.cpp
gerboland: Ok, I suspected that. A quick comment explaining that it will being used like that would do the trick.

src/client/cmd/list.h
gerboland: Ok I see now, ta

townsend2010:
Merging manually since bors will fail due to the failing MacOS build.

PR #472, opened 2018-11-02
tests/qemu: Make mock qemu a little more intelligent

townsend2010:
Also add a test for suspend


townsend2010:
Going to go ahead and merge this since the Mac build fails.

PR #484, opened 2018-11-08
common_cli: Need namespace prefix for function parameters

townsend2010:
This fixes build issues with some compilers.


Saviq:
Yups.

bors r+

PR #487, opened 2018-11-12
Fix crash if exception thrown during daemon startup

gerboland:
The "system" loggers have different lifecycle to the client loggers.
Client loggers register/unregister themselves with the
MultiplexingLogger on creation/destruction. However "system" loggers did
not, and would leave a dangling pointer in the MultiplexingLogger if
they were destroyed before it was.

Fix is to treat system logger ownership differently from client, i.e.
have MultiplexingLogger own the system logger.

Crash without this patch can be easily reproduced by adding a throw in
the QemuVirtualMachineFactory constructor.

### DIFF
--- a/src/logging/multiplexing_logger.cpp
+++ b/src/logging/multiplexing_logger.cpp
@@ -21,9 +21,15 @@
 
 namespace mpl = multipass::logging;
 
+multipass::logging::MultiplexingLogger::MultiplexingLogger(UPtr system_logger)
+    : system_logger{std::move(system_logger)}
+{
+}
+
 void mpl::MultiplexingLogger::log(mpl::Level level, CString category, CString message) const
 {
     std::shared_lock<decltype(mutex)> lock{mutex};
+    system_logger->log(level, category, message);
     for (auto logger : loggers)
         logger->log(level, category, message);
 }
--- a/include/multipass/logging/multiplexing_logger.h
+++ b/include/multipass/logging/multiplexing_logger.h
@@ -31,11 +31,13 @@ namespace logging
 class MultiplexingLogger : public Logger
 {
 public:
+    explicit MultiplexingLogger(UPtr system_logger);
     void log(Level level, CString category, CString message) const override;
     void add_logger(const Logger* logger);
     void remove_logger(const Logger* logger);
 
 private:
+    UPtr system_logger;
     mutable std::shared_timed_mutex mutex;
     std::vector<const Logger*> loggers;
 };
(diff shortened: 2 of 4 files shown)


gerboland:
Pet peeve: if you run multipassd via command line, no errors are printed to stdout. Can we do something about that?

gerboland:
OSX CI failed, think infrastructure problem

Saviq:
> Pet peeve: if you run multipassd via command line, no errors are printed to stdout. Can we do something about that?

```
$ multipassd --help
Usage: ./build/bin/multipassd [options]
multipass service daemon

Options:
  -h, --help                                  Displays this help.
  -v, --version                               Displays version information.
  --logger <platform|stderr>                  specifies which logger to use
  -V, --verbosity <error|warning|info|debug>  specifies the logging verbosity
                                              level
  --address <server_name:port>                specifies which address to use
                                              for the multipassd service; a
                                              socket can be specified using
                                              unix:<socket_file>
```

src/logging/multiplexing_logger.cpp
```diff
@@ -21,9 +21,15 @@
...
 
+multipass::logging::MultiplexingLogger::MultiplexingLogger(UPtr system_logger)
+    : system_logger{std::move(system_logger)}
+{
+}
+
 void mpl::MultiplexingLogger::log(mpl::Level level, CString category, CString message) const
 {
     std::shared_lock<decltype(mutex)> lock{mutex};
+    system_logger->log(level, category, message);
```
townsend2010: Just thinking if we should add the raw pointer to `system_logger` to the `loggers` array in the MultiplexingLogger constructor and then not worry about the extra `log()` call here.

Not going to block on that though.

gerboland: I could, but saving copying the raw pointer out of the unique_ptr to another location defeats the safety the unique_ptr gives. So I'd rather not.

gerboland:
@Saviq that's not an error. (With this patch) If an exception is thrown, the exception text is only sent to the journal, not stdout/stderr.

src/logging/multiplexing_logger.cpp
townsend2010: Ok, works for me:)

townsend2010:
Hmm, I see the exception text on stderr when starting multipassd like `multipassd --logger stderr`...

albaguirre:
Ohhh I see why. I failed to account for something in mp::DaemonConfigBuilder::build() throwing before DaemonConfig is actually constructed.

albaguirre:
> @Saviq that's not an error. (With this patch) If an exception is thrown, the exception text is only sent to the journal, not stdout/stderr.

You want to look at daemon_main. I changed the exception handler to use multipass::log, you could double down and also use fmt::print(stderr, "....").

The init sequence is a little trick - that's why the logger is initialized as soon as possible, so it goes to the correct places - if multipass::log is called before a logger is in place, by default will send it to stderr.

PR #492, opened 2018-11-14
Write to /proc/sys instead of using sysctl tool

gerboland:
One less utility to have to ship and confine

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
@@ -19,6 +19,7 @@
 #include "qemu_virtual_machine.h"
 
 #include <multipass/backend_utils.h>
+#include <multipass/logging/log.h>
 #include <multipass/optional.h>
 #include <multipass/utils.h>
 #include <multipass/virtual_machine_description.h>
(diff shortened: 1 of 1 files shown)


src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -93,7 +95,17 @@ void create_tap_device(const QString& tap_name, const QString& bridge_name)
...
-    mp::utils::run_cmd_for_status("sysctl", {"-w", "net.ipv4.ip_forward=1"});
+    // Command line equivalent: "sysctl -w net.ipv4.ip_forward=1"
+    QFile ip_forward("/proc/sys/net/ipv4/ip_forward");
+    if (!ip_forward.open(QFile::ReadWrite))
+    {
+        mpl::log(mpl::Level::warning, "daemon", fmt::format("Unable to open {}", qPrintable(ip_forward.fileName())));
+        return;
+    }
+
+    if (ip_forward.write("1") < 0) {
```
townsend2010: Please put curly brace below:)

gerboland: Done

PR #493, opened 2018-11-14
Avoid using bash script for text processing, use code instead

gerboland:
Confining Bash and its friends is much harder than the actual utility.

### DIFF
--- a/src/platform/backends/backend_utils/backend_utils.cpp
+++ b/src/platform/backends/backend_utils/backend_utils.cpp
@@ -69,7 +69,7 @@ auto virtual_switch_subnet(const QString& bridge_name)
 
     if (subnet.isNull())
     {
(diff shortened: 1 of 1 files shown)


src/platform/backends/backend_utils/backend_utils.cpp
```diff
@@ -50,10 +53,28 @@ bool can_reach_gateway(const std::string& ip)
...
+        if (line.contains(bridge_name))
+        {
+            subnet = line.section('.', 0, 2);
+            break;
+        }
+    }
+
+    if (subnet.isNull())
+    {
+        mpl::log(mpl::Level::warning, "daemon",
```
townsend2010: We don't really care if the subnet exists already or not, so if you really want to print out an information message here, then I would use `debug` as the log level.

PR #496, opened 2018-11-15
daemon: Reduce the number of times current_state() is called

townsend2010:
This will help with overhead on backends where it's necessary to query state of an instance via another daemon.

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -60,7 +60,14 @@ public:
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
     virtual void wait_for_cloud_init(std::chrono::milliseconds timeout) = 0;
     virtual void update_state() = 0;
-    virtual bool is_running() { return current_state() == State::running || current_state() == State::delayed_shutdown; }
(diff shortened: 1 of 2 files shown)


include/multipass/virtual_machine.h
```diff
@@ -60,7 +60,14 @@ class VirtualMachine
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
     virtual void wait_for_cloud_init(std::chrono::milliseconds timeout) = 0;
     virtual void update_state() = 0;
-    virtual bool is_running() { return current_state() == State::running || current_state() == State::delayed_shutdown; }
+    virtual bool is_running(const VirtualMachine::State& present_state)
+    {
+        return present_state == State::running || present_state == State::delayed_shutdown;
+    }
```
gerboland: This function is independent of the object, it's odd to have it be a method of the class. It means you're writing code like stopped_vm->is_running(another_vm_state) which could be true.

I think it would be better as a convenience function outside the VM class.

townsend2010: Yeah, that's a good idea.  Done:)

PR #498, opened 2018-11-20
WIP - Add AppArmor confinement for qemu, qemu-img and dnsmasq

gerboland:
This is step 1 of the required work to fully confine a Multipass snap.

As Multipass manages multiple VMs, it needs to take responsibility for confining them too. This PR does the following:
- create a ConfinementSystem abstract class, with 2 implementations: AppArmor and Unconfined.
- create a ProcessSpec that combines a process command line spec with its AppArmor profile.
- Process wraps QProcess, loading the AppArmor profile (if necessary), and exposing what is necessary to control the process while confined.

All QProcesses of qemu, qemu-img and dnsmasq are altered to use the above architecture.


snap/snapcraft.yaml
```diff
@@ -128,19 +139,30 @@ parts:
...
     source-type: deb
 
+  libssl:
+    plugin: dump
+    source:
+    - on amd64: http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.0g-2ubuntu4.1_amd64.deb
+    - on i386: http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.0g-2ubuntu4.1_i386.deb
+    - on armhf: http://ports.ubuntu.com/ubuntu-ports/pool/main/o/openssl/libssl1.1_1.1.0g-2ubuntu4.1_arm64.deb
+    - on arm64: http://ports.ubuntu.com/ubuntu-ports/pool/main/o/openssl/libssl1.1_1.1.0g-2ubuntu4.1_arm64.deb
+    source-type: deb
```
Saviq: Hmm why this? Doesn't core provide `libssl` for security reasons?

gerboland: I was focused on core16, which provides libssl1.0 only, so I added this. Core18 does provide libssl1.1 though. We could move to that..

gerboland:
@Saviq yes I can definitely be called out on being too pedantic. It's easy to go too far, and much of the argument for padantism is "well it _could_ happen" :)

I tend to be strict whenever dealing with downloaded image files, so verification, conversion and booting them.

I thought the little extra complexity in the qemu-img rule was worth it. Just in case qemuimg verify had a vulnerability that an ill-formed img file could force it to write a file. Perhaps that's too far-fetched, I dunno.

gerboland:
Interesting comment I just found on how OpenStack secures qemu: https://news.ycombinator.com/item?id=18588899

PR #501, opened 2018-11-27
Use wall for delayed shutdown

townsend2010:


src/daemon/daemon.cpp
```diff
@@ -1334,17 +1334,29 @@ try // clang-format on
...
                                 fmt::format("instance \"{}\" is not running", name), "");
         }
 
+        if (vm->state == VirtualMachine::State::delayed_shutdown)
+        {
+            if (delayed_shutdown_instances[name]->get_time_remaining() <= std::chrono::minutes(1))
+            {
+                return grpc::Status(
+                    grpc::StatusCode::FAILED_PRECONDITION,
+                    fmt::format("\"{}\" is in a delayed shutdown. Use 'multipass stop -c {}' to cancel.", name, name),
```
Saviq: I'd word this more like:
> {} is scheduled to shut down within a minute,
> use `multipass stop --cancel {}` to cancel the shutdown.

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -23,20 +23,38 @@
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
-mp::DelayedShutdownTimer::DelayedShutdownTimer(VirtualMachine* virtual_machine) : virtual_machine{virtual_machine}
+namespace
+{
+std::string shutdown_message(const std::chrono::minutes& time_left)
```
Saviq: I wouldn't expect the returned value to include the `wall `.

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -23,20 +23,38 @@
...
+mp::DelayedShutdownTimer::DelayedShutdownTimer(VirtualMachine* virtual_machine, SSHSession&& session)
+    : virtual_machine{virtual_machine}, ssh_session{std::move(session)}
 {
 }
 
 mp::DelayedShutdownTimer::~DelayedShutdownTimer()
 {
     if (shutdown_timer.isActive())
     {
+        ssh_session.exec("wall The system shutdown has been cancelled").exit_code();
```
Saviq: Hmm... "**The** system shutdown…"? Sounds weird?

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -48,19 +66,40 @@ void mp::DelayedShutdownTimer::start(std::chrono::milliseconds delay)
...
                              std::chrono::duration_cast<std::chrono::minutes>(delay).count(),
                              delay > std::chrono::minutes(1) ? "s" : ""));
-        QObject::connect(&shutdown_timer, &QTimer::timeout, [this]() { shutdown_instance(); });
+        ssh_session.exec(shutdown_message(std::chrono::duration_cast<std::chrono::minutes>(delay)));
+
+        time_remaining = delay;
+        std::chrono::minutes time_elapsed{1};
+        QObject::connect(&shutdown_timer, &QTimer::timeout, [this, delay, time_elapsed]() mutable {
+            time_remaining = delay - time_elapsed;
+            ssh_session.exec(shutdown_message(std::chrono::duration_cast<std::chrono::minutes>(time_remaining)));
```
Saviq: Wonder if we should not spam every minute? Maybe once when first scheduled, then at 5 minutes, only then every minute?

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -48,19 +66,40 @@ void mp::DelayedShutdownTimer::start(std::chrono::milliseconds delay)
...
+            {
+                time_elapsed += std::chrono::minutes(1);
+            }
+        });
 
         virtual_machine->state = VirtualMachine::State::delayed_shutdown;
 
-        shutdown_timer.setSingleShot(true);
-        shutdown_timer.start(delay);
+        shutdown_timer.start(delay < std::chrono::minutes(1) ? delay : std::chrono::minutes(1));
```
Saviq: Shouldn't this be `delay <=`?

src/daemon/delayed_shutdown_timer.cpp
townsend2010: Ok, but I did this to keep from having to add `wall` to every session.exec() call we make.  Would renaming the function make it more clear?  Maybe something like shutdown_message_command() or something?

src/daemon/delayed_shutdown_timer.cpp
townsend2010: That is what the `shutdown` command itself writes when it has been cancelled.

Saviq: Ohkay...

Saviq: Just noticed, why the `.exit_code()` here?

src/daemon/delayed_shutdown_timer.cpp
Saviq: Or even pass the ssh session in here?

src/daemon/delayed_shutdown_timer.cpp
townsend2010: The `shutdown` command itself sends a message every minute when the time is set to 15 minutes or less.  Any time over that and you don't see a message at all, even when the time remaining dips below 15 minutes (which seems strange).

But we could be more nice on times greater than 5 minutes as not to peeve the user:)

Saviq: Yeah we don't have to be *that* inspired by `shutdown` ;)

src/daemon/delayed_shutdown_timer.cpp
townsend2010: I think `<=` will accomplish the same thing as `<` in this case since both result in 1 minute timers if `delay` is set to ` minute.

I basically have the check for `delay` for the tests so I don't think it's really important since the delayed shutdown command itself uses minute resolution for the dealy.

src/daemon/delayed_shutdown_timer.cpp
townsend2010: It's a hack to make sure the wall command completes before the dtor finishes.

src/daemon/daemon.cpp
townsend2010: Ok, will do.

src/daemon/delayed_shutdown_timer.cpp
townsend2010: Yeah, that works too:)

src/daemon/delayed_shutdown_timer.cpp
Saviq: Oh and yeah, DUH!

src/daemon/delayed_shutdown_timer.cpp
Saviq: Add a comment so it doesn't get removed as a drive-by fix?

src/daemon/daemon.cpp

src/daemon/delayed_shutdown_timer.cpp
townsend2010: Done

src/daemon/delayed_shutdown_timer.cpp
townsend2010: This should be better now.

src/daemon/delayed_shutdown_timer.cpp
townsend2010: Comment added

Saviq:
Yup, all seems fine.

bors r+

Issue #502, opened 2018-11-29
Transient errors when performing mounts


sergiusens:
During start, I get,
    start failed: The following errors occurred: Removing "/root/project": failed to obtain exit status for remote process: 'id -ng'

Maybe mounting should be more resilient?

Saviq:
I believe this happens when we try to mount and the VM may just not be ready yet for real (it responds to SSH, but doesn't actually allow login).

ricab:
@sergiusens do you get this consistently in any particular setting? Do you know of a simple way to reproduce it? Thanks.

townsend2010:
I don't think this is an issue ssh logins not being ready because other commands have already ran successfully before `id -ng` runs, like `id -nu` and 'mkdir -p` for example.

We have a default 5 second timeout waiting for the exit code, which seems like more than enough time for a command to complete in the instance and return the exit code.  So I wonder if there are some cases where the exit code is simply being lost or the callback for obtaining the exit code never runs??

townsend2010:
After a little more looking at this code:
```
int rc{SSH_OK};
while ((std::chrono::steady_clock::now() < deadline) && rc == SSH_OK && !exit_status)
{
    rc = ssh_event_dopoll(event.get(), timeout.count());
}
```

it's also possible that `ssh_event_dopoll()` returned an error before the timeout and thus, broke out of the loop.  We should definitely check if there is an error in this case and log accordingly.

ricab:
Good point. We would still need to check whether or not there was a timeout, because that is one of the reasons for `ssh_event_dopoll` not to return `SSH_OK`. Strangely, the [documentation](https://git.libssh.org/projects/libssh.git/tree/src/poll.c?h=libssh-0.8.0&id=d66ea0b3b100664916cbab965e79a870e31ac7fe#n868) for `ssh_event_dopoll` isn't clear about the return value when a timeout occurs. But according to the [documentation](https://git.libssh.org/projects/libssh.git/tree/src/poll.c?h=libssh-0.8.0&id=d66ea0b3b100664916cbab965e79a870e31ac7fe#n580) of `ssh_poll_ctx_dopoll`, it must return SSH_AGAIN in that case. 

Perhaps we should also log `errno` to see if it gives any useful indication. I wish there was a way to get some logging out of libssh directly.

PR #519, opened 2018-12-10
Add some logging around mounting in daemon.

ricab:
Some more logging resolution around mounting, which may be useful to diagnose bugs that are hard to reproduce (e.g. #502)

### DIFF
--- a/include/multipass/exceptions/exitless_sshprocess_exception.h
+++ b/include/multipass/exceptions/exitless_sshprocess_exception.h
@@ -28,8 +28,10 @@ namespace multipass
 class ExitlessSSHProcessException : public std::runtime_error
 {
 public:
-    ExitlessSSHProcessException(const std::string& command)
-        : runtime_error(fmt::format("failed to obtain exit status for remote process: '{}'", command)), cmd{command}
+    ExitlessSSHProcessException(const std::string& command, const std::string& cause)
+        : runtime_error(fmt::format("failed to obtain exit status for remote process '{}': {}", command, cause)),
+          cmd{command},
+          cause{cause}
     {
     }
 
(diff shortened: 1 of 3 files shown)


include/multipass/exceptions/exitless_sshprocess_exception.h
```diff
@@ -40,6 +42,7 @@ class ExitlessSSHProcessException : public std::runtime_error
 
 private:
     const std::string cmd;
+    const std::string cause;
```
townsend2010: Not sure why we need to have `cause` as a member of the class since the `runtime_error` ctor is called during the `ExitlessSSHProcessException` ctor and `cause` is never used afterwards.

Along with that, but separate from this PR, why is `cmd` used as well?  I see the `command()` accessor method that uses `cmd`, but I don't see it used anywhere.  Not a blocker for this PR but just curious...

ricab: Both are superfluous indeed. When I added the exception I tried to stick with what was done in `mp::StartException`, which also keeps the instance name in an attribute which I don't think is ever used. This time I just repeated the same thing, but I notice now I forgot to add the corresponding accessor method...

ricab: So, I can either remove both attributes, arguably breaching this PR's scope, or add an accessor method, keeping attributes for no real reason. If you agree I will remove them.

townsend2010: Right, the `name()` accessor is used in the daemon code.  I think it's safe to remove the private members and the accessor method.  If we ever need them in the future, it's trivial to add:)

townsend2010: Thanks!

PR #520, opened 2018-12-10
daemon/delete: Explicitly stop mounts when deleting instances

townsend2010:
This avoids a race when delete & purge will start tearing down the VM itself before mounts are completely stopped and causes unexpected behavior.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1685,19 +1685,14 @@ try // clang-format on
             auto& sshfs_mounts = mount_threads.at(name);
             for (const auto& sshfs_mount : sshfs_mounts)
             {
-                mpl::log(mpl::Level::debug, "delete",
+                mpl::log(mpl::Level::debug, category,
                          fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, name));
                 sshfs_mount.second->stop();
             }
         }
         catch (const std::out_of_range&)
         {
-            mpl::log(mpl::Level::debug, "delete", fmt::format("No mounts to stop for instance \"{}\"", name));
-        }
-        catch (const std::exception& e)
-        {
-            mpl::log(mpl::Level::error, "delete",
-                     fmt::format("Error stopping mount(s) in instance \"{}\": {}", name, e.what()));
+            mpl::log(mpl::Level::debug, category, fmt::format("No mounts to stop for instance \"{}\"", name));
         }
 
         it->second->shutdown();
@@ -1950,7 +1945,7 @@ void mp::Daemon::start_mount(const VirtualMachine::UPtr& vm, const std::string& 
     QObject::connect(mount_threads[name][target_path].get(), &SshfsMount::finished, this,
                      [this, name, target_path]() {
                          mount_threads[name].erase(target_path);
-                         mpl::log(mpl::Level::debug, "mount",
+                         mpl::log(mpl::Level::debug, category,
                                   fmt::format("Mount '{}' in instance \"{}\" has stopped", target_path, name));
                      },
                      Qt::QueuedConnection);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1680,6 +1680,26 @@ try // clang-format on
             delayed_shutdown_instances.erase(name);
         }
 
+        try
+        {
+            auto& sshfs_mounts = mount_threads.at(name);
+            for (const auto& sshfs_mount : sshfs_mounts)
+            {
+                mpl::log(mpl::Level::debug, "delete",
+                         fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, name));
+                sshfs_mount.second->stop();
+            }
+        }
+        catch (const std::out_of_range&)
+        {
+            mpl::log(mpl::Level::debug, "delete", fmt::format("No mounts to stop for instance \"{}\"", name));
+        }
+        catch (const std::exception& e)
+        {
+            mpl::log(mpl::Level::error, "delete",
+                     fmt::format("Error stopping mount(s) in instance \"{}\": {}", name, e.what()));
+        }
+
         it->second->shutdown();
         if (purge)
         {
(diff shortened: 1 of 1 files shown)


src/daemon/daemon.cpp
```diff
@@ -1680,6 +1680,21 @@ try // clang-format on
...
+            for (const auto& sshfs_mount : sshfs_mounts)
+            {
+                mpl::log(mpl::Level::debug, "delete",
+                         fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, name));
+                sshfs_mount.second->stop();
+            }
+        }
+        catch (const std::exception& e)
+        {
+            mpl::log(mpl::Level::debug, "delete", fmt::format("No mounts to stop for instance \"{}\"", name));
```
Saviq: Is this the only reason why the above would throw? I'm worried we're catching too broad here?

townsend2010: I don't think there could be anything other than std::out_of_range thrown in this block.  However, there is probably some place some other exception can be thrown, so I'll tighten this up:)

src/daemon/daemon.cpp
```diff
@@ -1680,6 +1680,26 @@ try // clang-format on
...
+            for (const auto& sshfs_mount : sshfs_mounts)
+            {
+                mpl::log(mpl::Level::debug, "delete",
+                         fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, name));
+                sshfs_mount.second->stop();
+            }
+        }
+        catch (const std::out_of_range&)
+        {
+            mpl::log(mpl::Level::debug, "delete", fmt::format("No mounts to stop for instance \"{}\"", name));
```
ricab: If I am not mistaken, all other logs in `daemon.cpp`  use the constant `category` as the 2nd param. Wouldn't it be better to stick to that here?

src/daemon/daemon.cpp
```diff
@@ -1680,6 +1680,26 @@ try // clang-format on
...
+        }
+        catch (const std::exception& e)
+        {
+            mpl::log(mpl::Level::error, "delete",
```
ricab: idem- category

src/daemon/daemon.cpp
```diff
@@ -1928,7 +1948,12 @@ void mp::Daemon::start_mount(const VirtualMachine::UPtr& vm, const std::string&
     mount_threads[name][target_path] = std::move(sshfs_mount);
 
     QObject::connect(mount_threads[name][target_path].get(), &SshfsMount::finished, this,
-                     [this, name, target_path]() { mount_threads[name].erase(target_path); }, Qt::QueuedConnection);
+                     [this, name, target_path]() {
+                         mount_threads[name].erase(target_path);
+                         mpl::log(mpl::Level::debug, "mount",
```
ricab: idem- category

src/daemon/daemon.cpp
```diff
@@ -1680,6 +1680,26 @@ try // clang-format on
...
+                mpl::log(mpl::Level::debug, "delete",
+                         fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, name));
+                sshfs_mount.second->stop();
+            }
+        }
+        catch (const std::out_of_range&)
+        {
+            mpl::log(mpl::Level::debug, "delete", fmt::format("No mounts to stop for instance \"{}\"", name));
+        }
+        catch (const std::exception& e)
```
ricab: I wonder if this catch is really useful (since the whole function is wrapped in a try-catch). Do you think it would be harder to pinpoint the origin of an exception arising here if it was only caught below?

src/daemon/daemon.cpp
```diff
@@ -1680,6 +1680,26 @@ try // clang-format on
...
+        {
+            mpl::log(mpl::Level::error, "delete",
+                     fmt::format("Error stopping mount(s) in instance \"{}\": {}", name, e.what()));
+        }
```
ricab: If the catch is maintained for logging, shouldn't the exception be re-thrown? `std::exception` is very general and it could signal an important problem (an unrecoverable one, in the worst case). I am not sure we'd want to claim success to the client with only a message in the log in that case.

src/daemon/daemon.cpp
townsend2010: Yeah, I thought about that.  I guess I should use that instead for consistency's sake, but I do think at a later time, we should change that to be tell operation in the daemon logged the message.

For example, category would just be `daemon` and we don't really have a clue in the logs what operation logged that message.

ricab: I see your point. I don't have a strong opinion on category contents yet (i.e. entity vs operation). It is a question of how you split information, I suppose. The message can be constructed in such a way that it conveys all necessary meaning; or you can leave a bit of info to the category. 

But quickly screening relevant messages based on the category works better if it is used consistently. Of course if you see "daemon" everywhere that doesn't help much, but a single "delete" could be confusing ("didn't that originate on the daemon like the other msgs?"). So perhaps a separate PR for that?

src/daemon/daemon.cpp
townsend2010: I suppose letting the general catch below to the work would be roughly the same except that the delete operation would fail if a mount can't be stopped for whatever reason.

src/daemon/daemon.cpp
townsend2010: The re-throw depends on if we want to fail the delete or not if an exception besides std::out_of_range is thrown during the mount stop.  I don't necessarily view that as a fatal error since the instance is going to be shut down and "deleted" any ways.

src/daemon/daemon.cpp
townsend2010: Yep, separate PR.  I'll change this PR to use `category` for now.

ricab: I just noticed the category is not part of the default output of `journalctl`. Wouldn't it be better to clarify in the message? Something like "No mounts to stop for instance ... before deleting"

townsend2010: Hmm, I never noticed that about journalctl.  I imagine there are lots of places that would need better wording in this case.  Let's leave that one alone for now as well:)

src/daemon/daemon.cpp
townsend2010: done

src/daemon/daemon.cpp
townsend2010: BTW, I changed the category string.

src/daemon/daemon.cpp
townsend2010: done

src/daemon/daemon.cpp
townsend2010: Ok, after discussion on IRC and thinking about it some, the only reasons why we would get into this catch block would be for very strange reasons and would be worthy of failing a delete.  That said, I removed this:)

src/daemon/daemon.cpp
townsend2010: This block as been removed altogether, so no need to re-throw:)

PR #523, opened 2018-12-12
[daemon] try `admin` group for socket ownership, too

Saviq:


ricab:
oops, forgot the bors thing

PR #524, opened 2018-12-12
[cmake] mark dirty tree in the version string

Saviq:


CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ add_subdirectory(3rd-party)
 find_package(Qt5Core REQUIRED)
 find_package(Qt5Network REQUIRED)
 
-execute_process(COMMAND git describe
+execute_process(COMMAND git describe --dirty
```
ricab: Won't this be wrong when the version is not dirty (i.e. an exact tag)?

ricab:
For my understanding, what was the motivation for this?

CMakeLists.txt
Saviq: No, see [`man git-describe`](https://git-scm.com/docs/git-describe#git-describe---dirtyltmarkgt)

ricab: Hmm, my earlier comment seems to have been lost. Anyway, I just said you were right, that I had seen the man but had overlooked that bit.

PR #528, opened 2018-12-13
Autocomplete recover instances

ricab:


ricab:
Note: do not merge/review yet, there is still something I need to do here

ricab:
Oh I hadn't updated the page :unamused: Well, separate PR then

PR #537, opened 2018-12-14
Fix recovering all on deleted

### LINKED ISSUE #534: `multipass recover foo` recovers all deleted instances if `foo` is not deleted
```sh
$ multipass list
Name                    State             IPv4             Release
pliable-greyhound       STOPPED           --               Ubuntu 16.04 LTS
definite-shark          DELETED           --               Not Available
$ multipass recover pliable-greyhound
$ multipass list
Name                    State             IPv4             Release
definite-shark          STOPPED           --               Ubuntu 16.04 LTS
pliable-greyhound       STOPPED           --               Ubuntu 16.04 LTS
```

I'm afraid this is my fault: I inadvertently introduced this with #528 

Fix coming shortly.

ricab:
Fixes #534 
This can only be merged after #535 !

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1489,63 +1489,24 @@ try // clang-format on
 {
     mpl::ClientLogger<StopReply> logger{mpl::level_from(request->verbosity_level()), *config->logger, server};
 
-    fmt::memory_buffer errors;
-    std::vector<decltype(vm_instances)::key_type> instances_to_stop;
-    for (const auto& name : request->instance_names().instance_name())
-    {
-        auto it = vm_instances.find(name);
-        if (it == vm_instances.end())
-        {
-            it = deleted_instances.find(name);
-            if (it == deleted_instances.end())
-                fmt::format_to(errors, "instance \"{}\" does not exist\n", name);
-            else
-                fmt::format_to(errors, "instance \"{}\" is deleted\n", name);
-            continue;
-        }
-        instances_to_stop.push_back(name);
-    }
+    auto instances_and_status =
+        find_requested_instances(request->instance_names().instance_name(), vm_instances, deleted_instances);
+    const auto& instances = instances_and_status.first; // use structured bindings instead in C++17
+    auto& status = instances_and_status.second;         // idem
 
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -1277,39 +1270,32 @@ try // clang-format on
...
-        if (it == deleted_instances.end())
+        for (const auto& name : instances)
         {
-            it = vm_instances.find(name);
-            if (it == vm_instances.end()) // noop otherwise
-                fmt::format_to(errors, "instance \"{}\" does not exist\n", name);
+            auto it = deleted_instances.find(name);
+            if(it != std::end(deleted_instances))
+            {
+                it->second->shutdown(); // not sure why this is needed, but leaving it alone
```
townsend2010: I imagine this is a copy & pasta error from back when:)  I think it's safe to remove.

src/daemon/daemon.cpp
```diff
@@ -1910,6 +1898,28 @@ void mp::Daemon::start_mount(const VirtualMachine::UPtr& vm, const std::string&
                      Qt::QueuedConnection);
 }
 
+std::string mp::Daemon::check_instance_alive(const std::string& instance_name) const
```
townsend2010: IMO, the name of this method implies that we are checking that the instance is up and running.  I think maybe a better name would be something like `check_instance_usable` or `check_instance_startable`.

src/daemon/daemon.cpp
ricab: Yes it looks like that. Alright, I will remove it.

src/daemon/daemon.cpp
ricab: You're right, I was struggling with naming there... From your suggestions I prefer `usable`. I am still not sure it conveys the meaning we want (not deleted), but it's certainly usable :stuck_out_tongue: 

Ahh I am sure there is an English word for it, I just can't seem to find it...

townsend2010: Lets go with `operational`:)

ricab: OK, sounds good, will do that.

PR #544, opened 2018-12-18
Make grpc_status_for handle no error case.

ricab:
This makes `grpc_status_for` return `Status::OK` when there are no errors, rather than repeating the same check in each call.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -307,7 +307,7 @@ auto grpc_status_for_mount_error(const std::string& instance_name)
 
 auto grpc_status_for(fmt::memory_buffer& errors)
 {
-    if(!errors.size())
+    if (!errors.size())
         return grpc::Status::OK;
 
     // Remove trailing newline due to grpc adding one of it's own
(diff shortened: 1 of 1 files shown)


ricab:
Done!

ricab:
BTW, @townsend2010 do you use a git hook for clang-format? Or do you actually remember all the time?

townsend2010:
@ricab,

I use https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format and put it in my home's bin/ and it just works.

townsend2010:
Ok, bors, try it again...

bors r+

townsend2010:
Ugh, one last time bors...

bors r+

ricab:
@townsend2010 thanks, I use the same as you, but do you remember to run it manually all the time? Or do you have some sort of git hook that runs it for you before committing?

ricab:
And what happened with bors there? why did this need reopening?

ricab:
OK, so manually but with a habit.

PR #546, opened 2018-12-19
[daemon] use a better `core` image

Saviq:
`/stable/current` is a better place to get the `core` image from, the former is 4 months older.


Saviq:
The images generally only get rebuilt on point releases only, so they will not get updated too often.

PR #547, opened 2018-12-19
Release resources consistently

ricab:
This picks up refactoring changes to how VM resources are released and extends them to other places that do the same thing, simplifying the corresponding code along the way.


townsend2010:
Umm, why did bors close this PR???  No one even asked it to do anything with this PR????

townsend2010:
Oh, I see...it's because the base branch `rework_delete` was deleted by bors, so there is nothing to compare this PR to now.

ricab:
Thanks for catching this @townsend2010 I did not realize the PR would be closed when the base was deleted.

FWIW,  there is an [(unofficial) issue](https://github.com/isaacs/github/issues/959) requesting this use case. Until then, I think we are better off expressing dependencies only in comments for humans. Otherwise there would be PRs slipping through unmerged.

PR #556, opened 2019-01-07
Add cloud-init usage example and link to blog post

aral:
The readme currently does not provide an example of using cloud-init with multipass. This pull request fixes that and links to the blog post on the subject from the Ubuntu blog.


ricab:
It looks good @aral, thank you!

PR #558, opened 2019-01-09
client: Put CLI client files in their own subdirectory

townsend2010:
This is in prep of adding the GUI client.


townsend2010:
Ok, after discussion on IRC, we have decided not to do the big refactoring at this time, and instead, the tray icon status menu code will wrap the cli and parse the output since it is a much quicker approach to get something working.

That said, re-opening this to review:)

ricab:
I postponed reviewing this before because I always saw the travis build as pending here. Do you also see that? When I check the details, though, the mac job actually failed: https://travis-ci.org/CanonicalLtd/multipass/jobs/478381512

I am not sure what to make of it. It appears as though there was some merge conflict...

ricab:
OK, there seems to be some trouble with Travis: I am now getting "We couldn't find the repository
CanonicalLtd/multipass"

Saviq:
Yeah there was a merge conflict on Mac, this needs to be merged manually then.

townsend2010:
Actually, I don't think this is necessary considering that I'm going to just everything regarding the systray and the UI itself in the `multipass` client.  Going to close.

PR #564, opened 2019-01-17
client: Switch to using to Qt signal/slots & event handling

townsend2010:
This is for supporting Qt GUI like the system tray icon support.


gerboland:
IMO there's no reason for the command line client to need an event loop like this. It starts, parses command line, calls the command and when that returns, exits. Introducing this event loop is just adding complexity for no gain.

This "command_done" signal really is just equivalent to the function returning. I understand you needed to introduce it to tell the loop to quit, after the rpc callback fires, but the previous design was just fine.

My next concern is that the implementation of each command in src/client/cmd is very command line specific - they all print strings to stdout/stderr, they all do arg parsing of cli inputs - neither of which are suitable for a GUI program.

So I'm concerned that trying to utilise the existing CLI code in the GUI will result in complicated code that suits neither purpose very well.

But OFC I will reserve judgement until I see how you implement the GUI part, as I need to understand what your plan is.

townsend2010:
@gerboland, I understand your scepticism here.  I did force push a commit to get rid of the signal and just fire a QTimer::singleShot to quit.

Perhaps instead of just proposing this as a separate PR, I'll just make it a commit in a much larger PR that includes the systray support, in hopes that it will be clearer what my intent is here.

townsend2010:
Going to close this and open a new PR later that has all of the code.

PR #565, opened 2019-01-18
Client systray support

townsend2010:
Add a client systray icon and menu.  This allows:
- Getting a list of all instances and state.
- The ability to open a shell to an instance.
- Stopping an instance.
- Getting version information.
- Displaying when an update to Multipass is available.


townsend2010:
To @gerboland's comments in #564, I wonder if the main event loop I added is indeed unnecessary.

My question is if we can use a local `QEventLoop` in `Systray::run()` and Qt will be happy processing events?  If so, then the first commit in this can be removed.  I think I'll experiment with that.

src/client/main.cpp
```diff
@@ -48,7 +48,7 @@ std::string get_server_address()
 
 int main(int argc, char* argv[])
 {
-    QCoreApplication app(argc, argv);
+    QApplication app(argc, argv);
```
gerboland: Hmm, can you check if the command line runs when no display server is running (i.e. "DISPLAY=")

QApplication subclasses QGuiApplication, which loads Qt platform backends for various graphical shells. Those are needed for the systray icon ofc, but for a pure command-line util, I'm unsure if it'll run or just bail.

townsend2010: No, it doesn't:(  Offhand, do you if there is some way to load the platfrom backends after the fact, like when only the systray command runs?  I'll do some research, but hoping you have a quick answer:)

gerboland: Backends load on Q{Gui,}Application construction.

One workaround might be to read the CLI very early looking for systray:
```
if (!argv_has_systray())
    setenv("QT_QPA_PLATFORM", "minimal");

QApplication app(argc, argv);
```

townsend2010: That did the trick.  Thanks!

ricab: Wouldn't it be better to have a separate command altogether then? Just a separate main instead of the if?

townsend2010: @ricab,
I want to use the same binary, ie, `multipass` for the cli, systray icon, and the down-the-road full blown gui, so I think the simple `if` will do the trick.

townsend2010:
BTW, the design of this is based closely on https://github.com/CanonicalLtd/multipass-design#status-menu with added changes as I saw necessary.

ricab:
OK, so I finally managed to build the snap and launch multipass from it... but crash again :/

```
$ multipass systray
This application failed to start because it could not find or load the Qt platform plugin "xcb"
in "".

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.

Reinstalling the application may fix this problem.
Aborted (core dumped)
```

`gdb /snap/multipass/x1/bin/multipass core` [says](https://pastebin.canonical.com/p/stZ5Q8sqyj/).

Some more research [here](https://pastebin.canonical.com/p/wGGVkkTZxq/) and [here](https://pastebin.canonical.com/p/dTHHd9kMYC/) does not help much...

ricab:
OK, the issue seems to be a missing "xinerama" library. Here is the output [from Qt itself](https://pastebin.canonical.com/p/NgR3947Rf6/) with `QT_DEBUG_PLUGINS=1` (see towards the end). So we just need a new multipass dependency I guess...

src/platform/client/client_platform_linux.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <multipass/cli/client_platform.h>
+
+#include <QProcess>
+
+namespace mp = multipass;
+namespace mcp = multipass::cli::platform;
+
+void mcp::open_multipass_shell(const QString& instance_name)
+{
+    QProcess::startDetached("xterm", {"-e", "multipass", "shell", instance_name});
```
Saviq: Add a `-T <instance_name>` so the title is set to the instance name by default? Also `-j` might be desirable.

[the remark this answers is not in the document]
townsend2010: I originally tried it that way and it didn't work.  I didn't look any more into it since this way works:)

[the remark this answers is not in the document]
townsend2010: Right, I had planned on giving extra feedback when the terminal cannot be started for whatever reason.  Ideally, I'd like it to pop up a window with the failure and reason, but for a short term solution, your idea will work.

townsend2010:
> I wouldn't put the instance status verbatim in the menu

Right, I have plans to fix that.  I do think we should be consistent with the output of `multipass list` though, so we should fix it there and then carry it over here.

> It seems to disappear during the transition when the menu is closed. Maybe it's the separator?

Yeah, I'm sure it's the separator and I missed that.  Will fix.

> If there's no way to add the status like designed (right-aligned and greyed out), maybe we could add some Unicode trickery.

I think this is going to be real tricky.  I spent some time trying to make it work like the design with no luck.  I may end up having to subclass QMenu and try some things in that in forcing Qt to display it how we want to.

src/platform/client/client_platform_linux.cpp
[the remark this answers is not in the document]
Saviq: `xterm -e 'multipass shell foo || read'` works, so it's something `QProcess` does wrong?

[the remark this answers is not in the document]
Saviq: Let's avoid windows whenever possible to keep the experience coherent. Multipass is anchored in command line usage so it shouldn't be surprising that we don't have many dialogs and windows. Those are always focus hogs.

townsend2010:
> Detail: systray missing autocompletion

Ack, will add that to the autocompletions.

> OK, the issue seems to be a missing "xinerama" library.

I think once we switch the snap to use `base: core18`, then I can use the `desktop-qt5` remote part and that should pull in all of the dependencies and such that we need.

Saviq:
`clang` on Mac says:
```
../src/client/cmd/systray.cpp:105:77: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
            QObject::connect(instance_actions.back(), &QAction::triggered, [this, name] {
                                                                            ^
../src/client/cmd/systray.cpp:178:24: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
    auto on_success = [this, &list_reply](ListReply& reply) {
                       ^
2 errors generated.
```

src/platform/client/client_platform_linux.cpp
townsend2010: Ack

townsend2010:
@Saviq,

> When there's no instances, there's a stray blank section:

Would it be better to display something like "No instances" or "No instances defined" or something when there are no instances?  When it just has 'About' and 'Quit' there, it kind of looks like it's missing something.

Saviq:
@townsend2010 there ultimately will be a _Launch_, won't there? And the whole pet story…

src/platform/client/client_platform_linux.cpp
townsend2010: Ok, I tried `-T <instance_name>` and the instance name is in the title of the `xterm` window until bash is up and running, and then the title switches to `multipass@<instance_name>: ~` as before.

Saviq: Yeah and that is fine, but if we don't do that, the whole cmdline will be shown (including `|| read` when you add it) in case the command fails.

Saviq:
@townsend2010 right, but that means there will always be an item above About - and there will never be "No instances" in that sense.

ricab:
Some things that I noticed (but could belong in polishing PRs):

* The contents of the menu are sometimes updated, sometimes not:
    * when clicking "start" on an instance and then clicking again to monitor its state: it will go from STOPPED to STARTING, but will never update to RUNNING in the menu. So STOPPED-->STARTING-/->RUNNING. The RUNNING state is only shown after clicking outside and clicking the widget again.
    * the same happens when clicking suspend: RUNNING-->SUSPENDING-/->SUSPENDED
    * I tried sending commands from the terminal and immediately clicking on the icon. The same thing happens.
        * Clicking it right after launch, no new instance appears (until click out and in)
    * stop seems to be good: RUNNING-->STOPPED directly
* When the contents of the menu are updated, it briefly flickers out-of and into existence. It would be nice if the transition could be smooth.

Saviq:
>     * When the contents of the menu are updated, it briefly flickers out-of and into existence. It would be nice if the transition could be smooth.

What desktop environment is this? I don't think we can influence how they deal with this (unless in fact we're deleting the whole menu and recreating it).

ricab:
Regarding:

```
$ multipass systray
```

[This](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables) says:

> `$XDG_RUNTIME_DIR` defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

I noticed it is mentioned [here](https://blog.ubuntu.com/2019/01/31/where-eagles-snap-a-closer-look) and from some quick googling I got the impression it is relevant for security concerns.

ricab:
@Saviq Unity on 16.04 here. Do you see something else?

ricab:
@Saviq other similar menus update *continuously* here (e.g. system monitor)

townsend2010:
@ricab,

Thanks for the feedback.  In general, there is much work to be done on this:)  But to reply to your feedback:

> The contents of the menu are sometimes updated, sometimes not: 

Yes, there is absolutely no polling or event handling when the menu has been opened.  Currently, the only way to refresh the states is to close the menu and then reopen it.   Polling will be implemented first and then eventually some event driven model.

> When the contents of the menu are updated, it briefly flickers out-of and into existence. It would be nice if the transition could be smooth.

Right, this will be fixed.  But what you are seeing is the static menu from the last query, ie, the last time you opened the menu, and then once the client receives the list info from the daemon, it clears the menu and then creates a new one based on what it received.

src/client/gui/gui_cmd.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
+{
+    if (!QSystemTrayIcon::isSystemTrayAvailable())
+    {
+        cerr << "System tray not supported\n";
+        return ReturnCode::CommandFail;
+    }
+
+    QEventLoop loop;
```
gerboland: Is this necessary? There should be an event loop created by the QCoreApplication already that you can use.

src/client/gui/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <multipass/cli/client_common.h>
+
+#include <QApplication>
+
+namespace mp = multipass;
+
+int main(int argc, char* argv[])
+{
+    QApplication app(argc, argv);
+    QCoreApplication::setApplicationName("multipass-gui");
```
gerboland: I'd rather using the app:
```app.setApplicationName("multipass-gui")```

File: src/client/gui/main.cpp [hunk 5]
```diff
@@ -0,0 +1,36 @@
...
+
+    mp::ClientConfig config{mp::client::get_server_address(), mp::RpcConnectionType::ssl,
+                            std::move(mp::client::get_cert_provider()), std::cout, std::cerr};
+    mp::ClientGui client{config};
+
+    return client.run(QCoreApplication::arguments());
```
gerboland: Similarly can use ```app.arguments()```

src/client/gui/argparser.h
```diff
@@ -13,22 +13,24 @@
...
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
- *
  */
 
-#ifndef MULTIPASS_CLI_H
-#define MULTIPASS_CLI_H
+#ifndef ARGPARSER_H
+#define ARGPARSER_H
```
gerboland: Don't we always have "MULTIPASS_" at the start of our #ifdefs?

townsend2010:
@gerboland,

Thanks for the review.  Glad you like this iteration better.  Now that I put in the hard work, I think it's better too:)

> Nitpick: there's many files in src/client/cmd that are getting solely formatting changes, making the diff a lot bigger than it needs to be. Has clang-format changed its mind about some things?

Yes, silly `clang-format` decided to change a bunch of things and I really didn't want to have to keep backing those out.  Not sure why it decided to change now, but it did:/

src/client/gui/gui_cmd.cpp
townsend2010: Sure, I can use the static `QCoreApplication::exec()` here instead of the explicit `QEventLoop`.  Accomplishes the same thing in effect, but is cleaner just using the static.

src/client/gui/main.cpp
townsend2010: Sure, I was just following the cli client's convention here.  But yeah, looks cleaner to use the object's method instead of the static method.

src/client/gui/main.cpp
townsend2010: Yep, see above:)

src/client/gui/argparser.h
townsend2010: Right, I was just blindly following the cli's `ArgParser` header which does this:)  But yeah, will fix it to follow the convention.

src/client/gui/gui_cmd.cpp
townsend2010: Ok, pushed fix.

src/client/gui/main.cpp
townsend2010: Fix pushed.

src/client/gui/main.cpp
townsend2010: Fixed

src/client/gui/argparser.h
townsend2010: Fixed

townsend2010:
To address @gerboland's comment about the unnecessary clang-format changes, I did revert all of those to make the review a bit easier despite what I said previously:)

src/client/gui/client_gui.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef MULTIPASS_CLIENT_GUI_H
+#define MULTIPASS_CLIENT_GUI_H
+
+#include "gui_cmd.h"
+
+#include <multipass/cert_provider.h>
+#include <multipass/rpc/multipass.grpc.pb.h>
+#include <multipass/rpc_connection_type.h>
+
+#include <map>
```
gerboland: unused

src/client/gui/gui_cmd.cpp
```diff
@@ -0,0 +1,299 @@
...
+    {
+        auto name = instance.name();
+        auto state = QString::fromStdString(mp::format::status_string_for(instance.instance_status()));
+
+        auto it = instances_entries.find(name);
+
+        if (it != instances_entries.end())
+        {
+            auto instance_state = it->second.state;
+            if (state == "DELETED")
```
gerboland: This diff changes status_string_for to replace "DELETED" with "Deleted" - so this branch will never be followed afaics

gerboland: It would be safer to do comparisons with the enum returned by instance.instance_status()

src/client/gui/gui_cmd.cpp
```diff
@@ -0,0 +1,299 @@
...
+                instance_menu->setTitle(QString("%1 (%2)").arg(QString::fromStdString(name)).arg(state));
+                instance_menu->clear();
+                set_menu_actions_for(name, state);
+                instances_entries[name].state = state;
+            }
+
+            continue;
+        }
+
+        if (state != "DELETED")
```
gerboland: ditto

src/client/gui/gui_cmd.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+void cmd::GuiCmd::create_menu()
+{
+    tray_icon.setContextMenu(&tray_icon_menu);
+
+    tray_icon.setIcon(QIcon{":images/multipass-icon.png"});
+
+    QObject::connect(&list_watcher, &QFutureWatcher<ListReply>::finished, this, &GuiCmd::update_menu);
+
+    QObject::connect(&menu_update_timer, &QTimer::timeout, [this]() { initiate_menu_layout(); });
```
gerboland: For safety, could you make "this" the third argument, and the lambda the fourth? That'll ensure if the GuiCmd object is deleted, the connection will be broken so the lambda won't fire. It might save a crash on shutdown.

src/client/gui/gui_cmd.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    // Use a singleShot here to make sure the event loop is running before the quit() runs
+    QObject::connect(quit_action, &QAction::triggered, [this] {
+        future_synchronizer.waitForFinished();
+        QTimer::singleShot(0, [] { QCoreApplication::quit(); });
+    });
+
+    QObject::connect(&version_watcher, &QFutureWatcher<VersionReply>::finished, this, &GuiCmd::update_about_menu);
+    QObject::connect(&about_update_timer, &QTimer::timeout, [this]() { initiate_about_menu_layout(); });
```
gerboland: Ditto, add "this" 3rd argument

src/client/gui/gui_cmd.cpp
townsend2010: Yeah, why didn't I just use the enum?????  /me shrugs

ricab:
I just tried this too and I have the same problem with About and "no instances" as Gerry above. It keeps going away.

Also, usually I would expect something with the little arrow right to expand after a short delay of mouse-over, rather than require a click.

Saviq:
> Also, usually I would expect something with the little arrow right to expand after a short delay of mouse-over, rather than require a click. From the design drawings, I would also expect it to expand right (or left, if no space).

All of that the platform decides. In your case, probably GNOME. Remember we're only passing the structure of the menu around, we're not drawing it or receiving any input events.

src/client/cli/main.cpp
```diff
@@ -13,52 +13,26 @@
...
     mp::Console::setup_environment();
     auto term = mp::Terminal::make_terminal();
 
-    auto data_dir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-    auto client_cert_dir = mp::utils::make_dir(data_dir, "client-certificate");
-    auto cert_provider = std::make_unique<mp::SSLCertProvider>(client_cert_dir);
-
-    mp::ClientConfig config{get_server_address(), mp::RpcConnectionType::ssl, std::move(cert_provider), term.get()};
+    mp::ClientConfig config{mp::client::get_server_address(), mp::RpcConnectionType::ssl,
+                            std::move(mp::client::get_cert_provider()), term.get()};
```
Saviq: Building this on macOS results in:
```
../src/client/cli/main.cpp:35:29: error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]                                                                                                                   
                            std::move(mp::client::get_cert_provider()), term.get()};
                            ^
../src/client/cli/main.cpp:35:29: note: remove std::move call here
                            std::move(mp::client::get_cert_provider()), term.get()};
                            ^~~~~~~~~~                               ~
```

townsend2010:
@gerboland, @ricab,

Regarding `About` closing when there are no instances, it's because I'm adding a `Retrieving instances...` message when no instances are found.  I did this in case the operation is taking a long time to complete.  However, since the daemon is really non blocking for many operations now, that message is unnecessary and I will remove it.

That said, any changes to the menu such as instance status updates, etc. will redraw the menu, so if the `About` info is expanded, it will automatically close.  I think we can with that for now.

gerboland:
"About" issue fixed.

New issue: when VM in "Stopped" state, I am shown useless "Open Shell" option. I cannot restart VM from the UI.

Saviq:
> New issue: when VM in "Stopped" state, I am shown useless "Open Shell" option.
It shouldn't be useless, it should start the instance when invoked.

> I cannot restart VM from the UI.
It would be better to replace with _Start_, when stopped, indeed.

gerboland:
> > New issue: when VM in "Stopped" state, I am shown useless "Open Shell" option.
> 
> It shouldn't be useless, it should start the instance when invoked.

Maybe so, but that's not obvious from the UI.

townsend2010:
Regarding `Open shell`, it's analogous to `multipass shell foo`, which will automatically start the instance if it's not running and then once running, it will connect.  This is no different and we all agreed that is the behavior we wanted.  However, I'm not really sure we really want to force the user to go through 2 separate steps in the Ui to open the shell, ie, an explicit `Start`, then make them wait and then an explicit `Open shell`.

Maybe if the instance is `stopped`, then it says `Start & open shell` there?  And if it's already running, it just says, `Open shell` like now?

gerboland:
I like that, it tells you what it will do.

Saviq:
> then it says `Start & open shell` there?

A bit verbose, but OK…

townsend2010:
> A bit verbose, but OK…
I agree, but not sure what else to do since having an active action is not enough to signal to the user they can do something.

Saviq:
> I agree, but not sure what else to do since having an active action is not enough to signal to the user they can do something.

In a well-designed UI, isn't it? ;)

townsend2010:
> In a well-designed UI, isn't it? ;)

Not sure what you mean?  I thought it was clear that if the `Open shell` action is active, ie, clickable, that a user would click on it if the instance is even stopped and then they would see the terminal open with start status and then the shell is connected once the start succeeds.

But @gerboland disagrees and he's a UI guy, so not sure what else to do except say `Start & open shell`.  Otherwise, we make the user do two explicit steps.

ricab:
In this case, I think I'd go with `open shell` only. I normally prefer explicit stuff, but `Start & open shell` seems verbose for a UI button.

I understand it may feel unclear at first, because `start` is also offered to the user separately. But, as @townsend2010 pointed out, `shell` already starts in the CLI. In both cases, I think it is reasonable to view shell opening as an end goal that may require implicit intermediary steps. And client consistency has some non-negligable value IMO. Just my two cents.

ricab:
I just went to check the [design](https://github.com/CanonicalLtd/multipass-design#status-menu) for this. Why not use "Connect" as described there?

townsend2010:
> I just went to check the design for this. Why not use "Connect" as described there?

Because the "Connect" verbiage has been deprecated in favor of "Shell" like in the CLI and the design doc was never updated...at least that is how I took it when we switched the CLI from using `connect` to `shell` for this.

ricab:
@townsend2010 Ah ok, fair enough. I have the feeling you had said so already? Sorry if that was the case, I did not memorize that history bit.

townsend2010:
@ricab, I don't think I mentioned that already:)  Just getting you up to speed and also leaving it open a bit that "Connect" in the GUI is still what is wanted.  Things change and move yet docs aren't updates, so I'm not always sure:)

gerboland:
Overall I'm ok with this, and we can apply any polish in later PRs
bors r+

PR #579, opened 2019-01-28
custom_image_host: Update custom images information when requested

### LINKED ISSUE #569: Transient "Downloaded image hash does not match" error
People have stumbled upon:

```
$ multipass launch snapcraft:core
Launching a VM.
launch failed: Downloaded image hash does not match
```

The likely issue is that somehow we get out of sync with the [`SHA256SUMS`](http://cloud-images.ubuntu.com/minimal/releases/xenial/release/SHA256SUMS) file and trying to compare a previous hash to a new image.

townsend2010:
Fixes #569

### DIFF
--- a/src/daemon/custom_image_host.h
+++ b/src/daemon/custom_image_host.h
@@ -52,7 +52,7 @@ public:
 protected:
     void for_each_entry_do_impl(const Action& action) override;
     VMImageInfo info_for_full_hash_impl(const std::string& full_hash) override;
-    void update_manifests_impl() override;
+    void fetch_manifests() override;
     bool empty() const override;
     void clear() override;
 
--- a/src/daemon/custom_image_host.h
+++ b/src/daemon/custom_image_host.h
@@ -18,11 +18,10 @@
 #ifndef MULTIPASS_CUSTOM_IMAGE_HOST
 #define MULTIPASS_CUSTOM_IMAGE_HOST
 
-#include <multipass/vm_image_host.h>
+#include "common_image_host.h"
 
 #include <QString>
 
-#include <chrono>
 #include <memory>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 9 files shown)


townsend2010:
@ricab,

Well,`VMImageHost` is an interface, so putting common code here will blow up that concept.  I understand what you mean on there is commonality here (and many other places), but we designed many aspects of Multipass to use interfaces and then put the implementation in the derived class.

I'm open to discussing changing this and perhaps we should do that in Malta.  But changing will cause quite a bit of refactoring and I'm not sure if we have the resources for that right now.

But for now, I'm not akin to changing the code design for this fix.

ricab:
@townsend2010 

If we decided (e.g. in Malta) to dedicate some quality time to existing code, I would be happy. But irrespective of that, I think we should strive not to introduce *new* repetition.

I think a sensible approach is to do the refactoring that is strictly required to avoid any new piece of code increasing duplication. Of course, seeing how best to achieve that and considering effort on a case-by-case basis.

As discussed in IRC, I think the intermediate `CommonVMImageHost` that @Saviq suggested is the way to go on this occasion. I will pick up this PR and try to do that. The new class will only get the updating code for now. Other common code that could go in there is deferred to separate PRs.

src/daemon/custom_image_host.h
```diff
@@ -52,9 +53,13 @@ class CustomVMImageHost final : public VMImageHost
     std::vector<std::string> supported_remotes() override;
 
 private:
+    void update_manifest();
     CustomManifest* manifest_from(const std::string& remote_name);
+    std::chrono::seconds manifest_time_to_live;
```
ricab: @townsend2010 Just noticed this is being default constructed (TTL=0). This was just an oversight, correct?

townsend2010: @ricab,

Yes, you are correct, that is wrong.  In the `UbuntuVMImageHost` ctor, that value is passed in and assigned and the same should be done for the `CustomVMImageHost` or whatever common class you are implementing.

ricab:
OK I just pushed something that is kind of half-way to where I think we should be. I want to request feedback on whether or not I should move on to the next part:

1. Merge `CustomManifest` with `SimpleStreamsManifest`
2. Use the same type to map remotes (i.e. "snapcraft"/"daily"/"release") to manifests (ATM, one specialization uses `std::unordered_map`, the other a vector of pairs)
3. Move the manifest structure to the common base, so that we can do away with the [new *virtuals*](https://github.com/CanonicalLtd/multipass/blob/a59c899bdec8d1988c423b6ef7978d3cceec77fa/src/daemon/common_image_host.h#L38). Populating the manifests would still be left to specializations.

This would simplify things a lot, but require a bit more effort. WDYT?

townsend2010:
@ricab,

Thanks for taking this on.  The new common code is better:)

Regarding your questions, I think we should wait on doing any of that.  But I do think we should tackle that at some point and I'm thinking that is when we put in the actual support for custom image repositories the user can set up.

PR #581, opened 2019-01-29
Fix help message for find command.

### LINKED ISSUE #580: Imprecision in `multipass help find` message
```
$ multipass help find
Usage: multipass find [options] [<remote:>][<string>]
Lists available images matching <string> for creating instances from.
With no search string, lists all aliases for supported Ubuntu releases.

Options:
  -h, --help     Display this help
  -v, --verbose  Increase logging verbosity, repeat up to three times for more
                 detail

Arguments:
  string         An optional value to search for in [<remote:>]<string> format,
                 where <remote> can be either ‘release’ or ‘daily’. If <remote>
                 is omitted, it will search ‘release‘ first, and if no matches
                 are found, it will then search ‘daily‘. <string> can be a
                 partial image hash or an Ubuntu release version, codename or
                 alias.
```

`multipass find snapcraft:` is a valid command, so the help message for the argument should not limit remotes to `release` and `daily`. Perhaps adding `snapcraft` to the message is enough for now. 

Additionally, there should be brackets around `<string>` in `[<remote:>]<string>` (as is already the case in the "usage" message.

ricab:
Fixes #580.

The text is more complex than I would hope, but it reflects the behaviour of find.

### DIFF
--- a/src/client/cmd/find.cpp
+++ b/src/client/cmd/find.cpp
@@ -125,11 +125,14 @@ QString cmd::Find::description() const
 mp::ParseCode cmd::Find::parse_args(mp::ArgParser* parser)
 {
     parser->addPositionalArgument("string",
-                                  "An optional value to search for in [<remote:>]<string> format, where "
-                                  "<remote> can be either ‘release’ or ‘daily’. If <remote> is omitted, "
-                                  "it will search ‘release‘ first, and if no matches are found, it will "
-                                  "then search ‘daily‘. <string> can be a partial image hash or an "
-                                  "Ubuntu release version, codename or alias.",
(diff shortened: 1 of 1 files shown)


Saviq:
Yeah agreed, `snapcraft` is an implementation detail. I wonder if we should mention any remotes in particular. Instead I think we should point at the output of `multipass find`.

townsend2010:
> I wonder if we should mention any remotes in particular.

Yeah, I've been thinking this too.

ricab:
OK thanks for the feedback. Continuing discussion in the corresponding issue.

ricab:
back to the drawing board

PR #589, opened 2019-02-04
delayed_shutdown: Only broadcast `wall` messages if SSHSession is valid 

### LINKED ISSUE #539: Stopping an instance in `UNKNOWN` state results in bad error message
When the instance fails to get an IP, it ends up in `UNKNOWN` state and it's unstoppable:
```
$ multipass stop snapcraft-multipass
stop failed: libssh failed to set host option to '': 'Invalid argument in ssh_options_set'
```

townsend2010:
Fixes #539

### DIFF
--- a/include/multipass/delayed_shutdown_timer.h
+++ b/include/multipass/delayed_shutdown_timer.h
@@ -33,7 +33,7 @@ class DelayedShutdownTimer : public QObject
     Q_OBJECT
 
 public:
-    DelayedShutdownTimer(VirtualMachine* virtual_machine, SSHSession&& session);
+    DelayedShutdownTimer(VirtualMachine* virtual_machine, optional<SSHSession>&& session);
     ~DelayedShutdownTimer();
 
     void start(const std::chrono::milliseconds delay);
@@ -47,7 +47,7 @@ private:
 
     QTimer shutdown_timer;
     VirtualMachine* virtual_machine;
-    SSHSession ssh_session;
+    optional<SSHSession> ssh_session;
     std::chrono::milliseconds delay;
     std::chrono::milliseconds time_remaining;
 };
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1972,7 +1972,15 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
     {
         delayed_shutdown_instances.erase(name);
 
-        mp::SSHSession session{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
+        mp::optional<mp::SSHSession> session{mp::nullopt};
+        try
+        {
+            session = mp::SSHSession{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
+        }
+        catch (...)
+        {
+        }
+
         auto& shutdown_timer = delayed_shutdown_instances[name] =
             std::make_unique<DelayedShutdownTimer>(&vm, std::move(session));
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -1972,13 +1972,15 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
     {
         delayed_shutdown_instances.erase(name);
 
-        mp::optional<mp::SSHSession> session{mp::nullopt};
+        mp::optional<mp::SSHSession> session;
         try
         {
             session = mp::SSHSession{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
         }
-        catch (...)
+        catch (const std::exception& e)
         {
+            mpl::log(mpl::Level::info, category,
+                     fmt::format("Cannot open ssh session on \"{}\" shutdown: {}", name, e.what()));
         }
 
         auto& shutdown_timer = delayed_shutdown_instances[name] =
--- a/src/daemon/delayed_shutdown_timer.cpp
+++ b/src/daemon/delayed_shutdown_timer.cpp
@@ -25,22 +25,27 @@ namespace mpl = multipass::logging;
 
 namespace
 {
-void write_shutdown_message(mp::SSHSession& ssh_session, const std::chrono::minutes& time_left, const std::string& name)
+void write_shutdown_message(mp::optional<mp::SSHSession>& ssh_session, const std::chrono::minutes& time_left,
+                            const std::string& name)
 {
+    if (ssh_session == mp::nullopt)
+        return;
+
     if (time_left > std::chrono::milliseconds::zero())
     {
-        ssh_session.exec(fmt::format("wall \"The system is going down for poweroff in {} minute{}, use 'multipass stop "
-                                     "--cancel {}' to cancel the shutdown.\"",
-                                     time_left.count(), time_left > std::chrono::minutes(1) ? "s" : "", name));
+        ssh_session.value().exec(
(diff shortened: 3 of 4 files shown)


src/daemon/daemon.cpp
```diff
@@ -1968,7 +1968,17 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
...
-    if (std::none_of(cbegin(skip_states), cend(skip_states), [&state](const auto& st) { return state == st; }))
+    if (state == St::unknown)
+    {
+        if (delay.count() > 0)
+        {
+            return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
+                                "Cannot delay shutdown when instance is in the unknown state");
+        }
+
+        vm.shutdown();
```
ricab: I must be missing something but, when `delay` is 0, how is this different from what the `shutdown_timer` did?

src/daemon/daemon.cpp
```diff
@@ -1968,7 +1968,17 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
     using St = VirtualMachine::State;
     const auto skip_states = {St::off, St::stopped, St::suspended};
 
-    if (std::none_of(cbegin(skip_states), cend(skip_states), [&state](const auto& st) { return state == st; }))
+    if (state == St::unknown)
+    {
+        if (delay.count() > 0)
+        {
+            return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
+                                "Cannot delay shutdown when instance is in the unknown state");
```
ricab: why?

src/daemon/daemon.cpp
townsend2010: `shutdown_timer` requires a valid ssh_session, which we are not going to get if we can't determine the IP address of the instance.  The `UNKNOWN` status indicates we were unable to get an ssh session into to the instance to validate it was up and running.  See `wait_until_ssh_up`.

So, to be able to still stop the instance, we just force it here.

src/daemon/daemon.cpp
townsend2010: A delay of > 0 is a delayed shutdown, which we cannot do since we cannot get a valid ssh session into the instance.  My thinking is if someone is trying to do a delayed shutdown on an instance that is in an `UNKNOWN` state, then delayed shutdown is not valid.  I suppose the other option is to just kill the instance even when a delayed shutdown is requested, but that defeats the purpose of one wanting a delayed shutdown and would probably alarm the user.

src/daemon/daemon.cpp
ricab: I believe I got that same issue in states other than `UNKNOWN` (e.g. `STARTING`). IIUC, `VirtualMachine::shutdown` is also what shuts the machine down in the timer, the ssh session being used only to print wall messages. 

Wouldn't it then be better then to keep the `VM::shutdown` call in a single place but make the ssh session optional, rather than making the shutdown procedure depend on the state?

src/daemon/daemon.cpp
ricab: But if we made the ssh session optional in the delayed shutdown timer, it could work the same way, right? I mean, we would not need special behaviour for different states, other than printing or not printing wall messages depending on whether we can ssh into the machine (we keep the qemu interface in all cases)

src/daemon/daemon.cpp
townsend2010: Ok, changed:)

src/daemon/daemon.cpp
townsend2010: Changed

src/daemon/daemon.cpp
ricab: Yeah, something along those lines is what I had in mind. It may be a bit more work but I thought it could save us hunting down state-dependent code later on. 

Of course there may be disadvantages I am not thinking of, so please do counter if you think that's the best! :)

ricab: LOL ok thanks, crossed messages here, I hadn't seen your last when I replied

include/multipass/delayed_shutdown_timer.h
```diff
@@ -47,7 +47,7 @@ class DelayedShutdownTimer : public QObject
 
     QTimer shutdown_timer;
     VirtualMachine* virtual_machine;
-    SSHSession ssh_session;
+    optional<SSHSession> ssh_session;
```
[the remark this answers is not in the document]
townsend2010: hahaha, yeah:)  But really, it was the only way to do it without jumping through hoops of making an SSHSession default constructor with no arguments since we had to have the variable outside the try block.

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -25,22 +25,27 @@ namespace mpl = multipass::logging;
 
 namespace
 {
-void write_shutdown_message(mp::SSHSession& ssh_session, const std::chrono::minutes& time_left, const std::string& name)
+void write_shutdown_message(mp::optional<mp::SSHSession>& ssh_session, const std::chrono::minutes& time_left,
+                            const std::string& name)
 {
+    if (ssh_session == mp::nullopt)
```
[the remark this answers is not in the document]
townsend2010: Right, I'll probably change this since it will look cleaner.

src/daemon/delayed_shutdown_timer.cpp
```diff
@@ -50,7 +55,8 @@ mp::DelayedShutdownTimer::~DelayedShutdownTimer()
     if (shutdown_timer.isActive())
     {
         // exit_code() is here to make sure the command finishes before continuing in the dtor
-        ssh_session.exec("wall The system shutdown has been cancelled").exit_code();
+        if (ssh_session != mp::nullopt)
+            ssh_session.value().exec("wall The system shutdown has been cancelled").exit_code();
```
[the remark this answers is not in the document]
townsend2010: Sure, I see what you mean.  I'll add it in a try/catch to kill the redundancy.

src/daemon/daemon.cpp
```diff
@@ -1972,7 +1972,15 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
     {
         delayed_shutdown_instances.erase(name);
 
-        mp::SSHSession session{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
+        mp::optional<mp::SSHSession> session{mp::nullopt};
+        try
+        {
+            session = mp::SSHSession{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
+        }
+        catch (...)
```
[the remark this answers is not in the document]
townsend2010: Sure, I know what you mean, but we are in total control here of what is being thrown.  libssh does not throw at all (it's C based) and so any throws come from the `SSHSession` code.  I could be a bit more specific and catch `std::runtime_error` here.  But I don't think it's worth the effort for this particular issue to create a new exception class for this.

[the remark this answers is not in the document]
ricab: I am not convinced: even before entering the `SSHSession` code we could trip in [totally different I/O code](https://github.com/CanonicalLtd/multipass/blob/78e93672c765d4258b8533775482df7bf653a4bf/src/platform/backends/qemu/dnsmasq_server.cpp#L72)... but OK, if you think that is the best.... I don't want to insist more.

townsend2010:
@ricab,

I've pushed the latest changes, so please re-review at your convenience:)

src/daemon/daemon.cpp
```diff
@@ -1972,13 +1972,15 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
...
         delayed_shutdown_instances.erase(name);
 
-        mp::optional<mp::SSHSession> session{mp::nullopt};
+        mp::optional<mp::SSHSession> session;
         try
         {
             session = mp::SSHSession{vm.ssh_hostname(), vm.ssh_port(), vm.ssh_username(), *config->ssh_key_provider};
         }
-        catch (...)
+        catch (const std::exception& e)
```
ricab: For the record, as [previously discussed](https://github.com/CanonicalLtd/multipass/pull/589#discussion_r253998933), I am not keen on this, but don't want to insist.

townsend2010:
bors r=ricab

PR #593, opened 2019-02-06
[libvirt] use `current` in `<emulator>` path (Fixes #592)

### LINKED ISSUE #592: Upgrading while `driver=LIBVIRT` may result in broken instances
```
$ snap set multipass driver=LIBVIRT
$ multipass launch --name libvirt-test
$ grep emulator /var/snap/multipass/common/libvirt/qemu/libvirt-test.xml
  <emulator>/snap/multipass/593/usr/bin/qemu-system-x86_64</emulator>
# where 593 is a revision of the snap that may be removed
```

It's not clear when is the domain rewritten (but it is, as I couldn't find a reliable way to reproduce with the published revisions), but I stumbled upon this while working on #591.

Saviq:


src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Canonical, Ltd.
+ * Copyright (C) 2019 Canonical, Ltd.
```
townsend2010: Just curious why you didn't use `2018-2019` here since this file was created in 2018?

ricab: Oops, I have also been doing this. I started after seeing it done in some change, I think...

PR #591, opened 2019-02-06
[snap] switch to core18 base

Saviq:

PR #596, opened 2019-02-06
 [snap] switch to core18 base (Fixes #598)

### LINKED ISSUE #598: switch to core18 base
Most of this already happens in #596, but there's a twist: any instances suspended or running on the core16 version of the snap won't be resume-able on the core18 version.

`snapd` does not give us much notice about the refresh, the only thing we could do is to fail the `pre-refresh` hook, but without [epochs' stepped upgrades](https://forum.snapcraft.io/t/epochs-stepped-upgrades/1757) that snap revision may be skipped, and in fact [would be for sure](https://forum.snapcraft.io/t/how-to-cause-a-snap-rollback/9848/4), if we failed `pre-refresh`.

The only option for safe, no-user-intervention transition would be to convert/upgrade the snapshot, but it's unclear whether that's possible.

Plan of record:
- create a `core16` snap branch and keep the latest `core16`-based release there
- add a `start` option allowing for the user to discard the snapshot
- detect the incompatible snapshot either by its metadata or the `qmp` error and mark the instance appropriately
- inform the user of the problem and how they can solve it (refresh to `core16`, shut all instances down, refresh back)

Saviq:


Saviq:
This is a resubmit of #591.

I left
```
The 'qemu' part needs the following libraries that are not included in the snap or base: 
usr/lib/x86_64-linux-gnu/libGL.so.1
usr/lib/x86_64-linux-gnu/libGLU.so.1
usr/lib/x86_64-linux-gnu/libGLX.so.0
usr/lib/x86_64-linux-gnu/libGLdispatch.so.0
These dependencies can be satisfied via more stage-packages, more parts, or content sharing.
```
On purpose, since we don't use GL in our VMs.

Saviq:
bors try

Saviq:
Reopened targetting into `master`.

Saviq:
There's a new development… `snapd` [has a bug](https://bugs.launchpad.net/snappy/+bug/1819875) when transitioning services between different bases. It's recommended that we wait for this bug to be fixed before we do the switch.

Saviq:
@townsend2010 this is ready for another look.

Saviq:
Snap available in the `edge/core18` channel.

townsend2010:
Actually, since it's Friday, let's wait until Monday before landing :grin:

Saviq:
It's only going into `edge`, it's _fine_ ;)

bors r=townsend2010

townsend2010:
lol, ok:)

PR #607, opened 2019-02-12
qemu: Save qemu "machine type" as metadata for the instance

### LINKED ISSUE #598: switch to core18 base

townsend2010:
This is to allow resuming from suspend when changing QEMU versions in the snap.

Fixes #598

### DIFF
--- a/include/multipass/vm_status_monitor.h
+++ b/include/multipass/vm_status_monitor.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -13,14 +13,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
- *
  */
 
 #ifndef MULTIPASS_VM_STATUS_MONITOR_H
 #define MULTIPASS_VM_STATUS_MONITOR_H
 
 #include <string>
+
+#include <QJsonObject>
 
 namespace multipass
 {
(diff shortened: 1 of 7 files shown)


townsend2010:
@Saviq,

The `metadata` is not qemu specific per se.  It is *any* data that a specific backend wants to put in there, as shown in that `QemuVirtualMachine` is the one setting it and telling the daemon to save it.  

For example, the HyperKit backend (or any other backend) could potentially set any metadata it wants in there, just as long as it's sent as a JSON object.

That said, there needs to be some way to persist the data, so when `multipassd` comes up after a restart, the backend can get that saved data.  And unless we totally want each backend to have direct access to the json file, then I think the way I've implemented it is the most straight forward and safest way to go.

townsend2010:
@Saviq,

Just read your edit.  I'm not introducing anything qemu specific to the generic classes.  It's a QJsonObject that can contain *anything* the backend wants it to contain.

Saviq:
> Still, think it prudent to introduce all the qemu-specific bits after all to the generic classes?

I had (some food and) another look and I agree with your replies :)

I suppose it's not weird to require that the metadata is JSON, if the on-disk data format is that anyway. Only other option would be to employ some kind of marshalling, which doesn't seem to make much sense when it's not public API anyway.

townsend2010:
Right, I chose JSON since it's already JSON:)   Glad you had food and agree with me now;)

townsend2010:
Another issue is that the `QTemporaryFile` I use to output `-dump-vmstate` does not work inside a snap.  Need to fix that as well.

townsend2010:
Ok, it wasn't the QTemporaryFile after all, but I found the issue regardless and fixed it.  This should be ready again for review.

Saviq:
bors r+

Thanks!

PR #619, opened 2019-02-20
Remove sshfs install from cloud-init

### LINKED ISSUE #615: Do not install `sshfs` in the instance during `cloud-init`
We should not update `apt` caches and install `sshfs` during the `cloud-init` stage in order to decrease the time it takes to launch.  Instead, we'll use the already implemented "just in time" `sshfs` install during `mount`.

townsend2010:
Also move "just in time" sshfs install logic into the daemon.

Fixes #615

### DIFF
--- a/src/client/cmd/mount.cpp
+++ b/src/client/cmd/mount.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -17,8 +17,8 @@
 
 #include "mount.h"
 #include "common_cli.h"
-#include "exec.h"
 
+#include "animated_spinner.h"
 #include <multipass/cli/argparser.h>
 #include <multipass/cli/client_platform.h>
 #include <multipass/logging/log.h>
@@ -59,29 +59,24 @@ mp::ReturnCode cmd::Mount::run(mp::ArgParser* parser)
         return parser->returnCodeFrom(ret);
     }
 
-    auto on_success = [](mp::MountReply& reply) {
+    mp::AnimatedSpinner spinner{cout};
+
+    auto on_success = [&spinner](mp::MountReply& reply) {
+        spinner.stop();
         return ReturnCode::Ok;
     };
 
-    auto on_failure = [this, &parser](grpc::Status& status) {
-        auto ret = standard_failure_handler_for(name(), cerr, status);
-        if (!status.error_details().empty())
-        {
-            mp::MountError mount_error;
-            mount_error.ParseFromString(status.error_details());
+    auto on_failure = [this, &spinner](grpc::Status& status) {
+        spinner.stop();
 
-            if (mount_error.error_code() == mp::MountError::SSHFS_MISSING)
-            {
-                cmd::install_sshfs_for(mount_error.instance_name(), parser->verbosityLevel(), rpc_channel, stub, cout,
-                                       cerr);
-            }
-        }
-
-        return ret;
+        return standard_failure_handler_for(name(), cerr, status);
     };
 
+    auto streaming_callback = [&spinner](mp::MountReply& reply) { spinner.start(reply.mount_message()); };
+
     request.set_verbosity_level(parser->verbosityLevel());
-    return dispatch(&RpcMethod::mount, request, on_success, on_failure);
+
+    return dispatch(&RpcMethod::mount, request, on_success, on_failure, streaming_callback);
 }
 
 std::string cmd::Mount::name() const { return "mount"; }
--- a/src/client/cmd/mount.cpp
+++ b/src/client/cmd/mount.cpp
@@ -72,7 +72,10 @@ mp::ReturnCode cmd::Mount::run(mp::ArgParser* parser)
         return standard_failure_handler_for(name(), cerr, status);
     };
 
-    auto streaming_callback = [&spinner](mp::MountReply& reply) { spinner.start(reply.mount_message()); };
+    auto streaming_callback = [&spinner](mp::MountReply& reply) {
+        spinner.stop();
+        spinner.start(reply.mount_message());
+    };
 
     request.set_verbosity_level(parser->verbosityLevel());
 
--- a/src/client/cmd/start.cpp
+++ b/src/client/cmd/start.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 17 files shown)


Saviq:
@townsend2010 this needs a conflict resolve

townsend2010:
Ok, merge conflict resolved and is ready for review:)

ricab:
bors try

src/client/cmd/mount.cpp
```diff
@@ -59,29 +59,24 @@ mp::ReturnCode cmd::Mount::run(mp::ArgParser* parser)
...
-            }
-        }
+    auto on_failure = [this, &spinner](grpc::Status& status) {
+        spinner.stop();
 
-        return ret;
+        return standard_failure_handler_for(name(), cerr, status);
     };
 
+    auto streaming_callback = [&spinner](mp::MountReply& reply) { spinner.start(reply.mount_message()); };
```
ricab: Shouldn't start be preceded by a stop for intermediate messages? I believe the spinner currently checks whether it is already running.

src/client/cmd/start.cpp
```diff
@@ -59,25 +61,19 @@ mp::ReturnCode cmd::Start::run(mp::ArgParser* parser)
                                "instance.\n");
                 }
             }
-            else
```
ricab: Since this else is deleted, I would prefer to see the corresponding `if` incorporated into the parent `if`, to reduce one nesting level

src/client/cmd/start.cpp
```diff
@@ -59,25 +61,19 @@ mp::ReturnCode cmd::Start::run(mp::ArgParser* parser)
...
-                                           cout, cerr);
-                }
-            }
         }
 
         return ret;
     };
 
+    auto streaming_callback = [&spinner](mp::StartReply& reply) {
+        spinner.stop();
```
ricab: I wonder if `start` should just restart in the animated spinner, rather than requiring stop/start calls.

src/daemon/daemon.cpp
```diff
@@ -1261,7 +1258,18 @@ try // clang-format on
             }
             catch (const mp::SSHFSMissingError&)
             {
-                return grpc_status_for_mount_error(name);
+                try
+                {
+                    MountReply mount_reply;
+                    mount_reply.set_mount_message("Enabling support for mounting");
```
ricab: This seems like a good reason for the spinner to be restarted in the client.

src/daemon/daemon.cpp
```diff
@@ -2024,3 +2043,38 @@ grpc::Status mp::Daemon::cmd_vms(const std::vector<std::string>& tgts, std::func
...
+void mp::Daemon::install_sshfs(const VirtualMachine::UPtr& vm, const std::string& name)
+{
+    auto& key_provider = *config->ssh_key_provider;
+
+    SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
+
+    mpl::log(mpl::Level::info, category, fmt::format("Installing sshfs in \'{}\'", name));
+
+    int retries{0};
+    while (++retries <= 3)
```
ricab: 3 should be a constant somewhere IMO. Either at the top of this file, if only used here, or globally available if we do this sort of retrying elsewhere. There is something [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es45-avoid-magic-constants-use-symbolic-constants), but I think the main argument is encapsulation (e.g. to avoid changing this 3 in the future but missing the one below)

src/daemon/daemon.cpp
```diff
@@ -2024,3 +2043,38 @@ grpc::Status mp::Daemon::cmd_vms(const std::vector<std::string>& tgts, std::func
...
+                break;
+            }
+        }
+        catch (const mp::ExitlessSSHProcessException&)
+        {
+            mpl::log(mpl::Level::info, category, fmt::format("Timeout while installing 'sshfs' in '{}'", name));
+        }
+    }
+
+    if (retries > 3)
```
ricab: idem

src/rpc/multipass.proto
```diff
@@ -217,16 +217,6 @@ message TargetPathInfo {
     string target_path = 2;
 }
 
-message MountError {
```
ricab: are we OK with backward incompatible changes of the protocol? Do we need to consider client-server version mismatches? Perhaps we should have a deprecation period?

src/client/cmd/start.cpp
```diff
@@ -37,11 +37,13 @@ mp::ReturnCode cmd::Start::run(mp::ArgParser* parser)
...
-    auto on_success = [](mp::StartReply& reply) {
+    AnimatedSpinner spinner{cout};
+
+    auto on_success = [&spinner](mp::StartReply& reply) {
+        spinner.stop();
         return ReturnCode::Ok;
     };
 
-    AnimatedSpinner spinner{cout};
     auto on_failure = [this, &spinner, &parser](grpc::Status& status) {
```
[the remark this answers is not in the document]
townsend2010: Ugh, these darn things get me every time:)

src/client/cmd/mount.cpp
townsend2010: Yeah, it won't hurt if it's not running and will address any intermediate messages as you say.

src/client/cmd/start.cpp
townsend2010: Yeah, I should be able to tighten this up some.

src/client/cmd/start.cpp
townsend2010: Yeah, probably some sort of check that is the message string is different, then it resets it or something...but for a different PR:)

src/daemon/daemon.cpp
townsend2010: Unless I misunderstand what you are saying here, adding the `spinner.stop()` in mount's `streaming_callback` should do this??

src/daemon/daemon.cpp
townsend2010: Agreed...I must have had a moment of insanity when I did this.  I know better:)

src/daemon/daemon.cpp
townsend2010: Yep

src/rpc/multipass.proto
townsend2010: I'm of the opinion that until we draw the 1.0 line in the sand, the rpc protocol stability is off the table.  Anyone who has created a client using the rpc itself will just have to fix their client.

@Saviq, @gerboland, any opinions?

src/client/cmd/start.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/client/cmd/mount.cpp
townsend2010: Done

src/client/cmd/start.cpp
townsend2010: Tightened this up some.  Let me know what you think.

src/daemon/daemon.cpp

src/daemon/daemon.cpp

src/rpc/multipass.proto
Saviq: Yeah, we don't consider the protocol stable until we make it official. Also client/server split is not really possible yet, so +1 here.

src/daemon/daemon.cpp
ricab: yes, this was just me referring back to the same thing

ricab:
It seems the method removal needs to be propagated to hyperkit: https://travis-ci.org/CanonicalLtd/multipass/jobs/500462368#L3975 

I suppose the windows one too?

townsend2010:
@ricab,

Yes, that occurs when we merge `master` into the `full` branch.

ricab:
@townsend2010 delegating as I am not sure what's best in face of bors failing.

bors delegate=townsend2010

PR #623, opened 2019-02-21
[utils] add throw/log process wrapers

Saviq:


Saviq:
There's a more complete approach for this [in here](https://github.com/CanonicalLtd/multipass/compare/process-refactoring) so will drop this for now.

PR #628, opened 2019-02-21
Enable passing cloud-init config via stdin

gerboland:


src/client/cmd/launch.cpp
```diff
@@ -160,7 +161,27 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     {
         try
         {
-            auto node = YAML::LoadFile(parser->value(cloudInitOption).toStdString());
+            YAML::Node node;
+            const QString& cloudInitFile = parser->value(cloudInitOption);
+            if (cloudInitFile == "-")
+            {
+#ifdef MULTIPASS_PLATFORM_WINDOWS
```
townsend2010: I prefer not to have any Windows specific code in `master`, so instead, we could add a new utility function definition in  `include/multipass/cli/client_platform.h` and then put the implementation in the appropriate file under `src/platform/client`.  For non-Windows platforms, it would just be a no-op.

src/client/cmd/launch.cpp
```diff
@@ -160,7 +161,27 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     {
         try
         {
-            auto node = YAML::LoadFile(parser->value(cloudInitOption).toStdString());
+            YAML::Node node;
+            const QString& cloudInitFile = parser->value(cloudInitOption);
+            if (cloudInitFile == "-")
```
townsend2010: We should check if we are running in a TTY before going any further here.  There is already a convenience function defined in `include/multipass/cli/client_platform.h` and implemented already for each platform.

gerboland: Good point, done

src/client/cmd/launch.cpp
gerboland: Done

PR #631, opened 2019-02-22
client/shell: Start instance if it's not running

### LINKED ISSUE #621: `multipass shell <instance_name>` should start the instance if it's not running
`shell` should start a `STOPPED` or `SUSPENDED` instance instead of reporting an error.

townsend2010:
Fixes #621

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1354,8 +1354,7 @@ try // clang-format on
         auto& vm = it->second;
         if (!mp::utils::is_running(vm->current_state()))
         {
-            return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
-                                fmt::format("instance \"{}\" is not running", name), "");
+            return grpc::Status(grpc::StatusCode::ABORTED, fmt::format("'{}' needs starting", name));
         }
 
         if (vm->state == VirtualMachine::State::delayed_shutdown)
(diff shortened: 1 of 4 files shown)


src/daemon/daemon.cpp
```diff
@@ -1353,8 +1353,7 @@ try // clang-format on
         auto& vm = it->second;
         if (!mp::utils::is_running(vm->current_state()))
         {
-            return grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
-                                fmt::format("instance \"{}\" is not running", name), "");
+            return grpc::Status(grpc::StatusCode::ABORTED, fmt::format("'{}' needs starting", name));
```
gerboland: I'd prefer message said what is current state, rather than what might need to be done. I.e. "{} is not running"

gerboland: Hmm, this is printed to the console, as I see "shell failed: 'b' needs starting" before the spinner.
That's not great UI, as nothing has failed really. Can that message be suppressed?

townsend2010: @gerboland,

Sure, we can suppress the message.  I think the "Starting foo" message on the console would be enough to clue the user that something is happening and tells why the shell hasn't started.

townsend2010:
@gerboland,

Regarding tests, yeah, we need to do a bunch of work in these types of places to enable testing, ie, mocking many things, etc.  It's something we really, really need to do, but this is not the place to do it.

gerboland:
Re: tests, ok

src/daemon/daemon.cpp
townsend2010: Ok, switched back to the original message **and** the client doesn't print this out when `grpc::StatusCode::ABORTED` is returned.

townsend2010:
@gerboland, this is ready for another review pass.  Thanks!

PR #633, opened 2019-02-22
Mem value sanitizing

ricab:
Improves the handling of memory and disk size inputs. Means to address three issues, one of which is still pending.


src/daemon/daemon.cpp
```diff
@@ -666,7 +671,19 @@ try // clang-format on
     if (metrics_opt_in.opt_in_status == OptInStatus::ACCEPTED)
         metrics_provider.send_metrics();
 
-    auto name = name_from(request, *config->name_generator, vm_instances);
+    auto mem_disk_name_and_errors = validate_create_arguments(request);
+    auto mem_size = std::get<0>(mem_disk_name_and_errors);      // use structured bindings instead in C++17
+    auto disk_space = std::get<1>(mem_disk_name_and_errors);    // use structured bindings instead in C++17
+    auto instance_name = std::get<2>(mem_disk_name_and_errors); // use structured bindings instead in C++17
+    auto option_errors = std::get<3>(mem_disk_name_and_errors); // use structured bindings instead in C++17
```
gerboland: I dislike this because there's no clear connection between 0 and mem_size, 1 and disk_space, etc. Sure it works, but when we add an entry in the tuple, or swap entries with same type, it's easy to refer to the wrong index and create a bug, yet the code will still compile.

A quick struct will give names to these entries, establishing a stronger key/value connection. e.g. VirtualMachineDescription

Note, you could correctly point out that we do something very similar in the return statements of validate_create_arguments, to_machine_desc, etc - where we use brace initializers to populate a struct - as again there's no clear key/value connection. I'm not a giant fan of that style either, but at least it is localised to the creation of the struct, and not the consumer, so the person who changes it will (I think) be more aware of the consequences.

src/utils/utils.cpp
```diff
@@ -54,18 +55,51 @@ QDir mp::utils::base_dir(const QString& path)
...
-    QRegExp matcher("\\d+((K|M|G)(B){0,1}){0,1}$");
+    static constexpr auto kilo = 1024LL;
 
-    return matcher.exactMatch(mem_string);
+    QRegExp matcher("^(\\d+)([KMG])?B?$", Qt::CaseInsensitive);
+
+    if (matcher.exactMatch(QString::fromStdString(mem_value)))
+    {
+        auto val =
+            std::stoll(matcher.cap(1).toStdString()); // value is in the second capture (1st one is the whole match)
```
gerboland: Qt has a QString::toLongLong

include/multipass/utils.h
```diff
@@ -51,8 +52,9 @@ QString make_uuid();
 std::string contents_of(const multipass::Path& file_path);
 bool has_only_digits(const std::string& value);
 void validate_server_address(const std::string& value);
-bool valid_memory_value(const QString& mem_string);
-bool valid_hostname(const QString& name_string);
+optional<long long> in_bytes(const std::string& mem_value);
```
gerboland: We don't need sign, so unsigned long long, or uint64_t ?

src/daemon/daemon.cpp
```diff
@@ -154,24 +158,22 @@ void prepare_user_data(YAML::Node& user_data_config, YAML::Node& vendor_config)
...
                                               const mp::SSHKeyProvider& key_provider)
 {
     const auto num_cores = request->num_cores() < 1 ? 1 : request->num_cores();
-    const auto mem_size = request->mem_size().empty() ? "1G" : request->mem_size();
-    const auto disk_size = request->disk_space().empty() ? "5G" : request->disk_space();
     const auto instance_dir = mp::utils::base_dir(image.image_path);
     const auto cloud_init_iso =
         make_cloud_init_image(name, instance_dir, meta_data_config, user_data_config, vendor_data_config);
-    return {num_cores, mem_size, disk_size, name, mac_addr, ssh_username, image, cloud_init_iso, key_provider};
+    return {num_cores, mem_size, disk_space, name, mac_addr, ssh_username, image, cloud_init_iso, key_provider};
```
gerboland: Why the rename disk_size->disk_space?

src/utils/utils.cpp
ricab: Thank you, I did not know that. Changing accordingly

src/daemon/daemon.cpp
gerboland: Taking your points
1. as I said, swapping entries of same type will still compile, but create bug. Public/private is irrelevant to that argument, and the greatest strength structs have.
2. I'm only taking this case, not every return. Named returns would be ideal, closest is a struct IMO. 
3. same problem, just replace indexes with key names.
4. I'm not saying it's ideal, but least worst option IMO
5. not just public IMO
6. readability isn't my priority here, avoiding bugs is - especially those possibly introduced later by a different author.
7. it's a struct just to pass a list of values with labels. My objection is purely down to the lack of labels with your approach here.
8. I dunno, ValidatedVMOptions?

Sorry, my opinion hasn't changed. Note I'm not making a big deal here, it's just code after all :) 

I'll let @townsend2010 have his say.

include/multipass/utils.h
gerboland: Well I didn't think we were doing arithmetic on it, just passing it around. But I won't object to keeping it as-is.

src/daemon/daemon.cpp
townsend2010: I do feel like the proposed approach is more prone to bugs accidentally introduced if some other maintainer or contributor down the road added another check here, like for example, validating the number of cores/CPU's as well.  What I'm not a fan of, is that the consumer **has** to know the exact order of the elements that are returned from `validate_create_arguments`.  With a struct, it only needs to know the struct's member name and the rest still are valid, even if a contributor puts a new member variable anywhere in the struct.

But as @gerboland said, it's just code and it works now, so I wouldn't block on this either:)

ricab: Thanks for the comments. No big deal here either! Just debating an interesting issue with implications in future coding practices :) I actually agree with you to a large extent, but I don't quite follow what you say in 1 and 6.

Suppose I used a `struct Foo`, with ctor `Foo(mem, disk, name, errors)`. The function would say `return Foo{a,b,c,d};` right? Aren't argument swapping bugs as likely? The way I see it, a change in the producer requires a change in the consumer in both cases. The compiler will do the same. What changes is where those responsibilities lie, which is why I mentioned public/private.

A struct-based approach encapsulates the consumer, preventing impact on the function's client(s). But if the function has a single client and is only there to offload code into the anonymous namespace, the benefit is small IMO. When considering also drawbacks (struct proliferation, *a la carte* types representing no clear concept), I place the ideal compromise a bit differently (barring named returns).

Anyway... I just saw @townsend2010's comment and will modify the code to use the struct.

ricab:
This currently uses C++17 features (fallthrough attribute and structured bindings), so rebased on that branched.

src/daemon/daemon.cpp
ricab: I just committed changes with the ordering now encapsulated in a struct. Does that address the concerns?

src/utils/utils.cpp
ricab: I did this, so resolving

src/daemon/daemon.cpp
ricab: @gerboland did my answer make sense?

gerboland: I also wish we were consistent. Oh well.. no objection then

src/daemon/daemon.cpp
```diff
@@ -281,25 +283,35 @@ auto fetch_image_for(const std::string& name, const mp::FetchType& fetch_type, m
...
+        option_errors.add_error_codes(mp::LaunchError::INVALID_HOSTNAME);
 
-    return launch_error;
+    struct CheckedArguments
+    {
+        std::string mem_size;
+        std::string disk_space;
+        std::string instance_name;
+        mp::LaunchError option_errors;
+    } ret{mem_size, disk_space, instance_name, option_errors};
```
gerboland: This doesn't address my concern. There is still no connection between this ordering of params and the order in the consumer. E.g. this will compile, but is wrong:
```
auto [disk_space, mem_size, instance_name, option_errors] = validate_create_arguments(request);
```
This is the root of my dislike of this syntactic sugar.

If the params were all different types, the type system would catch errors. So I'm ok with this syntax being used with care. But here, I think using a struct like this would give more safety. Here's my thinking:
https://pastebin.ubuntu.com/p/RsQ3sf6dZy/

My motivation is that this moves the responsibility for parameter order checking to solely the location the param list is created (i.e. in validate_create_arguments). Not where it is consumed. 

With this struct approach, if someone swaps mem_size and disk_space, they only need to care about reordering those params in the creation of the struct. No other code needs to change, and no bugs created.

Sure, we could go all the way and insist on this at creation:
```
CheckedArguments args;
args.mem_size = mem_size;
args.disk_space = ....
return args;
```
but I think that's unnecessary. I trust that if an author wants to change the CheckedArguments list order, they'll be sure to do the right thing. But I don't trust they'll update all the consumers correctly too.

Is it pretty? No. But IMO it's worth it for safety.

townsend2010: Just my $0.02, but that pastebin is what I envisioned too when I made my comments.

ricab: OK, I will do it that way then, but...

> My motivation is that this moves the responsibility for parameter order checking to solely the location the param list is created (i.e. in validate_create_arguments). Not where it is consumed.

Well, the ordering of public attributes is part of a struct's interface, so it is not being contained. IOW, anyone changing the order would still have to check all clients (think also construction, assignment...). But I understand your concern is that they may forget so it is perhaps safer not to rely on that here.

ricab:
OK take 3 :)

@gerboland did as you requested, keeping the struct inside the function as discussed.

Saviq:
Moved back onto `move-to-cpp17` as we had to revert it temporarily.

ricab:
@gerboland thanks for noticing that!

I will have to move this because changing the base of this PR closed it and github does not allow reopening. New PR comming...

Issue #637, opened 2019-02-26
Build failure (unknown type name QEventLoop)


Saviq:
Due to our CI troubles last week some things got merged manually and we seem to have missed this [build fail](https://travis-ci.org/CanonicalLtd/multipass/jobs/497430678):

```
../src/daemon/daemon.cpp:1567:13: error: unknown type name 'QEventLoop'
            QEventLoop event_loop;
            ^
../src/daemon/daemon.cpp:1569:77: error: use of undeclared identifier 'QEventLoop'
            QObject::connect(this, &Daemon::suspend_finished, &event_loop, &QEventLoop::quit, Qt::QueuedConnection);
                                                                            ^
../src/daemon/daemon.cpp:1570:70: error: use of undeclared identifier 'QEventLoop'
            QObject::connect(&timer, &QTimer::timeout, &event_loop, &QEventLoop::quit);
                                                                     ^
3 errors generated.
```

PR #632, opened 2019-02-22
qemu: Remove QCoreApplication::processEvents() to avoid a crash

townsend2010:
Due to how we are interacting with the Qt event loop, it's possible to crash
when deleting an instance while it's still launching. The after effect of this
is that concurrent client connections will block until launch (or other long
running commands) finishes.

A plan has been fleshed out to alleviate the concurrent blocking and will be
implemented later.

Fixes #585

PR #638, opened 2019-02-26
[daemon] add missing QEventLoop include

Saviq:
Removing QCoreApplication in #632 resulted in this symbol being missing.


ricab:
Fixes #637

PR #640, opened 2019-02-27
[travis] use candidate snapcraft temporarily

Saviq:
To work around https://forum.snapcraft.io/t/psa-classic-confinement-and-core-snapcraft-issue/10149


Saviq:
This won't work, snapcraft stable needs to go inside the container...

PR #641, opened 2019-02-27
Move to cpp17

ricab:
Opening this to discussion: an attempt to move to C++17. I tested locally with gcc 8 and clang 7. Want to see what Travis has to say.


townsend2010:
This broke our current way of building snaps: https://pastebin.ubuntu.com/p/ysjjqWqYmZ/

PR #642, opened 2019-02-27
backends/libvirt: Match invalid ROM size error on instance restore

townsend2010:
When starting an instance with a managed save after snap base core18 upgrade,
match the error condition of mismatched ROM size and pass recovery steps to
the client.

### DIFF
--- a/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
+++ b/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
@@ -278,7 +278,26 @@ void mp::LibVirtVirtualMachine::start()
     update_state();
 
     if (virDomainCreate(domain.get()) == -1)
-        throw std::runtime_error(virGetLastErrorMessage());
+    {
+        state = State::suspended;
+        update_state();
+
+        std::string error_string{virGetLastErrorMessage()};
+        if (error_string.find("virtio-net-pci.rom: 0x80000 in != 0x40000") != std::string::npos)
+        {
+            error_string = fmt::format("Unable to start suspended instance due to incompatible save image.\n"
+                                       "Please use the following steps to recover:\n"
+                                       "  1. snap revert multipass\n"
+                                       "  2. multipass start {}\n"
+                                       "  3. Save any data in the instance\n"
+                                       "  4. multipass stop {}\n"
+                                       "  5. snap refresh multipass\n"
+                                       "  6. multipass start {}\n",
+                                       vm_name, vm_name, vm_name);
+        }
+
+        throw std::runtime_error(error_string);
+    }
 
     monitor->on_resume();
 }
--- a/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
+++ b/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
@@ -287,11 +287,11 @@ void mp::LibVirtVirtualMachine::start()
         {
             error_string = fmt::format("Unable to start suspended instance due to incompatible save image.\n"
                                        "Please use the following steps to recover:\n"
-                                       "  1. snap revert multipass\n"
+                                       "  1. snap refresh multipass --channel core16/beta\n"
                                        "  2. multipass start {}\n"
                                        "  3. Save any data in the instance\n"
                                        "  4. multipass stop {}\n"
-                                       "  5. snap refresh multipass\n"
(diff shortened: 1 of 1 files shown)


src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -278,7 +278,26 @@ void mp::LibVirtVirtualMachine::start()
     update_state();
 
     if (virDomainCreate(domain.get()) == -1)
-        throw std::runtime_error(virGetLastErrorMessage());
+    {
+        state = State::suspended;
+        update_state();
+
+        std::string error_string{virGetLastErrorMessage()};
+        if (error_string.find("virtio-net-pci.rom: 0x80000 in != 0x40000") != std::string::npos)
```
ricab: Just a question: are you sure these will always be the exact numbers in the error?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -278,7 +278,26 @@ void mp::LibVirtVirtualMachine::start()
...
+        {
+            error_string = fmt::format("Unable to start suspended instance due to incompatible save image.\n"
+                                       "Please use the following steps to recover:\n"
+                                       "  1. snap revert multipass\n"
+                                       "  2. multipass start {}\n"
+                                       "  3. Save any data in the instance\n"
+                                       "  4. multipass stop {}\n"
+                                       "  5. snap refresh multipass\n"
+                                       "  6. multipass start {}\n",
```
ricab: Another question: why will this last step now succeed?

Saviq: Because the instance will be stopped, not suspended.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
Saviq: Yes, the ROM size doubled between the `core16` and `core18` libvirt versions, and we only want to catch this exact case here. If we get any other message, we want to surface it as-is.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -278,7 +278,26 @@ void mp::LibVirtVirtualMachine::start()
...
+    {
+        state = State::suspended;
+        update_state();
+
+        std::string error_string{virGetLastErrorMessage()};
+        if (error_string.find("virtio-net-pci.rom: 0x80000 in != 0x40000") != std::string::npos)
+        {
+            error_string = fmt::format("Unable to start suspended instance due to incompatible save image.\n"
+                                       "Please use the following steps to recover:\n"
+                                       "  1. snap revert multipass\n"
```
gerboland: Definitely less likely case, but possible user has >1 multipass snap updates between creation of the old save image, and getting this error. Can we specify a particular snap revision/channel?

Saviq: Yeah that was my only concern indeed. We can't specify a revision, but we can maintain a `core16` channel for a while.

townsend2010: Ok, right.  What would the `snap` command be for something like this?

Saviq: `snap refresh multipass --channel core16/beta` and then `snap refresh multipass --channel beta`.

townsend2010: @Saviq, cool, thanks.  I'll change this to reflect that.

townsend2010: @Saviq, for the `core16` fallback, do we care about users on the `edge` channel?

Saviq: I think they'll know what to do (and no, I don't think we'll have `core16/edge`).

We'll actually move this to `--stable` once we have one.

townsend2010: Ok, changed the `snap` instructions.

Saviq:
Let's!

bors r+

Saviq:
Grrh.

bors r+

PR #650, opened 2019-02-28
Introduce Process, ProcessSpec and ProcessFactory

gerboland:
This is foundation work for adding security features to Multipass. The
goal is for processes that use external data to be confined with
technologies like AppArmor, and drop privileges if possible. This will
eventually impact qemu, qemu-img and dnsmasq.

Note Multipass spawns many processes that we consider trusted, i.e
"ip" and "iptables" - these will remain unconfined.

This MP does not include any security mechanisms, just refactors
process management by Multipass in preparation for it.

The design is the following:
- Process is a light wrapper around QProcess. It is intended solely for
  managing confined processes.
- ProcessSpec specifies the process and additional data relevant to the
  process. In future MP, this will include AppArmor rules and privilege dropping
  details.
- ProcessFactory takes a ProcessSpec, and returns a Process. In future MP,
  the process will have all the specified security mechanisms applied.


src/platform/backends/CMakeLists.txt
```diff
@@ -14,6 +14,6 @@
 #
 # Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
 
-add_subdirectory(backend_utils)
+add_subdirectory(shared_linux)
```
Saviq: As discussed yesterday, I think that limits it too much. Most of this is cross-platform, some of it unix-specific, but I don't think there's anything Linux-specific in this PR yet?

src/platform/backends/qemu/dnsmasq_process_spec.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef MULTIPASS_DNSMASQ_PROCESS_SPEC_H
+#define MULTIPASS_DNSMASQ_PROCESS_SPEC_H
+
+#include <multipass/ip_address.h>
+#include <multipass/optional.h>
+#include <multipass/path.h>
+#include "process_spec.h"
+
+#include <QDir>
```
Saviq: We have `mp::Path`, think we should use that instead?

src/platform/backends/libvirt/libvirt_virtual_machine_factory.cpp
```diff
@@ -18,7 +18,7 @@
 #include "libvirt_virtual_machine_factory.h"
 #include "libvirt_virtual_machine.h"
 
-#include <multipass/backend_utils.h>
+#include "backend_utils.h"
```
Saviq: Why quoted and not bracketed?

There's more instances of this below. So far we've only been using quoted includes for headers in the same directory as the implementation?

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -28,39 +31,24 @@ namespace mpl = multipass::logging;
...
-mp::DNSMasqServer::DNSMasqServer(const Path& path, const QString& bridge_name, const IPAddress& bridge_addr,
-                                 const IPAddress& start, const IPAddress& end)
+mp::DNSMasqServer::DNSMasqServer(const ProcessFactory* process_factory, const Path& path, const QString& bridge_name,
+                                 const IPAddress& bridge_addr, const IPAddress& start, const IPAddress& end)
     : data_dir{QDir(path)},
-      dnsmasq_cmd{start_dnsmasq_process(data_dir, bridge_name, bridge_addr, start, end)},
+      dnsmasq_cmd{make_dnsmasq_process(process_factory, data_dir, bridge_name, bridge_addr, start, end)},
       bridge_name{bridge_name}
 {
+    QObject::connect(dnsmasq_cmd.get(), &Process::readyReadStandardError,
```
Saviq: Would it make sense to integrate that into Process? It feels like we often would want to log stderr of launched processes?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -176,10 +128,11 @@ auto get_qemu_machine_type()
         return QString();
     }
 
-    auto process = set_qemu_exec();
-    process->setArguments({"-nographic", "-dump-vmstate", dump_file.fileName()});
-    process->start();
-    process->waitForFinished();
+    QProcess process;
```
Saviq: Would this not benefit from the Process treatment?

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "backend_utils.h"
+#include <QHash>
+#include <QString>
+
+namespace mp = multipass;
+
+
+mp::QemuVMProcessSpec::QemuVMProcessSpec(const mp::VirtualMachineDescription& desc, const QString& tap_device_name,
+                                     const QString& mac_addr)
```
Saviq: Weird indent, did you run clang-format on this?

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return "qemu-system-" + mp::backend::cpu_arch();
+}
+
+QStringList mp::QemuVMProcessSpec::arguments() const
+{
+    auto mem_size = QString::fromStdString(desc.mem_size);
+    if (mem_size.endsWith("B"))
+    {
+        mem_size.chop(1);
+    }
```
Saviq: This will conflict with #633, won't it?

src/platform/backends/shared_linux/backend_utils.cpp
```diff
@@ -130,23 +134,48 @@ void mp::backend::check_hypervisor_support()
...
+    if (!qemuimg_process->run_and_return_status({"resize", image_path, disk_size}))
         throw std::runtime_error("Cannot resize instance image");
 }
 
-std::string mp::backend::image_format_for(const mp::Path& image_path)
+mp::Path mp::backend::convert_to_qcow_if_necessary(const ProcessFactory* process_factory, const mp::Path& image_path)
 {
-    auto image_info = QString::fromStdString(
-        mp::utils::run_cmd_for_output("qemu-img", {QStringLiteral("info"), "--output=json", image_path}));
+    // Check if raw image file, and if so, convert to qcow2 format.
```
Saviq: Add a TODO that we could convert from other image formats as necessary?

src/platform/backends/shared_linux/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void mp::Process::start(const QStringList& extra_arguments)
+{
+    QProcess::start(program(), arguments() << extra_arguments);
+}
+
+bool mp::Process::run_and_return_status(const QStringList& extra_arguments, const int timeout)
+{
+    start(extra_arguments);
+    if (!waitForFinished(timeout))
```
Saviq: `waitForFinished` returns `false` if the process already finished, should we not check `exitStatus` here?

src/platform/backends/shared_linux/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return false;
+    }
+
+    return exitStatus() == QProcess::NormalExit && exitCode() == 0;
+}
+
+QString mp::Process::run_and_return_output(const QStringList& extra_arguments, const int timeout)
+{
+    start(extra_arguments);
+    if (!waitForFinished(timeout))
```
Saviq: Ditto.

src/platform/backends/shared_linux/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+bool mp::Process::run_and_return_status(const QStringList& extra_arguments, const int timeout)
+{
+    start(extra_arguments);
+    if (!waitForFinished(timeout))
+    {
+        mpl::log(mpl::Level::info, qPrintable(process_spec->program()), qPrintable(errorString()));
+        return false;
+    }
+
+    return exitStatus() == QProcess::NormalExit && exitCode() == 0;
```
Saviq: `exitStatus` will always be `NormalExit` here, no?

tests/stub_process_factory.h
```diff
@@ -0,0 +1,56 @@
...
+{
+    std::unique_ptr<Process> create_process(std::unique_ptr<ProcessSpec>&&) const override
+    {
+        return std::make_unique<StubProcess>();
+    }
+};
+} // namespace test
+} // namespace multipass
+
+#endif // MULTIPASS_STUB_LOGGER_H
```
Saviq: Copypasta

src/platform/backends/CMakeLists.txt
gerboland: I did look and notice that none of our non-linux code depends on the contents of the directory. Hence the name!

What we discussed did make sense, but I was wary of making too big an MP by breaking up this directory into shared, shared/linux etc. But I'll do so now if you like.

src/platform/backends/shared_linux/process.cpp
gerboland: Unless it crashes, but yeah it's not a good check. This returns true if the binary isn't even found!

I just took this line from the existing code in mp::utils::run_cmd_for_status. I didn't want to accidentally change behaviour during this refactor. But since you point it out, I'll make it more solid now.

src/platform/backends/CMakeLists.txt
Saviq: Yeah I'd say that makes sense.

src/platform/backends/shared_linux/process.cpp
gerboland: Likely yes. I didn't try to improve the logic for this iteration, just move it around. But I'll improve it now, it's not a bad opportunity to do so.

src/platform/backends/shared_linux/process.cpp
Saviq: Wouldn't `waitForFinished()` catch those above, though?

src/platform/backends/libvirt/libvirt_virtual_machine_factory.cpp
gerboland: My bad, at one stage they were in same dir. Will fix

src/platform/backends/qemu/qemu_virtual_machine.cpp
gerboland: Original goal was for Process to only wrap processes that need security confinement.

In this particular case, I don't think qemu here needs to be confined, as we ship it, and it isn't acting on any user-supplied data/file/arguments (I straced it to be sure, afaics it's just dumping internal data).

However I see demand for something more suitable to our needs than QProcess. Process is a candidate. Perhaps make it constructable, so it can be used just like QProcess too? Then rename ProcessFactory to ConfinedProcessFactory, to make clear the distinction in creation? WDYT?

Saviq: Yeah I think we might want to generalize Process a bit, but it can come later.

src/platform/backends/shared_linux/process.cpp
gerboland: This confused me a little, so I looked into it. The docs for QProcess::waitForFinished do indeed say

> Returns true if the process finished; otherwise returns false (if the operation timed out, if an error occurred, or if this QProcess is already finished).

but the last clause is subtle.

QProcess lives on the thread of its instantiator, and the code it has internally for updating its known state of the child process, also lives on this thread. QProcess doesn't create a thread to monitor the child process it spawns, instead it creates [QSocketNotifiers](https://doc.qt.io/qt-5/qsocketnotifier.html) to monitors file descriptors of the process (stdin/out/err and an [FD for the process itself](https://www.macieira.org/blog/2012/07/forkfd-part-4-proposed-solutions/)) and emit signals on changes.

QSocketNotifiers docs state that: `File descriptor action is detected in Qt's main event loop`, i.e. a signal is emitted from this thread.

Here we are running
```
start();
waitForFinished();
```
one command after another. Say if the child process did run and quit, before waitForFinished was called. In that case, a signal is emitted on the Qt main loop (_q_processDied) to the QProcess, which will be processed on a following event loop tick.

Since we start and wait in the same function, there is no opportunity in between those two function calls for the event loop to tick, so the QProcess won't change its state from starting to finished just yet.

When we call waitForFinished(), QProcess will then actively poll the FDs until the process goes away (or finds it already has died), and the FD closes, and then update its state to finished.

So I think the way we use it is correct here.
- also, if the binary is missing, waitForFinished also returns false

If start() and waitForFinished() were called in different event loop ticks, it would be possible for the internal state to be finished before waitForFinished() is called (and so would return false). We should avoid using QProcess like that, and instead rely on the finished() signal.


But you are right that we should check exitStatus() to ensure the process didn't crash, as a crashing process will still have waitForFinished return true.

src/platform/backends/qemu/dnsmasq_process_spec.h
gerboland: Done

src/platform/backends/shared_linux/backend_utils.cpp
gerboland: Done

src/platform/backends/shared_linux/process.cpp
gerboland: Addressed

tests/stub_process_factory.h
gerboland: Good spot

src/platform/backends/shared_linux/process.cpp
gerboland: waitForFinished() returns true if the process exits normally, or crashes. It doesn't care about return code.
exitStatus tells us if it crashed.

I've moved this code around a tad, see what you think.

src/platform/backends/libvirt/libvirt_virtual_machine_factory.cpp
gerboland: Done, but now suspect explicitly having <shared/linux/backend_utils.h" might be clearer...

Saviq: Yeah +1 on that.

src/platform/backends/qemu/dnsmasq_server.cpp
gerboland: It would, the only niggle is that desired the log level differs depending on the program (i.e. currently dnsmasq's stderr is logged as Error, but qemu's is logged as warning).

This also raises the question, should stdout be logged then too? Perhaps yes, with a non-default-displayed log level?

src/platform/backends/CMakeLists.txt
gerboland: I'm struggling with this change, as while now there isn't anything linux specific in the Process* bits, the next MP will introduce apparmor, which does make them linux specific.

So I'd be moving things around, just to move them back to where they are now.

(Backend_utils is more than 50% linux-only code too)

I agree with you that Process would be great as a shared thing, and likely there'll be more duplicate code we can combine. I'll definitely do that. But I don't think I've headspace to tackle that problem and add the apparmor bits simultaneously. I'd rather add the apparmor bits first, then identify the platform commonalities after.

So can I request a "let me do it later?"

src/platform/backends/libvirt/libvirt_virtual_machine_factory.cpp
gerboland: Done

src/platform/backends/qemu/dnsmasq_server.cpp
gerboland: I've added a error_log_level specifier to ProcessSpec, so each process can set how critical the child process' stderr should appear in the log

Saviq: > This also raises the question, should stdout be logged then too? Perhaps yes, with a non-default-displayed log level?

Yeah, all stdout should be logged to `debug` level I'd say and all stderr to `warning` by default.

Saviq: The `stdout` handling can come later though.

src/platform/backends/shared/linux/process.cpp
```diff
@@ -31,7 +31,9 @@ mp::Process::Process(std::unique_ptr<mp::ProcessSpec>&& spec) : process_spec{std
     if (!process_spec->working_directory().isNull())
         setWorkingDirectory(process_spec->working_directory());
 
-    setProcessChannelMode(QProcess::ForwardedErrorChannel); // default to forwarding child stderr to console
+    QObject::connect(this, &Process::readyReadStandardError, [this]() {
+        mpl::log(mpl::Level::error, qPrintable(process_spec->program()), qPrintable(readAllStandardError()));
```
Saviq: Add `TODO` please to handle newlines better here, so that it doesn't get chunked together?

src/platform/backends/shared/linux/process_spec.cpp
```diff
@@ -38,3 +39,9 @@ QString mp::ProcessSpec::working_directory() const
 {
     return QString();
 }
+
+// Set what multipass logging level the stderr of the child process should have
+mpl::Level mp::ProcessSpec::error_log_level() const
+{
+    return mpl::Level::error;
```
Saviq: I think we can default to `warning` here, wdyt?

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -0,0 +1,90 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "qemu_vm_process_spec.h"
+
+#include <QHash>
```
townsend2010: I don't believe we need this include here.

gerboland:
@townsend2010 thanks for the comments. On updating copyright, I usually only update it if a non-trivial change happens in the file. Do we have a team policy on that?

Yes I'll definitely address the process stuff later. My initial attempt was to make it cross-platform, but I found platform differences made the API challenging to implement. It may become clearer why once I introduce apparmor.

Unfortunately I'm breaking CI again, Mac has a conflict on merge.

src/platform/backends/shared/linux/process.cpp
gerboland: Done

src/platform/backends/qemu/qemu_vm_process_spec.cpp
gerboland: Quite right, removed

src/platform/backends/shared/linux/process_spec.cpp
gerboland: I really don't have a strong opinion. I thought sensible default for a process printing to stderr was to record it as an error. I'm happy to change it. Any other opinions?

townsend2010:
@gerboland, well, I don't think we have an official policy in place on when to update the copyright, but I have been under the assumption that the copyright needs to be updated when the file has been changed for any reason.  I do think we decided to remove the `Authored by` stuff when we are updating the preamble header though.

@Saviq, opinions?

src/platform/backends/shared/linux/process_spec.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <shared/linux/process_spec.h>
+
+#include <QFileInfo>
```
townsend2010: Actually, just saw that is not needed either.

src/platform/backends/shared/linux/process_spec.cpp
Saviq: The problem is `error` gets printed to the user's console, so defaulting to that may be too verbose on the user.

Saviq:
> @Saviq, opinions?

Just apply reason :) And we should probably just get rid of all the `Authored-by` headers in one go…

src/platform/backends/shared/linux/process_spec.cpp
gerboland: Done

src/platform/backends/shared/linux/process_spec.cpp
gerboland: Changed to warning. 

Note I kept existing behaviour of dnsmasq logging stderr as "error" which I presume was intentional

src/platform/backends/qemu/qemu_vm_process_spec.cpp
ricab: @gerboland is it safe to just remove 38-41 there?

ricab: @gerboland nevermind, I realize now this just changed places, but it is still required

PR #653, opened 2019-03-01
Move to cpp17 - take 2

ricab:
Retry moving to C++17. The first attempt (#641) got reverted because it breaks snaps until they use a more recent toolchain, which is why the PR is against on `snap-core18`.


Saviq:
Let's see if travis decides to act on this now :)

ricab:
not o0

Saviq:
Hey @ricab can you rebase this and re-propose? Thanks :)

PR #654, opened 2019-03-01
Mem value sanitizing - take 3

ricab:
Moved from #633 (after rebase and force-push).


ricab:
Commit to address last review issue in #633 coming shortly...

ricab:
@gerboland I just fixed the last issue you detected

ricab:
@townsend2010 only the `[[fallthrough]]` attribute remains I think.

ricab:
@townsend2010 without further changes, compilation breaks if I remove it (because a warning is issued which is treated as an error)

townsend2010:
@ricab, Ok, I was afraid of that:/  Oh well...

gerboland:
Perhaps the previous "solution": a fallthrough comment, will avoid a warning and mean we can land this?
https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them

ricab:
@gerboland is that enough for clang and msvc too?

gerboland:
I've seen people say clang supports same thing, but I'll test it on Mac & Windows to be sure.

ricab:
OK, just finished rebasing on master and resolving all conflicts. I hope everything is sane...

I had to make a few changes to integrate the new stream encapsulation. The most significant were in [this](https://github.com/CanonicalLtd/multipass/pull/654/commits/37ae02a76340e56cbacacb6d30b4d2ff2ecf2389) commit, but there were others. Dropped `[[fallthrough]]` as requested to loose the dependency on C++17.

tests/test_cli_client.cpp
```diff
@@ -66,8 +63,11 @@ struct Client : public Test
     mpt::StubCertProvider cert_provider;
     mpt::StubCertStore cert_store;
     mp::DaemonRpc stub_daemon{server_address, mp::RpcConnectionType::insecure, cert_provider, cert_store};
+    static std::stringstream trash_stream; // this may have contents (that we don't care about)
 };
 
+std::stringstream Client::trash_stream; // replace with inline in C++17
```
gerboland: Why not this: https://pastebin.ubuntu.com/p/BjsYZB6zWr/
Simpler, works now, no need for C++17 magic.

ricab: > Simpler

For the writer only, and then only marginally.

> no need for C++17 magic

No "need" in either case.

> Why not this [...]

Moving it from its natural scope and making it a global arguably reduces clarity, so I don't think it is an improvement. Don't get me wrong, I don't think it would be bad either, this being tests at all. Almost indifferent perhaps, but not better IMO.

gerboland: I'm not gonna block on it

PR #656, opened 2019-03-01
[travis] run on all pull requests

Saviq:
This way we can chain PRs behind one another.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,9 @@ language: cpp
 git:
   depth: false
 
-branches:
-  only:
-  - staging
-  - trying
-  - master
+if: |
+  branch IN (staging, trying, master)
+  OR type = pull_request
 
 addons:
   snaps:


.travis.yml
```diff
@@ -6,11 +6,9 @@ language: cpp
 git:
   depth: false
 
-branches:
-  only:
-  - staging
-  - trying
-  - master
+if: |
+  branch IN (staging, trying, master)
```
ricab: doesn't this need a slash to continue the line? https://docs.travis-ci.com/user/conditions-v1#line-continuation-multiline-conditions

ricab: according to [this](https://yaml-multiline.info/) newlines would be kept in the string with `|`

ricab: However, Travis seems happy...

Saviq: Yeah I think the important bit is "We liked the idea, so we allowed the following:"... FWIW [these](https://github.com/MirServer/mir/blob/master/.travis.yml#L15) have been working for a while.

ricab:
bors r+ 

I know this will fail for now for unrelated reasons, but it might detect other problems here.

ricab:
bors delegate=Saviq

PR #658, opened 2019-03-01
downloader: Reset timer when readReady is emitted for general downloads

### LINKED ISSUE #386: Network timeout too aggressive
If my network is not at its best state with regards to latency, it is rather easy to get:
```
$ multipass launch
failed to launch: Network timeout                                               
$ multipass launch                        
failed to launch: Network timeout
```

As a side data point, using snapcraft or having lxd auto trigger the new image download does not timeout like this. I have not timed it, but feels like it waits only 2 seconds to decide on a timeout.

townsend2010:
Fixes #386

### DIFF
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -44,6 +44,7 @@ QByteArray download(QNetworkAccessManager& manager, const Time& timeout, QUrl co
 {
     QEventLoop event_loop;
     QTimer download_timeout;
+    download_timeout.setInterval(timeout);
 
     QNetworkRequest request{url};
     request.setRawHeader("Connection", "Keep-Alive");
@@ -62,7 +63,7 @@ QByteArray download(QNetworkAccessManager& manager, const Time& timeout, QUrl co
         reply->abort();
     });
 
-    download_timeout.start(timeout);
+    download_timeout.start();
     event_loop.exec();
     if (reply->error() != QNetworkReply::NoError)
     {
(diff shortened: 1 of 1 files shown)


src/network/url_downloader.cpp
```diff
@@ -128,8 +129,11 @@ void mp::URLDownloader::download_to(const QUrl& url, const QString& file_name, i
 
 QByteArray mp::URLDownloader::download(const QUrl& url)
 {
-    return ::download(manager, timeout, url, [](QNetworkReply*, qint64, qint64) {}, [](QNetworkReply*, QTimer&) {},
-                      [] {});
+    // This will connect to the QNetworkReply::readReady signal and when emitted,
+    // reset the timer.
+    auto on_download = [](QNetworkReply* reply, QTimer& download_timeout) { download_timeout.start(); };
```
ricab: Why are lines 111-114 needed there but not here?

townsend2010: @ricab, Because in lines 111-114, we want to stop the timer to allow time reading from the `QNetworkReply` and write to the file.  Here, we are not doing that and calling `start()` just resets the timer.

ricab: @townsend2010 OK thanks, I missed that and thought could be part of the resetting

ricab:
bors delegate=townsend2010

townsend2010:
bors r=ricab

Issue #651, opened 2019-03-01
Client.launch_cmd_cloudinit_option_reads_stdin_ok test fails in CI


Saviq:
Now we got CI back, the below [happened there](https://travis-ci.org/CanonicalLtd/multipass/jobs/500005982):

```
[ RUN      ] Client.launch_cmd_cloudinit_option_reads_stdin_ok
/root/build_multipass/parts/multipass/src/tests/test_cli_client.cpp:233: Failure
Value of: send_command({"launch", "--cloud-init", "-"})
Expected: is equal to 0
  Actual: 1 (of type int)
[  FAILED  ] Client.launch_cmd_cloudinit_option_reads_stdin_ok (18 ms)
```

gerboland:
Problem is that CI runs tests on a terminal with no tty connected, unlike running test on our own machines. Multipass has internal checks to change behaviour depending on if tty around or not.

I don't see a way to trick these multipass internal checks, without a bit of refactoring. Branch attached with what I think is least worst solution

townsend2010:
Closed by #659.

PR #659, opened 2019-03-04
Pipe all stdin/out/err via dedicated Terminal interface

gerboland:
Create and use a Terminal interface which defines stdin/out/err for the client.

It more tightly couples stdin/out/err with their respective FDs, and TTY detection, which eases mocking for testing.

This also loosens the coupling of SSHClient with Console which was an impediment during this refactor.

Fixes bug #651


include/multipass/terminal.h
```diff
@@ -0,0 +1,47 @@
...
+    virtual int cin_fd() const = 0;
+    virtual bool cin_is_tty() const = 0;
+
+    virtual int cout_fd() const = 0;
+    virtual bool cout_is_tty() const = 0;
+
+    using UPtr = std::unique_ptr<Terminal>;
+    static UPtr make_terminal();
+};
+} // namespace multipass
```
ricab: Naming here confused me a bit: "how can a terminal's cout not be a tty?" I don't have a better name for the class (shell is inappropriate, console is taken, virtual terminal would only confuse things). But perhaps the `cin_is_tty` could be `is_interactive` and `cout_is_tty` could be `is_live`? I don't know, just a thought, maybe it would be even worse. WDYT?

src/platform/console/unix_console.cpp
```diff
@@ -82,15 +81,15 @@ class mp::WindowChangedSignalHandler
...
+    : term{term}, handler{std::make_unique<WindowChangedSignalHandler>(channel, term->cout_fd())}
 {
     setup_console();
 
-    if (interactive)
+    if (term->cout_is_tty())
     {
         ssh_channel_request_pty(channel);
-        change_ssh_pty_size(channel);
+        change_ssh_pty_size(channel, term->cout_fd());
```
ricab: A bit of a side question, how come `cerr` does not receive the same treatment as `cout`?

include/multipass/cli/argparser.h
```diff
@@ -75,8 +75,7 @@ class ArgParser
     bool help_requested;
     int verbosity_level{0};
 
-    std::ostream& cout;
-    std::ostream& cerr;
+    std::ostream &cout, &cerr;
```
ricab: Quite some formatting changes in code that was not otherwise modified. Was this some IDE lapse?

src/platform/terminal/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+add_library(terminal STATIC
+  terminal.cpp
+  unix_terminal.cpp)
+
+target_link_libraries(terminal
+  client_platform)
```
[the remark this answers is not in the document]
gerboland: It's not. Branch WIP so still a bit rough :)

src/platform/console/unix_console.cpp
gerboland: I didn't see it being used much. Totally could, if we want consistency

include/multipass/cli/argparser.h
gerboland: I merged the lines, clang-format did the rest. Will undo if you dislike

ricab: Nope, that's fine by me. The only thing is that clang-formatting of untouched code adds a bit of noise to review, but it wasn't much here. In general, running clang-format on only the changes of a commit(-range) rather than whole files avoids that.

src/platform/console/unix_console.cpp
ricab: Right, I wondered why it wasn't used, but that is probably a separate thing

include/multipass/terminal.h
gerboland: "how can a terminal's cout not be a tty?" - when you do this: multipass list | cat, multipass's cout is not a tty, but cin is. Similarly, echo "" | multipass, cin is not a tty, but cout is.

Naming, I'm not sure. My first thought seeing "is_interactive" is that it implies both cin and cout are backed by a tty. But I didn't want this to be too clever. 
"cin/cout_is_live" maybe?

ricab: OK, my confusion was probably caused more by intrinsic complexity than how is was presented. Yeah go with what you think is best.

ricab:
My review is in progress, but I was unable to finish it today.

src/client/cmd/launch.cpp
```diff
@@ -207,7 +194,7 @@ mp::ReturnCode cmd::Launch::request_launch()
 
         if (reply.metrics_pending())
         {
-            if (mcp::is_tty())
+            if (term->cin_is_live())
```
townsend2010: Should this check both cin and cout since there is both output and input?  Maybe another helper function in the Terminal class that checks both at the same time?

Although previously we only checked if stdin is a tty, so I'm not sure how much it matters...

src/platform/console/console.cpp
```diff
@@ -16,14 +16,16 @@
...
 #include "unix_console.h"
+#include "unix_terminal.h"
 
 namespace mp = multipass;
 
-mp::Console::UPtr mp::Console::make_console(ssh_channel channel)
+mp::Console::UPtr mp::Console::make_console(ssh_channel channel, Terminal* term)
 {
-    return std::make_unique<UnixConsole>(channel);
+    return std::make_unique<UnixConsole>(channel, static_cast<UnixTerminal*>(term));
```
ricab: For my instruction, how does this platform stuff work? Are the contents of the `/src/platform` directory somehow different in other platforms? And is that why this `static_cast` is expected to succeed?

src/client/cmd/launch.cpp
gerboland: I had similar thought, but was concerned in case you knew something I didn't, so left it. I've done as you asked.

src/client/cmd/launch.cpp
```diff
@@ -165,20 +165,7 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
             const QString& cloudInitFile = parser->value(cloudInitOption);
             if (cloudInitFile == "-")
             {
-                if (!mcp::is_tty())
-                {
-                    throw std::runtime_error("cannot read from stdin without a TTY");
-                }
-                mcp::prepare_stdin_for_read();
```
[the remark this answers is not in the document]
gerboland: Quite right, I forgot that, thanks

src/platform/console/console.cpp
gerboland: Yep we add extra bits for other platforms. The #ifdefs will prevent accidental mixing of platform components, which is why that static_cast will succeed - on a unix-system, only UnixTerminal will exist.

ricab:
bors r=[townsend2010, ricab]

PR #660, opened 2019-03-04
daemon/suspend: Stop mounts before suspending instance

### LINKED ISSUE #608: Resuming instance with a mount results in an error
```
$ multipass info snapcraft-multipass
Name:           snapcraft-multipass
State:          SUSPENDED
IPv4:           --
Release:        --
Image hash:     56e742b3e689 (Ubuntu Snapcraft builder for Core 18)
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         /home/michal/dev/canonical/multipass/multipass => /root/project
                    UID map: 10000:0
                    GID map: 10001:0
$ multipass start snapcraft-multipass
Removing "/root/project": mkdir: cannot create directory ‘/root/project’: File exists
```

It seems we're trying to mount over an existing mount. We should kill (`fusermount -uz`?) the mounts before suspending.

townsend2010:
Fixes #608

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1563,6 +1563,8 @@ try // clang-format on
 
         for (const auto& name : instances_to_suspend)
         {
+            stop_mounts_for_instance(name);
+
             QTimer timer;
             QEventLoop event_loop;
 


townsend2010:
Just going to invoke bors for posterity's sake.

bors r=gerboland

PR #661, opened 2019-03-04
tests: Use namespace on function paramaters to fix build errors

townsend2010:
Some platforms are sensitive to name spacing and have build errors
when the namespace is left out.


townsend2010:
bors r=ricab

PR #665, opened 2019-03-05
daemon: Remove timer that waits for the suspend operation to finish

### LINKED ISSUE #609: SSH timeout error on `suspend`
Suspending an instance took longer than usual (`iowait` likely) and multipass threw an error out:
```
$ multipass suspend snapcraft-multipass                 
suspend failed: [ssh proc] failed to open session channel: '' 
```

townsend2010:
The EventLoop tied to the QTimer was causing unintended behavior.  Instead, just wait
on suspend to finish in the particular backend code.

Fixes #609

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1574,7 +1574,7 @@ try // clang-format on
         }
 
         status = cmd_vms(instances_to_suspend, [this](auto& vm) {
-            stop_mounts_for_instance(vm.vm_name);
+            this->stop_mounts_for_instance(vm.vm_name);
             vm.suspend();
             return grpc::Status::OK;
         });
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1573,13 +1573,11 @@ try // clang-format on
                 instances_to_suspend.push_back(pair.first);
         }
 
-        for (const auto& name : instances_to_suspend)
-        {
-            stop_mounts_for_instance(name);
-
-            auto it = vm_instances.find(name);
-            it->second->suspend();
-        }
+        status = cmd_vms(instances_to_suspend, [this](auto& vm) {
+            stop_mounts_for_instance(vm.vm_name);
+            vm.suspend();
+            return grpc::Status::OK;
+        });
     }
 
     return status;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1577,18 +1577,8 @@ try // clang-format on
         {
             stop_mounts_for_instance(name);
 
-            QTimer timer;
-            QEventLoop event_loop;
-
-            QObject::connect(this, &Daemon::suspend_finished, &event_loop, &QEventLoop::quit, Qt::QueuedConnection);
-            QObject::connect(&timer, &QTimer::timeout, &event_loop, &QEventLoop::quit);
-
             auto it = vm_instances.find(name);
             it->second->suspend();
-
-            timer.setSingleShot(true);
-            timer.start(std::chrono::seconds(30));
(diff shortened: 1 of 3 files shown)


src/daemon/daemon.cpp
```diff
@@ -1778,7 +1768,6 @@ void mp::Daemon::on_stop()
 
 void mp::Daemon::on_suspend()
```
ricab: I suppose this override could be removed from the daemon entirely?

src/daemon/daemon.cpp
```diff
@@ -1563,18 +1563,8 @@ try // clang-format on
 
         for (const auto& name : instances_to_suspend)
```
ricab: Might as well replace the loop with a call to cmd_vms...

src/daemon/daemon.h
```diff
@@ -123,9 +123,6 @@ public slots:
     grpc::Status version(grpc::ServerContext* context, const VersionRequest* request,
                          grpc::ServerWriter<VersionReply>* response) override;
 
-signals:
```
ricab: Question: isn't there a corresponding connect call to remove too?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -370,6 +370,7 @@ void mp::QemuVirtualMachine::suspend()
             update_state();
 
             update_shutdown_status = false;
+            vm_process->waitForFinished();
```
ricab: No parallel in other backends?

src/daemon/daemon.cpp
townsend2010: It could, but I think that is out of scope for this PR.  The other `on_foo()` methods are the same as well.

src/daemon/daemon.h
townsend2010: Yes, and it has been removed in line 1569 in daemon.cpp above.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Not the `libvirt` backend since all operations go through `libvirtd`, there is really nothing to wait on.

src/daemon/daemon.cpp
townsend2010: Sure, this is a good candidate to switch over to that.

src/daemon/daemon.cpp
ricab: I understand what you say in regard to the other dangling `on_foo` methods, but this one *becomes* dangling in this PR if not removed. Personally I would have nothing against cleaning other `on_foo` methods here (that is justified by consistency with other PR changes IMO), but that's a separate question.

src/daemon/daemon.h
ricab: Ah sorry I overlooked that

src/platform/backends/qemu/qemu_virtual_machine.cpp
ricab: Ack

src/daemon/daemon.cpp
townsend2010: I understand what you are saying about it becoming dangling, but I'm not keen on messing with the VMStatusMonitor class with this PR.

src/daemon/daemon.cpp
townsend2010: Ok, I updated this to use `cmd_vms()`.

townsend2010:
bors r=ricab

townsend2010:
bors r=ricab

townsend2010:
bors r=ricab

PR #674, opened 2019-03-11
Locate socket in $SNAP_COMMON while snapped

gerboland:
The multipass socket can be located in $SNAP_COMMON while both the daemon and client are in the same snap.

While doing this, deduplicate reads of the SNAP env var by using a small utility.

### DIFF
--- a/include/multipass/snap_utils.h
+++ b/include/multipass/snap_utils.h
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2019 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


gerboland:
For context, having the socket in $SNAP_COMMON, it means there's no need to create separate client & daemon interfaces in snapd. And until we're guaranteeing our RPC, the CLI is the only interface we support.

Naturally, when we want to expose our socket to other snaps, the snapd interfaces can be amended to suit this use-case.

gerboland:
In retrospect, I dislike the premature optimisation here (i.e. statics). Makes everything that uses this harder to test. Will change

Saviq:
Hey, there's conflicts here now, rebase please :)

townsend2010:
Hmm, where are you bors...

bors r+

PR #675, opened 2019-03-11
Have git ignore snapcraft build artifacts

gerboland:
For those of us who build on host


Saviq:
Will this still matter with `core18` though?

gerboland:
It makes my life easier as I build with `SNAPCRAFT_BUILD_ENVIRONMENT=host` for core18 too. It lets me iterate faster, and use `snap try` for testing.

Saviq:
FWIW `--destructive-mode` is the only public version of this. But OK!

bors r+

PR #663, opened 2019-03-05
Extract instance creation from launch

ricab:
Place instance creation in a separate method, to prepare the way to create without launching.

PR #676, opened 2019-03-12
Extract instance creation from launch and provide it over RPC

ricab:
Reissue of #663 

This can be merged presently. Client side to come separately.


tests/test_daemon.cpp
```diff
@@ -96,6 +102,94 @@ struct StubNameGenerator : public mp::NameGenerator
...
+    {
+        auto on_success = [](mp::CreateReply& /*reply*/) { return mp::ReturnCode::Ok; };
+        auto on_failure = [this](grpc::Status& status)
+        {
+            mp::CreateError create_error;
+            create_error.ParseFromString(status.error_details());
+            const auto errors = create_error.error_codes();
+
+            cerr << "fail: ";
+            if(errors.size() == 1)
```
gerboland: Clang-format inserts space between `if` and `{` for me. And for below few lines. Did you run it?

src/daemon/daemon.cpp
```diff
@@ -1993,6 +1911,111 @@ std::string mp::Daemon::check_instance_exists(const std::string& instance_name)
...
+                               config->ssh_username,
+                               VirtualMachine::State::off,
+                               {},
+                               false,
+                               QJsonObject()};
+    persist_instances();
+
+    if (start)
+    {
+        LaunchReply reply;
```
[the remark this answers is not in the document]
ricab: Yeah, that was also my thought process. I initially tried as you suggested, but then found it better to avoid complex returns. 

Now that you mention this though, `create_vm` could be better named, to reserve the term "create" for creation without starting (which is what the RPC does). "Make" and "generate" would kind of imply returning the vm. "Spawn" could kind of imply starting it. Perhaps `sprout_vm`? WDYT? Better suggestions?

tests/test_daemon.cpp
ricab: Must have forgotten, sorry

ricab: OK, I just pushed the fix. Man I need to write myself a clang-format hook...

src/daemon/daemon.cpp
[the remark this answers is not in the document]
gerboland: TBH I'm good with it as "create"

Saviq:
Sorry, got confused with `bors` here. All in business now.

PR #681, opened 2019-03-14
Cleanse backend memory args

### LINKED ISSUE #677: qemu backend tries to allocate millions of megabytes of memory
Between #654 and #650 the qemu backend started allocating petabytes of memory... Failing, obviously.

ricab:
Fixes #677

On investigation, back-end tools use slightly different formats to represent digital info sizes (bytes). So, our previous assumption that a *bytes string* would be correctly interpreted everywhere is not actually true. IOW, neither plain strings nor plain numbers are adequate to convey memory sizes uniformly, so the earlier *shortcut implementation* falls short. 

This PR encapsulates memory sizes in a dedicated type: `MemorySize`. This type does the necessary arithmetic, but it is the responsibility of each (internal) back-end implementation to derive the string representation they need (unit and unit representation).


include/multipass/exceptions/invalid_memory_size_exception.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (C) 2018-2019 Canonical, Ltd.
```
Saviq: New file, should probably be 2019?

src/client/cmd/launch.cpp
```diff
@@ -258,14 +258,12 @@ mp::ReturnCode cmd::Launch::request_launch()
             if (error == LaunchError::INVALID_DISK_SIZE)
             {
                 error_details =
-                    fmt::format("Invalid disk size value supplied: {}. Note that a minimum of {} is required.",
-                                request.disk_space(), min_disk_size);
+                    fmt::format("Invalid disk size value supplied: {}.", request.disk_space(), min_disk_size);
```
Saviq: This doesn't mention the minimum disk size now, but still passes it to `fmt::format()`?

src/client/cmd/launch.cpp
```diff
@@ -258,14 +258,12 @@ mp::ReturnCode cmd::Launch::request_launch()
...
             }
             else if (error == LaunchError::INVALID_MEM_SIZE)
             {
                 error_details =
-                    fmt::format("Invalid memory size value supplied: {}. Note that a minimum of {} is required.",
-                                request.mem_size(), min_memory_size);
+                    fmt::format("Invalid memory size value supplied: {}.", request.mem_size(), min_memory_size);
```
Saviq: And here?

src/daemon/daemon.cpp
```diff
@@ -278,23 +277,40 @@ auto fetch_image_for(const std::string& name, const mp::FetchType& fetch_type, m
...
+
+    auto mem_size_str = request->mem_size();
+    auto disk_space_str = request->disk_space();
     auto instance_name = request->instance_name();
     auto option_errors = mp::LaunchError{};
 
-    const auto opt_mem_size = mp::utils::in_bytes(mem_size.empty() ? "1G" : mem_size);
-    const auto opt_disk_space = mp::utils::in_bytes(disk_space.empty() ? "5G" : disk_space);
+    const auto opt_mem_size = try_mem_size(mem_size_str.empty() ? "1G" : mem_size_str);
+    const auto opt_disk_space = try_mem_size(disk_space_str.empty() ? "5G" : disk_space_str);
```
Saviq: IIUC this silently accepts invalid entries and "converts" them into the defaults? Should this not be an error instead?

src/utils/memory_size.cpp
```diff
@@ -0,0 +1,121 @@
...
+{
+constexpr auto kilo = 1024LL;
+constexpr auto mega = kilo * kilo;
+constexpr auto giga = mega * kilo;
+
+long long in_bytes(const std::string& mem_value)
+{
+    QRegExp matcher("^(\\d+)([KMG])?B?$", Qt::CaseInsensitive);
+
+    if (matcher.exactMatch(QString::fromStdString(mem_value)))
```
Saviq: If using `exactMatch()`, you can omit the `^` and `$` in the expression above.

src/utils/memory_size.cpp
```diff
@@ -0,0 +1,121 @@
...
+    {
+        auto val = matcher.cap(1).toLongLong(); // value is in the second capture (1st one is the whole match)
+        const auto unit = matcher.cap(2);       // unit in the third capture (idem)
```
Saviq: This is standard for regex captures, not sure if the comments are necessary.

tests/test_memory_size.cpp
```diff
@@ -0,0 +1,309 @@
...
+    EXPECT_EQ(size.in_bytes(), val);
+}
+
+TEST(MemorySize, interprets0G)
+{
+    constexpr auto val = 0;
+    const auto size = mp::MemorySize{std::to_string(val) + "G"};
+
+    EXPECT_EQ(size.in_bytes(), val);
+}
```
Saviq: No luck with parameterized tests?

include/multipass/exceptions/invalid_memory_size_exception.h
ricab: Sorry, will fix it

src/client/cmd/launch.cpp
ricab: Oops, nice catch. I removed the "minimum size" hint because, when testing, I found it was actually confusing when the cause for failure was something else. But did forget that, will fix. (When we have time to spare, we should move to format's compilation time checking)

src/client/cmd/launch.cpp
ricab: thanks

src/daemon/daemon.cpp
ricab: I thought the RPC parameter would be empty only when the user did not provide it, in which case we want the default, no? Anyway, I believe this comes directly from transforming what we had before, but I will have a better look.

src/utils/memory_size.cpp
ricab: Good point, will do that (to make the regex simpler to read).

src/utils/memory_size.cpp
ricab: I understand your point. I just commented what I thought I would forget ("document what you know when you know it" I read somewhere and caught in my ear). I suppose I have used regex groups as much as the next programmer, but that is still sparsely enough that I always end up forgetting these details. I wanted to ease it a bit on the reader and save myself (and others) from having to go consult the documentation when I read this code in a couple of months.

But I agree comments should add something and the code is hopefully clear enough. So let me know if you still think they are not helpful and I will remove it.

src/daemon/daemon.cpp
Saviq: Sorry, just read properly below, it does error out there.

src/utils/memory_size.cpp
Saviq: It's ok, they can stay :)

tests/test_memory_size.cpp
ricab: Heh, I had, but in other places :P 

Yeah... I agree this would be the exact sort of place to have them. In my defense, I just wanted to be quick about it and so transferred existing tests from the previous "mem-sanitizing" approach with minimal change (multiple cursors in sublime make that easier). I think I wrote those tests before the ones I referred to when I talked about parameterized tests. But let me do that now.

ricab:
Hopefully this now addresses all review points. 

Parameterizing the tests was the challenging bit. I eventually stopped trying to wrangle with [parameter generators](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-value-parameterized-tests) and implemented generation myself. 

Concerning the remaining plain tests, I found they were not such good candidates for parameterizing without dramatically reducing readability. Let me know if you think otherwise.

As an aside, I suggest we consider [property-based tests](http://www.se-radio.net/2018/05/se-radio-episode-322-bill-venners-on-property-based-tests/) at some point.

As another aside, we're about to cross the 2/3 coverage threshold! :smiley:

Saviq:
Yes, thanks!

bors r+

PR #695, opened 2019-03-25
set static version

Saviq:


Saviq:
This has served its purpose.

PR #699, opened 2019-03-29
Release 0.6.0

Saviq:


CMakeLists.txt
```diff
@@ -68,6 +68,18 @@ function(determine_version OUTPUT_VARIABLE)
     message(FATAL_ERROR "Failed to parse version number: ${GIT_VERSION}")
   endif()
 
+  if(APPLE OR MSVC)
+    string(REGEX MATCH "\\+full$" FULL_SUFFIX_MATCH ${NEW_VERSION})
+    if(NOT FULL_SUFFIX_MATCH)
+      set(NEW_VERSION ${NEW_VERSION}.full)
```
ricab: Shouldn't this be `+full` instead?

Saviq: No, the only two cases I'm expecting here are:
```
v0.7.0-dev.123+g456abc
v0.7.0+full
```
So we need to append `.full` to end up with dot-separated metadata.

ricab: OK thanks

townsend2010:
Hmm, I think I need to make sure all the ducks are in a row before this merges.

bors r-

townsend2010:
Ok, I think it's good.

bors r+

townsend2010:
bors r=ricab,townsend2010

townsend2010:
bors r=ricab,townsend2010

PR #720, opened 2019-04-10
 daemon: Make multiple instance starts completely asynchronous 

townsend2010:
Also allow for concurrent starts on the same instance.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -45,6 +45,7 @@
 
 #include <QDir>
 #include <QEventLoop>
+#include <QFutureSynchronizer>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
@@ -1403,12 +1404,13 @@ try // clang-format on
     for (const auto& name : vms)
     {
         auto it = vm_instances.find(name);
-        it->second->start();
+        if (it->second->current_state() != VirtualMachine::State::starting)
+            it->second->start();
     }
 
     auto future_watcher = create_future_watcher();
     future_watcher->setFuture(
-        QtConcurrent::run(this, &Daemon::async_wait_for_ssh_and_start_mounts<StartReply>, server, vms, status_promise));
+        QtConcurrent::run(this, &Daemon::async_wait_for_ready_all<StartReply>, server, vms, status_promise));
 }
 catch (const std::exception& e)
 {
@@ -1512,8 +1514,8 @@ try // clang-format on
         if (status.ok())
         {
             auto future_watcher = create_future_watcher();
-            future_watcher->setFuture(QtConcurrent::run(
-                this, &Daemon::async_wait_for_ssh_and_start_mounts<RestartReply>, server, instances, status_promise));
+            future_watcher->setFuture(QtConcurrent::run(this, &Daemon::async_wait_for_ready_all<RestartReply>, server,
+                                                        instances, status_promise));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/daemon/daemon.cpp
```diff
@@ -2070,40 +2072,19 @@ QFutureWatcher<mp::Daemon::AsyncOperationStatus>* mp::Daemon::create_future_watc
...
-
-    return grpc::Status::OK;
-}
-
 template <typename Reply>
-mp::Daemon::AsyncOperationStatus
-mp::Daemon::async_wait_for_ssh_and_start_mounts(grpc::ServerWriter<Reply>* server, const std::vector<std::string>& vms,
-                                                std::promise<grpc::Status>* status_promise)
+std::string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::string& name,
+                                                                grpc::ServerWriter<Reply>* server)
```
gerboland: Using return of an empty string to indicate success, and non-empty to indicate error, bothers me greatly.

I know in standard C++, the recommended paradigm for passing exceptions threads is to use `exception_ptr` to  make a copy of the exception and pass it to the future and rethrow it. Something like this:
https://stackoverflow.com/questions/233127/how-can-i-propagate-exceptions-between-threads

Could you give it a try?

src/daemon/daemon.cpp
```diff
@@ -2131,17 +2112,66 @@ mp::Daemon::async_wait_for_ssh_and_start_mounts(grpc::ServerWriter<Reply>* serve
...
+
+template <typename Reply>
+mp::Daemon::AsyncOperationStatus mp::Daemon::async_wait_for_ready_all(grpc::ServerWriter<Reply>* server,
+                                                                      const std::vector<std::string>& vms,
+                                                                      std::promise<grpc::Status>* status_promise)
+{
+    QFutureSynchronizer<std::string> start_synchronizer;
+    for (const auto& name : vms)
+    {
+        std::lock_guard<decltype(start_mutex)> lock{start_mutex};
```
gerboland: Instead of locking and unlocking for each iteration of the loop, why not lock outside the for loop?

townsend2010: Sure, I can do that.  I was just trying to take advantage of the automatic unlock in the scope of the `for` loop, but sure I can just add a new scope encapsulating the `for` loop and only do one lock for the whole thing.

src/daemon/daemon.cpp
townsend2010: @gerboland,

Sure, I understand what you are saying, but in this case, it's really not a case of exception detection per se, but more about passing **all** error strings back to the client via `grpc_status_for()`.  You can see many examples of where we do that in the daemon code.

In `async_wait_for_ready_all()`, `start_synchronizer` is responsible for coalescing the error strings returned back from the thread(s) and then call `grpc_status_for()` on the full error string and based on that, the proper grpc status is set along with the error string.

Does that make sense or are you still bothered by it?  I kind of feel we will accomplish the same exact thing by trying to pass the exception via the thread as to what we do here at the cost of complication.

gerboland: I'm still bothered by it, but could live with it if we make it more explicit that the string returned by this method are error string(s)? Because right now it is not obvious what this is returning.

Even a `typedef std::string errorstring` would help satisfy me.

townsend2010: Ok, I made at least the change to define the return type to be something more meaningful.  Perhaps later we can explore exception passing in the future if we find there is a need for it.

src/daemon/daemon.cpp
townsend2010: I moved this into its own scope and moved the lock outside of the `for` loop.

PR #730, opened 2019-04-18
WIP: Cache default img

ricab:


ricab:
This is missing two essential parts:
- offloading to a separate thread (so that the daemon does not block upon construction)
- unit tests

ricab:
stupid bors...

PR #731, opened 2019-04-18
Cache default img, new attempt

ricab:
bors closed #730 after the base there got merged and I could not reopen it or change the base


ricab:
This is too outdated now. I am keeping the branch locally for future reference.

PR #739, opened 2019-04-22
backends/qemu: Use '-drive' option for already suspended instances

### LINKED ISSUE #738: Resume from suspend potentially broken due to the fix introduced by #727
The fix introduced by #727 can break resuming from suspend due to the incompatibility of the device type use.  Here is the error:
```
[warning] [qemu-system-x86_64] qemu-system-x86_64: Unknown savevm section or instance '0000:00:04.0/virtio-blk' 0

[warning] [qemu-system-x86_64] qemu-system-x86_64: Error -22 while loading VM state
```
This **only** occurs when an instance was suspended with a daemon version prior to #727 and the resumed with a daemon version containing that fix such as when the Multipass snap updates.

townsend2010:
This fixes the case where an instance is suspended using '-drive' for the cloud-init
iso and then the newer version of the daemon is started that uses the '-cdrom' option
instead and qemu fails to load the saved memory snapshot.

Fixes #738

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -145,6 +145,7 @@ auto get_metadata()
     QJsonObject metadata;
 
     metadata["machine_type"] = get_qemu_machine_type();
+    metadata["use_cdrom"] = true;
 
     return metadata;
 }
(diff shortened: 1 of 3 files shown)


Saviq:
This will bite us every time we try to change the `qemu` options, shouldn't we just save the whole argument list instead of adding stuff on a case-by-case basis?

townsend2010:
@Saviq,

Yes, that makes sense for a long term solution, but in this case, it won't help.  The problem being that when multipass upgrades with the new `-cdrom` option, the `qemu` arguments won't be saved (because the instance is already running) and so we are in the same situation as now.

Saviq:
Right, whereas your approach actually makes existing suspended instances work…

bors r+

PR #742, opened 2019-04-23
launch/find: Allow using and searching for unsupported cloud images

### LINKED ISSUE #736: multipass shouldn't refuse to launch unsupported images
In #734, we ran in to problems launching trusty VMs because streams had ([incorrectly](https://bugs.launchpad.net/cloud-images/+bug/1825553)) started reporting trusty as unsupported.

A reasonable use of VMs is to isolate applications that rely on unsupported releases of Ubuntu; I've had to do this a couple of times for abandonware that I needed to use temporarily. With the current multipass behaviour, I wouldn't be able to do that.

I think it would be preferable for multipass to allow me to bypass this check somehow, to enable this use case.

townsend2010:
This uses an `--allow-unsupported` option for `launch` and `find`.

Fixes #736

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -61,7 +61,7 @@ _multipass_complete()
             opts="${opts} --all --purge"
         ;;
         "launch")
-            opts="${opts} --cpus --disk --mem --name --cloud-init --allow-unsupported"
+            opts="${opts} --cpus --disk --mem --name --cloud-init"
         ;;
         "mount")
             opts="${opts} --gid-map --uid-map"
@@ -70,7 +70,7 @@ _multipass_complete()
             opts="${opts} --all"
         ;;
         "find")
-            opts="${opts} --allow-unsupported"
+            opts="${opts} --show-unsupported"
         ;;
     esac
 
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -61,13 +61,16 @@ _multipass_complete()
             opts="${opts} --all --purge"
         ;;
         "launch")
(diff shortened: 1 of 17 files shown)


townsend2010:
Trying to launch an unsupported image without using the `--unsupported` option will fail until #741 is merged.

Saviq:
> "multipass launch --unsupported" ends up launching bionic, which surprised me. IMO it should complain an image name wasn't specified.

That would mean changing how the command treats positional arguments (i.e. that it requires one), when it doesn't without that option… While I agree with what you wrote above, I'm not sure I'd welcome that behavioral change… I suppose if the error and help messages would be clear…

townsend2010:
Ok, I changed the option name to `--allow-unsupported` to be more explicit about the intent of the option.

Saviq:
Slight twist from a brainstorm:
```
multipass launch trusty  # make this Just Work 
multipass find  # only lists supported
multipass find --allow-unsupported  # lists them all
```

townsend2010:
@Saviq,

So if use `--allow-unsupported` in `find`, we need that in `launch` as well in order to launch unsupported images.

Also, so we need quirk `trusty` to Just Work?  What about other aliases associated with Trusty?  The way we have this implemented and rely so much on SimpleStreams for what's supported and not supported, I don't think this will be real trivial to implement.

Saviq:
As discussed on IRC:
```
multipass launch *  # this should Just Work™, the user explicitly asked for it
multipass find  # only lists supported
multipass find --show-unsupported  # lists them all
```

townsend2010:
Ok, this is updated with the new way of doing this.

PR #757, opened 2019-04-30
Warn unused param

### LINKED ISSUE #697: Activate compiler warnings on unused parameters
We currently add [`-Wno-unused-parameter`](https://github.com/CanonicalLtd/multipass/blob/3febdd7481189fe4b220d0ca842313c7f2337028/CMakeLists.txt#L88) in cmake. This makes it easy to leave unused parameters around, sometimes propagating them to a lot of code. 

For instance, there is an unused `grpc::ServerContext*` parameter that is propagated from the Rpc to all daemon handling methods and beyond (this particular case is currently being fixed in #682).

ricab:
Activates unused parameter warnings. Fixes all such cases, homogenizing implicated code.

Fixes #697

### DIFF
--- a/3rd-party/CMakeLists.txt
+++ b/3rd-party/CMakeLists.txt
@@ -63,6 +63,8 @@ target_link_libraries(gRPC INTERFACE
   libprotobuf
   zlibstatic)
 
+target_compile_options(gRPC INTERFACE "-Wno-unused-parameter")
+
 # YAML C++
 # Disable tests here to avoid double-including gtest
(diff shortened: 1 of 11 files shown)


ricab:
@gerboland Thanks for the feedback. I did try to mention it in the messages. See fd1dd09 and 
c62cf00. Did you perhaps not click the ellipsis at the end of the commit title in Github?

gerboland:
So you did. I was more looking at the PR title.
bors r+

PR #759, opened 2019-05-02
image_vault: Allow for simultaneous launching of new images

### LINKED ISSUE #664: Concurrent `launch` requests duplicate downloads
Steps:
- run `multipass launch <not-cached-image>` twice

Expected:
- only one download follows

Current:
- both requests download the image separately

townsend2010:
The first launch will download and prepare the image while other launches of the
same image will block until the download and prepare is complete.

Fixes #664

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -56,7 +56,6 @@
 #include <cassert>
 #include <functional>
 #include <stdexcept>
-#include <unordered_set>
 #include <utility>
 
 namespace mp = multipass;
@@ -1849,7 +1848,19 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                                                       create_error.SerializeAsString()));
     }
 
+    if (preparing_instances.find(name) != preparing_instances.end())
+    {
+        CreateError create_error;
+        create_error.add_error_codes(CreateError::INSTANCE_EXISTS);
+
+        return status_promise->set_value(grpc::Status(grpc::StatusCode::INVALID_ARGUMENT,
+                                                      fmt::format("instance \"{}\" is being prepared", name),
+                                                      create_error.SerializeAsString()));
+    }
+
     config->factory->check_hypervisor_support();
+
+    preparing_instances.insert(name);
 
     auto prepare_future_watcher = new QFutureWatcher<VirtualMachineDescription>();
 
@@ -1870,6 +1881,8 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                                            {},
                                            false,
                                            QJsonObject()};
+                preparing_instances.erase(name);
+
                 persist_instances();
 
                 if (start)
@@ -1897,6 +1910,7 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
             }
             catch (const std::exception& e)
             {
+                preparing_instances.erase(name);
                 status_promise->set_value(grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), ""));
             }
 
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -268,235 +268,187 @@ mp::DefaultVMImageVault::DefaultVMImageVault(std::vector<VMImageHost*> image_hos
 mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, const Query& query,
                                                  const PrepareAction& prepare, const ProgressMonitor& monitor)
 {
-    auto name_entry = instance_image_records.find(query.name);
-    if (name_entry != instance_image_records.end())
     {
-        const auto& record = name_entry->second;
+        std::lock_guard<decltype(fetch_mutex)> lock{fetch_mutex};
+        auto name_entry = instance_image_records.find(query.name);
+        if (name_entry != instance_image_records.end())
+        {
+            const auto& record = name_entry->second;
 
-        return record.image;
+            return record.image;
+        }
     }
 
-    if (query.query_type != Query::Type::Alias)
-    {
-        if (!mp::platform::is_image_url_supported())
-            throw std::runtime_error(fmt::format("http and file based images are not supported"));
+    if (query.query_type != Query::Type::Alias && !mp::platform::is_image_url_supported())
+        throw std::runtime_error(fmt::format("http and file based images are not supported"));
 
+    if (query.query_type == Query::Type::LocalFile)
+    {
         QUrl image_url(QString::fromStdString(query.release));
         VMImage source_image, vm_image;
 
-        if (image_url.isLocalFile())
+        if (!QFile::exists(image_url.path()))
+            throw std::runtime_error(fmt::format("Custom image `{}` does not exist.", image_url.path().toStdString()));
+
+        source_image.image_path = image_url.path();
+
+        if (source_image.image_path.endsWith(".xz"))
         {
-            if (!QFile::exists(image_url.path()))
-                throw std::runtime_error(
-                    fmt::format("Custom image `{}` does not exist.", image_url.path().toStdString()));
-            source_image.image_path = image_url.path();
-
-            if (source_image.image_path.endsWith(".xz"))
-            {
-                source_image = extract_image_from(query.name, source_image, monitor);
-            }
-            else
-            {
-                source_image = image_instance_from(query.name, source_image);
-            }
-
-            if (fetch_type == FetchType::ImageKernelAndInitrd)
-            {
-                Query kernel_query{query.name, "default", false, "", Query::Type::Alias};
-                auto info = info_for(kernel_query);
-
-                source_image = fetch_kernel_and_initrd(info, source_image,
-                                                       QFileInfo(source_image.image_path).absoluteDir(), monitor);
-            }
-
-            vm_image = prepare(source_image);
-            remove_source_images(source_image, vm_image);
+            source_image = extract_image_from(query.name, source_image, monitor);
         }
         else
         {
+            source_image = image_instance_from(query.name, source_image);
+        }
+
+        if (fetch_type == FetchType::ImageKernelAndInitrd)
+        {
+            auto info = get_kernel_query_info(query.name);
+
+            source_image =
+                fetch_kernel_and_initrd(info, source_image, QFileInfo(source_image.image_path).absoluteDir(), monitor);
+        }
+
+        vm_image = prepare(source_image);
+        remove_source_images(source_image, vm_image);
+
+        {
+            std::lock_guard<decltype(fetch_mutex)> lock{fetch_mutex};
+            instance_image_records[query.name] = {vm_image, query, std::chrono::system_clock::now()};
+            persist_instance_records();
+        }
+
+        return vm_image;
+    }
+    else
+    {
+        std::string id;
+        optional<VMImage> source_image{nullopt};
+        VMImageInfo info;
+        QDir image_dir;
+
+        if (query.query_type == Query::Type::HttpDownload)
+        {
+            QUrl image_url(QString::fromStdString(query.release));
+
             // Generate a sha256 hash based on the URL and use that for the id
-            auto hash =
-                QCryptographicHash::hash(query.release.c_str(), QCryptographicHash::Sha256).toHex().toStdString();
+            id = QCryptographicHash::hash(query.release.c_str(), QCryptographicHash::Sha256).toHex().toStdString();
             auto last_modified = url_downloader->last_modified(image_url);
 
-            auto entry = prepared_image_records.find(hash);
+            std::lock_guard<decltype(fetch_mutex)> lock{fetch_mutex};
+            auto entry = prepared_image_records.find(id);
             if (entry != prepared_image_records.end())
             {
                 auto& record = entry->second;
 
                 if (last_modified.isValid() && (last_modified.toString().toStdString() == record.image.release_date))
                 {
-                    record.last_accessed = std::chrono::system_clock::now();
-                    auto vm_image = finalize_image_records(query, record.image);
-                    return vm_image;
+                    return finalize_image_records(query, record.image, id);
                 }
                 else
                 {
-                    source_image = record.image;
+                    *source_image = record.image;
                 }
             }
             else
             {
+                auto kernel_info = get_kernel_query_info(query.name);
+                info = static_cast<const VMImageInfo>(VMImageInfo{{},
+                                                                  {},
+                                                                  {},
+                                                                  {},
+                                                                  true,
+                                                                  image_url.url(),
+                                                                  kernel_info.kernel_location,
+                                                                  kernel_info.initrd_location,
+                                                                  QString::fromStdString(id),
+                                                                  {},
+                                                                  0,
+                                                                  false});
                 const auto image_filename = filename_for(image_url.path());
                 // Attempt to make a sane directory name based on the filename of the image
+
                 const auto image_dir_name = QString("%1-%2")
                                           .arg(image_filename.section(".", 0, image_filename.endsWith(".xz") ? -3 : -2))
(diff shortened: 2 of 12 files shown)


src/daemon/default_vm_image_vault.cpp
```diff
@@ -438,36 +454,47 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
...
+        auto prepared_image = future.result();
 
-        persist_image_records();
-        persist_instance_records();
+        lock.lock();
+        prepared_image_records[id] = {prepared_image, query, std::chrono::system_clock::now()};
+        auto vm_image = finalize_image_records(query, prepared_image);
+
+        in_progress_image_fetches.erase(id);
+        lock.unlock();
```
ricab: I am having a lot of trouble reading this and following all the locking and unlocking. Why is it needed? 

Repeated locking and unlocking means one thing: identifiable work units that could be extracted out. I think this would really profit from a different sort of organization, extracting individual operations and having them lock internally where needed. Something like:

```c++
QFuture<VMImage> get_existing_or_create_future_vm(const std::string& id)
{
    std::unique_lock<decltype(fetch_mutex)> lock{fetch_mutex};
    //...
    return future;
}

VMImage DefaultImageVault::fetch_image(...)
{
    auto future = get_existing_or_create_future_vm(id);
    auto my_image = future.result();
    further_operation(my_image); // if this needs to lock, it does inside
    // ...
    return my_image;
}
```

src/daemon/default_vm_image_vault.cpp
```diff
@@ -438,36 +454,47 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
         }
         DeleteOnException image_file{source_image.image_path};
 
-        url_downloader->download_to(info.image_location, source_image.image_path, info.size, LaunchProgress::IMAGE,
-                                    monitor);
+        auto future = QtConcurrent::run([&]() -> VMImage {
```
ricab: This is a clearly identifiable work unit, can it be extracted out into a proper function rather than local lambda?

src/client/cmd/launch.cpp
```diff
@@ -292,7 +292,8 @@ mp::ReturnCode cmd::Launch::request_launch()
             {LaunchProgress_ProgressTypes_KERNEL, "Retrieving kernel image: "},
             {LaunchProgress_ProgressTypes_INITRD, "Retrieving initrd image: "},
             {LaunchProgress_ProgressTypes_EXTRACT, "Extracting image: "},
-            {LaunchProgress_ProgressTypes_VERIFY, "Verifying image: "}};
+            {LaunchProgress_ProgressTypes_VERIFY, "Verifying image: "},
+            {LaunchProgress_ProgressTypes_WAITING, "Waiting for in progress image preparation"}};
```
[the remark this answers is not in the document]
townsend2010: I don't think it requires a hyphen as 'in progress' is perfectly acceptable for meaning "taking place" or "under way".  See https://www.collinsdictionary.com/us/dictionary/english/in-progress for how they spell it.

I agree about the colon- I just missed putting that in there.

I'd prefer to be more specific to let the user know that there is already another process of downloading and preparing the image taking place.  It's not necessarily from another client since, at some point, the automatic image updater could be doing this or the pet-env stuff (although neither of those do it right now).

Since I want to be specific, I'm not sure how `Waiting for in progress image preparation:` is less readable than "Image preparation in progress; waiting: ".  My way only adds the word "for", but your way adds a semicolon, so meh.  And my way keeps the consistency of beginning each phase with the present participle verb.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -605,6 +632,22 @@ mp::VMImage mp::DefaultVMImageVault::fetch_kernel_and_initrd(const VMImageInfo&
     return image;
 }
 
+mp::VMImage mp::DefaultVMImageVault::finalize_image_records(const Query& query, const VMImage& prepared_image)
+{
+    VMImage vm_image;
+
+    if (!query.name.empty())
```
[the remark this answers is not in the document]
townsend2010: Ok, I'll see if this can be dropped altogether.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -394,6 +393,24 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
 
         auto id = info.id.toStdString();
 
+        std::unique_lock<decltype(fetch_mutex)> lock{fetch_mutex};
```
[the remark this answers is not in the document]
townsend2010: I'll see if I can that, but it might actually be better to extract many parts out of this function.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -394,6 +393,24 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
...
+        {
+            auto future = it->second;
+            monitor(LaunchProgress::WAITING, -1);
+            lock.unlock();
+
+            auto prepared_image = future.result();
+
+            lock.lock();
+            prepared_image_records[id].last_accessed = std::chrono::system_clock::now();
+            auto vm_image = finalize_image_records(query, prepared_image);
```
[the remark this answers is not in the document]
townsend2010: It's not the only place it is locked.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -394,6 +393,24 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
...
+            lock.unlock();
```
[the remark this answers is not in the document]
townsend2010: Yes, I tried that originally, but with how this function is laid out, it would be nearly impossible to use scoped locks.  Perhaps with breaking the function up in pieces, it would be easier, but that's for another day.

src/daemon/default_vm_image_vault.cpp
townsend2010: The locking an unlocking is needed so different threads don't read/write to the json files that is done in `finalize_image_records` at the same time and potentially corrupt the files if writing at the same time.

I agree the locking/unlocking is messy and organizing and breaking down this function would be beneficial- we just need to weigh the time vs. benefit of doing such a thing.

src/daemon/default_vm_image_vault.cpp
townsend2010: Yes, sure, it's possible.

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: After looking at this, `query.name` can indeed be empty, such as when the background image updater runs.  That only downloads new images and doesn't prepare any instance images.  So no, it can't be dropped and I prefer to keep it here as well.

src/client/cmd/launch.cpp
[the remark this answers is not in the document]
ricab: Alright, good point, I am convinced on going with your version... except the hyphen :stuck_out_tongue: See https://getitwriteonline.com/articles/hyphenated-adjectives/ or https://www.grammarbook.com/punctuation/hyphens.asp . Saying that "Foo is in progress" is of course correct, but calling it an "in-progress Foo" turns it into a compound adjective and requires an hyphen to be perfectly correct. Just a tiny pedantic detail though...

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
ricab: Sure

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
ricab: Right, I really think it would be better to extract those pieces along with locking. It would make it much easier to understand and also maintain in the future.

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
ricab: OK, I meant there are other places where `finalize_image_records` is called but not locked. That looked suspicious to me.

src/daemon/default_vm_image_vault.cpp
ricab: IMO it is bound to save us time in the ~~long~~ medium-run.

src/daemon/default_vm_image_vault.cpp
ricab: Thanks

src/daemon/default_vm_image_vault.cpp
townsend2010: Ok, I'm looking at this and I think I can do better, especially around using `lock_guard`'s.

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: There is one place where the call is not locked, line 330 in this files diff, but that is in a different use case where we are not accounting for parallelism yet.  The other places it is called, it is locked.  That said, I'm making `finalize_image_records` itself use a `lock_guard` and then it won't be an issue:)

src/client/cmd/launch.cpp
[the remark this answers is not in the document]
townsend2010: lol, ok, point taken and learned something new:)

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
ricab: I see, alright then

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
ricab: OK, yeah that's what I had in mind. Sounds good, thanks!

ricab:
I left conversations open to easy tracking, but the plans above all sound good to me.

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: Based on how some other critical sections needed a `lock_guard`, I decided not to put the `lock_guard` in `finalize_image_records` itself.  Let me know what you think based on the new code.  I won't resolve this review comment yet:)

src/daemon/default_vm_image_vault.cpp
townsend2010: As mentioned above, I used `lock_guard`'s everywhere to take advantage of scoped based locking.  Hopefully this addresses some of your concerns.

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
gerboland: Can you add a comment then? As that's not obvious to me either

src/daemon/default_vm_image_vault.cpp
```diff
@@ -634,6 +709,13 @@ mp::VMImageInfo mp::DefaultVMImageVault::info_for(const mp::Query& query)
     throw std::runtime_error(fmt::format("Unable to find an image matching \"{}\"", query.release));
 }
 
+mp::VMImageInfo mp::DefaultVMImageVault::get_kernel_query_info(const std::string& name)
+{
+
```
gerboland: extra line break

townsend2010:
@gerboland,

I agree that this is becoming unwieldy wrt to the locking, etc.  That said, I don't think we should go down this particular path and then have you change it all at some other point.  There has already been much time spent doing this work and then to go forward with this implementation to have it changed again is not a good use of resources.  Now is the time to either move forward with this design, but make it less dependent on thread locking or stop before this is merged and change it all to the Observer pattern you mention.

I can take a step back from how I've proposed doing this and see if I can make it work in such a way that the daemon thread does all of the critical section work and only the `run()` threads just do downloading and image preparation.

gerboland:
> @gerboland,
> 
> I agree that this is becoming unwieldy wrt to the locking, etc. That said, I don't think we should go down this particular path and then have you change it all at some other point. There has already been much time spent doing this work and then to go forward with this implementation to have it changed again is not a good use of resources. Now is the time to either move forward with this design, but make it less dependent on thread locking or stop before this is merged and change it all to the Observer pattern you mention.

I don't want to block this landing this because (1) it is here and works and honestly it is a perfectly fine change, (2) it is something we need for the 1.0, and (3) the refactor I mention is simply theoretical and in my head. It will require a large amount of code changes across various components, which increases risk of new regressions being introduced. And it may end up being a worse option with its own deficiencies! In the next cycle we've scheduled time for dealing with tech debt, IMO I think it best to try it then.

townsend2010:
@gerboland,

Ok, then we'll go this route, pending my fixes to the bugs you found.  Thanks!

src/daemon/default_vm_image_vault.cpp
townsend2010: Fixed

gerboland:
>     * I can call "multipass launch --name x1 xenial" in two separate in quick succession, and the second one proceeds as if everything is ok. Then one of them fails with a "launch failed: Cannot resize instance image" error.

This bug still reproducable.

gerboland:
As is the second bug. I've definitely got this branch. To clarify, I always purge my downloaded image vault and restart the daemon before trying the steps I give for each bug.

gerboland:
> Actually there's a simpler to reproduce bug here: Start a download, then Ctrl+C it. Then try start download again. It fails for me with "Operation canceled"

This is fixed.

> However, I will fix the issue where it won't allow another download again after this.

Yep, this is also fixed.

> I can call "multipass launch --name x1 xenial" in two separate in quick succession, and the second one proceeds as if everything is ok. Then one of them fails with a "launch failed: Cannot resize instance image" error.

I can still reproduce this. Are you clearing the download cache before attempting to reproduce? (I do `sudo rm -rf $HOME/.cache/multipassd/`).

src/daemon/daemon.cpp
```diff
@@ -1849,8 +1848,20 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                                                       create_error.SerializeAsString()));
     }
 
+    if (preparing_instances.find(name) != preparing_instances.end())
+    {
+        CreateError create_error;
+        create_error.add_error_codes(CreateError::INSTANCE_EXISTS);
+
+        return status_promise->set_value(grpc::Status(grpc::StatusCode::INVALID_ARGUMENT,
+                                                      fmt::format("instance \"{}\" is being prepared", name),
```
gerboland: "already exists" - IMO clearer, and consistent with if VM was already ready

townsend2010: @gerboland, to me, "already exists" implies that if I do `list` or `info foo`, then it will show up, but it does not since the instance doesn't really exist yet.  This is why I chose the "being prepared" wording to differentiate the state to the user in a meaningful way.

townsend2010: I will also say I'm open to better wording there, but I don't think "already exists" is the right wording.

ricab: Suggestions:
- "instance name ... is reserved/locked/unavailable"
- "instance name ... has already been requested"

gerboland: @townsend2010 I understand your motivation, but I see no reason why a user would be interested in such a distinction. IMO the fact that a being-downloaded VM doesn't appear in `list` is a bug.

townsend2010: @gerboland, we can agree to disagree on this:)

As a data point, lxd/lxc behaves exactly like this in that a container that is launched and is downloading the image will not show up in `lxc list` until the container's download & prepare is complete.

townsend2010: But I do like @ricab's suggestion of putting something like "...has already been requested".  I think I will use that instead.

townsend2010:
@ricab, since you did some initial review of this and requested changes, are you good with it now?

ricab:
@townsend2010 I did not look closely again, but I see you separated things into a few more separate functions and used lock scoping rather than unlock, so that addresses my requests.

townsend2010:
bors r=gerboland,ricab

PR #762, opened 2019-05-03
Handle empty memory specs in old instance DBs

### LINKED ISSUE #760: Daemon fails to start when db has empty memory/disk string
This was reported on IRC:

>  I just tried to upgrade from an old Mac version 2018.4.2-full-65-g0f77459 to 0.6.1+mac, and the daemon is failing to start with "[error] [daemon]  is not a valid memory size" in the log

It turned out the instance db had empty strings for fields that are now supposed to be read into `MemorySize`, e.g.:

```
   "speciously-spotless-skimmer": {
        "disk_space": "",
        "mac_addr": "52:54:00:ef:15:57",
        "mem_size": "1G",
        "mounts": [
        ],
        "num_cores": 1,
        "ssh_username": "ubuntu",
        "state": 0
    },
```

Filling in `5G` in the disk_space fixed the issue.

The daemon should handle such cases by falling back to a default value when [loading the db]( https://github.com/CanonicalLtd/multipass/blob/ed3429a1da344d77d5556e70de5387b5dcb46df0/src/daemon/daemon.cpp#L262).

ricab:
Fixes #760

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -77,6 +77,8 @@ constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
 constexpr auto max_install_sshfs_retries = 3;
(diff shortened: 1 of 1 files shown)


Saviq:
bors r=gerboland

PR #778, opened 2019-05-10
[cli] rename `copy-files` to `transfer` (Fixes #748)

### LINKED ISSUE #748: Rename 'copy-files' to 'transfer'
Given the intent, 'transfer' is a cleaner verb for the movement of files from host to instance. It also gets rid of the dash :)

Saviq:


src/client/cmd/transfer.cpp
```diff
@@ -83,26 +83,31 @@ mp::ReturnCode cmd::CopyFiles::run(mp::ArgParser* parser)
...
+QString cmd::Transfer::short_help() const
+{
+    return QStringLiteral("Transfer files between the host and instances");
+}
+
+QString cmd::Transfer::description() const
 {
     // TODO: Don't mention directories until we support that
     // return QStringLiteral("Copy files and directories between the host and instances.");
     return QStringLiteral("Copy files between the host and instances.");
```
ricab: Good choice leaving this as is (clarifies what is meant by transfer).

src/client/cmd/transfer.cpp
```diff
@@ -83,26 +83,31 @@ mp::ReturnCode cmd::CopyFiles::run(mp::ArgParser* parser)
...
 {
-    return "copy-files";
+    return "transfer";
 }
 
-QString cmd::CopyFiles::short_help() const
+std::vector<std::string> cmd::Transfer::aliases() const
 {
-    return QStringLiteral("Copy files between the host and instances");
+    return {name(), "copy-files"};
```
ricab: I notice this is not reflected in the help, so the alias is hidden. Is that intended?

Saviq: Yes, we've just left it here for muscle-memory purposes.

PR #779, opened 2019-05-10
[qemu,libvirt] add "multipass" search domain (Fixes #666)

### LINKED ISSUE #666: Configure DNS with a local `.multipass` domain
To be able to `ping <instance>` from inside another instance, we need to set up a search domain through the DHCP server. Otherwise `systemd-resolved` refuses to resolve the name.

On QEMU, it should be as easy as adding `--domain multipass,local` to the `dnsmasq` command line.

On Hyper-V, `.mshome.net` is added by default, we should ensure it's in the search domains via `cloud-init`.

On macOS `bootp` isn't set up to give out a search domain, so we'd need to configure `systemd-resolved` to allow no-suffix lookups.

Saviq:


Saviq:
bors r=gerboland

PR #780, opened 2019-05-10
[travis] add clang build

Saviq:


Saviq:
Will close until we can #598.

PR #781, opened 2019-05-14
Try stdio transfer

Saviq:


townsend2010:
We merged this in via #789.

PR #784, opened 2019-05-15
[snap] add autostart for multipass.gui

Saviq:


townsend2010:
The other thing to consider is if we really want to force the systray icon to autostart?  I think this should be configurable by the user somehow or at least have some to allow the user to keep if it from autostarting, like an opt-out.

Saviq:
OK I'm taking this back. There doesn't seem a way to _not_ autostart an app that has this declaration. Folks will have to resort to the standard "Startup applications" GUI.

Issue #773, opened 2019-05-09
startup crash on corrupt instance database


Saviq:
A problem reported today showed we're fragile when it comes to logically-corrupt instance databases and crashing when we encounter anything out of the ordinary.

We should improve our resiliency there - recover the instance (if it's deleted, it definitely should not be running) and log a warning. If we can't recover without losing data, we should skip the instance and log an error. Crashing is the worst :)

A deleted instance (`deleted: true`) with a running state (`state: 4`) will crash in [`daemon.cpp:593`](https://github.com/CanonicalLtd/multipass/blob/v0.6.1/src/daemon/daemon.cpp#L593).

The instance should have been shut down / suspended when deleting, but that did not happen for some reason and `multipassd` started [crashing](https://github.com/CanonicalLtd/multipass/files/3161538/gdb.txt) since.

```
{
    "lpt": {
        "deleted": true,
        "disk_space": "5368709120",
        "mac_addr": "52:54:00:ac:7a:50",
        "mem_size": "1073741824",
        "metadata": {
            "machine_type": "pc-i440fx-xenial"
        },
        "mounts": [
        ],
        "num_cores": 1,
        "ssh_username": "multipass",
        "state": 4
    },
}
```

gerboland:
I've failed to reproduce this with current Multipass. Perhaps it was possible in an older release?

The only workaround I can propose is this:
https://github.com/CanonicalLtd/multipass/compare/avoid-crash-on-invalid-state

My other idea would be to remove the "deleted" property from the JSON file, and instead use `state: -1` - that I'd prefer as then it wouldn't be possible to have contradictory state in the JSON.

ricab:
I find the approach there to be reasonable, although I would think the log level should be higher than `debug`, to me it is a plain `error`. Error != bug, but if we're concerned with scaring the user, perhaps warning...

As to changing file content, notice we would still have to cope with old DBs...

gerboland:
Yep, supporting old DB would be a wart to drag around

PR #790, opened 2019-05-17
Detect invalid saved state (deleted VM still running), log and rectify

gerboland:
I can't figure out how the user got it in that state, but at least we should avoid crashing.
Fixes bug #773

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -595,6 +595,13 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
             config->vault->remove(name);
         }
 
+        // FIXME: somehow we're writing contradictory state to disk.
+        if (spec.deleted && spec.state != VirtualMachine::State::stopped)
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("{} is deleted but has state", name, spec.state));
+            spec.state = VirtualMachine::State::stopped;
+        }
+
         if (spec.state == VirtualMachine::State::running && vm_instances[name]->state != VirtualMachine::State::running)
         {
             assert(!spec.deleted);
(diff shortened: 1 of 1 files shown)


ricab:
Trying to put things in more practical light, what happens now on `recover`? or `mount`? We have no idea. We are now in our own *UB land*. Most likely we'll get leaks and eventually crashes, whose origin is harder to understand or overcome.

src/daemon/daemon.cpp
```diff
@@ -595,9 +595,9 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
             config->vault->remove(name);
         }
 
-        if (spec.state == VirtualMachine::State::running && vm_instances[name]->state != VirtualMachine::State::running)
+        if (spec.state == VirtualMachine::State::running && !spec.deleted &&
```
ricab: I thought this a bit more over at lunch. If we're getting inconsistent state from disk, I think we should log and correct, (to stopped or unknown).

```c++
if(spec.state == VirtualMachine::State::running)
    if(spec.deleted)
    {
        // log
        spec.state = VirtualMachine::State::unknown;
    }
    else if(vm_instances[name]->state != VirtualMachine::State::running)
    {
        [...]
    }

File: src/daemon/daemon.cpp [hunk 2]
```diff
@@ -595,6 +595,13 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
             config->vault->remove(name);
         }
 
+        // FIXME: somehow we're writing contradictory state to disk.
+        if (spec.deleted && spec.state != VirtualMachine::State::stopped)
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("{} is deleted but has state", name, spec.state));
+            spec.state = VirtualMachine::State::stopped;
```
townsend2010: Can't assign here as spec is a `const`.

gerboland:
I'm gonna remove this as I think I've a hand on the core problem

PR #794, opened 2019-05-20
[cli] format `find` output

Saviq:


src/client/cli/cmd/find.cpp
```diff
@@ -19,6 +19,7 @@
 #include "common_cli.h"
 
 #include <multipass/cli/argparser.h>
+#include <multipass/cli/formatter.h>
 
 #include <iomanip>
```
ricab: Is this now a leftover header?

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -135,3 +136,28 @@ std::string mp::TableFormatter::format(const ListReply& reply) const
...
+    {
+        auto aliases = image.aliases_info();
+
+        mp::format::filter_aliases(aliases);
+
+        auto alias = aliases.cbegin();
+        fmt::format_to(buf, "{:<24}{:<18}{:<17}{:<}\n",
+                       (alias->remote_name().empty() ? "" : alias->remote_name() + ":") + alias->alias(),
+                       fmt::format("{}", fmt::join(aliases.cbegin() + 1, aliases.cend(), ",")), image.version(),
+                       fmt::format("{} {}", image.os(), image.release()));
```
ricab: `snapcraft:*` images apparently have an empty os field, so the description starts with a space here, which breaks the alignment:

```
$ multipass find
Image                   Aliases           Version          Description
snapcraft:core          core16            20190515          Snapcraft builder for Core 16
snapcraft:core18                          20190515          Snapcraft builder for Core 18
core                    core16            20190424         Ubuntu Core 16
core18                                    20190213         Ubuntu Core 18
[...]
```

Do you see the same?

gerboland: yep I see it too

src/client/cli/cmd/find.cpp
Saviq: No, `chosen_formatter` is a `Formatter`.

src/client/cli/formatter/table_formatter.cpp
Saviq: Indeed, fixing.

src/client/cli/cmd/find.cpp
ricab: It looks confusing on github, but I was referring to `<iomanip>`

Saviq:
> Right, I'll need to include the remote name in there after all.

@gerboland fixed, there's some duplication with the `image_id` thus, think it worth it abstracting out into `format_utils.cpp`? At first I thought I'd use a `fmtlib` formatter, but I'd have to modify the `_ImageInfo` instance to filter the aliases out…

src/client/cli/cmd/find.cpp
Saviq: d'oh, yes - both that and `<sstream>` below it. Fixed.

tests/test_output_formatter.cpp
```diff
@@ -735,6 +771,39 @@ TEST_F(JsonFormatter, no_images_find_output)
...
+                           "        \"snapcraft:core18\": {\n"
+                           "            \"aliases\": [\n"
+                           "            ],\n"
+                           "            \"os\": \"\",\n"
+                           "            \"release\": \"Snapcraft builder for core18\",\n"
+                           "            \"remote\": \"snapcraft\",\n"
+                           "            \"version\": \"20190520\"\n"
+                           "        }\n"
+                           "    }\n"
+                           "}\n";
```
gerboland: Just a suggestion, raw string literals might make this nicer on the eye: https://stackoverflow.com/a/5460235

gerboland:
> @gerboland fixed, there's some duplication with the `image_id` thus, think it worth it abstracting out into `format_utils.cpp`? At first I thought I'd use a `fmtlib` formatter, but I'd have to modify the `_ImageInfo` instance to filter the aliases out…

I don't think it's worth the trouble tbh.

ricab:
> I don't think it's worth the trouble tbh.

@Saviq @gerboland I beg to differ. The "trouble" is negligible and quite worth eliminating tripplication. This is the sort of thing that accumulates with time and makes bigger trouble in the long run. Just my opinion.

Saviq:
@ricab @gerboland abstracted out a tiny utility for this then.

ricab:
@Saviq perfect, happy codebase says thanks :smiley: And me too, thanks a lot!

PR #797, opened 2019-05-21
tests: Use proper temp file for scp pull testing

### LINKED ISSUE #782: Test suite leaves a `bar` test behind
Whenever I run the test suite, an empty `bar` file gets created in `$PWD`.

townsend2010:
Fixes #782

### DIFF
--- a/tests/test_scp_client.cpp
+++ b/tests/test_scp_client.cpp
@@ -123,16 +123,22 @@ TEST_F(SCPClient, throws_on_push_file_invalid_source)
 
 TEST_F(SCPClient, throws_when_pull_file_request_not_newfile)
 {
+    mpt::TempDir temp_dir;
+    auto file_name = temp_dir.path() + "/test-file";
+
     auto scp = make_scp_client();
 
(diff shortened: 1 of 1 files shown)


tests/test_scp_client.cpp
```diff
@@ -123,16 +123,22 @@ TEST_F(SCPClient, throws_on_push_file_invalid_source)
 
 TEST_F(SCPClient, throws_when_pull_file_request_not_newfile)
 {
+    mpt::TempDir temp_dir;
+    auto file_name = temp_dir.path() + "/test-file";
```
ricab: Does the scp client handle the forward slash on windows?

ricab: I suppose so, as I see this in the code... https://github.com/CanonicalLtd/multipass/blob/4a9707473e1104ecb2d406fb37b0ebb469cfdbee/src/ssh/scp_client.cpp#L55

townsend2010: @ricab,

Yes, it's fine on Windows as other tests already use forward slashes and it works without issue.

ricab: ok, approving then

PR #800, opened 2019-05-22
backends/daemon: Allow for backend to determine its data/cache dirs

### LINKED ISSUE #774: Separate data for different backends
It just so happens that the currently supported backends - `qemu` and `libvirt` - work with the same subsystem, and so it's possible to switch between them and keep the same instances.

If we introduce other backends, dealing with the differences between them seems overcomplicated, when we can just keep a completely separate dataset for them - switching between backends would mean working with data in a different place altogether.

The backends should have an ability to declare a `data-format` string which would be used to separate the data for that particular backend.

Only new backends would implement that feature.

townsend2010:
Fixes #774

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -54,6 +54,7 @@ enum class TimeoutAction
 QDir base_dir(const QString& path);
 multipass::Path make_dir(const QDir& a_dir, const QString& name);
 bool is_dir(const std::string& path);
+QString backend_directory_path(const Path& path, const QString& subdirectory);
 std::string filename_for(const std::string& path);
 QString make_uuid();
 std::string contents_of(const multipass::Path& file_path);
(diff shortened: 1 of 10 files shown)


townsend2010:
The meat of this will be implemented in the Windows and Mac platforms.

Saviq:
Doesn't this complicate things a bit too much? Couldn't the platform+backend just return a string for the daemon to do with as it pleases?

townsend2010:
@Saviq,

I thought about the approach you suggest, and I thought that approach is at least as complicated as this approach.  However, thinking about it some more, it might be better to implement this in each backend's factory.  I'll look into that.

townsend2010:
@Saviq,

I updated this to have the backend's factory be the keeper of the data/cache directories.

townsend2010:
@Saviq,

Ok, updated again:)

PR #803, opened 2019-05-27
exec command should use non-interactive console

### LINKED ISSUE #667: getting output from multipass hangs when redirecting output to a file
When redirecting output of stdout to a file the following command will hang:

multipass exec my-instance -- systemd-analyze blame > output.log

I've found a workaround:

multipass exec my-instance -- systemd-analyze blame > output.log  < /dev/null

..however I'd rather not rely on a workaround.

gerboland:
Fixes #667

### DIFF
--- a/include/multipass/console.h
+++ b/include/multipass/console.h
@@ -45,7 +45,7 @@ public:
     virtual void write_console() = 0;
     virtual void exit_console() = 0;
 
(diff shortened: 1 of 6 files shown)


gerboland:
Hmm no, not quite there yet

gerboland:
There's a better way

Issue #766, opened 2019-05-06
Allow configuring the primary name


ricab:
It should be possible to configure the name of the primary instance (currently "primary"). Following current state of affairs in #756, this should be achievable with `multipass set client.primary_name <name>`.

ricab:
Initially, this may involve very basic `set`(/`get`/`config`) command(s) that only accept(s) the one key. To accommodate further configuration, however, these commands would probably need to rely on some intermediary object knowing what keys should be in the output, what to fill them with, and how to reflect config input on program state.

ricab:
@Saviq IIUC this should be in 0.7.0, right? I added it there, please change if needed.

ricab:
Looking better into this issue, I don’t think it makes sense to implement anything without a config intermediary.

Being an adept of _solving the problem, then coding_, I put together [some thoughts](https://docs.google.com/document/d/1pyQsMTPnR-iDqKeffo0bS7rSq-BwbZ0tbxdfx_l89QU) on how to approach this. @Saviq, @townsend2010, @gerboland, your feedback would be appreciated.

ricab:
After further team discussion, the current plans for this issue evolved:
- use `QSettings` directly
- use a specific per-user file for client settings (separate from the daemon's)
    * this file is shared by both CLI and systray clients
- create a `QSettings` object as needed, throwing away after use (both read and write use, no internal state, only the file keeps record of the settings)
- rely on Qt file watchers/locks to protect concurrent access

gerboland:
>     * rely on Qt file watchers/locks to protect concurrent access
Note that QSettings does this for you. We do need Qt file watcher for the daemon to watch for if the client changes the settings file though.

Saviq:
> Note that QSettings does this for you. We do need Qt file watcher for the daemon to watch for if the client changes the settings file though.

And for the systray, too.

ricab:
For the systray *only* actually, right? Since the daemon was supposed to be killed by the client: #776

Saviq:
Correct.

ricab:
Fixed by #805

Issue #776, opened 2019-05-09
Allow configuring the daemon driver


ricab:
It should be possible to configure the driver for the daemon to use. Following the current state of affairs in #756, this should be achievable with `multipass set local.driver <driver>`, where `<driver>` is one of `qemu`, `hyperkit`, `hyper-v`, `libvirt`.

The current driver should be obtainable with `multipass get local.driver`.

(This issue is a sister of #766)

ricab:
The team discussion mentioned [here](https://github.com/CanonicalLtd/multipass/issues/766#issuecomment-491014781) resulted in the following plan:
- as in #766 use `QSettings` directly, now in both client and daemon (file-backed settings, no RPC)
- use a specific file for daemon settings (separate from the client's)
- in the daemon, the file is only read (not written) and only upon startup
- client also reads/writes to the daemon file, which is how the settings get communicated to the daemon (no RPC)
- consider only one local daemon (follows from local in the single supported key)
- have the client restart the daemon when its config changes (Linux only to begin with)
    * with `snapctl restart $SNAP_NAME` if `$SNAP_NAME` ([ref](https://github.com/CanonicalLtd/multipass/blob/fc2d5d6516c0503f4a794c2009d2fa99a0ebddc0/snap/hooks/configure#L67))
    * `pkill multipassd` otherwise (or similar)
    * note: both will require root privileges, which is fine for now, but need to sort out interplay of settings file and sudo (home dir)

matthewpaulthomas:
Should this setting be provided in the GUI as well?

ricab:
I think this is an essential setting, so I guess it would be appropriate. But I'll let @Saviq, @townsend2010, @gerboland comment.

Saviq:
Hey @matthewpaulthomas I posted an RFC for a full `multipass config` experience #756, a lot of that would have to grow into the GUI in the end. I'd love your feedback there, including any remarks you might have for how to ensure GUI-compatibility from the get go.

For this particular one, "not yet" would be the answer.

ricab:
After some more consideration, we decided to drop daemon restarting from the client and replace it with settings-file monitoring in the daemon + exit (or reinit, possibly) on modification.

PR #805, opened 2019-05-28
Settings primary name

ricab:
This addresses #766


ricab:
Sys-tray GUI side still pending

ricab:
Note: although settings have a hierarchical appearance (i.e. their keys are structured with dots like `client.primary_name`), they are currently saved to disk in INI format (`QSettings` default). QSettings does not offer a YAML backend, but we should be able to [register one](https://doc.qt.io/qt-5.9/qsettings.html#registerFormat).

ricab:
This is ready to merge, as per discussion in IRC:
1.  the petenv does not intersect with the systray yet, so the file-watcher bit is postponed until then (when the systray gets top-level commands).
2. I am creating a separate issue for saving settings in YAML format

ricab:
> This is ready to merge

Travis disagrees, looking into it.

ricab:
It should be good now...

ricab:
oO really ? let me check

ricab:
@townsend2010 thanks for catching that, now I understand why I kept downloading images when switching branches :) (thought they were updating a lot)

I have pushed the fix. The issue was that, by setting organization name and domain in `QCoreApplication`, as advised in `QSettings` docs, I was also affecting other `QStandardPaths`. But simply leaving them out would result in an "Unknown Organization" folder for the settings. 

Although the organization name can be passed directly to `QSettings`' ctor (rather than determined centrally) on mac the domain would be used instead, and `QSettings` does not provide a ctor for that. I ended up deriving a custom filename altogether, based on `QStandardPaths::AppConfigLocation`, the application name, and the suffix `.conf` (which `QSettings` used by default). 

Note that this code will still have to change with #776 (the client will need access to both its and the daemon's settings file) and possibly for yaml backend.

townsend2010:
@ricab,

Cool, thanks.  There is also a merge conflict happening now too.

ricab:
@townsend2010 rebased and force-pushed

ricab:
@gerboland fair questions, I started typing answers a couple of times but I am not inspired ATM, I'll have to reply later.

Meanwhile I pushed fixes for compilation in buggy compilers ;)

ricab:
@gerboland sorry for the long reply. I hope it is clear, but let me know if you would like to chat about this or if there are particular places in the code that I could explain better.

What brought me to this approach was really the combination of:

- having something globally accessible (to avoid param injection)
- being able to mock it

As you suggest, let's do away with option 1 (direct `QSettings`). It would not be mockable and it would also mean implementing extra logic (for checking, separate files (coming for the daemon), custom format (need to move to YAML)) in calling sites.

So we needed a wrapper. Your option 2 (regular interface + inheriting implementation and mock classes), was my departure point. It would normally mean passing `SettingsInterface` objects as a param whose ultimate type was determined elsewhere. As the config grew, multiple entities would need to access it, so the param could start polluting many signatures. Repeating the dependency would *add entropy, but not much information*, because we could approximate it by just saying "everyone depends on it" in a single place. I *tried* to explain this in the [doc I wrote](https://docs.google.com/document/d/1pyQsMTPnR-iDqKeffo0bS7rSq-BwbZ0tbxdfx_l89QU/edit).

OK, so what if we just dropped the param but created a new object whenever needed? Well, we could not use the ctor directly if we wanted to mock it, because which type would we use? We would need some sort of factory. This factory would be an interface with an implementation for tests and one for *real code*. But which one do we call? Well, we'd need to *pass* one factory implementation in normal execution and another in tests. But for that we'd need to inject a dependency everywhere again. Basically we'd be back to the same problem as before.

What we needed was a way to say "here is the settings object, everyone can use it", in a single place. We'd determine the exact object (and thus its type) only once.

The Settings singleton does that. By default it provides a "real" Settings object: one of its own implementation, who does "the real job" we need. But it allows registering a derivation that overrides things for mocking.

Note that *this is totally dependent on the dynamic polymorphism* (the usual sort). The `Settings` class declares `get`/`set` as virtual and they are overridden by the mock. Relatively to what you propose, the only difference is that I skipped the pure interface part. That's perhaps not as clean looking, but it avoided another inheritance level and provided a ready default implementation that the mock could delegate to ([which it does](https://github.com/CanonicalLtd/multipass/pull/805/files#diff-bdfc5d8bf04bd0bbd0ba1f8aa5371998R39)).

The rest is the implementation of the `Singleton`. That is where the static polymorphism comes into play, and only so we can have a `Singleton` base class that ensures it really has a single object. (Note: this restriction is actually a bit relaxed, to "only a single object at any point in time". There can be multiple objects over the course of a program, because I provide a protected `reset`, but that is really only there for mocking.)

To ensure that only one object is created, the `Singleton` needs to control ctor access. If its ctor was private, descendants could not be constructed (because they would not be able to call the parent's ctor). If it was protected, descendants would be able to construct as many as they liked, so they wouldn't be singletons.

So we use a different mechanism to restrict ctor calling that does not rely on visibilities at all. We actually declare it public, but give it a token parameter that callers need to be able to produce ([here](https://github.com/CanonicalLtd/multipass/pull/805/files#diff-94f1189f2cf6ebf978226433286ebc6bR38)). Descendants then require this same token in their ctor and pass it to the parent ctor. We keep a single such token private in the `Singleton` and we provide it only in calls made by the `Singleton` itself. In other words, the constructors of `Singleton` derivations are only called by the `Singleton` (and intermediate des/ascendants).

Since the Singleton is [the one creating the object](https://github.com/CanonicalLtd/multipass/pull/805/files#diff-94f1189f2cf6ebf978226433286ebc6bR99), it needs to know its derived type. It gets that in the template parameter. But because we also allow mocking, we allow an additional template for the mock. That is the bit that introduces more template complexity.

The final bit is the `PrivatePassProvider`. This is the root of the singleton inheritance chain. It is just the class that provides the private token discussed above. We could pretty much paste its contents in the singleton directly, but that would make it even harder to read. Besides, having a separate class honors separation of concerns and modularity, so the `PrivatePassProvider` can be used for other purposes &ndash; I actually had it in my code for totally different purposes; it can be useful in general whenever we want fine control of accessibility that goes beyond what normal private/protected/public visibilities provide.

I suspect part of the "boiler-plate" you mention is due to separation of declaration from definition (even if the implementation is in the header). This is arguably less important today, with `=default`, `constexpr`, and default initializers, going the other way. But I think it still makes for a much clearer view. You look at the class and get an immediate overview of the interface without having to care about implementation details. If you want those, they are below, (but they do have to repeat templates and namespaces).

As for testability, it can be tested by attempting to compile something and see if it breaks. I provided examples in `test_singleton.cpp` and `test_private_pass_provider.cpp`. It is automated testing that is more difficult. I did that [here](https://github.com/ricab/scope_guard/blob/master/compile_time_tests.cpp) with ifdefs and [cmake](https://github.com/ricab/scope_guard/blob/760de0a9ea0fec980a5244604ffa5b1acff2db7a/CMakeLists.txt#L126). It seemed like too much here, but I can do it if requested. There is also [DejaGNU](https://www.gnu.org/software/dejagnu/) but I am not familiar with it.

An alternative I thought of was a free factory function that would be "premocked" or have a header injected. Both sounded hackish and I was unsure if they were doable (I am not familiar with premock).

Another alternative that occurs to me now would be to provide copies or a static "template" object, rather than a reference to it as in a singleton. We would have to return `unique_ptrs` or something, to avoid chopping. I did not think of that thought. Nevertheless most of the complexity (for registration/resetting/mocking) would still be there. And having a single object makes it easier to reason about, either if it is a simple proxy to what could otherwise be global functions or if it gets state in the future.

ricab:
Note to self: merge master and check if/where the new QString/fmt util can be used.

PR #810, opened 2019-05-29
daemon: Lower logging level of unknown hash exception

### LINKED ISSUE #563: multipass complains about missing image data when launching an image
```
$ multipass launch http://cloud.debian.org/images/cloud/OpenStack/current/debian-9.6.3-20190115-openstack-amd64.qcow2                                                                                            
Waiting for initialization to complete -[2019-01-17T13:11:50.943] [error] [daemon] Error fetching image information: Unable to find an image matching hash "ea609ceb1c6d90bc269713134e78312cd67afedc3512f0919a96c173f15814ed"                

Launched: elemental-rockhopper
$ 
```

It should be a warning instead.

townsend2010:
Fixes #563

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -992,7 +992,7 @@ try // clang-format on
             }
             catch (const std::exception& e)
             {
-                mpl::log(mpl::Level::info, category, fmt::format("Cannot fetch image information: {}", e.what()));
+                mpl::log(mpl::Level::warning, category, fmt::format("Cannot fetch image information: {}", e.what()));
             }
         }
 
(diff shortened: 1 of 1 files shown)


Saviq:
Hmm is this the right approach? Can't we determine whether we're supposed to see that data, and complain if we don't?

townsend2010:
@Saviq,

I totally agree with you.  However, I looked into doing this properly and it won't be trivial by any stretch.  

We made many assumptions originally about images, ie, information about them will come from SimpleStreams or some other remote type service.  Obviously, http type downloads have no information at all about the image.  So in order to handle that case, we'd have to rework all of our assumptions built into this.  That said, I thought it would be quick to at least suppress the messages to not be errors, but instead just be informational and then go back and rework this correctly.

Saviq:
Ack, shouldn't we do warning, though? Those don't get propagated to the client by default?

townsend2010:
I originally had `warning`, but then changed my mind since these are the only two places that this is called, but really, it doesn't matter to me, so I'll change it back to `warning`.

PR #812, opened 2019-05-30
launch: Clean up instance data if an error occurs during image prep

### LINKED ISSUE #680: Daemon leaks instances when launching with low `--disk`
The `--disk` option does not work with a size below the *virtual size* of the relevant image. This makes sense (to avoid data loss). However, it is currently only enforced by the underlying `qemu-img` command. 

`multipassd` itself does not perform this check and leaves files behind when `qemu-img` fails. The most significant of which is the copy of the image file.

This means that a few hundreds of MB of disk are leaked on such occasions. The regular user does not realize this and leaves the files behind.

```sh
$ multipass launch --disk 1G
launch failed: Cannot resize instance image

$ multipass list
No instances found.

$ du -sh ~/.local/share/multipassd/vault/instances/deep-midge/*
20K     /home/ricab/.local/share/multipassd/vault/instances/deep-midge/cloud-init-config.iso
328M    /home/ricab/.local/share/multipassd/vault/instances/deep-midge/ubuntu-18.04-server-cloudimg-amd64.img

$ qemu-img info /home/ricab/.local/share/multipassd/vault/instances/deep-midge/ubuntu-18.04-server-cloudimg-amd64.img | grep size
virtual size: 2.2G (2361393152 bytes)
disk size: 328M
cluster_size: 65536
```

The daemon logs

> [warning] [qemu-img] qemu-img: warning: Shrinking an image will delete all data beyond the shrunken image's end. Before performing such an operation, make sure there is no important data there.
qemu-img: Use the --shrink option to perform a shrink operation.

townsend2010:
Fixes #680

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1927,6 +1927,9 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
             catch (const std::exception& e)
             {
                 preparing_instances.erase(name);
+                release_resources(name);
+                vm_instances.erase(name);
+                persist_instances();
                 status_promise->set_value(grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), ""));
             }
 
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -50,6 +50,7 @@
 #include <memory>
 #include <ostream>
 #include <sstream>
+#include <stdexcept>
 #include <string>
 
 namespace mp = multipass;
@@ -127,7 +128,7 @@ public:
             create_error.ParseFromString(status.error_details());
             const auto errors = create_error.error_codes();
 
-            cerr << "fail: ";
+            cerr << "failed: " << status.error_message();
             if (errors.size() == 1)
             {
                 const auto& error = errors[0];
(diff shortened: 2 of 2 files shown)


townsend2010:
@ricab,

Yeah, I'd really like to add a test for this, but as you know, we have absolutely no infrastructure in place to test the daemon.  In order to test this particular case, we'd need to mock quite a bit of things.  And also, the code in question is the result of a concurrent operation, so we'd also have to account for that as well.

I think we really need to spend some time on coming up with a way to test the daemon better and once we do that, we can then begin to break down areas of unit testing the daemon.

ricab:
@townsend2010 Perhaps we're not talking about the same sort of testing, but all the mocking needed for what I meant is in place. And no need to deal with concurrency, just follow the same pattern from existing tests... I just pushed what I meant and will do a PR into this PR.

ricab:
https://github.com/CanonicalLtd/multipass/pull/813

townsend2010:
@ricab,

Ah, ok, I see.  So you're just basically overriding the `prepare_instance_image` call to throw the exception and then testing to make sure `remove_resources_for` is called in the mock factory and that the exception string matches what is expected.

I'll just cherry-pick your PR commit into this PR and we can close your PR.

ricab:
@townsend2010 yes exactly. Really only to make sure the resources are cleaned up. Feel free to tweak or improve. The test name in in particular may be misleading.

townsend2010:
@ricab,

I think the test is fine as-is.

ricab:
Right, borsing...

bors r=townsend2010,ricab

PR #813, opened 2019-05-31
[test] Add test for instance-image-leak fix

ricab:
Adds the test for the leak fix. It fails before the fix and passes afterward.


townsend2010:
Thanks for this.  I just cherry-picked this into the other PR, so I'll go ahead and close this.

PR #814, opened 2019-05-31
Teach fmtlib to deal with QString, can remove lots of toStdString() calls

gerboland:
RFC. Do we want?


townsend2010:
I'm definitely a fan of trying to commonize between `QString` and `std::string` as much as we can:)

PR #815, opened 2019-05-31
Remove unused passing around of key_provider

gerboland:
I do not see it being used in any VM


townsend2010:
Fails on Mac build, so manually merging.

PR #827, opened 2019-06-07
Create VMInstance class

gerboland:
I want to share my thinking into a refactor.

We currently keep a lot of state and operations code related to VMs in the Daemon. For instance state, i.e. preparing a VM, VM being deleted, VM shutdown delay, we track that state by moving the instance name between different sets/maps. Operations code are mostly platform-independent VM-control methods (start/shutdown/reboot/install_sshfs/mounts).

I'm in a situation where I want to add another per-VM operation (a maintenance operation which will fire on a timer while VM running). It is a platform independent op, so doesn't belong in the VirtualMachine implementations themselves. Right now, platform independent VM ops are private methods in Daemon itself. So adding this new op means I'll have to add another list to Daemon, and carefully add/remove from the list in sync with the VM state - clunky.

My thinking is to introduce a "VMInstance" class, which will hold per-VM state and platform independent operations (i.e. start/shutdown/reboot/install_sshfs/mounts). 

I hope to replace the state sets/maps in Daemon with VMInstance::state() which for example would return Prepared/Exists/Deleted/ - I've to decide out of the actual VM state should be included too.

And I also want to move the ops like start/shutdown/reboot/install_sshfs/mounts our of Daemon and create as members of this.

This PR is the first step on this journey, and make it easier to add my per-VM operation.

What do you folks think? Right direction or no?

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -25,7 +25,8 @@ add_library(daemon STATIC
   daemon_rpc.cpp
   default_vm_image_vault.cpp
   json_writer.cpp
-  ubuntu_image_host.cpp)
+  ubuntu_image_host.cpp
+  vm_instance.cpp)
 
 add_library(delayed_shutdown STATIC
   delayed_shutdown_timer.cpp
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -585,7 +585,8 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         try
         {
             auto& instance_record = spec.deleted ? deleted_instances : vm_instances;
-            instance_record[name] = config->factory->create_virtual_machine(vm_desc, *this);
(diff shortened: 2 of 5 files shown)


townsend2010:
I think this is a really good idea.  The growing list of `unordered_maps` is getting a bit large along with all of the things the daemon itself is responsible for regarding instances.  The daemon code itself is simply getting too large and starting to refactor it in a way like this where another object is charge of the instance is a good thing.

I will look over the code some more, but on the first pass, it looks quite reasonable.

gerboland:
To give you idea of where this brings us, this is how VMInstances looks when I move more related bits out of Daemon:
https://pastebin.ubuntu.com/p/JBvBK5Nkxq/

townsend2010:
That should also make this much easier to test too as we'd only need to worry about this class and no need to mock so many other parts of the daemon.

ricab:
Yes, this sounds like a good idea to me too. Thanks!

PR #830, opened 2019-06-10
url_downloader: Remove QNetworkDiskCache caching

### LINKED ISSUE #825: Image hashes may not match when a new cloud-image is uploaded
We ran into a situation where new minimal cloud images were uploaded, but we were getting image hash mismatch errors.

Turns out that the SHA256SUM file is cached by Qt and it was using an older cached version of that file instead or retrieving the new one. 

We need to force Qt to always hit the actual URL when retrieving the hash file.

townsend2010:
The Qt network caching doesn't work for our needs since it doesn't necessarily
get the latest file from the server and using "AlwaysNetwork" defeats the purpose
of caching.

In the future, custom caching will be introduced that fits Multipass's needs.

Fixes #825

### DIFF
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -26,7 +26,6 @@
 #include <QEventLoop>
 #include <QFile>
 #include <QNetworkAccessManager>
-#include <QNetworkDiskCache>
 #include <QNetworkReply>
 #include <QTimer>
 #include <QUrl>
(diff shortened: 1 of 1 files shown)


townsend2010:
Hmm, there may be a better way to handle this and still use `QNetworkDiskCache`.  Instead of having `QNetworkAccessManager` own the cache, the `URLDownloader` class can instead own it and use it how it sees fit.

I think I will work on that instead of just whacking the whole thing here.

PR #834, opened 2019-06-11
url_downloader: Directly use network cache when downloading data

### LINKED ISSUE #825: Image hashes may not match when a new cloud-image is uploaded

### LINKED ISSUE #829: `multipass launch…` should work offline for all cached images
Today when the daemon is started with no network, the daemon just decides there's no images:

```
$ multipass find
No images found.
$ multipass launch 
launch failed: Remote "" is unknown or unreachable. 
```

We should cache at least enough of the image metadata along with the images to make this work.

townsend2010:
Due to how QNetwork handles caching via QNetworkAccessManager and issues it
causes, instead directly access the network cache. This includes comparing the
Last-Modified header of the URL to what is cached and also use cached data (if
availble) when there is an issue downloading the data from the URL.

Fixes #825, fixes #829

### DIFF
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -163,22 +163,18 @@ QByteArray mp::URLDownloader::download(const QUrl& url)
 
     if (metadata.isValid())
     {
-        QDateTime modified;
-
         try
         {
-            modified = last_modified(url);
+            if (last_modified(url) == metadata.lastModified())
+            {
+                return get_network_cache_data(network_cache, url);
+            }
         }
         catch (const std::exception& e)
         {
-            // Force using the cached data if the URL is unreachable
-            modified = metadata.lastModified();
-            mpl::log(mpl::Level::info, category,
-                     fmt::format("Cannot get last modified date for {}: {}", url.toString(), e.what()));
-        }
-
-        if (modified == metadata.lastModified())
-        {
+            mpl::log(
+                mpl::Level::info, category,
+                fmt::format("Cannot get last modified date for {}: {}. Using cached data.", url.toString(), e.what()));
             return get_network_cache_data(network_cache, url);
         }
     }
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -56,6 +56,14 @@ auto make_network_manager(const mp::Path& cache_dir_path)
     return manager;
 }
 
+auto get_network_cache_data(QAbstractNetworkCache* network_cache, const QUrl& url)
+{
+    auto contents = network_cache->data(url);
+    auto data = contents->readAll();
+    contents->deleteLater();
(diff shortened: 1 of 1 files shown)


Saviq:
Hey, could we have tests for this?

townsend2010:
@Saviq,

I thought about it and we would need to mock things like `QNetworkDiskManager` and be able to populate it with cache data in a way that works for `QAbstractNetworkCache`.  I didn't see any straightforward way to do that, although thinking about it as I write this, I wonder if it's possible to do something with our premock stuff we use for testing.  I'll look into that before declaring that this would take too much time for a small gain.

townsend2010:
Ok doesn't look like premock will help us on this.  So basically, what would need to be done is pre-populate a cache that is understood by `QNetworkDiskCache`.  This would be fragile since in the Qt code, they use internal versioning of the cache and use obscure file names along with internal metadata in the cache files we'd have to get correct.

I can add this basing this on the real cache on my system, but it could be prone to breakage down the line if Qt decides to change how it handles the cache internally, including path versioning that it uses and special metadata in the cache file.  I'm just not sure it's worth it.

townsend2010:
After some more looking at this, I really don't think it's worth the effort to put in the static caching infrastructure (which would potentially be fragile to Qt changes) in place for testing this small chunk of code.

src/network/url_downloader.cpp
```diff
@@ -149,11 +158,53 @@ QByteArray mp::URLDownloader::download(const QUrl& url)
...
+        QDateTime modified;
+
+        try
+        {
+            modified = last_modified(url);
+        }
+        catch (const std::exception& e)
+        {
+            // Force using the cached data if the URL is unreachable
+            modified = metadata.lastModified();
```
gerboland: Could you not just do

    return get_network_cache_data(network_cache, url);

from here instead of setting `modified` to a value that makes the later if() true, that calls the above anyway?

townsend2010: @gerboland,

Thanks for the review.  It is difficult to test and that is how I tested it too.

Sure, I can change that and it would make it more clear.

townsend2010: Ok, i tightened up this code.  Thanks!

townsend2010:
Shall we merge?

gerboland:
Nobody else has opinion then? Ok, then
bors r+

Saviq:
Well, I *do* have an opinion.

~~Everything~~This is awesome!

Issue #756, opened 2019-04-30
RFC: `multipass config/get/set` structure and user experience


Saviq:
#307 has some details on specifics of the "custom repository" feature, but I wanted to collect the user experience and the imaginable extent of the configuration structure here.

### The CLI
Three new commands would be introduced, all operating on YAML-formatted data, with `.` (periods) separating depth levels. All operations need to be atomic and return validation errors.
```
multipass config [--no-defaults] [--expand] [<subtree>]
# opens an editor in interactive mode
# or prints the YAML of the subtree to standard out in non-interactive mode
# if `--no-defaults` given, only explicitly configured values will be shown
# some parts of the tree may not be shown by default (think remote and instance configuration), passing `--expand` would mean that the whole tree is shown
# on errors returns to the editor with annotation about validation or configuration issues

multipass get <key>
# prints the value of the requested configuration key

multipass set [<key>=<value> …]
# sets the values of the given configuration keys
# or accepts a YAML of a subset of the configuration tree on standard input
```

### The structure
To avoid unnecessary nesting, I propose top-level keys to only be: `client` and remote names, with remote configuration nested under the remote name.

```yaml
client:
  default-remote: local       # name of one of the remotes configured below
  primary-name: primary       # name of the primary instance

  launch-defaults:
    image: default            # see `multipass find` for available images or use file:// or http:// URLs
    cpus: 1                   # recommended at most one below your host's cores
    disk: 5GB                 # this is the maximum the instance can use, so go big
    memory: 2GB               # maximum the instance can use, shared between host and instances
    cloud-init: {}            # see http://cloudinit.readthedocs.io/en/latest/topics/examples.html
    mounts:
      /local/path: remote/path
      /other/path:
        target: other/remote/path
        uid_maps:
          *: default          # "*" for "all", "default" for the default user's UID inside the instance,…
        gid_maps:
          *: default          # …or the numeric ID

  images:
    example:                  # this is the image name to use for `multipass launch`
      aliases: [ex]           # an optional list of alternative names
      image: default          # defaults to this image's key above
      cpus: 2                 # overrides the default launch options from above
 
    lts:                      # because the `lts` alias exists on the remote…
      disk: 25GB              # …this only overrides the disk size when doing `multipass launch lts`

local:                        # configured on first start
  address: unix:/run/multipass_socket  # platform default
  
  # below is remote configuration
  listen-address: unix:/run/multipass_socket
[...]

PR #835, opened 2019-06-11
Settings daemon driver - take 2

### LINKED ISSUE #776: Allow configuring the daemon driver

ricab:
Fixes #776.

New attempt, using a central file for daemon settings and requiring admin privileges to set.

### DIFF
--- a/include/multipass/exceptions/settings_exceptions.h
+++ b/include/multipass/exceptions/settings_exceptions.h
@@ -38,8 +38,8 @@ public:
 class PersistentSettingsException : public SettingsException
 {
 public:
-    PersistentSettingsException(const QString& attempted_operation, const QString& failure)
-        : SettingsException{fmt::format("Could not {} settings: {} error", attempted_operation, failure)}
+    PersistentSettingsException(const QString& attempted_operation, const QString& detail)
+        : SettingsException{fmt::format("Unable to {} settings: {}", attempted_operation, detail)}
     {
     }
 };
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -26,6 +26,8 @@
 
 #include <libssh/sftp.h>
 
+#include <QString>
+
 #include <string>
 
 namespace multipass
@@ -33,6 +35,7 @@ namespace multipass
 namespace platform
 {
 std::string default_server_address();
+QString default_driver();
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -36,6 +36,7 @@ namespace platform
 {
 std::string default_server_address();
 QString default_driver();
+QString daemon_config_home(); // temporary
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -37,6 +37,7 @@ namespace platform
 std::string default_server_address();
 QString default_driver();
 QString daemon_config_home(); // temporary
+bool is_backend_supported(const QString& backend); // temporary
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -35,7 +35,7 @@ apps:
       XDG_DATA_HOME: $SNAP_USER_DATA/data
       XDG_CACHE_HOME: $SNAP_USER_DATA/cache
       XDG_CONFIG_HOME: $SNAP_USER_DATA/config
-    command: bin/multipass
+    command: bin/launch-multipass
     completer: etc/bash_completion.d/snap.multipass
   gui:
     environment: *client-environment
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,25 +17,28 @@ apps:
   multipassd:
     command: bin/launch-multipassd
     environment:
-      LD_LIBRARY_PATH: $SNAP/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET
-      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
+      LD_LIBRARY_PATH: &library-path |
+        $SNAP/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET
+      PATH: &path |
+        $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
       QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
       XDG_DATA_HOME: $SNAP_COMMON/data
       XDG_CACHE_HOME: $SNAP_COMMON/cache
+      XDG_CONFIG_HOME: $SNAP_DATA/config
     daemon: simple
     plugs: [libvirt]
   multipass:
-    environment:
-      LD_LIBRARY_PATH: $SNAP/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET
-      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
+    environment: &client-environment
+      LD_LIBRARY_PATH: *library-path
+      PATH: *path
       QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
+      XDG_DATA_HOME: $SNAP_USER_DATA/data
+      XDG_CACHE_HOME: $SNAP_USER_DATA/cache
+      XDG_CONFIG_HOME: $SNAP_USER_DATA/config
     command: bin/multipass
     completer: etc/bash_completion.d/snap.multipass
   gui:
-    environment:
-      LD_LIBRARY_PATH: $SNAP/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET
-      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
-      QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
+    environment: *client-environment
     command: bin/multipass-gui
 
 parts:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,6 +25,7 @@ apps:
       XDG_DATA_HOME: $SNAP_COMMON/data
       XDG_CACHE_HOME: $SNAP_COMMON/cache
       XDG_CONFIG_HOME: &daemon-config $SNAP_DATA/config
+      DAEMON_CONFIG_HOME: *daemon-config # temporary
     daemon: simple
     plugs: [libvirt]
   multipass:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -24,17 +24,19 @@ apps:
       QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
       XDG_DATA_HOME: $SNAP_COMMON/data
       XDG_CACHE_HOME: $SNAP_COMMON/cache
-      XDG_CONFIG_HOME: $SNAP_DATA/config
+      XDG_CONFIG_HOME: &daemon-config $SNAP_DATA/config
     daemon: simple
     plugs: [libvirt]
   multipass:
-    environment: &client-environment
-      LD_LIBRARY_PATH: *library-path
-      PATH: *path
-      QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
-      XDG_DATA_HOME: $SNAP_USER_DATA/data
-      XDG_CACHE_HOME: $SNAP_USER_DATA/cache
-      XDG_CONFIG_HOME: $SNAP_USER_DATA/config
+    environment:
+      <<: &client-environment
+        LD_LIBRARY_PATH: *library-path
+        PATH: *path
+        QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
+        XDG_DATA_HOME: $SNAP_USER_DATA/data
+        XDG_CACHE_HOME: $SNAP_USER_DATA/cache
+        XDG_CONFIG_HOME: $SNAP_USER_DATA/config
+      DAEMON_CONFIG_HOME: *daemon-config # temporary
     command: bin/launch-multipass
     completer: etc/bash_completion.d/snap.multipass
   gui:
--- a/src/client/cli/cmd/set.cpp
+++ b/src/client/cli/cmd/set.cpp
@@ -72,15 +72,15 @@ mp::ParseCode cmd::Set::parse_args(mp::ArgParser* parser)
     if (status == ParseCode::Ok)
     {
         const auto args = parser->positionalArguments();
-        if (args.isEmpty())
+        if (args.size() != 1)
         {
-            cerr << "Insufficient arguments.\n";
+            cerr << "Need exactly one key-value pair.\n";
             status = ParseCode::CommandLineError;
         }
         else
         {
             const auto keyval = args.at(0).split('=');
-            if (keyval.count() != 2 || keyval.contains(QStringLiteral("")))
+            if (keyval.size() != 2 || keyval.contains(QStringLiteral("")))
             {
                 cerr << "Bad key-value format.\n";
                 status = ParseCode::CommandLineError;
--- a/src/client/cli/cmd/set.cpp
+++ b/src/client/cli/cmd/set.cpp
@@ -63,22 +63,33 @@ QString cmd::Set::description() const
 
 mp::ParseCode cmd::Set::parse_args(mp::ArgParser* parser)
 {
-    parser->addPositionalArgument("key", "Path to the option to configure.", "<key>");
-    parser->addPositionalArgument("value", "Value to set the option to.", "<value>");
+    parser->addPositionalArgument(
+        "keyval",
+        "A key-value pair. The key specifies a path to the option to configure. The value is its intended value.",
+        "<key>=<value>");
 
     auto status = parser->commandParse(this);
     if (status == ParseCode::Ok)
     {
         const auto args = parser->positionalArguments();
-        if (args.count() == 2)
+        if (args.isEmpty())
         {
-            key = args.at(0);
-            val = args.at(1);
+            cerr << "Insufficient arguments.\n";
+            status = ParseCode::CommandLineError;
         }
         else
         {
-            cerr << "Need exactly one option key and one value\n";
-            status = ParseCode::CommandLineError;
+            const auto keyval = args.at(0).split('=');
+            if (keyval.count() != 2 || keyval.contains(QStringLiteral("")))
+            {
+                cerr << "Bad key-value format.\n";
+                status = ParseCode::CommandLineError;
+            }
+            else
+            {
+                key = keyval.at(0);
+                val = keyval.at(1);
+            }
         }
     }
 
--- a/src/daemon/daemon_main.cpp
+++ b/src/daemon/daemon_main.cpp
@@ -79,6 +79,12 @@ void set_server_permissions(const std::string& server_address)
         throw std::runtime_error("Could not set permissions for the multipass socket.");
 }
 
+void init_settings(const QString& filename)
+{
+    QFile file{filename};
+    file.open(QIODevice::WriteOnly | QIODevice::Append);
+}
+
 class UnixSignalHandler
 {
 public:
@@ -108,8 +114,10 @@ private:
 
 void monitor_and_quit_on_settings_change()
 {
-    // TODO @ricab touch the file (QFileWatcher can't watch if not there)
-    static QFileSystemWatcher monitor{{mp::Settings::get_daemon_settings_file_path()}};
+    static const auto filename = mp::Settings::get_daemon_settings_file_path();
+    init_settings(filename); // create if not there
+
+    static QFileSystemWatcher monitor{{filename}};
     QObject::connect(&monitor, &QFileSystemWatcher::fileChanged, QCoreApplication::instance(), QCoreApplication::quit);
 }
 
--- a/src/daemon/daemon_main.cpp
+++ b/src/daemon/daemon_main.cpp
@@ -26,6 +26,7 @@
 #include <multipass/name_generator.h>
 #include <multipass/platform.h>
 #include <multipass/platform_unix.h>
+#include <multipass/settings.h>
 #include <multipass/utils.h>
 #include <multipass/version.h>
 #include <multipass/virtual_machine_factory.h>
@@ -35,6 +36,8 @@
 #include <multipass/format.h>
 
 #include <QCoreApplication>
+#include <QFileSystemWatcher>
+#include <QObject>
 
 #include <csignal>
 #include <cstring>
@@ -102,6 +105,14 @@ public:
 private:
     mp::AutoJoinThread signal_handling_thread;
 };
+
+void monitor_and_quit_on_settings_change()
+{
+    // TODO @ricab touch the file (QFileWatcher can't watch if not there)
+    static QFileSystemWatcher monitor{{mp::Settings::get_daemon_settings_file_path()}};
+    QObject::connect(&monitor, &QFileSystemWatcher::fileChanged, QCoreApplication::instance(), QCoreApplication::quit);
+}
+
 } // namespace
 
 int main(int argc, char* argv[]) // clang-format off
@@ -117,6 +128,7 @@ try // clang-format on
     auto config = builder.build();
     auto server_address = config->server_address;
 
+    monitor_and_quit_on_settings_change();
     mp::Daemon daemon(std::move(config));
 
     set_server_permissions(server_address);
--- a/src/daemon/daemon_main.cpp
+++ b/src/daemon/daemon_main.cpp
@@ -51,6 +51,7 @@ namespace mpp = multipass::platform;
 
 namespace
 {
+constexpr const int settings_changed_code = 42;
 const std::vector<std::string> supported_socket_groups{"sudo", "adm", "admin"};
 
 void set_server_permissions(const std::string& server_address)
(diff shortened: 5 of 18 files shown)


src/platform/platform_linux.cpp
```diff
@@ -49,16 +49,26 @@ std::string mp::platform::default_server_address()
     return {"unix:/run/multipass_socket"};
 }
 
-mp::VirtualMachineFactory::UPtr mp::platform::vm_backend(const mp::Path& data_dir)
+QString mp::platform::default_driver()
 {
-    auto driver = qgetenv("MULTIPASS_VM_DRIVER");
+    return QStringLiteral("qemu");
```
Saviq: We should think about transitioning from the environment variable to this setting for those that have previously used it.

This should probably be done in:
1) check if `MULTIPASS_VM_DRIVER` is set and use it as the initial value for the setting (for platforms other than snap)
2) `on-refresh` snap hook should `snapctl get driver`, `multipass set…` it and reset it (set to `~`, which is YAML's `null`?)
3) `configure` hook should fail and tell the user to use `multipass set…` for this key (I don't think supporting `snap set` along `multipass set` will do us good in the long run)

It doesn't all have to be this PR, as long as we have an upgrade story for both `edge` now and `beta` users when we release.

src/utils/settings.cpp
```diff
@@ -32,31 +34,56 @@ namespace mp = multipass;
...
+ *   - /root/.config/multipass/multipassd.conf
+ */
+QString file_for(const QString& key) // the key should have passed checks at this point
 {
-    return {{mp::petenv_key, petenv_name}};
+    static const auto file_path_base = QStringLiteral("%1/%2.%3"); // static consts ensure these stay fixed
+    static const auto client_dir_path = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
+    static const auto daemon_dir_path = QStringLiteral("/etc/xdg"); // temporary
+    static const auto client_file_path = file_path_base.arg(client_dir_path, mp::client_name, file_extension);
+    static const auto daemon_file_path = file_path_base.arg(daemon_dir_path, mp::daemon_name, file_extension);
```
Saviq: I know you just picked the code from below, but maybe use `QDir` for those? We do have a `multipass::Path`, too, but it's a plain `QString` for some reason…

src/utils/settings.cpp
```diff
@@ -32,31 +34,56 @@ namespace mp = multipass;
...
-
-    return std::make_unique<QSettings>(file_path, format); /* unique_ptr to circumvent absent copy-ctor and no RVO
-                                                              guarantee (until C++17) */
+    static const auto format = QSettings::defaultFormat(); // static const to make sure these stay fixed
+
+    return std::make_unique<QSettings>(file_for(key), format); /* unique_ptr to circumvent absent copy-ctor and no RVO
+                                                                  guarantee (until C++17) */
+}
+
+bool is_unreadable(const QString& filename)
```
Saviq: I'd probably make this `is_readable()` and move the negation to the consumer.

src/platform/platform_linux.cpp
```diff
@@ -49,16 +49,26 @@ std::string mp::platform::default_server_address()
...
+    else if (driver == QStringLiteral("libvirt"))
         return std::make_unique<LibVirtVirtualMachineFactory>(::process_factory(), data_dir);
+    else
+    {
+        std::string msg_tail{};
+        if (driver == QStringLiteral("hyperkit") || driver == QStringLiteral("hyper-v"))
+            msg_tail = " on this platform";
 
-    throw std::runtime_error("Invalid virtualization driver set in the environment");
+        throw std::runtime_error(fmt::format("Unsupported virtualization driver{}: {}", msg_tail, driver));
```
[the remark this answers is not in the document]
ricab: Are you referring to the leading space on `msg_tail`? Avoiding that split requires duplication. Would this satisfy you?

```c++
std::string msg;
if (driver == QStringLiteral("hyperkit") || driver == QStringLiteral("hyper-v"))
    msg = "Unsupported virtualization driver (on this platform): {}";
else
    msg = "Unsupported virtualization driver: {}";
throw std::runtime_error(fmt::format(msg), driver);
```
 This at least avoids duplicating the throw and format (one less place to change if we need a more specific exception in the future). I could also remove the distinction and just `throw runtime_error{fmt::format("Unsupported virtualization driver: {}", driver)};`

src/platform/platform_linux.cpp
ricab: OK. I'd create a separate issue for this, just for the sake of my organization. Are epochs already available? Otherwise, would we leave that transition code indefinitely?

src/utils/settings.cpp
ricab: Yeah I understand your point. `is_readable` would be more elementary and so more general. But here `is_unreadable` means the file *exists and cannot be read*. I think this makes sense in English, as you would not say a non-existing file is unreadable. But, if I negated that, *I'd be saying non-existing files were readable*. While that can be argued to be true (you know exactly what the contents of an absent file are, and you can "read" it's absence provided you have read permission to the parent), it sounds less intuitive to me. But ok, if you insist I don't mind doing that. Give me a verdict.

src/utils/settings.cpp
ricab: Sorry I don't quite follow, where do you mean? The `*_dir_path` vars? Because those are consumed as strings. The `*_file_path` vars OTH are files, not dirs.

src/platform/platform_linux.cpp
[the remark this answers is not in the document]
Saviq: Yeah what you propose is fine, it's just string concatenation is evil when it comes to translations.

src/utils/settings.cpp
Saviq: It's probably just my bad experiences from elsewhere, I just cringe on using string formatting to create paths… I'm especially thinking path separators. Even if Qt can deal with `/` on Windows, if we end up passing those strings to something other than Qt, we'll see failures there.

So I would rather do something like:
```cpp
    const auto file_pattern = QStringLiteral("%1.conf");
    …
    static const auto client_file_path = QDir(client_dir_path).absoluteFilePath(file_pattern.arg(mp::client_name));
    static const auto daemon_file_path = QDir(daemon_dir_path).absoluteFilePath(file_pattern.arg(mp::daemon_name));
```

Won't block on this, though.

src/utils/settings.cpp
```diff
@@ -32,31 +34,56 @@ namespace mp = multipass;
...
+ * Examples:
+ *   - ${HOME}/.config/multipass/multipass.conf
+ *   - /root/.config/multipass/multipassd.conf
+ */
+QString file_for(const QString& key) // the key should have passed checks at this point
 {
-    return {{mp::petenv_key, petenv_name}};
+    static const auto file_path_base = QStringLiteral("%1/%2.%3"); // static consts ensure these stay fixed
+    static const auto client_dir_path = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
+    static const auto daemon_dir_path = QStringLiteral("/etc/xdg"); // temporary
```
Saviq: Oh and this, won't work on Windows or macOS, or when we move to a strict snap. We might need to make this a platform thing?

src/utils/settings.cpp
Saviq: So `is_unreadable("non-existent-file");` is false… i.e. a missing file is "not unreadable"? I'm getting confused here ;)

If this is just about permissions, then I'd say we need a more descriptive name for the function, then, or even move the `.exists()` out of it.

src/platform/platform_linux.cpp
Saviq: Right, we do need to do *something* here though, or people using `libvirt` on edge will get moved back to `qemu`, because we're not looking at the environment variable any more.

I think as a minimum we'd need to respect the environment variable as long as it's set, and once we implement migration we can warn that it is deprecated.

As for epochs… even if they are available, that'd only work for snaps.

src/utils/settings.cpp
ricab: Right, I meant to do that but forgot. On the strict snap though, won't it see its own filesystem /etc/xdg?

src/utils/settings.cpp
ricab: I see, ok.

src/utils/settings.cpp
ricab: Yeah I thought it made sense to say that unreadable required reading to be locked, which required the file to be there. But ok, I'll rework this then.

ricab:
@Saviq I replied to your points, but github is not dealing well with the multiple reviews. You can see them in the files tab or in each original comment's thread in this page.

src/platform/platform_linux.cpp
Saviq: I *think* we should respect the environment variable if it's set, for now. Otherwise people using `snap set` will get ignored.

So, precedence:
1) `MULTIPASS_VM_DRIVER`
1) `multipass set local.driver`
1) platform default

src/utils/settings.cpp
Saviq: No, even strict snaps see *a* rootfs, in their case it's the base (i.e. `core16`) in our case. They still can only write things in any of `$SNAP{,_USER}_{DATA,COMMON}` by default.

[Here](https://docs.snapcraft.io/environment-variables) is a handful reference to the environment variables.

Saviq:
> @Saviq I replied to your points, but github is not dealing well with the multiple reviews. You can see them in the files tab or in each original comment's thread in this page.

Yeah I thought I was being smart by "reviewing" all the comments rather than commenting one-by-one. Seems Github outsmarted me by making a mess out of that.

ricab:
> Yeah I thought I was being smart by "reviewing" all the comments rather than commenting one-by-one. Seems Github outsmarted me by making a mess out of that.

That is what I did also. It usually works, but maybe not if both sides use that approach.

src/platform/platform_linux.cpp
ricab: You're right, no way to check if there was an explicit configuration with the current interface.

src/utils/settings.cpp
ricab: OK, so we need to move this and implement differently depending on the platform. But should I also use `SNAP_COMMON` on Linux if available? Or leave that to deal with separately in the confinement story?

BTW, the strict snap would still be able to see `/etc/xdg` but not write to it, is that it?

ricab: Or maybe SNAP_DATA, to allow changing it between versions. I suppose those contents won't be lost from one version to the next?

Saviq: I think yes, using `$SNAP_COMMON` if it's set is the right approach.

For things that people should generally not see/edit, the better default place on Linux is `/var/lib/`.

We shouldn't spend too much though on this though, as it's going to go away once we push the configs through the RPC.

Saviq: Ah and about the strict snap, by default it's not going to be able to read /etc/xdg I don't think. The policies are whitelists, so unless there's something that all snaps should have access to, or we connect to an interface, it only has access to its own bits and data (and a handful of things from the base, like libc and whatever's shipped there).

ricab:
@Saviq

I remembered why we discarded `/root` before: it is not generally readable and so would require `sudo` for `multipass get`. I therefore went back to the earlier agreement of `/var/lib`, which best matches the requirements I was given (as I understood them).

If you deem the `/root`-based path preferable &ndash; because it could be preserved when moving away from file-based communication and despite the added permission restriction &ndash; let me know, so I can change the current *speculative implementation*. In that event, please also clarify whether you prefer:
1. `/root/.local/share`, which corresponds to `AppDataLocation`, but is different from what is already used in the client
 2. `/root/.config`, which corresponds to `AppConfigLocation`, matching the client but having `/etc/xdg` as fallback path (althought I don't see when, if ever, that would be returned by Qt).

If you want to stick with previous decisions despite newly found implications, or if you simply do not have a preference and/or just want *me* to decide, please also let me know that. If it were up to me (and excluding new findings) I would probably go with 2 above and consider `sudo` a temporary consequence of the file-based comm choice.

Saviq:
I'm fine with `sudo` and 2. above - provided the file location will not change when we move to RPC.

ricab:
OK, I will make that change then.

ricab:
Unfortunately I again stumbled upon `QSettings`' inability to distinguish absent from unreadable files. I now realized my previous attempt to circumvent this (with `exists_but_unreadable`) was not working in all cases. In particular, if the file was in a directory that could itself not be read (e.g. `/root/.config`), `QFileInfo("/root/.config/multipassd/multipassd.conf").exists()` would return `false`, even if the file was there. So, with the new path, we would happily return the default value on a `multipass get` without `sudo`. I did not think this would make for an acceptable UX:

```
$ sudo multipass set local.driver libvirt
$ multipass get local.driver
qemu
```

After some investigation I ended up solving this by checking `errno` directly after trying to *manually* open the file. Permissions checks would suffer from the same issue as above. Checking for parent permissions would be doing the OS's job. `QFile.error` did not distinguish our cases either (it just returned `QFileDevice::OpenError`). `QFile::errorString` did, but string matching would be unreliable, as the actual returned string is not part of any interface. After digging through Qt's code, I found `errorString` was just relying on `errno`, so I used that directly. 

The remaining "dir case" is reported by `QSettings` itself, so we don't need to account for it explicitly.

src/platform/platform_linux.cpp
```diff
@@ -42,23 +44,52 @@ mp::ProcessFactory* process_factory()
...
+}
 
-    if (driver.isEmpty() || driver == "QEMU")
+QString mp::platform::daemon_config_home() // temporary
+{
+    auto ret = QString{qgetenv("DAEMON_CONFIG_HOME")};
+    if (ret.isEmpty())
+        ret = QStringLiteral("/root/.config");
+
+    ret += QStringLiteral("/%1").arg(mp::daemon_name);
```
Saviq: Similar to the paths below?

```suggestion
    ret = QDir{ret}.absoluteFilePath(mp::daemon_name);
```

src/platform/platform_linux.cpp
```diff
@@ -42,23 +44,52 @@ mp::ProcessFactory* process_factory()
...
-    else if (driver == "LIBVIRT")
+    else if (driver == QStringLiteral("libvirt"))
         return std::make_unique<LibVirtVirtualMachineFactory>(::process_factory(), data_dir);
+    else
+    {
+        std::string msg;
+        if (driver == QStringLiteral("hyperkit") || driver == QStringLiteral("hyper-v"))
+            msg = "Unsupported virtualization driver (on this platform): {}";
+        else
+            msg = "Unsupported virtualization driver: {}";
```
Saviq: This really is protecting against someone messing with the config file manually, isn't it? I'd just throw the same error regardless of platform.

Saviq:
This could use a nicer error message:
```
$ multipass set local.driver libvirt
Could not read/write settings: access error
```

snap/snapcraft.yaml
```diff
@@ -17,31 +17,36 @@ apps:
...
-      QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
-    command: bin/multipass
+      <<: &client-environment
+        LD_LIBRARY_PATH: *library-path
+        PATH: *path
+        QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
+        XDG_DATA_HOME: $SNAP_USER_DATA/data
+        XDG_CACHE_HOME: $SNAP_USER_DATA/cache
+        XDG_CONFIG_HOME: $SNAP_USER_DATA/config
+      DAEMON_CONFIG_HOME: *daemon-config # temporary
```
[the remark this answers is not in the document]
ricab: Nope, it was intentional, to keep `DAEMON_CONFIG_HOME` outsite of the `<<: &client-environment`: I saw no point in having it propagated to the GUI client (unlike the `XDG*` ones, which I thought made sense).

snap/snapcraft.yaml
```diff
@@ -17,31 +17,36 @@ apps:
...
-      QT_PLUGIN_PATH: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qt5/plugins
-    command: bin/multipass-gui
+    environment: *client-environment
+    command: bin/launch-multipass-gui
   libvirt-bin:
     command: bin/launch-libvirtd
     environment:
-      LD_LIBRARY_PATH: $SNAP/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET
-      PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
+      <<: *client-environment
```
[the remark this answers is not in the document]
ricab: OK, I noticed that but did not bother changing it since it is going away.

src/platform/platform_linux.cpp
```diff
@@ -42,23 +44,52 @@ mp::ProcessFactory* process_factory()
...
+QString mp::platform::default_driver()
 {
-    auto driver = qgetenv("MULTIPASS_VM_DRIVER");
+    return QStringLiteral("qemu");
+}
 
-    if (driver.isEmpty() || driver == "QEMU")
+QString mp::platform::daemon_config_home() // temporary
+{
+    auto ret = QString{qgetenv("DAEMON_CONFIG_HOME")};
```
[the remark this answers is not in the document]
ricab: We need to implement `mp::platform::daemon_config_home` there, but it does not have to use the env var.

src/platform/platform_linux.cpp
ricab: OK, will do.

src/platform/platform_linux.cpp
ricab: It was just trying to help the user in case he was used to those drivers elsewhere. But yes, it is probably best to remove, I will do that.

snap/snapcraft.yaml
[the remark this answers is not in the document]
Saviq: Ah, now I see what you did here. Should've looked at the whole file, not just the diff ;)

src/platform/platform_linux.cpp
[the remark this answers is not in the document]
Saviq: Right.

ricab:
> nicer error message

Like what? Do you mean adding "try sudo" or something? Or just totally different message?

> Just realized:

Yes, I thought the equals-sign was just a quirk of description, since we separate all positional arguments w/ space currently. I will change it if it really was intentional.

> This works, but shouldn't:

Well, those will fail the daemon when it tries to restart (just like the env vars and `snapctl`). I thought it appropriate to maintain this behavior for now because it keeps the responsibility of semantics-checking in the consumer &ndash; the daemon, where it will naturally be when we move to RPC. Do you want a temporary `mp::platform` check being called from the client instead?

Saviq:
> Like what? Do you mean adding "try sudo" or something? Or just totally different message?

Yeah, something like that.

> Yes, I thought the equals-sign was just a quirk of description, since we separate all positional arguments w/ space currently. I will change it if it really was intentional.

It was intentional so you can provide multiple key/value pairs without those getting confusing.

> Well, those will fail the daemon when it tries to restart (just like the env vars and `snapctl`). I thought it appropriate to maintain this behavior for now because it keeps the responsibility of semantics-checking in the consumer – the daemon, where it will naturally be when we move to RPC. Do you want a temporary `mp::platform` check being called from the client instead?

Right, the problem is that the client then just says:
```
$ multipass find
find failed: cannot connect to the multipass socket
Please ensure multipassd is running and '/run/multipass_socket' is accessible
```

And setting a different value doesn't resolve that, since the daemon's dead by then and won't be restarted automatically.

ricab:
I see. Alright, fixing pending items.

src/daemon/daemon_main.cpp
```diff
@@ -102,6 +111,16 @@ class UnixSignalHandler
...
     mp::AutoJoinThread signal_handling_thread;
 };
+
+void monitor_and_quit_on_settings_change()
+{
+    static const auto filename = mp::Settings::get_daemon_settings_file_path();
+    init_settings(filename); // create if not there
+
+    static QFileSystemWatcher monitor{{filename}};
+    QObject::connect(&monitor, &QFileSystemWatcher::fileChanged, [] { QCoreApplication::exit(EXIT_FAILURE); });
```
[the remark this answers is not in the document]
ricab: Yeah, done.

ricab:
Only last couple of points missing (`=` and driver-value check).

ricab:
OK, so those two items are in. But reading back #756 and comments above, there is actually a 3rd item that I did not push yet: supporting multiple key-value pairs in `set`. For that, I would need a strategy to handle invalid keys/values. I considered the following options (in increasing preference):

1. Process arguments in order and stop setting upon error.
2. Set as much as we can from the command line arguments, skipping failures and continuing with remaining arguments, showing all errors in the end.
3. Only set anything if all settings succeed - requires transactional setting API.

Option 1 would mean committing to an order or leaving the settings in an undefined state. Either way, it does not sound very user-friendly.

Option 2 would require bookkeeping shenanigans and clarifying what caused what in the output (think different error causes in different arguments, building error text in an iterating maner from multiple exception msgs and corresponding args...). This would probably feel clunky to the user (not to mention the programmer).

Both 1 and 2 suffer from other downsides as well:
  * the various options may be related, meaning the user does not want to set one without the other
  * easy to do something like `multipass set key1=val1K ey2=val2` and get `key1=val1K` recorded (if there were no particular requirements on `key1`'s value)
  * these approaches are not scalable if we want to accept plain YAML in the future

Option 3 is the cleaner one IMO. We only let the request go through if it is 100% correct. While this could be done by keeping former values to revert to in case of error, that could be a bit clumsy &ndash; there could be problems when trying to revert and things would be left in an undefined state. 

Doing option 3 properly requires changing the current interface of the `Settings` class to receive multiple options at once. I have an implementation, which is a bit *hackish* as long as we save to separate files, but that would go away. It *mostly* works, though it would still need some work, particularly on tests. I say "mostly" for two reasons:

  1. Since we use separate files, one of them needs to be written before the other. If there is a failure writing the second one, it was not transactional.
  2. Because we need `sudo` for daemon settings, the following undesirable interaction sequence becomes likely:

```
$ multipass set client.primary_name=yoyo local.driver=libvirt
Unable to read/write settings: access error (consider `sudo`)
$ sudo multipass set client.primary_name=yoyo local.driver=libvirt
$ multipass set client.primary_name=asdf
Unable to read/write settings: access error (consider `sudo`)
```

*Sudoing* the second step above would create a root-owned client file. Client settings would not work without `sudo` anymore.

For this reason (reason 2 above), and because we have only two settings for the time being, I thing our best option is to allow only one argument in `set` until we move to RPC for daemon settings. However, we will still have to deal with problem 1, even then. If, in order to address that, we decide to disallow `set` on multiple entities simultaneously (i.e. clients/daemons), that could be done today. It would solve problem 2 today, albeit only in abstract: since we have no more than one option per entity at present, no multiple-arg `set` would be accepted.

ricab:
PS: In either case, perhaps this PR can already be merged. I would then deal with the other platforms before moving on to further functionality.

Saviq:
@ricab yeah, multi-set would be a later step towards full config story, I didn't mean it should happen in this PR, just that the `<key>=<value>` format is more friendly for that case.

src/client/cli/cmd/set.cpp
```diff
@@ -63,22 +63,33 @@ QString cmd::Set::description() const
...
+        "keyval",
+        "A key-value pair. The key specifies a path to the option to configure. The value is its intended value.",
+        "<key>=<value>");
 
     auto status = parser->commandParse(this);
     if (status == ParseCode::Ok)
     {
         const auto args = parser->positionalArguments();
-        if (args.count() == 2)
+        if (args.isEmpty())
```
Saviq: If someone provides more than one, it will be silently dropped?

include/multipass/exceptions/settings_exceptions.h
```diff
@@ -38,8 +38,8 @@ class SettingsException : public std::runtime_error
 class PersistentSettingsException : public SettingsException
 {
 public:
-    PersistentSettingsException(const QString& attempted_operation, const QString& failure)
-        : SettingsException{fmt::format("Could not {} settings: {} error", attempted_operation, failure)}
+    PersistentSettingsException(const QString& attempted_operation, const QString& detail)
+        : SettingsException{fmt::format("Unable to {} settings: {}", attempted_operation, detail)}
```
Saviq: I won't block on this, but we'll have to move away from constructing the error message like this later.

src/utils/settings.cpp
```diff
@@ -85,9 +85,9 @@ void check_status(const QSettings& settings, const QString& attempted_operation)
 {
     auto status = settings.status();
     if (status || exists_but_unreadable(settings.fileName()))
-        throw mp::PersistentSettingsException{attempted_operation, status == QSettings::FormatError
-                                                                       ? QStringLiteral("format")
-                                                                       : QStringLiteral("access")};
+        throw mp::PersistentSettingsException{
+            attempted_operation, status == QSettings::FormatError ? QStringLiteral("format error")
+                                                                  : QStringLiteral("access error (consider `sudo`)")};
```
Saviq: How will we approach this for Windows, where `sudo` is not a thing?

src/client/cli/cmd/set.cpp
```diff
@@ -63,22 +63,33 @@ QString cmd::Set::description() const
...
-            val = args.at(1);
+            cerr << "Insufficient arguments.\n";
+            status = ParseCode::CommandLineError;
         }
         else
         {
-            cerr << "Need exactly one option key and one value\n";
-            status = ParseCode::CommandLineError;
+            const auto keyval = args.at(0).split('=');
+            if (keyval.count() != 2 || keyval.contains(QStringLiteral("")))
```
Saviq: Long-term we'll need to be smarter here, I can imagine some values including `=` characters. So either we split at the first one only, or require some escape sequences.

Not in this PR, though.

src/client/cli/cmd/set.cpp
ricab: Good catch, I did not do this because I meant to evolve to multiple arguments and then forgot.

src/utils/settings.cpp
ricab: Good point. How about "consider running this with an administrative role"? (Replacing it here and having the same msg in all platforms)

src/client/cli/cmd/set.cpp
ricab: OK, leaving it for now. `multipass config` would be an alternative for those cases in the future.

include/multipass/exceptions/settings_exceptions.h
ricab: I am not sure what the issue is. Note that `attempted_operation` could be an enum here, representing read and write ops, that we would then turn into string. Then `detail` is a separate phrase. How could I make this friendlier to translation? Perhaps an example of what problems this could cause would help me understand better.

ricab:
> I didn't mean it should happen in this PR

OK, the reasoning stands for *when* it should happen, and can perhaps elucidate *whether* it should happen.

src/utils/settings.cpp
Saviq: Works for me.

include/multipass/exceptions/settings_exceptions.h
Saviq: Basically what I mean is that "Unable to format settings" may translate to something completely different than "Unable to access settings" depending on language.

Different languages work in very different ways, building a sentence from a set of strings is bound to break at one point or another. It's a discussion to have over beers, though :)

I'm just sensitive to this because I've translated a lot of software over the years and Polish is one of the more complex languages out there so it easily highlights problems in approaches like that.

ricab: I think I understand the idea. To avoid duplication, I think I would then try a more "square" way, e.g. "Unable to operate on settings. Operation: {}; Details: {}". But this might not feel so good for the user... Isn't this the sort of thing that one would do when preparing the code for translation? Preemptively duplicating messages in the code does not feel right to me either.

BTW, the operation here is just "read" or "read/write", "format/access" goes in the second param as a free string.

include/multipass/platform.h
```diff
@@ -26,13 +26,18 @@
...
 #include <string>
 
 namespace multipass
 {
 namespace platform
 {
 std::string default_server_address();
+QString default_driver();
+QString daemon_config_home(); // temporary
+bool is_backend_supported(const QString& backend); // temporary
```
townsend2010: What is your plan for these on the other platforms?

ricab: For 38 and 40 I was planning to maintain the behavior I find there. So, for instance, on win accept "hyperv" and "virtualbox" (keeping with the lowercase theme, but we could of course accept capitals as well if prefered). 

As for 39, I was planning to do something analogous to what we did here. That is, find the folder where the settings would naturally be saved by the daemon if we were already using RPC and have the client save them there, requiring admin privileges if necessary. 

How does that sound?

ricab: On a more practical note, now would probably be the right time to get the windows install back up to speed in my other machine. As for mac, I would need help testing or remote access.

townsend2010: Ok ,works for me.  Just wanted to make sure there was a plan in place;)

PR #837, opened 2019-06-12
utils: Re-add wait_for_cloud_init()

### LINKED ISSUE #696: multipass edge is sometimes late executing cloud-init directives
This is a script I was working on before running into https://github.com/CanonicalLtd/multipass/issues/690

It works perfectly in the beta channel, but fails in the edge channel. The installation of nginx happens much later after multipass launch finishes. If I test on a single VM, it works, but when doing two VMs back-to-back, it fails:

```
karl@tp-work:~/Projects/linux-ha-notes/keepalived-quickstart$ snap info multipass
name:      multipass
summary:   Ubuntu at your fingertips
publisher: Canonical✓
license:   unset
description: |
  Multipass gives you Ubuntu VMs in seconds. Just run `multipass launch` and it'll do all the setup
  for you. Check out `multipass find` for a list of available images. More details you can find
  under `multipass help`.
commands:
  - multipass
  - multipass.virsh
services:
  multipass.libvirt-bin: simple, enabled, inactive
  multipass.multipassd:  simple, enabled, active
snap-id:      mA11087v6dR3IEcQLgICQVjuvhUUBUKM
tracking:     edge
refresh-date: yesterday at 17:54 CET
channels:
  stable:           –                                             
  candidate:        –                                             
  beta:             v0.5                   2019-01-22 (587) 111MB classic
  edge:             0.6.0-dev.218+g3febdd7 2019-03-27 (700) 112MB classic
  core16/stable:    –                                             
  core16/candidate: –                                             
  core16/beta:      v0.5                   2019-03-07 (587) 111MB classic
  core16/edge:      v0.5-128-g47ceef4      2019-03-07 (659) 112MB classic
installed:          0.6.0-dev.214+gec3848d            (697) 112MB classic
```
keepalived-cloud-config.yaml:
```yaml
#cloud-config

bootcmd:
  # https://github.com/CanonicalLtd/multipass/issues/666
  - sudo ln -fs /run/systemd/resolve/resolv.conf /etc/resolv.conf

packages:
  - nginx
  - keepalived
```
keepalived-quickstart.sh:
```bash
#!/bin/bash

set -eu

echo
echo "### Rebuilding ka-node1 and ka-node2"

multipass delete -p ka-node1 || true
multipass delete -p ka-node2 || true

multipass launch daily:cosmic --cloud-init ./keepalived-cloud-config.yaml --name ka-node1
multipass launch daily:cosmic --cloud-init ./keepalived-cloud-config.yaml --name ka-node2

# -------------------------------------------------------------------

echo
echo "### Gathering network information"

ip_node1=$(multipass exec ka-node1 -- ip addr | grep 'inet ' | grep global | head -1 | sed 's/.*inet \([0-9]*\.[0-9]*\.[0-9]*\.[0-9]*\).*/\1/g')
ip_node2=$(multipass exec ka-node2 -- ip addr | grep 'inet ' | grep global | head -1 | sed 's/.*inet \([0-9]*\.[0-9]*\.[0-9]*\.[0-9]*\).*/\1/g')
ip_dot3=$(multipass exec ka-node1 -- ip addr | grep 'inet ' | grep global | head -1 | sed 's/.*inet \([0-9]*\.[0-9]*\.[0-9]*\)\..*/\1.3/g')

echo "ka-node1 is at $ip_node1, ka-node2 is at $ip_node2, and our virtual IP will be at the .3 address of this subnet: $ip_dot3"

# -------------------------------------------------------------------

echo
echo "### We're using nginx as our highly available service. Create home pages so we can see when it's running."
echo
echo "Note: You'd normally use the same content on both sites, but we want to see via the browser which node is running!"

multipass exec ka-node1 -- sudo bash -c "echo '<html><h1>Primary</h1></html>' >/var/www/html/index.html"
multipass exec ka-node2 -- sudo bash -c "echo '<html><h1>Secondary</h1></html>' >/var/www/html/index.html"
```
Result:
```
karl@tp-work:~/Projects/linux-ha-notes/keepalived-quickstart$ ./keepalived-quickstart.sh 
/snap/bin/multipass
### Rebuilding ka-node1 and ka-node2
Launched: ka-node1                                                              
Launched: ka-node2                                                              
                                                                                
### Gathering network information
ka-node1 is at 10.214.41.37, ka-node2 is at 10.214.41.99, and our virtual IP will be at the .3 address of this subnet: 10.214.41.3

### We're using nginx as our highly available service. Create home pages so we can see when it's running.

Note: You'd normally use the same content on both sites, but we want to see via the browser which node is running!
bash: /var/www/html/index.html: No such file or directory
```
This looks like a timing issue:

```
karl@tp-work:~/Projects/linux-ha-notes/keepalived-quickstart$ multipass exec ka-node1 -- bash -c "apt search nginx |grep '^nginx/'"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

nginx/cosmic-updates,cosmic-security,now 1.15.5-0ubuntu2.1 all [installed]
karl@tp-work:~/Projects/linux-ha-notes/keepalived-quickstart$ multipass exec ka-node2 -- bash -c "apt search nginx |grep '^nginx/'"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

nginx/cosmic-updates,cosmic-security 1.15.5-0ubuntu2.1 all
```
30 seconds later:

```
karl@tp-work:~/Projects/linux-ha-notes/keepalived-quickstart$ multipass exec ka-node2 -- bash -c "apt search nginx |grep '^nginx/'"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

nginx/cosmic-updates,cosmic-security,now 1.15.5-0ubuntu2.1 all [installed]
```

townsend2010:
This allows launch to wait for a long running custom cloud-init.

Fixes #696

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -59,7 +59,7 @@ public:
     virtual std::string ipv4() = 0;
     virtual std::string ipv6() = 0;
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
-    virtual void ensure_vm_is_running(){};
+    virtual void ensure_vm_is_running() = 0;
     virtual void update_state() = 0;
 
     VirtualMachine::State state;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -59,6 +59,7 @@ public:
     virtual std::string ipv4() = 0;
     virtual std::string ipv6() = 0;
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
+    virtual void ensure_vm_is_running(){};
     virtual void update_state() = 0;
 
     VirtualMachine::State state;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2115,14 +2115,14 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
 }
 
 QFutureWatcher<mp::Daemon::AsyncOperationStatus>*
-mp::Daemon::create_future_watcher(std::function<void()> const& finished_operation)
+mp::Daemon::create_future_watcher(std::function<void()> const& finished_op)
 {
     async_future_watchers.emplace_back(std::make_unique<QFutureWatcher<AsyncOperationStatus>>());
 
     auto future_watcher = async_future_watchers.back().get();
     QObject::connect(future_watcher, &QFutureWatcher<AsyncOperationStatus>::finished,
-                     [this, future_watcher, finished_operation] {
-                         finished_operation();
+                     [this, future_watcher, finished_op] {
+                         finished_op();
                          finish_async_operation(future_watcher->future());
                      });
 
@@ -2140,14 +2140,17 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
         auto vm = it->second;
         vm->wait_until_ssh_up(up_timeout);
 
-        if (server)
+        if (std::is_same<Reply, LaunchReply>::value)
         {
-            Reply reply;
-            reply.set_reply_message("Waiting for initialization to complete");
-            server->Write(reply);
-        }
+            if (server)
+            {
+                Reply reply;
+                reply.set_reply_message("Waiting for initialization to complete");
+                server->Write(reply);
+            }
 
-        mp::utils::wait_for_cloud_init(vm.get(), cloud_init_timeout, *config->ssh_key_provider);
+            mp::utils::wait_for_cloud_init(vm.get(), cloud_init_timeout, *config->ssh_key_provider);
+        }
 
         std::vector<std::string> invalid_mounts;
         auto& mounts = vm_instance_specs[name].mounts;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -74,6 +74,7 @@ constexpr auto uuid_file_name = "multipass-unique-id";
 constexpr auto metrics_opt_in_file = "multipassd-send-metrics.yaml";
 constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
+constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
 constexpr auto max_install_sshfs_retries = 3;
 constexpr auto default_mem = "1G";
@@ -1911,11 +1912,12 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                     auto& vm = vm_instances[name];
                     vm->start();
 
-                    reply.set_vm_instance_name(name);
-                    config->update_prompt->populate_if_time_to_show(reply.mutable_update_info());
-                    server->Write(reply);
-
-                    auto future_watcher = create_future_watcher();
+                    auto future_watcher = create_future_watcher([this, server, name] {
+                        LaunchReply reply;
+                        reply.set_vm_instance_name(name);
+                        config->update_prompt->populate_if_time_to_show(reply.mutable_update_info());
+                        server->Write(reply);
+                    });
                     future_watcher->setFuture(QtConcurrent::run(this, &Daemon::async_wait_for_ready_all<LaunchReply>,
                                                                 server, std::vector<std::string>{name},
                                                                 status_promise));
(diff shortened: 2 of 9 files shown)


Saviq:
GAH! I just hate that this isn't inherited from a base `VirtualMachine`…

townsend2010:
@Saviq,

Yeah, I too would have liked many of the classes to not be pure abstract so we could just inherit base functionality from them.  I tried to do this at one time, but was overruled.

townsend2010:
I re-worked this into a hopefully better piece of code:)

src/daemon/daemon.cpp
```diff
@@ -2134,6 +2140,15 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
         auto vm = it->second;
         vm->wait_until_ssh_up(up_timeout);
 
+        if (server)
+        {
+            Reply reply;
+            reply.set_reply_message("Waiting for initialization to complete");
```
ricab: I can't see this message in my testing. I confirm multipass waited for cloud init. Debugging, the code does go through here, but it does not cause the message to be shown in the client.

include/multipass/virtual_machine.h
```diff
@@ -59,6 +59,7 @@ class VirtualMachine
     virtual std::string ipv4() = 0;
     virtual std::string ipv6() = 0;
     virtual void wait_until_ssh_up(std::chrono::milliseconds timeout) = 0;
+    virtual void ensure_vm_is_running(){};
```
ricab: Hmm, so this is just a way to detect whether or not we are in a mock? Not sure that is advisable... And having the implementation here drops the pure interface.

I would prefer keeping this an interface and having a pure virtual `wait_for_cloud_init` here. Then an intermediate descendant (i.e. `VirtualMachineBase` or something) to implement it, and then real implementations inheriting from that one. The mock one could then throw or do whatever appropriate. If we later need to move to the alternative *composition-over-inheritance* approach that would be easy.

src/daemon/daemon.cpp
townsend2010: Right, because it's nearly instantaneous for our default cloud-init.  In order to see it, you'll need a custom cloud-init that has a long running operation.

include/multipass/virtual_machine.h
townsend2010: No, this is not used a just a way to detect whether we are in a mock or not.  It's used to ensure the particular backend's instance is actually running.  I had to make it public so utils can call it.  But it's been private before.  If we want this to remain an interface, then I can make `ensure_vm_is_running` a pure virtual since it's needed by all the backends any ways.  Does that work for you?

I don't think it's necessary to have `wait_for_cloud_init` as a pure virtual since everything that is needed by that call can be handled via the utils call.

src/daemon/daemon.h
```diff
@@ -163,7 +163,8 @@ public slots:
                                                   const std::vector<std::string>& vms,
                                                   std::promise<grpc::Status>* status_promise);
     void finish_async_operation(QFuture<AsyncOperationStatus> async_future);
-    QFutureWatcher<AsyncOperationStatus>* create_future_watcher();
+    QFutureWatcher<AsyncOperationStatus>* create_future_watcher(std::function<void()> const& finished_operation = []() {
```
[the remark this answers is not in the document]
townsend2010: I guess I can call it `finished_op` or something???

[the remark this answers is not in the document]
ricab: That would do perfectly. The newline with the closing bracket makes it look like a default implementation or something.

src/daemon/daemon.cpp
ricab: I don't think that's it &ndash; I tried with a *sleep-30* cloud-init, and indeed the launch waits until the sleep is completed, but the message never shows.

Here is what I mean. The cloud init file:

```yaml
output: { all: "| tee -a /var/log/cloud-init-output.log" }
runcmd:
  - date
  - echo sleeping 30
  - sleep 30
  - echo slept
  - date
```
Then launching with `multipass launch --cloud-init ... && date`. Comparing timestamps you can see that the command indeed finishes only after cloud-init completes, but the msg does not show.

townsend2010: Hmm, ok, I'll look into it.  Thanks!

include/multipass/virtual_machine.h
ricab: Oh you're right! I looked only at the diff and thought `ensure_vm_is_running` had been reintroduced now, but then only saw an empty implementation in the base and a throwing one in the stub... fogettaboudit 

If you could just keep it pure virtual then, to make it more tidy...

src/daemon/daemon.h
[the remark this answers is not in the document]
townsend2010: Ok, done

src/daemon/daemon.cpp
townsend2010: Found the issue and now fixed.

include/multipass/virtual_machine.h
townsend2010: Ok, `ensure_vm_is_running` is now a pure virtual.

src/daemon/daemon.cpp
townsend2010: Well, it prints this out now on start too.  I don't think we want that.

townsend2010: Ok, that's better :wink:

PR #838, opened 2019-06-13
Make Process an interface, LinuxProcess the implementation.

### LINKED ISSUE #705: `Process` arguments treatment unclear
#703 showed an issue in the [`Process`](https://github.com/CanonicalLtd/multipass/blob/v0.6.0-rc/src/platform/backends/shared/linux/process.cpp) class that `extra_arguments` pile up on subsequent runs.

 The reason is that [`QProcess::start()`](https://doc.qt.io/qt-5/qprocess.html#start) actually overwrites the `arguments` property, which then is used on subsequent runs in [`Process::start()`](https://github.com/CanonicalLtd/multipass/blob/v0.6.0-rc/src/platform/backends/shared/linux/process.cpp#L41).

The [immediate solution](../pull/704) was to use `process_spec::arguments()` instead of `QProcess::arguments()` in `Process::start()`, but that means the behaviour of the underlying `QProcess` is modified - calling `QProcess::setArguments()` on it will be ignored.

This deserves a re-think on how the `Process` API should look and a suite of tests.

gerboland:
Tidies up the Process API so doesn't inherit all of QProcess bits (and also fix #705).

This is mostly a refactor with the goal of making Process mockable/fake-able. I added
simple tests to verify QemuBackend launches dnsmasq and qemu, and plan to add more.


gerboland:
> Code looks good overall. Just a few questions to help me understand this better.
> 
>     1. This looks like a wrapper around `QProcess`.  Is that correct?  And you said the main point of this is so we can mock a `QProcess`, right?

Ultimately yes.

>     2. You have a specific `LinuxProcess` class.  Can this class be used on Mac as well for HyperKit?  If so, maybe calling it `UnixProcess` would be better?  If not used on Mac, then nevermind 😛

It's probably going to change again, but I just kept the name on-topic for now. It's not used on Mac or Windows (yet).

For future there will be tiny differences between each platform's *Process implementation, due to the different ways each implements the "kill child if parent dies" function.

Saviq:
bors r=townsend2010

PR #850, opened 2019-06-19
snap: Remove libvirt from the snap

### LINKED ISSUE #482: Connect to the libvirt interface
multipass has libvirt support, but currently that means spawning its own copy, siloed. Not very useful is you have libvirt networks, storage, images, etc already available on the host.

It would be useful if multipass could connect to the host libvirt. Either by providing a socket path, in the case of the classic snap, or by using the libvirt snap interface.

### LINKED ISSUE #846: snap: Remove libvirt from the snap
Instead, use a host installed libvirt and connect to it via the `libvirt` snap interface.

* Need to ensure migrations of current Multipass libvirt users to the host libvirt work correctly.
* Have a meaningful failure if libvirt is not available on the host or if the `libvirt` snap interface does not work.

townsend2010:
Maintaining libvirt in the snap causes burdens for not much gain. Anyone wanting to
use libvirt can just use the one installed on the host.

Fixes #482, fixes #846

### DIFF
--- a/snap-wrappers/bin/launch-libvirtd
+++ b/snap-wrappers/bin/launch-libvirtd
@@ -1,7 +1,0 @@
-#!/bin/sh -e
-
-driver="$(snapctl get driver)"
-
-if [ "$driver" = "LIBVIRT" ]; then
-    exec "$SNAP/sbin/libvirtd"
-fi
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -1,5 +1,0 @@
-#!/bin/sh
-set -e
-
-install -D $SNAP/var/snap/multipass/common/libvirt/libvirtd.conf $SNAP_COMMON/libvirt/libvirtd.conf
-sed -i 's/unix_sock_group = "libvirtd"/unix_sock_group = "sudo"/' $SNAP_COMMON/libvirt/libvirtd.conf
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -1,7 +1,0 @@
-#!/bin/sh
-set -e
-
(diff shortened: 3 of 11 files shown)


Saviq:
I built the snap clean and this happened:

```
The 'multipass' part needs the following libraries that are not included in the snap or base: 
usr/lib/x86_64-linux-gnu/libavahi-client.so.3
usr/lib/x86_64-linux-gnu/libavahi-common.so.3
usr/lib/x86_64-linux-gnu/libicudata.so.55
usr/lib/x86_64-linux-gnu/libicuuc.so.55
usr/lib/x86_64-linux-gnu/libvirt.so.0
usr/lib/x86_64-linux-gnu/libxml2.so.2
These dependencies can be satisfied via more stage-packages, more parts, or content sharing.
…
multipass list
/snap/multipass/x4/bin/multipass: /snap/core/current/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /usr/lib/x86_64-linux-gnu/libxml2.so.2)
```

Looks like we lost some dependencies.

Saviq:
In general this works and the only issue I found is most likely preexisting:
1) `snap set multipass driver=LIBVIRT`
1) launch an instance
1) stop the instance
1) `snap set multipass driver=QEMU`

The domain remains defined in libvirt, which may result in *weird* situations when someone tries to manage it with virsh / virt-manager or some such.

townsend2010:
@Saviq,

> The domain remains defined in libvirt

Yes, this is an already occurring issue.  `multipassd` is ignorant about which backend it used in the previous running instance of the process...which is really the crux of the issue in #854.

Saviq:
A'ight!

bors r+

PR #852, opened 2019-06-21
rpc: Unify virtual machine state values via the rpc enum

townsend2010:
Previously, possible virtual state values were kept in two different places, so
a "translation" was necessary to switch back and forth between the two. Now, just
use the InstanceState enum in the rpc everywhere.

### DIFF
--- a/include/multipass/cli/format_utils.h
+++ b/include/multipass/cli/format_utils.h
@@ -33,7 +33,7 @@ class Formatter;
 
 namespace format
 {
-std::string status_string_for(const InstanceStatus& status);
+std::string state_string_for(const InstanceState& state);
 std::string image_string_for(const multipass::FindReply_AliasInfo& alias);
 Formatter* formatter_for(const std::string& format);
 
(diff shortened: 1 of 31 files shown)


townsend2010:
Hey @gerboland,

You said you'd give this some more thought since you said de-duplication is also a Good Thing :smiley: 

Have you given it any more thought?  I'd like some resolution of this so I can move on and add a new state that is needed.

gerboland:
> Hey @gerboland,
> 
> You said you'd give this some more thought since you said de-duplication is also a Good Thing 😃
> 
> Have you given it any more thought? I'd like some resolution of this so I can move on and add a new state that is needed.

I don't think deduplication is good at a cost of this kind of coupling.

I looked but found nothing nice. gRPC/protobuf is a closed box. The translating enums approach we do now is only realistic option I can find.

townsend2010:
@gerboland,

Ok, I'll agree to disagree in that I think there are many other numerous places in our code that have tightly coupled technologies for the sake of convenience and de-duplication.  That said, it's not worth going back and forth on it and I think I can make what we have better and still satisfy the de-coupling here.

PR #856, opened 2019-06-24
Make ProcessFactory a (testable) singleton

gerboland:
We need to launch Processes all around our codebase. Piping a ProcessFactory pointer around to all those location is tedious. Instead I was inspired by @ricab 's Singleton work, and wish to propose this.

We'll be able to use this to launch a Process as conveniently as we do with QProcess, but it will be fully mockable & testable.

I've added StubProcessFactory to enhance our testing, as it keeps track of each process we launch, so we can check their arguments (just applied to Qemu backend tests currently, will do others next). I'll have a Mock ready shortly in a followup PR.

### DIFF
--- a/tests/test_dnsmasq_server.cpp
+++ b/tests/test_dnsmasq_server.cpp
@@ -21,10 +21,8 @@
 #include <multipass/logging/logger.h>
 
 #include "file_operations.h"
-#include "stub_process_factory.h"
 #include "temp_dir.h"
 #include "test_with_mocked_bin_path.h"
-
 #include <QDir>
 
 #include <memory>
@@ -48,7 +46,7 @@ struct CapturingLogger : public mp::logging::Logger
     mutable std::vector<std::string> logged_lines;
 };
 
-struct DNSMasqServer : public ::testing::Test
+struct DNSMasqServer : public mpt::TestWithMockedBinPath
 {
     DNSMasqServer()
     {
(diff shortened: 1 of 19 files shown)


gerboland:
Bah I break Mac build due to merge conflict. Rest should be fine tho

ricab:
I had a (quick) look and I like the direction this is going.

tests/test_dnsmasq_server.cpp
```diff
@@ -67,7 +65,6 @@ struct DNSMasqServer : public mpt::TestWithMockedBinPath
 
     mpt::TempDir data_dir;
     std::shared_ptr<CapturingLogger> logger = std::make_shared<CapturingLogger>();
-    mpt::StubProcessFactory process_factory;
     const QString bridge_name{"dummy-bridge"};
```
ricab: Isn't this missing a `StubProcessFactory::Inject()`, here and in libvirt tests? Otherwise this now creates actual processes, while that was not the case before.

tests/test_libvirt_backend.cpp
```diff
@@ -60,7 +60,6 @@ struct LibVirtBackend : public Test
     }
     mpt::TempFile dummy_image;
     mpt::TempFile dummy_cloud_init_iso;
-    mpt::StubProcessFactory process_factory;
```
ricab: idem

tests/test_dnsmasq_server.cpp
gerboland: This test actually does create actual processes, its test harness `mpt::TestWithMockedBinPath`  messes with the $PATH to launch a test binary.

I thought this change was most minimal to keep the test working - and restores it to be more like the original design. Likely I'll refactor this test to use MockProcessFactory later.

tests/test_libvirt_backend.cpp
gerboland: Above reply covers this too.

tests/test_dnsmasq_server.cpp
ricab: Alright

PR #857, opened 2019-06-24
 Transition Qemu snapshot checking to ProcessFactory and test it 

gerboland:
Introduces MockProcesFactory, which creates mockable Processes. Use this to test Qemu snapshot checking.

This involved some refactoring:
- create concept of Qemu command versioning (so when we can support resuming suspended VMs run with older command lines). This already existed in the "use_cdrom" metadata flag, I've generalised it to be more future-proof (support for "use_cdrom" has been maintained however).
- move all Qemu command line generation into the QemuVMProcessSpec class, and test it.
- add a convenience API: `ProcessFactory::create_process(command, arguments)` so one doesn't need to create a Spec for every process.
- this allowed me to remove an API wart in `Process`, namely the "extra_arguments" idea. This was confusing, and made testing difficult. Now once `Process` is constructed, its command, arguments, environment etc are fixed.


ricab:
I had a (quick) look and I like the direction this is going.

PR #860, opened 2019-06-26
Qemu: enable image compaction/garbage collection

gerboland:
Please land #857 first.

This adds a Qemu command version 2, that enable Qemu's image "discard=unmap" feature, which causes the disk image to shrink when blocks are trimmed by the VM kernel.

A way to see this in action is to monitor the disk image size, while running these commands in the VM:

    dd if=/dev/zero of=bigfile bs=1M count=1024

which will increase its size, and then 

    rm bigfile
    sudo fstrim -v /

which should cause it to shrink again.


src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -17,13 +17,49 @@
...
+ *      Replaced: "-hda image.img"
+ *      With:     "-device virtio-scsi-pci,id=scsi0
+ *                 -drive file=image.img,if=none,format=qcow2,discard=unmap,id=hda
+ *                 -device scsi-hd,drive=hda,bus=scsi0.0"
+ *  1 - changed how cloud-init ISO was specified:
+ *      Replaced: "-drive file=cloud-init.iso,if=virtio,format=raw,snapshot=off,read-only"
+ *      With:     "-cdrom cloud-init.iso"
+ *      Note this was originally encompassed in the metadata as "use_cdrom" being true.
+ *  0 - original
+ */
```
Saviq: I would prefer we just save the whole `qemu` command line used on startup and discard on shutdown, this way we'd achieve 100% the same with much less complexity…

gerboland: I thought the tap network device interface name could change between suspend/resume. But aside from that, sure, that can be done.

PR #862, opened 2019-06-26
snap: Remove the libvirt plug since we are using classic confinement

townsend2010:
It's considered an error to define a plug while using classic confinement.


townsend2010:
This is a result of the error seen in https://build.snapcraft.io/user/CanonicalLtd/multipass/600227.

PR #864, opened 2019-06-26
[snap] classic doesn't allow plugs

Saviq:


Saviq:
See e.g. https://dashboard.snapcraft.io/snaps/multipass/revisions/911/

ricab:
I think you missed #862 which I just approved

Saviq:
Indeed :)

PR #874, opened 2019-06-28
image_vault: Use proper id when finding an existing image

### LINKED ISSUE #867: Launching an instance is resulting in the image being downloaded every time
As reported by @OddBloke, launching an instance always resulted in downloading the image despite it being cached.

Some debugging revealed the `multipassd-image-records.json` file is out of sync somehow for the xenial image:
```
    "ccd6e054990d4b5f119f3a4e74cada135f2f3e9452458738230624ed8a604ed4": {
        "image": {
            "aliases": [
                {
                    "alias": "16.04"
                },
                {
                    "alias": "x"
                },
                {
                    "alias": "xenial"
                }
            ],
            "current_release": "",
            "id": "74baf6380a51398877a7e7d27d2319b765fee0efbf94a3be6ebc29575658c5a4",
            "initrd_path": "",
            "kernel_path": "",
            "original_release": "16.04 LTS",
            "path": "/var/snap/multipass/common/cache/multipassd/vault/images/xenial-20190530.3/ubuntu-16.04-server-cloudimg-amd64-disk1.img",
            "release_date": ""
        },
        "last_accessed": 1560951473721076700,
        "query": {
            "persistent": false,
            "query_type": 0,
            "release": "xenial",
            "remote_name": ""
        }
    }
```
Note the entry's key does not match the "id" field and they should.  Need to debug further to figure out how such a state can occur and how to recover from it.

townsend2010:
This fixes an issue where the wrong hash id was being used when finalizing
an image used in launch.

Also, always use all info when updating the prepared_image_records map.

Fixes #867

### DIFF
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -421,7 +421,7 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
                         const auto prepared_image = record.second.image;
                         try
                         {
-                            return finalize_image_records(query, prepared_image, id);
+                            return finalize_image_records(query, prepared_image, record.first);
                         }
                         catch (const std::exception& e)
                         {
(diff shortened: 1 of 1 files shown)


Saviq:
Could we have tests?

townsend2010:
@Saviq,

I looked into adding a test for this, but it will require a large amount of test code in order to reproduce the scenario that caused this issue.  I'm not sure if spending a bunch of time for one test is the best use of my time given our backlog.  That's your call though.

Saviq:
In that case:

bors r=gerboland

PR #876, opened 2019-06-28
Fix absent settings dir and consequent inotify watch failure

ricab:
Fixes absent settings directory, which prevented the daemon from initializing its settings file to watch for changes. Also drops XDG dir creation in the snap launchers (see corresponding commit msg). The `Settings` can create its files on its own (relying on `QSettings`) for that.


Saviq:
The launchers for `multipass` and `multipass-gui` are now NOOP, just drop them please.

ricab:
@Saviq Thanks for noticing that, I just pushed the fix. BTW, how can I check the results of snap testing you recently added? Is it automatic in our existing CI?

Saviq:
@ricab yeah, it's just an additional step to the `RelWithDebInfo` job, [here's an example](https://travis-ci.org/CanonicalLtd/multipass/jobs/551938661#L3465) from the previous run of this PR.

ricab:
OK got it, thanks.

PR #877, opened 2019-07-01
Cleanup vm state translation

townsend2010:


include/multipass/virtual_machine.h
```diff
@@ -33,15 +33,15 @@ class VirtualMachine
 public:
     enum class State
     {
+        unknown,
```
ricab: These are saved as numbers in the json instance db, so changing the order here will alter the meaning there. So, for instance, an instance that was saved as `suspended` would now be loaded as `suspending`, no?

src/rpc/multipass.proto
```diff
@@ -172,15 +172,15 @@ message MountInfo {
...
-        RUNNING = 0;
-        STARTING = 1;
-        RESTARTING = 2;
-        STOPPED = 3;
-        DELETED = 4;
-        DELAYED_SHUTDOWN = 5;
-        SUSPENDING = 6;
-        SUSPENDED = 7;
-        UNKNOWN = 8;
+        UNKNOWN = 0;
```
ricab: The order change also requires the client and daemon versions to match for the states to be interpreted coherently. I suppose that is alright, as long as this does not get into a bug release.

gerboland: Agreed this is slightly concerning, but since we typically tie the CLI closely to the server, it shouldn't be an issue. And we've not frozen our RPC yet.

But I do like this change as gRPC has a nasty habit of using the default initialiser in the case that the variable is not set. So having 0 be an error state is a good decision.

src/daemon/daemon.cpp
```diff
@@ -556,6 +556,30 @@ QStringList filter_unsupported_aliases(const QStringList& aliases, const std::st
...
+    case mp::VirtualMachine::State::running:
+        return mp::InstanceStatus::RUNNING;
+    case mp::VirtualMachine::State::delayed_shutdown:
+        return mp::InstanceStatus::DELAYED_SHUTDOWN;
+    case mp::VirtualMachine::State::suspending:
+        return mp::InstanceStatus::SUSPENDING;
+    case mp::VirtualMachine::State::suspended:
+        return mp::InstanceStatus::SUSPENDED;
+    case mp::VirtualMachine::State::unknown:
+    default:
```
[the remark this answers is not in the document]
townsend2010: I thought `fallthrough` is only needed when matching a `case` and then wanting to go to the next case, ie, no `break` for the matched case.  I've never had an issue with stacked cases like what I've done here, but maybe `clang` on mac is really finicky.

But CI on mac is busted atm, so I'll need to find some other way to try to compile this on mac...

include/multipass/virtual_machine.h
townsend2010: Oh, hmm, I think you're right.  I'd really like for `unknown` to be the first & default state in the enum, but if this breaks things, then I'm outta luck.  I'll try it to make sure though.

src/rpc/multipass.proto
townsend2010: Yeah, I'm not aware of any 3rd party multipass clients, so if anyone upgrades multipass, they'll get the new client at the same time.  Besides, as @gerboland points out, our RPC is not deemed stable yet, so we have the freedom to make changes like this :smile:

include/multipass/virtual_machine.h
townsend2010: Yep, you are definitely right, so we'll just live with this `enum` the way it is and can only add states at the end in order not to break anything.  That said...fixed.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
ricab: Ah alright, maybe it only complains when there are intermediate instructions. I tried to compile this locally w/ clang and it passed ok, but then I added said intermediate instructions and still went on fine, so I believe only the mac build detects this.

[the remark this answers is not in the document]
ricab: I just pushed #879 to test this

include/multipass/virtual_machine.h
ricab: Yeah another instance where epochs would come in handy... An alternative would be to detect the version (e.g. w/ some slight change in the json that we could recognize) and interpret the state differently, but that would mean adding historical complexity and I am not sure it would be worth it...

townsend2010: Yeah, definitely not worth it.

src/rpc/multipass.proto
ricab: Sure, I was mostly thinking aloud here... And then just pointing out this is a backward-incompatible change, so it means increasing the middle part of the semantic version.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
ricab: That experiment PR ended up not being enough for what we wanted, but @Saviq managed to build manually, so that resolves it.

townsend2010:
Any takers on approving/request changes/rejecting this?

PR #878, opened 2019-07-02
Transition Qemu snapshot checking to ProcessFactory and test it

gerboland:
Introduces MockProcessFactory, which creates mockable Processes. Use this to test Qemu snapshot checking.

This involved some refactoring:

   -  move all Qemu command line generation into the QemuVMProcessSpec class, and test it. This includes support for resuming an existing VM.
   - remove the "use_cdrom" hack and instead save all the Qemu command line arguments to the JSON file instead. Retain support for "use_cdrom" to ease transition. Tests for all transition cases.
   -  add a convenience API: ProcessFactory::create_process(command, arguments) so one doesn't need to create a Spec for every process.
   - this allowed me to remove an API wart in Process, namely the "extra_arguments" idea. This was confusing, and made testing difficult. Now once Process is constructed, its command, arguments, environment etc are fixed.


gerboland:
Mac fails on CI due to merge conflict

PR #879, opened 2019-07-02
Experiment mac build

ricab:
Try mac build of #877 in saner context


ricab:
For the record, as @Saviq explained elsewhere, we check out `+refs/pull/.../merge`, which is a merge commit into the base of the PR. We would have to make the base point at the old commit for this to work (we currently build all PRs in CI, independently of the target).

The experiment is now being done manually, so this PR can be thrown away.

PR #883, opened 2019-07-03
[snap] transition driver from `snap set` to `multipass set` (Fixes #866)

### LINKED ISSUE #866: [snap] migrate from `snap set driver…` to `multipass set local.driver…`
Now that #835 landed we still respect `$MULTIPASS_VM_DRIVER` to avoid breaking consumers.

We need to transition away from that, and I believe there's two parts to it:
1) `multipassd` to only respect the internal config
   - warn about `MULTIPASS_VM_DRIVER` being ignored, if set
   - and tell people to use `multipass set`
1) snap hooks to transition the setting from `snap set`
   - either the `refresh` or the `configure` hook (I believe it runs on refresh, too) move the value from `snapctl get driver` through to `multipass set`
   - and reset the snap value to an empty string

Outside of snaps, 1. above solves it all, I think.

Saviq:


snap/hooks/configure
```diff
@@ -9,38 +9,15 @@ fi
...
-    fi
-
-    driver_saved=$driver
-    needs_restart=true
+    echo "`snap set` is no longer used for driver selection, please use:"
+    echo
+    echo "multipass set local.driver=$( echo $driver | awk '{print tolower($0)}' )"
+    exit 1
+elif [[ $driver_saved != "" ]]; then
+    multipass set local.driver=$( echo $driver_saved | awk '{print tolower($0)}')
```
gerboland: I get a failure installing the snap from this line. Error:

    error: cannot perform the following tasks:
    - Run configure hook of "multipass" snap if present (run hook "configure": /snap/multipass/x3/meta/hooks/configure: line 18: multipass: command not found)

I guess the full path is necessary

snap/hooks/configure
```diff
@@ -9,38 +9,15 @@ fi
...
-        echo "All instances need to be stopped."
-        echo "Please save any data and stop the instances before proceeding:"
-        echo
-        echo "multipass stop --all"
-        exit 1
-    fi
-
-    driver_saved=$driver
-    needs_restart=true
+    echo "`snap set` is no longer used for driver selection, please use:"
```
gerboland: I get this output when I run `sudo snap set multipass driver=QEMU`

    error: cannot perform the following tasks:
    - Run configure hook of "multipass" snap (run hook "configure": 
    -----
    error: the required arguments `<snap>` and `<conf value> (at least 1 argument)` were not provided
     is no longer used for driver selection, please use:

    multipass set local.driver=qemu
    -----)

Looks like its executing "snap set" inside the backticks!

Saviq:
> Causes the daemon to restart anyway. Can you repro?

Yeah, expected as it is today - `Settings` doesn't check if it changes anything.

PR #884, opened 2019-07-04
[settings] Force INI format on all platforms

ricab:


ricab:
Thanks @gerboland, I don't think the settings were released yet. If I am not mistaken, 0.7 had the primary instance but without a configurable name at that point.

gerboland:
@ricab I wasn't sure of the timing to be honest, but double-checking I see you're right.
bors r+

PR #887, opened 2019-07-05
Portability fixes

ricab:
To make certain compilers happy.

### DIFF
--- a/src/platform/CMakeLists.txt
+++ b/src/platform/CMakeLists.txt
@@ -25,9 +25,7 @@ include_directories(platform
   backends)
 
 target_link_libraries(platform
-  update)
-
-target_link_libraries(platform
+  update
   utils)
 
 add_subdirectory(backends)
--- a/src/platform/CMakeLists.txt
+++ b/src/platform/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright © 2017-2018 Canonical Ltd.
+# Copyright © 2017-2019 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -11,8 +11,6 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-# Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
 
 add_library(platform STATIC
   platform_linux.cpp
@@ -28,6 +26,9 @@ include_directories(platform
 
 target_link_libraries(platform
   update)
+
+target_link_libraries(platform
+  utils)
 
 add_subdirectory(backends)
 add_subdirectory(client)
--- a/src/platform/platform_linux.cpp
+++ b/src/platform/platform_linux.cpp
@@ -19,7 +19,7 @@
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
-#include <multipass/settings.h>
+#include <multipass/utils.h>
 #include <multipass/virtual_machine_factory.h>
 
 #include "backends/libvirt/libvirt_virtual_machine_factory.h"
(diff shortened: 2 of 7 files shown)


ricab:
Need this on full

src/platform/CMakeLists.txt
```diff
@@ -29,6 +27,9 @@ include_directories(platform
 target_link_libraries(platform
   update)
 
+target_link_libraries(platform
+  utils)
```
gerboland: Why not add to the above target_link_libraries?

src/platform/CMakeLists.txt
```diff
@@ -29,6 +27,9 @@ include_directories(platform
 target_link_libraries(platform
   update)
 
+target_link_libraries(platform
```
townsend2010: Why not put this in the `target_link_libraries` above such that it looks like this:
```
target_link_libraries(platform
  update
  utils)
```

tests/mock_process_factory.cpp
```diff
@@ -32,7 +32,7 @@ std::unique_ptr<mp::Process> mpt::MockProcessFactory::create_process(std::unique
     return process;
 }
 
-mpt::MockProcess::MockProcess(std::unique_ptr<ProcessSpec>&& spec,
+mpt::MockProcess::MockProcess(std::unique_ptr<mp::ProcessSpec>&& spec,
```
gerboland: Ah sorry I missed that, I know vscpp is fussy about these

src/platform/platform_linux.cpp
```diff
@@ -76,7 +57,7 @@ bool mp::platform::is_backend_supported(const QString& backend)
 
 mp::VirtualMachineFactory::UPtr mp::platform::vm_backend(const mp::Path& data_dir)
 {
-    const auto& driver = get_driver_str();
+    const auto& driver = utils::get_driver_str();
```
townsend2010: Definitely not going to block on it, but I prefer putting the `mp` namespace before any `utils` calls.

src/platform/CMakeLists.txt
ricab: Not sure anymore to be honest, at first I added it to the first `target_link_libraries` but that created a conflict, so I separated it. But it could have gone there, let me do that.

src/platform/CMakeLists.txt
ricab: OK, see reply to Gerry

src/platform/platform_linux.cpp
ricab: Why?

townsend2010: Just seems clearer to me where is comes from.  But like I said, I'm not going to block or anything if you don't want to change it.  It's just a preference of mine.

src/platform/CMakeLists.txt
townsend2010: lol, didn't realize @gerboland was reviewing as well.  I'll just resolve this one.

src/platform/platform_linux.cpp
ricab: OK, I don't quite see it that way. We're in the multipass namespace, it is only natural to assume it (and that is what the compiler does too - I believe it is an msvc bug that it does not do that for parameters and trailing returns). If we were to prefix every multipass function call with a `mp::` namespace the code would be much noisier to read.

gerboland: I'm of @ricab 's disposition here, but I'm also ok with defining "mpu = mp::utils" and using that.

townsend2010: Sure, that's fine.  I just remember when we were evaluating a 3rd party library (I can't remember what it was now), it also had a `utils` namespace, so there was a collision.  I found that `utils` is a fairly common namespace for other libraries, and so I went with the `mp::utils` convention.  But we aren't using anything like that now, so meh.  This was only meant as a suggestion since it's like that in most places in the code and didn't mean for it turn into a longer conversation :smiley:

src/utils/settings.cpp
```diff
@@ -74,9 +74,13 @@ std::unique_ptr<QSettings> persistent_settings(const QString& key)
 
 bool exists_but_unreadable(const QString& filename)
 {
-    QFile file(filename);
-    return !file.open(QIODevice::ReadOnly) && errno != ENOENT; /* note file.error() not enough for us: it would not
-                                                                  distinguish the cause of failure */
+    std::ifstream stream;
+    stream.open(filename.toStdString(), std::ios_base::in);
+    return stream.fail() && errno && errno != ENOENT; /*
+        Note: file.error() is not enough for us: it would not distinguish the cause of failure;
```
townsend2010: I'm not sure what `file.error()` is referring to in these comments.

ricab: Indeed, `file` went away now, will have to update it, thanks.

ricab: Done, I used QFile::error - although `QFile` isn't used there anymore, `QFile::error` would be a suitable alternative if it gave us the same info as `QFile::errorString`. If you prefer I can remove the comment entirely.

townsend2010: I think leaving the comment for now is fine for historical purposes.

PR #905, opened 2019-07-10
snap: Check for an empty 'driver' string first before comparing

### LINKED ISSUE #904: [snap] Refreshing to latest edge version causes the configure hook to fail
Here is the output:
```
error: cannot perform the following tasks:
- Run configure hook of "multipass" snap if present (run hook "configure": 
-----
`snap set` is no longer used for driver selection, please use:

sudo multipass set local.driver=
-----)
```

townsend2010:
If there is no driver set, configure will always fail.

Fixes #904

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -10,7 +10,7 @@ fi
 
 driver="$(snapctl get driver)"
 
-if [[ $driver != $driver_saved ]]; then
+if [[ -n $driver && $driver != $driver_saved ]]; then
     echo -e '`snap set` is no longer used for driver selection, please use:\n'
     echo "sudo multipass set local.driver=$( echo $driver | awk '{print tolower($0)}' )"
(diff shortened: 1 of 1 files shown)


ricab:
Giving it a little more thought, perhaps we could have only a warning in the configure hook plus a check in the multipass clients, bailing with an error if the configuration got spotted...

townsend2010:
@ricab,

If `driver` is set and the snap refreshes, the `driver` string won't be different than the `driver_saved` string, so the logic will go to the `elif`.

With this change the only way to get into the error is if someone is trying to do a `snap set multipass driver=blah`, not during a refresh when it's impossible to set the driver to a different value.

ricab:
@townsend2010 right, of course. Good to go then.

bors r+

Saviq:
Thanks :)

Issue #772, opened 2019-05-08
long output to exec hangs


knitzsche:
With multipass  0.6.1    752  beta      canonical✓  classic

A command with very long output hangs, such as
$ multipass exec vm -- dpkg -l

This gets part way through displaying the long output and then the terminal becomes unresponsive (until stopping the vm).

Saviq:
Hmm I can't reproduce that @knitzsche, could you please try on the `edge` channel?

knitzsche:
@Saviq I tried it with multipass from edge, and it also happens (intermittently, but most times). This time I tried with a very simple snap: I ran snapcraft init, then modified the snapcraft.yaml to have base: core18. then about 2 times in 3 the command result does not fully show and the terminal because unresponsive until I delete the vm (with snapcraft clean or multipass delete)

Saviq:
@knitzsche so you're seeing this through `snapcraft`? Care to share the `yaml`?

knitzsche:
host is xenial. snapcraft.yaml is:
`name: test-multipass-issue # you probably want to 'snapcraft register <name>'
base: core16 # the base snap is the execution environment for this snap
version: '0.1' # just for humans, typically '1.2+git' or '1.3.2'
summary: Single-line elevator pitch for your amazing snap # 79 char long summary
base: core16
description: |
  This is my-snap's description. You have a paragraph or two to tell the
  most important story about your snap. Keep it under 100 words though,
  we live in tweetspace and your description wants to look good in the snap
  store.

grade: devel # must be 'stable' to release into candidate/stable channels
confinement: devmode # use 'strict' once you have the right plugs and slots

parts:
  my-part:
    # See 'snapcraft plugins'
    plugin: nil`

Saviq:
It just snapped fine here…
```plain
snapcraft
Using 'snapcraft.yaml': Project assets will be searched for from the 'snap' directory.
Launching a VM.
Launched: snapcraft-test-multipass-issue                                        
2019-05-09T21:26:11Z INFO Waiting for restart...
core 16-2.38.1 from Canonical✓ installed
snapcraft 3.4.1 from Canonical✓ installed
core16 16-20190110 from Canonical✓ installed
Hit:1 http://archive.ubuntu.com/ubuntu xenial InRelease
Hit:2 http://archive.ubuntu.com/ubuntu xenial-updates InRelease                                      
Hit:3 http://archive.ubuntu.com/ubuntu xenial-backports InRelease                                    
[...]

PR #906, opened 2019-07-11
libssh: upgrading to the latest from their stable-0.9 branch

davidkrauser:
The current hypothesis for the cause of issue #772 is a bug in libssh. It manifests as a hung ssh connection, even after the process spawned by ssh has finished. We get stuck in this loop:

```
while (ssh_channel_is_open(channel.get()) && !ssh_channel_is_eof(channel.get()))
{
  // ...
}
```

We could get stuck there if the ssh channel mishandles an EOF. Looking at the changes to libssh from what we have to the latest stable-0.9 branch, there are at least a few directly related to EOF handling:

```
multipass/3rd-party/libssh/libssh $ git log e42a423a247c8c9f0bd58d32a8b80c522f7b43f6..origin/stable-0.9 --oneline | grep EOF
95ab5f0d channel: Don't send EOF on channel more than once
9340a0af connector: Stop socket-to-channel EOF flooding
```

There have been lots of other bug fixes and improvements in the stable-0.9 branch that are probably worth bringing in. Since upgrading, I have been unable to reproduce #772.

### DIFF
--- a/3rd-party/libssh/CMakeLists.txt
+++ b/3rd-party/libssh/CMakeLists.txt
@@ -33,6 +33,17 @@ set(LIBSSH_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/libssh/include)
 set(LIB_INSTALL_DIR lib)
 set(BIN_INSTALL_DIR bin)
 
+# Since the main CMake file is bypassed, we have to specifically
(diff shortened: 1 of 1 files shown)


davidkrauser:
After more testing, I was able to reproduce the issue, even with the latest from their stable-0.9 branch. Going to close for now while I do more digging.

PR #912, opened 2019-07-11
[cli] Improve `set` argument format error

### LINKED ISSUE #908: `set` syntax-error message should be improved
As @gerboland pointed out:

> `multipass set local.driver hyperkit` printing "Need exactly one key-value pair." - message unclear it should be key=value.

ricab:
Fixes #908.


ricab:
@gerboland I adapted the msg rather than replacing it because the code just checks the number of arguments and so this would be output also on something like `multipass set k1=v1 k2=v2` (I don't think it merits changing to look for `=` in the arguments)

Issue #907, opened 2019-07-11
`set` command assumes "libvirt" is valid before `Settings` has a chance to check it


ricab:
On platforms that do not support libvirt, we still get

```
sudo multipass set local.driver=libvirt
All instances need to be stopped.
Please save any data and stop the instances before proceeding:

multipass stop --all
```

Need to add check for option validity

PR #914, opened 2019-07-12
Fix set tests

ricab:
This fixes new failures when moving to other platforms, on tests that rely on the qemu->libvirt switch specifics. It basically bypasses those tests when not applicable, changing names accordingly. It also adds a grpc-related test to complement the previous one.

### DIFF
--- a/tests/test_cli_client.cpp
+++ b/tests/test_cli_client.cpp
@@ -1451,10 +1451,14 @@ TEST_F(Client, set_cmd_falls_through_instances_when_another_driver)
     aux_set_cmd_rejects_bad_val(mp::driver_key, "other");
 }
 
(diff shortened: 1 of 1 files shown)


ricab:
These tests failed because they assumed that providing a "libvirt" driver was enough to trigger the behavior of checking for running instance. This is no longer the case: on other platforms, we don't want to do this check if the user says `multipass set local.driver=libvirt` by mistake, because that is not a valid driver (see #907).

For instance, the test previously named `set_cmd_fails_when_grpc_problem` expected list to be called on the mock daemon, but after #910, that does not happen when on other platforms.

ricab:
After discussing with @townsend2010, it was decided to go in favor or disabling the tests entirely with `#ifdef`: #915.

PR #926, opened 2019-07-19
libssh: Add patch to increase connector buffer size to 64K

### LINKED ISSUE #772: long output to exec hangs

townsend2010:
This is a temporary workaround and more work in libssh is needed in order to fix
this properly.

Fixes #772

### DIFF
--- a/3rd-party/libssh/CMakeLists.txt
+++ b/3rd-party/libssh/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright © 2017 Canonical Ltd.
+# Copyright © 2017-2019 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


Saviq:
Hmm I'm not sure I like this more than going back to our own branch… This leaves the source tree dirty, has the CMake hackery to know if it's been applied already… I don't think the maintenance of this will actually be less than just our own branch that we rebase on…

Sorry I didn't realize this when we first spoke.

Saviq:
Made this into a proper submodule change.

PR #935, opened 2019-07-29
Send metrics to correct endpoint.

cmars:
api.staging.jujucharms.com is a test deployment and has no guarantee of
reliability or data retention. This updates multipass to send KPI
metrics to the correct deployment.


Saviq:
Thanks :)

bors r+

Issue #936, opened 2019-07-30
Mounts don't appear to be properly restored on host reboot


OddBloke:
I just rebooted my machine, and when I `multipass exec my-vm /bin/bash`'d my way in to a multipass VM (`my-vm`), a mountpoint (`/my/mountpoint`) that I had configured previously was just an empty directory. To remedy this, I figured I would just perform the mount again:

```
$ multipass mount /the/same/local/path my-vm:/my/mountpoint
mount failed: The following errors occurred:                                    
There is already a mount defined for "my-vm:/my/mountpoint"
```

After this, `/my/mountpoint` had the contents I would expect, and I was able to proceed on with my work.

OddBloke:
I've seen this a couple of times now (I don't use multipass every time I boot my machine, so I can't be sure that it's consistent, though), so if there's any extra debugging info I could provide that would be helpful, let me know and I'll try to grab it next time I see this occur.

townsend2010:
The mount handling has undergone some major changes in #2835 and I believe when that PR is merged, this issue will be fixed.

PR #937, opened 2019-07-31
 sshfs_mount: Recover mounts when sshfs exits for unexpected reasons 

### LINKED ISSUE #509: [mount] multipassd fails to notice sshfs going away inside the instance
Steps:
- create a mount
- kill sshfs inside the instance
- `multipass info --all`

Expected:
- an error logged in the multipassd log
- restarting sshfs

Current:
- multipassd thinks the mount is still going

townsend2010:
Fixes #509

### DIFF
--- a/include/multipass/sshfs_mount/sftp_server.h
+++ b/include/multipass/sshfs_mount/sftp_server.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


Saviq:
Does that relate to #936 do you think?

townsend2010:
@Saviq 
> Does that relate to #936 do you think?

It _might_...

townsend2010:
@gerboland,

I'm not sure where you think I'm polling for the `sshfs` process inside the VM on each `sshfs` event.  I only check for the `sshfs` process if `sftp_get_client_message` returns a `NULL` client message, indicating that something bad happened to the connection.

gerboland:
> @gerboland,
> 
> I'm not sure where you think I'm polling for the `sshfs` process inside the VM on each `sshfs` event. I only check for the `sshfs` process if `sftp_get_client_message` returns a `NULL` client message, indicating that something bad happened to the connection.

Ah ok, I hadn't noticed that null check was so important.

townsend2010:
> Ah ok, I hadn't noticed that null check was so important.

Ok, good, I hoped I wasn't missing something :smile:

townsend2010:
@gerboland,

I'm not sure of the exact reasons why `sshfs` goes away in the instance, but I suspect it's etiher crashing or erroring out.

The archive version of `libfuse` and `sshfs` are quite old, so I have another project at https://github.com/CanonicalLtd/multipass-sshfs that builds a snap package using upstream `libfuse` and `sshfs`.  This works quite well and should resolve many crashing/error issues.  I'm just waiting for the `bare` base to be promoted to stable and then support for that to be merged into snapcraft.  That way, the whole `core18` snap will not need to be downloaded and installed when installing the `multipass-sshfs` snap.

PR #942, opened 2019-08-05
[test] Replace literal with constant in test

ricab:
Just a little tweek to keep it tidy &ndash; I failed to use the constant here before.


Saviq:
bors r=gerboland

PR #950, opened 2019-08-09
daemon: Use 'ubuntu' user instead of 'multipass' user

### LINKED ISSUE #949: Replace 'multipass' user with 'ubuntu' user
Instead of using the default `multipass` user, use an `ubuntu` user instead.  This is a temporary solution until we can implement #929 properly.

townsend2010:
Fixes #949


Saviq:
> LGTM. `mount` behaved as if it mapped my file ownership automatically, although I suspect that is because I have the same uid in the host as the ubuntu user in the vm (1000).

We actually add a default mapping of `$UID:1000, $GID:1000`, you can see it in `multipass info …`

ricab:
@Saviq indeed, thanks for clearing it up

PR #959, opened 2019-08-12
Process - add API to fetch process state and close_write_channel

gerboland:
This renames ProcessExitState with ProcessState, and allows user to fetch process state from Process at any stage.

And quoting myself:
> The wait_for functions return false if the process exits unexpectedly, or the timeout hits. 

Not always true! wait_for_finished will return true if process crashes. wait_for_started return true after forking, so if the program doesn't exist, it doesn't notice.

### DIFF
--- a/tests/mock_process.cpp
+++ b/tests/mock_process.cpp
@@ -1,4 +1,3 @@
-#include <cassert>
 #include <cstdlib>
 #include <iostream>
 
--- a/tests/mock_process.cpp
+++ b/tests/mock_process.cpp
@@ -1,8 +1,31 @@
+#include <cassert>
 #include <cstdlib>
 #include <iostream>
 
-int main(int /*argc*/, char* argv[])
+int main(int argc, char* argv[])
 {
-    // deliberately will crash if argument 1 missing
+    if (argc == 1)
+    {
+        // deliberately will crash if argument 1 missing
+        assert(false);
+    }
+    // Exit immediately if only 1 argument
+    else if (argc == 2)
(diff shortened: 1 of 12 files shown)


gerboland:
Thanks for the feedback @Saviq and @ricab  - I've added a simpler API instead. Please take a look.

tests/mock_process.cpp
```diff
@@ -1,8 +1,31 @@
...
 #include <iostream>
 
-int main(int /*argc*/, char* argv[])
+int main(int argc, char* argv[])
 {
-    // deliberately will crash if argument 1 missing
+    if (argc == 1)
+    {
+        // deliberately will crash if argument 1 missing
+        assert(false);
```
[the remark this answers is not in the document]
gerboland: Oh ofc, duhh. Thanks!

tests/mock_process.cpp
```diff
@@ -1,8 +1,31 @@
+#include <cassert>
```
ricab: Dangling include

gerboland: fixed

Saviq:
bors r=ricab

Saviq:
Harumph…

bors r=ricab

gerboland:
Trying once more, bors landed something a few hours ago successfully
bors r=ricab

Saviq:
I just restarted the standard PR run and it went fine…

bors r=ricab

If this fails, I'll have to look tomorrow into what's really going on…

PR #974, opened 2019-08-15
 dnsmasq_server: Better handle when dnsmasq does not run or dies

### LINKED ISSUE #325: Daemon does not react to `dnsmasq` dying
When `named` is installed, it seems to jump onto all interfaces on port 53 to serve DNS queries. Our `dnsmasq` dies in this situation, failing to bind on port 53 on `mpqpbr0`:

```
Aug 02 12:22:10 foo-host dnsmasq[15529]: failed to create listening socket for 10.227.59.1: Address already in use
Aug 02 12:22:10 foo-host dnsmasq[15529]: FAILED to start up
```

Instances then lack DHCP and fail to start. We need to do something smarter in this case.

townsend2010:
Fixes #325

### DIFF
--- a/src/platform/backends/qemu/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/dnsmasq_server.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Canonical, Ltd.
+ * Copyright (C) 2018-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -25,40 +25,63 @@
 
 #include <multipass/format.h>
 
+#include <QDir>
+
 #include <fstream>
+
+#include <signal.h>
+#include <sys/types.h>
+
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
 namespace
 {
-auto make_dnsmasq_process(const mp::Path& data_dir, const QString& bridge_name, const mp::IPAddress& bridge_addr,
-                          const mp::IPAddress& start, const mp::IPAddress& end)
+auto make_dnsmasq_process(const mp::Path& data_dir, const QString& bridge_name, const QString& pid_file_path,
+                          const std::string& subnet)
 {
-    auto process_spec = std::make_unique<mp::DNSMasqProcessSpec>(data_dir, bridge_name, bridge_addr, start, end);
+    auto process_spec = std::make_unique<mp::DNSMasqProcessSpec>(data_dir, bridge_name, pid_file_path, subnet);
     return mp::ProcessFactory::instance().create_process(std::move(process_spec));
+}
+
+auto get_dnsmasq_pid(const mp::Path& pid_file_path)
+{
+    std::ifstream pid_file{pid_file_path.toStdString()};
+    std::string pid;
+
+    getline(pid_file, pid);
+
+    return std::stoi(pid);
 }
 } // namespace
 
-mp::DNSMasqServer::DNSMasqServer(const Path& data_dir, const QString& bridge_name, const IPAddress& bridge_addr,
-                                 const IPAddress& start, const IPAddress& end)
+mp::DNSMasqServer::DNSMasqServer(const Path& data_dir, const QString& bridge_name, const std::string& subnet)
     : data_dir{data_dir},
-      dnsmasq_cmd{make_dnsmasq_process(data_dir, bridge_name, bridge_addr, start, end)},
-      bridge_name{bridge_name}
+      bridge_name{bridge_name},
+      pid_file_path{QDir(data_dir).filePath("dnsmasq.pid")},
+      subnet{subnet}
 {
-    dnsmasq_cmd->start();
+    try
+    {
+        check_dnsmasq_running();
+    }
+    catch (const std::exception&)
+    {
+        start_dnsmasq();
+    }
 }
 
 mp::DNSMasqServer::~DNSMasqServer()
 {
-    dnsmasq_cmd->kill();
-    dnsmasq_cmd->wait_for_finished();
+    auto dnsmasq_pid = get_dnsmasq_pid(pid_file_path);
+    kill(dnsmasq_pid, SIGKILL);
 }
 
 mp::optional<mp::IPAddress> mp::DNSMasqServer::get_ip_for(const std::string& hw_addr)
 {
     // DNSMasq leases entries consist of:
     // <lease expiration> <mac addr> <ipv4> <name> * * *
-    const auto path = data_dir.filePath("dnsmasq.leases").toStdString();
+    const auto path = QDir(data_dir).filePath("dnsmasq.leases").toStdString();
     const std::string delimiter{" "};
     const int hw_addr_idx{1};
     const int ipv4_idx{2};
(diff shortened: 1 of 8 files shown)


src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -105,3 +135,21 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr)
...
+void mp::DNSMasqServer::check_dnsmasq_running()
+{
+    auto dnsmasq_pid = get_dnsmasq_pid(pid_file_path);
+    if (kill(dnsmasq_pid, 0) == 0)
+        return;
+
+    // exit_code == 2 signifies dnsmasq network-related error. See `man dnsmasq`.
+    throw std::runtime_error(
+        fmt::format("Multipass dnsmasq is not running.{}",
+                    (dnsmasq_exit_state.exit_code == 2) ? " Ensure nothing is using port 53." : ""));
```
gerboland: Unlikely, but if dnsmasq crashed, dnsmasq_exit_state.exit_code will not be set, so it will throw too

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -277,6 +274,17 @@ void mp::QemuVirtualMachineFactory::hypervisor_health_check()
...
+    try
+    {
+        dnsmasq_server.check_dnsmasq_running();
+    }
+    catch (const std::exception&)
+    {
+        // Try starting dnsmasq if it's not running
+        dnsmasq_server.start_dnsmasq();
+        dnsmasq_server.check_dnsmasq_running();
+    }
```
gerboland: Do we need this here? This removes the RAII-ness of DNSMasqServer, where if it can be created, we can trust dnsmasq is running. 

Could not DNSMasqServer's constructor do this, and throw if it cannot create the process? And then these methods can be made private.

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -105,3 +135,21 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr)
...
+}
+
+void mp::DNSMasqServer::start_dnsmasq()
+{
+    dnsmasq_cmd = make_dnsmasq_process(data_dir, bridge_name, pid_file_path, subnet);
+    dnsmasq_exit_state = dnsmasq_cmd->execute();
```
gerboland: Nitpick: `dnsmasq_exit_state` contains the return code of the now daemonized dnsmasq process - could you rename to "dnsmasq_daemon_fork_state" or something like that?

You're not handling cases if dnsmasq binary isn't found. Wouldn't hurt...

townsend2010:
@gerboland,

Apologies for not explaining what I was intending here.  For posterity's sake, I'll write a brief synopsis of what we discussed in the Hangout.

I agree, I'd rather have kept this under the control of `QProcess`, but I ran into a big issue with this.


When there are no running instances, a `launch` or `start` needs to check whether `dnsmasq` is up and running as it can fail for a few reasons, one being the reason given in #325.  If `dnsmasq` did fail on start, we want to notify the user of this, so the check should try to start `dnsmasq` and see if it fails.

If we use `QProcess` and run `dnsmasq` in the foreground, the only way to know if it fails is to connect a slot to one of the `QProcess` signals such as `stateChanged(QProcess::ProcessState newState)` or `readyReadStandardError()` and do something there.  But the pitfall with this is that we need to know if `dnsmasq` starts or not synchronously and using the signal/slot method does not allow for that.

The only way I found to reliably know this in a synchronous manner is to start `dnsmasq` as a daemon and get the return code from the parent process that forks `dnsmasq`. Since it can detect the error when `dnsmasq` starts and return an exit code that clues us into what may have happened, we wait on it to return and that time, we can bubble up a (hopefully) meaningful error to the client and user.

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -25,40 +25,70 @@
...
-                          const mp::IPAddress& start, const mp::IPAddress& end)
+auto make_dnsmasq_process(const mp::Path& data_dir, const QString& bridge_name, const QString& pid_file_path,
+                          const std::string& subnet)
 {
-    auto process_spec = std::make_unique<mp::DNSMasqProcessSpec>(data_dir, bridge_name, bridge_addr, start, end);
+    auto process_spec = std::make_unique<mp::DNSMasqProcessSpec>(data_dir, bridge_name, pid_file_path, subnet);
     return mp::ProcessFactory::instance().create_process(std::move(process_spec));
 }
+
+auto get_dnsmasq_pid(const mp::Path& pid_file_path)
```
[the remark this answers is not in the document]
townsend2010: Sure, I'll add a comment saying `std::stoi` will throw when the string is not valid for conversion.

Regarding returning a negative `int` there, I'm not really sure how `dnsmasq` would ever put a negative pid in that file, but it's trivial to cast it to an `unsigned int`.

src/platform/backends/qemu/dnsmasq_server.cpp
```diff
@@ -105,3 +135,21 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr)
 
     dhcp_release.waitForFinished();
 }
+
+void mp::DNSMasqServer::check_dnsmasq_running()
+{
+    auto dnsmasq_pid = get_dnsmasq_pid(pid_file_path);
```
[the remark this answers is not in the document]
townsend2010: I don't think so as we want the throw from `std::stoi` to bubble up to the caller of `check_dnsmasq_running()`.

src/platform/backends/qemu/dnsmasq_server.cpp
townsend2010: Isn't that what we want?  I'm not I follow what you mean here.  We want to throw if `dnsmasq` is not running for whatever reason.

src/platform/backends/qemu/dnsmasq_server.cpp
townsend2010: Sure, I can fix up both of those :grin:

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
townsend2010: `DNSMasqServer` is currently and has been a non-pointer class member, so RAII doesn't really apply to that.  That said, we could certainly move it to be a `std::unique_ptr` and only need to check if it's a `nullptr` or not, and if it is, then try to initialize it within a `try`/`catch` block.  I'll look into that.

src/platform/backends/qemu/dnsmasq_server.cpp
gerboland: Ok. This is what I dislike about exceptions, it's very hard to know what error handling was planned when reading the code.

gerboland:
@townsend2010 thanks for your explanation, I now understand this better. I don't have a large objection to the change

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
townsend2010: So I thought about this some more and I don't think assuming that if the `DNSMasqServer` ctor doesn't throw, we can trust `dnsmasq` is running is enough.  What happens if `dnsmasq` dies after construction?  We need some way to query that `dnsmasq` is alive which is the reason for the `check_dnsmasq_running()` method.  And then we need a method for trying to start it if it's not running.  I suppose I could wrap that up in a single method, but I think it's better to leave `DNSMasqServer` as a non-pointer object and use `check_dnsmasq_running()` for querying when needed.

gerboland: Before you explained why you daemonized dnsmasq, I had thought we could reliably manage the dnsmasq process with this DNSMasqServer class. With the way you're doing it here, you can't really.

src/platform/backends/qemu/dnsmasq_server.cpp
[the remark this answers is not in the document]
townsend2010: Ok, done.

src/platform/backends/qemu/dnsmasq_server.cpp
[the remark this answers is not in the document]
townsend2010: I redid this, so now it does indeed have a `try`/`catch` around this.

src/platform/backends/qemu/dnsmasq_server.cpp
townsend2010: I redid this a bit, but it still throws here, so I think your dislike still applies.

src/platform/backends/qemu/dnsmasq_server.cpp
townsend2010: Fixed both of these nits.

townsend2010:
@gerboland,

I made some changes based on your feedback and it's ready for re-review.

PR #978, opened 2019-08-16
gui: Put the petenv instance at the top of the tray icon menu

### LINKED ISSUE #966: Reshuffle system tray for primary instance
Add primary instance handling in the systray.

townsend2010:
The petenv instance gets its own section at the top of the tray icon menu.

Fixes #966

---
I followed the design spec found at https://github.com/CanonicalLtd/multipass-design#status-menu as closely as I could.  A couple of things though.
- Currently, there is no clean way to determine if the petenv instance is using the default name or another name, so I always put the petenv name in the `Start` line whereas the spec (I think) specifies only putting the alternate petenv name and don't put a name if it's the default.
- We currently have no way to determine if `suspend` is available in the backend, so I always use `Stop` for now.

### DIFF
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -21,6 +21,8 @@
 #include <multipass/cli/client_common.h>
 #include <multipass/cli/client_platform.h>
 #include <multipass/cli/format_utils.h>
+#include <multipass/format.h>
+#include <multipass/settings.h>
 #include <multipass/version.h>
 
 #include <QCoreApplication>
@@ -51,6 +53,17 @@ void cmd::GuiCmd::create_actions()
 {
     about_separator = tray_icon_menu.addSeparator();
     quit_action = tray_icon_menu.addAction("Quit");
+
+    QObject::connect(&petenv_shell_action, &QAction::triggered,
+                     [] { mp::cli::platform::open_multipass_shell(QString()); });
+    QObject::connect(&petenv_stop_action, &QAction::triggered, [this] {
+        future_synchronizer.addFuture(
+            QtConcurrent::run(this, &GuiCmd::stop_instance_for, Settings::instance().get(petenv_key).toStdString()));
+    });
+    QObject::connect(&petenv_start_action, &QAction::triggered, [this] {
+        future_synchronizer.addFuture(
+            QtConcurrent::run(this, &GuiCmd::start_instance_for, Settings::instance().get(petenv_key).toStdString()));
+    });
 }
 
 void cmd::GuiCmd::update_menu()
@@ -58,6 +71,8 @@ void cmd::GuiCmd::update_menu()
     std::vector<std::string> instances_to_remove;
 
     auto reply = list_future.result();
+
+    handle_petenv_instance(reply.instances());
 
     for (auto it = instances_entries.cbegin(); it != instances_entries.cend(); ++it)
     {
@@ -78,6 +93,9 @@ void cmd::GuiCmd::update_menu()
     for (const auto& instance : reply.instances())
     {
         auto name = instance.name();
+        if (name == Settings::instance().get(petenv_key).toStdString())
+            continue; // Already handled
+
         auto state = instance.instance_status();
         auto state_string = QString::fromStdString(mp::format::status_string_for(state));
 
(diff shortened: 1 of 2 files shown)


src/client/gui/gui_cmd.cpp
```diff
@@ -255,6 +273,71 @@ void cmd::GuiCmd::set_menu_actions_for(const std::string& instance_name, const m
...
+    {
+        petenv_stop_action.setEnabled(true);
+        petenv_start_action.setEnabled(false);
+    }
+    else if (state.status() == InstanceStatus::STOPPED || state.status() == InstanceStatus::SUSPENDED)
+    {
+
+        petenv_start_action.setEnabled(true);
+        petenv_stop_action.setEnabled(false);
+    }
```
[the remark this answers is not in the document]
townsend2010: Hmm, yeah, I should take some other states into account here.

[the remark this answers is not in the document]
townsend2010: Ok, I updated this.  Note that the `DELETED` state is handled in `handle_petenv_instance()` since we don't want to show deleted instances.

ricab:
@townsend2010 If you'd prefer to get this in faster and tackle some of the aspects above as separate issues, that would be fine by me. I think the first two derive directly from this PR and would probably make sense to tackle together, but it's up to you. Let me know how you prefer to proceed.

townsend2010:
@ricab,

Thanks for reviewing!  I'll take these one by one :wink: 

> Instance list grows one empty item every second

I'll look into that and fix that.

> Instance dangles in list when the pet env name is updated

Yep, another thing that needs fixing.

> stop enabled on starting instance

Well, with the async work I did before, it should be possible to stop an instance while it's starting.  I don't think it's completely waiting to finish starting since you see `shutdown called while starting` which indicates the start up was aborted by a shutdown.  That said, this is really a design decision on how this should be handled and we should defer this.

> Start/Open Shell not shown when pet env absent

Right, we do eventually want that, but this will require more work, especially when the image needs to be downloaded and getting that progress into the gui.  Let's defer this for later.

> Open shell on deleted

This should be fixed.  I'll look into it.

> Whether to show petenv name after start

Right, if I read the design spec correctly, _only_ a name other than the default name of `primary` should be shown in quotes.  I guess we can do a prerequisite PR to change getting the settings defaults public and then I can make the change here to match the design.

> Other points (for separate PRs)

Yep, there are a number of issues I've seen when doing this and plan on doing a big update soon-ish to clean those issues up along with what you point out.

townsend2010:
> Start/Open Shell not shown when pet env absent

After some further discussion on IRC, I'll enable `Open Shell` for a non-existent petenv and other functions will be disabled.

ricab:
@townsend2010 *Yes to all*, except

> > stop enabled on starting instance
>
> [...] this is really a design decision [...]

I agree, and indeed the design is not very detailed about this. I think the relevant bit is:

> *For both primary and non-primary instances, “Start” should be enabled when the instance isn’t running, and “Suspend” when it is.*

So, strictly speaking, it follows that "stop" ("suspend" really<sup>1</sup>) wouldn't be available while starting (only when running). OTH, a strict interpretation would make "start" available when starting, which I don't think is intentional. So I agree clarification is needed.

But meanwhile, notice that the current implementation of the GUI does not offer "stop" for starting "regular" instances. It is only this PR that allows it (on the petenv). In my reproduce steps I used different clients, but it could also be all done in the GUI: click "start" and then "stop" while starting.

I think it should replicate what is done for ordinary instances. That would more coherent, both internally and with the CLI, which does not offer "stop" for a starting vm in the same client instance (it blocks).
<br/>
<sup>1</sup> Another separate issue

townsend2010:
@ricab,

I had planned on changing the regular instances way of handling the start/stop stuff, but after reading your comments and reconsidering, I'll change the petenv behavior as you describe for now.

townsend2010:
Ok, this is all updated to address most of the feedback.

src/client/gui/gui_cmd.cpp
```diff
@@ -255,6 +278,72 @@ void cmd::GuiCmd::set_menu_actions_for(const std::string& instance_name, const m
...
+    else if (std::find(can_start_states.begin(), can_start_states.end(), state.status()) != can_start_states.end())
+    {
+        petenv_shell_action.setEnabled(true);
+        petenv_start_action.setEnabled(true);
+        petenv_stop_action.setEnabled(false);
+    }
+    else
+    {
+        if (state.status() == InstanceStatus::DELETED || state.status() == InstanceStatus::SUSPENDING)
```
ricab: Why SUSPENDING?

townsend2010: Why would you "shell" into an instance that is suspending?  The instance is not in a state ready to handle shell, so I don't think "Open Shell" should be available.

src/client/gui/gui_cmd.cpp
```diff
@@ -90,12 +94,10 @@ void cmd::GuiCmd::update_menu()
         instances_entries.erase(instance);
     }
 
+    const auto petenv_name = Settings::instance().get(petenv_key).toStdString();
```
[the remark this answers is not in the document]
townsend2010: I can see about cutting down the number of calls, but really, I want to see if the `primary-name` has changed each time we poll, so `create_actions` is not the right place.  But I should only call it once per poll, so I'll look into that.

townsend2010:
> start disabled on absent primary

Right, as discussed on irc with @Saviq, only `Open Shell` should be available.  The reason being is if the image is not downloaded yet and the user chooses `Start`, then they currently don't have any feedback about the image downloading and being prepared in the gui.  `Open Shell` will provide the feedback in the terminal that it opens though.  If/when we have proper `launch` support in the gui, then `Start` can be made available.

> all actions disabled after primary purged

Ok, I'll look into that.

> petenv not updated, primary listed twice

Ok, I thought I had this working.  I guess not :frowning_face:

src/client/gui/gui_cmd.cpp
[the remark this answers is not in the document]
townsend2010: Ok, I made a class attribute that keeps track of the current petenv name and gets updated when it changes, so now, getting the setting is only called once per poll and is the only place it is called.

townsend2010:
@ricab,

I appreciate the testing and feedback.  I've addressed everything here except the "start disabled on absent primary" and the `SUSPENDING` state you pointed out.

Thanks again!

src/client/gui/gui_cmd.cpp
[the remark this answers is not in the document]
ricab: Perfect, thanks.

src/client/gui/gui_cmd.cpp
ricab: Indeed, but that is also true of `STOPPING`, right? And `UNKNOWN`, which is handled separately above. 

`stop` and `start` cover complement sets of the possible states (they are mutually exclusive and, together, they cover all of them). This is not true of `shell` (it covers some states from each).So I think the code would be cleaner if the handling of `petenv_shell_action` got separated, with its own `can_shell_states` + if-else. WDYT?

ricab:
@townsend2010 Thanks for the updates

I had another look and it looks good, but I haven't tested yet. I will try to do that later today. Meanwhile, I added some feedback on `SUSPENDING`. Let me know what you think. I would then test when that was in too.

> [...] only Open Shell should be available.

OK

src/client/gui/gui_cmd.cpp
ricab: ... or maybe `cant_shell_states` (would be shorter).

townsend2010: Well, there is no `STOPPING` state, although there should be, but I get what you mean here.  I have thought of a different way to handle this, so I'll push that soon.

townsend2010: Ok, I updated this section to hopefully be more clear and better capable of handling different states and the action combinations.

PR #986, opened 2019-08-21
Replace driver key literals with constant

ricab:
Little PR to keep it tidy


townsend2010:
Ok #987 is now merged.

bors r+

PR #987, opened 2019-08-21
snap: Fix missing libssl package in the archive

townsend2010:
Seems the archive has removed a newer version of libssl that is compatible with
Qt, so need to download the older version instead.

Note: The newer version of libssl was tried, but QSslSocket failed with
unresolved symbols.


ricab:
bors retry

Saviq:
bors r=ricab

Issue #968, opened 2019-08-14
Systray icon integration on macOS


Saviq:
Some references for the available APIs:

https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html

ricab:
This issue has two parts:
- [ ] Shortcut to Multipass GUI, with the multipass icon and the name "Multipass" (wherever mac shortcuts usually are)
- [ ] Autostarting

Issue #944, opened 2019-08-06
Systray icon integration on Windows


Saviq:
The missing bits:
- [x] icon in Start menu
- [x] autostart handling

The former should be straightforward, with the one caveat that ideally we'd check if it's already running to avoid spawning a second copy. That can come later, though.

The latter I think we can attack in steps:
1. add launching the icon to the current user's `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run` - use `--autostart` flag to tell the binary that it is autostarted
1. add a `multipass set client.gui.autostart` boolean and, if false, exit when ran with `--autostart`
1. have the above `multipass set…` add/remove the `Run` key above
1. add an option in the NSIS installer whether to add the `Run` entry at all

We can spawn separate issues from the above bits, and not all of them need to be done in v0.9.0.

ricab:
Ouch, I missed the edit! Only saw the text now (was empty before IIRC)... Fortunately I was going in the same direction 🙂  A few comments:

- Nice thinking with the flag, but I'd propose calling it `--autostarting` instead, to better signal this is really a declaration rather than a request/option
- why the redundancy between 2 and 3, as well as these two and 4? I'd rather have a single source of truth, so just 1 and 2 all the time. That would apply nicely on other platforms too. Keeping the setting from one installation to the next would also be simpler (just keep the settings file). 
- Do we always want to use HKCU? Shouldn't that be tied to the already existing question on the scope of the PATH change? There is also SHCTX... (perhaps this can be tweaked later)

Saviq:
As for the flag, it's quite internal, so whatever we call it is fine.
As for redundancy my thinking was that we could avoid running the binary altogether if the user disabled it. People also look into their registry and might be unhappy if they see Multipass there even if they disabled it.
We want the NSIS bit to let the user choose during install whether they want it autostarted at all or not.

ricab:
#177 adds an item to the start menu. Not really an "icon" though I am afraid...

PR #991, opened 2019-08-23
Config gui autostart

### LINKED ISSUE #967: Systray icon autostart on Linux

ricab:
Fixes #967 and addresses (lays the ground for) #944 and #968.

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -34,6 +34,7 @@ namespace multipass
 {
 namespace platform
 {
+void preliminary_gui_autostart_setup();
 std::string default_server_address();
 QString default_driver();
 QString daemon_config_home(); // temporary
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -34,7 +34,7 @@ namespace multipass
 {
 namespace platform
 {
-QString preliminary_gui_autostart_setup(); // returns a platform-specific string
+QString setup_gui_autostart_prerequisites(); // returns a platform-specific string
 std::string default_server_address();
 QString default_driver();
 QString daemon_config_home(); // temporary
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -34,7 +34,8 @@ namespace multipass
 {
 namespace platform
 {
-QString setup_gui_autostart_prerequisites(); // returns a platform-specific string
+QString autostart_test_data(); // returns a platform-specific string, for testing purposes
+void setup_gui_autostart_prerequisites();
 std::string default_server_address();
 QString default_driver();
 QString daemon_config_home(); // temporary
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -34,7 +34,7 @@ namespace multipass
 {
 namespace platform
 {
-void preliminary_gui_autostart_setup();
+QString preliminary_gui_autostart_setup(); // returns a platform-specific string
 std::string default_server_address();
 QString default_driver();
 QString daemon_config_home(); // temporary
--- a/snap-wrappers/bin/launch-multipass
+++ b/snap-wrappers/bin/launch-multipass
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+. client-common.sh
+link_autostart
+
+exec "$SNAP/bin/multipass" "$@"
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -50,6 +50,26 @@
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
+namespace
+{
+void set_logger(const mp::ArgParser& parser)
+{
+    mpl::set_logger(std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel())));
+}
+
+void preliminary_setup()
+{
+    try
+    {
+        mp::platform::preliminary_gui_autostart_setup();
+    }
+    catch (std::runtime_error& e)
+    {
+        mpl::log(mpl::Level::warning, "client", e.what());
+    }
+}
+} // namespace
+
 mp::Client::Client(ClientConfig& config)
     : cert_provider{std::move(config.cert_provider)},
       rpc_channel{mp::client::make_channel(config.server_address, config.conn_type, *cert_provider)},
@@ -96,21 +116,13 @@ int mp::Client::run(const QStringList& arguments)
     parser.setApplicationDescription(description);
 
     ParseCode parse_status = parser.parse();
+
+    set_logger(parser);  // we need logging for...
+    preliminary_setup(); // ... something we want to do even if the command was wrong
+
     if (parse_status != ParseCode::Ok)
     {
         return parser.returnCodeFrom(parse_status);
-    }
-
-    auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
-    mpl::set_logger(logger);
-
-    try
-    {
-        mp::platform::preliminary_gui_autostart_setup();
-    }
-    catch (std::runtime_error& e)
-    {
-        mpl::log(logging::Level::warning, "client", e.what());
     }
 
     return parser.chosenCommand()->run(&parser);
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -45,6 +45,7 @@
 #include <multipass/cli/client_common.h>
 #include <multipass/logging/log.h>
 #include <multipass/logging/standard_logger.h>
+#include <multipass/platform.h>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
@@ -102,6 +103,15 @@ int mp::Client::run(const QStringList& arguments)
 
     auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
     mpl::set_logger(logger);
+
+    try
+    {
+        mp::platform::preliminary_gui_autostart_setup();
+    }
+    catch (std::runtime_error& e)
+    {
+        mpl::log(logging::Level::warning, "client", e.what());
+    }
 
     return parser.chosenCommand()->run(&parser);
 }
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -44,31 +44,10 @@
 #include <multipass/cli/argparser.h>
 #include <multipass/cli/client_common.h>
 #include <multipass/logging/log.h>
-#include <multipass/logging/standard_logger.h>
 #include <multipass/platform.h>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
-
-namespace
-{
-void set_logger(const mp::ArgParser& parser)
-{
-    mpl::set_logger(std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel())));
-}
-
-void preliminary_setup()
-{
-    try
-    {
-        mp::platform::preliminary_gui_autostart_setup();
-    }
-    catch (std::runtime_error& e)
-    {
-        mpl::log(mpl::Level::warning, "client", e.what());
-    }
-}
-} // namespace
 
 mp::Client::Client(ClientConfig& config)
     : cert_provider{std::move(config.cert_provider)},
@@ -117,8 +96,8 @@ int mp::Client::run(const QStringList& arguments)
 
     ParseCode parse_status = parser.parse();
 
-    set_logger(parser);  // we need logging for...
-    preliminary_setup(); // ... something we want to do even if the command was wrong
+    mp::client::set_logger(mpl::level_from(parser.verbosityLevel())); // we need logging for...
+    mp::client::preliminary_setup(); // ... something we want to do even if the command was wrong
 
     if (parse_status != ParseCode::Ok)
     {
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -115,6 +115,6 @@ void mp::client::preliminary_setup()
     catch (AutostartSetupException& e)
     {
         mpl::log(mpl::Level::error, "client", fmt::format("Failed to set up autostart prerequisites: {}", e.what()));
-        mpl::log(mpl::Level::debug, "client", fmt::format("Tried:\n  {}", e.get_detail()));
+        mpl::log(mpl::Level::debug, "client", fmt::format("Tried: {}", e.get_detail()));
     }
 }
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -113,6 +113,6 @@ void mp::client::preliminary_setup()
     }
     catch (std::runtime_error& e)
     {
-        mpl::log(mpl::Level::error, "client", fmt::format("Error setting up autostart prerequisites: {}", e.what()));
+        mpl::log(mpl::Level::error, "client", fmt::format("Failed to setup autostart prerequisites: {}", e.what()));
     }
 }
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -109,7 +109,7 @@ void mp::client::preliminary_setup()
 {
     try
     {
-        mp::platform::preliminary_gui_autostart_setup();
+        mp::platform::setup_gui_autostart_prerequisites();
     }
     catch (std::runtime_error& e)
     {
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -22,8 +22,11 @@
 #include <QStandardPaths>
 
 #include <fmt/ostream.h>
+#include <multipass/logging/log.h>
+#include <multipass/logging/standard_logger.h>
 
 namespace mp = multipass;
+namespace mpl = multipass::logging;
 
 namespace
 {
@@ -90,4 +93,26 @@ std::unique_ptr<mp::SSLCertProvider> mp::client::get_cert_provider()
     auto data_dir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
     auto client_cert_dir = mp::utils::make_dir(data_dir, "client-certificate");
     return std::make_unique<mp::SSLCertProvider>(client_cert_dir);
+}
+
+void mp::client::set_logger()
+{
+    set_logger(mpl::Level::info);
+}
+
+void mp::client::set_logger(mpl::Level verbosity)
+{
+    mpl::set_logger(std::make_shared<mpl::StandardLogger>(verbosity));
+}
+
+void mp::client::preliminary_setup()
+{
+    try
+    {
+        mp::platform::preliminary_gui_autostart_setup();
+    }
+    catch (std::runtime_error& e)
+    {
+        mpl::log(mpl::Level::warning, "client", e.what());
+    }
 }
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -22,6 +22,7 @@
 #include <QStandardPaths>
 
 #include <fmt/ostream.h>
+#include <multipass/exceptions/autostart_setup_exception.h>
 #include <multipass/logging/log.h>
 #include <multipass/logging/standard_logger.h>
 
@@ -109,10 +110,11 @@ void mp::client::preliminary_setup()
 {
     try
     {
-        mp::platform::setup_gui_autostart_prerequisites();
+        platform::setup_gui_autostart_prerequisites();
     }
-    catch (std::runtime_error& e)
+    catch (AutostartSetupException& e)
     {
         mpl::log(mpl::Level::error, "client", fmt::format("Failed to set up autostart prerequisites: {}", e.what()));
+        mpl::log(mpl::Level::debug, "client", fmt::format("Tried:\n  {}", e.get_detail()));
     }
 }
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -113,6 +113,6 @@ void mp::client::preliminary_setup()
     }
     catch (std::runtime_error& e)
     {
-        mpl::log(mpl::Level::warning, "client", e.what());
+        mpl::log(mpl::Level::error, "client", fmt::format("Error setting up autostart prerequisites: {}", e.what()));
     }
 }
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -113,6 +113,6 @@ void mp::client::preliminary_setup()
     }
     catch (std::runtime_error& e)
     {
-        mpl::log(mpl::Level::error, "client", fmt::format("Failed to setup autostart prerequisites: {}", e.what()));
+        mpl::log(mpl::Level::error, "client", fmt::format("Failed to set up autostart prerequisites: {}", e.what()));
     }
 }
--- a/src/client/gui/client_gui.cpp
+++ b/src/client/gui/client_gui.cpp
@@ -19,8 +19,10 @@
 #include "argparser.h"
 
 #include <multipass/cli/client_common.h>
+#include <multipass/constants.h>
 #include <multipass/logging/log.h>
 #include <multipass/logging/standard_logger.h>
+#include <multipass/settings.h>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
@@ -35,10 +37,25 @@ mp::ClientGui::ClientGui(ClientConfig& config)
 
 int mp::ClientGui::run(const QStringList& arguments)
 {
+    auto ret = 0;
     ArgParser parser;
 
-    auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
-    mpl::set_logger(logger);
+    QCommandLineOption autostart{"exit-unless-autostart",
+                                 "Exit right away, not actually creating a GUI, unless configured to autostart. "
+                                 "Pass this option when auto-starting to honor the autostart setting."};
+    autostart.setFlags(QCommandLineOption::HiddenFromHelp);
 
-    return gui_cmd->run(&parser);
+    parser.addOption(autostart);
+    parser.addHelpOption();
+    parser.process(arguments);
+
+    if (!parser.isSet(autostart) || Settings::instance().get_as<bool>(autostart_key))
+    {
+        auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
+        mpl::set_logger(logger);
+
+        ret = gui_cmd->run(&parser);
+    }
+
+    return ret;
 }
--- a/src/client/gui/client_gui.cpp
+++ b/src/client/gui/client_gui.cpp
@@ -40,7 +40,7 @@ int mp::ClientGui::run(const QStringList& arguments)
     auto ret = 0;
     ArgParser parser;
 
-    QCommandLineOption autostart{"exit-unless-autostart",
+    QCommandLineOption autostart{"autostarting",
                                  "Exit right away, not actually creating a GUI, unless configured to autostart. "
                                  "Pass this option when auto-starting to honor the autostart setting."};
     autostart.setFlags(QCommandLineOption::HiddenFromHelp);
--- a/src/client/gui/client_gui.cpp
+++ b/src/client/gui/client_gui.cpp
@@ -20,8 +20,6 @@
 
 #include <multipass/cli/client_common.h>
 #include <multipass/constants.h>
-#include <multipass/logging/log.h>
-#include <multipass/logging/standard_logger.h>
 #include <multipass/platform.h>
 #include <multipass/settings.h>
 
@@ -39,6 +37,9 @@ mp::ClientGui::ClientGui(ClientConfig& config)
 int mp::ClientGui::run(const QStringList& arguments)
 {
     auto ret = 0;
+    mp::client::set_logger();        // we need logging for...
+    mp::client::preliminary_setup(); // ... something we want to do even if the command was wrong
+
     ArgParser parser;
 
     QCommandLineOption autostart{"autostarting",
@@ -52,18 +53,6 @@ int mp::ClientGui::run(const QStringList& arguments)
 
     if (!parser.isSet(autostart) || Settings::instance().get_as<bool>(autostart_key))
     {
-        auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
-        mpl::set_logger(logger);
-
-        try
-        {
-            mp::platform::preliminary_gui_autostart_setup();
-        }
-        catch (std::runtime_error& e)
-        {
-            mpl::log(logging::Level::warning, "client", e.what());
-        }
-
         ret = gui_cmd->run(&parser);
     }
 
--- a/src/client/gui/client_gui.cpp
+++ b/src/client/gui/client_gui.cpp
@@ -22,6 +22,7 @@
 #include <multipass/constants.h>
 #include <multipass/logging/log.h>
 #include <multipass/logging/standard_logger.h>
+#include <multipass/platform.h>
 #include <multipass/settings.h>
 
 namespace mp = multipass;
@@ -53,6 +54,15 @@ int mp::ClientGui::run(const QStringList& arguments)
     {
         auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
         mpl::set_logger(logger);
+
+        try
+        {
+            mp::platform::preliminary_gui_autostart_setup();
+        }
+        catch (std::runtime_error& e)
+        {
+            mpl::log(logging::Level::warning, "client", e.what());
+        }
 
         ret = gui_cmd->run(&parser);
     }
--- a/src/client/gui/client_gui.cpp
+++ b/src/client/gui/client_gui.cpp
@@ -36,7 +36,6 @@ mp::ClientGui::ClientGui(ClientConfig& config)
 
 int mp::ClientGui::run(const QStringList& arguments)
 {
-    auto ret = 0;
     mp::client::set_logger();        // we need logging for...
     mp::client::preliminary_setup(); // ... something we want to do even if the command was wrong
 
@@ -51,10 +50,9 @@ int mp::ClientGui::run(const QStringList& arguments)
     parser.addHelpOption();
     parser.process(arguments);
 
+    auto ret = ReturnCode::Ok;
     if (!parser.isSet(autostart) || Settings::instance().get_as<bool>(autostart_key))
-    {
         ret = gui_cmd->run(&parser);
-    }
 
     return ret;
 }
--- a/src/platform/platform_linux.cpp
+++ b/src/platform/platform_linux.cpp
@@ -29,13 +29,35 @@
 #include "logger/journald_logger.h"
 #include <disabled_update_prompt.h>
 
+#include <QStandardPaths>
+
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 namespace mu = multipass::utils;
 
+namespace
+{
+constexpr auto autostart_desktop_contents = "[Desktop Entry]\n"
+                                            "Name=Multipass\n"
+                                            "Exec=multipass.gui --autostarting\n"
+                                            "Icon=${SNAP}/meta/gui/multipass-gui.svg\n"
+                                            "Type=Application\n"
+                                            "Terminal=false\n"
+                                            "Categories=Utility;\n";
+}
+
 void mp::platform::preliminary_gui_autostart_setup()
 {
-    // TODO
+    static const auto config_dir = QDir{QStandardPaths::writableLocation(QStandardPaths::GenericConfigLocation)};
+    static const auto autostart_dir = QDir{config_dir.filePath("autostart")};
+    static const auto fname = autostart_dir.filePath(QStringLiteral("multipass-gui.conditional-autostart.desktop"));
+    // TODO @ricab make base filename constant
+
+    autostart_dir.mkpath(".");
+    QFile f{fname};
+    if (!f.exists())                                        // assuming correct contents otherwise
+        if (f.open(QIODevice::WriteOnly | QIODevice::Text)) // TODO @ricab handle failure
+            f.write(autostart_desktop_contents);
 }
 
 std::string mp::platform::default_server_address()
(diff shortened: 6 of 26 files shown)


ricab:
Snaps have an `autostart` field that is supposed to be coupled with a desktop file. This desktop file should go into `$SNAP_USER_DATA/.config`. But that is per user, and I'm not sure we want that &ndash; couldn't get it there at install (which user?) and don't need it per user since the autostart setting already covers that. 

I wonder if there is some global autostart place we could use... More digging needed.

ricab:
To clarify my earlier comment, it is currently the application's responsibility (not the snap's) to get the desktop file in `$SNAP_USER_DATA/.config` (e.g. on first run). 

For future reference, we discussed this a bit and, for now, we decided to make `multipass` and `multipass-gui` write it there when they first come up. Later on, we can drop the underlying flag entirely, presenting it to the user but not saving it as such: `multipass set client.gui.autostart` would just do the necessary to (de)activate autostarting on the present platform.

ricab:
@Saviq I have been thinking and researching snaps/autostart.

Some conclusions:
  - apparently we still need `passthrough` to make this work. I needed it with

```console
$ snapcraft --version
snapcraft, version 2.43.1+18.10
```
... which is the version I get in disco (despite 18.10 in the version string). Someone says it is still [needed in 3.4](https://forum.snapcraft.io/t/how-to-add-autostart-to-the-app/8866/3).

  - `/etc/xdg/autostart/snap-userd-autostart.desktop` is how snaps are autostarted

Considerations on way forward:

  - In general, if you want your linux application to provide an autostart toggle, you'll probably rely on the [freedesktop.org specification](https://specifications.freedesktop.org/autostart-spec/0.5/ar01s02.html): you'd put a desktop file in `XDG_CONFIG_HOME/autostart`
  - How you package and distribute your application *should* have minimal impact on what it does (ideally none). 
  - snaps seem to apply this principle when they emulate the same interface for autostarting (*almost* the same, see next point). They in turn use the same approach to implement it (as noted above).
  - However, snaps expect the desktop file to be in `$SNAP_USER_DATA/.config/autostart` rather than `XDG_CONFIG_HOME`, which they do not define by default.
  - But `XDG_CONFIG_HOME`'s utility goes beyond autostart and we rely on it for other things
  - Since we define `XDG_CONFIG_HOME` to be `$SNAP_USER_DATA/config`, we could make the general autostart behavior work with snaps by adding a `.config -> config` link.
  - This would treat people compiling from source as first class citizens.
  - This would be consistent with how we have been using the XDG vars so far: transparently. It becomes transparent to the application whether or not it is in a snap.

This, quite frankly, seems to me to be the right thing to do. Packaging should be an additive layer, not a disruptive one. It is in agreement with the plan we had until yesterday afternoon when I brought up a separate matter (below). Namely, that we meant to move `multipass set autostart` to do the necessary to (de)activate autostarting in the present platform. Wrappers would be incompatible with that, and they would require branching somewhere too (they could not be merely addictive, there would behavior to subtract).

This leaves one aspect to solve, which is what I sought opinions on yesterday: the cmd for `Exec` in the desktop file is different in snaps. There are multiple ways to address this without departing from existing plans. A very simple one would be to have Exec="if multipass.gui exists, use that, else use multipass-gui". Another, even simpler one, would be to change the executable name to `multipass.gui` directly. The latter would maximize transparency, but I am not sure if this could have implications I am not considering. I recognize these options are also an instance of "packaging disrupting application", but much less than branching in the code or adding separate wrappers. If we only renamed, it would be a one time thing (not a permanent branch in code). Another option that comes to mind would be a link. Feedback needed.

townsend2010:
@ricab,

> apparently we still need passthrough to make this work. I needed it with

Why aren't you using the snapcraft snap, even from stable?  This is the best supported way of getting updated snapcraft.

townsend2010:
After reading this, a few questions/comments come to mind.

- I assume we can't define `XDG_CONFIG_HOME` to be `$SNAP_USER_DATA/.config` is because we have some releases out now that define it the current way?  Or is there some other reason?  If there is no other reason, we could redefine that and then in the postinstall script, look for `$SNAP_USER_DATA/config` and rename it to `$SNAP_USER_DATA/.config`.
- I may be wrong, but it seems like much of the hassle is to try to treat multipass compiled from source to work the same.  Is this correct?  If so, I think there may be too much emphasis put on this.  We really only care and have cared about the snap package for end users and so that should be the focus.

Just my $0.02 after jumping in a bit late on this conversation :wink:

ricab:
@townsend2010 Thanks for the hint. I wonder why the discrepancy in versions...

Anyway, with 3.7.2 passthrough is not needed anymore, so that is good news

townsend2010:
@ricab,

> I wonder why the discrepancy in versions...

Because they aren't maintaining the snapcraft Debian package any more.  Last update was ~1 year ago.

ricab:
@townsend2010 

- Re: renaming `config` to `.config`: maybe it could be done, but I think adding a link would be safer and easier. `config` would be kept parallel to `cache` and `data`
- Re hassle: there needn't be any hassle, a couple of links would do it! My plans did get upturned when I brought the matter up yesterday, but it didn't have to be that way... 

Having a feature dependent on the packaging does not seem very sensible to me. We have cared about the non-snap case in the past (maybe not always, I don't know). The XDG vars are proof of that. IMO that is a good thing. Maybe it sounds theoretical, but it would soon get very practical whenever we wanted to manually test it and needed to snapcraft. Unit testing would be excluded. That would be a much bigger hassle.

This is like "should I do a link, rename, if-else? that would get me there &mdash; no let's revisit the whole thing, turn back, start going a different direction &mdash; but I am almost there anyway and this direction has advantages a, b, and c, &mdash; those advantages are not important, let's change plans, they are not worth the hassle". I bet it's in a Dilbert comic strip somewhere...

ricab:
It occurred to me that, if my path was perceived to take longer, that could cause resistance. Not really the case, at the point I was at, It would actually have taken much longer to change to wrappers.

There were only a few missing bits, and I wanted to put them together to show that it works as I intended. [This](https://github.com/CanonicalLtd/multipass/compare/master...ricab:config-gui-autostart) makes autostart work on linux (my fork because I have some tags that I block in a push hook which would get messed up if I pushed it here). Both snap and non-snap (w/ mp on the PATH). A couple of open ends (marked with TODO @ricab), but this serves as a proof of concept.

src/client/gui/client_gui.cpp
```diff
@@ -35,10 +37,25 @@ mp::ClientGui::ClientGui(ClientConfig& config)
...
 int mp::ClientGui::run(const QStringList& arguments)
 {
+    auto ret = 0;
     ArgParser parser;
 
-    auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
-    mpl::set_logger(logger);
+    QCommandLineOption autostart{"exit-unless-autostart",
+                                 "Exit right away, not actually creating a GUI, unless configured to autostart. "
+                                 "Pass this option when auto-starting to honor the autostart setting."};
```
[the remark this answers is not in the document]
ricab: Yes, I did that already here: https://github.com/CanonicalLtd/multipass/compare/master...ricab:config-gui-autostart

I just did not push to this repo yet because of the hook issues I mentioned (I've a push check for `//TODO @ricab` tags and it relies on them not being in the base... nevermind).

tests/test_cli_client.cpp
```diff
@@ -1390,6 +1401,35 @@ TEST_F(Client, set_handles_persistent_settings_errors)
...
+TEST_F(Client, set_cmd_rejects_bad_autostart_value)
+{
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "asdf");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "True");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "trueasdf");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "False");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "0");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "1");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "123");
+    aux_set_cmd_rejects_bad_val(mp::autostart_key, "");
```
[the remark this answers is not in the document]
ricab: OK, I can do that.

ricab:
> They could've (and IMO they should have, to not facilitate packagers equating, or worse, setting `$HOME` to `$SNAP_USER_DATA`) very well placed the `.desktop` file in `$SNAP_USER_DATA/foobar/autostart`

Hmm, ok I see it differently. I am not saying they equate it to home, but they don't constrain it either (I see no reason why they should do so). But the `.config` thing gives away that is what they are assuming. IMO they should accept any path under `$SNAP_USER_DATA` and let the user specify it in the `autostart` flag (along with the file name). That way, the client would be free to choose they layout they wanted.

ricab:
@Saviq 

> I'd be OK with us saying that we'll be treating $SNAP_USER_DATA as $HOME for the purposes of the clients.

We could also stick to our preferred layout and do the minimum change to conform to the `.config` constraint snaps are imposing. I don't see a technical reason to adhere to a layout strictly reflecting home. Not that it would be a bad option, just not worth the effort IMO.

On top of that, do you think we could convince the snapcraft folk to free the autostart param? If so, and if we used a `.config -> config` link, we could just remove it when that was supported.

Saviq:
> We could also stick to our preferred layout and do the minimum change to conform to the `.config` constraint snaps are imposing. I don't see a technical reason to adhere to a layout strictly reflecting home. Not that it would be a bad option, just not worth the effort IMO.

If at all, I'd rather do a `.config/autostart → config/autostart` link rather than the "global" `.config → config`. That would clearly show what its purpose is, and not put things in `.config` that might have no place there.

> On top of that, do you think we could convince the snapcraft folk to free the autostart param? If so, and if we used a `.config -> config` link, we could just remove it when that was supported.

I very much doubt that. This is public, released API now.

ricab:
> I'd rather do a `.config/autostart → config/autostart`

Yes, that does sound better.

tests/test_cli_client.cpp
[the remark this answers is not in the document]
ricab: I made it support "true", "on", "yes", "1", "false", "off", "no", "0", and all upper case variations thereof. We can only cover so much natural language, so I use an explicit conversion and  still reject everything else and, rather than giving the input directly to Qt and have it interpreting non-empty strings (e.g. "-", "nope", "n", "000", "nein") as true.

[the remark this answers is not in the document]
Saviq: Yeah totally. Could you make it case insensitive?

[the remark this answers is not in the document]
ricab: Sure, I did already.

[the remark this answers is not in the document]
Saviq: Oh, I misunderstood "upper case variations" :)

ricab:
This is now ready for review.

ricab:
Hints for reviewers: 
- you can use `dex -ad` to do a dry run and see what commands would be autostarted for you. When not in a snap, `multipass.gui --autostarting` should be included. That will immediately quit unless `multipass get client.gui.autostart` gives `true`.
- to actually get it to autostart on desktop login, the the gui needs to be found in the path. To avoid polluting system dirs with `make install`, you can instead edit your `~/.profile` and extend `PATH` there.
- no tricks needed with snap, just:
  1. snap install
  2. `multipass set client.gui.autostart=true`
  3. log out
  4. log in
- if you want to inspect or delete autostart files, they are in `~/{snap/multipass/,}.config/autostart/multipass.gui.conditional-autostart.desktop`

src/client/gui/client_gui.cpp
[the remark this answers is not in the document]
Saviq: Can you change the help text to say something like "Tell the GUI to apply whatever behaviour is applicable for autostarting." rather than explicitly saying what it is today?

src/client/cli/client.cpp
```diff
@@ -103,5 +104,14 @@ int mp::Client::run(const QStringList& arguments)
     auto logger = std::make_shared<mpl::StandardLogger>(mpl::level_from(parser.verbosityLevel()));
     mpl::set_logger(logger);
 
+    try
+    {
+        mp::platform::preliminary_gui_autostart_setup();
```
Saviq: Same here.

snap-wrappers/bin/launch-multipass
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+. client-common.sh
+link_autostart
```
Saviq: I wonder if we should be doing this for the CLI, too?

src/platform/platform_linux.cpp
```diff
@@ -29,10 +29,45 @@
...
+#include <cstring>
+#include <stdexcept>
+
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 namespace mu = multipass::utils;
 
+namespace
+{
+constexpr auto autostart_filename = "multipass.gui.conditional-autostart.desktop";
```
Saviq: Just `.autostart.desktop`, we don't want to have to change that name when it's not conditional any more ;)

src/platform/platform_linux.cpp
```diff
@@ -29,10 +29,45 @@
...
+
+QString mp::platform::preliminary_gui_autostart_setup()
+{
+    const auto config_dir = QDir{QStandardPaths::writableLocation(QStandardPaths::GenericConfigLocation)};
+    const auto autostart_dir = QDir{config_dir.absoluteFilePath("autostart")};
+    const auto filepath = autostart_dir.absoluteFilePath(autostart_filename);
+
+    autostart_dir.mkpath(".");
+    QFile f{filepath};
+    if (!f.exists()) // assuming correct contents otherwise
```
Saviq: Should we be assuming correct contents? I wonder if this can be a symlink to a file in `$SNAP` instead?

BTW, if you wanted this generic, the `Icon=` path probably shouldn't point at `${SNAP}`?

ricab:
Some more notes for reviewers and future reference:

- useful tools:
    + `desktop-file-validate`
    + `dex` (`dex --verbose -a` to autostart stuff)
    + `xdg-autostart`
- the snapd facility that does snap autostarts:
    + `/etc/xdg/autostart/snap-userd-autostart.desktop`

tests/test_cli_client.cpp
```diff
@@ -1392,6 +1403,31 @@ TEST_F(Client, set_handles_persistent_settings_errors)
     EXPECT_THAT(send_command({"set", keyval_arg(key, val)}), Eq(mp::ReturnCode::CommandFail));
 }
 
+TEST_F(Client, get_returns_acceptable_autostart_value_by_default)
+{
+    EXPECT_THAT(get_setting(mp::autostart_key), AnyOf("true", "false"));
```
Saviq: Are you testing the acceptance of "on", "off" etc. somewhere?

townsend2010:
Not sure if this has been noticed, but Mac is not happy with some of the C++ things in the code: https://travis-ci.org/CanonicalLtd/multipass/jobs/581221612#L3853

You may want to address those here.

snap-wrappers/bin/launch-multipass
ricab: My reasoning is: the CLI is our interface to the autostart flag; without this, `multipass set client.gui.autostart=true` wouldn't work (until the gui was manually started); the clients share the same environment (same settings file even), so whichever comes up first can get the required infrastructure in place.

Saviq: OK true.

src/utils/settings.cpp
```diff
@@ -37,11 +37,13 @@ namespace
 const auto file_extension = QStringLiteral("conf");
 const auto daemon_root = QStringLiteral("local");
 const auto petenv_name = QStringLiteral("primary");
+const auto autostart_default = QStringLiteral("false");
```
Saviq: Hmm I wonder if this should be `true` instead? This way when the user installs and runs the GUI, it will autostart from then on, until they disable it. We should be able to add a toggle of some kind to the tray menu, too.

src/platform/platform_linux.cpp
ricab: Good point. My intention was to let users looking around know that this does not mean the gui is really autostarted. But that does not justify dealing with a lingering file later. I'll do that.

src/platform/platform_linux.cpp
ricab: 1. I initially thought of doing it along those lines, but later opted for this approach to avoid file ownership/permissions issues
2. Good catch. I am not sure what to do here though, any ideas? Maybe just dropping the icon altogether?

tests/test_cli_client.cpp
ricab: No, so far there are no tests checking acceptance/rejection of settings values unfortunately :/  This is because, to do that, I would need another wrapping layer around the `QSettings` backend, in order to mock it. That would be a considerable investment that I have been avoiding so far.

src/utils/settings.cpp
ricab: It won't unfortunately, not until either CLI or GUI are run manually. Here is the msg from that commit:

```
[settings] Disable autostart by default

Otherwise, the flag would not be respected until the user ran a client.
This way, autostarting requires a `multipass set`, at which point the
autostart infrastructure is setup. This also accommodates the plan to
implement the autostart setting with the underlying platform 
rather than a plain flag.
```

ricab:
> Mac is not happy

Ahh, pity, mac still not up to par with c++17. Thanks for pointing it out @townsend2010, I'll fix it.

src/utils/settings.cpp
Saviq: Not convinced, on Windows we have the installer where we can enable autostart (long-term - conditionally). That's likely the same on macOS, really only snaps are problematic here, and I actually think they will fix it long term.

When we implement it with the platform, the default here won't matter anyway.

tests/test_cli_client.cpp
Saviq: Ack.

src/platform/platform_linux.cpp
Saviq: 1. We determine ownership of files in the snap, so that shouldn't be a problem.
2. Nothing comes to mind that will happen automagically… Snapcraft does rewrite the .desktop files, but only those that are actually bound to an `app` in `snapcraft.yaml`… We'd have to register a dummy "Multipass (autostart)" app to use that :(

src/utils/settings.cpp
ricab: Are you sure? If set to true now, how do we handle it in the platform implementation? How will we be able to have a default of `true` then?

Saviq: On Windows, you add the registry key unconditionally already. The config key becomes moot and `multipass.gui --autostart` won't look at it, if called it will just start. Migration from the key to platform implementation will be a bit awkward, but I think solvable.

src/platform/platform_linux.cpp
ricab: 1. Let me break this down into a few points:
    1. "assuming correct contents": I could simply overwrite, if you wanted. That would kind of rule out manual tweaks, in a potentially confusing way (accepted at first, overwritten on 1st run). Maybe that's not a big problem and still something you could be happy with?
    2. I don't see how a link helps there: we can either overwrite it or it can still have/point-to anything
    3. I believe writing our own strings to files is something we already do in other places (cloud init stuff, libvirt xmls, apparmor profiles now)?
    4. given all this, I don't see much advantage in the symlink approach, and it would be more work before I can move on to other platforms.
    5. if you still want to do it, what ownership would you give the target? Currently the file is owned by the user.
2. OK, I'll just drop the icon

Saviq: 1. Sure:
    1. I thought of overwriting, but that seems excessive. That's exactly where a symlink would help - it would point to a read-only location in `${SNAP}` so there's no modification that could be made. The right place to do modifications to this would be `~/.config/autostart`. In any case, people can remove the symlink and they're on their own.
    1. a symlink to a read-only file in `${SNAP}` is clearer and simpler than writing the file out (which may also lead to corruption of the file)
    1. the examples you gave are all per-instance (not sure about apparmor, if we could keep those in static files and link, we should)
    1. it would be less code to maintain in the end, though
    1. the file goes in the snap, and as long as it's world-readable, it should be fine
1. ACK

@gerboland @townsend2010 care to weigh in on this?

src/utils/settings.cpp
Saviq: Sorry, I seem to have edited out "When we move to platform implementation of this…".

I agree it may lead to inconsitent behaviour today (it will auto-start upon installation on Windows and hopefully macOS, but not on Linux), but as soon as you run anything `multipass…` it will sort itself out.

src/platform/platform_linux.cpp
ricab: 1. 
    1. "people can remove the symlink and they're on their own" &ndash; generalizing, you can say "overwrite" in place of "remove"; the situation is parallel to a regular file
    2. I concede it may be clearer. Simpler, hmm maybe. I still don't see how it helps integrity: modification and corruption can happen in either case.
    3. ack
    4. ok, marginally: the string just moves to a separate file, but you replace open+write with a single link; error handling still needed though. Still more work now :stuck_out_tongue: 
    5. ofc, good point.

OK, I'll proceed with the other review items and wait for feedback from the others for this.

Saviq: Any corruption would have to happen on the packaging level, so a much bigger problem.

Also realized there's another advantage - handling packaging-specific differences (like `Icon=${SNAP}`) becomes much easier, if the packaging ships a static .desktop file to static location and we only link to it.

ricab: > Any corruption would have to happen on the packaging level, so a much bigger problem.

packaging, deployment, reading... anyway, splitting hairs here

> another advantage - handling packaging-specific differences (like Icon=${SNAP}) becomes much easier

Sorry, I don't quite understand what you mean, could you explain?

Saviq: > Sorry, I don't quite understand what you mean, could you explain?

You were trying to find a way to keep the packaging differences away from the app itself, and having a packaged static file (that may have a different `Icon=` path, for example, or a different `Exec=` line) seems like a good approach to me. It makes any differentiation contained in packaging alone.

ricab: Ah OK, I think I see what you mean.

Saviq: Just to summarize, as I stewed with this a bit again :)

I think the symlink approach makes for the best separation between app and packaging:
- app is only responsible for placing the link in `${XDG_CONFIG_HOME}/autostart`
- packaging is responsible for placing the actual `.desktop` file, applicable to the packaging, in the place where app is expecting it

The one scenario where this gets wonky is if the file needs to differ based on runtime properties of the system, rather than build time.

ricab: OK, I'll do that then.

ricab: @Saviq I starting doing this, but immediately stumbled into another hole: where to place the actual file in the first place (in the build dir)? Options I see:
- place it in `bin` -> I would still have to get the path to the current executable to find the desktop file (cwd can be something else entirely)
- somewhere relative to `bin` -> same problem + we'd have to keep the same relative path on `make install`
- use cmake to bake the file into the executable, to write it down later -> not much different from the current approach

So far autostart would work as long as `multipass.gui` was in the PATH. Now it would need to find another file. This seems to do away with the marginal simplicity gain of replacing "open+write" with "link". On the contrary, it feels like adding complexity for no tangible reason honestly.

As for packaging differences, there needn't be any. The `Icon` thing wouldn't work either way as you pointed above.

Saviq: Rather than baking the file contents into the executable in CMake, you can bake a "resources path" into it. Today it may only be there for the .desktop file, but we're bound to need more static resources with time.

ricab: Having an absolute path in the executable? Is that what you're saying? Even if generated at build time, that would constrain the exe to a particular filesystem. And it would not address the `make install` question. If that is not what you're saying, I didn't get it.

Saviq: If the packaging system puts all the files in certain locations, I wouldn't expect that executable to be able to deal with changing those locations, unless that was documented behaviour.
Why would `make install` not be addressed by that? The `.desktop` file would be installed to a location based on install prefixes, and that location would also be where the executable would be looking for it.

An environment variable allowing to override that path on runtime is an option. I don't know of a standard environment variable that we could set to affect e.g. [`AppDataLocation`](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum), but I haven't read the code and even `XDG_` vars aren't mentioned in that documentation, so…

Saviq: Actually, it [seems to respect `XDG_DATA_DIRS`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/io/qstandardpaths_unix.cpp#n276).

ricab: Sorry, I am getting lost in all this... I don't really see what you are proposing

ricab: What about we offload that to a separate issue?

Saviq: You didn't like "absolute path in the executable", so another option is to use a standard path - `QStandardPaths` use the `XDG_DATA_DIRS` (note that's different to `_HOME`) environment variable to determine where data files for an application live. So instead of baking in an absolute path, we could use that, and rely on `QStandardPaths::AppDataLocation` (or one of the other ones, if applicable) to locate the static `.desktop` file.
Packaging would then be the one to make sure that either the file is indeed placed in the standard location (`/usr/share/…`, really), or `XDG_DATA_DIRS` is set as appropriate (in the snap case).

It doesn't help In the `make install` case unless you export `XDG_DATA_DIRS` to the right place in your install prefix.

I still think the link/unlink step is a much cleaner story for toggling autostart in the long run rather than writing out the file whole.

We're bound to need more resource files in the long run, so we can decide today how to make this happen.

Saviq: I don't want to merge code in that we'll have to deal with in another PR, because then we'll need to do a migration of some kind. We have the chance to set a precedent here, so let's do that.

ricab: OK, so something like this? (please confirm or adapt):

- src:
    - new `resources` dir (`755` permissions)
    - new `.desktop` file there (`644` permissions)
- at `make`:
    - copy the file to a new `resources/multipass` dir
- at `make install`
    - find the first existing directory from `${XDG_DATA_DIRS}`, if not empty, `/usr/local/share/:/usr/share/` otherwise;
    - create a `multipass` subdir there
    - copy the file there
- in the snap:
    - use dump plugin to copy the `resources/multipass` dir
    - prepend the `resources` dir (not `multipass`) it to `XDG_DATA_DIRS`
- in the executable:
    - use [`QStandardPaths::locate`](https://doc.qt.io/qt-5/qstandardpaths.html#locate) to find the file (warning on failure)
    - create a link to it
- to test without installing
    - beside extending `PATH` we now need an additional step, namely adding the `resources` dir to `XDG_DATA_DIRS`

Except this won't really work because it assumes the app name `multipass` which is not the case in `multipass.gui`. So much for simplicity :joy:

ricab:
This last push hopefully addresses all review items with the exception of the link stuff. 

Not sure about the mac issue, will have to see the result in CI...

src/platform/platform_linux.cpp
Saviq: > * at `make install`
>   * find the first existing directory from `${XDG_DATA_DIRS}`, if not empty, `/usr/local/share/:/usr/share/` otherwise;
>   * create a `multipass` subdir there
>   * copy the file there

No, packaging only ever installs in one destination. Using [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) and `${CMAKE_INSTALL_DATADIR}` as destination in a call to [`install`](https://cmake.org/cmake/help/latest/command/install.html), or potentially [`configure_file`](https://cmake.org/cmake/help/latest/command/configure_file.html) when packaging differences needed.

> * in the snap:
>   * use dump plugin to copy the `resources/multipass` dir
>   * prepend the `resources` dir (not `multipass`) it to `XDG_DATA_DIRS`

Rather than the above, we just need to set `XDG_DATA_DIRS` to `${SNAP}/share`.

> Except this won't really work because it assumes the app name multipass which is not the case in multipass.gui. So much for simplicity :joy:

Not sure why `multipass` vs. `multipass.gui` matters here.
This moves responsibility to build/packaging time, over which we have much more control. The application only then needs to `locate` and link / unlink the `.desktop` file every time it wishes to enable/disable autostart. A much simpler operation than writing the file out.

Saviq: My quick stab at this: http://paste.ubuntu.com/p/9sDyYFypkS/

src/client/common/client_common.cpp
```diff
@@ -91,3 +94,25 @@ std::unique_ptr<mp::SSLCertProvider> mp::client::get_cert_provider()
...
+
+void mp::client::preliminary_setup()
+{
+    try
+    {
+        mp::platform::preliminary_gui_autostart_setup();
+    }
+    catch (std::runtime_error& e)
+    {
+        mpl::log(mpl::Level::warning, "client", e.what());
```
Saviq: This should probably be an error, no?

include/multipass/platform.h
```diff
@@ -34,6 +34,7 @@ namespace multipass
 {
 namespace platform
 {
+QString preliminary_gui_autostart_setup(); // returns a platform-specific string
```
Saviq: Is this string only used for testing?

src/platform/platform_linux.cpp
Saviq: I can see now what you meant with `multipass` vs. `multipass-gui`, that does put a proverbial stick between the spokes (not sure if that translates…). I'm looking into what we could do here.

Saviq: OK this is my proposal: http://paste.ubuntu.com/p/RJCPKQZwnh/

ricab: > No, packaging only ever installs in one destination.

That is what I meant with "first existing directory". 

I like the standard `GNUInstallDirs` approach, but we have to hope it will always be included in `XDG_DATA_DIRS`. I suppose we can go with it and look back if anyone ever reports a failure.

> Rather than the above, we just need to set XDG_DATA_DIRS to ${SNAP}/share.

OK, so set rather than prepend and use the standard dir directly.

> This moves responsibility to build/packaging time, over which we have much more control. The application only then needs to locate and link / unlink the .desktop file every time it wishes to enable/disable autostart. A much simpler operation than writing the file out.

There is an elegance element in extracting the data from the app logic to packing. I am sensitive to that. I just don't think it is right to say "simpler" when it involves many more steps and possible failure points. I balance the pros and cons differently, so I see a different best approach. But of course I respect your preference. 

But my main point here was that the current approach is not quite enough yet. `QStandardPaths::AppDataLocation` will return something different in the two clients. Something else is still needed to make it work.

src/client/common/client_common.cpp
ricab: Sure! I thought you wouldn't like that.

include/multipass/platform.h
ricab: Yes

ricab: Testing will need directions of where to look for the autostart stuff in other platforms too, be it file or registry paths. Duplicating info in the test did not feel right, nor did creating a separate function or variable for that.

Saviq: Yeah I don't have a better approach right now…

src/client/common/client_common.cpp
Saviq: Yeah it ultimately means there's a problem with how the app is being run / packaged, so this should be and error, IMO.

src/platform/platform_linux.cpp
ricab: OK, crossed comments and edits... 

It makes me a bit itchy to give two different programs the same "AppName". This would affect all QStandardPaths using the app name. How about we use `QStandardPaths::GenericDataLocation` instead:

*To obtain a path to store data to be shared with other applications, use QStandardPaths::GenericDataLocation*

We can append the `multipass` suffix manually when we want to share data. That allows both shared and app-specific data. For instance, imagine we want a separate settings file for the gui, we could still do that (not that it is needed, but just for the sake of example). Or if we wanted to place separate logs there or something...

Saviq: To me it really depends on the definition of AppName, but I suppose we're lacking that already.

I wanted to avoid coming up with our own version of `QStandardPaths::locate`, but maybe we really just need a `mp::StandardPaths` that has the properties we need. WFM.

ricab: @Saviq, I just tested and `QStandardPaths::locate` works with dirs and relative paths:
```c++
qDebug() << QStandardPaths::locate(QStandardPaths::GenericDataLocation, "testdir", QStandardPaths::LocateDirectory);
qDebug() << QStandardPaths::locate(QStandardPaths::GenericDataLocation, "testdir/testfile", QStandardPaths::LocateFile);
```

Works :slightly_smiling_face:

Saviq: Oh ok, lacking in documentation again, then… Yeah, that works, then.

src/client/common/client_common.cpp
ricab: Done.

Saviq:
@ricab macOS seems fine, it just fails linking because of the new symbols not existing there:
```
Undefined symbols for architecture x86_64:
  "multipass::platform::preliminary_gui_autostart_setup()", referenced from:
      multipass::client::preliminary_setup() in libclient_common.a(client_common.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

ricab:
@Saviq Yeah I know, I had seen it in the previous commit.

src/platform/platform_linux.cpp
ricab: @Saviq FYI I was doing some research into a possible alternative: install a symlink `multipass/bla.desktop` -> `multipass.gui/bla.desktop`. This would have allowed us to use a transparent app data path. 

It turned out to be a bit of a rabbit hole. There does not seem to be a nice way to achieve that. The best I got was something like

```cmake
set(autostart_filename multipass.gui.autostart.desktop)
set(install_dir ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR})
set(autostart_main_dir ${install_dir}/multipass.gui)
set(autostart_secondary_dir ${install_dir}/multipass)
install(CODE "execute_process( \
        COMMAND ${CMAKE_COMMAND} -E create_symlink \
        \"\${DESTDIR}${autostart_main_dir}/${autostart_filename}\" \
        \"\${DESTDIR}${autostart_secondary_dir}/${autostart_filename}\"   \
    )"
        )
```

Notice that I escaped `\${DESTDIR}`, to interpret it only when actually doing `make install DESTDIR=foo`. Unfortunately this does not seem to work. The executed process apparently does not inherit the environment `make install` is executed with and `DESTDIR` always expands to an empty string.

Reverting back to the previous plan.

Saviq: @ricab yeah, I think the `::GenericDataLocation` approach is sane. The only caveat with it is that it's wonky when we want a writable location (for `multipass.conf` for example). So some wrapper around `QStandardPaths` may be in order at some point, so that we can easily and reliably determine a location common for all of Multipass, not each binary on its own.

ricab: @Saviq ack.

Separate issue: I don't think the proposed change to snapcraft.yaml will work, for two reasons:

- `make install` actually installs it to `/usr/local` by default (see [this](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html))
- it would overwrite the entry `/var/lib/snapd/desktop`, which snapd presumably needs:

```console
$ snap run --shell multipass
$ echo $XDG_DATA_DIRS
/usr/local/share:/usr/share:/var/lib/snapd/desktop
```

Unless you see an issue with that, I am changing it to:

```yaml
XDG_DATA_DIRS: $SNAP/usr/local/share:$XDG_DATA_DIRS
```

Saviq: @ricab I confirmed this worked, though?
```
ls -lh /snap/multipass/current/usr/share/multipass
razem 512
-rw-r--r-- 1 root root 117 wrz  9 10:06 multipass.gui.autostart.desktop
```

Saviq: @ricab and we can override `snapd`'s value, it's not for its consumption, but for the snap's - and we know we won't need other app's desktop files. And if we need it, we can {pre,ap}pend.

Saviq: @ricab it's `CMAKE_INSTALL_PREFIX` that defaults to `/usr/local`, but it [gets overriden by Snapcraft](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/cmake.py#L116).

ricab: @Saviq I see, I'll revert it then.

ricab:
@Saviq this ought to address all pending items. Let me know if I missed something.

src/platform/platform_linux.cpp
```diff
@@ -42,29 +42,39 @@ namespace mu = multipass::utils;
...
+QString find_desktop_target()
+{
+    const auto target_subpath = QDir{mp::client_name}.filePath(autostart_filename);
+    const auto target_path = QStandardPaths::locate(QStandardPaths::GenericDataLocation, target_subpath);
+
+    if (target_path.isEmpty())
+        throw std::runtime_error(fmt::format(
+            "could not locate the autostart .desktop file '{}', tried:\n  {}", autostart_filename,
+            (QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation) << QStringLiteral("\b\b\b"))
+                .join(QString{"/%1\n  "}.arg(target_subpath))));
```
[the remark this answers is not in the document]
ricab: OK, I did this by throwing a specific exception, separating the main error message from the detail. I also dropped the (stylish but) hackish abuse of `join` for formatting that actually needs to be applied to each line.

ricab:
@Saviq I did as you requested, but that involved a bit of new code. Could you please have a look and confirm you're good with it?

bors delegate=Saviq

Saviq:
Yup.

bors r+

townsend2010:
Since Travis has been flaky lately and the Mac build fails, will just merge this manually.

ricab:
Thanks @townsend2010 I had actually restarted the faulty build, but there was no need, since the problem was just an unexpected 404 during `apt upgrade`.

PR #994, opened 2019-08-26
[list] sort list reply alphabetically (Fixes #836)

### LINKED ISSUE #836: Sort `multipass list` output
```
$ multipass list
trusty-190611-1542      Running           10.76.88.44      Ubuntu 14.04 LTS
trusty-190611-1535      Running           10.76.88.69      Ubuntu 14.04 LTS
trusty-190611-1539      Running           10.76.88.59      Ubuntu 14.04 LTS
trusty-190611-1529      Running           10.76.88.101     Ubuntu 14.04 LTS
```

These names are based on launch time, so I'm not seeing either alphabetical or launch-time ordering.

Saviq:
As part of that, refactor the formatter test suite into a parametrized one.

---

Take that, C++ tests!

### DIFF
--- a/tests/test_output_formatter.cpp
+++ b/tests/test_output_formatter.cpp
@@ -320,10 +320,6 @@ private:
     std::locale saved_locale;
 };
 
-class CSVFormatter : public LocaleSettingTest
-{
-};
-
 class YamlFormatter : public LocaleSettingTest
 {
 };
@@ -354,6 +350,7 @@ auto print_petenv_param_name(const testing::TestParamInfo<PetenvFormatterSuite::
 
 const mp::TableFormatter table_formatter;
 const mp::JsonFormatter json_formatter;
+const mp::CSVFormatter csv_formatter;
 const auto empty_list_reply = mp::ListReply();
 const auto single_instance_list_reply = construct_single_instance_list_reply();
 const auto multiple_instances_list_reply = construct_multiple_instances_list_reply();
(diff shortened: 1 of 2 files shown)


tests/test_output_formatter.cpp
```diff
@@ -320,6 +347,22 @@ TEST_F(TableFormatter, multiple_instance_list_output)
...
+    auto list_reply = construct_unsorted_list_reply();
+
+    auto expected_table_output = "Name                    State             IPv4             Image\n"
+                                 "trusty-190611-1529      Deleted           --               Ubuntu N/A\n"
+                                 "trusty-190611-1535      Stopped           --               Ubuntu N/A\n"
+                                 "trusty-190611-1539      Suspended         --               Ubuntu N/A\n"
+                                 "trusty-190611-1542      Running           --               Ubuntu N/A\n";
+
+    mp::TableFormatter table_formatter;
+    auto output = table_formatter.format(list_reply);
```
ricab: To test treatment of petenv, you could do something like the following (did not actually try it):

```c++
EXPECT_CALL(mpt::MockSettings::mock_instance(),
        get(Eq(mp::petenv_key))).WillOnce(Return("trusty-190611-1539"));
```

And get that entry first in the expected output `trusty-190611-1539`.

tests/test_output_formatter.cpp
```diff
@@ -871,6 +914,22 @@ TEST_F(CSVFormatter, multiple_instance_list_output)
...
+                           "trusty-190611-1529,Deleted,,,N/A\n"
+                           "trusty-190611-1535,Stopped,,,N/A\n"
+                           "trusty-190611-1539,Suspended,,,N/A\n"
+                           "trusty-190611-1542,Running,,,N/A\n";
+
+    mp::CSVFormatter csv_formatter;
+    auto output = csv_formatter.format(list_reply);
+
+    EXPECT_EQ(output, expected_output);
+}
```
ricab: ditto

tests/test_output_formatter.cpp
```diff
@@ -1070,6 +1129,37 @@ TEST_F(YamlFormatter, multiple_instance_list_output)
...
+                           "  - state: Running\n"
+                           "    ipv4:\n"
+                           "      - \"\"\n"
+                           "    release: N/A\n";
+
+    mp::YamlFormatter formatter;
+    auto output = formatter.format(list_reply);
+
+    EXPECT_EQ(output, expected_output);
+}
```
ricab: ditto

tests/test_output_formatter.cpp
Saviq: I tend to keep unit tests single-purpose, this one tests that (non-petenv) instances are sorted alphabetically. I'd like to keep it that way.

If theory we could extend the `…_petenv_first` tests a bit, but then again those test exactly what they should :)

ricab: Unit tests are indeed supposed to test "units" of functionality, but where you decide the unit is can often subjective. IMO it would be perfectly appropriate to consider the entire sorting function as the unit here.

The lexicographical ordering is implemented elsewhere, so this really only tests whether it is being called properly on our side. If you then test the petenv comes first without lexicographical ordering of the remaining entries, nothing guarantees that they work together: you could write code that passes both tests without doing both things simultaneously ("if petenv in, put it first, else sort"). IOW the conjunction of two bits of functionality is itself functionality that requires testing. 

In practice some parsimony is needed when deciding unit boundaries, lest we end up testing single instructions :wink:

Saviq: Sure, which is why I suggested adding another test covering the conjuction of the two. And will do that.

I still find it important to know whether it's the alphabetical sorting, or the petenv precedence, or the conjunction of them failed, so all three tests make sense, IMO.

ricab: OK, I am fine with separate testing for the conjunction. 

You may want to consider parameterized tests to reduce redundancy &ndash; they could check for both properties but be fed input in those 3 separate conditions. It would also be easy to add coverage for other limit cases too (like empty input, only petenv, only a single non-petenv).

Saviq:
@ricab @townsend2010 this should be better then.

tests/test_output_formatter.cpp
```diff
@@ -276,1056 +295,697 @@ class LocaleSettingTest : public testing::Test
...
-};
-
-class JsonFormatter : public LocaleSettingTest
-{
-};
-
-class CSVFormatter : public LocaleSettingTest
-{
-};
+typedef std::tuple<const mp::Formatter*, const ::google::protobuf::Message*, std::string, std::string>
```
ricab: It's a bit difficult to track what each element means (here and in the petenv suite). Could we maybe have a comment to help the reader?

tests/test_output_formatter.cpp
```diff
@@ -276,1056 +295,697 @@ class LocaleSettingTest : public testing::Test
...
+     "    remote: snapcraft\n",
+     "yaml_find_multiple_duplicate_image"}};
+} // namespace
+
+TEST_P(FormatterSuite, properly_formats_output)
+{
+    const auto param = GetParam();
+    const auto formatter = std::get<0>(param);
+    const auto reply = std::get<1>(param);
+    const auto expected_output = std::get<2>(param);
```
ricab: Did you mean to copy `param` and then its contents? Not that it matters much, I suppose...

tests/test_output_formatter.cpp
Saviq: Done.

tests/test_output_formatter.cpp
Saviq: Moved to structured bindings now. Could be even nicer with `[[maybe_unused]]` and nesting, but not possible yet.

PR #1007, opened 2019-08-28
[settings] Drop leftover include after c++17 move

ricab:
Submitting this separately just for the sake of focused PRs.


ricab:
Header not required anymore after removal of unnecessary `unique_ptr` in C++17.

PR #1008, opened 2019-08-28
[submodule] Add scope_guard

ricab:
Can I?

@Saviq I know we discussed copy-pasting the header instead in Malta. I ended up not needing it in that occasion, so dropped it. Now I again needed it and thought I might as well add it as a submodule, to pick up any updates (e.g. to deal with compiler bugs).


Saviq:
@ricab sure, but maybe not standalone like this, but rather when it actually is used?

ricab:
@Saviq OK, I will drop this then and include it in the auto-start PR. Only leaving it around for a bit as no-merge to use CI.

PR #1013, opened 2019-08-30
backend/libvirt: Fix not detecting new instance state

### LINKED ISSUE #993: [libvirt] fails to notice instance is already running
```
$ virsh list
 Id   Name                   State
--------------------------------------
 5    snapcraft-subsurface   running

$ multipass ls                       
snapcraft-subsurface    Stopped           --               Ubuntu Snapcraft builder for Core 18

$ multipass start snapcraft-subsurface
start failed: Requested operation is not valid: domain is already running
```

townsend2010:
Better detect when an instance is started/stopped outside of Multipass.

Fixes #993

### DIFF
--- a/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
+++ b/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
@@ -209,7 +209,12 @@ auto domain_state_for(const mp::VirtualMachine::State& current_instance_state, v
     if (virDomainHasManagedSaveImage(domain, 0) == 1)
         return mp::VirtualMachine::State::suspended;
 
-    if (domain_state == VIR_DOMAIN_SHUTOFF)
+    // Most of these libvirt domain states don't have a Multipass instance state
+    // analogue, so we'll treat them as "off".
+    const auto domain_off_states = {VIR_DOMAIN_BLOCKED, VIR_DOMAIN_PAUSED,  VIR_DOMAIN_SHUTDOWN,
+                                    VIR_DOMAIN_SHUTOFF, VIR_DOMAIN_CRASHED, VIR_DOMAIN_PMSUSPENDED};
+
+    if (std::find(domain_off_states.begin(), domain_off_states.end(), domain_state) != domain_off_states.end())
         return mp::VirtualMachine::State::off;
 
     if (domain_state == VIR_DOMAIN_RUNNING && current_instance_state == mp::VirtualMachine::State::off)
@@ -395,7 +400,7 @@ void mp::LibVirtVirtualMachine::ensure_vm_is_running()
         state = State::off;
         state_wait.notify_all();
 
-        throw mp::StartException(vm_name, "Instance shut down while starting");
+        throw mp::StartException(vm_name, "Instance failed to start");
     }
 }
 
--- a/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
+++ b/src/platform/backends/libvirt/libvirt_virtual_machine.cpp
@@ -199,7 +199,7 @@ auto domain_by_definition_for(const mp::VirtualMachineDescription& desc, const s
     return domain;
 }
 
-auto domain_state_for(const mp::VirtualMachine::State& current_instance_state, virDomainPtr domain)
+auto refresh_instance_state_for_domain(virDomainPtr domain, const mp::VirtualMachine::State& current_instance_state)
 {
     auto domain_state{0};
 
@@ -266,7 +266,7 @@ void mp::LibVirtVirtualMachine::start()
     else
         domain = domain_by_name_for(vm_name, connection.get());
 
-    state = domain_state_for(state, domain.get());
+    state = refresh_instance_state_for_domain(domain.get(), state);
     if (state == State::running)
         return;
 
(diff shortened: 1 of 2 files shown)


src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -199,22 +199,33 @@ auto domain_by_definition_for(const mp::VirtualMachineDescription& desc, const s
...
-    {
-        return mp::VirtualMachine::State::unknown;
-    }
+    if (domain_state == VIR_DOMAIN_SHUTOFF)
+        return mp::VirtualMachine::State::off;
+
+    if (domain_state == VIR_DOMAIN_RUNNING && current_instance_state == mp::VirtualMachine::State::off)
+        return mp::VirtualMachine::State::running;
+
+    return current_instance_state;
```
gerboland: Why are you passing Multipass' thinking of the current state through? Can we not read all possible states directly from libvirt?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -199,22 +199,33 @@ auto domain_by_definition_for(const mp::VirtualMachineDescription& desc, const s
...
         return mp::VirtualMachine::State::unknown;
 
     if (virDomainHasManagedSaveImage(domain, 0) == 1)
         return mp::VirtualMachine::State::suspended;
 
-    if (virDomainGetState(domain, &state, nullptr, 0) == -1)
-    {
-        return mp::VirtualMachine::State::unknown;
-    }
+    if (domain_state == VIR_DOMAIN_SHUTOFF)
```
gerboland: Why don't you deal with the other possible states like SHUTDOWN, CRASHED, PAUSED & BLOCKED?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -380,17 +388,14 @@ void mp::LibVirtVirtualMachine::ensure_vm_is_running()
...
     {
         // Have to set 'off' here so there is an actual state change to compare to for
         // the cond var's predicate
         state = State::off;
         state_wait.notify_all();
-        if (present_state == VirtualMachine::State::unknown)
-            throw mp::StartException(vm_name, "Unable to determine state");
 
-        throw mp::StartException(vm_name, "Instance shutdown during start");
+        throw mp::StartException(vm_name, "Instance shut down while starting");
```
gerboland: I dunno, maybe "failed to start" ? Since a crash would result in this too, no?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: There are a number of Multipass states that have no analogous `libvirt` states such as `DELAYED_SHUTDOWN`, `STARTING`, `SUSPENDING`, and Multipass's way of handling `UNKNOWN`. So in this particular piece of code, this is the "fallthrough" in which we just return what Multipass has stored since no libvirt state matches.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: I suppose I should account for those libvirt states as well.  I'll add them.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: Yeah, I suppose being more generic here will be better.

townsend2010:
> And can this fix get a test?

I'll think about how best to go about adding a test that mimics the bug's behavior.  I will say, this PR adds a significant amount more coverage, so the current tests actually test some of this.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: Ok, I have accounted for all of the domain states and basically set them to `off` for the returned instance state.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: Done

townsend2010:
>> And can this fix get a test?

>I'll think about how best to go about adding a test that mimics the bug's behavior. I will say, this PR adds a significant amount more coverage, so the current tests actually test some of this.

Ok, I added a test that should mimic the behavior that caused #993.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
gerboland: I see. Could I request a function name change of `domain_state_for` to something like `resolve_domain_state_with_current` ?

gerboland:
Thanks for the improvements & test. I've one niggle, but am otherwise happy

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
townsend2010: I kept thinking the function name morphed into a weak name, and as we all know, naming functions (and variables) is one of the hardest software engineering principles :grinning:, I was drawing a blank on what to rename it.  I'll change it for sure.

townsend2010: Ok, I renamed it.  Let me know what you think.

PR #1035, opened 2019-09-10
Change LinuxProcess into more platform independent BasicProcess

gerboland:
Platform specific behaviours can be implemented by subclassing it. `BasicProcess` itself just uses `QProcess` internally, so is cross-platform.

Some other minor changes
- fixes stderr printing to console not clearing `readAllStandardError()`
- remove `ProcessFactory` method that can operate on program name & argument, instead have intermediary `simple_process_spec` function to enable that usage. This eases testing of `BasicProcess` later

This will enable testing `BasicProcess` on all platforms.


tests/test_with_mocked_bin_path.cpp
```diff
@@ -18,36 +18,23 @@
...
-} // namespace
-
 void mpt::TestWithMockedBinPath::SetUp()
 {
-    old_path = getenv("PATH");
-    std::string new_path{mpt::mock_bin_path()};
-    new_path.append(":");
-    new_path.append(old_path);
-    set_path(new_path);
+#ifdef WIN32
```
townsend2010: Let's avoid platform ifdef's. You could use QDir::listSeparator() here instead.

gerboland: Darn, it was right there under my nose. Thanks!

townsend2010: Cool, thanks!

Issue #1043, opened 2019-09-11
Unnecessary error message printed to client


gerboland:
I see
```
[error] [client] Failed to set up autostart prerequisites: could not locate the autostart .desktop file 'multipass.gui.autostart.desktop
```
on every command launch with multipass currently (running from build dir).

I think a warning to the logs would be more appropriate

Saviq:
The problem is this suggests a packaging problem when installed.

You can see in debug logs where it's looking for it and create it, or set `XDG_DATA_DIRS` to point somewhere you have the file. Unless there's other ideas of making this obvious when installed?

ricab:
@gerboland @townsend2010 

Taking the discussion here to address a few points together.

- the title of the issue suggests confusion: this does not come from the daemon, it really is an error that arises in the clients
- it is logged using our normal logging facilities; they happen to sent it to the terminal
- as @Saviq pointed out, this should not happen with normal installation (neither snap nor `make install`). We do want to know about it otherwise. There are simple ways to make it work from the build directory
- anticipating dislike, I initially had this as a WARNING, even though I agree with Saviq that ERROR is the right category: it is an unexpected problem that prevents a feature from working (something that prevents the app from running would be a FATAL). I changed it on his request.
- this is a consequence of having gone "the link way". I did point out it would introduce a new requirement on the clients. The plain write I had before did not have this problem, but it is less elegant for mixing data and logic.

PR #1045, opened 2019-09-11
Account for autostart file in dev instructions

ricab:


ricab:
With #991 multipass now needs to find the autostart desktop file (`multipass/multipass.gui.autostart.desktop`) in either `~/.local/share` or one of `XDG_DATA_DIRS` (elements in `QStandardPaths::GenericDataLocation`). 

This addresses outcome of discussion relating to #1043.

PR #1047, opened 2019-09-11
Compile fix for older compiler

gerboland:
Needed to fix VSC++17 error: 
```
error C2131: expression did not evaluate to a constant
note: failure was caused by non-constant arguments or reference to a non-constant symbol
```


ricab:
@gerboland well CI is now failing because of conflicts, but as mentioned in IRC that is what I had [before](https://github.com/CanonicalLtd/multipass/commit/8c42885ebbab6eae4a604c8e1566850b5718cd44#diff-955fee021b76c99185946f67e1571f95R115). Unfortunately [it failed](https://travis-ci.org/CanonicalLtd/multipass/jobs/581221612#L3855)

gerboland:
I must be using different version of clang so.

PR #1048, opened 2019-09-11
Fix autostart tests

ricab:
Autostart tests were failing when instructions in #1045 were followed. This is because they were not accounting for a matching desktop file in HOME. This fixes that.


townsend2010:
Doing this to get the proper bors commit message in staging and will manually merge staging at that point.

bors r+

PR #1064, opened 2019-09-17
Small Process API improvements/fixes

gerboland:
- If an error occurs, while ProcessError does say why, the errorString() can reveal more details as to what happened (e.g. on Windows it can return Windows API error messages that are useful)
- fix bug where ProcessError enum wasn't registered with the signal/slot mechanism


Saviq:
bors r=townsend20101

townsend2010:
Mac build failing for other reasons, so going to merge manually.

PR #1071, opened 2019-09-18
transfer: Use the sftp client now instead of scp

### LINKED ISSUE #1069: Switch to using sftp client code and remove scp client code
It's time to switch over to using the sftp client code and remove the scp client code.  SFTP is much more powerful and the preferred way of handling file transfers.

townsend2010:
- Remove the scp code.
- Clean up the sftp client code.

Fixes #1069

---
This should be functionally equivalent to the previous implementation.


townsend2010:
Hey @gerboland,

The motivation here is that we are basically carrying 2 separate code bases for doing the same thing.  However, the `scp` protocol is very limited and does not support streaming `stdin`/`stdout` which is why we use `sftp` for that.  But why still use `scp` for file transfer when we have `sftp` as well?

Also, the reason we didn't switch over to using `sftp` for file transfer immediately was that this was a community contribution and I needed to audit the code a bit more and I finally had some time to do so:grin:

Regarding the removal of the two tests, I removed these because after the code audit, we don't need to use `sftp_stat` for anything the the `SFTClient` and so what these tests are testing no longer applies.

I hope this helps explain things :smile:

ricab:
@townsend2010 what about splitting the sftp_client changes into a separate PR, so this one could go in right away? The other one could then deal with the error message.

townsend2010:
@ricab,

As mentioned earlier on IRC, some (most?) of the sftp client code changes are to fix (an) error(s) that I saw when testing this.  While I don't remember the exact details, I do know most of it is necessary to fix error conditions.  But I do know any cleanup went well with what issue I was trying to fix.  I think we should leave it as-is.

Also, I do agree about the error with directories.  We simply have not supported directory transfers and so that is all a separate issue.  It's just that `libssh`'s handling of the error between the scp protocol and sftp protocol are different.  We should enter an issue to support directory transfers.

PR #1073, opened 2019-09-19
Add AppArmor profile for sshfs_server

gerboland:


townsend2010:
Doing this just to get bors to add it's message to staging...

bors r+

PR #1074, opened 2019-09-19
Convert to strict snap

gerboland:
Requires on #1073 for mounts to function

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -47,6 +47,7 @@ apps:
     daemon: simple
     plugs:
       - kvm
+      - libvirt
       - multipass-support
       - network-bind
       - network-control
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -10,7 +10,7 @@ description: |
 license: GPL-3.0
 
 adopt-info: multipass
-confinement: classic
+confinement: strict
 base: core18
 
 architectures:
@@ -24,6 +24,11 @@ architectures:
 layout:
   /usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu:
     bind: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu
+
+plugs:
+  all-home:
+    interface: home
+    read: all
 
 apps:
   multipassd:
@@ -40,6 +45,14 @@ apps:
       DAEMON_CONFIG_HOME: *daemon-config # temporary
       XTABLES_LIBDIR: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/xtables
     daemon: simple
+    plugs:
+      - kvm
+      - multipass-support
+      - network-bind
+      - network-control
+      - network-manager
+      - firewall-control
+
   multipass:
     environment:
       <<: &client-environment
@@ -53,11 +66,17 @@ apps:
       DAEMON_CONFIG_HOME: *daemon-config # temporary
     command: bin/launch-multipass
     completer: etc/bash_completion.d/snap.multipass
+    plugs:
+      - all-home
+      - network
   gui:
     environment: *client-environment
     command: bin/launch-multipass-gui
     desktop: usr/share/applications/multipass.gui.desktop
     autostart: multipass.gui.autostart.desktop
+    plugs:
+      - all-home
+      - network
 
 parts:
   multipass:
@@ -90,6 +109,7 @@ parts:
     - libvirt0
     - dnsmasq-base
     - dnsmasq-utils
+    - apparmor
     source: .
     configflags:
     - -DCMAKE_BUILD_TYPE=RelWithDebInfo
@@ -124,6 +144,9 @@ parts:
       usr/lib/*/pulseaudio/libpulsecommon-*.so: usr/lib/
       usr/share/seabios/*: qemu/
       usr/lib/ipxe/qemu/*: qemu/
+    stage:
+    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libcurl-gnutls.so.4.5.0
+    - -usr/share/doc/libcurl3-gnutls/changelog.Debian.gz
 
   kvm-support:
     plugin: nil
@@ -152,6 +175,9 @@ parts:
     - -usr/share/doc/libfreetype6/TODO
     - -usr/share/doc/libfreetype6/changelog.Debian.gz
     - -usr/share/doc/libfreetype6/pcf/README
+    - -lib/$SNAPCRAFT_ARCH_TRIPLET/libexpat.so.1.6.7
+    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libexpatw.so.1.6.7
+    - -usr/share/doc/libexpat1/changelog.Debian.gz
 
   glue:
     plugin: dump
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,11 +25,6 @@ layout:
   /usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu:
     bind: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu
 
-plugs:
-  all-home:
-    interface: home
-    read: all
-
 apps:
   multipassd:
     command: bin/launch-multipassd
@@ -51,9 +46,7 @@ apps:
       - multipass-support
       - network-bind
       - network-control
-      - network-manager
       - firewall-control
-
   multipass:
     environment:
       <<: &client-environment
@@ -68,7 +61,7 @@ apps:
     command: bin/launch-multipass
     completer: etc/bash_completion.d/snap.multipass
     plugs:
-      - all-home
+      - home
       - network
   gui:
     environment: *client-environment
@@ -76,7 +69,7 @@ apps:
     desktop: usr/share/applications/multipass.gui.desktop
     autostart: multipass.gui.autostart.desktop
     plugs:
-      - all-home
+      - home
       - network
 
 parts:
@@ -145,9 +138,6 @@ parts:
       usr/lib/*/pulseaudio/libpulsecommon-*.so: usr/lib/
       usr/share/seabios/*: qemu/
       usr/lib/ipxe/qemu/*: qemu/
-    stage:
-    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libcurl-gnutls.so.4.5.0
-    - -usr/share/doc/libcurl3-gnutls/changelog.Debian.gz
 
   kvm-support:
     plugin: nil
@@ -176,9 +166,6 @@ parts:
     - -usr/share/doc/libfreetype6/TODO
     - -usr/share/doc/libfreetype6/changelog.Debian.gz
     - -usr/share/doc/libfreetype6/pcf/README
-    - -lib/$SNAPCRAFT_ARCH_TRIPLET/libexpat.so.1.6.7
-    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libexpatw.so.1.6.7
-    - -usr/share/doc/libexpat1/changelog.Debian.gz
 
   glue:
     plugin: dump
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -71,6 +71,9 @@ apps:
     plugs:
       - home
       - network
+      - x11
+      - unity7
+      - wayland
 
 parts:
   multipass:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -46,13 +46,13 @@ apps:
       XTABLES_LIBDIR: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/xtables
     daemon: simple
     plugs:
+      - all-home
+      - firewall-control
       - kvm
       - libvirt
       - multipass-support
       - network-bind
       - network-control
-      - firewall-control
-      - all-home
   multipass:
     environment:
       <<: &client-environment
@@ -77,9 +77,9 @@ apps:
     plugs:
       - home
       - network
-      - x11
       - unity7
       - wayland
+      - x11
 
 parts:
   multipass:
@@ -99,6 +99,7 @@ parts:
     - qtbase5-dev
     - qtbase5-dev-tools
     stage-packages:
+    - apparmor
     - on amd64: [libgl1]
     - on armhf: [libgles2-mesa]
     - on arm64: [libgles2-mesa]
@@ -112,7 +113,6 @@ parts:
     - libvirt0
     - dnsmasq-base
     - dnsmasq-utils
-    - apparmor
     source: .
     configflags:
     - -DCMAKE_BUILD_TYPE=RelWithDebInfo
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,6 +25,11 @@ layout:
   /usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu:
     bind: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu
 
+plugs:
+  all-home:
+    interface: home
+    read: all
+
 apps:
   multipassd:
     command: bin/launch-multipassd
@@ -47,6 +52,7 @@ apps:
       - network-bind
       - network-control
       - firewall-control
+      - all-home
   multipass:
     environment:
       <<: &client-environment
--- a/src/platform/backends/shared/linux/process_factory.cpp
+++ b/src/platform/backends/shared/linux/process_factory.cpp
@@ -68,11 +68,6 @@ mp::optional<mp::AppArmor> create_apparmor()
         mpl::log(mpl::Level::warning, "apparmor", "AppArmor disabled by environment variable");
         return mp::nullopt;
     }
-    else if (mp::utils::get_driver_str() == "libvirt")
-    {
-        mpl::log(mpl::Level::info, "apparmor", "libvirt backend disables Multipass' AppArmor support");
-        return mp::nullopt;
-    }
     else
     {
         try
--- a/src/platform/backends/shared/linux/process_factory.cpp
+++ b/src/platform/backends/shared/linux/process_factory.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2019-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -21,6 +21,7 @@
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/process_spec.h>
+#include <multipass/snap_utils.h>
 #include <multipass/utils.h>
 
 namespace mp = multipass;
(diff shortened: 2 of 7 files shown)


gerboland:
CI failing the snap due to requiring human approval to autoconnect to multipass-support and home/all slots.

snap/snapcraft.yaml
```diff
@@ -39,6 +44,14 @@ apps:
       XDG_CONFIG_HOME: &daemon-config $SNAP_DATA/config
       DAEMON_CONFIG_HOME: *daemon-config # temporary
     daemon: simple
+    plugs:
```
townsend2010: We're gonna need to include the `libvirt` plug as well, right?

snap/snapcraft.yaml
```diff
@@ -115,6 +135,9 @@ parts:
       usr/share/seabios/vgabios-stdvga.bin: qemu/
       usr/share/seabios/kvmvapic.bin: qemu/
       usr/lib/ipxe/qemu/efi-virtio.rom: qemu/
+    stage:
```
townsend2010: Curious why these need to be unstaged?

snap/snapcraft.yaml
```diff
@@ -143,6 +166,9 @@ parts:
     - -usr/share/doc/libfreetype6/TODO
     - -usr/share/doc/libfreetype6/changelog.Debian.gz
     - -usr/share/doc/libfreetype6/pcf/README
+    - -lib/$SNAPCRAFT_ARCH_TRIPLET/libexpat.so.1.6.7
```
townsend2010: Same here, why these 3 are being unstaged now?

snap/snapcraft.yaml
gerboland: My local snapcraft runs claims 2 parts supply these files but with different contents. Dunno why

snap/snapcraft.yaml
gerboland: Same as above. Just to make it build really

snap/snapcraft.yaml
gerboland: Quite right, added

snap/snapcraft.yaml
townsend2010: Ack, ok

snap/snapcraft.yaml
townsend2010: Alright

Saviq:
The stage exclusions shouldn't be necessary, it just means your build environment was stale. A `snapcraft clean <part>` should help.

gerboland:
> The stage exclusions shouldn't be necessary, it just means your build environment was stale. A `snapcraft clean <part>` should help.

I tried that, but perhaps lxd backend isn't as good. I just wanted the thing to build.

gerboland:
> Also, there is still the check for `libvirt` that disables the AppArmor profile for Multipass:
> 
> https://github.com/CanonicalLtd/multipass/blob/1d58dc522a103b1e546bdd5c6aff0833d87be401/src/platform/backends/shared/linux/process_factory.cpp#L62
> 
> That block needs to be removed.

That will require a little more that just removing that branch. libvirt implements the apparmor protections itself, so the only things multipass would need to apparmor itself are likely the qemu-img calls.

townsend2010:
I should add that it looks like the `sshfs_server` process gets a slew of apparmor denials in the journal.  I think this is to do with the aforementioned logical `libvirt` branch that disables AppArmor support.

townsend2010:
Another issue I've noticed is if a mount is active and the instance is shutdown, there is an error when the `sshfs_server` profile is unloaded which leads to an unhandled exception in `multipassd` causing it to terminate.  Need to address this as well.

townsend2010:
As I'm taking over this PR, the two issues I've found thus far that need fixing are:
* Mounts don't work when using the `libvirt` backend
* Stopping an instance with an active mount leads to an unhandled exception causing `multipassd` to terminate.

townsend2010:
If I remove the following line:
https://github.com/CanonicalLtd/multipass/blob/eb5fb696dddd06fbf4d107c76e4025ccd7a035f7/src/platform/backends/shared/linux/process_factory.cpp#L62-L66
then it seems everything works fine with mounts with the `libvirt` backend.  I also did not observe any regressions or negative behavior with apparmor enabled using the `libvirt` backend.

I'm now confused as to why this was done and the push back I received when I mentioned this needs to be removed in https://github.com/CanonicalLtd/multipass/pull/1074#pullrequestreview-291114611.

I'll test further just to make sure I didn't miss something.

Still have the abort to deal with when stopping an instance with an active mount.

Saviq:
> I'm now confused as to why this was done and the push back I received when I mentioned this needs to be removed in [#1074 (review)](https://github.com/CanonicalLtd/multipass/pull/1074#pullrequestreview-291114611).

I think the "push back" was that that just wasn't enough to properly confine the libvirt backend. Not that it was the wrong thing to do :)

townsend2010:
> I think the "push back" was that that just wasn't enough to properly confine the libvirt backend. Not that it was the wrong thing to do :)

Well, just removing those lines and the `libvirt` backend does indeed appear to work properly, so it seems to me it is enough to properly confine the `libvirt` backend.  So I'm still confused regarding it not being enough to just remove these lines.  I looked at the code in my original review and thought it should be enough, which is why I mentioned it in the first place.  What am I missing?

townsend2010:
The abort when stopping an instance with an active mount is not new to this PR.  I'm going to split that out as a separate issue and not fix it with this PR.

Saviq:
> Well, just removing those lines and the `libvirt` backend does indeed appear to work properly, so it seems to me it is enough to properly confine the `libvirt` backend. So I'm still confused regarding it not being enough to just remove these lines. I looked at the code in my original review and thought it should be enough, which is why I mentioned it in the first place. What am I missing?

It sounded to me like `qemu-img` wasn't confined back then? That was the concern IIUC?

townsend2010:
> It sounded to me like qemu-img wasn't confined back then? That was the concern IIUC?

That was an example of my concerns.  My main concern was that when we moved to a strict snap, then the `libvirt` backend wouldn't work due to apparmor denials on external apps that `multipassd` uses.  And this ended up being the case, but only with `sshfs_server`.

I looked at the code during the review and *thought* that just removing the aforementioned lines would be enough to make it work. Gerry said there was more to it than just removing those lines, but no real explanation as to what the extra work entails.

That said, I didn't think we should merge this knowing it will break the `libvirt` backend, so I began trying to fix it.  Since my initial analysis led me to believe that all I needed to do was to remove those lines, that is what I did.  Lo and behold, it does seem to work.  Gerry worked on this for quite some time, so I trust him when he said there is more to it, but I simply don't know what more there is to it.  I'm uncomfortable with just removing those lines without knowing the reasoning behind saying there is more to it. I suppose we can be reactive instead of proactive and Just Do It and wait for any bug reports saying something is broken :grin:

I hope this explains my reasoning and concerns about this.  Anybody have any other thoughts about moving forward with this?

Saviq:
@townsend2010 I think the thing to do is have a look at all the processes `multipassd` spawns and ensure they're all confined, is all.

snap/snapcraft.yaml
```diff
@@ -119,6 +140,9 @@ parts:
       usr/share/seabios/vgabios-stdvga.bin: qemu/
       usr/share/seabios/kvmvapic.bin: qemu/
       usr/lib/ipxe/qemu/efi-virtio.rom: qemu/
+    stage:
+    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libcurl-gnutls.so.4.5.0
+    - -usr/share/doc/libcurl3-gnutls/changelog.Debian.gz
```
Saviq: We should remove these lines, I believe these were a build env problem.

snap/snapcraft.yaml
```diff
@@ -25,6 +25,11 @@ layout:
   /usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu:
     bind: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/qemu
 
+plugs:
+  all-home:
+    interface: home
+    read: all
```
Saviq: Hmm why would we need this, instead of the plain `home`? (ref. https://snapcraft.io/docs/home-interface)

townsend2010:
@Saviq,

lol, that is what happens when I remove those lines.

I think I'm just going to commit the removal of those lines and be done with until further review :wink:

snap/snapcraft.yaml
```diff
@@ -147,6 +171,9 @@ parts:
     - -usr/share/doc/libfreetype6/TODO
     - -usr/share/doc/libfreetype6/changelog.Debian.gz
     - -usr/share/doc/libfreetype6/pcf/README
+    - -lib/$SNAPCRAFT_ARCH_TRIPLET/libexpat.so.1.6.7
+    - -usr/lib/$SNAPCRAFT_ARCH_TRIPLET/libexpatw.so.1.6.7
+    - -usr/share/doc/libexpat1/changelog.Debian.gz
```
Saviq: ditto

snap/snapcraft.yaml
```diff
@@ -40,6 +45,15 @@ apps:
       DAEMON_CONFIG_HOME: *daemon-config # temporary
       XTABLES_LIBDIR: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/xtables
     daemon: simple
+    plugs:
+      - kvm
+      - libvirt
+      - multipass-support
+      - network-bind
+      - network-control
+      - network-manager
```
Saviq: I _think_ we should be able to do without this.

snap/snapcraft.yaml
townsend2010: I *think* this is here to de-duplicate some `plugs` definitions for both `multipass` and `multipass-gui`.  I think...

Saviq: Well, sure, but we wouldn't need to deduplicate if we didn't need the `read: all` bit. That's really what I'm referring to.

From https://snapcraft.io/docs/home-interface:
> - `read` (plug):
    optional, when set to ‘all’ allow reading non-hidden files in home directories of all users as traditional file permissions allow.

Why would either `multipass` or `multipass-gui` need to read from `/home/*/` rather than just `$HOME`?

townsend2010: Oh, I see what you mean.  But it might be because the mount runs through `multipassd` so it's the `root` user and needs access the user's actual home directory?  Again, I'm not really sure, so I'll have to experiment with it if you think this is really unnecessary.

townsend2010: Right after I wrote that, I think I'm wrong :grin: I'll experiment.

Saviq: The only thing the client does is resolve the path, so I don't think that's true. And even if the resolution required access to the mount source, that doesn't cover all the cases for mounts outside of user homes… Some other commands like `transfer` (obviously) and `launch` (for `--cloud-init`) actually access the files client-side, but `read: all` doesn't really help there, does it :)

snap/snapcraft.yaml
townsend2010: I only removed the `network-manager` plug here, which is what I think you are referring to.

snap/snapcraft.yaml
```diff
@@ -40,6 +40,14 @@ apps:
       DAEMON_CONFIG_HOME: *daemon-config # temporary
       XTABLES_LIBDIR: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/xtables
     daemon: simple
+    plugs:
+      - kvm
+      - libvirt
+      - multipass-support
+      - network-bind
+      - network-control
+      - firewall-control
```
Saviq: So we need `kvm`, `network-control` and `firewall-control` autoconnected. `network-bind` [already is](https://snapcraft.io/docs/network-bind-interface), and we just got `multipass-support`, too.

Saviq: > `<jdstrand> done`

snap/snapcraft.yaml
Saviq: Yup.

townsend2010:
Some testing notes:
- Built the snap package locally and installed it.
- Manually connect some plugs/slots.  See `snap connections multipass`.
- Created a core18 instance.
- Transferred the snap package to the instance and installed it there as well.
- Manually connect the plugs/slots like before.
- The `multipass` client could not connect to the daemon.  This is due to the daemon's socket being owned by `root.sudo 660` and on core, you cannot add the `ubuntu` user to the `sudo` group.  Instead, changing the socket's owner to `root.lxd` will get by this.
- Created an 18.04 instance in the core18 instance.
- Profit!

Also, tested `snapcraft` with the strict Multipass snap installed and it errors out:
```
$ snapcraft 
Launching a VM.
Launched: snapcraft-multipass                                                   
error: no changes of type "auto-refresh" found                                  
2020-03-07T08:06:59Z INFO Waiting for restart...
snapd 2.43.3 from Canonical✓ installed
Source path "/tmp/tmpufpsfrak/core18.snap" does not exist
An error occurred with the instance when trying to copy files with 'multipass': returned exit code 1.
```
Changes will be necessary in snapcraft.

Saviq:
@townsend2010 I can't get mounts to work now:
```console
$ multipass start
$ multipass exec primary -- ls Home/
# nothing
```

And `dmesg`:
```
[150326.786403] audit: type=1400 audit(1584726078.383:3248): apparmor="DENIED" operation="capable" profile="snap.multipass.multipassd" pid=2430460 comm="sshfs_server" capability=1  capname="dac_override"
[150383.603852] audit: type=1400 audit(1584726135.203:3249): apparmor="DENIED" operation="open" profile="snap.multipass.multipass" name="/etc/ssh/ssh_config" pid=2432073 comm="multipass" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
[150384.115318] audit: type=1400 audit(1584726135.715:3250): apparmor="DENIED" operation="open" profile="snap.multipass.multipassd" name="/home/michal/snap/multipass/1819/" pid=2430460 comm="sshfs_server" requested_mask="r" denied_mask="r" fsuid=0 ouid=1000
```

Nothing interesting in the journal…

Saviq:
I can't seem to get the GUI to start either:
```
$ multipass.gui         
qt.qpa.screen: QXcbConnection: Could not connect to display :0
Could not connect to any X display.
```

dmesg:
```
[36360.114187] audit: type=1400 audit(1584965568.560:843): apparmor="DENIED" operation="connect" profile="snap.multipass.gui" pid=509771 comm="multipass.gui" family="unix" sock_type="stream" protocol=0 requested_mask="send receive connect" denied_mask="send connect" addr=none peer_addr="@/tmp/.X11-unix/X0" peer="unconfined"
```

townsend2010:
@Saviq, ok, will look into these.  Thanks!

townsend2010:
Well, I get a different outcome, but is still not what we want:
```
$ multipass start
Launched: primary                                                               
Mounted '/home/townsend/snap/multipass/1863' into 'primary:Home'
$ multipass exec primary -- ls Home
cache  config  data
$
```

Regarding the gui, yeah, we'll need to add some additional interface connections such as `unity`, `wayland`, and `x11`.

townsend2010:
If I `multipass umount primary:Home` and then do `multipass mount $HOME primary:Home`, then it works as expected for me.  Seems we are using something that resolves to a different path when doing the automatic `primary` mount.

@Saviq, yeah, I think something is messed up on your system regarding mounts.  It's working as expected here with the strict snap.

Saviq:
> If I `multipass umount primary:Home` and then do `multipass mount $HOME primary:Home`, then it works as expected for me. Seems we are using something that resolves to a different path when doing the automatic `primary` mount.

`multipass info primary`?

> @Saviq, yeah, I think something is messed up on your system regarding mounts. It's working as expected here with the strict snap.

D'oh… I shot myself in the foot with https://github.com/canonical/multipass/issues/1423#issuecomment-597790138. I wonder if we can/should bail if we're confined, but this is set.

townsend2010:
We are using `QDir::homePath()` and that resolves to `$HOME`, which, within a snap, resolves to `/home/$USER/snap/multipass/$VERSION` which is `/home/townsend/snap/multipass/1863` currently on my system.

Saviq:
> We are using `QDir::homePath()`

Yes, we need something better… when in a snap, we'll need to do the equivalent of `getent passwd $USER`…

Saviq:
> Yes, we need something better… when in a snap, we'll need to do the equivalent of `getent passwd $USER`…

This can probably go to our CLI and GUI command wrappers. Something along the ways of `export HOME=$( getent passwd $USER | cut -d: -f6)`. Need to ensure we're not reading/writing from the standard `$HOME` locations, though.

Maybe we should rather do `export SNAP_REAL_HOME=…` instead, and use that over `::homePath()`.

townsend2010:
The `multipass.gui` issue should be taken care of now.

townsend2010:
> Need to ensure we're not reading/writing from the standard $HOME locations, though.

Not sure what you mean by this.  I thought the whole point of mounting $HOME was to have all of the user's $HOME available as a mount in the primary instance for reading & writing.  I'm probably misunderstanding this though.

> Maybe we should rather do export SNAP_REAL_HOME=… instead, and use that over ::homePath()

Yeah, we'll have to quirk this for just snap's since we still want `::homePath()` on other platforms (and Linux source builds).

townsend2010:
> Need to ensure we're not reading/writing from the standard $HOME locations, though.

Oh, I think I know what you mean.  You mean don't mess with $HOME within the snap, but instead use some other env var such as `SNAP_REAL_HOME`, right?

townsend2010:
Ok, I pushed a change that makes this work.

townsend2010:
The proper home directory should be mounted by default now in the `primary` instance.  The `multipass.gui` should start now.  Lastly, it the `DISABLE_APPARMOR` env var is set, it will be ignored when running as a snap.

src/platform/backends/shared/linux/process_factory.cpp
```diff
@@ -63,16 +64,11 @@ class AppArmoredProcess : public mp::BasicProcess
 
 mp::optional<mp::AppArmor> create_apparmor()
 {
-    if (qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
```
Saviq: I think we should error out here. If someone sets this var, they're unlikely to want it ignored :)

townsend2010: Hmm, this is really just an internal env var (used for debugging/testing) that really makes no sense in a strict snap.  I think if someone is expecting Apparmor to be disabled for a strict snap, then they really don't understand snaps at all.  I think just being nice here and ignoring this scenario is the "nice" thing to do.

Saviq: Hmm I could've sworn I've answered this… I just worry ignoring it may lead to confusion (similar to when I had it set last week)… I'd rather it fails here _if_ we're actually confined (rather than _if_ we're in snap). But it's not a big deal - if it's difficult to detect if we're confined, it's OK to ignore, then.

townsend2010: @Saviq, we discussed this on a side channel :wink: I'll see if it's possible to tell if we are strict.  Otherwise, yeah, we should leave it as-is.

luis4a0:
I'm having an issue when launching a vm from a local file.
`multipass launch -n fedora file:///home/luis/Fedora-Cloud-Base-30-1.2.x86_64.qcow2
launch failed: Cannot read image format: qemu-img failed (Process returned exit code: 1) with output:
qemu-img: Could not open '/var/snap/multipass/common/data/multipassd/vault/instances/fedora/Fedora-Cloud-Base-30-1.2.x86_64.qcow2': Could not open '/var/snap/multipass/common/data/multipassd/vault/instances/fedora/Fedora-Cloud-Base-30-1.2.x86_64.qcow2': No such file or directory`

It seems it doesn't copy the file to `/var/snap/multipass/...`. Maybe the access to the filesystem outside the snap is forbidden.

townsend2010:
Hey @luis4a0!

Thanks for letting me know! I think I might know what's going on and will take a look.

snap/snapcraft.yaml
townsend2010: Actually, this is needed to allow `file:\\` based launches.  Without this, we get the issue reported in https://github.com/canonical/multipass/pull/1074#issuecomment-617162865.

Saviq:
Alright, let's get this in :)

bors merge

PR #1079, opened 2019-09-20
[desktop] make multipass.gui.desktop snap-agnostic

Saviq:
---
The main difference:
```
$ curl --unix-socket /run/snapd.socket http://localhost/v2/snaps/multipass | jq
…
    "apps": [
      {
        "snap": "multipass",
        "name": "gui"
      },
…
```

vs.

```
…
    "apps": [
      {
        "snap": "multipass",
        "name": "gui",
        "desktop-file": "/var/lib/snapd/desktop/applications/multipass_gui.desktop"
      },
…
```

Unfortunately [LP#1844791](https://bugs.launchpad.net/ubuntu/+source/gnome-software/+bug/1844791) means it doesn't yet work the way we want it to (Launch in Software Center should run the GUI).


data/multipass.gui.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Multipass
 Exec=multipass.gui
-Icon=${SNAP}/meta/gui/multipass-gui.svg
+Icon=multipass.gui.svg
```
ricab:
```
 $ desktop-file-validate data/multipass.gui.desktop
data/multipass.gui.desktop: error: (will be fatal in the future): value "multipass.gui.svg" for key "Icon" in group "Desktop Entry" is an icon name with an extension, but there should be no extension as described in the Icon Theme Specification if the value is not an absolute path
```

It will be a path in the snap, but not outside, and since this PR is to make it snap independent... Perhaps the extension should be added in the `sed` line, together with the path.

goffinf:
Making multipass snap agnostic in general would get my vote.

I would like to run multipass under Windows Subsystem for Linux (WSL)
but currently that’s not possible because WSL doesn’t support systemd
or snap.

The related issue this causes is that scripts used to provision Linux
nodes launched by multipass are almost entirely written to execute
within a Linux shell where often those commands don’t play nice with
git bash, Powershell or the native Windows shell.

As an example I like to use multipass to create a local kubernetes
cluster using Rancher k3s, but to do so I have to hack the scripts
provided and for some steps go manual. This undermines the automation
experience that is really needed to support a portable developer
desktop.

My 4 cents

Fraser.

townsend2010:
@goffinf,

I'm not sure of your use case of running Multipass under WSL.  First, Multipass runs natively on Windows, using either Hyper-V or VirtualBox as its hypervisor backend.  Second, it definitely won't run under WSL1 as it requires a KVM enabled kernel and no kernel is available on WSL1.  If you're using the WSL2 preview, then the question is if the Microsoft provided kernel has KVM support and if the "lightweight" hypervisor it runs supports nested virtualization.

Saviq:
@goffinf hey, can you please file an issue (issues?) with more info on what breaks in your workflow for the k8s cluster?

Saviq:
> Thanks, this makes it tidier and I like the overall approach of having the snap as an addition only :slightly_smiling_face: However, I just tried to snap and the multipass menu entry has no icon now (it does launch the gui).

Hmf, I've now fixed the original `.desktop` file, have snapped and the icon is there fine (and the `/var/lib/snapd/desktop/applications/multipass_gui.desktop` file looks sane).

> Also, could you please explain this PR just a little bit better? My understanding is that there are two parts to it: one to make the file snap-agnostic (making things tidier along the way) and another to make it known in the snap. Is this correct? If so, what are the expected effects of the second part? Only the icon in start and the launch thing for which you opened a bug in launchpad?

There are no visible effects of this due to the above LP bug. But you can check the result with:

```
$ curl  --unix-socket /run/snapd.socket http://localhost/v2/snaps/multipass | jq '.result.apps[] | select(.name=="gui")'
{
  "snap": "multipass",
  "name": "gui",
  "desktop-file": "/var/lib/snapd/desktop/applications/multipass_gui.desktop"
}
```

The `desktop-file` key is missing in today's snap.

Saviq:
> The `desktop-file` key is missing in today's snap.

FWIW this alone could be fixed by a simple `git mv snap/gui/{multipass-,}gui.desktop` (the `.desktop` file name needs to exactly match the app name), but I thought it would be good to clean the situation up.

Saviq:
> I confirm the multipass menu entry got its icon back. On the `gui.desktop` workaround, do you still intend to do it? What about using just a link instead?

Oh no, that's an alternative to what's done in this PR. It's all done here.

ricab:
OK, I didn't quite get it then...

PR #1099, opened 2019-09-27
Provide available settings (Fixes #1023)

### LINKED ISSUE #1023: Document `get/set` options in `multipass help`
**example**

> `c:\>multipass config`
> Error: Unknown Command 'config' (try "multipass help")

>`c:\> multipass set local.network="192.168.0.1/24" `
>Unrecognized settings key: 'local.network'

ricab:
Addresses #1023 

This lists available configuration keys in `get` and `set`. I ideally they should offer guidance on what each key means and what values are allowed, but that will require more thought.

### DIFF
--- a/include/multipass/settings.h
+++ b/include/multipass/settings.h
@@ -25,6 +25,7 @@
 #include <QVariant>
 
 #include <map>
+#include <set>
 
 namespace multipass
 {
@@ -33,6 +34,7 @@ class Settings : public Singleton<Settings>
 public:
     Settings(const Singleton<Settings>::PrivatePass&);
 
+    std::set<QString> keys() const;
     virtual QString get(const QString& key) const;            // throws on unknown key
     virtual void set(const QString& key, const QString& val); // throws on unknown key or bad settings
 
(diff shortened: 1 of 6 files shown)


ricab:
Currently each key is registered with `Settings` with only a default value, upon construction. Then there are hardcoded checks when setting each key.

I was thinking of evolving things to get a description and a domain (e.g. "qemu" or "libvirt") registered with the keys. We'd probably also need an optional checker (for when we need to check more than just the domain).

This info could be gathered nicely into the help page, the same way the keys are (displayed alphabetically, which means automatically grouping config branches together). The help page would then be self-contained, avoiding the problem of keeping the docs and code in sync. It would grow with time, but I thought that's ok. A lot of help and man pages are huge, and this would not be that big.

WDYT?

ricab:
uhhh, why did bors keep this open?

Saviq:
@ricab did it? Looks merged?

ricab:
@Saviq bors merged the branch but kept the PR open and did not delete the branch. When I manually closed the PR, bors deleted the branch.

PR #1107, opened 2019-10-02
image_vault: Delete old images from the cache

### LINKED ISSUE #1014: The background image updater does not remove the older image
When the background image updater runs and finds an image that needs updating, the older image is left lying around taking up disk space.

townsend2010:
This will delete old images in the cache when updating an image and will also remove
any "leaked" images that are no longer in the database.

Fixes #1014


townsend2010:
@ricab, I'll see if adding some testing is prudent for this.

ricab:
@townsend2010 OK, thanks a lot!

townsend2010:
I think it will be possible to add some tests, but I won't have it done before my holiday.  I'll put this on hold until I'm back.

townsend2010:
Hmm, try again...

bors r=ricab

PR #1110, opened 2019-10-03
image_vault: Allow background image updater downloads to be aborted

### LINKED ISSUE #1109: Background image updater downloads block daemon exit
If there is/are in-progress background image updater download(s), the daemon will block exiting until the downloads are complete.

townsend2010:
Fixes #1109

### DIFF
--- a/include/multipass/url_downloader.h
+++ b/include/multipass/url_downloader.h
@@ -13,9 +13,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Chris Townsend <christopher.townsend@canonical.com>
- *              Alberto Aguirre <alberto.aguirre@canonical.com>
- *
  */
 
 #ifndef MULTIPASS_URL_DOWNLOADER_H
@@ -43,6 +40,7 @@ public:
                              const ProgressMonitor& monitor);
     virtual QByteArray download(const QUrl& url);
     virtual QDateTime last_modified(const QUrl& url);
+    virtual void abort_all_downloads();
 
 private:
     URLDownloader(const URLDownloader&) = delete;
@@ -50,6 +48,7 @@ private:
 
     const Path cache_dir_path;
     std::chrono::milliseconds timeout;
+    bool abort_download{false};
 };
 }
 #endif // MULTIPASS_URL_DOWNLOADER_H
--- a/include/multipass/url_downloader.h
+++ b/include/multipass/url_downloader.h
@@ -43,13 +43,15 @@ public:
     virtual QDateTime last_modified(const QUrl& url);
     virtual void abort_all_downloads();
 
+protected:
+    std::atomic_bool abort_download{false};
+
 private:
     URLDownloader(const URLDownloader&) = delete;
     URLDownloader& operator=(const URLDownloader&) = delete;
 
     const Path cache_dir_path;
     std::chrono::milliseconds timeout;
-    std::atomic_bool abort_download{false};
 };
 }
 #endif // MULTIPASS_URL_DOWNLOADER_H
--- a/include/multipass/url_downloader.h
+++ b/include/multipass/url_downloader.h
@@ -24,6 +24,7 @@
 #include <QByteArray>
 #include <QDateTime>
 
+#include <atomic>
 #include <chrono>
 
 class QUrl;
@@ -48,7 +49,7 @@ private:
 
     const Path cache_dir_path;
     std::chrono::milliseconds timeout;
-    bool abort_download{false};
+    std::atomic_bool abort_download{false};
 };
 }
 #endif // MULTIPASS_URL_DOWNLOADER_H
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -20,7 +20,6 @@
 
 #include <multipass/exceptions/aborted_download_exception.h>
 #include <multipass/exceptions/create_image_exception.h>
-#include <multipass/exceptions/download_exception.h>
 #include <multipass/exceptions/unsupported_image_exception.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -18,7 +18,9 @@
 #include "default_vm_image_vault.h"
 #include "json_writer.h"
 
+#include <multipass/exceptions/aborted_download_exception.h>
 #include <multipass/exceptions/create_image_exception.h>
+#include <multipass/exceptions/download_exception.h>
 #include <multipass/exceptions/unsupported_image_exception.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
@@ -279,6 +281,11 @@ mp::DefaultVMImageVault::DefaultVMImageVault(std::vector<VMImageHost*> image_hos
     }
 }
 
+mp::DefaultVMImageVault::~DefaultVMImageVault()
+{
+    url_downloader->abort_all_downloads();
+}
+
 mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, const Query& query,
                                                  const PrepareAction& prepare, const ProgressMonitor& monitor)
 {
@@ -469,6 +476,10 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, co
             in_progress_image_fetches.erase(id);
             return finalize_image_records(query, prepared_image, id);
         }
+        catch (const AbortedDownloadException&)
+        {
+            throw;
+        }
         catch (const std::exception& e)
         {
             std::lock_guard<decltype(fetch_mutex)> lock{fetch_mutex};
@@ -577,7 +588,7 @@ void mp::DefaultVMImageVault::update_images(const FetchType& fetch_type, const P
             prepared_image_records.erase(key);
             persist_image_records();
         }
-        catch (const std::exception& e)
+        catch (const CreateImageException& e)
         {
             mpl::log(mpl::Level::warning, category,
                      fmt::format("Cannot update source image {}: {}", record.query.release, e.what()));
@@ -636,6 +647,10 @@ mp::VMImage mp::DefaultVMImageVault::download_and_prepare_source_image(
         remove_source_images(source_image, prepared_image);
 
         return prepared_image;
+    }
+    catch (const AbortedDownloadException&)
+    {
+        throw;
     }
     catch (const std::exception& e)
     {
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -67,7 +67,7 @@ auto get_network_cache_data(QAbstractNetworkCache* network_cache, const QUrl& ur
 
 template <typename ProgressAction, typename DownloadAction, typename ErrorAction, typename Time>
 QByteArray download(QNetworkAccessManager* manager, const Time& timeout, QUrl const& url, ProgressAction&& on_progress,
-                    DownloadAction&& on_download, ErrorAction&& on_error, const bool& abort_download)
+                    DownloadAction&& on_download, ErrorAction&& on_error, const std::atomic_bool& abort_download)
 {
     QEventLoop event_loop;
     QTimer download_timeout;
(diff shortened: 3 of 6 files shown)


include/multipass/url_downloader.h
```diff
@@ -43,13 +40,15 @@ class URLDownloader
...
     virtual QDateTime last_modified(const QUrl& url);
+    virtual void abort_all_downloads();
 
 private:
     URLDownloader(const URLDownloader&) = delete;
     URLDownloader& operator=(const URLDownloader&) = delete;
 
     const Path cache_dir_path;
     std::chrono::milliseconds timeout;
+    bool abort_download{false};
```
ricab: `abort_download` is read and written to concurrently, so strictly speaking it requires synchronization. 

I have used the same technique as you a few times before because, in practice, since it is a boolean, the worse that I could see happening is `on_download` not seeing it as true until a later iteration. Even it the underlying memory got messed up &ndash; I thought &ndash; it would still resolve to `true`. However, this assumes a memory model, and forgets the messing up could actually mean setting it to 0 (see below). Formally it is undefined behavior, so, better keep it sane.

A second issue is that, if the compiler caches it in a register, the second thread might not see it changing ever (until there is a sync with memory, which may never happen IIUC). This can be thought of as a particular case of the memory being messed up &ndash; to the previous value, in this case. That could be prevented by declaring it `volatile`, but that does nothing for our undefined behavior yet...

So, thr right app approach I think is to declare it `std::atomic_bool`. I believe at that point it can't be cached, so no point declaring it volatile.

Note for completeness: `std::atomic` (which `atomic_bool` specializes) uses a lock underneath, so there is a little performance penalty. It does not matter in our case but, if it did, there would be another alternative: `std::atomic_flag`. Unlike the former, which is an `std::atomic` specialization, the latter is guaranteed to be lock-free. Its interface is awkward though.

townsend2010: @ricab, hey, thanks for the review.  I had thought about using an atomic, but thinking that we would only ever have one writer (and potentially many readers), that it would be safe.  But after reading what you wrote and doing a bit of research, yeah, atomic is the way to go.  I'll change that.

townsend2010:
I found a race condition that can cause a crash when the daemon is exiting.  Will need to investigate further.

include/multipass/url_downloader.h
townsend2010: Done

townsend2010:
Ok, this is updated now and *should* avoid the aforementioned race.

tests/test_image_vault.cpp
```diff
@@ -165,6 +167,26 @@ struct HttpURLDownloader : public mp::URLDownloader
...
+struct RunningURLDownloader : public mp::URLDownloader
+{
+    RunningURLDownloader() : mp::URLDownloader{std::chrono::seconds(10)}
+    {
+    }
+    void download_to(const QUrl& url, const QString& file_name, int64_t size, const int download_type,
+                     const mp::ProgressMonitor&) override
+    {
+        while (!abort_download)
+            QThread::yieldCurrentThread();
```
ricab: Nice :)

townsend2010:
@ricab,

For the issues you see, I too, ran into these when trying to test these via `launch`.  But these issues are due to some issue in grpc that needs to be investigated in that the grpc server won't shutdown when the client has an active connection.

The way to test his particular PR is to modify the json to a different hash and to modify the 6 hour check to something like 30 or 60 seconds.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -18,7 +18,9 @@
 #include "default_vm_image_vault.h"
 #include "json_writer.h"
 
+#include <multipass/exceptions/aborted_download_exception.h>
 #include <multipass/exceptions/create_image_exception.h>
+#include <multipass/exceptions/download_exception.h>
```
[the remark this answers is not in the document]
townsend2010: Oops, left over cruft from when I was trying a different way to fix the race.  Will remove.

src/network/url_downloader.cpp
```diff
@@ -97,7 +98,11 @@ QByteArray download(QNetworkAccessManager* manager, const Time& timeout, QUrl co
         on_error();
 
         const auto msg = download_timeout.isActive() ? reply->errorString().toStdString() : "Network timeout";
-        throw mp::DownloadException{url.toString().toStdString(), msg};
+
+        if (reply->error() == QNetworkReply::OperationCanceledError)
+            throw mp::AbortedDownloadException{msg};
+        else
+            throw mp::DownloadException{url.toString().toStdString(), msg};
```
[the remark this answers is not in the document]
townsend2010: Could you please explain this and the previous suggested change?  I fail to see what it gains us.

ricab:
@townsend2010  As discussed in IRC, as of today, this PR:
- is called *image_vault: Allow in-progress downloads to be aborted*
- claims to fix #1109, which is called *In-progress downloads block daemon exit* and whose description is *If there is/are in-progress download(s), the daemon will block exiting until the downloads are complete.*

This is just not true. Some downloads (most of them really) still block the daemon. We need to keep track of that. #1109 does that very succinctly, but would be marked fixed with this PR.

If you think this is a useful fix on its own, fine, just update things on github. That would have saved me some time. ATM it just doesn't do what it says.

townsend2010:
I will look into a bit further.  Perhaps the order of constructors in daemon_config is causing some issue when the destructors are being called in the reverse order when the daemon is exiting.

townsend2010:
Ok, constructor/destructor order does not matter.  We are definitely seeing #1126 here, which is absolutely not related to image downloads.  I even looked at all of the threads running and the thread responsible for doing the download is not even running!

As far as the semantics of what this is fixing, it _is_ true that this PR is fixing a case where the daemon would block when an in-progress download is occurring.  As mentioned here and in IRC, the case were the background image updater is actively doing a download will no block the daemon from exiting with this PR.

Would it be better if I changed the wording of #1109, this PR, and the commits to only reference the background image updater instead of generically referring to "in-progress downloads"?

townsend2010:
> Some downloads (most of them really) still block the daemon.

Simply not true after debugging what is going on and I stated as such in comments in this PR and on IRC.  The visible behavior appears that way, but it's not the download at all causing the issue.

> That would have saved me some time.

When you initially reviewed this on October 4, I assumed you had tested it already. You would have run into the "hang" issue on `launch` then as well, if you had tested it the same then as you did today.  Until yesterday, I was unaware of the `launch` hang issue since I had only tested using the background image updater methodology.  After my holiday, I ran into the `launch` hang as well and debugged it to determine it is a grpc server shutdown issue (#1126).  As such, I felt this still addressed "in-progress downloads" since I felt and still feel the grpc server shutdown issue is a completely unrelated issue only in that the visible behavior of the daemon and the client is the same.

I apologize for not being more specific in the bug and this PR once I discovered the grpc server shutdown issue.  I did not intend to waste your time.

townsend2010:
Ok, I replaced "in-progress" with "background image updater" where appropriate.  Hopefully that is more clear.

ricab:
Hi @townsend2010, I understand what you are saying better now. However, I am apparently seeing something slightly different from you, because the download thread is still there. When I ctrl+c or SIGKILL the daemon during a launch download, it keeps going (and completes). That may be because it is wrapped in a grpc thread and the vault never gets destroyed, but it is effectively still a *download that I could not abort*.

> I assumed you had tested it already.

No, I don't think I had, I try to leave that to the end if I can. I experienced being unable to cancel background downloads, but that was in the other PR. I honnestly forgot the details after the interregnum. I find the advantage of recording issues and PRs is being able to recover the info. And when I tested *allow in-progress downloads to be aborted*, I could not find it to be true.

I agree the new cause should be considered a separate issue and I think your updates in GH make that precise and clear now. Thank you for that and I am sorry for any miscommunisation on my part.

src/network/url_downloader.cpp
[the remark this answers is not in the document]
ricab: Yes, I meant them as the same change, but I could not make the new multiline thing work with that. My goal was two-fold:
- to make sure there was no corner case where we were throwing `AbortedDownloadException` with "Network timeout" (e.g. ctrl+c immediately followed by a network timeout)
- to make it clear to the reader that "Network timeout" is only applicable to `DownloadException`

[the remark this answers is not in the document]
ricab: I've added the suggestions to a batch, to make it clear they're intended to go together.

townsend2010:
@ricab,

Regarding not seeing the download thread, I misspoke about that and left out an important detail.  I did an experiment where I created `~Deamon()` and in that I explicitly called `config->url_downloader->abort_all_downloads()` to ensure that was the first thing to happen during tear down.  In this case, the download thread did indeed stop yet the grcp server shutdown hang still occurred.

I still do think this PR will cancel in-progress downloads during `launch` once we solve the grpc issue, but since there is no functional way of proving that, we'll just refer to the background image updater.

src/network/url_downloader.cpp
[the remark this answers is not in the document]
townsend2010: Ok, I see.  Although I really don't think it matters what the message is here (`Network timeout` vs. `Operation canceled`) when the daemon exits, I'll incorporate this for completeness's sake.

ricab:
> I still do think this PR will cancel in-progress downloads during `launch` once we solve the grpc issue

I understand that, but the way I see it that is not enough to say this PR does it. To me that would be like "Bug: `increment(1) == 1`; investigation: increment decrements before incrementing; PR: 'fixes increment'; test: `increment(2) == 2`; investigation: separate problem means `increment(a)` always treated as `increment(1)`; What I see you saying: "Calling this PR 'fix increment' is still valid because it will fix increment once we fix that other part". 

This is perhaps not too relevant anymore, but just so you see my point of view.

src/network/url_downloader.cpp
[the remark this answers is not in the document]
ricab: Yeah sure, this was not the most important thing, just a detail.

townsend2010:
@ricab,

I already saw your point of view:)  That is why I changed all of the references of the generic "in-progress" wording to the specific "background image updater" wording.  It is also why I said in the same sentence you partially quoted that I said there is no functional way of testing it for `launch` due to the grpc issue, but with me being the developer of this fix, I am confident it will also fix the `launch` version of downloading as well.  It was just meant as a statement of confidence, not a statement that it should still apply to the `launch` scenario.

Let's please be done with this particular discussion and move on to either approving this PR or addressing any additional issues with this PR.  Thanks!

ricab:
I am glad you see it, I wasn't sure after the "I still do think" sentence. And thank you again for the GH updates!

Yes let's move on indeed. I really want to get this PR done with, but I need to test it. At least that's my understanding of what I am tasked with.

I have just confirmed the background cancel. I now only need to check whether the second issue I mentioned is a regression.

ricab:
Well, the second issue I mentioned really appears to be a regression. After this PR, a user who ctrl+c's a launch can't launch anymore... Unless I am missing something?

I am really sorry, I don't want to upset you and I honestly wanted to send this in, but what am I supposed to do? Not test? Ignore? You know, I don't even mind if you want to merge yourself... Go right ahead, I just can't take responsibility for that... Please let me know if there is something I can do to make your life easier here.

townsend2010:
@ricab,

I'll look into it.  Thanks!

townsend2010:
Confirmed there is a regression.  I misread your steps yesterday for issue 2 and thought you were Ctrl-C'ing `multipassd`, not the client.  Sorry about that.

townsend2010:
Ok, I believe the issue #2 regression is fixed.  Thanks again!

src/network/url_downloader.cpp
```diff
@@ -66,7 +67,7 @@ auto get_network_cache_data(QAbstractNetworkCache* network_cache, const QUrl& ur
 
 template <typename ProgressAction, typename DownloadAction, typename ErrorAction, typename Time>
 QByteArray download(QNetworkAccessManager* manager, const Time& timeout, QUrl const& url, ProgressAction&& on_progress,
-                    DownloadAction&& on_download, ErrorAction&& on_error)
+                    DownloadAction&& on_download, ErrorAction&& on_error, const bool& abort_download)
```
ricab: This will still be read concurrently, so should still be atomic, no?

townsend2010: Oh, right, duh...

townsend2010: Thanks, updated.

Issue #1003, opened 2019-08-28
Tests fail when shuffled


ricab:
```console
$ ./bin/multipass_tests --gtest_shuffle
[...]
[==========] 820 tests from 51 test suites ran. (4727 ms total)
[  PASSED  ] 816 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] QemuBackend.returns_version_string_when_failed_parsing
[  FAILED  ] QemuBackend.verify_qemu_arguments_when_resuming_suspend_image
[  FAILED  ] QemuBackend.returns_version_string_when_errored
[  FAILED  ] QemuBackend.returns_version_string_when_exec_failed
```

Each failing test says:

```
unknown file: Failure
C++ exception with description "ProcessFactory::instance() called before MockProcessFactory::Inject()" thrown in the test body.
```

ricab:
We have experienced manifestations of this when adding new tests or changing their build structure. A useful case for investigation, with a fixed ordering that fails, is reported [here](https://github.com/CanonicalLtd/multipass/pull/1120#issuecomment-542657291).

PR #1120, opened 2019-10-08
Check min img size

### LINKED ISSUE #701: Multipass should provide better error message when unable to resize
```sh
$ multipass launch ubuntu --mem 1500M --disk 2g
launch failed: Cannot resize instance image
```

This happens because

```sh
qemu-img info ubuntu-18.04-server-cloudimg-amd64.img | grep "virtual size"
virtual size: 2.2G (2361393152 bytes)
```

The error message should explain this. Otherwise, the user will have no idea why this happened.

Alternatively &ndash; or additionally &ndash; this can be explained in the help.

ricab:
Check that the disk size requested on launch is not below the virtual size of the corresponding image. Fixes #701.


src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -86,6 +87,34 @@ auto virtual_switch_subnet(const QString& bridge_name)
...
+        std::make_unique<mp::QemuImgProcessSpec>(QStringList{"info", image_path}));
+    auto process_state = qemuimg_process->execute();
+
+    if (!process_state.completed_successfully())
+        throw std::runtime_error(fmt::format("Cannot get image info: qemu-img failed ({}) with output:\n{}",
+                                             process_state.failure_message(),
+                                             qemuimg_process->read_all_standard_error()));
+
+    const auto img_info = QString{qemuimg_process->read_all_standard_output()};
+    const auto capture_name = QStringLiteral("size");
```
Saviq: This seems a bit excessive… I don't think repeating "size" would be that big of a deal here, and having to substitute this in the pattern below… Your call.

src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -86,6 +87,34 @@ auto virtual_switch_subnet(const QString& bridge_name)
...
+    const auto pattern = QStringLiteral("^virtual size: .+ \\((?<%1>\\d+) bytes\\)$").arg(capture_name);
+    const auto re = QRegularExpression{pattern, QRegularExpression::MultilineOption};
+
+    if (const auto match = re.match(img_info); match.hasMatch())
+    {
+        const auto min_size = match.captured(capture_name).toStdString();
+        if (requested_size < mp::MemorySize{min_size})
+            throw std::runtime_error(fmt::format("Requested disk ({} bytes) below minimum for this image ({} bytes)",
+                                                 requested_size.in_bytes(), min_size));
```
Saviq: I think we should print this human-readable, we can probably assume GBs here, although now I see `MemorySize` floors, would have to round up for this use case…

I'd be OK with this with a TODO to improve the message later.

src/platform/backends/shared/linux/backend_utils.cpp
ricab: OK, I will do that.

src/platform/backends/shared/linux/backend_utils.cpp
ricab: I considered that, but then users might get `Requested disk (2.2G) below minimum for this image (2.2G)`, because of approximations. I used bytes to be precise and avoid that.

ricab: So, this would happen when a user wanted to save space:

```
$ multipass launch --disk=1G
Requested disk (1G) below minimum for this image (2.2G)
$ multipass launch --disk=2.2G  # user trying next-best thing
Requested disk (2.2G) below minimum for this image (2.2G)
```

ricab:
> Could we unit-test this please?

Sure, I am working on it.

src/platform/backends/shared/linux/backend_utils.cpp
Saviq: Well of course, which is why I mentioned we'd need to `ceil()` the minimum, in which case the minimum would be `2.3G`.

ricab: I see, I did not get what you meant before. OK then, fine by me. It would be imprecise, but I agree it looks friendlier.

ricab:
Test are in. I still have to address the other review comments.

ricab:
CI-detected conflict addressed in canonical/multipass-private#187.

Saviq:
@ricab if you rebase / merge this, the CI failure should go away.

ricab:
Bahh new test failure...

ricab:
The failure is just another manifestation of #1003. I tried a shot at changing the order of our sources in cmake and it made it go away! At least in my machine, let's see what CI says.

I kept that change as a separate commit on purpose. In summary, the failure in c5eacaf (merge commit) goes away in f502847. Something that will byte us back again for sure, but not really related to this PR.

src/platform/backends/shared/linux/backend_utils.cpp
ricab: Bahh I totally forgot about this and just noticed this PR never got merged,

src/platform/backends/shared/linux/backend_utils.cpp
ricab: `MemorySize` does not support decimals ATM, so I will add the TODO for now.

ricab:
The conflict in CI is handled in canonical/multipass-private#190

ricab:
Thanks! Yeah, the tests took most of the time. 

I considered parameterized tests but ended up deciding against. I forget the exact reasoning at the time, but I think the main reasons were:
- `test_image_resizing` is a template: I remember looking briefly and not finding any obvious way to mix parameterized tests with templates
- it would not gain me much: despite the many parameters, there weren't many combinations to test and the automated facilities to produce parameter combinations to test would not have been useful
- test names would suffer: output would consist of a repeated name plus a load of parameters, instead of a specific, more descriptive, name for each test

Saviq:
>     * it would not gain me much: despite the many parameters, there weren't many combinations to test and the automated facilities to produce parameter combinations to test would not have been useful

There would be less `const auto`s though ;). But then, less readable because need to remember argument order, I suppose.

>     * test names would suffer: output would consist of a repeated name plus a load of parameters, instead of a specific, more descriptive, name for each test

FWIW this one can be solved by providing [a naming function](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#specifying-names-for-value-parameterized-test-parameters) when instantiating the suite.

Thanks.

polarapfel:
I tried edge/pr1120 and I am still running into the issue with qemu-img being unable to resize the image and timing out.

townsend2010:
Hi @polarapfel,

It sounds to me the issue you are seeing is unrelated to this PR.  Could you please submit a new issue and put in your reproduction steps?

polarapfel:
Sure thing.

polarapfel:
Hey @townsend2010, it's #1408.

Issue #1132, opened 2019-10-21
sshfs_server not cleanly shut down when stopping an instance


townsend2010:
I noticed the following behavior when using the apparmored `sshfs_server` when shutting down an instance:
```
QProcess: Destroyed while process ("/snap/multipass/1244/bin/sshfs_server") is still running.
```

It doesn't appear `multipassd` is waiting for `sshfs_server` to fully exit before continuing the stopping of the instance and deleting the `QProcess` object.

Since this occurs during shutdown, it's probably not too extreme of a bug, but I suppose there could be some data that still needs to be written to the mount before the mount itself stops.

PR #1149, opened 2019-10-24
platform: Do not detect parent dying and send SIGHUP to ssfs_server

### LINKED ISSUE #1143: Mounts stop working after 30 when starting an instance with mounts defined
Mounts will stop working after 30 seconds when starting an instance that has mounts defined.  This can be one or more mounts.

Steps to reproduce:
* Launch an instance
* Define a mount
* Stop the instance
  * You will hit #1131 until #1136 is merged.  So start and stop again to overcome that.
* Start the instance

Observe after 30 seconds, the `sshfs_server` process will unexpectedly go away, leaving the mount dead.

townsend2010:
When starting mounts asynchronously, ie, on start, when the thread that starts the mount
exits, this code would send a SIGHUP to sshfs_process, causing it to exit prematurely.

This code is mostly unnecessary since it's only for Linux and systemd takes care of killing
child processes if the main process dies unexpectedly when running the snap.

Fixes #1143

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2019 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -12,8 +12,6 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
  *
  */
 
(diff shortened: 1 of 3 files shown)


Saviq:
Hmm, something's fishy:
```
paź 25 10:19:01 michal-laptop multipassd[16130]: QProcess: Destroyed while process ("/snap/multipass/1291/bin/sshfs_server") is still running.
paź 25 10:19:01 michal-laptop multipassd[16130]: Mount 'dev' in instance "primary" has stopped
…
paź 25 10:22:47 michal-laptop multipassd[16130]: No mounts to stop for instance "primary"
```

But the mount's still up. So we seem to be losing the handle on `sshfs_server`, as if the process move from the QtConcurrent thread to the daemon thread didn't work?

Aaand now I can't reproduce…

I can confirm systemd is taking care of all the processes when the main one (`multipassd`) goes away, though.

townsend2010:
@Saviq,

For what it's worth, I think you saw a manifestation of #1132- at least the symptom looks the same as what I observed.

Saviq:
Ack. I haven't been able to prove otherwise.

bors r+

Saviq:
Hrmf, just to get `staging` up:

bors r+

PR #1148, opened 2019-10-24
just testing, nothing to see here ;)

Saviq:

PR #1152, opened 2019-10-25
[travis] publish `trying` builds on its PR

Saviq:


tools/report_build.py
```diff
@@ -245,31 +286,49 @@ class RepoCall(GitHubV3Call):
...
 
 
-class UpdateCommitComment(RepoCall):
+class UpdateCommitComment(RepoCall, CommentURLMixin):
     def run(self, variables={}):
         super().run(variables)
-        return (self._repo.get_comment(
+        self.data = self._repo.get_comment(
+            self._vars(variables)["databaseId"])
+        (self._repo.get_comment(
```
ricab: This seems forgotten, `edit` should just be called on `self.data`.

tools/report_build.py
```diff
@@ -347,18 +410,19 @@ def main():
             # found a recent commit we can update
             for line in event["node"]["body"].splitlines():
                 # include all builds with a different build_type
-                if not line.startswith(args.build_type):
+                if not line.startswith(f"{args.build_type} "):
```
ricab: Why the change? What difference does the space make?

Saviq: In case of "Type" and "Type2", "Type" would take "Type2" over when updating the comment.

This is not entirely future-proof, but good enough for the formatting we have now.

tools/report_build.py
Saviq: Fixed.

tools/report_build.py
ricab: OK, fair enough

Saviq:
> OK, lgtm. Delegating as I'm not sure if you still want to do something about the label discrepancy here.
Just want to wait for [the `bors try`](https://travis-ci.org/CanonicalLtd/multipass/builds/602899644) to complete on #1148, it should report here (because that's the last PR for the commit).

Thanks!

.travis.yml
```diff
@@ -64,12 +64,19 @@ before_install:
...
-    # other branches only publish on the commits
+    # other branches publish on the commits by default
     # skip staging as the same commit will be published from master if all is good
     elif [[ "${TRAVIS_EVENT_TYPE}" == "push" \
             && "${TRAVIS_BRANCH}" != "staging" ]]; then
       MULTIPASS_PUBLISH="true"
 
+      # except for `trying`, which publishes on the parent's PR
+      if [[ "${TRAVIS_BRANCH}" == "trying" \
+            && "${TRAVIS_COMMIT_MESSAGE}" =~ ^Try\ #([0-9]+): ]]; then
```
Saviq: @ricab one thing I realized I forgot:

```suggestion
        MULTIPASS_SNAP_CHANNEL="edge/pr${BASH_REMATCH[1]}"
```

This makes the channel different from the build label, but consistent in terms of snap channels, which I think is more important. I could also just keep the build label `pr*`, so it's consistent regardless if it's a PR build or a `trying` one… WDYT?

ricab: Hmm so without this, a trying snap wouldn't be published in snapcraft, right? Do we really need it there? With this, it would overwrite the snap in the corresponding pr channel, correct?

If so, I think I'd prefer a separate `edge/tryXYZ` channel, to keep the distinction transparent and consistent with the build label. If there is any difference between that and the previous PR build, better be able to investigate it. Otherwise, what's the real point of publishing? 

I recognize the proliferation of snap channels, but I think that is fair reflex of the many sorts of build uploads.

Saviq: @ricab we only really need `bors try` on external contributions, on which the PR builds won't get published in the store because of the lack of credentials.

Saviq: Oh and the only difference between a PR build and a `bors try` build would be the merge commit, so in the package - the version string. And even if a channel is "overwritten", collaborators have access to previous revisions via `snap install --revision`.

The argument for keeping it under the `edge/pr*` branch is to avoid reviewers having to check / remember. The most recent build for a PR, whether from a `pull_request` travis build, or a `bors try` one, would be published in the `edge/pr<number>` channel.

Maybe the `try` build label is an unnecessary distinction?

Saviq: > I am not sure I follow. We use `bors try` ourselves sometimes. But if the only use would fail uploading to the store, why attempt to upload in the first place?

It's exactly `bors try` that won't fail to upload, as opposed to a `pull_request` build from outside of the CanonicalLtd/multipass repository (for security reasons - we can't run untrusted code with credentials). `bors try` means that someone has looked at the code and verified there's nothing malicious in it and it's safe to run with credentials, we can very well publish for testing. When the PR is "internal", `bors try` does virtually the same thing a normal PR run does.

> OK, but then what is the use of publishing it?

Testing, no?

ricab: Ah ok, much clearer now. 

Well, if there is no reason to distinguish snap channels, indeed I'd expect no real reason to distinguish snap package names. So yeah, IMO it should be coherent: either `try*` or `pr*` in both label and channel.

Saviq:
snap try

Saviq:
bors try

Saviq:
Once more, with feeling.

bors try

PR #1154, opened 2019-10-28
[mount] fix mounting to a default target (Fixes #1145)

### LINKED ISSUE #1145: `mount` requires a target path
When the help states:
```
  target                           Target mount points, in <name>[:<path>]
                                   format, where <name> is an instance name, and
                                   optional <path> is the mount point. If
                                   omitted, the mount point will be the same as
                                   the source's absolute path
```

Saviq:


src/client/cli/cmd/mount.cpp
```diff
@@ -153,12 +153,12 @@ mp::ParseCode cmd::Mount::parse_args(mp::ArgParser* parser)
...
     for (auto i = 1; i < parser->positionalArguments().count(); ++i)
     {
-        auto parsed_target = QString(parser->positionalArguments().at(i)).split(":");
+        auto parsed_target = QString(parser->positionalArguments().at(i)).split(":", QString::SkipEmptyParts);
 
         auto entry = request.add_target_paths();
         entry->set_instance_name(parsed_target.at(0).toStdString());
 
-        if (parsed_target.count() == 1)
+        if (parsed_target.count())
```
townsend2010: This doesn't work.
`$ multipass mount ~/Downloads/ foo:Downloads` will still mount it in the absolute path of the source, but it should be in the `ubuntu` user's directory.

I think it still should be `== 1` here since that indicates *only* the instance name was found.

Saviq: Of course, bad refactoring on my part. Fixed.

PR #1155, opened 2019-10-29
[travis] fix release handling

Saviq:
We incorrectly published on PRs of `release/0.9` branches, and the version determination required for `release/*` to be a local ref instead of one in the `origin` remote.

---

Testing:
1. `MULTIPASS_UNLOCK` is required on a `release/*` branch
   ```console
   $ git checkout -b release/0.0
   $ cmake .
   # ...
   CMake Error at CMakeLists.txt:87 (message):
     You need to set MULTIPASS_UPSTREAM for a release build.

     Use an empty string to make local the authoritative repository.
   ```
1. setting `MULTIPASS_UPSTREAM=origin` makes it ignore the `release/*` branch
   ```console
   $ git checkout -b release/0.0
   $ MULTIPASS_UPSTREAM=origin cmake .
   # ...
   -- Setting version to: 1.1.0-dev.84+g4a33a0a9
   $ git tag -as v0.0.0 -m Foo
   $ MULTIPASS_UPSTREAM=origin cmake .
   # ...
   -- Setting version to: 1.1.0-dev.84+g4a33a0a9
   $ git tag -d v0.0.0
   ```

1. setting `MULTIPASS_UPSTREAM=` makes it respect the `release/*` branch
   ```console
   $ git checkout -b release/0.0
   $ MULTIPASS_UPSTREAM=origin cmake .
   # ...
   -- Setting version to: 1.1.0-rc.84+g4a33a0a9
   $ git tag -as v0.0.0 -m Foo
   $ MULTIPASS_UPSTREAM=origin cmake .
   # ...
   -- Setting version to: 0.0.0
   $ git tag -d v0.0.0
   ```

1. `release/*` pull requests don't get published
   ```console
   ( python3 -c 'import yaml
   t = yaml.safe_load(open(".travis.yml")) 
   print("\n".join(t["env"]["global"]))
   print(t["before_install"][0])'
   for var in PUBLISH BUILD_LABEL SNAP_CHANNEL REPORT_OPTIONS; do
     echo "echo MULTIPASS_${var}=\${MULTIPASS_${var}}"
   done
   ) \
   | env TRAVIS_SECURE_ENV_VARS=true TRAVIS_EVENT_TYPE=pull_request TRAVIS_PULL_REQUEST=123 TRAVIS_PULL_REQUEST_BRANCH=release/0.0 bash

   MULTIPASS_PUBLISH=               # rather than "true"
   MULTIPASS_BUILD_LABEL=pr123
   MULTIPASS_SNAP_CHANNEL=          # rather than "edge/pr123"
   MULTIPASS_REPORT_OPTIONS=
   ```

1. `release/*` pushes still get published on `--branch`
   ```console
   ( python3 -c 'import yaml
   t = yaml.safe_load(open(".travis.yml")) 
   print("\n".join(t["env"]["global"]))
   print(t["before_install"][0])'
   for var in PUBLISH BUILD_LABEL SNAP_CHANNEL REPORT_OPTIONS; do
     echo "echo MULTIPASS_${var}=\${MULTIPASS_${var}}"
   done
   ) \
   | env TRAVIS_SECURE_ENV_VARS=true TRAVIS_EVENT_TYPE=push TRAVIS_BRANCH=release/0.0 bash
   MULTIPASS_PUBLISH=true
   MULTIPASS_BUILD_LABEL=
   MULTIPASS_SNAP_CHANNEL=candidate/0.0
   MULTIPASS_REPORT_OPTIONS=--branch
   ```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ env:
   # regexes for release branch and tag
   - MULTIPASS_RELEASE_BRANCH_PATTERN="^release/([0-9\.]+)$"
   - MULTIPASS_RELEASE_TAG_PATTERN="^v([0-9]+\.[0-9]+)\.[0-9]+$"
+  # upstream to use as reference for release version matching
+  - MULTIPASS_UPSTREAM="origin"
   # build label added to the version string
   - MULTIPASS_BUILD_LABEL=""
   # whether to publish the built packages
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,17 +51,18 @@ before_install:
 
     # when it's a release branch or tag
     if [[ "${TRAVIS_BRANCH}" =~ ${MULTIPASS_RELEASE_BRANCH_PATTERN} \
+          || "${TRAVIS_PULL_REQUEST_BRANCH}" =~ ${MULTIPASS_RELEASE_BRANCH_PATTERN} \
           || "${TRAVIS_BRANCH}" =~ ${MULTIPASS_RELEASE_TAG_PATTERN} ]]; then
 
-      # publish pushes on candidate/*
+      # only publish pushes, on candidate/*
       if [ "${TRAVIS_EVENT_TYPE}" == "push" ]; then
         MULTIPASS_PUBLISH="true"
         MULTIPASS_SNAP_CHANNEL="candidate/${BASH_REMATCH[1]}"
-      fi
 
-      # but report on the PR if it's a release branch
-      if [[ "${TRAVIS_BRANCH}" == release/* ]]; then
-        MULTIPASS_REPORT_OPTIONS+=("--branch" "${TRAVIS_BRANCH}")
+        # but report on the PR if it's a release branch
+        if [[ "${TRAVIS_BRANCH}" == release/* ]]; then
+          MULTIPASS_REPORT_OPTIONS+=("--branch" "${TRAVIS_BRANCH}")
+        fi
       fi
 
     # other branches publish on the commits by default
@@ -78,7 +79,7 @@ before_install:
         MULTIPASS_REPORT_OPTIONS+=("--parent")
       fi
 
-    # non-release publish on edge/pr*
+    # publish all other PRs on edge/pr*
     elif [ "${TRAVIS_EVENT_TYPE}" == "pull_request" ]; then
       MULTIPASS_PUBLISH="true"
       MULTIPASS_SNAP_CHANNEL="edge/${MULTIPASS_BUILD_LABEL}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ add_subdirectory(3rd-party)
 find_package(Qt5 COMPONENTS Core Network Widgets REQUIRED)
 
 function(determine_version OUTPUT_VARIABLE)
-  execute_process(COMMAND git describe --all --match release/*
+  execute_process(COMMAND git describe --all --match origin/release/*
                   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
                   ERROR_QUIET
                   OUTPUT_VARIABLE GIT_BRANCH
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,13 @@ add_subdirectory(3rd-party)
 find_package(Qt5 COMPONENTS Core Network Widgets REQUIRED)
 
 function(determine_version OUTPUT_VARIABLE)
-  execute_process(COMMAND git describe --all --match origin/release/*
+  # default to the "origin" remote when checking for release status
+  set(MULTIPASS_UPSTREAM $ENV{MULTIPASS_UPSTREAM})
+  if(NOT MULTIPASS_UPSTREAM)
+    set(MULTIPASS_UPSTREAM origin)
+  endif()
+
+  execute_process(COMMAND git describe --all --match ${MULTIPASS_UPSTREAM}/release/*
                   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
                   ERROR_QUIET
                   OUTPUT_VARIABLE GIT_BRANCH
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,32 +62,38 @@ add_subdirectory(3rd-party)
 find_package(Qt5 COMPONENTS Core Network Widgets REQUIRED)
 
 function(determine_version OUTPUT_VARIABLE)
-  # default to the "origin" remote when checking for release status
+  # use upstream repo as the authoritative reference when checking for release status
+  # set MULTIPASS_UPSTREAM="*" to use the local repository
   set(MULTIPASS_UPSTREAM $ENV{MULTIPASS_UPSTREAM})
-  if(NOT MULTIPASS_UPSTREAM)
-    set(MULTIPASS_UPSTREAM origin)
+  if(MULTIPASS_UPSTREAM)
+    set(MULTIPASS_UPSTREAM "${MULTIPASS_UPSTREAM}/")
   endif()
 
-  execute_process(COMMAND git describe --all --match ${MULTIPASS_UPSTREAM}/release/*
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -51,17 +51,18 @@ before_install:
 
     # when it's a release branch or tag
     if [[ "${TRAVIS_BRANCH}" =~ ${MULTIPASS_RELEASE_BRANCH_PATTERN} \
+          || "${TRAVIS_PULL_REQUEST_BRANCH}" =~ ${MULTIPASS_RELEASE_BRANCH_PATTERN} \
```
ricab: This is getting hard to follow. Right now the branching is all intertwined:

```
if branch...
  if even_type...
    if branch
elif event... and branch...
  if branch...
elif event...
```

And `TRAVIS_PULL_REQUEST_BRANCH` itself depends on the event type. 

Could this be reorganized/simplified, so that tests for even type would be in one level, tests for branch in another, etc.? To get closer to a table structure, with non-overlapping cases and sub-cases.

CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ add_subdirectory(3rd-party)
 find_package(Qt5 COMPONENTS Core Network Widgets REQUIRED)
 
 function(determine_version OUTPUT_VARIABLE)
-  execute_process(COMMAND git describe --all --match release/*
+  execute_process(COMMAND git describe --all --match origin/release/*
```
ricab: Could you please clarify what cases were malfunctioning without this?

Saviq: This only matched `release/0.9` if you had it as a local branch (i.e. `heads/release/0.9`). The new match matches on the `origin` remote (i.e. `remotes/origin/release/0.9`), so you can check out the commit alone (i.e. `git fetch origin/release/0.9; git checkout FETCH_HEAD`).

That said, I _think_ we're using newer git now when building the snap, so I should be able to use a more precise regex here instead. Will circle back.

.travis.yml
Saviq: Yeah you're right, I'll rework.

CMakeLists.txt
ricab: Yep, that's clear. I was more asking if there was something in CI or some real use-case that requires this. I am concerned that
- this assumes an `origin` remote. For instance, I don't have one with that name here (I call it `upstream`). And I could actually have a remote called `origin` pointing to some other repo (e.g. a personal fork). One way to address this would be `git describe --all --match *release/0.9`, but then there could be multiple/different remotes...
- determining the version depends on something external. It feels like it should be determined according to the repository this is run in. So, no branch named `release/0.9` there &rarr; not version 0.9 in that repo. To determine an official version, this would just need to run in a representative repository, with the required branches and tags.

On a side note, `git fetch origin/release/0.9` does not work for me, git says it "does not appear to be a git repository". But this can be tested under a branch with a different name (e.g. `git checkout -b special_branch upstream/release/0.9`).

.travis.yml
ricab: Cool, thanks!

CMakeLists.txt
Saviq: > I was more asking if there was something in CI or some real use-case that requires this

Yes, CI gets told to build a particular commit (because the referencing branch might've moved since the job was triggered).

>     * this assumes an `origin` remote. For instance, I don't have one with that name here (I call it `upstream`). And I could actually have a remote called `origin` pointing to some other repo (e.g. a personal fork).

Sure, but that's the default remote name, so not a crazy assumption. And where it really matters (CI), we're in control of the remote name.

> One way to address this would be `git describe --all --match *release/0.9`, but then there could be multiple/different remotes...

Yeah, that.

>     * determining the version depends on something external. It feels like it should be determined according to the repository this is run in. So, no branch named `release/0.9` there → not version 0.9 in that repo. To determine an official version, this would just need to run in a representative repository, with the required branches and tags.

That's exactly the problem, the context we get in CI is just the commit, and I'm arguing here that authority of whether that commit is a release candidate (or a release, for that matter) is the upstream repository (by means of a branch name, or a tag), not the local one. Of course you can spoof that by overwriting the tag or the remote, if you so wish.

An alternative is to use a CMake variable and be explicit about what constitutes a release canidate. But then it becomes a function of _how_, rather than _what_ is being built.

> On a side note, `git fetch origin/release/0.9` does not work for me, git says it "does not appear to be a git repository". But this can be tested under a branch with a different name (e.g. `git checkout -b special_branch upstream/release/0.9`).

This should've been `git fetch origin release/0.9`, sorry. And it's different to what you did in that there is no branch if you `git checkout FETCH_HEAD` - you end up in _detached_ state.

Saviq: > Ahh I see. OK, I understand it better now, I agree we need to depend on the remote then. I'd be happier with the star version then. A different remote config would affect the output, but I guess it then makes sense, and then it wouldn't depend on remote names. WDYT?

> Yeah that the command? Or yeah that there could be different remotes?

More importantly, `*` also matches `/`, so a local "foorelease/0.9" branch would also match. Which is why I'm looking at the regular expression approach (that's what I started with, but couldn't use due to older git on Core 16).

Different remotes, partial matches… Wildcards are wild.

Saviq: > Which is why I'm looking at the regular expression approach

Ugh, there isn't a regular expression approach, it's only globs. I misremembered - the difference was multiple `--match` support, not regexes.

I'd go for being explicit about `origin/` in the match, as even this:
```
git describe --all --match */release/*
```
Can be fooled by a local `foo/release/0.9` branch… It's meh that I can't do `--match refs/remotes/*/release/0.9` - that would at least limit it to remotes…

ricab: A `foo/release/0.9` branch sounds less likely than an absent (or fork-pointing) `origin`. I agree the ideal would be a regex, but failing that, IMO it makes more sense to impose restrictions on branches than on remotes.

We could rule out branches with slashes except `release/<num>` and error out in the regex when the glob pattern you mentioned (`*/release/*`) caught anything before "release" that was not a remote.

Orrr... we use `${AUTHORITATIVE_REMOTE_PREFIX}release/*` and pass that var in CI. Perhaps this would be the most sensible... WDYT?

Saviq: Yeah, how about a40e7a65f41b25d12ef3df3d6dfc251a9a1d844e?

ricab: That is a bit better, but now I need to remember to setup an environment or I get the wrong version. Don't you think it would make more sense to use local stuff and override in CI? The detached HEAD would be the special situation needing the var, and that would be in code (rather than local copies needing manual env setup).

ricab: Yeah, I think that is actually better than getting the wrong version unless you have the right remote. I will ask you to reconsider. Proposal in #1160 (using this as base).

That way working copies are more independent by default, as until now. But we can give them an overriding reference.

ricab: BTW, shouldn't the remote be used in other `describe` commands too?

Saviq: Hey @ricab, I've incorporated your proposal and tweaked it a bit:
- when `HEAD` exactly matches `release/*`, `MULTIPASS_UPSTREAM` is required
- you need to set `MULTIPASS_UPSTREAM=""` to match in the local repo

.travis.yml
Saviq: As discussed on IRC, we'll leave it as a separate step, ideally including a test suite that would run as a first step.

ricab:
Hey @Saviq, a few questions on the tests you posted:
  - `gco` is `git checkout`, right?
  - `MULTIPASS_UNLOCK` in step 3 just a typo, correct?
  - also step 3, you meant to have `MULTIPASS_UPSTREAM= cmake .` in the second cmake call, correct?

Saviq:
>     * `gco` is `git checkout`, right?

Sorry, yes, it's `zsh`'s git plugin.

>     * `MULTIPASS_UNLOCK` in step 3 just a typo, correct?

Yup.

>     * also step 3, you meant to have `MULTIPASS_UPSTREAM= cmake .` in the second cmake call, correct?

Yes.

PR #1160, opened 2019-10-30
[cmake, travis] Default to local release matching

ricab:
But support preferring a remote as the authoritative reference. Define
that remote to "origin" in travis builds.


ricab:
> Merging #1160 into fix-release-handling will decrease coverage by 0.2%

:thinking:

Saviq:
Incorporated into #1155

PR #1161, opened 2019-10-30
[launch] do not require a TTY for --cloud-init via stdin

Saviq:
---
The only place this i used is https://github.com/CanonicalLtd/multipass/blob/a91c9c9a214841ba6f97e5c99b953180f8efd8bc/src/client/cli/cmd/launch.cpp#L177-L187

TTY is not a requirement for stdio, this check is wrong.

To test:
```
$ multipass launch --cloud-init - < file
# before
error loading cloud-init config: cannot read from stdin without a TTY

# after
Launched: …
```

### DIFF
--- a/src/platform/console/terminal.cpp
+++ b/src/platform/console/terminal.cpp
@@ -58,11 +58,6 @@ bool mp::Terminal::is_live() const
 
 std::string mp::Terminal::read_all_cin()
 {
-    if (!cin_is_live())
-    {
-        throw std::runtime_error("cannot read from stdin without a TTY");
-    }
-
     std::string content;
     char arr[1024];
     while (!cin().eof())


Saviq:
I want this in our next release, impact is minimal and otherwise it'd delay our ability to switch to `strict`, for Snapcraft will rely on this.

src/platform/console/terminal.cpp
```diff
@@ -58,11 +58,6 @@ bool mp::Terminal::is_live() const
 
 std::string mp::Terminal::read_all_cin()
 {
-    if (!cin_is_live())
-    {
-        throw std::runtime_error("cannot read from stdin without a TTY");
-    }
-
```
ricab: The windows terminal sets the mode of `stdin` to binary before calling `read_all_cin()`. I am not sure what the intended effects are, but are we sure they're maintained outside of a live tty?

Saviq: Looking at [the docs for it](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode) I can't see any exceptions. It just sets the mode on the file descriptor.

ricab: Yes, but to what end? Weren't there problems with this sort of stuff in the past? Like blocking exec commands and the like? This affects newlines, so could that be a cause? Could Gerry have put this in to fix something there?

Side note:

> If you apply _setmode to the file descriptor for a stream, call _setmode before you perform any input or output operations on the stream.

I wonder what happens if we don't, I don't think we ensure it.

Saviq: > Yes, but to what end? Weren't there problems with this sort of stuff in the past? Like blocking exec commands and the like? This affects newlines, so could that be a cause?

The problems you refer to were around the metrics question being asked, which should never be shown if we're not on a tty, or when there is anything to be read from stdin, because how do we differentiate between cloud-init being read and the answer to the question. Reading stdin and interactivity are conflicting. 

`exec` commands never reach `read_all_cin()`, it's only the cloud-init input handling that does.

>  Could Gerry have put this in to fix something there?

I can only guess that this was a wrong assumption that people will only feed it in from TTYs…

> Side note:
> 
> > If you apply _setmode to the file descriptor for a stream, call _setmode before you perform any input or output operations on the stream.
> 
> I wonder what happens if we don't, I don't think we ensure it.

One may wonder ;) You could also argue that for the YAML we _should_ be translating.  But it's out of scope for this PR.

PR #1182, opened 2019-11-15
submodules/grpc: Update grpc submodule branch

townsend2010:


ricab:
Hey @townsend2010 , just to let you know I tried this to see if it would help fix the debug-build failure in disco and I got earlier errors there (in boringssl). Those appeared in both debug and release. 

In both cases the problems are caused by warnings that are promoted to errors (I guess the compiler is more thorough). I wonder if we should disable that in 3rd party stuff, since it is out of our reach...

townsend2010:
Even though this was originally my PR, the work needed to get the coverage building was done by @ricab, so I'll review this :grin:

ricab:
@townsend2010 hey, just a heads-up that I had to force-push to fix a wrong comment.

ricab:
OK, thanks @townsend2010. 

bors r+

PR #1183, opened 2019-11-15
Add update balloon notification

### LINKED ISSUE #1072: [gui] add a notification for "New version of Multipass available"
We currently have a system tray item when there's an update available, [where possible](https://doc.qt.io/qt-5/qsystemtrayicon.html#supportsMessages) we should show a bubble, too - ideally resulting in opening https://multipass.run.

townsend2010:
Adds a notification for when an update is available.
Adds a first run notification to alert the user that the Multipass tray icon menu is running.

Fixes #1072 

---
The easiest way to test this is on Windows and modify the following files:
- `src/version.h.in`
  -  Change `constexpr auto version_string = "@MULTIPASS_VERSION@";` to `constexpr auto version_string = "0.6.0";` (or a version less than what is currently released).
- `src/client/gui/gui_cmd.cpp`
  - Change `about_update_timer.start(24h);` to `about_update_timer.start(60s);` (or whatever duration you'd like the notification to show).

Also, for the first run notification, due to a bug in Qt, the custom Multipass icon will not show in this notification on Ubuntu.  It should be there for the other platforms though.

### DIFF
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -208,6 +208,7 @@ void cmd::GuiCmd::update_about_menu()
     {
         update_action.setWhatsThis(QString::fromStdString(reply.update_info().url()));
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
+        tray_icon.showMessage("New Multipass update available", QString::fromStdString(reply.update_info().url()));
     }
     else
     {
@@ -220,6 +221,9 @@ void cmd::GuiCmd::create_menu()
     tray_icon.setContextMenu(&tray_icon_menu);
 
     tray_icon.setIcon(QIcon{":images/multipass-icon.png"});
+
+    QObject::connect(&tray_icon, &QSystemTrayIcon::messageClicked,
+                     [this] { QDesktopServices::openUrl(QUrl(update_action.whatsThis())); });
 
     QObject::connect(&list_watcher, &QFutureWatcher<ListReply>::finished, this, &GuiCmd::update_menu);
 
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -106,6 +106,16 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
     create_menu();
     tray_icon.show();
 
+    QFile first_run_file(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + "/first_run");
+
+    if (!first_run_file.exists())
+    {
+        tray_icon.showMessage("Multipass System Tray Menu", "The Multipass system tray menu is now running",
+                              tray_icon.icon());
+        first_run_file.open(QIODevice::WriteOnly);
+        first_run_file.close();
+    }
+
     return static_cast<ReturnCode>(QCoreApplication::exec());
 }
 
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -110,7 +110,11 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
 
     if (!first_run_file.exists())
     {
-        tray_icon.showMessage("Multipass System Tray Menu", "The Multipass system tray menu is now running",
+        // Each platform refers to the "system tray", icons, and the "menu bar" by different terminology.
+        // A platform dependent mechanism is used to get the terms via a QStringList.
+        auto notification_area_terms = mp::cli::platform::gui_tray_notification_strings();
+        tray_icon.showMessage(QString("Multipass is in your %1").arg(notification_area_terms[0]),
+                              QString("Click on the %1 for available options").arg(notification_area_terms[1]),
                               tray_icon.icon());
         first_run_file.open(QIODevice::WriteOnly);
         first_run_file.close();
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -111,11 +111,9 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
     if (!first_run_file.exists())
     {
         // Each platform refers to the "system tray", icons, and the "menu bar" by different terminology.
-        // A platform dependent mechanism is used to get the terms via a QStringList.
-        auto notification_area_terms = mp::cli::platform::gui_tray_notification_strings();
-        tray_icon.showMessage(QString("Multipass is in your %1").arg(notification_area_terms[0]),
-                              QString("Click on the %1 for available options").arg(notification_area_terms[1]),
-                              tray_icon.icon());
+        // A platform dependent mechanism is used to get the messages via a QStringList.
+        auto notification_area_strings = mp::cli::platform::gui_tray_notification_strings();
+        tray_icon.showMessage(notification_area_strings[0], notification_area_strings[1], tray_icon.icon());
         first_run_file.open(QIODevice::WriteOnly);
         first_run_file.close();
     }
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -221,7 +221,7 @@ void cmd::GuiCmd::update_about_menu()
         update_action.setWhatsThis(QString::fromStdString(reply.update_info().url()));
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
         tray_icon.showMessage("New Multipass update available",
-                              QString("Version %1 is availble. Click for more information.")
+                              QString("Version %1 is available. Click here for more information.")
                                   .arg(QString::fromStdString(reply.update_info().version())));
     }
     else
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -25,8 +25,9 @@
 #include <multipass/settings.h>
 #include <multipass/version.h>
 
-#include <QCoreApplication>
+#include <QApplication>
 #include <QDesktopServices>
+#include <QStyle>
 #include <QtConcurrent/QtConcurrent>
 
 namespace mp = multipass;
@@ -206,9 +207,12 @@ void cmd::GuiCmd::update_about_menu()
 
     if (update_available(reply.update_info()))
     {
+        update_action.setIcon(QApplication::style()->standardIcon(QStyle::SP_MessageBoxInformation));
         update_action.setWhatsThis(QString::fromStdString(reply.update_info().url()));
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
-        tray_icon.showMessage("New Multipass update available", QString::fromStdString(reply.update_info().url()));
+        tray_icon.showMessage("New Multipass update available",
+                              QString("Version %1 is availble. Click for more information.")
+                                  .arg(QString::fromStdString(reply.update_info().version())));
     }
     else
     {
@@ -255,7 +259,7 @@ void cmd::GuiCmd::create_menu()
     initiate_about_menu_layout();
 
     menu_update_timer.start(1s);
-    about_update_timer.start(12h);
+    about_update_timer.start(24h);
 }
 
 void cmd::GuiCmd::initiate_menu_layout()
--- a/include/multipass/cli/client_platform.h
+++ b/include/multipass/cli/client_platform.h
@@ -33,6 +33,7 @@ void parse_transfer_entry(const QString& entry, QString& path, QString& instance
 int getuid();
 int getgid();
 void open_multipass_shell(const QString& instance_name);
+QStringList gui_tray_notification_strings();
 }
 }
 }
--- a/src/platform/client/client_platform_linux.cpp
+++ b/src/platform/client/client_platform_linux.cpp
@@ -27,3 +27,8 @@ void mcp::open_multipass_shell(const QString& instance_name)
     QProcess::startDetached(
         "xterm", {"-title", instance_name, "-j", "-e", QString("multipass shell %1 || read").arg(instance_name)});
 }
+
+QStringList mcp::gui_tray_notification_strings()
+{
+    return {"System tray", "status icon"};
+}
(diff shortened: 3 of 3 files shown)


townsend2010:
Ok, this is ready for the first review pass.  I'm sure the strings will need to be tweaked :grin:

src/client/gui/gui_cmd.cpp
```diff
@@ -105,6 +106,16 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
...
 
+    QFile first_run_file(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + "/first_run");
+
+    if (!first_run_file.exists())
+    {
+        tray_icon.showMessage("Multipass System Tray Menu", "The Multipass system tray menu is now running",
+                              tray_icon.icon());
+        first_run_file.open(QIODevice::WriteOnly);
+        first_run_file.close();
+    }
```
Saviq: I wonder if we should just use the `Settings` class for this to avoid multiplying entities?

src/client/gui/gui_cmd.cpp
```diff
@@ -206,8 +217,12 @@ void cmd::GuiCmd::update_about_menu()
 
     if (update_available(reply.update_info()))
     {
+        update_action.setIcon(QApplication::style()->standardIcon(QStyle::SP_MessageBoxInformation));
         update_action.setWhatsThis(QString::fromStdString(reply.update_info().url()));
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
+        tray_icon.showMessage("New Multipass update available",
+                              QString("Version %1 is availble. Click for more information.")
```
Saviq:
```suggestion
                              QString("Version %1 is availble. Click here for more information.")
```
I know the "here" bit can't actually be a link, but "Click for…" kinda begs the question "where do I click again?".

src/client/gui/gui_cmd.cpp
townsend2010: Hmm, that would work, but it does somewhat expose a key to the user to muck around with.  That said, we wouldn't explicitly expose the key, so the user would have to look at source code to figure out how to set/unset it.  If we're fine with that, then I can move it into the `Settings` handler.

src/client/gui/gui_cmd.cpp
townsend2010: Ok, sure.

src/client/gui/gui_cmd.cpp
ricab: Well, it is not a setting, is it? It's an internal flag that we want to persist to and read from disk. The `Settings` class includes that capability but, as Chris points out, it goes beyond it, exposing things to the user. I'd rather not abuse it and leak internal implementation details. 

That would affect future settings frontends, including support for text editing (we'd have to identify and filter things). It is safer and cleaner to keep invariants alone than changing concepts just to include extra entities.

The extra entity will be odd either way. It has less of an impact placed alone than in the wrong group. If such entities got multiplied, a separate group could be created, similar to settings, but with a separate file and no user exposure.

Saviq:
Works! Only meh is the `multipass.gui.exe` identifier on Windows, think there's a way to make it say "Multipass"?

obraz

src/client/gui/gui_cmd.cpp
Saviq: I can easily see this as a "hidden" setting, when we introduce more "first run" behaviours, the user may find out that there is a "first_run" hidden setting if they want to go through them again (think a wizard or something).

src/client/gui/gui_cmd.cpp
```diff
@@ -105,6 +106,16 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
     create_menu();
     tray_icon.show();
 
+    QFile first_run_file(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + "/first_run");
+
+    if (!first_run_file.exists())
+    {
+        tray_icon.showMessage("Multipass System Tray Menu", "The Multipass system tray menu is now running",
```
Saviq: I'll sleep on this string, "system tray" apparently not a common enough term…

src/client/gui/gui_cmd.cpp
Saviq: At the very least, we should be able to use the same concept that Settings does for any such internal state storage.

I would rather make this a hidden setting for users to "muck about" with rather than have them muck about with the files directly.

src/client/gui/gui_cmd.cpp
townsend2010: Yeah, I struggled with the wording here...

townsend2010:
> Only meh is the `multipass.gui.exe` identifier on Windows,

Ok, I'll see if there is some way to change that.

src/client/gui/gui_cmd.cpp
ricab: One can always find possible uses for an implementation detail. If they come about, it can be consciously promoted to interface.

Settings are part of our interface, internal files aren't. If the user tampers with those, they're on their own. 

I understand the impulse to reuse functionality, I still don't see a strong reason to move the interface/impl boundary, especially since it's 3 LOC above.

src/client/gui/gui_cmd.cpp
townsend2010: Done.  And I also fixed a type in what I originally had :grin:

src/client/gui/gui_cmd.cpp
townsend2010: Thinking it about some more and reading the feedback, it just seems unnatural a "one and done" thing would be a setting.  To me, if we wanted to make this a setting, it should either always show the startup notification or not, something like `client.gui.show-start-notification` and be a boolean.  Otherwise, for true "first run" type of things, we should rely on something else such as what I currently have.

Saviq: > I understand the impulse to reuse functionality, but I still don't see a strong reason to move the interface/impl boundary, especially since it's 3 LOC above.

3 LOC here, 3 in a bunch of other places around the code… Just look in `/var/snap/multipass/common/data/multipassd` at the amount of files we're already storing some state in. Each of those can get corrupted and we have little protection against that. We're heavily multithreaded and it's mostly by chance that we're not stepping on one another's toes when writing / reading them.

Which is why I suggested that the _concept_ behind the Settings class can be useful for this, even if I got outvoted about allowing `multipass set/get …` to affect this state.

> it just seems unnatural a "one and done" thing would be a setting

To me, on the other hand, this is just state. I don't see a difference between a file that is created (and can be removed) and a boolean stored in a file somewhere. It's not like we can blow a fuse :).

That got me thinking, btw, that this shouldn't be a blanket "first run done". We should be more granular - "first run notification shown", because if we introduce some other do-once behaviours, we're likely to want them to run for users that ran Multipass already, but that have not, say, gone through a first-run wizard.

ricab: > Just look in /var/snap/multipass/common/data/multipassd at the amount of files we're already storing some state in.

So what? Files and dirs are nice units to organize non-volatile info. I am sure one or two things could be improved, but I don't see a structural problem.

> Each of those can get corrupted and we have little protection against that.

Anything can get corrupted. Whether the bytes are in separate or the same file has no bearing on it.

> To me, on the other hand, this is just state. I don't see a difference between a file that is created (and can be removed) and a boolean stored in a file somewhere.

The difference is public vs private &ndash; crucial distinction, probably the 1st consideration to organize data. The direct correspondence between our `Settings` concept and interface data is valuable. It makes everything clean and easy to reason about. It is worth preserving. Lumping everything together would hamper organization, not improve it.

> We should be more granular

I'd favor introducing complexity when it is needed.

townsend2010: > To me, on the other hand, this is just state. I don't see a difference between a file that is created (and can be removed) and a boolean stored in a file somewhere.

I definitely understand wanting to keep all persistent gui states in a single file. I do think these states should not be modifiable by the user other than potential interactions with the gui itself, so the keys should not be exposed to the user and not be modifiable by `multipass set`.  

I read up on the `QSettings` documentation and it suggests using it for storing persistent gui states.  So I think in the long run, using our `Settings` infrastructure is the right thing to do, but that requires expanding it to be able handle "hidden" keys, etc.

Given the time crunch of wanting another release out very soon, I don't see that it is possible to implement it in our `Settings` code right now.  we are going to have tech debt for this, so I say we use the simplest way to make it work, like what I currently have.  Either that, or drop this notification for now and add it in when we have time to implement it correctly.

Saviq: > > Just look in /var/snap/multipass/common/data/multipassd at the amount of files we're already storing some state in.
> 
> So what? Files and dirs are nice units to organize non-volatile info. I am sure one or two things could be improved, but I don't see a structural problem.
>
> > Each of those can get corrupted and we have little protection against that.
> 
> Anything can get corrupted. Whether the bytes are in separate or the same file has no bearing on it.

IMO the structural problem is that we don't have a central place that we handle this.
 
> > To me, on the other hand, this is just state. I don't see a difference between a file that is created (and can be removed) and a boolean stored in a file somewhere.
> 
> The difference is public vs private – crucial distinction, probably the 1st consideration to organize data. The direct correspondence between our `Settings` concept and interface data is valuable. It makes everything clean and easy to reason about. It is worth preserving. Lumping everything together would hamper organization, not improve it.

Which is why I said I'd be OK with the _concept_ behind Settings (the singleton class, the storage) being used for state, too - without exposing it to the user.

> > We should be more granular
> 
> I'd favor introducing complexity when it is needed.

That approach does extend the complexity though to migration. When we later want to introduce other first-run behaviours, we'd need to migrate the "first-run" state to a new one.

> > To me, on the other hand, this is just state. I don't see a difference between a file that is created (and can be removed) and a boolean stored in a file somewhere.
> 
> I definitely understand wanting to keep all persistent gui states in a single file. I do think these states should not be modifiable by the user other than potential interactions with the gui itself, so the keys should not be exposed to the user and not be modifiable by `multipass set`.

WFM

> I read up on the `QSettings` documentation and it suggests using it for storing persistent gui states. So I think in the long run, using our `Settings` infrastructure is the right thing to do, but that requires expanding it to be able handle "hidden" keys, etc.

Or just have a separate `State` singleton that will be internal only.

> Given the time crunch of wanting another release out very soon, I don't see that it is possible to implement it in our `Settings` code right now. we are going to have tech debt for this, so I say we use the simplest way to make it work, like what I currently have. Either that, or drop this notification for now and add it in when we have time to implement it correctly.

Yeah, agreed. If we don't want it in `Settings`, that is too much for now.

ricab: To be clear, I am not proposing a file per flag, but certificates separate from instances separate from metrics, I think is good. It's like single responsibility applied to files. Unfortunately this new flag just has no good company ATM.

Notice that more info in a single file -> more race potential. By using separate files for different purposes, you effectively delegate synchronization to the OS (threads don't collide in disk IO if they use their own file).

src/client/gui/gui_cmd.cpp
Saviq: We said this will need to be platform specific though?

It's ["System menu"](https://support.apple.com/guide/mac-help/menu-bar-mchlp1446/mac) on macOS, ["Notification area"](https://en.wikipedia.org/wiki/Taskbar#Taskbar_elements) on Windows, ["App indicators"/"Status icons"/"System tray"](https://blogs.gnome.org/aday/2017/08/31/status-icons-and-gnome/) on Linux…

I'm thinking:
- "Multipass is in your System menu", "Click on the icon in the menu bar for available options"
- "Multipass is in your Notification area", "Click on the icon in the taskbar for available options"
- "Multipass is in your System tray", "Click on the status icon for available options"

I can't see a better way…

src/client/gui/gui_cmd.cpp
ricab: Oops, I did not see the last message.

I will say I recognize some appeal in having a dedicated entity to route and gate disk IO, as long as private/public is distinguished. But I agree that would go beyond this PR.

src/client/gui/gui_cmd.cpp
townsend2010: > We said this will need to be platform specific though?

We originally said that because of what I thought at the time that Mac did not support clicking on the notification message and emitting a signal to invoke a slot (which is to open the default web browser to multipass.run).  However, that does work, so thus far, this code works across all platforms.  The only thing I've seen so far being platform specific is packaging changes to make this work.

That said, it's just strings and it's trivial to add platform specific string fetching via the platform interface.  I'll add that with the strings you suggest and we'll go from there.

src/client/gui/gui_cmd.cpp
Saviq: > certificates separate from instances separate from metrics, I think is good

FWIW I would say at least these three would be "good company" for this incoming flag:
- `multipassd-send-metrics.yaml`
- `network/multipass_subnet`
- `multipass-unique-id`

townsend2010:
The `multipass.gui.exe` identifier in Windows is fixed via a PR for `full`.

src/client/gui/gui_cmd.cpp
townsend2010: Ok, this is done.  It will of course break the other builds, but is trivial to fix that in full with the correct terminology.

src/client/gui/gui_cmd.cpp
ricab: Suppose we merge with those. If the gui starts around the time metrics are being treated, there is a collision. You now need to implement disk IO synchronization across processes, while that was previously done by the OS for free. I fail to see what exact gain would justify the cost.

Of course we could generalize this flag to a launch count and decide that was useful metrics, or make it some sort of setting, but that would be different motivation altogether.

Perhaps a disk IO entity could be useful in cases where there is already a race for the same data. But do we have many such cases? If we don't, aren't we better off just avoiding them? Having entities encapsulate their data in files?

ricab: One other point. Settings, by definition, are used by multiple entities (setter and settee). That's why they need a coordinating entity (our `Settings` class). 

(Talking about this, I realized the synchronization mechanism is missing: #1221)

ricab:
I just noticed one detail that might be improved: on windows, a right click is needed, perhaps the msg could reflect that.

Saviq:
@ricab thanks, will do.

bors r+

PR #1192, opened 2019-11-25
add core20 image urls

sd-hd:
Please update CI/CD so, the missing url's exists.

### DIFF
--- a/src/daemon/custom_image_host.cpp
+++ b/src/daemon/custom_image_host.cpp
@@ -69,6 +69,14 @@ const QMap<QString, CustomImageInfo> multipass_image_info{
       "core-18",
       "Core 18",
       "",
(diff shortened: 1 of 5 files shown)


Saviq:
Hi @sd-hd as you noticed, these images don't exist yet, so it's a bit early to be doing this ;)

Making a `core20` image is not a simple "replace 18 with 20" - the team is working hard to get it done.

You don't need the apparmor changes, though, those relate to how Multipass is built, not to what it's launching.

I'll close this until we know we have working `core20` images, thanks!

sd-hd:
HI @saviq,

do not mention.

I know, i'm new in the snapd / multipass universe,
but  if I can help, it would make me fun ;-)

PR #1196, opened 2019-11-29
[linux] Stop looking for libvirt in cmake

ricab:


townsend2010:
Hmm, `libvirt-dev` is still required.  Does this not break when building directly from source and `libvirt-dev` is not installed?

ricab:
Hey @townsend2010 `libvirt-dev` is part of the build dependencies [here](https://github.com/CanonicalLtd/multipass/blob/4a403fb8b9ee6ea32be2631ba810636c3528ccca/debian/control#L7), so it indeed needs to be installed for building.

Saviq:
Having discussed on IRC, I've reverted this. It's the config phase where missing dependencies should be caught.

PR #1207, opened 2019-12-05
WIP: Fix coverage build

ricab:


ricab:
So... this makes the coverage build pass by setting the required coverage options at the top level (as opposed to only `src`). It also replaces previous options with only `--coverage` in both compilation and linking, letting gcc take care of the details.

However, having this at the top level has new effects in the [coverage report](https://codecov.io/gh/canonical/multipass/pull/1207/tree):
  - `3rd-party` - does not seem to be included, not sure why. Exact delta is between #1282 with #1284 and the cov reports match.
  - `tests` - this becomes part of the report. Intuitively, I would say it should not be, but this is C++ and many of our headers lines are only hit when including it...

I did not find much about what the advised practice is, but perhaps we could try something like this: https://docs.codecov.io/docs/commit-status#section-excluding-tests-example . Or we can keep bypassing many header lines that seem to actually be tested... 

Here are the effects of different coverage configurations:
  * [this](https://codecov.io/gh/canonical/multipass/tree/3138e27c744fe2ed889f2c73cff69b517becc9a6) is the same as this PR but with options applied to `src`, `tests`, and `3rd-party` sub-folders separately (just a base for comparison). As expected, the results are pretty much the same (except for a test I deleted locally and ended up pushing in the throwaway commits). 
  * [This](https://codecov.io/gh/canonical/multipass/tree/ad2b9c2c4f884c9186ff68f248ed4a36a0764b75) adds the link options to fix the build, but only passes `--coverage` compilation option in `src`. As expected, the results are comparable to what we [currently have](https://codecov.io/gh/canonical/multipass/tree/ff85ac387c150d093991f4c9566860b32a348801).
  * I also tried to remove optimization (`-O0`), because gcc man says "Coverage data matches the source files more closely if do not optimize.". It did not seem to have [any effect](https://codecov.io/gh/canonical/multipass/tree/0c14a8d39e1622a5454259ce6c4744b5e31a030c).
  * Comparing [this](https://codecov.io/gh/canonical/multipass/tree/ad2b9c2c4f884c9186ff68f248ed4a36a0764b75) with [this](https://codecov.io/gh/canonical/multipass/tree/1a3e2600fec9cca2c56855a15897a8f6cb44b2dc) shows that enabling coverage in 3rd-party does not seem to have any impact.

PR #1210, opened 2019-12-09
[update] use multipass.run rather than github direct

Saviq:
---

This will work once https://github.com/canonical-web-and-design/multipass.run/pull/62 lands.

### DIFF
--- a/src/platform/update/CMakeLists.txt
+++ b/src/platform/update/CMakeLists.txt
@@ -15,7 +15,7 @@
 set(CMAKE_AUTOMOC ON)
 
 add_library(update STATIC
-  github_update_prompt.cpp
+  default_update_prompt.cpp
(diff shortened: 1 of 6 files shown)


Saviq:
@ricab this should be good to review again, the JSON file is in place.

Saviq:
> @Saviq I pushed a commit to update field names, otherwise it wouldn't work.
> 
> That shows this wasn't really tested, so I am not really comfortable approving. And I don't think I should be the one reviewing anymore ? (since I pushed multiple commits)

Well, yeah, testing was waiting for the JSON to show up at the URL… And then they used the field names from YAML rather than the original ones from GitHub I had there originally.

@townsend2010 would you please have a last-pass look?

townsend2010:
@Saviq, yup, will review

PR #1226, opened 2019-12-12
gui: Only connect the messageClicked signal when there is an actual u…

townsend2010:
…pdate


townsend2010:
@Saviq,
> You can't know the click was on the update notification only?

Unfortunately not :frowning_face: 

All Qt allows is to connect a slot to the `QSystemTrayIcon::messageClicked` signal for *anytime* a notification is clicked when `showMessage()` is invoked.  There is no granularity here.

PR #1227, opened 2019-12-12
WIP: [primary] Auto-mount home in primary on launch

### LINKED ISSUE #1224: [primary] mount $HOME on creation
`multipass shell` and `multipass start` should add a mount of the host's home directory into `primary:Home`.

ricab:
Fixes #1224.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -22,11 +22,13 @@
 #include <multipass/cli/argparser.h>
 #include <multipass/cli/client_platform.h>
 #include <multipass/constants.h>
+#include <multipass/settings.h>
 
 #include <multipass/format.h>
 
 #include <yaml-cpp/yaml.h>
 
+#include <QDir>
 #include <QTimeZone>
 
 #include <regex>
@@ -47,15 +49,23 @@ const std::regex show{"s|show", std::regex::icase | std::regex::optimize};
 
 mp::ReturnCode cmd::Launch::run(mp::ArgParser* parser)
 {
-    auto ret = parse_args(parser);
-    if (ret != ParseCode::Ok)
+    if (auto ret = parse_args(parser); ret != ParseCode::Ok)
     {
         return parser->returnCodeFrom(ret);
     }
 
     request.set_time_zone(QTimeZone::systemTimeZoneId().toStdString());
 
-    return request_launch();
+    auto ret = request_launch();
+    const auto petenv_name = Settings::instance().get(petenv_key); // TODO@ricab make attr and reuse in help
+    if (ret == ReturnCode::Ok && request.instance_name() == petenv_name.toStdString())
+    {
+        const auto mount_source = QDir::toNativeSeparators(QDir::homePath()); // TODO@ricab test on other platforms
+        const auto mount_target = QString{"%1:Home"}.arg(petenv_name);        // TODO@ricab extract home
+        ret = run_cmd_and_retry({"multipass", "mount", mount_source, mount_target}, parser, cout, cerr);
+    }
+
+    return ret;
 }
 
 std::string cmd::Launch::name() const
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -62,7 +62,7 @@ mp::ReturnCode cmd::Launch::run(mp::ArgParser* parser)
     {
         const auto mount_source = QDir::toNativeSeparators(QDir::homePath()); // TODO@ricab test on other platforms
         const auto mount_target = QString{"%1:Home"}.arg(petenv_name);        // TODO@ricab extract home
-        ret = run_cmd_and_retry({"multipass", "mount", mount_source, mount_target}, parser, cout, cerr);
+        ret = run_cmd({"multipass", "mount", mount_source, mount_target}, parser, cout, cerr);
     }
 
     return ret;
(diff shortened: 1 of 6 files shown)


ricab:
Voila, this should do it :smile: @Saviq could you please confirm this works as expected overall?

Polishing and tests missing.

Saviq:
@ricab yeah this seems good :)

It's a shame it takes a bit long… Could we fire it off and not wait for it to complete in the client?

ricab:
For the record, discussed in IRC and leaving that as a future improvement.

src/client/cli/cmd/launch.cpp
```diff
@@ -47,15 +49,27 @@ const std::regex show{"s|show", std::regex::icase | std::regex::optimize};
...
 
     request.set_time_zone(QTimeZone::systemTimeZoneId().toStdString());
 
-    return request_launch();
+    auto ret = request_launch();
+    if (ret == ReturnCode::Ok && request.instance_name() == petenv_name.toStdString())
+    {
+        const auto mount_source = QDir::toNativeSeparators(QDir::homePath());
+        const auto mount_target = QString{"%1:%2"}.arg(petenv_name, mp::home_automount_dir);
+        cout << fmt::format("Now auto-mounting '{}' in '{}'\n", mount_source, mount_target);
```
Saviq: I think this reads a tad better:

```suggestion
        cout << fmt::format("Mounting '{}' into '{}'\n", mount_source, mount_target);
```

src/client/cli/cmd/launch.cpp
```diff
@@ -47,15 +49,27 @@ const std::regex show{"s|show", std::regex::icase | std::regex::optimize};
...
+
+        ret = run_cmd({"multipass", "mount", mount_source, mount_target}, parser, cout, cerr);
+        if (ret == ReturnCode::Ok)
+            cout << "Auto-mount succeeded\n";
```
Saviq: I don't think we should print anything on success. And failure would have printed the problem already?

ricab: I initially did not have this, but then thought it looked strange to leave the last message on the screen saying we were mounting, while the previous one said "Launched". OTH, If I just let the mount spinner overwrite the "now mounting" message, it would not be seen (I tried this first, it would quickly be replaced with the "enabling mount support" msg). I also considered saying "Launched primary. Mounting..." in the same line so that we'd have only two lines when saying "mounted". WDYT?

src/client/cli/cmd/launch.cpp
ricab: I was about to commit but I thought we will probably have a setting with "automount" in the name? If so, saying "auto-mounting" here might help the user make the connection. Please confirm if you don't think that's relevant so I can commit.

Saviq: I think "auto-" is implied here, since it happens on its own :)

Saying "auto-mounting" here suggests it differs from "mounting", so I'd stick to the latter.

src/client/cli/cmd/launch.cpp
Saviq: @ricab or maybe we only print on success?

Remove the above and then:
```cpp
        if (ret == ReturnCode::Ok)
            cout << fmt::format("Mounted '{}' into '{}'\n", mount_source, mount_target);
```

We still get feedback about "Enabling support for mounting…" etc, so maybe that's all that's needed?

src/client/cli/cmd/launch.cpp
Saviq: But see below!

src/client/cli/cmd/launch.cpp
Saviq: "Enabling…" will be almost non-existent when we go for the `sshfs` snap, so that would result in:
```
Launched: primary
Mounted '/home/michal' in 'primary:Home'
```
WDYT?

src/client/cli/cmd/launch.cpp
ricab: Ack, I will incorporate below.

src/client/cli/cmd/launch.cpp
ricab: Yep, sounds good to me. The spinner erases the "enabling..." and later msgs, so that example is also what it will look like now after it is finished.

ricab:
@Saviq Sure, just trying to test this on windows and having weird trouble there.

ricab:
bors try

ricab:
@Saviq leaving the final call to you as per msgs in IRC

Saviq:
Yeah this is fine.

bors r+

ricab:
I am dropping the plan of adapting the remaining test to windows (`Client.launch_cmd_automounts_home_in_petenv`). The reason is that there doesn't seem to be a way to mock the user's home there without the Windows API. 

Qt obtains it from [`GetUserProfileDirectory`](https://docs.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-getuserprofiledirectorya) and [elsewhere](https://docs.microsoft.com/en-us/windows/win32/api/profinfo/ns-profinfo-profileinfoa?redirectedfrom=MSDN#remarks) they say:

> Do not use environment variables when specifying a path. The LoadUserProfile function does not expand environment variables, such as %username%, in a path. 

Found it out after trying that and other vars. There might be a way by changing things in the registry, but I'd rather not mess with that for this.

PR #1231, opened 2019-12-13
[daemon] throw a better error on sshfs installation issue

Saviq:


Saviq:
Aha, so it looks like this is falling through somewhere else? Will look.

Saviq:
@ricab found another place where this was thrown.

PR #1234, opened 2019-12-13
update: Use the website's JSON title and description in the update ve…

townsend2010:
…rbiage

### DIFF
--- a/src/client/cli/cmd/common_cli.cpp
+++ b/src/client/cli/cmd/common_cli.cpp
@@ -120,10 +120,8 @@ std::string cmd::instance_action_message_for(const mp::InstanceNames& instance_n
 
 std::string cmd::update_notice(const mp::UpdateInfo& update_info)
 {
-    return ::message_box("A new Multipass version " + update_info.version() +
-                         " is available!\n"
-                         "Find out more: " +
-                         update_info.url());
+    return ::message_box(update_info.title() + " is now available!\n\n" + "Major changes and enhancements include:\n" +
+                         update_info.description() + "\n\n" + "Find out more: " + update_info.url());
 }
 
 mp::ReturnCode cmd::run_cmd(const QStringList& args, const mp::ArgParser* parser, std::ostream& cout,
--- a/src/client/cli/cmd/common_cli.cpp
+++ b/src/client/cli/cmd/common_cli.cpp
@@ -120,8 +120,8 @@ std::string cmd::instance_action_message_for(const mp::InstanceNames& instance_n
 
 std::string cmd::update_notice(const mp::UpdateInfo& update_info)
 {
-    return ::message_box(update_info.title() + " is now available!\n\n" + "Major changes and enhancements include:\n" +
-                         update_info.description() + "\n\n" + "Find out more: " + update_info.url());
+    return ::message_box(fmt::format("{}\n{}\n\nGo here for more information: {}", update_info.title(),
+                                     update_info.description(), update_info.url()));
 }
 
 mp::ReturnCode cmd::run_cmd(const QStringList& args, const mp::ArgParser* parser, std::ostream& cout,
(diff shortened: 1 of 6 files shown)


src/client/cli/cmd/common_cli.cpp
```diff
@@ -120,10 +120,8 @@ std::string cmd::instance_action_message_for(const mp::InstanceNames& instance_n
 
 std::string cmd::update_notice(const mp::UpdateInfo& update_info)
 {
-    return ::message_box("A new Multipass version " + update_info.version() +
-                         " is available!\n"
-                         "Find out more: " +
-                         update_info.url());
+    return ::message_box(update_info.title() + " is now available!\n\n" + "Major changes and enhancements include:\n" +
+                         update_info.description() + "\n\n" + "Find out more: " + update_info.url());
```
Saviq: Don't base it off what's in the JSON today ;)

We should drive it from the JSON 1:1, otherwise we're stuck with whatever format you came up with here.

There's no reason we can't have CLI- and GUI-specific strings in there, or even translations later.

src/client/gui/gui_cmd.cpp
```diff
@@ -222,9 +222,9 @@ void cmd::GuiCmd::update_about_menu()
         update_action.setIcon(QApplication::style()->standardIcon(QStyle::SP_MessageBoxInformation));
         update_action.setWhatsThis(QString::fromStdString(reply.update_info().url()));
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
-        tray_icon.showMessage("New Multipass update available",
-                              QString("Version %1 is available. Click here for more information.")
-                                  .arg(QString::fromStdString(reply.update_info().version())));
+        tray_icon.showMessage(QString("%1 is now available!").arg(QString::fromStdString(reply.update_info().title())),
+                              QString("Major changes and enhancements include:\n%1")
+                                  .arg(QString::fromStdString(reply.update_info().description())));
```
Saviq: ditto

src/client/cli/cmd/common_cli.cpp
Saviq: Long-term we could even think of making these strings different depending on platform, something like:
```json
"title": {
  "default": "The default title",
  "windows-10": "Something specific to Windows",
  ...
}
```

And then query [`QSysInfo`](https://doc.qt.io/qt-5/qsysinfo.html) when displaying them.

But I don't think we need to do that just yet (OTOH we'll need a release to "enable" this before it actually works) ;)

townsend2010: @Saviq,

I see the value in just using the JSON 1:1 since you say we can be in control of what it says via the website including translations.  I'll change this to reflect that.

Regarding the actual translations and platform specific strings, I say we should wait on that and work out the true requirements needed to support that.

src/platform/update/new_release_monitor.cpp
```diff
@@ -35,7 +35,9 @@ namespace
 {
 constexpr auto timeout = std::chrono::minutes(1);
 constexpr auto json_tag_name = "version";
-constexpr auto json_html_url = "release_url";
+constexpr auto json_html_url = "download_url";
```
Saviq: Why this? It's `release_url` that has more details about the release, no?

townsend2010: @Saviq, I thought we wanted to drive folks to `multipass.run` and not `github`.

townsend2010:
Ugh, those tests *were* fixed in my original notifications branch, but that get reverted :frowning:

townsend2010:
I take it back, those tests are now broken in a different way than what I referred to before :grin:  Will fix, but it depends on what URL we want to return.

src/platform/update/new_release_monitor.cpp
Saviq: Not for this usecase I don't think, since they won't get any more details on https://multipass.run…

PR #1238, opened 2019-12-16
gui: Add a checkable "autostart on login" option in the About menu

townsend2010:
---

Merry Christmas!

Since this is kind of off the cuff, let's see if this looks ok and if we want it.

### DIFF
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -270,12 +270,17 @@ void cmd::GuiCmd::create_menu()
 
     about_menu.setTitle("About");
 
+    autostart_option.setCheckable(true);
+    autostart_option.setChecked(Settings::instance().get_as<bool>(autostart_key));
+    QObject::connect(&autostart_option, &QAction::toggled, this,
+                     [](bool checked) { Settings::instance().set(autostart_key, QVariant(checked).toString()); });
+
     about_client_version.setEnabled(false);
     about_daemon_version.setEnabled(false);
     about_copyright.setText("Copyright © 2017-2019 Canonical Ltd.");
     about_copyright.setEnabled(false);
 
-    about_menu.insertActions(0, {&about_client_version, &about_daemon_version, &about_copyright});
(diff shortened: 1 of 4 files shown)


Saviq:
obraz
obraz
Purrfect :)

src/client/gui/gui_cmd.cpp
```diff
@@ -133,6 +133,16 @@ mp::ReturnCode cmd::GuiCmd::run(mp::ArgParser* parser)
...
+    config_watcher.addPath(Settings::get_client_settings_file_path());
+    QObject::connect(&config_watcher, &QFileSystemWatcher::fileChanged, this, [this](const QString& path) {
+        autostart_option.setChecked(Settings::instance().get_as<bool>(autostart_key));
+        // Needed since the original watched file may be removed and opened as a new file
+        if (!config_watcher.files().contains(path) && QFile::exists(path))
+        {
+            config_watcher.addPath(path);
+        }
+    });
+
```
Saviq: This looks like it could be incorporated into Settings some way in the long run?

src/utils/settings.cpp
```diff
@@ -173,6 +173,11 @@ QString mp::Settings::get_daemon_settings_file_path() // temporary
     return file_for(daemon_root);
 }
 
+QString mp::Settings::get_client_settings_file_path() // idem
+{
+    return file_for("client");
```
Saviq: Nitpick: `"client"` looks like a candidate for `const auto client_root`?

townsend2010: Right, I thought of that, but then if I do that, I should change any other places that use the `client` string and that's unrelated to this PR.  If you're fine with that, then I'll change to a `constexpr`.

src/client/gui/gui_cmd.cpp
townsend2010: I think @ricab has some ideas on how to accomplish monitoring settings better, but that is for another day.

src/utils/settings.cpp
Saviq: Yeah, let's :)

townsend2010: Done.

Saviq:
Yess!

bors r+

PR #1246, opened 2019-12-18
backends/qemu: Create/destroy Process as needed to avoid stale args

### LINKED ISSUE #1243: Timeout on stop/start after daemon restart
1. `multipass start`
2. `snap restart multipass`
3. `multipass stop`
4. `multipass start`
=> Timeout

Daemon's log says `qemu-system-x86_64: Device 'hda' does not have the requested snapshot 'suspend'` when resuming from stopped (not when resuming from the daemon restart).

townsend2010:
Arguments needed to support suspend/resume were getting missed since arguments were only
defined on QemuVirtualMachine construction.

Fixes #1243

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -227,10 +227,7 @@ void mp::QemuVirtualMachine::start()
     if (state == State::suspending)
         throw std::runtime_error("cannot start the instance while suspending");
 
-    vm_process = make_qemu_process(
-        desc, ((state == State::suspended) ? mp::make_optional(monitor->retrieve_metadata_for(vm_name)) : mp::nullopt),
-        tap_device_name);
-    set_process_connections();
+    initialize_vm_process();
 
     if (state == State::suspended)
     {
@@ -452,8 +449,12 @@ void mp::QemuVirtualMachine::wait_until_ssh_up(std::chrono::milliseconds timeout
     }
 }
 
-void mp::QemuVirtualMachine::set_process_connections()
+void mp::QemuVirtualMachine::initialize_vm_process()
 {
+    vm_process = make_qemu_process(
+        desc, ((state == State::suspended) ? mp::make_optional(monitor->retrieve_metadata_for(vm_name)) : mp::nullopt),
+        tap_device_name);
+
     QObject::connect(vm_process.get(), &Process::started, [this]() {
         mpl::log(mpl::Level::info, vm_name, "process started");
         on_started();
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -209,10 +209,19 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
 
 mp::QemuVirtualMachine::~QemuVirtualMachine()
 {
-    update_shutdown_status = false;
-    if (state == State::running)
+    if (vm_process)
     {
-        suspend();
+        update_shutdown_status = false;
+
+        if (state == State::running)
+        {
+            suspend();
+        }
+        else
+        {
+            shutdown();
+        }
+
         vm_process->wait_for_finished();
     }
 
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -27,7 +27,6 @@
 #include <multipass/process.h>
 #include <multipass/ssh/ssh_session.h>
 #include <multipass/utils.h>
-#include <multipass/virtual_machine_description.h>
 #include <multipass/vm_status_monitor.h>
 
 #include <multipass/format.h>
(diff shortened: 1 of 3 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -303,12 +211,12 @@ mp::QemuVirtualMachine::~QemuVirtualMachine()
...
     if (state == State::running)
+    {
         suspend();
-    else
-        shutdown();
+        vm_process->wait_for_finished();
+    }
 
     remove_tap_device(QString::fromStdString(tap_device_name));
-    vm_process->wait_for_finished();
```
Saviq: So now Process's destruction deals with reaping the qemu process?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -319,12 +227,15 @@ void mp::QemuVirtualMachine::start()
     if (state == State::suspending)
         throw std::runtime_error("cannot start the instance while suspending");
 
+    vm_process = make_qemu_process(
+        desc, ((state == State::suspended) ? mp::make_optional(monitor->retrieve_metadata_for(vm_name)) : mp::nullopt),
+        tap_device_name);
+    set_process_connections();
```
Saviq: I think I'd be a little happier if we had a `set_vm_process()` that would deal with the connections, lower chance of forgetting to `set_process_connections()` after, wdyt? We don't have much of a precedent of this though…

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: @Saviq,

Not sure what you mean here.  I found the former logic to be incorrect, irregardless of this PR, albeit this PR exposed it.  The new logic is basically saying to reap the qemu process only when its running.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Again, I'm not really sure what you mean here.  Are you saying that I should rename the method to `set_vm_process` instead or move all of the creation of the vm_process object and setting up its signal/slot connections all in one method?  Actually, as I type this out, I prefer doing the later and will change that.  But please let me know :grin:

townsend2010:
> The failing tests:

Yep, totally forgot about fixing those up in my haste :smile:

src/platform/backends/qemu/qemu_virtual_machine.cpp
Saviq: @townsend2010 I mean that we're not calling `shutdown()` any more, so is it that `Process`/`QProcess`'s d'tor handling killing the process?

src/platform/backends/qemu/qemu_virtual_machine.cpp
Saviq: @townsend2010 basically, I'm used to setters and getters in Qt code :)

Pseudo-code:
```cpp
void mp::QemuVirtualMachine::start()
{
    // ...
    set_vm_process(std::move(make_qemu_process(...)));
    // ...
}

// ...

void set_vm_process(std::unique_ptr<Process> process)
{
    vm_process = process;
    // set the connections up...
}
```

But I won't block on that ;)

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: @Saviq,

This is what I mean by the incorrect logic.  If the process is running, we want to `suspend()` the instance when the dtor is called, ie, the daemon is exiting.  Otherwise, there is nothing to do and the process is not running.  The `shutdown()` there was really unnecessary the begin with.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Yeah, I'll do something like that.  I think it's cleaner and is done all in one place.

src/platform/backends/qemu/qemu_virtual_machine.cpp
Saviq: @townsend2010 and we don't care about any of the other states the vm might be in?

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Ok, renamed the private method and added the `make_qemu_process` handling in the method to have it all in one place.  I made this a private method instead of an anonymous namespace function because many class members are needed for this.

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: @Saviq, hmm, ok I see your point.  For other "running" types of states, I suppose it should shut down the instance.  However, I'm not sure for `delayed_shutdown`. Do we suspend and when the daemon starts again, the instance is just `running`?  Or do we just shut it down and not start on daemon start?

townsend2010: Ok, I added the logic back in, but wrapped it in a check to make sure `vm_process` is not null.

PR #1255, opened 2019-12-19
utils: Add function to check for a config file

### LINKED ISSUE #1242: GUI settings watch fails if no client settings were made yet
```
ricab@ricab-xps13:~ $ snap install --classic --channel beta/1.0 multipass
multipass (beta/1.0) 1.0.0 from Canonical✓ installed
ricab@ricab-xps13:~ $ multipass.gui
[2019-12-18T13:02:39.797] [warning] [Qt] inotify_add_watch("/home/ricab/snap/multipass/1525/config/multipass/multipass.conf") failed: "No such file or directory"
```

The GUI then does not see subsequent `multipass set client.gui.autostart=false` until it is restarted.

townsend2010:
Creates an empty config file if it doesn't exists.

Fixes #1242

### DIFF
--- a/tests/test_utils.cpp
+++ b/tests/test_utils.cpp
@@ -19,7 +19,9 @@
 
 #include "file_operations.h"
 #include "temp_dir.h"
+#include "temp_file.h"
 
+#include <QDateTime>
 #include <QRegExp>
 
 #include <gmock/gmock.h>
@@ -371,7 +373,7 @@ TEST(Utils, vm_stopped_returns_false)
     EXPECT_FALSE(mp::utils::is_running(state));
 }
 
-TEST(Utils, no_config_file_and_dir_are_created)
+TEST(Utils, absent_config_file_and_dir_are_created)
 {
     mpt::TempDir temp_dir;
     const QString config_file_path{QString("%1/config_dir/config").arg(temp_dir.path())};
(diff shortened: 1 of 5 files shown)


tests/test_utils.cpp
```diff
@@ -370,3 +370,13 @@ TEST(Utils, vm_stopped_returns_false)
 
     EXPECT_FALSE(mp::utils::is_running(state));
 }
+
+TEST(Utils, no_config_file_and_dir_are_created)
```
ricab:
```suggestion
TEST(Utils, absent_config_file_and_dir_are_created)
```

townsend2010:
@ricab, Sure, I'll do the `lastModifed` test and change the test name.

ricab:
@townsend2010 thanks! Just tested the snap and all works well :)

townsend2010:
@ricab, both parts are done.  Thanks!

tests/test_utils.cpp
townsend2010: Done

ricab:
Unfortunately bors won't like this because of the merge conflict in full...

ricab:
I was reminded bors is actually out after the org move, so merging will have to be done manually in 2020. The mac build has been restarted and we should get a mac package soon.

Saviq:
Okies,

bors r+

PR #1278, opened 2020-01-06
Fix delete crash

### LINKED ISSUE #1266: multipassd crashes when deleting a VM
I can't delete a stopped VM:

```
$ multipass list              
quiet-burbot            Stopped           --               Ubuntu 18.04 LTS
$ multipass delete --all
delete failed: cannot connect to the multipass socket
```

dmesg right after that:

```
[ 1669.814708] multipassd[32262]: segfault at 0 ip 00005590fa3efa14 sp 00007ffede3a4650 error 4 in multipassd[5590fa196000+736000]
[ 1669.814714] Code: 44 24 48 31 c0 8b 47 18 83 f8 07 0f 84 95 00 00 00 8d 50 fc 48 8b bf e8 01 00 00 83 fa 01 76 5e 83 f8 08 74 59 83 f8 02 74 44 <48> 8b 07 ff 90 90 00 00 00 48 8b bb e8 01 00 00 be 30 75 00 00 48
```

```
$ lsb_release -d
Description:	Ubuntu 18.04.3 LTS
$ snap info multipass | grep installed
installed:   1.0.1                           (1555) 185MB classic
```

I've also tried switching to the edge channel, but the problem persisted.

ricab:
Fixes #1266. `vm_process` is a pointer now and it is null when in off state. Shutdown is called when deleting, which was dereferencing the ptr to kill the process, even if it wasn't there.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -295,8 +295,11 @@ void mp::QemuVirtualMachine::shutdown()
         if (state == State::starting)
             update_shutdown_status = false;
 
-        vm_process->kill();
-        vm_process->wait_for_finished();
+        if (vm_process)
+        {
+            vm_process->kill();
+            vm_process->wait_for_finished();
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Saviq:
bors try

Saviq:
Some travis issue. Let's get bors through it again.

bors r+

PR #1285, opened 2020-01-08
[Trial] Cov separate tests

ricab:
Adds coverage measurement to test code, so that includes are also taken into account. Adds a codecov yaml to separate coverage reporting of tests.


ricab:
Changed the base so that we get a comparison with master.

Saviq:
Closing and reopening so the codecov app picks up.

PR #1286, opened 2020-01-08
ssh_session: Use SSH_OPTIONS_SSH_DIR to point libssh to a different d…

### LINKED ISSUE #1259: Complex `~/.ssh` may cause `multipass` to crash
I'm getting a segfault when running 'multipass shell'

```
bash-5.0$ multipass shell
Segmentation fault: 11
```

It works as the root user:

```
bash-5.0$ sudo multipass shell
Password:
Welcome to Ubuntu 18.04.3 LTS (GNU/Linux 4.15.0-72-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Fri Dec 20 10:20:08 NZDT 2019

  System load:  0.0               Processes:             106
  Usage of /:   24.0% of 4.67GB   Users logged in:       0
  Memory usage: 12%               IP address for enp0s2: 192.168.64.3
  Swap usage:   0%


0 packages can be updated.
0 updates are security updates.


Last login: Fri Dec 20 09:36:44 2019 from 192.168.64.1
ubuntu@primary:~$
```

Running MacOS Catalina:

```
bash-5.0$ uname -a
Darwin Robs-MacBook-Pro.local 19.0.0 Darwin Kernel Version 19.0.0: Thu Oct 17 16:17:15 PDT 2019; root:xnu-6153.41.3~29/RELEASE_X86_64 x86_64
```

townsend2010:
…irectory

Fixes #1259

### DIFF
--- a/src/ssh/ssh_session.cpp
+++ b/src/ssh/ssh_session.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -25,6 +25,8 @@
 
 #include <multipass/format.h>
 
+#include <QStandardPaths>
+
 #include <sstream>
 #include <stdexcept>
 #include <string>
@@ -40,6 +42,8 @@ mp::SSHSession::SSHSession(const std::string& host, int port, const std::string&
 
     const long timeout_secs = std::chrono::duration_cast<std::chrono::seconds>(timeout).count();
     const int nodelay{1};
+    auto ssh_dir =
+        QString("%1/.ssh").arg(QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)).toStdString();
 
     set_option(SSH_OPTIONS_HOST, host.c_str());
     set_option(SSH_OPTIONS_PORT, &port);
@@ -48,6 +52,7 @@ mp::SSHSession::SSHSession(const std::string& host, int port, const std::string&
     set_option(SSH_OPTIONS_NODELAY, &nodelay);
     set_option(SSH_OPTIONS_CIPHERS_C_S, "chacha20-poly1305@openssh.com,aes256-ctr");
     set_option(SSH_OPTIONS_CIPHERS_S_C, "chacha20-poly1305@openssh.com,aes256-ctr");
+    set_option(SSH_OPTIONS_SSH_DIR, ssh_dir.c_str());
 
     SSH::throw_on_error(session, "ssh connection failed", ssh_connect);
     if (key_provider)
@@ -106,6 +111,8 @@ const char* name_for(ssh_options_e type)
         return "client to server ciphers";
     case SSH_OPTIONS_CIPHERS_S_C:
         return "server to client ciphers";
+    case SSH_OPTIONS_SSH_DIR:
+        return "ssh config directory";
     default:
         break;
     }
@@ -120,6 +127,7 @@ std::string as_string(ssh_options_e type, const void* value)
     case SSH_OPTIONS_USER:
     case SSH_OPTIONS_CIPHERS_C_S:
     case SSH_OPTIONS_CIPHERS_S_C:
+    case SSH_OPTIONS_SSH_DIR:
         return std::string(reinterpret_cast<const char*>(value));
     case SSH_OPTIONS_PORT:
     case SSH_OPTIONS_NODELAY:
--- a/src/ssh/ssh_session.cpp
+++ b/src/ssh/ssh_session.cpp
@@ -25,6 +25,7 @@
 
 #include <multipass/format.h>
 
+#include <QDir>
 #include <QStandardPaths>
 
 #include <sstream>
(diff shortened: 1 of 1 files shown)


Saviq:
> Snap build available: `snap refresh multipass --channel edge/pr1286`

@lucj you should be able to test this now with this version ↑.

src/ssh/ssh_session.cpp
```diff
@@ -40,6 +42,8 @@ mp::SSHSession::SSHSession(const std::string& host, int port, const std::string&
 
     const long timeout_secs = std::chrono::duration_cast<std::chrono::seconds>(timeout).count();
     const int nodelay{1};
+    auto ssh_dir =
+        QString("%1/.ssh").arg(QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)).toStdString();
```
ricab: Two requests on the `QString("%1/.ssh")` bit (if it is to support an actual directory):
  - could this use `QDir::filePath` or a similar way to abstract the path for all platforms
  - could the directory not start with dot, just to play nicer on windows?

townsend2010:
Heya @ricab,

> If this is only meant to prevent ~/.ssh from being used, wouldn't it be better to use an impossible directory?

For now, yes, the main objective to not use the user's `.ssh` directory.  But I thought down the road, we might want to supply a Multipass specific ssh config, hence the use of a location that (hopefully) makes sense for that.

> Perhaps an empty string?

Unfortunately, `libssh` has a check for that and will return an error for such as case.  It seems strange though that `libssh` doesn't do the same if the directory doesn't exist :confused:

> OTH, if supporting a valid dir is intended, could we have a test to check that config there is actually picked up?

Picked up by `libssh` you mean?  Not sure if it would be worth the effort to mock whatever is needed in `libssh` to see if it actually sees and uses the config file.  Much of that code is "internal" to `libssh` (I know, it's C, so "internal" is relative).

src/ssh/ssh_session.cpp
townsend2010: You're right, it would be better to use `QDir` for that.  And yeah, really no need for the `.`, so I'll remove that.

ricab:
Hi @townsend2010, I was thinking a test that the ssh-config dir that Multipass now offers was actually honored. But you're right, that would require mocking the dir itself and then, as you point out, it would be mostly checking libssh functionality. 

Alternatively, there could be a test to check that the user's home config isn't used. That one would only have to mock HOME, put something in .ssh, and confirm it had no effect...

townsend2010:
> Alternatively, there could be a test to check that the user's home config isn't used. That one would only have to mock HOME, put something in .ssh, and confirm it had no effect...

Well, in order to do that, we'd need a mock ssh server that libssh could connect to and confirm via the mock ssh server that some options weren't being set.  Since this is a `libssh` issue itself and we don't test `libssh` itself, then adding a test for this is beyond the scope of what we do.  I really don't think trying to add a test here will accomplish much especially for the amount of work needed to really test what we need.

src/ssh/ssh_session.cpp
townsend2010: Ok, I pushed the changes requested here.

ricab:
@townsend2010 OK, I won't insist if you think it is too much work to only check the var is being set. 

Back to my earlier question, do you have a quick recipe for (manually) testing? Like some simple ssh setting you used to check that multipass does not pick it up?

townsend2010:
@ricab,

> Back to my earlier question, do you have a quick recipe for (manually) testing?

The bug report (#1259) attached to this bug only mentions some settings such as `proxyjump` or "pointers to keys that don't exist".  TBH, I'm just counting on `libssh` to DTRT and didn't test it since it wasn't clear to me what a valid ssh config was that caused `libssh` to crash.

In my opinion, if setting this option is not working in `libssh`, then that's a pretty severe `libssh` bug and we can report it there if people still claim this doesn't fix it.

ricab:
Man, sorry but I don't think things should go in without some form of testing! I guess we can accept the little decrease in coverage, given the cost-benefit you argued about, but at least some manual testing... 

I agree looking at the code, it does not seem likely to fail on multipass's side, but it rarely does and then that small probability sometimes materializes. There could be typos, other needed flags or vars for this to take effect, or just something unforeseen. That's why there are protocols for tests, reviews, etc. We're fallible, of course we never see a bug when we introduce it, testing helps mitigate the likelihood. 

If the team decides code can go in without any testing, I don't think that is the correct approach, but I'll abide by it. Otherwise, I don't really like to be in this position, but I can't go on approving something untested.

townsend2010:
@ricab,

I asked in the bug for a simple ssh config that can cause the issue.

Also, the code that sets the option is in a `throw`, so if `libssh` has a problem with how the new option is set, it will throw an exception and it doesn't when I've tried it which tells me `libssh` accepts the option.

townsend2010:
I also asked for any affected users to try it out and verify it fixes the issue for them.

ricab:
I don't follow what you mean by the option being in a `throw`. IIUC, all you need is an option in ~/.ssh with an effect you can observe before this PR, then confirming the option has no effect after this PR.

Is this not so? Is it unfeasible for some reason? If so, please let me know. 

If there is nothing stopping it but it is just something you don't want to do, I see two options:
  - you take the onus or merging yourself.
  - we clarify our reviewing guidelines and decide testing is not needed for code going in.

As it stands, I am asked to review PRs (which we said before should not decrease coverage). I find it contradictory to have a review procedure in place and then approve things that step over any sort of testing. This seems very obvious. Not honoring this is a sure recipe for more bugs. I would only ignore it if I was clearly told that is what is expected of me. Until then, please don't blame me for the PR not being in, it wasn't me deciding we need reviews, it wasn't me failing to test the change.

And yes, if the users said it fixed it for them, and since this is pretty one-dimensional in this case, I think that could reasonably be considered testing enough.

townsend2010:
I agree that we should figure out some way to verify this and should not be merged until verification is done.  I think the best way is to have an affected user do this for us.

PR #1290, opened 2020-01-09
Fix include coverage

ricab:
Enable coverage in test compilation, to measure header code execution properly (from include). Then, exclude the tests themselves from coverage reporting.


ricab:
Grrrr why won't github let me update the base...

PR #1294, opened 2020-01-09
Fix include coverage - based on master

ricab:
#1290 got closed after its base was merged and gh doesn't let me change the base or reopen.


Saviq:
bors r+

You should've known from the "Yups!"!

PR #1300, opened 2020-01-13
xz_decoder: Update submodule

townsend2010:


townsend2010:
@ricab,

I've seen this behavior when running `multipassd` from source builds.  Snap builds don't have the problem.  I haven't taken the time to figure out what is happening, but it's not related to this submodule update.

ricab:
@townsend2010 OK, good to go then.

Issue #1097, opened 2019-09-27
YAML colon handling incompatible with cloud-init's pyyaml


gucharbon:
### Context

- I'm running `multipass 0.8.0+win` on Windows 10 (17763).
- I'm trying to install Docker using cloud-init. 
- I'm using multipass to start a virtual machine using the following command:

`multipass launch -n swarm --cpus 2 --mem 4G --cloud-init cloud-config.yaml`

> Note: I didn't know how to name this issue, feel free to rename it

### Issue

`cloud-init` fails to load YAML with the following error:

```
2019-09-27 14:59:06,516 - util.py[WARNING]: Failed loading yaml blob. Invalid format at line 19 column 16: "while scanni
ng a plain scalar
  in "<unicode string>", line 19, column 16:
        -   [echo, Installing docker using script f ...
                   ^
found unexpected ':'
  in "<unicode string>", line 19, column 57:
     ... g docker using script from https://get.docker.com]
                                         ^
Please check http://pyyaml.org/wiki/YAMLColonInFlowContext for details."
2019-09-27 14:59:06,550 - util.py[WARNING]: Failed loading yaml blob. Invalid format at line 19 column 16: "while scanni
ng a plain scalar
  in "<unicode string>", line 19, column 16:
        -   [echo, Installing docker using script f ...
                   ^
found unexpected ':'
  in "<unicode string>", line 19, column 57:
     ... g docker using script from https://get.docker.com]
                                         ^
```

### Given cloud-config

```yaml
#cloud-config

groups:
  - docker

users:
  - name: gcharbon
    gecos: Guillaume Charbonnier
    primary_group: gcharbon
    groups: docker
    home: /home/gcharbon
    shell: /bin/bash
    lock_passwd: true
    sudo: ALL=(ALL) NOPASSWD:ALL
    ssh_authorized_keys:
      - ssh-rsa <ssh_key>

package_upgrade: true

rumcmd:
  - [ mkdir, /run/install_docker ]
  - [ echo, "Installing docker using script from https://get.docker.com" ]
  - [ curl, -fsSL, "https://get.docker.com", -o, /run/install_docker/get-docker.sh ]
  - [ sh, /run/install_docker/get-docker.sh ]
  - [ docker, swarm, init ]

final_message: >
   The system is finally up, after $UPTIME seconds.
   Docker has been installed and packages are up to date
```

### Additional information

The command `sudo cat /var/lib/cloud/instances/swarm/user-data.txt` returns the following:

[...]

PR #1301, opened 2020-01-13
yaml-cpp: Update submodule to version 0.6.3

townsend2010:


tests/test_daemon.cpp
```diff
@@ -532,10 +532,10 @@ MATCHER_P(YAMLSequenceContainsStringMap, values, "")
             continue;
         for (auto it = values.cbegin();; ++it)
         {
-            if (node[it->first].template as<std::string>() != it->second)
-                break;
-            else if (it == values.cend())
+            if (it == values.cend())
                 return true;
+            else if (node[it->first].template as<std::string>() != it->second)
+                break;
```
Saviq: Hey, can you briefly explain why this change? I can't immediately see the difference?

src/client/cli/formatter/yaml_formatter.cpp
```diff
@@ -66,7 +66,7 @@ std::string mp::YamlFormatter::format(const InfoReply& reply) const
             std::locale::global(std::locale("C"));
             auto loads = mp::utils::split(info.load(), " ");
             for (const auto& entry : loads)
-                instance_node["load"].push_back(std::stod(entry));
+                instance_node["load"].push_back(entry);
```
Saviq: This didn't work with new yaml-cpp, or did you just tweak it to look better?

townsend2010:
@Saviq,

> Does this have any bearing on #1097?

I've no idea, I haven't tried it :grin:  I only made sure it isn't broken for our normal use.  When I have a moment, I'll give it a try and see though.

tests/test_daemon.cpp
townsend2010: This has been a bug all along, but only exposed by the changes in the updated `yaml-cpp`.  We really want to check if `it == cend()` first, because if it is and using the previous code, then `it->first` is an invalid access.  With the new `yaml-cpp`, the previous code would result in a segfault.

src/client/cli/formatter/yaml_formatter.cpp
townsend2010: The new `yaml-cpp` switched to a new precision type, so we'd end up with long decimal values if `load` was kind of high.  When I saw this, I thought to myself that we should leave it as a string just as the `load` outputs and not modify it.

tests/test_daemon.cpp
Saviq: D'uh, of course :)

src/client/cli/formatter/yaml_formatter.cpp
Saviq: ACKies

Saviq:
This works still:
```yaml
$ multipass launch --cloud-init - <<EOF
users:
- default                                                                                                                                                                                                                                                             
- name: michal
  ssh_authorized_keys:
  - "$( cat ~/.ssh/id_rsa.pub )"
  sudo: "ALL=(ALL) NOPASSWD:ALL"
EOF
```

And that:
```yaml
$ multipass info --format yaml --all
errors:
  - ~
primary:
  - state: Running
    image_hash: a720c34066dce5521134c0efa63d524c53f40c68db24cf161d759356a24aad0e
    image_release: 18.04 LTS
    release: Ubuntu 18.04.3 LTS
    load:
      - 0.11
      - 0.13
      - 0.06
    disks:
      - sda1:
          used: 1205604352
          total: 5019643904
    memory:
      usage: 79708160
      total: 1032953856
    ipv4:
      - 10.253.230.172
    mounts:
      Home:
        uid_mappings:
          - 10000:default
        gid_mappings:
          - 10000:default
        source_path: /home/michal
```

bors r+

Saviq:
Network issue.

Once more, with feeling:

bors r+

Saviq:
OK, my fault now, this will work:

bors r+

PR #1302, opened 2020-01-14
submodules: Add txt file that lists submodule info

townsend2010:
- Includes version, repo url, and release url.


Saviq:
Hey, could you make this Markdown please? Also, I wonder if we should list the patches we're keeping as well, rather than just mentioning them? Or maybe link to a `$upstream_version..$our_head` diff on GitHub?

townsend2010:
> could you make this Markdown please?

Yep, will do.

> Also, I wonder if we should list the patches we're keeping as well, rather than just mentioning them? Or maybe link to a $upstream_version..$our_head diff on GitHub?

I think I'll go with the latter as it will be a single URL with all the info needed to see what the differences are.

townsend2010:
Ok, all updated.

townsend2010:
@Saviq,

I cherry picked your commit from the other PR and pushed it here.  You can approve at your convenience :smile:

townsend2010:
lol, I think coverage statistics are messed up.  Coverage drops with this PR???  I'm not even touching any covered files.

ricab:
Strange indeed. [This](https://docs.codecov.io/docs/unexpected-coverage-changes) may be relevant, although I still don't see an obvious reason. Maybe Travis got some update...

PR #1303, opened 2020-01-14
[settings] Lock checked get and set

### LINKED ISSUE #1221: Make Settings thread-safe
To protect from different threads/processes colliding when reading/writing settings. Currently it is not very likely, but it is still something to protect against.

ricab:
To make settings thread safe. Fixes #1221.

### DIFF
--- a/include/multipass/settings.h
+++ b/include/multipass/settings.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2019-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/include/multipass/settings.h
+++ b/include/multipass/settings.h
@@ -25,6 +25,7 @@
 #include <QVariant>
 
 #include <map>
+#include <mutex>
 #include <set>
 
 namespace multipass
@@ -51,6 +52,7 @@ private:
     void set_aux(const QString& key, QString val);
 
     std::map<QString, QString> defaults;
+    mutable std::mutex mutex;
 };
 } // namespace multipass
 
--- a/src/utils/settings.cpp
+++ b/src/utils/settings.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2019-2020 Canonical, Ltd.
  *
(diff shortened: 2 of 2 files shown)


ricab:
Unit tests for this would require mocking `QSettings` (beyond the existing `MockSettings`), which was so far deferred. Let me know if you think that should get higher priority.

I tested manually with the following patch:

```patch
diff --git a/src/client/cli/cmd/set.cpp b/src/client/cli/cmd/set.cpp
index a894102..4c52323 100644
--- a/src/client/cli/cmd/set.cpp
+++ b/src/client/cli/cmd/set.cpp
@@ -25,6 +25,9 @@
 #include <multipass/rpc/multipass.grpc.pb.h>
 #include <multipass/settings.h>
 
+#include <thread> // FIXME testing
+#include <iostream> // FIXME testing
+
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
 
@@ -36,10 +39,34 @@ bool allow_only_stopped_instances(const QString& key, const QString& val) // tem
            val != mp::Settings::instance().get(key); /* if we are switching between qemu and libvirt drivers (on linux),
                                                         we can only have stopped instances */
 }
+
+void run_test_threads() // FIXME testing
+{
+    std::vector<std::thread> threads{};
+
+    for(int i = 0; i < 1000; ++i)
+    {
+        threads.emplace_back([]() {
+            auto x = mp::Settings::instance().get("client.primary-name");
+            std::cout << x.toStdString() << "\n";
+        });
+        threads.emplace_back([i]() {
+            auto val = (i % 2) ? "foo" : "bar";
+            std::cout << "setting " << val << "\n";
+           mp::Settings::instance().set("client.primary-name", val);
+        });
+    }
+
+    std::cout << "joining\n";
+    for(auto& t : threads)
+        t.join();
+
+}
 } // namespace
 
 mp::ReturnCode cmd::Set::run(mp::ArgParser* parser)
 {
+    run_test_threads(); // FIXME testing
     auto parse_code = parse_args(parser);
     auto ret = parser->returnCodeFrom(parse_code);
     if (parse_code == ParseCode::Ok)
```

ricab:
Of course, well spotted.

ricab:
Done @townsend2010.

townsend2010:
@ricab, great, thanks!

bors r+

townsend2010:
Transient build issue it seems.  Try again...

bors r+

legns:
How my second user...s8

angel labelle

PR #1312, opened 2020-01-17
[qemu] Keep qemu error when shutdown during start

### LINKED ISSUE #882: Cryptic error when instance fails to start
`multipass` worked ok with `snapcraft` the first time. Then I renamed snap to `trx-audio` and now it fails.
```
$ snapcraft
Using 'snap/snapcraft.yaml': Project assets will be searched for from the 'snap' directory.
Launching a VM.
snapcraft-trx-audio: shutdown called while starting
An error occurred with the instance when trying to start with 'multipass': returned exit code 2.
$ snapcraft --version
snapcraft, version 3.6
$ multipass --version
multipass  0.7.1
multipassd 0.7.1
$ sudo snap logs -f multipass
2019-07-03T07:03:25Z multipassd[13419]: instance "snapcraft-trx-audio" does not need stopping
2019-07-03T07:03:37Z multipassd[13419]: process state changed to Starting
2019-07-03T07:03:37Z multipassd[13419]: process state changed to Running
2019-07-03T07:03:37Z multipassd[13419]: process started
$ multipass start snapcraft-trx-audio -vvv
Starting snapcraft-trx-audio /[2019-07-03T10:07:07.230] [info] [snapcraft-trx-audio] process state changed to Starting

 /[2019-07-03T10:07:07.234] [info] [snapcraft-trx-audio] process state changed to Running

  [2019-07-03T10:07:07.234] [info] [snapcraft-trx-audio] process started        

snapcraft-trx-audio: shutdown called while starting
```

ricab:
Fixes #882. The error message from qemu says it all, and we had it already saved, but it was being overwritten in `on_shutdown`. This is useful also for other cases beside OOM (see tests below).

I tried adding tests for this, but I am being bitten by a variation of #1003. That is something that I would like to have a go at, otherwise it may always byte us back when we try to add tests. So keeping this no-merge for now, will check next week.

Here is how I tested manually:

### Test too much memory

1. `multipass launch --mem 1234G`

Output before fix:

```
first-crane: shutdown called while starting
```

And after fix:

```
qemu-system-x86_64: cannot set up guest memory 'pc.ram': Cannot allocate memory
equipped-robin: shutdown called while starting
```

### Test by removing firmware

This on a local build.

1. `sudo mv /usr/share/seabios{,.bak}` don't forget to undo this
2. run daemon
3. `multipass launch`

Output before fix:

```
speedy-grison: shutdown called while starting
```
And after fix:

```
qemu: could not load PC BIOS 'bios-256k.bin'
casual-haddock: shutdown called while starting
```

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -365,7 +365,9 @@ void mp::QemuVirtualMachine::on_shutdown()
     std::unique_lock<decltype(state_mutex)> lock{state_mutex};
     if (state == State::starting)
     {
-        saved_error_msg = fmt::format("{}: shutdown called while starting", vm_name);
+        if (!saved_error_msg.empty() && saved_error_msg.back() != '\n')
+            saved_error_msg.append("\n");
+        saved_error_msg.append(fmt::format("{}: shutdown called while starting", vm_name));
         state_wait.wait(lock, [this] { return state == State::off; });
     }
     else
--- a/tests/qemu/test_qemu_backend.cpp
+++ b/tests/qemu/test_qemu_backend.cpp
@@ -35,6 +35,7 @@
 #include <multipass/virtual_machine_description.h>
 
 #include <QJsonArray>
+#include <thread>
 
 namespace mp = multipass;
 namespace mpt = multipass::test;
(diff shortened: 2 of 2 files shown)


ricab:
Had to rebase, since this needs #1316, but created duplicate branch, to avoid closing when that is merged. Please change base to master when that is in and delete branch  `fix-mock-injection-dup` before merging this.

ricab:
Note: the second test above no longer applies, it will now produce a different error (after rebase).

ricab:
I have been cracking my head trying to get that diff coverage (earlier commits were stepping stones that increased coverage only outside of this PR's diff). It turned out to be much trickier than I thought. I kept running into strange crashes, deadlocks, and other problems, leaving it just beyond reach each time.

This should do it, but I had to resort to thread sleeping. The tested code blocks on a condition variable, which is supposed to be notified from another thread. The notifying thread (the main thread here) sleeps to give the blocking thread time to reach the blocking instruction, so it can then be unblocked. I didn't see a better way, but suggestions would be welcome. I got so buried in this I may be missing something obvious.

ricab:
Finally ready :)

PR #1319, opened 2020-01-21
daemon: Also check for the restarting state on start()

### LINKED ISSUE #1315: multipassd crash
After one of my multipass VM's went to restart (it was a Ubuntu Core 16 VM if that's relevant), multipassd crashed: https://pastebin.ubuntu.com/p/Grm5WNznRb/

There were no updates and no `snap restart` commands, etc.

townsend2010:
No need to issue a backend start() when the instance is in a restarting state.

Fixes #1315

---

I tested using the following method:

- Create an instance.
- `multipass exec <foo> -- sudo reboot`
- Confirm instance is in the `restarting` state.
- Issue `multipass start <foo>` while instance is in the `restarting` state.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1435,7 +1435,8 @@ try // clang-format on
     for (const auto& name : vms)
     {
         auto it = vm_instances.find(name);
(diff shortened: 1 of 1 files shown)


ricab:
bors still fishing

bors r+

Issue #1322, opened 2020-01-23
`Transport endpoint is not connected` with snapcraft (large files)


cjp256:
Ian Johnson reported an issue he was having with snapcraft on a project with large images.  I tried to reproduce with roughly the following steps:

Create a 10G file in a project (any project):

```
chris@linvy:~/git/figlet$ dd if=/dev/zero of=foobar.bin bs=1M count=10000
```

Builds with multipass fine some number of times:

```
chris@linvy:~/git/figlet$ SNAPCRAFT_BUILD_ENVIRONMENT=multipass snapcraft

<snipped>

Snapping 'figlet' |                                    
Snapped figlet_2.2.5+git23.5bbcd73-dirty_amd64.snap
```

At some point:

```
chris@linvy:~/git/figlet$ SNAPCRAFT_BUILD_ENVIRONMENT=multipass snapcraft
Launching a VM.
                                                                                                                                                                                                              Updating pull step for figlet (source changed)
[Errno 107] Transport endpoint is not connected
```

Journal logs show an interesting crash:

```
chris@linvy:~/git/figlet$ journalctl -f -n 200 | grep multipass 

<snipped>

Jan 23 12:44:18 linvy audit[19574]: AVC apparmor="STATUS" operation="profile_remove" info="profile does no
t exist" error=-2 profile="snap.multipass.multipassd" name="multipass.snapcraft-figlet.73b2fff1.sshfs_server" pid=19574 comm="apparmor_parser"
Jan 23 12:44:18 linvy kernel: audit: type=1400 audit(1579801458.274:14859): apparmor="STATUS" operation="profile_remove" info="profile does not exist" error=-2 profile="snap.multipass.multipassd" name="multipass.snapcraft-figlet.73b2fff1.sshfs_server" pid=19574 comm="apparmor_parser"
Jan 23 12:44:18 linvy multipassd[32469]: Removing AppArmor policy: 
                                         profile multipass.snapcraft-figlet.73b2fff1.sshfs_server flags=(attach_disconnected) {
                                             # Allow multipassd send sshfs_server signals
                                             signal (receive) peer=snap.multipass.multipassd,
                                             /snap/multipass/1597/bin/sshfs_server ixr,
                                             /snap/multipass/1597/{usr/,}lib/** rm,
Jan 23 12:44:18 linvy multipassd[32469]: Failed to remove AppArmor policy 
[...]

Issue #1310, opened 2020-01-16
Unaligned ttyname/ioctl errors


ricab:
Noticed some unaligned ttyname/ioctl errors when running snapcraft with edge-channel multipass (1.1.0-dev.47+g31ad44c). 

```console
ricab@ricab-xps13:~/mydev/test-snap/snap $ snap refresh --edge multipass
multipass (edge) 1.1.0-dev.47+g31ad44c from Canonical✓ refreshed
ricab@ricab-xps13:~/mydev/test-snap/snap $ snapcraft
Launching a VM.
Launched: snapcraft-test-snap
mesg: ttyname failed: Inappropriate ioctl for device
                                                    mesg: ttyname failed: Inappropriate ioctl for device
   mesg: ttyname failed: Inappropriate ioctl for device
                                                       error: no changes of type "auto-refresh" found
2020-01-16T18:28:40Z INFO Waiting for restart...
snapd 2.42.5 from Canonical✓ installed
[...]
```

Snapcraft otherwise completes well. These messages don't appear with stable (1.0.2). Could this be related to libssh now polling for stderr? 

I do not observe this when using multipass directly. This may well be under snapcraft scope, but opening an issue anyway so it can be discussed and maybe routed.

townsend2010:
@ricab,

Yes, this is due to both the `libssh` patch to fix reading `stderr` and now handling redirected `stdio` properly.  It's basically the result of `snapcraft` using the Python `subprocess.check_output` when using `sudo -i` inside the instance.  An example would be:
```
python -c "import subprocess;subprocess.check_output(['bin/multipass', 'exec', 'foo', '--', 'sudo', '-i', 'echo', 'hi'])"
```

As far as the unaligned output, yeah, I saw that and it needs to be fixed but it's only a visual thing and doesn't affect functionality.  BTW, I think the fix needs to happen in `libssh`.

ricab:
Hi @townsend2010 , I see. So IIUC the issue is them using a login shell, which ends up executing `mesg`, which doesn't get a tty but needs one, correct? Not much we can do on our side I suppose.

As to the unaligned output, is that something you have seen in other cases too?

Issue #962, opened 2019-08-13
Verbose mode doesn't work for launch


Andras-Csanyi:
I try to figure out why my Virtualbox backed VM doesn't start. Verbose mode doesn't work at all.

```
PS C:\Users\Andras Csanyi> multipass launch ubuntu -n sdf -vvv
launch failed: The following errors occurred:
sdf: timed out waiting for response
PS C:\Users\Andras Csanyi>
```

Saviq:
You're right, somehow we're losing the extended logging info in the `launch` command. Thanks for noticing!

PR #1328, opened 2020-01-30
Install the multipass-sshfs Snap for mount support 

### LINKED ISSUE #627: Add static sshfs binary
To avoid issues with installing sshfs on instances that lack it, we can ship our own static version of it into the instance.

townsend2010:
Fixes #627

---

Notes about testing:
1. If an instance already has a mount defined, then `multipassd` should just use the `sshfs` that is already installed.
1. Any new instances or existing instances without a mount defined that then has a mount defined should install the `multipass-sshfs` snap and use that.
1. Mounts on `core16` and `core18` should work now.

### DIFF
--- a/include/multipass/sshfs_mount/sshfs_mount.h
+++ b/include/multipass/sshfs_mount/sshfs_mount.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -29,6 +29,12 @@ class SftpServer;
 class SshfsMount
 {
 public:
+    enum class SshfsPkgType
+    {
+        debian,
+        snap
+    };
+
     SshfsMount(SSHSession&& session, const std::string& source, const std::string& target,
                const std::unordered_map<int, int>& gid_map, const std::unordered_map<int, int>& uid_map);
     SshfsMount(SshfsMount&& other);
--- a/include/multipass/sshfs_mount/sshfs_mount.h
+++ b/include/multipass/sshfs_mount/sshfs_mount.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2020 Canonical, Ltd.
+ * Copyright (C) 2017 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -29,12 +29,6 @@ class SftpServer;
 class SshfsMount
 {
 public:
-    enum class SshfsPkgType
-    {
-        debian,
-        snap
-    };
-
     SshfsMount(SSHSession&& session, const std::string& source, const std::string& target,
                const std::unordered_map<int, int>& gid_map, const std::unordered_map<int, int>& uid_map);
     SshfsMount(SshfsMount&& other);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -77,8 +77,9 @@ constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
 constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
-const std::string sshfs_error_template = "Error enabling mount support in '{}'"
-                                         "\n\nPlease install 'sshfs' manually inside the instance.";
+const std::string sshfs_error_template =
+    "Error enabling mount support in '{}'"
+    "\n\nPlease install the 'multipass-sshfs' snap package manually inside the instance.";
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
@@ -2067,9 +2068,10 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
     {
         mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed for \'{}\'", name));
         throw std::runtime_error(
-            fmt::format("Snap support needs to be installed for \'{}\' in order to support mounts. "
-                        "Please see https://docs.snapcraft.io/installing-snapd for information on "
-                        "how to install Snap support for your instance's distribution.",
+            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts.\n\n"
+                        "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
+                        "\nhow to install Snap support for your instance's distribution."
+                        "\n\nAlternatively, install `sshfs` manually inside the instance.",
                         name));
     }
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2063,8 +2063,8 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
     mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap in \'{}\'", name));
 
     // Check if snap support is installed in the instance
-    auto proc = session.exec("which snap");
-    if (proc.exit_code() != 0)
+    auto which_proc = session.exec("which snap");
+    if (which_proc.exit_code() != 0)
     {
         mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in \'{}\'", name));
         throw std::runtime_error(
@@ -2072,6 +2072,18 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
                         "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
                         "\nhow to install snap support for your instance's distribution."
                         "\n\nAlternatively, install `sshfs` manually inside the instance.",
+                        name));
+    }
+
+    // Check if /snap exists for "classic" snap support
+    auto test_file_proc = session.exec("[ -e /snap ]");
+    if (test_file_proc.exit_code() != 0)
+    {
+        mpl::log(mpl::Level::warning, category, fmt::format("Classic snap support symlink is needed in \'{}\'", name));
+        throw std::runtime_error(
+            fmt::format("Classic snap support is needed in \'{}\' in order to support mounts."
+                        "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
+                        "\nhow to enable classic snap support for your instance's distribution.",
                         name));
     }
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2060,17 +2060,17 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
 
     SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
 
-    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs Snap for \'{}\'", name));
+    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap for \'{}\'", name));
 
     // Check if Snap support is installed in the instance
     auto proc = session.exec("which snap");
     if (proc.exit_code() != 0)
     {
-        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed for \'{}\'", name));
+        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in \'{}\'", name));
         throw std::runtime_error(
             fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts.\n\n"
                         "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
-                        "\nhow to install Snap support for your instance's distribution."
+                        "\nhow to install snap support for your instance's distribution."
                         "\n\nAlternatively, install `sshfs` manually inside the instance.",
                         name));
     }
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -77,9 +77,8 @@ constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
 constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
-const std::string sshfs_error_template =
-    "Error enabling mount support in '{}'"
-    "\n\nPlease install the 'multipass-sshfs' snap manually inside the instance.";
+const std::string sshfs_error_template = "Error enabling mount support in '{}'"
+                                         "\n\nPlease install the 'multipass-sshfs' snap manually inside the instance.";
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -79,7 +79,7 @@ constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
 const std::string sshfs_error_template =
     "Error enabling mount support in '{}'"
-    "\n\nPlease install the 'multipass-sshfs' snap package manually inside the instance.";
+    "\n\nPlease install the 'multipass-sshfs' snap manually inside the instance.";
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
@@ -2060,15 +2060,15 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
 
     SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
 
-    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap for \'{}\'", name));
+    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap in \'{}\'", name));
 
-    // Check if Snap support is installed in the instance
+    // Check if snap support is installed in the instance
     auto proc = session.exec("which snap");
     if (proc.exit_code() != 0)
     {
         mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in \'{}\'", name));
         throw std::runtime_error(
-            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts.\n\n"
+            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts."
                         "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
                         "\nhow to install snap support for your instance's distribution."
                         "\n\nAlternatively, install `sshfs` manually inside the instance.",
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -77,7 +77,6 @@ constexpr auto reboot_cmd = "sudo reboot";
 constexpr auto up_timeout = 2min; // This may be tweaked as appropriate and used in places that wait for ssh to be up
 constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
-constexpr auto max_install_sshfs_retries = 3;
 const std::string sshfs_error_template = "Error enabling mount support in '{}'"
                                          "\n\nPlease install 'sshfs' manually inside the instance.";
 
@@ -2060,33 +2059,36 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
 
     SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
 
-    mpl::log(mpl::Level::info, category, fmt::format("Installing sshfs in \'{}\'", name));
+    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs Snap for \'{}\'", name));
 
-    int retries{0};
-    while (++retries <= max_install_sshfs_retries)
+    // Check if Snap support is installed in the instance
+    auto proc = session.exec("which snap");
+    if (proc.exit_code() != 0)
     {
-        try
-        {
-            auto proc = session.exec("sudo apt update && sudo apt install -y sshfs");
-            if (proc.exit_code(std::chrono::minutes(5)) != 0)
-            {
-                auto error_msg = proc.read_std_error();
-                mpl::log(mpl::Level::warning, category,
-                         fmt::format("Failed to install 'sshfs', error message: '{}'", mp::utils::trim_end(error_msg)));
-            }
-            else
-            {
-                break;
-            }
-        }
-        catch (const mp::ExitlessSSHProcessException&)
-        {
-            mpl::log(mpl::Level::info, category, fmt::format("Timeout while installing 'sshfs' in '{}'", name));
-        }
+        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed for \'{}\'", name));
+        throw std::runtime_error(
+            fmt::format("Snap support needs to be installed for \'{}\' in order to support mounts. "
+                        "Please see https://docs.snapcraft.io/installing-snapd for information on "
+                        "how to install Snap support for your instance's distribution.",
+                        name));
     }
 
-    if (retries > max_install_sshfs_retries)
-        throw mp::SSHFSMissingError();
+    try
+    {
+        auto proc = session.exec("sudo snap install multipass-sshfs");
+        if (proc.exit_code(std::chrono::minutes(5)) != 0)
+        {
+            auto error_msg = proc.read_std_error();
+            mpl::log(mpl::Level::warning, category,
+                     fmt::format("Failed to install \'multipass-sshfs\', error message: \'{}\'",
+                                 mp::utils::trim_end(error_msg)));
+            throw mp::SSHFSMissingError();
+        }
+    }
+    catch (const mp::ExitlessSSHProcessException&)
+    {
+        mpl::log(mpl::Level::info, category, fmt::format("Timeout while installing 'sshfs' in '{}'", name));
+    }
 }
 
 QFutureWatcher<mp::Daemon::AsyncOperationStatus>*
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1248,6 +1248,7 @@ try // clang-format on
         }
 
         auto& vm = it->second;
+        auto& vm_specs = vm_instance_specs[name];
 
         if (vm->current_state() == mp::VirtualMachine::State::running)
         {
@@ -1266,7 +1267,10 @@ try // clang-format on
                     MountReply mount_reply;
                     mount_reply.set_mount_message("Enabling support for mounting");
                     server->Write(mount_reply);
-                    install_sshfs(vm.get(), name);
+
+                    mp::SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm_specs.ssh_username,
+                                           *config->ssh_key_provider};
+                    mp::utils::install_sshfs_for(name, session);
                     instance_mounts.start_mount(vm.get(), request->source_path(), target_path, gid_map, uid_map);
                 }
                 catch (const mp::SSHFSMissingError&)
@@ -1281,7 +1285,6 @@ try // clang-format on
             }
         }
 
-        auto& vm_specs = vm_instance_specs[name];
         if (vm_specs.mounts.find(target_path) != vm_specs.mounts.end())
         {
             fmt::format_to(errors, "There is already a mount defined for \"{}:{}\"\n", name, target_path);
@@ -2054,57 +2057,6 @@ grpc::Status mp::Daemon::cmd_vms(const std::vector<std::string>& tgts, std::func
     return grpc::Status::OK;
 }
 
-void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
-{
-    auto& key_provider = *config->ssh_key_provider;
-
-    SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
-
-    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap in \'{}\'", name));
-
-    // Check if snap support is installed in the instance
-    auto which_proc = session.exec("which snap");
-    if (which_proc.exit_code() != 0)
-    {
-        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in \'{}\'", name));
-        throw std::runtime_error(
-            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts."
-                        "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
-                        "\nhow to install snap support for your instance's distribution."
-                        "\n\nAlternatively, install `sshfs` manually inside the instance.",
-                        name));
-    }
-
-    // Check if /snap exists for "classic" snap support
-    auto test_file_proc = session.exec("[ -e /snap ]");
-    if (test_file_proc.exit_code() != 0)
-    {
-        mpl::log(mpl::Level::warning, category, fmt::format("Classic snap support symlink is needed in \'{}\'", name));
-        throw std::runtime_error(
-            fmt::format("Classic snap support is needed in \'{}\' in order to support mounts."
-                        "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
-                        "\nhow to enable classic snap support for your instance's distribution.",
-                        name));
-    }
-
-    try
-    {
-        auto proc = session.exec("sudo snap install multipass-sshfs");
-        if (proc.exit_code(std::chrono::minutes(5)) != 0)
-        {
-            auto error_msg = proc.read_std_error();
-            mpl::log(mpl::Level::warning, category,
-                     fmt::format("Failed to install \'multipass-sshfs\', error message: \'{}\'",
-                                 mp::utils::trim_end(error_msg)));
-            throw mp::SSHFSMissingError();
-        }
-    }
-    catch (const mp::ExitlessSSHProcessException&)
-    {
-        mpl::log(mpl::Level::info, category, fmt::format("Timeout while installing 'sshfs' in '{}'", name));
-    }
-}
-
 QFutureWatcher<mp::Daemon::AsyncOperationStatus>*
 mp::Daemon::create_future_watcher(std::function<void()> const& finished_op)
 {
@@ -2145,6 +2097,7 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
 
         std::vector<std::string> invalid_mounts;
         auto& mounts = vm_instance_specs[name].mounts;
+        auto& vm_specs = vm_instance_specs[name];
         for (const auto& mount_entry : mounts)
         {
             auto& target_path = mount_entry.first;
@@ -2167,7 +2120,9 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
                         server->Write(reply);
                     }
 
-                    install_sshfs(vm.get(), name);
+                    mp::SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm_specs.ssh_username,
+                                           *config->ssh_key_provider};
+                    mp::utils::install_sshfs_for(name, session);
                     instance_mounts.start_mount(vm.get(), source_path, target_path, gid_map, uid_map);
                 }
                 catch (const mp::SSHFSMissingError&)
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -37,9 +37,6 @@ namespace mpl = multipass::logging;
 namespace
 {
 constexpr auto category = "sftp server";
-constexpr auto multipass_sshfs_bin_path = "/snap/multipass-sshfs/current/bin";
-constexpr auto multipass_sshfs_lib_path = "/snap/multipass-sshfs/current/lib";
-
 using SftpHandleUPtr = std::unique_ptr<ssh_string_struct, void (*)(ssh_string)>;
 using namespace std::literals::chrono_literals;
 
@@ -245,26 +242,10 @@ void check_sshfs_status(mp::SSHSession& session, mp::SSHProcess& sshfs_process)
     }
 }
 
-auto create_sshfs_process(mp::SSHSession& session, mp::SshfsMount::SshfsPkgType sshfs_pkg_type,
-                          const std::string& source, const std::string& target)
+auto create_sshfs_process(mp::SSHSession& session, const std::string& sshfs_exec_line, const std::string& source,
+                          const std::string& target)
 {
-    std::string options{"-o slave -o transform_symlinks -o allow_other"};
-    std::string sshfs_exec{"sshfs"};
-
-    if (sshfs_pkg_type == mp::SshfsMount::SshfsPkgType::debian)
-    {
-        // FIXME: Need query the libfuse version to see if `nonempty` is valid.
-        // The option was removed in libfuse 3.0
-        options += " -o nonempty";
-    }
-    else
-    {
-        // TODO: Need to see if there are any new options that could be a benefit
-
-        sshfs_exec = fmt::format("LD_LIBRARY_PATH={} {}/sshfs", multipass_sshfs_lib_path, multipass_sshfs_bin_path);
-    }
-
-    auto sshfs_process = session.exec(fmt::format("sudo {} {} :\"{}\" \"{}\"", sshfs_exec, options, source, target));
+    auto sshfs_process = session.exec(fmt::format("sudo {} :\"{}\" \"{}\"", sshfs_exec_line, source, target));
 
     check_sshfs_status(session, sshfs_process);
 
@@ -274,9 +255,9 @@ auto create_sshfs_process(mp::SSHSession& session, mp::SshfsMount::SshfsPkgType 
 
 mp::SftpServer::SftpServer(SSHSession&& session, const std::string& source, const std::string& target,
                            const std::unordered_map<int, int>& gid_map, const std::unordered_map<int, int>& uid_map,
-                           int default_uid, int default_gid, const mp::SshfsMount::SshfsPkgType sshfs_pkg_type)
+                           int default_uid, int default_gid, const std::string& sshfs_exec_line)
     : ssh_session{std::move(session)},
-      sshfs_process{create_sshfs_process(ssh_session, sshfs_pkg_type, mp::utils::escape_char(source, '"'),
+      sshfs_process{create_sshfs_process(ssh_session, sshfs_exec_line, mp::utils::escape_char(source, '"'),
                                          mp::utils::escape_char(target, '"'))},
       sftp_server_session{make_sftp_session(ssh_session, sshfs_process->release_channel())},
       source_path{source},
@@ -285,7 +266,7 @@ mp::SftpServer::SftpServer(SSHSession&& session, const std::string& source, cons
       uid_map{uid_map},
       default_uid{default_uid},
       default_gid{default_gid},
-      sshfs_pkg_type{sshfs_pkg_type}
+      sshfs_exec_line{sshfs_exec_line}
 {
 }
 
@@ -455,7 +436,7 @@ void mp::SftpServer::run()
                 }
 
                 sshfs_process =
-                    create_sshfs_process(ssh_session, sshfs_pkg_type, mp::utils::escape_char(source_path, '"'),
+                    create_sshfs_process(ssh_session, sshfs_exec_line, mp::utils::escape_char(source_path, '"'),
                                          mp::utils::escape_char(target_path, '"'));
                 sftp_server_session = make_sftp_session(ssh_session, sshfs_process->release_channel());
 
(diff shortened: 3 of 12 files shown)


hellsworth:
I setup this multipass for testing (mostly hoping it gets around #1322 ). After doing `snap refresh multipass --channel edge/pr1328`, I 

Right off the bat, I notice a couple of things:

1. I see that every so often when I"m building something with it, I see this message logged to stdout:
```
mesg: ttyname failed: Inappropriate ioctl for device
```

2. If I do `snapcraft clean` in ProjectB while ProjectA is building, then start building Project B (while Project A is still building), it generates this journalctl log immediately regarding ProjectB:
```
Feb 06 15:49:11 bmo multipassd[10326]: QProcess: Destroyed while process ("/snap/multipass/1673/bin/sshfs_server") is still running.              
Feb 06 15:49:11 bmo multipassd[10326]: There was an error with sshfs_server for instance "snapcraft-drawing" with path '/root/project': Crashed - 
program: /snap/multipass/1673/bin/sshfs_server; error: Process crashed                                                                            
Feb 06 15:49:11 bmo multipassd[10326]: Mount '/root/project' in instance "snapcraft-drawing" has stopped unexpectedly: program: /snap/multipass/16
73/bin/sshfs_server; error: Process crashed
```
Then eventually ProjectA stops building because it hits #1322 
Project B keeps building but does eventually fail (because of an error in the snapcraft.yaml).

Here is the complete journalctl snippet from the time I started building ProjectA.. did snapcraft clean on ProjectB.. started building ProjectB and ProjectA stopped building.

[pr1328-journalctl.txt](https://github.com/canonical/multipass/files/4168160/pr1328-journalctl.txt)

ricab:
Hi, just want to point out that the `ttyname` part is reported in #1310 and unrelated to the rest &ndash; just a msg that we failed to return before, caused by snapcraft using login shells.

src/daemon/daemon.cpp
```diff
@@ -2053,33 +2052,36 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
...
+    try
+    {
+        auto proc = session.exec("sudo snap install multipass-sshfs");
+        if (proc.exit_code(std::chrono::minutes(5)) != 0)
         {
-            mpl::log(mpl::Level::info, category, fmt::format("Timeout while installing 'sshfs' in '{}'", name));
+            auto error_msg = proc.read_std_error();
+            mpl::log(mpl::Level::warning, category,
+                     fmt::format("Failed to install \'multipass-sshfs\', error message: \'{}\'",
+                                 mp::utils::trim_end(error_msg)));
```
Saviq: This doesn't go into the error the user sees, wonder if it should? OTOH `-v` makes it show…

src/sshfs_mount/sftp_server.cpp
```diff
@@ -37,6 +37,9 @@ namespace mpl = multipass::logging;
 namespace
 {
 constexpr auto category = "sftp server";
+constexpr auto multipass_sshfs_bin_path = "/snap/multipass-sshfs/current/bin";
+constexpr auto multipass_sshfs_lib_path = "/snap/multipass-sshfs/current/lib";
```
Saviq: It irks me that this is hardcoded. Wonder if `snap run --shell multipass-sshfs.sshfs` would allow us to keep this dynamic?

`/snap` is unfortunately not uniform across distros.

src/sshfs_mount/sftp_server.cpp
```diff
@@ -242,10 +245,26 @@ void check_sshfs_status(mp::SSHSession& session, mp::SSHProcess& sshfs_process)
...
+    {
+        // FIXME: Need query the libfuse version to see if `nonempty` is valid.
+        // The option was removed in libfuse 3.0
+        options += " -o nonempty";
+    }
+    else
+    {
+        // TODO: Need to see if there are any new options that could be a benefit
+
+        sshfs_exec = fmt::format("LD_LIBRARY_PATH={} {}/sshfs", multipass_sshfs_lib_path, multipass_sshfs_bin_path);
```
Saviq: Could this go into a wrapper inside the snap somehow?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -242,10 +245,26 @@ void check_sshfs_status(mp::SSHSession& session, mp::SSHProcess& sshfs_process)
...
-    auto sshfs_process = session.exec(fmt::format(
-        "sudo sshfs -o slave -o nonempty -o transform_symlinks -o allow_other :\"{}\" \"{}\"", source, target));
+    std::string options{"-o slave -o transform_symlinks -o allow_other"};
+    std::string sshfs_exec{"sshfs"};
+
+    if (sshfs_pkg_type == mp::SshfsMount::SshfsPkgType::debian)
+    {
+        // FIXME: Need query the libfuse version to see if `nonempty` is valid.
+        // The option was removed in libfuse 3.0
+        options += " -o nonempty";
```
Saviq: What happens if it's removed? Could we try, fail, and try again without that option instead?

include/multipass/sshfs_mount/sshfs_mount.h
```diff
@@ -29,6 +29,12 @@ class SftpServer;
 class SshfsMount
 {
 public:
+    enum class SshfsPkgType
+    {
+        debian,
+        snap
+    };
+
```
Saviq: I would quite like to not have this enum…

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -33,30 +33,40 @@ namespace mpl = multipass::logging;
...
+    {
+        mpl::log(mpl::Level::debug, category, fmt::format("'sshfs' snap package is not installed: {}", e.what()));
+    }
+
+    try
+    {
+        // Fallback to looking for Debian version if snap is not found
+        run_cmd(session, "which sshfs");
+        return mp::SshfsMount::SshfsPkgType::debian;
+    }
```
Saviq: If anything, this should be more generic than "debian", since this may be any other distro.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -33,30 +33,40 @@ namespace mpl = multipass::logging;
...
 
-auto run_cmd(mp::SSHSession& session, std::string&& cmd)
-{
-    auto error_handler = [](mp::SSHProcess& proc) { throw std::runtime_error(proc.read_std_error()); };
-    return run_cmd(session, std::forward<std::string>(cmd), error_handler);
+    return ssh_process.read_std_output();
 }
 
-void check_sshfs_exists(mp::SSHSession& session)
+mp::SshfsMount::SshfsPkgType get_sshfs_package_type(mp::SSHSession& session)
```
Saviq: Could we do away with this, and just try the location for `multipass-sshfs`, then if that fails, `which sshfs`?

src/sshfs_mount/sftp_server.cpp
townsend2010: Ah, ok, didn't realize that `/snap` is not consistent.  Sure, I'll find a more general solution such as what you suggest.

src/daemon/daemon.cpp
```diff
@@ -2053,33 +2052,36 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
...
-            {
-                break;
-            }
-        }
-        catch (const mp::ExitlessSSHProcessException&)
+        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed for \'{}\'", name));
+        throw std::runtime_error(
+            fmt::format("Snap support needs to be installed for \'{}\' in order to support mounts. "
+                        "Please see https://docs.snapcraft.io/installing-snapd for information on "
+                        "how to install Snap support for your instance's distribution.",
```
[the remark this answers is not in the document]
townsend2010: Did you mean to have so many newlines, particularly between the first and second sentence?

src/sshfs_mount/sftp_server.cpp
townsend2010: Good idea, I'll look into that.

src/sshfs_mount/sftp_server.cpp
townsend2010: If what is removed?  The option?  The `multipass-sshfs` snap?  The packaged version if `libfuse`?

include/multipass/sshfs_mount/sshfs_mount.h
townsend2010: Yeah, me neither :grin:  I had this mainly to quirk the options based on what type of `libfuse` package is installed.  Maybe probing `libfuse` for the version or something would be better.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Well, it's just an internal naming implementation, so I don't think the name of the enum value is important here.  That said, I'd rather just get rid of it altogether as mentioned above.

src/daemon/daemon.cpp
townsend2010: Hmm, I suppose it should be surfaced to the client to give a clue to the user why it didn't install.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Ok, I'll see what I can do.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
Saviq: Yeah I did. It looked a bit squashed without them.

src/sshfs_mount/sftp_server.cpp
Saviq: The option removed :)

Oh so now we need `libfuse` on the system, too?

src/sshfs_mount/sftp_server.cpp
Saviq: I had an idea: just have an `env` app inside `multipass-sshfs`, and `snap run multipass-sshfs.env` to get the right environment from the snap?

src/sshfs_mount/sftp_server.cpp
townsend2010: Upon investigation, it's not trivial to use a wrapper in a snap using the `bare` base, ie, a static package.  `/bin/sh` is not available without having to do tricks.  That said, I have extracted the `LD_LIBRARY_PATH` out of the environment and just add that in the exec line.

include/multipass/sshfs_mount/sshfs_mount.h
townsend2010: I got rid of this.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: It's gone, so no longer an issue.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: I redid this into a new function that detects which `sshfs` to use and adds the command options based on the environment and the `libfuse` version.

src/sshfs_mount/sftp_server.cpp
townsend2010: Yes, `sshfs` requires `libfuse`, but I redid how to detect if the option is needed or not.

src/sshfs_mount/sftp_server.cpp
townsend2010: These are gone now :grin:

src/daemon/daemon.cpp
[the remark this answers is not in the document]
townsend2010: Ok, modified to use your suggestion.

townsend2010:
@Saviq, I've modified this quite a bit based on your review suggestions.  In order for this to work, CanonicalLtd/multipass-sshfs#4 needs to be merged and the new snap built and promoted to stable.

src/daemon/daemon.cpp
townsend2010: I've left this alone for now.  If you think it's necessary to surface to the client without `-v`, then I'll look into it.

include/multipass/sshfs_mount/sshfs_mount.h
Saviq: Did you though?

src/daemon/daemon.cpp
```diff
@@ -2053,33 +2053,37 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
 
     SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), key_provider};
 
-    mpl::log(mpl::Level::info, category, fmt::format("Installing sshfs in \'{}\'", name));
+    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs Snap for \'{}\'", name));
```
Saviq:
```suggestion
    mpl::log(mpl::Level::info, category, fmt::format("Installing the multipass-sshfs snap for \'{}\'", name));
```
Following https://snapcraft.io/docs at least…

src/daemon/daemon.cpp
```diff
@@ -2053,33 +2053,37 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
...
-                mpl::log(mpl::Level::warning, category,
-                         fmt::format("Failed to install 'sshfs', error message: '{}'", mp::utils::trim_end(error_msg)));
-            }
-            else
-            {
-                break;
-            }
-        }
-        catch (const mp::ExitlessSSHProcessException&)
+        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed for \'{}\'", name));
```
Saviq:
```suggestion
        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in \'{}\'", name));
```

src/daemon/daemon.cpp
```diff
@@ -2053,33 +2053,37 @@ void mp::Daemon::install_sshfs(VirtualMachine* vm, const std::string& name)
...
+        throw std::runtime_error(
+            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts.\n\n"
+                        "\n\nPlease see https://docs.snapcraft.io/installing-snapd for information on"
+                        "\nhow to install Snap support for your instance's distribution."
```
Saviq:
```suggestion
                        "\nhow to install snap support for your instance's distribution."
```
This one's on me ;P

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -26,37 +26,93 @@
...
+    std::string line;
+    while (std::getline(output, line, '\n'))
+    {
+        if (line.find(matcher) != std::string::npos)
+        {
+            return line;
+        }
+    }
+
+    return std::string{};
```
Saviq: MEH! ;P

Should this go into utils?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -26,37 +26,93 @@
...
+    try
+    {
+        // Prefer to use Snap package version first
+        std::istringstream sshfs_env{run_cmd(session, "sudo multipass-sshfs.env")};
+
+        auto ld_library_path = match_line(sshfs_env, ld_library_path_key);
+        auto snap_path = match_line(sshfs_env, snap_path_key);
+        snap_path = snap_path.substr(snap_path_key.length());
+
+        sshfs_exec = fmt::format("{} {}/bin/sshfs", ld_library_path, snap_path);
```
Saviq: Maybe have `env` at the beginning so we don't need a shell to set the variable up?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -26,37 +26,93 @@
...
+    }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::debug, category, fmt::format("'sshfs' snap package is not installed: {}", e.what()));
```
Saviq:
```suggestion
        mpl::log(mpl::Level::debug, category, fmt::format("'multipass-sshfs' snap package is not installed: {}", e.what()));
```

Or just drop `'sshfs' `, we know where this is.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -26,37 +26,93 @@
...
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
 namespace
 {
 constexpr auto category = "sshfs mount";
-template <typename Callable>
-auto run_cmd(mp::SSHSession& session, std::string&& cmd, Callable&& error_handler)
+const std::string fuse_version_string{"FUSE library version "};
```
Saviq: In a current multipass instance:
```
$ sshfs -V
...
FUSE library version: 2.9.7
$
```
The colon expected?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -26,37 +26,93 @@
...
+    auto fuse_version_line = match_line(version_info, fuse_version_string);
+    if (!fuse_version_line.empty())
+    {
+        fuse_version = fuse_version_line.substr(fuse_version_string.length());
+    }
+
+    sshfs_exec += " -o slave -o transform_symlinks -o allow_other";
+
+    // The option was removed in libfuse 3.0
+    if (fuse_version < "3")
```
Saviq: Should we `atoi` this?

src/sshfs_mount/sshfs_mount.cpp
townsend2010: D'oh, silly fragile output strings...

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Yeah, that's probably a better place for this.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Well, `atoi` would probably work in this case, but I do see a need that we may need to break down the version string into tokens and then compare each x, y, and z parts of the version.

For example, a new feature may be added in the minor version of `libfuse` or even `sshfs`.  So I think I should change this into a util to figure out the version we are interested in.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Yeah, good idea.

src/sshfs_mount/sshfs_mount.cpp
Saviq: We already have `semver`?

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Ugh, come to find out, the archive version of `sshfs -V` prints out some of the info on `stdout` as well as `stderr` for whatever reason and we are only reading `stdout` when getting output from a command running through `libssh`.  We are going to have to modify how we are getting output through `libssh` and read both `stdout` and `stderr` for any given command regardless if there is an error or not returned by the command.

townsend2010: And of course `FUSE library version:` is one of those that is written to `stderr` :frowning_face:

townsend2010: Ok, fix seems easy- just concatenate reading `stdout` and `stderr` in `run_cmd()`.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Yep, switched to using `semver`.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Ok, I fixed this by not including the space in `fuse_version_string` along with fixing up `run_cmd()` like mentioned previously.  Lastly, I use a regex to split on the returned string.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Done

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Yep, moved into utils now.  Will write a simple test for it.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -27,36 +27,78 @@
...
+    sshfs_exec += " -o slave -o transform_symlinks -o allow_other";
+
+    auto fuse_version_line = mp::utils::match_line_for(version_info, fuse_version_string);
+    if (!fuse_version_line.empty())
+    {
+        // split on the fuse_version_string along with 0 or more colons
+        auto tokens = mp::utils::split(fuse_version_line, fmt::format("{}:* ", fuse_version_string));
+        auto fuse_version = tokens[1];
+
+        // The option was removed in libfuse 3.0
```
Saviq:
```suggestion
        // The option was made the default in libfuse 3.0
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -27,36 +27,78 @@
...
+
+    std::istringstream version_info{run_cmd(session, fmt::format("sudo {} -V", sshfs_exec))};
+
+    sshfs_exec += " -o slave -o transform_symlinks -o allow_other";
+
+    auto fuse_version_line = mp::utils::match_line_for(version_info, fuse_version_string);
+    if (!fuse_version_line.empty())
+    {
+        // split on the fuse_version_string along with 0 or more colons
+        auto tokens = mp::utils::split(fuse_version_line, fmt::format("{}:* ", fuse_version_string));
```
Saviq: I wonder if we should be strict about it? I.e. match `^{}:? (0-9\.)+$`?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -27,36 +27,78 @@
...
+        auto fuse_version = tokens[1];
+
+        // The option was removed in libfuse 3.0
+        if (!fuse_version.empty() && (version::Semver200_version(fuse_version) < version::Semver200_version("3.0.0")))
+            sshfs_exec += " -o nonempty";
```
Saviq: Error that we failed to parse the version?

townsend2010: Yes, and it's really necessary for the testing.

townsend2010: Oh, you mean throw an error and/or log an error?

Saviq: Yes :)

townsend2010: Hmm, if I throw here, I'll have to modify the tests quite a bit to mock returning some valid version number here.  I can certainly add the logging, but maybe we just leave in the option as default if we can't figure out the version number.  WDYT?

src/sshfs_mount/sshfs_mount.cpp
townsend2010: I don't think it's necessary since I think we can trust the maintainers of FUSE to put in a correct version number here.  But if we do want to be very strict, then I think we need to account for more than a single digit followed by a `.` like `3.12.15` or something.

Also, `semver` should throw if the version is not something it can understand.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Oh, I think `semver` will throw if we try to use an invalid version number.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -27,36 +27,78 @@
...
-        error_handler(ssh_process);
-    return ssh_process.read_std_output();
-}
+        throw std::runtime_error(ssh_process.read_std_error());
 
-auto run_cmd(mp::SSHSession& session, std::string&& cmd)
-{
-    auto error_handler = [](mp::SSHProcess& proc) { throw std::runtime_error(proc.read_std_error()); };
-    return run_cmd(session, std::forward<std::string>(cmd), error_handler);
+    return ssh_process.read_std_output() + ssh_process.read_std_error();
```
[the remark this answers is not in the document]
townsend2010: Yeah, I think we have a few of these `SSHSession::exec` type calls in various places.  Ok, I'll add a `TODO` to remind us that we need to clean that up and have one, probably in `utils`, that can handle this and account for `stdout` and `stderr`.

src/sshfs_mount/sshfs_mount.cpp
Saviq: OK you're right, semver will tell us if it's good. I would still only accept zero-or-one colon:
```suggestion
        auto tokens = mp::utils::split(fuse_version_line, fmt::format("{}:? ", fuse_version_string));
```

src/sshfs_mount/sshfs_mount.cpp
Saviq: Yeah, just logging the error is fine, but: keep the option _in_, our _out_? I think I'm leaning _out_, since it's more compatible, and it's not very dangerous if it's left out?

townsend2010: If we don't specify `-o nonempty` on a directory that isn't empty, then `sshfs` will error out.  Not dangerous by any means.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
townsend2010: Ok, I put a `TODO` at the top of the function.

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Ok, I modified the logic a bit and logging.  I think the option should only be added if we can be sure the version is < 3.0.0, otherwise leave it out.

Saviq:
No worky on CentOS 7?
```
$ multipass launch --disk 10G https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2
$  multipass mount $PWD sensitive-nilgai:Home                                                                                                                                                                                                
mount failed: Snap support needs to be installed in 'sensitive-nilgai' in order to support mounts.                                                                                                                                                                    
                                                                                                                                                                                                                                                                      
Please see https://docs.snapcraft.io/installing-snapd for information on                                                                                                                                                                                              
how to install snap support for your instance's distribution.                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                      
Alternatively, install `sshfs` manually inside the instance.  
# follow installing-snapd for CentOS 7
$  multipass mount $PWD sensitive-nilgai:Home                         
mount failed: Error enabling mount support in 'sensitive-nilgai'                

Please install the 'multipass-sshfs' snap manually inside the instance.
# :(
```

townsend2010:
Ok, on CentOS:
```
$ sudo multipass-sshfs.env 
sudo: multipass-sshfs.env: command not found
```
which is what we are running in the instance to get the environment.  I need to remember why we need the `sudo` there in the first place or remove it if we don't.

townsend2010:
Bah, the environment is not correct on CentOS:
```
$ multipass exec test-centos -- snap run multipass-sshfs.env | grep SNAP=
SNAP=/snap/multipass-sshfs/134
$ multipass exec test-centos -- ls /snap/multipass-sshfs/134
ls: cannot access /snap/multipass-sshfs/134: No such file or directory
```
Back to square one on how to find where the executable is :frowning_face:

ricab:
:confounded: wouldn't that be a snap bug?

townsend2010:
Yes, I believe it's a bug either in snapd itself or how the snapd RPM package is done.  But unfortunately, until we have some sort of solution, we shouldn't merge this :disappointed:

townsend2010:
Ah, it's not a bug really.  In https://snapcraft.io/docs/installing-snap-on-centos, there is a section that says:
```
To enable classic snap support, enter the following to create a symbolic link between /var/lib/snapd/snap and /snap:

$ sudo ln -s /var/lib/snapd/snap /snap
```
So this is expected to have the path like this.  We just need to check if `/snap` exists, and if not, just print a message that they need to enable support for classic snaps per the directions (even though our snap is not classic, but...).

townsend2010:
This all seems to be working as expected now in my testing.

Saviq:
```console
snap refresh multipass --stable
multipass shell
snap refresh multipass --channel edge/pr1328
multipass delete -p --all
multipass shell
# broken "Home" mount
```

```
mar 26 15:23:43 michal-laptop multipassd[1195874]: mounting /home/michal => Home in primary
mar 26 15:23:43 michal-laptop multipassd[1195874]: process program '/snap/multipass/1882/bin/sshfs_server'
mar 26 15:23:43 michal-laptop multipassd[1195874]: process arguments '10.9.198.230, 22, ubuntu, /home/michal, Home, 1000:-1,, 1000:-1,'
mar 26 15:23:43 michal-laptop multipassd[1215259]: Applying AppArmor policy: multipass.primary.fc813973.sshfs_server
mar 26 15:23:44 michal-laptop multipassd[1195874]: SSHFS was not found on the host: 10.9.198.230
mar 26 15:23:44 michal-laptop multipassd[1195874]: Mount 'Home' in instance "primary" has stopped unexpectedly: Process returned exit code: 9
mar 26 15:23:44 michal-laptop multipassd[1195874]: Installing the multipass-sshfs snap in 'primary'
mar 26 15:24:10 michal-laptop multipassd[1195874]: mounting /home/michal => Home in primary
mar 26 15:24:10 michal-laptop multipassd[1195874]: process program '/snap/multipass/1882/bin/sshfs_server'
mar 26 15:24:10 michal-laptop multipassd[1195874]: process arguments '10.9.198.230, 22, ubuntu, /home/michal, Home, 1000:-1,, 1000:-1,'
mar 26 15:24:10 michal-laptop multipassd[1195874]: QMP: {"timestamp": {"seconds": 1585232644, "microseconds": 712820}, "event": "RTC_CHANGE", "data": {"offset": 1}}
mar 26 15:24:10 michal-laptop multipassd[1216602]: Applying AppArmor policy: multipass.primary.fc813973.sshfs_server
mar 26 15:24:13 michal-laptop multipassd[1195874]: QProcess: Destroyed while process ("/snap/multipass/1882/bin/sshfs_server") is still running.
mar 26 15:24:13 michal-laptop multipassd[1195874]: Mount 'Home' in instance "primary" has stopped
mar 26 15:24:13 michal-laptop multipassd[1195874]: Failed to remove AppArmor policy 
                                                   #include <tunables/global>
                                                   profile multipass.primary.fc813973.sshfs_server flags=(attach_disconnected) {
                                                       #include <abstractions/base>
                                                       #include <abstractions/nameservice>
                                                   
                                                       # Sshfs_server requires broad filesystem altering permissions, but only for the
                                                       # host directory the user has specified to be shared with the VM.
                                                   
                                                       # Required for reading and searching host directories
                                                       capability dac_override,
                                                       capability dac_read_search,
                                                       # Enables modifying of file ownership and permissions
                                                       capability chown,
                                                       capability fsetid,
                                                       capability fowner,
                                                       # Multipass allows user to specify arbitrary uid/gid mappings
                                                       capability setuid,
                                                       capability setgid,
                                                   
                                                       # Allow multipassd send sshfs_server signals
                                                       signal (receive) peer=snap.multipass.multipassd,
                                                   
                                                       # sshfs gathers some info about system resources
                                                       /sys/devices/system/node/ r,
                                                       /sys/devices/system/node/node[0-9]*/meminfo r,
                                                   
                                                       # binary and its libs
                                                       /snap/multipass/1882/bin/sshfs_server ixr,
                                                       /snap/multipass/1882/{usr/,}lib/** rm,
                                                   
                                                       # CLASSIC ONLY: need to specify required libs from core snap
                                                       /{,var/lib/snapd/}snap/core18/*/{,usr/}lib/@{multiarch}/{,**/}*.so* rm,
                                                   
                                                       # allow full access just to this user-specified source directory on the host
                                                       /home/michal/ rw,
                                                       /home/michal/** rwlk,
                                                   }
                                                       : errno=254 ()
mar 26 15:24:13 michal-laptop multipassd[1195874]: QProcess: Destroyed while process ("/snap/multipass/1882/bin/sshfs_server") is still running.
mar 26 15:24:13 michal-laptop multipassd[1195874]: There was an error with sshfs_server for instance "primary" with path 'Home': Crashed - program: /snap/multipass/1882/bin/sshfs_server; error: Process crashed
mar 26 15:24:13 michal-laptop multipassd[1195874]: Mount 'Home' in instance "primary" has stopped unexpectedly: program: /snap/multipass/1882/bin/sshfs_server; error: Process crashed
mar 26 15:35:16 michal-laptop multipassd[1195874]: QMP: {"timestamp": {"seconds": 1585233316, "microseconds": 508477}, "event": "RTC_CHANGE", "data": {"offset": 1}}
```

`multipass-sshfs` is installed inside the instance:
```
$ multipass exec primary -- snap list multipass-sshfs
Name             Version  Rev  Tracking  Publisher   Notes
multipass-sshfs  1.1      134  stable    canonical✓  -
```

Another mount worked, though - something's not resilient enough… And I can't reproduce again, of course… Hoping something from the logs above will be informative (potentially for a follow-up PR).

townsend2010:
@Saviq,

I think I know what may have happened and it's not specific to this PR.  When a mount is destroyed such as when deleting an instance, the associated `QObject` is called with `deleteLater()` in the dtor.  So if you delete an instance that has a mount and then fairly quickly create an instance with the same name with the same mount, that `deleteLater()` may be called when the new instance is started and the mount is complete.

src/utils/utils.cpp
```diff
@@ -212,6 +216,55 @@ void mp::utils::wait_for_cloud_init(mp::VirtualMachine* virtual_machine, std::ch
...
+
+    try
+    {
+        auto proc = session.exec("sudo snap install multipass-sshfs");
+        if (proc.exit_code(std::chrono::minutes(5)) != 0)
+        {
+            auto error_msg = proc.read_std_error();
+            mpl::log(mpl::Level::warning, category,
+                     fmt::format("Failed to install \'multipass-sshfs\', error message: \'{}\'",
+                                 mp::utils::trim_end(error_msg)));
```
Saviq: Hmm this didn't bubble up when [the Fedora](https://forum.snapcraft.io/t/installing-snap-on-fedora/6755/17) mount failed…

I suppose that's due to our log forwarding thing not working (#962, but maybe more…).

src/utils/utils.cpp
```diff
@@ -327,6 +380,22 @@ std::string mp::utils::timestamp()
     return time.toString(Qt::ISODateWithMs).toStdString();
 }
 
+std::string mp::utils::match_line_for(std::istringstream& output, const std::string& matcher)
+{
+    output.seekg(0, output.beg);
```
Saviq: Feels weird to be rewinding the stream here… Should we make a copy of the stream instead? Or require it to be moved, maybe?

tests/test_sshfsmount.cpp
```diff
@@ -53,14 +53,17 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+        auto request_exec = [this, expected_cmds, &invoked](ssh_channel, const char* raw_cmd) {
             std::string cmd{raw_cmd};
-            if (cmd.find(expected_cmd) != std::string::npos)
+            for (const auto expected_cmd : expected_cmds)
             {
-                invoked = true;
-                exit_status_mock.return_exit_code(SSH_ERROR);
+                if (cmd.find(expected_cmd) != std::string::npos)
+                {
+                    invoked = true;
```
Saviq: This irks me b/c it gets set to `true` if _any_ of the commands are invoked… But probably not something for this PR.

I think `test_command_execution` should be made into something that can also trigger `SSH_ERROR`, at which point all this can be made into a parameterized test.

tests/test_sshfsmount.cpp
```diff
@@ -73,11 +76,20 @@ struct SshfsMount : public mp::test::SftpServerTest
...
 
-            if (next_expected_cmd != commands.end())
+            auto it = default_cmds.find(cmd);
+            if (it != default_cmds.end())
+            {
+                output = it->second;
+                remaining = output.size();
+                invoked = true;
+            }
+            else if (next_expected_cmd != commands.end())
```
Saviq: This should be fall-through, not the first case. Otherwise you can't test the case when one of the default commands returns something different than the default.

tests/test_sshfsmount.cpp
```diff
@@ -73,11 +76,20 @@ struct SshfsMount : public mp::test::SftpServerTest
                                      CommandVector::const_iterator& next_expected_cmd, std::string& output,
                                      bool& invoked)
     {
-        auto request_exec = [&commands, &remaining, &next_expected_cmd, &output, &invoked](ssh_channel,
-                                                                                           const char* raw_cmd) {
+        auto request_exec = [this, &commands, &remaining, &next_expected_cmd, &output, &invoked](ssh_channel,
```
Saviq: Why `this`?

tests/test_sshfsmount.cpp
```diff
@@ -369,11 +413,50 @@ TEST_F(SshfsMount, works_with_absolute_paths)
...
+    EXPECT_THROW(mp::utils::install_sshfs_for("foo", session), mp::SSHFSMissingError);
+    EXPECT_TRUE(invoked);
+}
+
+TEST_F(SshfsMount, install_sshfs_no_failures_does_not_throw)
+{
+    mp::SSHSession session{"a", 42};
+
+    EXPECT_NO_THROW(mp::utils::install_sshfs_for("foo", session));
+}
```
Saviq: This should probably be a bit more verbose, checking that the expected commands are executed (using `test_command_execution`).

Saviq: Also, missing case for the timeout, if possible to test?

Saviq: And a couple more untested cases:

https://codecov.io/gh/canonical/multipass/compare/e4cab2e0ade0ebdb04e51d770da6a7fed224bce5...c5c9f0c2df427023c788eabc16f2ebbe534b74c2/diff#D9-101

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -28,38 +28,90 @@
...
+    auto fuse_version_line = mp::utils::match_line_for(version_info, fuse_version_string);
+    if (!fuse_version_line.empty())
+    {
+        // split on the fuse_version_string along with 0 or 1 colon(s)
+        auto tokens = mp::utils::split(fuse_version_line, fmt::format("{}:? ", fuse_version_string));
+        auto fuse_version = tokens[1];
+
+        if (fuse_version.empty())
+        {
+            mpl::log(mpl::Level::warning, category, fmt::format("Unable to parse the {}", fuse_version_string));
```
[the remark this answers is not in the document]
townsend2010: It will resolve to `Unable to parse the FUSE library version`.  I'm not sure what's wrong with that.  Can you please explain what you mean?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -28,38 +28,90 @@
...
+        sshfs_exec = fmt::format("env {} {}/bin/sshfs", ld_library_path, snap_path);
+    }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::debug, category, fmt::format("'multipass-sshfs' snap package is not installed: {}", e.what()));
+
+        // Fallback to looking for distro version if snap is not found
+        try
+        {
+            sshfs_exec = run_cmd(session, "sudo which sshfs");
```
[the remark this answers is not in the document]
townsend2010: We do run `sshfs` using `sudo`.

src/utils/utils.cpp
```diff
@@ -212,6 +216,55 @@ void mp::utils::wait_for_cloud_init(mp::VirtualMachine* virtual_machine, std::ch
...
+                        "Please see https://docs.snapcraft.io/installing-snapd for information on\n"
+                        "how to install snap support for your instance's distribution.\n\n"
+                        "If your distribution's instructions specify enabling classic snap support,\n"
+                        "please do that as well.\n\n"
+                        "Alternatively, install `sshfs` manually inside the instance.",
+                        name));
+    }
+
+    // Check if /snap exists for "classic" snap support
+    auto test_file_proc = session.exec("[ -e /snap ]");
```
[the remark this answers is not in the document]
townsend2010: Certainly more than one way to get the same result.  Is it really worth changing?  We use this type of checking in other parts of the code too, so I have this to remain consistent.

src/utils/utils.cpp
townsend2010: I did this because of https://github.com/canonical/multipass/blob/c5c9f0c2df427023c788eabc16f2ebbe534b74c2/src/sshfs_mount/sshfs_mount.cpp#L61-L64

But yeah, after you mentioning it, I don't think `match_line_for()` should have a rewind.  I'll look into a better way to handle this.

tests/test_sshfsmount.cpp
townsend2010: Because `default_cmds` is a member of `struct SshfsMount`, so we have to pass the object into the lambda.

tests/test_sshfsmount.cpp
townsend2010: Ack

tests/test_sshfsmount.cpp
townsend2010: Sorry, I'm not really following what you mean here.

tests/test_sshfsmount.cpp
townsend2010: Yeah, I think between this PR and recent merges (and probably upcoming other work), these tests will need refactoring into something better.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
Saviq: Oh, thought it would actually be the version string as returned from the above… We should debug-log the output we failed to parse, then.

tests/test_sshfsmount.cpp
townsend2010: For the timeout, I can add a timeout parameter to the function with a default of 5 minutes that we can override in the test.

tests/test_sshfsmount.cpp
Saviq: The defaults should be evaluated last, not first. Otherwise you can't have a `id -u` that returns "1001" in your `commands`.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
townsend2010: Ok, I'll add another logging message of the actual output.

tests/test_sshfsmount.cpp
Saviq: There's no way to mock the timeout passing regardless of the actual timeout? I'd rather avoid an actual timeout expiring in the tests :)

tests/test_sshfsmount.cpp
townsend2010: Ah, right, ok, I follow now :smiley:

townsend2010: Ah, ok, I follow you now :smiley:

tests/test_sshfsmount.cpp
townsend2010: No, we have no mocks in place for timeouts of any sort.  But like we do in other tests, we can put in a very short timeout and then wait for that.

src/utils/utils.cpp
townsend2010: Right, log messages...it's log level `warning` for this too, so I think that would require at least 1 `-v` to be specified for it to show in the client if log forwarding worked.

tests/test_sshfsmount.cpp
townsend2010: @Saviq,

Looking at this particular test some more (`install_sshfs_no_failures_does_not_throw`), we can't easily use `test_command_execution` since all that is being tested here is `mp::utils::install_sshfs_for()`.  `test_command_execution` creates a whole mock `SshfsServer` whereas this test is not.  Unless I refactor `test_command_execution` in this PR, I think it should wait and this test is the best we can do.

I will still try to address the timeout test along with other couple of missing coverage pieces.

tests/test_sshfsmount.cpp
townsend2010: Fixed

src/utils/utils.cpp
townsend2010: Fixed

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
townsend2010: Done

townsend2010:
Um, what happened to travis ci?????

tests/test_sshfsmount.cpp
townsend2010: I still have the timeout test to try to address.

Saviq:
Hmm unsure why this conflict:
```
CONFLICT (content): Merge conflict in include/multipass/utils.h
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -28,38 +28,95 @@
...
+
+        // split on the fuse_version_string along with 0 or 1 colon(s)
+        auto tokens = mp::utils::split(fuse_version_line, fmt::format("{}:? ", fuse_version_string));
+        if (tokens.size() == 2)
+            fuse_version = tokens[1];
+
+        if (fuse_version.empty())
+        {
+            mpl::log(mpl::Level::warning, category, fmt::format("Unable to parse the {}", fuse_version_string));
+            mpl::log(mpl::Level::debug, category, fmt::format("Value is {}", fuse_version_line));
```
Saviq:
```suggestion
            mpl::log(mpl::Level::debug, category, fmt::format("Unable to parse the {}: {}", fuse_version_string, fuse_version_line));
```

In case you filter for just debug messages, or something else logs in between (I know, unlikely)?

tests/test_sshfsmount.cpp
```diff
@@ -154,13 +189,23 @@ struct SshfsMount : public mp::test::SftpServerTest
     std::string default_target{"target"};
     std::unordered_map<int, int> default_map;
     int default_id{1000};
+    std::shared_ptr<NiceMock<mpt::MockLogger>> logger = std::make_shared<NiceMock<mpt::MockLogger>>();
+
+    const std::unordered_map<std::string, std::string> default_cmds{
```
Saviq: Add `pwd` → "/home/ubuntu" here, too?

tests/test_sshfsmount.cpp
```diff
@@ -350,17 +429,11 @@ TEST_F(SshfsMount, throws_when_unable_to_get_current_dir)
 TEST_F(SshfsMount, executes_commands)
 {
     CommandVector commands = {
-        {"which sshfs", "/usr/bin/sshfs"},
         {"pwd", "/home/ubuntu"},
```
Saviq:
```suggestion
```

src/sshfs_mount/sshfs_mount.cpp
townsend2010: Ok, sure.

tests/test_sshfsmount.cpp
townsend2010: Ok...I *thought* we could make it a default command, but it was added in #1359, so I left it as a tested command.  But since you mentioned it, then we'll do it :smile:

tests/test_sshfsmount.cpp
townsend2010: Done

townsend2010:
Ok, this should be ready for another review pass.  I should have addressed the latest round of comments one way or another.

Saviq:
Yes. Yes.

bors r+

PR #1331, opened 2020-01-31
Use std optional directly

ricab:


Saviq:
Because of #1392 we can't merge this today. Probably best to close?

ricab:
Right

PR #1332, opened 2020-01-31
Throwaway hunt optional deref

ricab:
Something like this should help us detect unprotected optional dereferencing cases. 

The build stops fast (even with `make --keep-going`), so the process would be to replace such cases with `value()` iteratively, noting down the ones we'd actually want to change.

### DIFF
--- a/include/multipass/optional.h
+++ b/include/multipass/optional.h
@@ -41,6 +41,14 @@ public:
     }
     using std::optional<T>::operator=;
     using std::optional<T>::optional;
+
+private:
+    constexpr const T* operator->() const;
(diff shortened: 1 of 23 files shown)


ricab:
The build and tests now pass using only checked optional referencing everywhere. Changes in the last commit basically signal all places where unchecked optional dereferencing was being used. Only a few of those are meant to go to master, I will make a new PR for that. One place had a clear undefined behavior. 

I push everything here for reference, so that changes that get picked can be compared against this.

ricab:
This has served its purpose: #1340

PR #1353, opened 2020-02-13
output: Use qUtf8Printable() instead of qPrintable()

townsend2010:
This is supposed to be used for logging and will also help with proper UTF8 output
on Windows.

---

This PR in conjunction with canonical/multipass-private#216 should make unicode characters in our help output properly.

Also note that this is Qt's recommendation: https://doc.qt.io/qt-5/qtglobal.html#qPrintable
> ...qUtf8Printable() should be used for logging strings instead of qPrintable().

### DIFF
--- a/src/client/cli/argparser.cpp
+++ b/src/client/cli/argparser.cpp
@@ -112,9 +112,9 @@ mp::ParseCode mp::ArgParser::parse()
     {
         if (!parser_result)
         {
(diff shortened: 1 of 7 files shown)


Saviq:
Works!

bors r+

Saviq:
CI breakdown…

bors r+

PR #1354, opened 2020-02-13
unix_console: Better handle SIGWINCH

### LINKED ISSUE #1327: stty size not updated on macOS
When resizing a terminal window (both Terminal.app and iTerm2), the value returned by `stty size` is not updated w.r.t. the initial value when it is run from a `multipass shell`.

townsend2010:
On OSX, SIGWINCH is ignored by default and thus, the way the console code was waiting on the signal,
it was never being delivered and the console size would never change.

Since OSX and Linux handle the console code the same, this change is for both platforms.

Fixes #1327

---

On Linux, `sigwait()` will not ignore any signals set in the `sigset` even if any of the signals are ignored by default.  However, OSX will ignore any signals in this case and `SIGWINCH` is ignored by default.

For testing, on Linux, the behavior should not change at all.  This does simplify the code though.  On OSX, you'll notice that the console stty would never change when `shell`ed into an instance, but with this change, that should work now.

### DIFF
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -28,8 +28,9 @@ namespace mp = multipass;
 
 namespace
 {
-const std::vector<int> blocked_sigs{SIGWINCH, SIGUSR1};
 mp::Console::ConsoleGeometry last_geometry{0, 0};
+ssh_channel global_channel;
+int global_cout_fd;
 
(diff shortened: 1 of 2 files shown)


Saviq:
We've no test harness for this yet, do we?

townsend2010:
@Saviq, no, it will require a way to manipulate a console's window size.

Saviq:
Yass!

bors r+

PR #1359, opened 2020-02-14
Correct ownership of mount directories

### LINKED ISSUE #953: multipass-created parent directories for mountpoints are root-owned
```shell
# Create something to mount
$ mkdir example

# Launch a VM and mount our directory
$ multipass launch bionic
Launched: steady-koel
$ multipass mount example steady-koel:/home/multipass/parent/mountpoint

# Examine permissions
$ multipass exec steady-koel -- ls -ld /home/multipass/parent/mountpoint/
drwxr-xr-x 1 multipass multipass 4096 Aug 12 06:40 /home/multipass/parent/mountpoint/
$ multipass exec steady-koel -- ls -ld /home/multipass/parent/
drwxr-xr-x 3 root root 4096 Aug 12 06:36 /home/multipass/parent/
```

**Expected behaviour:** Either multipass should refuse to mount into non-existent paths, or it should create the parent paths with the same permissions as the mountpoint.

**Version**: Reproduced on both beta (1017) and edge (1041).

(Over the weekend, I wrote [a blog post](https://write.wrestle.town/oddbloke/using-multipass-to-create-a-writefreely-dev-vm) guiding people through configuring a dev VM for [WriteFreely](https://writefreely.org/) using multipass. I found a few bugs and/or rough edges that I'm filing now.)

luis4a0:
Fixes #953

When a path to a mount point must be created, all created directories except the last were owned by root.root.

Instead of creating the full path with a mkdir, now we check the existence of each directory in the path, create it and set the correct owner.

### DIFF
--- a/src/sshfs_mount/sshfs_mount.cpp
+++ b/src/sshfs_mount/sshfs_mount.cpp
@@ -25,6 +25,7 @@
 
 #include <multipass/format.h>
 
+#include <QDir>
 #include <iostream>
 
 namespace mp = multipass;
@@ -42,12 +43,14 @@ auto run_cmd(mp::SSHSession& session, std::string&& cmd, Callable&& error_handle
     return ssh_process.read_std_output();
 }
 
+// Run a command on a given SSH session.
 auto run_cmd(mp::SSHSession& session, std::string&& cmd)
 {
     auto error_handler = [](mp::SSHProcess& proc) { throw std::runtime_error(proc.read_std_error()); };
     return run_cmd(session, std::forward<std::string>(cmd), error_handler);
 }
 
+// Check if sshfs exists on a given SSH session.
 void check_sshfs_exists(mp::SSHSession& session)
 {
     auto error_handler = [](mp::SSHProcess& proc) {
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
     run_cmd(session, "which sshfs", error_handler);
 }
 
-void make_target_dir(mp::SSHSession& session, const std::string& target)
+// Split a path into existing and to-be-created parts.
+std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, const std::string& target)
 {
-    run_cmd(session, fmt::format("sudo mkdir -p \"{}\"", target));
+    QDir complete_path(QString::fromStdString(target));
+    QString absolute;
+
+    if (complete_path.isRelative())
+    {
+        QString home = QString::fromStdString(run_cmd(session, "pwd")).trimmed();
+        absolute = home + '/' + complete_path.path();
+    }
+    else
+    {
+        absolute = complete_path.path();
+    }
+
+    QString existing =
+        QString::fromStdString(
+            run_cmd(session,
+                    fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'",
+                                absolute)))
+            .trimmed();
+
+    return {existing.toStdString(), QDir(existing).relativeFilePath(absolute).toStdString()};
 }
 
-void set_owner_for(mp::SSHSession& session, const std::string& target)
+// Create a directory on a given root folder.
+void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
+{
+    if (!relative_target.empty())
+        run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && mkdir -p \"{}\"'", root, relative_target));
+}
+
+// Set ownership of all directories on a path starting on a given root.
+// Assume it is already created.
+void set_owner_for(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
 {
     auto vm_user = run_cmd(session, "id -nu");
     auto vm_group = run_cmd(session, "id -ng");
     mp::utils::trim_end(vm_user);
     mp::utils::trim_end(vm_group);
 
-    run_cmd(session, fmt::format("sudo chown {}:{} \"{}\"", vm_user, vm_group, target));
+    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
+                                 relative_target.substr(0, relative_target.find_first_of('/'))));
 }
 
 auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const std::string& target,
@@ -81,8 +115,14 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
              fmt::format("{}:{} {}(source = {}, target = {}, …): ", __FILE__, __LINE__, __FUNCTION__, source, target));
 
     check_sshfs_exists(session);
-    make_target_dir(session, target);
-    set_owner_for(session, target);
+
+    // Split the path in existing and missing parts.
+    const auto& [leading, missing] = get_path_split(session, target);
+
+    // We need to create the part of the path which does not still exist,
+    // and set then the correct ownership.
+    make_target_dir(session, leading, missing);
+    set_owner_for(session, leading, missing);
 
     auto output = run_cmd(session, "id -u");
     mpl::log(mpl::Level::debug, category,
--- a/tests/test_sshfsmount.cpp
+++ b/tests/test_sshfsmount.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018-2019 Canonical, Ltd.
+ * Copyright (C) 2018-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -19,15 +19,23 @@
 #include "signal.h"
 
 #include <multipass/exceptions/sshfs_missing_error.h>
+#include <multipass/optional.h>
 #include <multipass/ssh/ssh_session.h>
 #include <multipass/sshfs_mount/sshfs_mount.h>
+#include <multipass/utils.h>
 
+#include <algorithm>
 #include <gmock/gmock.h>
+#include <iterator>
+#include <unordered_set>
+#include <vector>
 
 namespace mp = multipass;
 namespace mpt = multipass::test;
 
 using namespace testing;
+
+typedef std::vector<std::pair<std::string, std::string>> CommandVector;
 
 namespace
 {
@@ -39,10 +47,10 @@ struct SshfsMount : public mp::test::SftpServerTest
         channel_is_closed.returnValue(0);
     }
 
-    mp::SshfsMount make_sshfsmount()
+    mp::SshfsMount make_sshfsmount(mp::optional<std::string> target = mp::nullopt)
     {
         mp::SSHSession session{"a", 42};
-        return {std::move(session), default_source, default_target, default_map, default_map};
+        return {std::move(session), default_source, target.value_or(default_target), default_map, default_map};
     }
 
     auto make_exec_that_fails_for(const std::string& expected_cmd, bool& invoked)
@@ -59,10 +67,56 @@ struct SshfsMount : public mp::test::SftpServerTest
         return request_exec;
     }
 
+    // The 'invoked' parameter binds the execution and read mocks. We need a better mechanism to make them
+    // cooperate better, i.e., make the reader read only when a command was issued.
+    auto make_exec_to_check_commands(const CommandVector& commands, std::string::size_type& remaining,
+                                     CommandVector::const_iterator& next_expected_cmd, std::string& output,
+                                     bool& invoked)
+    {
+        auto request_exec = [&commands, &remaining, &next_expected_cmd, &output, &invoked](ssh_channel,
+                                                                                           const char* raw_cmd) {
+            std::string cmd{raw_cmd};
+
+            if (next_expected_cmd != commands.end())
+            {
+                // Check if the first of the remaining commands is being executed.
+                if (cmd == next_expected_cmd->first)
+                {
+                    output = next_expected_cmd->second;
+                    remaining = output.size();
+                    ++next_expected_cmd;
+                    invoked = true;
+                }
+                else
+                {
+                    // If not, then we have to check the rest of the remaining
+                    // commands. If the executed command is there, it means that the
+                    // executed commands do not follow the order of our expected
+                    // command vector.
+                    for (auto it = std::next(next_expected_cmd); it != commands.end(); ++it)
+                    {
+                        if (cmd == it->first)
+                        {
+                            output = it->second;
+                            remaining = output.size();
+                            ADD_FAILURE() << "\"" << (it->first) << "\" executed out of order; expected \""
+                                          << next_expected_cmd->first << "\"";
+                            break;
+                        }
+                    }
+                }
+            }
+
+            return SSH_OK;
+        };
+
+        return request_exec;
+    }
+
     auto make_channel_read_return(const std::string& output, std::string::size_type& remaining, bool& prereq_invoked)
     {
-        auto channel_read = [output, &remaining, &prereq_invoked](ssh_channel, void* dest, uint32_t count,
-                                                                  int is_stderr, int) {
+        auto channel_read = [&output, &remaining, &prereq_invoked](ssh_channel, void* dest, uint32_t count,
+                                                                   int is_stderr, int) {
             if (!prereq_invoked)
                 return 0u;
             const auto num_to_copy = std::min(count, static_cast<uint32_t>(remaining));
@@ -72,6 +126,24 @@ struct SshfsMount : public mp::test::SftpServerTest
             return num_to_copy;
         };
         return channel_read;
+    }
+
+    void test_command_execution(const CommandVector& commands, mp::optional<std::string> target = mp::nullopt)
+    {
+        bool invoked{false};
+        std::string output;
+        auto remaining = output.size();
+        CommandVector::const_iterator next_expected_cmd = commands.begin();
+
+        auto channel_read = make_channel_read_return(output, remaining, invoked);
+        REPLACE(ssh_channel_read_timeout, channel_read);
+
+        auto request_exec = make_exec_to_check_commands(commands, remaining, next_expected_cmd, output, invoked);
+        REPLACE(ssh_channel_request_exec, request_exec);
+
+        make_sshfsmount(target.value_or(default_target));
+
+        EXPECT_TRUE(next_expected_cmd == commands.end()) << "\"" << next_expected_cmd->first << "\" not executed";
     }
 
     mpt::ExitStatusMock exit_status_mock;
@@ -240,4 +312,53 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
 
     mount.join();
     EXPECT_TRUE(stopped_ok);
+}
+
+TEST_F(SshfsMount, throws_when_unable_to_change_dir)
+{
+    bool invoked{false};
+    auto request_exec = make_exec_that_fails_for("cd", invoked);
+    REPLACE(ssh_channel_request_exec, request_exec);
+
+    EXPECT_THROW(make_sshfsmount(), std::runtime_error);
+    EXPECT_TRUE(invoked);
+}
+
+TEST_F(SshfsMount, throws_when_unable_to_get_current_dir)
+{
+    bool invoked{false};
+    auto request_exec = make_exec_that_fails_for("pwd", invoked);
+    REPLACE(ssh_channel_request_exec, request_exec);
+
+    EXPECT_THROW(make_sshfsmount(), std::runtime_error);
+    EXPECT_TRUE(invoked);
+}
+
+TEST_F(SshfsMount, executes_commands)
+{
+    CommandVector commands = {
+        {"which sshfs", "/usr/bin/sshfs"},
+        {"pwd", "/home/ubuntu"},
+        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+         "/home/ubuntu/"},
+        {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", ""},
+        {"id -nu", "ubuntu"},
+        {"id -ng", "ubuntu"},
+        {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R ubuntu:ubuntu target'", ""},
+        {"id -u", "1000"},
+        {"id -g", "1000"},
+        {"sudo sshfs -o slave -o nonempty -o transform_symlinks -o allow_other :\"source\" \"target\"", "don't care"}};
+
+    test_command_execution(commands, std::string("target"));
+}
+
+TEST_F(SshfsMount, works_with_absolute_paths)
+{
+    CommandVector commands = {
+        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+         "/home/ubuntu/"},
+        {"id -u", "1000"},
+        {"id -g", "1000"}};
+
+    test_command_execution(commands, std::string("/home/ubuntu/target"));
 }


src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -64,6 +68,26 @@ void make_target_dir(mp::SSHSession& session, const std::string& target)
...
+{
+    std::string dir;
+
+    for (char c : path)
+    {
+        if (('\\' == c || '/' == c) && !dir.empty())
+        {
+            splitting.push_back(dir);
+        }
+        dir.push_back('\\' == c ? '/' : c);
```
Saviq: Are we ever expecting backslashes in this path?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -64,6 +68,26 @@ void make_target_dir(mp::SSHSession& session, const std::string& target)
...
+    }
+
+    splitting.push_back(dir);
+
+    return;
+}
```
Saviq: Could iterating over `std::filesystem::path` simplify this?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -81,8 +105,19 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
...
+    split_path(target, splitting);
+    bool needs_create = false;
+    for (std::string partial : splitting)
+    {
+        if (needs_create || !std::filesystem::exists(partial))
+        {
+            needs_create = true;
+            make_target_dir(session, partial);
+            set_owner_for(session, partial);
+        }
```
Saviq: We could optimize this by still running `mkdir --parents`, but only with the "remainder" of the path, and then `chown --recursive`, provided both processes' working dir was the subpath that does exist?

We'd then probably have to add an explicit check for file or directory, as changing `CWD` would fail if it was a file.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: To be honest, I don't know. A backslash in the path is maybe a mistake of the user?

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Yes, it can. I'll give it a try.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: If we do that, we'd have to check anyway ownership of each subpath, since we are not changing ownership of existing directories. Am I missing something?

Saviq: If we just do the equivalent of `cd /existing/paths; mkdir --parents remainder/to/be/created; chown … --recursive remainder/to/be/created`, we're good. So rather than `std::filesystem::exists`, use `std::filesystem::is_directory`.

It may also make sense to iterate in reverse.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: OK, so the code is wrong, fixing it.

townsend2010:
Thinking about this some, this is an operation that takes place inside the instance, so I think all of the work should be done in the instance using bash type file operations.  I don't see a need for using `std::filesystem` or any Qt file operation on the host.  Doing this on the host may not match up with what is being mounted in the instance target.

I agree with @Saviq's comment that we should be doing things like `cd /blah`, `mkdir -p`, etc. and use `test -d`, etc. to determine if it's a directory.  I see no reason to do anything on the host itself.

I will also say that I think [this](https://github.com/verybigbadboy/multipass/commit/94a74359b158beda6ea563a840972be6418ebd86) is not correct either :grin:

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Done, implemented as Saviq and ChrisTownsend suggested.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +61,54 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& target)
+{
+    if (!target.empty())
+        run_cmd(session, fmt::format("cd \"{}\" && sudo mkdir -p \"{}\"", root, target));
 }
 
-void set_owner_for(mp::SSHSession& session, const std::string& target)
+// Set ownership of all directories on a path starting on a given root.
+// Assume it is already created.
+void set_owner_for(mp::SSHSession& session, const std::string& root, const std::string& target)
```
Saviq:
```suggestion
void set_owner_for(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +61,54 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+        session, fmt::format("/bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'", path));
+
+    // Strip the trailing newline from the string.
+    existing.pop_back();
+
+    return existing;
+}
+
+// Create a directory on a given root folder.
+void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& target)
```
Saviq:
```suggestion
void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +61,54 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+{
+    if (!target.empty())
+        run_cmd(session, fmt::format("cd \"{}\" && sudo mkdir -p \"{}\"", root, target));
```
Saviq:
```suggestion
        run_cmd(session, fmt::format("cd \"{}\" && sudo mkdir -p \"{}\"", root, relative_target));
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +61,54 @@ void check_sshfs_exists(mp::SSHSession& session)
...
 {
     auto vm_user = run_cmd(session, "id -nu");
     auto vm_group = run_cmd(session, "id -ng");
     mp::utils::trim_end(vm_user);
     mp::utils::trim_end(vm_group);
 
-    run_cmd(session, fmt::format("sudo chown {}:{} \"{}\"", vm_user, vm_group, target));
+    run_cmd(session, fmt::format("/bin/bash -c 'cd \"{}\" && P=\"{}\"; Q=$P; while [ ! -z \"$Q\" ]; do sudo chown "
+                                 "{}:{} $P; Q=`echo $P | cut -f 1 -d / -s`; P=${{P%/*}}; done'",
+                                 root, target, vm_user, vm_group));
```
Saviq: This could then just be:
```cpp
    run_cmd(session, fmt::format("/bin/bash -c 'cd \"{}\" && sudo chown -R {}:{} \"{}\"'", root, vm_user, vm_group, relative_target));
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -81,8 +118,20 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
...
+    // Split the path in existing and missing parts.
+    std::string leading(get_existing_dirs(session, absolute_path));
+    std::string missing = leading.size() >= absolute_path.size()
+                              ? ""
+                              : absolute_path.substr(leading.size(), absolute_path.size() - leading.size());
+
+    // We need to create the part of the path which does not still exist,
+    // and set then the correct ownership.
+    make_target_dir(session, leading, missing);
+    set_owner_for(session, leading, missing);
```
Saviq: All this is then down to:
```cpp
    const auto& [leading, missing] = get_path_split(session, target);

    make_target_dir(session, leading, missing);
    set_owner_for(session, leading, missing);
```

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +61,54 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+std::string get_absolute_path(mp::SSHSession& session, const std::string& path)
+{
+    return *(path.begin()) == '/' ? path : get_home_dir(session) + '/' + path;
+}
+
+// Given a path, returns the part of the path which exists.
+std::string get_existing_dirs(mp::SSHSession& session, const std::string& path)
+{
+    std::string existing = run_cmd(
+        session, fmt::format("/bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'", path));
```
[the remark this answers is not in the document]
townsend2010: Why use `std::filesystem` when we use `QDir`, `QFileInfo`, etc. everywhere else?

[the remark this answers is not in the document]
townsend2010: Oh, I think it just occurred to me.  So `std::filesystem::path` works directly on the path it's given and doesn't do any platform abstraction to translate it to the host filesystem.  So yeah, I think I'm on board with this :grin:

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right, it could be that because we just created the path.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
...
 {
-    run_cmd(session, fmt::format("sudo mkdir -p \"{}\"", target));
+    QDir complete_path(QString::fromStdString(target));
+    std::string absolute;
+
+    if (complete_path.isRelative())
+    {
+        std::string home = run_cmd(session, "pwd");
+        mp::utils::trim_end(home);
+        absolute = home + '/' + complete_path.path().toStdString();
```
Saviq: Wanted to say `QDir::cd` could work for this, but can see now it actually tries to enter the directory…

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
     run_cmd(session, "which sshfs", error_handler);
 }
 
-void make_target_dir(mp::SSHSession& session, const std::string& target)
+// Split a path into existing and to-be-created parts.
+std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, const std::string& target)
```
Saviq: Maybe we could make all these `QString`s so there's not so many `::fromStdString`, `.toStdString()`?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+        mp::utils::trim_end(home);
+        absolute = home + '/' + complete_path.path().toStdString();
+    }
+    else
+    {
+        absolute = complete_path.path().toStdString();
+    }
+
+    std::string existing =
+        run_cmd(session, fmt::format("/bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'",
```
Saviq: Just realized this needs to run under `sudo`, in case part of the path is not accessible to the user.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+    return {existing,
+            QDir(QString::fromStdString(existing)).relativeFilePath(QString::fromStdString(absolute)).toStdString()};
 }
 
-void set_owner_for(mp::SSHSession& session, const std::string& target)
+// Create a directory on a given root folder.
+void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
+{
+    if (!relative_target.empty())
+        run_cmd(session, fmt::format("cd \"{}\" && sudo mkdir -p \"{}\"", root, relative_target));
```
Saviq: Probably best to wrap this in `sudo bash -c`, too.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,19 +62,50 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+// Assume it is already created.
+void set_owner_for(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
 {
     auto vm_user = run_cmd(session, "id -nu");
     auto vm_group = run_cmd(session, "id -ng");
     mp::utils::trim_end(vm_user);
     mp::utils::trim_end(vm_group);
 
-    run_cmd(session, fmt::format("sudo chown {}:{} \"{}\"", vm_user, vm_group, target));
+    run_cmd(session, fmt::format("/bin/bash -c 'cd \"{}\" && sudo chown -R {}:{} {}'", root, vm_user, vm_group,
```
Saviq: `sudo bash -c`, too.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: yes, I tried with cd() until I realized it tries to do operations on the local filesystem

src/sshfs_mount/sshfs_mount.cpp
luis4a0: right

src/sshfs_mount/sshfs_mount.cpp
luis4a0: ok

src/sshfs_mount/sshfs_mount.cpp
luis4a0: ok

src/sshfs_mount/sshfs_mount.cpp
luis4a0: I'd like to keep QString and QDir local to the split function. Note that if we convert some std::string to QString, we may need to use QString::toStdString in some places, thus we are in the same situation, I think.

luis4a0:
Code should be ok this time :)
Now I'm going for the tests.

tests/test_sshfsmount.cpp
```diff
@@ -39,20 +42,23 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-    mp::SshfsMount make_sshfsmount()
+    mp::SshfsMount make_sshfsmount(const std::string& target = "")
     {
         mp::SSHSession session{"a", 42};
-        return {std::move(session), default_source, default_target, default_map, default_map};
+        return {std::move(session), default_source, target == "" ? default_target : target, default_map, default_map};
     }
 
-    auto make_exec_that_fails_for(const std::string& expected_cmd, bool& invoked)
+    auto make_exec_that_fails_for(const std::vector<std::string>& expected_cmds, bool& invoked)
```
Saviq: You did not end up using this change as far I can tell?

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +247,107 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+    auto request_exec = make_exec_that_fails_for({"pwd"}, invoked);
+    REPLACE(ssh_channel_request_exec, request_exec);
+
+    EXPECT_THROW(make_sshfsmount(), std::runtime_error);
+    EXPECT_TRUE(invoked);
+}
+
+TEST_F(SshfsMount, executes_commands)
+{
+    std::unordered_map<std::string, std::string> commands = {
```
Saviq: The order of those actually matters, think you could easily make it so that this fails if commands are run out of order?

I suppose it still makes sense to ignore calls that we don't include in the map, but if we're interested about a call, ideally we'd check the order.

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +247,107 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+        {
+            if (executed.find(command.first) == executed.end() && cmd.find(command.first) != std::string::npos)
+            {
+                output = command.second;
+                remaining = output.size();
+                executed.insert(command);
+                invoked = true;
+                break;
+            }
+        }
```
Saviq: Could this be abstracted and shared between here and the test below?

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +247,107 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+        {"id -ng", "ubuntu"},
+        {"id -u", "1000"},
+        {"id -g", "1000"},
+        {"sudo sshfs -o slave -o nonempty -o transform_symlinks -o allow_other :\"source\" \"target\"", "don't care"}};
+
+    std::unordered_map<std::string, std::string> executed;
+
+    bool invoked{false};
+    std::string output{"1000"};
+    auto remaining = output.size();
```
Saviq: This seems arbitrary, shouldn't these be empty to start with?

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +247,107 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+                invoked = true;
+                break;
+            }
+        }
+
+        return SSH_OK;
+    };
+    REPLACE(ssh_channel_request_exec, request_exec);
+
+    make_sshfsmount();
```
Saviq: I'd be explicit about `target` here, otherwise I wasn't sure where "target" comes from in the commands matched above.

tests/test_sshfsmount.cpp
```diff
@@ -39,20 +42,23 @@ struct SshfsMount : public mp::test::SftpServerTest
         channel_is_closed.returnValue(0);
     }
 
-    mp::SshfsMount make_sshfsmount()
+    mp::SshfsMount make_sshfsmount(const std::string& target = "")
     {
         mp::SSHSession session{"a", 42};
-        return {std::move(session), default_source, default_target, default_map, default_map};
+        return {std::move(session), default_source, target == "" ? default_target : target, default_map, default_map};
```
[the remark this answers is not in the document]
luis4a0: Sure, changed.

tests/test_sshfsmount.cpp
luis4a0: No, I didn't using. I'll let it as it was.

tests/test_sshfsmount.cpp
luis4a0: Sure, they were a leftover from a previous implementation.

tests/test_sshfsmount.cpp
luis4a0: I'm not sure of getting your point here. 'target' is a string representing the mount point on the commands above, but I don't understand how to be explicit: using a variable to represent the mount point?

Saviq: I just meant `make_sshfsmount("target")` here - or maybe even better, `make_sshfsmount("different/target")` :)

luis4a0: OK, now I understand. It makes sense.

tests/test_sshfsmount.cpp
luis4a0: Done.

tests/test_sshfsmount.cpp
luis4a0: Done

tests/test_sshfsmount.cpp
luis4a0: Done

tests/test_sshfsmount.cpp
luis4a0: Sure, done

tests/test_sshfsmount.cpp
```diff
@@ -39,10 +47,10 @@ struct SshfsMount : public mp::test::SftpServerTest
         channel_is_closed.returnValue(0);
     }
 
-    mp::SshfsMount make_sshfsmount()
+    mp::SshfsMount make_sshfsmount(mp::optional<std::string> target = mp::nullopt)
     {
         mp::SSHSession session{"a", 42};
-        return {std::move(session), default_source, default_target, default_map, default_map};
+        return {std::move(session), default_source, target.value_or(default_target), default_map, default_map};
```
Saviq: If `std::optional::value` is unsupported on macOS, I doubt this will be…

tests/test_sshfsmount.cpp
```diff
@@ -21,14 +21,22 @@
...
+#include <vector>
+#include <algorithm>
+#include <iterator>
 
 namespace mp = multipass;
 namespace mpt = multipass::test;
 
 using namespace testing;
 
+typedef std::vector<std::pair<std::string, std::string>> command_vector;
```
Saviq: Use CamelCase for this so it's clearer it's a type?

tests/test_sshfsmount.cpp
```diff
@@ -59,10 +67,52 @@ struct SshfsMount : public mp::test::SftpServerTest
         return request_exec;
     }
 
+    auto make_exec_to_check_commands(const command_vector& commands, std::string::size_type& remaining, command_vector::const_iterator& next_expected_cmd, std::string& output, bool& invoked)
+    {
+        auto request_exec = [&commands, &remaining, &next_expected_cmd, &output, &invoked](ssh_channel, const char* raw_cmd) {
+            std::string cmd{raw_cmd};
+            mp::utils::trim_end(cmd);
```
Saviq: Do we need to trim this? I feel like this may hide some extra whitespace that we shouldn't be executing anyway.

tests/test_sshfsmount.cpp
```diff
@@ -59,10 +67,52 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+
+            if (next_expected_cmd != commands.end())
+            {
+                // Check if the first of the remaining commands is executed first.
```
Saviq:
```suggestion
                // Check if the first of the remaining commands is being executed.
```

tests/test_sshfsmount.cpp
```diff
@@ -59,10 +67,52 @@ struct SshfsMount : public mp::test::SftpServerTest
         return request_exec;
     }
 
+    auto make_exec_to_check_commands(const command_vector& commands, std::string::size_type& remaining, command_vector::const_iterator& next_expected_cmd, std::string& output, bool& invoked)
```
Saviq: Is `invoked` actually used still?

tests/test_sshfsmount.cpp
```diff
@@ -59,10 +67,52 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+                    invoked = true;
+                    ++next_expected_cmd;
+                }
+                else
+                {
+                    // If not, then we have to check the rest of the remaining
+                    // commands. If the executed command is there, it means that the
+                    // executed commands do not follow the order of our expected
+                    // command vector.
+                    for (auto it2 = std::next(next_expected_cmd); it2 != commands.end(); ++it2)
```
Saviq: It's the only iterator, just `it` would be enough?

tests/test_sshfsmount.cpp
```diff
@@ -88,7 +156,7 @@ struct SshfsMount : public mp::test::SftpServerTest
 TEST_F(SshfsMount, throws_when_sshfs_does_not_exist)
 {
     bool invoked{false};
-    auto request_exec = make_exec_that_fails_for("which sshfs", invoked);
+    auto request_exec = make_exec_that_fails_for({"which sshfs"}, invoked);
```
Saviq: Undo that change, please - and all the similar ones below.

tests/test_sshfsmount.cpp
```diff
@@ -59,10 +67,52 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+                    // executed commands do not follow the order of our expected
+                    // command vector.
+                    for (auto it2 = std::next(next_expected_cmd); it2 != commands.end(); ++it2)
+                    {
+                        if (cmd == it2->first)
+                        {
+                            output = it2->second;
+                            remaining = output.size();
+                            invoked = true;
+                            ADD_FAILURE() << "\"" << next_expected_cmd->first << "\" is not executed before \"" << (it2->first) << "\"";
```
Saviq: How about this:

```suggestion
                            ADD_FAILURE() << "\"" << (it2->first) << "\" executed out of order; expected \"" << next_expected_cmd->first << "\"";
```

tests/test_sshfsmount.cpp
```diff
@@ -74,6 +124,24 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+
+        auto channel_read = make_channel_read_return(output, remaining, invoked);
+        REPLACE(ssh_channel_read_timeout, channel_read);
+
+        auto request_exec = make_exec_to_check_commands(commands, remaining, next_expected_cmd, output, invoked);
+        REPLACE(ssh_channel_request_exec, request_exec);
+
+        make_sshfsmount(target.value_or(default_target));
+
+        EXPECT_EQ(next_expected_cmd, commands.end()) << "\"" << next_expected_cmd->first << "\" is not executed in the right place";
```
Saviq: If this doesn't pass, we know the command did not execute, no? Also, `EXPECT_EQ` is unnecessarily verbose here:
```
/home/michal/dev/canonical/multipass/multipass/tests/test_sshfsmount.cpp:146: Failure
Expected equality of these values:
  next_expected_cmd
    Which is: 8-byte object <10-93 30-F7 E3-55 00-00>
  commands.end()
    Which is: 8-byte object <10-94 30-F7 E3-55 00-00>
```

How about this?
```suggestion
        EXPECT_TRUE(next_expected_cmd == commands.end()) << "\"" << next_expected_cmd->first << "\" not executed";
```

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +309,52 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+         "/home/ubuntu/"},
+        {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", ""},
+        {"id -nu", "ubuntu"},
+        {"id -ng", "ubuntu"},
+        {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R ubuntu:ubuntu target'", ""},
+        {"id -u", "1000"},
+        {"id -g", "1000"},
+        {"sudo sshfs -o slave -o nonempty -o transform_symlinks -o allow_other :\"source\" \"target\"", "don't care"}};
+
+    test_command_execution(commands, "target");
```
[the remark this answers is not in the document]
luis4a0: Fixed, explicitly converted the `char*` to `std::string`.

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +309,52 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+
+TEST_F(SshfsMount, works_with_absolute_paths)
+{
+    command_vector commands = {
+        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+         "/home/ubuntu/"},
+        {"id -u", "1000"},
+        {"id -g", "1000"}};
+
+    test_command_execution(commands, "/home/ubuntu/target");
```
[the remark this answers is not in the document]
luis4a0: Same here.

tests/test_sshfsmount.cpp
luis4a0: Indeed, this works on macos.

tests/test_sshfsmount.cpp
luis4a0: Right, fixed.

tests/test_sshfsmount.cpp
luis4a0: You're right, it's unnecesary. I tested it.

tests/test_sshfsmount.cpp
luis4a0: Fixed.

tests/test_sshfsmount.cpp
luis4a0: When a command is invoked, `invoke` makes the return channel to read. AFAICS, it's needed. Am I missing something?

tests/test_sshfsmount.cpp
luis4a0: Done.

tests/test_sshfsmount.cpp
luis4a0: Sure, the name remains from a previous implementation. Changed.

tests/test_sshfsmount.cpp
luis4a0: Better

tests/test_sshfsmount.cpp
luis4a0: Better. Changed.

tests/test_sshfsmount.cpp
Saviq: OK now I see where it is used… My only problem then is that invoked becomes `true` as soon as the first command is invoked, and it never goes false. So in theory an out-of-band read could be called, and the tests wouldn't catch that.

I think this is fine for now, but maybe we should add a TODO to make `test_command_execution` mimic this better, probably replacing `ssh_channel_read_timeout` on each command being invoked, and failing inside it if it's being called beyond the end of the returned data. Or at least resetting `invoked` to `false` when `remaining` goes 0.

luis4a0: OK, now I see your point. I made a minor change, it should me much better now.

Saviq:
Yes!

bors r+

Issue #1370, opened 2020-02-19
[win] window size change not respected in Windows Terminal


Saviq:
When running `multipass shell` in the upcoming [Windows Terminal](https://www.microsoft.com/en-us/p/windows-terminal-preview/9n0dx20hk701), changing the window size often results in a broken console setup.

Easiest to see with `byobu`, which fills the whole screen.

marcodelapierre:
I am experiencing a similar issue on the mac

Saviq:
Hi @marcodelapierre, we've just fixed this in #1354, can you please try the release candidate from https://github.com/canonical/multipass/pull/1375#issuecomment-589779268 - if that doesn't help, please file a separate issue. Thanks!

marcodelapierre:
thanks @Saviq , just tested, my terminal resizing issue was fixed!

townsend2010:
Boo, [console WinEvents](https://docs.microsoft.com/en-us/windows/console/console-winevents) are not being used in the new Windows Terminal.  Will have to investigate what API it is using and see if there is some way to hook into it.

townsend2010:
Seems we're going to have to redo how we detect window size changes: https://github.com/microsoft/terminal/issues/1811

ricab:
Yeah I was planning to take a look as part of the work to integrate on the windows terminal.

ricab:
I've been investigating this and here are a few conclusions.

WT [supports](https://github.com/microsoft/terminal/issues/3238#issuecomment-543342919) built-in resize events. These [WINDOW_BUFFER_SIZE_EVENT](https://docs.microsoft.com/en-us/windows/console/input-record-str)s really should be triggered [whenever the viewport is resized](https://github.com/microsoft/terminal/issues/281#issuecomment-470692714). IIUC, these "events" can only be obtained from the buffer itself, mixed with the rest of the input, via [ReadConsoleInput](https://docs.microsoft.com/en-us/windows/console/readconsoleinput) or [PeekConsoleInput](https://docs.microsoft.com/en-us/windows/console/peekconsoleinput). However, we use [ReadConsole](https://docs.microsoft.com/en-us/windows/console/readconsole) to get the text, which discards those events. We also use a [flush](https://github.com/canonical/multipass-private/blob/625335bac8659eb8a8f6ad58cae4ba3cddae52c2/src/platform/console/windows_console.cpp#L118), whose purpose isn't clear to me. 

[...]

Issue #1258, opened 2019-12-19
Symbol not found trying to run on OSX 10.11.6


iammer:
Just installed 1.0.0 on OSX 10.11.6.  When I try to run I get the following error:

```
$ multipass
dyld: Symbol not found: __ZNSt19bad_optional_accessD1Ev
  Referenced from: /usr/local/bin/multipass
  Expected in: /usr/lib/libc++.1.dylib
 in /usr/local/bin/multipass
Trace/BPT trap: 5
```

Saviq:
Hi @iammer, it looks like we're using features from libraries of newer macOS than you're on… Any chance you could upgrade to the latest?

iammer:
Yes.
You might want to update your downloads page, it lists the minimum macOS version as: Yosemite 10.10.3 or later.

Saviq:
Hi @iammer, yes, we'll need to check what's the earliest we can run on - we didn't know about this new limitation. We may also try and work around the issue.

fclaerho:
Confirming I have the exact same issue, El Capitan (10.11.6).

robbles:
Same issue on Sierra (10.12.6).

townsend2010:
For reference: https://stackoverflow.com/questions/58469837/how-do-i-include-a-particular-version-of-the-c-library-in-a-mac-os-bundle-made

townsend2010:
Could someone affected by this please post the output of `nm -gU /usr/lib/libc++.1.dylib | grep optional`?

Thanks!

townsend2010:
I *think* I may have found the root cause of this.  Could anyone affected by this please try a [test Multipass package](https://transfer.sh/PatQj/multipass-1.1.0-dev.594+gbbfa89bd.mac-Darwin.pkg)?

Thanks!

townsend2010:
Hey all,

It looks like `transfer.sh` is having problems, so please try the package found at https://easyupload.io/noxhg8.

Thanks!

beilmit:
Hi @townsend2010 
I tested the operation using the specified test package on my environment. The results are as follows.

```
* 10.11
kurosawa-no-Mac:~ kurosawa$ multipass version
Sorry, "multipass" cannot be run on this version of macOS. Qt requires macOS 10.12.0 or later, you have macOS 10.11.6.

kurosawa-no-Mac:~ kurosawa$ /Library/Application\ Support/com.canonical.multipass/bin/multipassd -v
[...]

Issue #1170, opened 2019-11-07
[windows] `exec` hanging when output is redirected


Saviq:
There's a couple of similar bugs (#667 and #577 at least), but this is specifically about:

```powershell
PS> multipass exec <instance> -- lsb_release -a | Out-Null
```

This hangs indefinitely, and is a blocker for Snapcraft on Windows.

NickZ:
This is still present in the windows build of 1.1.0rc. I haven't been able to reproduce on guests using the 16.04 image, but I can get 18.04 guests to reliably trigger this. This affects snapcraft, which will hang during the build process

I can reliably reproduce using this python script: https://gist.github.com/NickZ/58e87d6beaf6473a61da6856802154da

NickZ:
You can also just run `multipass exec primary -- sudo -i env printenv HOME | Out-Null` in powershell a few times and it will eventually trigger

townsend2010:
Hey @NickZ,

Yeah, I can reproduce as well :frowning_face:  I'll just reopen this and continue to investigate.  Thanks for reporting!

NickZ:
Unfortunately, still able to trigger this on the latest build here: https://github.com/canonical/multipass/pull/1375#issuecomment-593880436

While it may take a while longer, given enough iterations it will trigger.

Using my python script above.
```
PS C:\Users\nick\Documents> python .\trigger_1170.py
Using primary
testing output...
mesg: ttyname failed: Inappropriate ioctl for device
b'/root\n'
testing output...
mesg: ttyname failed: Inappropriate ioctl for device
b'/root\n'
testing output...
mesg: ttyname failed: Inappropriate ioctl for device
b'/root\n'
testing output...
mesg: ttyname failed: Inappropriate ioctl for device
b'/root\n'
testing output...
mesg: ttyname failed: Inappropriate ioctl for device
b'/root\n'
testing output...
mesg: ttyname failed
```
where it hangs.

townsend2010:
Hey @NickZ,

Thanks for letting us know.  I think the behavior is much better than before as I think your script is an extreme case and in normal operation, this should work.  That said, we will continue to work on a better fix, but it won't block our upcoming release.

ricab:
Same thing in reverse:

```cmd
C:\Users\ricab>multipass exec -n rich-zorilla -- bash -c "echo some_file contents > some_file"

C:\Users\ricab>echo asdf | multipass exec -n rich-zorilla -- cat -E some_file - some_file
[...]

PR #1375, opened 2020-02-20
Release 1.1.0

Saviq:
---

Diff from master: https://paste.ubuntu.com/p/3P4BKf9q44/


ricab:
@Saviq Would you cherry-pick the new commit into master? 

Maybe we can have Travis test this but then get it in a regular PR, merged into master, and cherry-pick that into the release branch, as usual? If it is still just a small change and Travis does the right thing here, send it my way and I'll approve when I see it (=tomorrow).

townsend2010:
I tested the Mac package and all seems well.  That said, as mentioned on IRC, I'd like to set a new minimum OSX version.  Since we build on a `10.13` machine, I thought we should set the minimum version to that.  However, https://github.com/canonical/multipass/issues/1258#issuecomment-587455490 seems to indicate that `10.14` is the minimum, but it is not clear to me where that comes from.

townsend2010:
As an aside, I tried `snap refresh multipass --channel candidate/1.1` per the instructions above and I get:
```
error: requested a non-existing branch on latest/candidate for snap "multipass": 1.1
```

marcodelapierre:
Just tested the rc on a mac, to test for a terminal resizing issue I had with 1.0.
The release candidate has fixed my issue.
(see also my comment in #1370 )

Saviq:
> ```
> error: requested a non-existing branch on latest/candidate for snap "multipass": 1.1
> ```

Fixed: 79004a6

Saviq:
I've tested all of [the v1.1.0 milestone](https://github.com/canonical/multipass/milestone/7?closed=1) on Linux (and the other platforms, where the issue was platform-specific) and confirmed it all worked, didn't see any new issues.

Refreshing from `stable` to `candidate/1.1` also works, recovering instance state.

NickZ:
Issue #1170 is still present in this build: https://github.com/canonical/multipass/issues/1170#issuecomment-593051354

PR #1382, opened 2020-02-24
[logging] add trace level (Fixes #1371)

### LINKED ISSUE #1371: AppArmor profiles spamming the log
It became near impossible to look through multipassd logs with all the AppArmor profile loads / unloads.

I think we need a `trace` log level that these will go to. It will have to map to DEBUG in `journald`, but we'd have it disabled by default.

Saviq:
Move AppArmor profile logging contents to the trace level to avoid spamming the logs.

### DIFF
--- a/src/platform/logger/journald_logger.cpp
+++ b/src/platform/logger/journald_logger.cpp
@@ -29,6 +29,8 @@ constexpr auto to_syslog_priority(const mpl::Level& level) noexcept
 {
     switch (level)
     {
+    case mpl::Level::trace:
+        return LOG_DEBUG;
     case mpl::Level::debug:
         return LOG_DEBUG;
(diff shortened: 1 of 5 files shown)


src/platform/logger/journald_logger.cpp
```diff
@@ -29,6 +29,8 @@ constexpr auto to_syslog_priority(const mpl::Level& level) noexcept
 {
     switch (level)
     {
+    case mpl::Level::trace:
+        return LOG_DEBUG;
```
townsend2010: Unless you're accounting for a future `LOG_TRACE` in `syslog.h`, then the `trace` and `debug` cases can be combined.

Saviq: Right! Done.

PR #1393, opened 2020-02-28
Introduce bug template

Saviq:
---

The result of this is:

https://github.com/Saviq/multipass/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+Describe what your problem is.
+
+**To Reproduce**
+How, and what happened?
+1. `multipass ...`
+1.
+
+**Expected behavior**
+What did you expect to happen?
+
+**Logs**
+Please provide logs from the daemon, see [accessing logs](https://multipass.run/docs/accessing-logs) on where to find them on your platform.
+
+**Additional info**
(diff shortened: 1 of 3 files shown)


townsend2010:
By the way, I noticed this by going to https://github.com/lxc/lxd/blob/master/.github/ISSUE_TEMPLATE.md and saw the message at the top.

townsend2010:
Many bug reports we have been getting certainly are lacking the host OS, version, and logs.  And as we do more releases, the version of Multipass the user is running will become much more relevant.  I agree, bug templates that ask for quite a bit of detail are annoying at best and at worst will put of the user and they won't report anything.

I'm +1 as long as we keep it simple like this proposes.

luis4a0:
@ricab : the template specify the steps to get the necessary information, I don't see a big harm asking for this info. We can at least say that something (like pasting the logs) is optional. I agree with @townsend2010 in that we _need_ these data.

Saviq:
@townsend2010 I just used GitHub UI to create this… So their own UI generates an old template…

@ricab we can certainly reduce it if you feel it too heavy?

What I'd love in the long run is a `multipass report-bug` that will actually collect all of that data for the user. This is a first step in that direction, really.

townsend2010:
> I just used GitHub UI to create this… So their own UI generates an old template…

lol, ok then...

ricab:
@luis4a0, @townsend2010 I understand your points, thanks.

@Saviq nothing specific. Automating sounds interesting, even if we'd need to be careful with privacy stuff.

Saviq:
I added a feature request template and a couple contact links, the result can be seen here:

https://github.com/Saviq/multipass/issues/new/choose

Saviq:
bors r=townsend2010

PR #1399, opened 2020-03-02
[travis] Randomize tests

ricab:
Randomize tests in CI. Gtest should print the seed for reproducibility.


Saviq:
Yass.

bors r+

ricab:
For future reference, the seed that failed was 26454: https://travis-ci.org/canonical/multipass/jobs/657303264#L4989

ricab:
I was working with the singleton mocks and hit shuffled tests failing locally. I had another look at it and reported in #1469.


The seed above doesn't reproduce with current code anymore, but one seed that fails for me is 76165. 

The following will find a failing seed: `while multipass_tests --gtest_shuffle ; do : ; done`.

PR #1420, opened 2020-03-10
switched to https for image downloads

### LINKED ISSUE #1419: Download images over HTTPS versus HTTP
Just had a timeout and noticed that the images being downloaded are over HTTP and not HTTPS.
 (e.g. failed to download from 'http://cloud-images.ubuntu.com/releases/server/releases/bionic/release-20200218/ubuntu-18.04-server-cloudimg-amd64.img'). 

Switching to HTTPS works as expected: https://cloud-images.ubuntu.com/releases/bionic/release-20200218/ubuntu-18.04-server-cloudimg-amd64.img

Should these URLs be https instead?

jasonmccallister:
Resolves #1419

### DIFF
--- a/src/daemon/custom_image_host.cpp
+++ b/src/daemon/custom_image_host.cpp
@@ -55,7 +55,7 @@ struct CustomImageInfo
 
 const QMap<QString, CustomImageInfo> multipass_image_info{
     {{"ubuntu-core-16-amd64.img.xz"},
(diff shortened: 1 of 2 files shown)


Saviq:
Hi @jasonmccallister thanks for this - it does indeed make sense. I've filed #1453 to track catering for a certain use case that this will make problematic.

bors try

PR #1422, opened 2020-03-11
daemon/find: Account for no valid aliases being found

### LINKED ISSUE #1411: multipass find release FATAL
following: https://discourse.ubuntu.com/t/multipass-find-command/8351

```
$ multipass find release:
[libprotobuf FATAL ../3rd-party/grpc/third_party/protobuf/src/google/protobuf/repeated_field.h:1557] CHECK failed: (index) < (current_size_):
libc++abi.dylib: terminating with uncaught exception of type google::protobuf::FatalException: CHECK failed: (index) < (current_size_):
[1]    91774 abort      multipass find release:
```

townsend2010:
On particular platforms, it's possible that some aliases that are found for a
particular remote are not valid, so the created entry needs to be removed in
this case.

Fixes #1411

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -904,13 +904,13 @@ try // clang-format on
     {
         const auto remote = request->remote_name();
 
-        if (!mp::platform::is_remote_supported(remote))
-            throw std::runtime_error(fmt::format(
-                "{} is not a supported remote. Please use `multipass find` for list of supported images.", remote));
-
         auto it = remote_image_host_map.find(remote);
         if (it == remote_image_host_map.end())
             throw std::runtime_error(fmt::format("Remote \"{}\" is unknown.", remote));
+
+        if (!mp::platform::is_remote_supported(remote))
+            throw std::runtime_error(fmt::format(
+                "{} is not a supported remote. Please use `multipass find` for list of supported images.", remote));
 
         auto vm_images_info = it->second->all_images_for(remote, request->allow_unsupported());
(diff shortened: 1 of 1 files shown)


src/daemon/daemon.cpp
```diff
@@ -928,6 +928,13 @@ try // clang-format on
                     alias_entry->set_alias(alias.toStdString());
                 }
 
+                // If no aliases are found, then it's an invalid entry
+                if (entry->aliases_info().empty())
+                {
+                    response.mutable_images_info()->RemoveLast();
+                    continue;
+                }
+
```
Saviq: Hmm this feels like a workaround, shouldn't we be not getting invalid entries here?

townsend2010: Above this code in the `for` loop is where the check is for if an alias is supported or not.  If no aliases are found, then as this PR proposes, we should not have an entry for the response.  This whole thing is kind of a hack since we added the whole notion of which aliases are supported and which are not depending on the platform well after how we initially implemented remote image querying.  If we want to move all of the logic to determine that somewhere else, then it will require quite a bit of refactoring which beyond the scope of this PR.  This PR simply fixes a crash on Windows and Mac when issuing `multipass find release:`.

Saviq: Well, yeah, I just mean in the `for` above, we shouldn't even be adding an entry unless we know it's not going to be empty?

Saviq:
Discussed on IRC, this is the best we can do atm.

bors r+

PR #1424, opened 2020-03-11
Fix inconsistencies in mount/umount parameters.

luis4a0:
Path names are now normalized before mounting or unmounting.

This way, directories can be mounted using relative paths and unmounted using absolute paths or viceversa. Also, the final slash in the path can be omitted.

The fix works in the multipass client, so the daemon always receives an absolute mountpoint as parameter.

---

Actual diff: https://github.com/canonical/multipass/compare/fix-mount-dir-permissions...fix-mount-umount-inconsistencies

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -1241,9 +1241,40 @@ try // clang-format on
         }
 
         auto& vm = it->second;
+        auto& vm_specs = vm_instance_specs[name];
 
         if (vm->current_state() == mp::VirtualMachine::State::running)
         {
+            // If the given path is relative, we need to resolve the fill
+            // absolute path. For this, we execute the command 'pwd'.
+            if (QDir(QString::fromStdString(target_path)).isRelative())
+            {
+                mp::SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm_specs.ssh_username, *config->ssh_key_provider};
+
+                std::string home;
+
+                auto proc = session.exec("pwd");
+                if (proc.exit_code() != 0)
+                {
+                    auto error_msg = proc.read_std_error();
+                    mpl::log(
+                        mpl::Level::warning, category,
+                        fmt::format("failed to run 'pwd', error message: '{}'", mp::utils::trim_end(error_msg)));
+                    home = std::string{};
+                }
+
+                home = proc.read_std_output();
+                mp::utils::trim_end(home);
+
+                target_path = mp::utils::normalize_absolute_path(home + '/' + target_path);
+                mpl::log(mpl::Level::info, category, fmt::format("home directory: {}, target_path: {}", home, target_path));
+            }
+            else
+            {
+                target_path = mp::utils::normalize_absolute_path(target_path);
+                mpl::log(mpl::Level::info, category, fmt::format("target_path: {}", target_path));
+            }
+
             try
             {
                 instance_mounts.start_mount(vm.get(), request->source_path(), target_path, gid_map, uid_map);
@@ -1274,7 +1305,6 @@ try // clang-format on
             }
         }
 
-        auto& vm_specs = vm_instance_specs[name];
         if (vm_specs.mounts.find(target_path) != vm_specs.mounts.end())
         {
             fmt::format_to(errors, "There is already a mount defined for \"{}:{}\"\n", name, target_path);
@@ -1632,6 +1662,7 @@ try // clang-format on
         auto target_path = path_entry.target_path();
         auto& mounts = vm_instance_specs[name].mounts;
         auto& vm = it->second;
+        auto& vm_specs = vm_instance_specs[name];
 
         // Empty target path indicates removing all mounts for the VM instance
         if (target_path.empty())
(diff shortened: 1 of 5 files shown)


src/daemon/daemon.cpp
```diff
@@ -1241,9 +1241,40 @@ try // clang-format on
         }
 
         auto& vm = it->second;
+        auto& vm_specs = vm_instance_specs[name];
 
         if (vm->current_state() == mp::VirtualMachine::State::running)
         {
+            // If the given path is relative, we need to resolve the fill
```
ricab:
```suggestion
            // If the given path is relative, we need to resolve the full
```

src/daemon/daemon.cpp
```diff
@@ -1241,9 +1241,40 @@ try // clang-format on
...
+
+                target_path = mp::utils::normalize_absolute_path(home + '/' + target_path);
+                mpl::log(mpl::Level::info, category, fmt::format("home directory: {}, target_path: {}", home, target_path));
+            }
+            else
+            {
+                target_path = mp::utils::normalize_absolute_path(target_path);
+                mpl::log(mpl::Level::info, category, fmt::format("target_path: {}", target_path));
+            }
+
```
ricab: Since this code is repeated below, could it be extracted into a helper function, to avoid duplication? Something returning the resolved and normalized `target_path` from the given the VM name and `target_path` input.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -59,30 +60,56 @@ void check_sshfs_exists(mp::SSHSession& session)
     run_cmd(session, "which sshfs", error_handler);
 }
 
-void make_target_dir(mp::SSHSession& session, const std::string& target)
+// Split a path into existing and to-be-created parts. The input is assumed to
+// be an absolute path, normalized with mp::utils::normalize_absolute_path().
+std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, const std::string& target)
```
ricab: Isn't this part of the other PR? In such occasions, where you need the changes of a PR that still hasn't gone in, it is better to stack the PRs. So you'd branch this PR out of where the other PR left off.

src/utils/utils.cpp
```diff
@@ -342,3 +342,10 @@ void mp::utils::check_and_create_config_file(const QString& config_file_path)
         config_file.open(QIODevice::WriteOnly);
     }
 }
+
+// Given an absolute path, removes trailing slash, multiple slashes and
+// redundant "." and "..".
+std::string mp::utils::normalize_absolute_path(const std::string& path)
+{
+    return QDir::cleanPath(QString::fromStdString(path)).toStdString();
```
ricab: Nice, `QDir::cleanPath` indeed seems the right choice for this. Symlinks are left alone, but that is probably ok, and if we decide we should resolve them, that can always done in a separate PR.

luis4a0: Sure. Symlinks can be resolved exactly as we executed 'pwd' when mounting/unmounting. We'll see that later.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Yes, that's part of the other PR. There can be a problem when merging we'd need to solve. That can be done. I can try to stack the PR's, would be a good idea (altough I'll need to force-push at some point).

src/daemon/daemon.cpp
luis4a0: Nice, I'll do it.

src/daemon/daemon.cpp
luis4a0: Fixed.

src/utils/utils.cpp
ricab: Right, something along those lines, like `pwd -P` or `readlink -f`. Something to think about.

luis4a0:
It's strange, lcov says that nothing is tested from the daemon functions. I'll take a look to the tests.

src/sshfs_mount/sshfs_mount.cpp
ricab: Right, it is generally a good idea to avoid commit overlap. So stacking branches/PRs decouples things and is usually better (even though GH is not very good dealing with those). 

On force-pushing, I know it is generally not advised, but it is pretty much part of our usual workflow for PR branches (like when rebasing on the tip of master). It is fine as long as no one depends on the changing refs, so we usually avoid depending on each other's branches and we're ok if we don't force-push master.

luis4a0: Well, we can manually merge on this specific case, there will be at most one conflict and the file on this commit is what will prevail. Next time I'll pay attention to this issue.

About force pushing, I'm easy. Since I am the only one pushing to this branch, there will be no issues.

luis4a0: Solved, now this branch is on top of the other PR.

tests/test_sshfsmount.cpp
```diff
@@ -241,3 +309,52 @@ TEST_F(SshfsMount, unblocks_when_sftpserver_exits)
...
+{
+    bool invoked{false};
+    auto request_exec = make_exec_that_fails_for({"pwd"}, invoked);
+    REPLACE(ssh_channel_request_exec, request_exec);
+
+    EXPECT_THROW(make_sshfsmount(), std::runtime_error);
+    EXPECT_TRUE(invoked);
+}
+
+TEST_F(SshfsMount, executes_commands)
```
townsend2010: Sorry I'm late to this party, but this test is quite problematic.  I'm working on rebasing #1328 and have added a number of new commands and removed old ones.  This totally breaks this test and since this test is very specific to commands and order, it is not trivial to update it when we make changes elsewhere.   Looking at this test, I'm not sure what this test gains us that is not covered elsewhere.  Is this test really necessary?

townsend2010:
Sorry for the noise in my code comment.  It was directed at #1359, but having discussed this with @Saviq, I think we have a plan on accounting for unrelated sshfs-based command changes.

PR #1430, opened 2020-03-16
Top level catch all

### LINKED ISSUE #1429: Multipass needs a top-level catch-all
See [this](https://softwareengineering.stackexchange.com/a/308828). In general, things can throw, so in C++ there should be a top-level catch; or generically, a catch-all in the top-most function that calls anything that is not noexcept.

CLI and GUI clients seem to be missing this. #618 is related.

For a practical example, just change the permissions/format of one of the files we need.

ricab:
Fixes #1429.

---

Manual test examples (beyond unit tests):

- Daemon :
```console
$ ./bin/multipassd --logger=stderr -V debug  # non-admin user, no sudo
```
- CLI client:
```console
$ chmod -r ~/.local/share/multipass/client-certificate/multipass_cert.pem  # don't forget to revert
$ multipass
```
- GUI client:
```console
$ echo asdf > ~/.config/multipass/multipass.conf  # don't forget to revert
$ multipass.gui
```

### DIFF
--- a/include/multipass/top_catch_all.h
+++ b/include/multipass/top_catch_all.h
@@ -23,12 +23,12 @@
 
 namespace multipass
 {
-template <typename F, typename... Args>                           // F needs to return int
-int top_catch_all(const char* log_category, F f, Args&&... args); // not noexcept because logging isn't
+template <typename F, typename... Args>                                      // F needs to return int
+int top_catch_all(const logging::CString log_category, F f, Args&&... args); // not noexcept because logging isn't
 }
 
 template <typename F, typename... Args>
-inline int multipass::top_catch_all(const char* log_category, F f, Args&&... args)
+inline int multipass::top_catch_all(const logging::CString log_category, F f, Args&&... args)
 {
     namespace mpl = multipass::logging;
     try
--- a/include/multipass/top_catch_all.h
+++ b/include/multipass/top_catch_all.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2020 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


include/multipass/top_catch_all.h
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef MULTIPASS_TOP_CATCH_ALL_H
+#define MULTIPASS_TOP_CATCH_ALL_H
+
+#include <multipass/format.h>
+#include <multipass/logging/log.h>
+
+namespace multipass
+{
+template <typename F, typename... Args>                                      // F needs to return int
```
luis4a0: Altough the function is clear, I'd add a clear comment on what its purpose is. It's just a suggestion, I generally try to comment everything not trivial.

tests/test_top_catch_all.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+TEST_F(TopCatchAll, calls_function_with_args)
+{
+    int a = 5, b = 7, got = 0;
+    EXPECT_NO_THROW(got = mp::top_catch_all("", std::plus<int>{}, a, b););
+    EXPECT_EQ(got, a + b);
+}
+
+TEST_F(TopCatchAll, handles_unknown_error)
```
luis4a0: Besides `std::runtime_error` and unknown exceptions, I'd suggest to test also for other less common exceptions, such as those raised by ssh. Again, take this as a suggestion.

include/multipass/top_catch_all.h
ricab: Sure thing, I will do that!

tests/test_top_catch_all.cpp
ricab: Yup, ok I will. Do you have any specific exception type you could suggest? :)

include/multipass/top_catch_all.h
ricab: Done!

tests/test_top_catch_all.cpp
ricab: @luis4a0 I suppose you were referring to our own SSH exceptions (libssh itself is C), which are themselves derived from `std::runtime_error`. To account for other `std::exceptions`, I just created a test exception locally. Does that agree with what you had in mind?

luis4a0: That seems correct to me, thanks

PR #1439, opened 2020-03-20
Expand `~` in mount target (Fixes #1230)

### LINKED ISSUE #1230: unexpected handling of ~ by mount
Launched a new 18.04 VM:
```
$ multipass launch 18.04                                                                                                                         
Launched: saving-chamois   
$ multipass list                                                                                                                                 
saving-chamois          Running           10.235.69.2      Ubuntu 18.04 LTS 
```

Mounted a local directory using the following command:

```
$ multipass mount $PWD/some-dir saving-chamois:~/foo
```

The directory was mounted under `$HOME/~/foo` in the VM. Based on prior ssh/rsync experience, I expected `~` to be mapped to $HOME, and the mount to happen at `$HOME/foo`.

luis4a0:
This commit adds support for specifying mount points starting with ~.
It also handles the case on which mount points or home paths end with a space character.

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -72,6 +72,7 @@ std::string run_cmd_for_output(const QString& cmd, const QStringList& args, cons
 std::string& trim_end(std::string& s);
 std::string& trim_newline(std::string& s);
 std::string escape_char(const std::string& s, char c);
+std::string escape_for_shell(const std::string& s);
 std::vector<std::string> split(const std::string& string, const std::string& delimiter);
 std::string generate_mac_address();
 std::string timestamp();
--- a/src/sshfs_mount/sshfs_mount.cpp
+++ b/src/sshfs_mount/sshfs_mount.cpp
@@ -26,6 +26,8 @@
 #include <multipass/format.h>
 
 #include <QDir>
+#include <QString>
+#include <cassert>
 #include <iostream>
 
 #include <semver200.h>
@@ -122,26 +124,31 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
 // Split a path into existing and to-be-created parts.
 std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, const std::string& target)
 {
-    QDir complete_path(QString::fromStdString(target));
-    QString absolute;
+    std::string absolute;
 
-    if (complete_path.isRelative())
+    switch (target[0])
     {
-        std::string home = run_cmd(session, "pwd");
-        mp::utils::trim_newline(home);
-        absolute = QString::fromStdString(home) + '/' + complete_path.path();
-    }
-    else
-    {
-        absolute = complete_path.path();
+    case '~':
+        absolute =
+            run_cmd(session, fmt::format("echo ~{}", mp::utils::escape_for_shell(target.substr(1, target.size() - 1))));
+        mp::utils::trim_newline(absolute);
+        break;
+    case '/':
+        absolute = target;
+        break;
+    default:
+        absolute = run_cmd(session, fmt::format("echo $PWD/{}", mp::utils::escape_for_shell(target)));
+        mp::utils::trim_newline(absolute);
+        break;
     }
 
     std::string existing = run_cmd(
-        session,
-        fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'", absolute));
+        session, fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=\"${{P%/*}}\"; done; echo $P/'",
+                             absolute));
     mp::utils::trim_newline(existing);
 
-    return {existing, QDir(QString::fromStdString(existing)).relativeFilePath(absolute).toStdString()};
+    return {existing,
+            QDir(QString::fromStdString(existing)).relativeFilePath(QString::fromStdString(absolute)).toStdString()};
 }
 
 // Create a directory on a given root folder.
@@ -188,8 +195,8 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
         set_owner_for(session, leading, missing, default_uid, default_gid);
     }
 
-    return std::make_unique<mp::SftpServer>(std::move(session), source, target, gid_map, uid_map, default_uid,
-                                            default_gid, sshfs_exec_line);
+    return std::make_unique<mp::SftpServer>(std::move(session), source, leading + missing, gid_map, uid_map,
+                                            default_uid, default_gid, sshfs_exec_line);
 }
 
 } // namespace
--- a/tests/test_sshfsmount.cpp
+++ b/tests/test_sshfsmount.cpp
@@ -204,8 +204,8 @@ struct SshfsMount : public mp::test::SftpServerTest
     const std::unordered_map<std::string, std::string> default_cmds{
         {"snap run multipass-sshfs.env", "LD_LIBRARY_PATH=/foo/bar\nSNAP=/baz\n"},
         {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 3.0.0\n"},
-        {"pwd", "/home/ubuntu\n"},
-        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+        {"echo $PWD/target", "/home/ubuntu/target\n"},
+        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
          "/home/ubuntu/\n"},
         {"id -u", "1000\n"},
         {"id -g", "1000\n"},
@@ -299,57 +299,72 @@ TEST_P(SshfsMountExecuteThrowRuntErr, test_runtime_error_when_executing)
 //
 
 INSTANTIATE_TEST_SUITE_P(SshfsMountThrowWhenError, SshfsMountFail,
-                         testing::Values("mkdir", "chown", "id -u", "id -g", "cd", "pwd"));
+                         testing::Values("mkdir", "chown", "id -u", "id -g", "cd", "echo $PWD"));
 
 // Commands to check that a version of FUSE smaller that 3 gives a correct answer.
 CommandVector old_fuse_cmds = {{"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 2.9.0\n"},
                                {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -o slave -o transform_symlinks -o "
-                                "allow_other -o nonempty :\"source\" \"target\"",
+                                "allow_other -o nonempty :\"source\" \"/home/ubuntu/target\"",
                                 "don't care\n"}};
 
 // Commands to check that a version of FUSE at least 3.0.0 gives a correct answer.
 CommandVector new_fuse_cmds = {{"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 3.0.0\n"},
                                {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -o slave -o transform_symlinks -o "
-                                "allow_other :\"source\" \"target\"",
+                                "allow_other :\"source\" \"/home/ubuntu/target\"",
                                 "don't care\n"}};
 
 // Commands to check that an unknown version of FUSE gives a correct answer.
 CommandVector unk_fuse_cmds = {{"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "weird fuse version\n"},
                                {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -o slave -o transform_symlinks -o "
-                                "allow_other :\"source\" \"target\"",
+                                "allow_other :\"source\" \"/home/ubuntu/target\"",
                                 "don't care\n"}};
 
 // Commands to check that the server correctly creates the mount target.
 CommandVector exec_cmds = {
-    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
      "/home/ubuntu/\n"},
     {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", "\n"},
     {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R 1000:1000 \"target\"'", "\n"}};
+
+// Commands to check that it works with a path containing a space.
+CommandVector space_cmds = {
+    {"echo $PWD/space\\ odyssey", "/home/ubuntu/space odyssey\n"},
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/space odyssey\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
+     "/home/ubuntu/\n"}};
+
+// Commands to check that the ~ expansion works.
+CommandVector tilde1_cmds = {
+    {"echo ~/target", "/home/ubuntu/target\n"},
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
+     "/home/ubuntu/\n"}};
+
+// Commands to check that the ~user expansion works (assuming that user exists).
+CommandVector tilde2_cmds = {
+    {"echo ~ubuntu/target", "/home/ubuntu/target\n"},
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
+     "/home/ubuntu/\n"}};
 
 // Commands to check that the server works if an absolute path is given.
 CommandVector absolute_cmds = {
-    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
-     "/home/ubuntu/\n"},
-    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", "\n"},
-    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R 1000:1000 \"target\"'", "\n"}};
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
+     "/home/ubuntu/\n"}};
 
 // Commands to check that it works for a nonexisting path.
 CommandVector nonexisting_path_cmds = {
-    {"sudo /bin/bash -c 'P=\"/nonexisting/path\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'", "/\n"},
-    {"sudo /bin/bash -c 'cd \"/\" && mkdir -p \"nonexisting/path\"'", "\n"},
-    {"sudo /bin/bash -c 'cd \"/\" && chown -R 1000:1000 \"nonexisting\"'", "\n"}};
+    {"sudo /bin/bash -c 'P=\"/nonexisting/path\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'", "/\n"}};
 
 // Check the execution of the CommandVector's above.
-INSTANTIATE_TEST_SUITE_P(SshfsMountSuccess, SshfsMountExecute,
-                         testing::Values(std::make_pair("target", old_fuse_cmds), std::make_pair("target", exec_cmds),
-                                         std::make_pair("target", new_fuse_cmds), std::make_pair("target", exec_cmds),
-                                         std::make_pair("target", unk_fuse_cmds), std::make_pair("target", exec_cmds),
-                                         std::make_pair("/home/ubuntu/target", absolute_cmds),
-                                         std::make_pair("/nonexisting/path", nonexisting_path_cmds)));
+INSTANTIATE_TEST_SUITE_P(
+    SshfsMountSuccess, SshfsMountExecute,
+    testing::Values(std::make_pair("target", old_fuse_cmds), std::make_pair("target", new_fuse_cmds),
+                    std::make_pair("target", unk_fuse_cmds), std::make_pair("target", exec_cmds),
+                    std::make_pair("space odyssey", space_cmds), std::make_pair("~/target", tilde1_cmds),
+                    std::make_pair("~ubuntu/target", tilde2_cmds), std::make_pair("/home/ubuntu/target", absolute_cmds),
+                    std::make_pair("/nonexisting/path", nonexisting_path_cmds)));
 
 // Commands to test that when a mount path already exists, no mkdir nor chown is ran.
 CommandVector execute_no_mkdir_cmds = {
-    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=\"${P%/*}\"; done; echo $P/'",
      "/home/ubuntu/target/\n"}};
 
 INSTANTIATE_TEST_SUITE_P(SshfsMountSuccessAndAvoidCommands, SshfsMountExecuteAndNoFail,
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -154,6 +154,26 @@ std::string mp::utils::escape_char(const std::string& in, char c)
     return std::regex_replace(in, std::regex({c}), fmt::format("\\{}", c));
 }
 
+// Escape all characters which need to be escaped in the shell.
+std::string mp::utils::escape_for_shell(const std::string& in)
+{
+    std::string ret;
+    std::back_insert_iterator<std::string> ret_insert = std::back_inserter(ret);
+
+    for (char c : in)
+    {
+        // If the character is in one of these code ranges, then it must be escaped.
+        if (c < 0x25 || c > 0x7a || (c > 0x25 && c < 0x2b) || (c > 0x5a && c < 0x5f) || 0x2c == c || 0x3b == c ||
+            0x3c == c || 0x3e == c || 0x3f == c || 0x60 == c)
+        {
+            *ret_insert++ = '\\';
+        }
+        *ret_insert++ = c;
+    }
+
+    return ret;
+}
+
 std::vector<std::string> mp::utils::split(const std::string& string, const std::string& delimiter)
 {
     std::regex regex(delimiter);
--- a/tests/test_utils.cpp
+++ b/tests/test_utils.cpp
@@ -218,6 +218,13 @@ TEST(Utils, escape_char_actually_escapes)
     EXPECT_THAT(res, ::testing::StrEq("I've got \\\"quotes\\\""));
 }
 
+TEST(Utils, escape_for_shell_actually_escapes)
+{
+    std::string s{"I've got \"quotes\""};
+    auto res = mp::utils::escape_for_shell(s);
+    EXPECT_THAT(res, ::testing::StrEq("I\\'ve\\ got\\ \\\"quotes\\\""));
+}
+
 TEST(Utils, try_action_actually_times_out)
 {
     bool on_timeout_called{false};


src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -50,6 +50,25 @@ auto run_cmd(mp::SSHSession& session, std::string&& cmd)
     return run_cmd(session, std::forward<std::string>(cmd), error_handler);
 }
 
+// Run a command returning a string which can end in space on a given session and reads the result. It adds a
+// non-space character (actually, a '-') to signal the end of the resulting string, and later removes the trailing
+// spaces and the special character to gather the contents of the return string including trailing spaces of the
+// command output.
```
Saviq: This is catering for a very extreme edge case. IMO it adds complexity to handle something, which most often will be a mistake (trailing space in home folder).

@ricab @townsend2010 WDYT?

I wonder if the right thing to do would be to _not_ trim the string (other than removing any expected newlines)?

I mean, we're in control of the commands, and should be able to assume the behaviour of them to not return extraneous whitespace?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -62,6 +81,43 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+std::string expand_home_directory(mp::SSHSession& session, const std::string& target)
+{
+    std::string expanded_target;
+
+    if ('~' == target[0])
+    {
+        std::string home;
+
+        // Get the user name, which can be empty.
+        auto pos = (1 == target.size()) ? 1 : target.find('/', 1);
```
Saviq: Maybe overkill, but use a regular expression here? Would make for more concise code on our part.

Something like [`^~([^/]+)?(.+)?$`](https://regex101.com/r/P2pk8f/1/tests) should do.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -62,6 +81,43 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+        if (1 == pos)
+        {
+            home = run_string_cmd(session, "pwd");
+        }
```
Saviq: I would rather execute `getent passwd $USER...` for this case.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -62,6 +81,43 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+        else
+        {
+            std::string username = target.substr(1, pos - 1);
+            home = run_string_cmd(session, "getent passwd " + username + " | cut -d : -f 6");
```
Saviq: Any reason to not use `fmt::format`?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -62,6 +81,43 @@ void check_sshfs_exists(mp::SSHSession& session)
...
+            if (0 == home.size())
+            {
+                throw std::runtime_error("user " + username + " does not exist or does not have a home defined");
```
Saviq: Here, too, why not `fmt::format`? Please put the username in quotes, too.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -50,6 +50,25 @@ auto run_cmd(mp::SSHSession& session, std::string&& cmd)
...
 }
 
+// Run a command returning a string which can end in space on a given session and reads the result. It adds a
+// non-space character (actually, a '-') to signal the end of the resulting string, and later removes the trailing
+// spaces and the special character to gather the contents of the return string including trailing spaces of the
+// command output.
+std::string run_string_cmd(mp::SSHSession& session, std::string&& cmd)
+{
+    std::string ret;
+    ret = run_cmd(session, "echo `" + cmd + "`-");
```
Saviq: One thing that irks me here is that if you try to `run_string_cmd` that has backticks (a subshell, I mean), this would break. At least, these should be `$( ... )`.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -99,13 +152,16 @@ void make_target_dir(mp::SSHSession& session, const std::string& root, const std
...
 
-    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
-                                 relative_target.substr(0, relative_target.find_first_of('/'))));
+    // Get the first directory of the relative path.
+    std::string::size_type first_slash = relative_target.find_first_of('/');
+    std::string first_dir =
+        (std::string::npos == first_slash) ? relative_target : relative_target.substr(0, first_slash);
+
+    run_cmd(session,
+            fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group, first_dir));
```
Saviq: Why this change? FWIW we try to avoid, what we call, "drive-by" changes. If you want to refactor something unrelated to the fix / feature you're working on, please file a separate PR.

Also just realized these should be numeric IDs, not strings, otherwise we need to cater for spaces in them.

tests/test_sshfsmount.cpp
```diff
@@ -80,9 +80,11 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-                if (cmd == next_expected_cmd->first)
+                if (cmd.find(next_expected_cmd->first) != std::string::npos)
                 {
-                    output = next_expected_cmd->second;
+                    // We add a character to the output because the function gathering the output of the executed
+                    // commands uses it to mark the end of the output.
+                    output = std::string{next_expected_cmd->second + '-'};
                     remaining = output.size();
                     ++next_expected_cmd;
                     invoked = true;
```
Saviq: I don't like this, we really want to test for equality. One more reason against the `echo` approach IMO.

The test now assumes the command is executed via `run_string_cmd`. What if one of the calls goes through `run_cmd` directly? And if it's not meant to, why have `run_cmd` and `run_string_cmd` separate?

tests/test_sshfsmount.cpp
```diff
@@ -95,9 +97,9 @@ struct SshfsMount : public mp::test::SftpServerTest
                     // command vector.
                     for (auto it = std::next(next_expected_cmd); it != commands.end(); ++it)
                     {
-                        if (cmd == it->first)
+                        if (cmd.find(it->first) != std::string::npos)
                         {
-                            output = it->second;
+                            output = std::string{it->second + '-'}; // Same as above...
```
Saviq: Same as above indeed ;)

src/sshfs_mount/sshfs_mount.cpp
Saviq: Just realized there's a problem with this - `getent passwd 1000` works, too, but I'm unsure if `multipass mount . instance:~1000` should work (I mean, getent works with UIDs, not only usernames, but `~` in a shell does not).

src/sshfs_mount/sshfs_mount.cpp
ricab: Hmm, I am uncertain. It is not very complex, and if valid homes can have trailing spaces, I don't think we should remove them. But I also don't understand why we would remove *any* space. 

I would like to know what @luis4a0 has to say about it. What is the motivation for trimming in the first place? This seems to deal with output with an unknown amount of space padding, but what is the actual use case for that? Why would we be getting unexpected spaces from the commands we issue? I mean, when we parse output, we need to be able to assume some format, right? So why aren't spaces accounted for in that format? Some quirk of different shells? Or just being careful, just in case?

Also, please remind me, why parse bits of the path ourselves? Why not just leave the whole thing to the shell? IOW, isn't this just a particular case of the earlier issue, where we could just do `readlink -f input`? It requires the instance to be running, but so does this, right?

Saviq: > But I also don't understand why we would remove any space.

It's kinda the default when dealing with shell output, because you will always get a trailing newline that's not part of the _interesting_ output. But yeah that's what I meant by suggesting to not trim it in the first place, or rather - only trimming the "expected" trailing character.

We (`multipass exec`) actually seem to be dealing better with this than `ssh`:
```console
$ echo "foo " | hexdump 
0000000 6f66 206f 000a                         
0000005
$ echo -n "foo " | hexdump
0000000 6f66 206f                              
0000004
$ ssh localhost echo "foo " | hexdump  
0000000 6f66 0a6f                              
0000004
$ ssh localhost echo -n "foo " | hexdump  # notice SSH trims the space in the echoed string
0000000 6f66 006f                              
0000003
$ mp exec primary -- echo "foo " | hexdump
0000000 6f66 206f 000a                         
0000005
$ mp exec primary -- echo -n "foo " | hexdump
0000000 6f66 206f                              
0000004
```

> Also, please remind me, why parse bits of the path ourselves?

The only other way to expand `~` would be evaluating it in the shell… While it should generally be OK, since users "have root" on the instances anyway, so they don't _need_ to do "BASH injection" attacks, it felt cleaner to attack this through `getent`. We can error out a bit better.
The adverse effect is that `multipass mount . instance:~1000/path` does work, when it shouldn't…

src/sshfs_mount/sshfs_mount.cpp
Saviq: Looking at [`man adduser.conf`](http://manpages.ubuntu.com/manpages/trusty/man5/adduser.conf.5.html), usernames by default need to begin with a `[a-z]`.

Maybe even more importantly, in [`man getent`](http://manpages.ubuntu.com/manpages/xenial/en/man1/getent.1.html):
> pass each numeric key to getpwuid(3) and each nonnumeric key to getpwnam(3)

So there's three routes that I can see:
1. ignore this, and support UIDs as well as usernames, and let `getent` deal with it
1. only accept non-numeric usernames and treat e.g. "~1000" verbatim
1. accept the "unsafe" option of letting the shell evaluate the whole string

Only other option would be to somehow call [`getpwnam()`](http://manpages.ubuntu.com/manpages/eoan/en/man3/getpwnam.3.html) explicitly inside the instance…

src/sshfs_mount/sshfs_mount.cpp
Saviq: #1443 deals with the numer IDs issue.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: My intention here was to make reading output from commands more robust. But it's true that it adds complexity to the implementation for very small improvement, for a corner case, at the cost of adding code which is not so easy to maintain. I agree that forgetting about this improvement will simplify things a lot.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Yes, it can be done more concisely and elegantly.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: No reason :)
From the pdf you sent, spaces are not allowed.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: ok

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right

src/sshfs_mount/sshfs_mount.cpp
luis4a0: It's a good thing to avoid drive-by changes. I'll forget about this change, which is targeted at making the function more robust in case `relative_target` does not have any slashes.

tests/test_sshfsmount.cpp
luis4a0: Note I introduced the == in a previous PR, but now the `.find()` is needed. I never liked how the tests looked after introducing `run_string_cmd` :)

After reading _all_ yout comments, I see `run_string_cmd` pollutes the code, and introduces these glitches in the tests.

tests/test_sshfsmount.cpp
luis4a0: ack

src/sshfs_mount/sshfs_mount.cpp
luis4a0: I didn't think about numeric ID's.

My thought here is that it is safe to use option 1. We can make some check before calling `getent`, to avoid wrong strings. Then `getent` deals with it, and errors back if the user does not exist. Option 2 is not good I think, because using option 1 the user can still mount on a path containing `~`. Option 3 sounds to me unsafer than option 1.

luis4a0:
Thanks for your comments. I think that we can forget about trailing spaces in usernames and paths. Then we can consider this again if some use case appears where trailing spaces become more than a corner case.

src/sshfs_mount/sshfs_mount.cpp
Saviq: The problem is that if you _do_ have a user "1000:1001:1001…" in `/etc/passwd` (just add manually), that will not be selected over the user with UID 1000…

Saviq: > spaces are not allowed.

…by default :)

src/sshfs_mount/sshfs_mount.cpp
townsend2010: I agree this adds quite a bit of complexity for a very small corner case.  I'm +1 on removing this.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Oh, right, we were discussing corner cases :P
So we don't have a best option. In this case, 3 would be the best because the shell will resolve things that we can't do without much job. But at the cost of loosing a bit of safety, we should perform some checks there. But then, again, we'll have to make some assumptions on the user name, i.e., it cannot contain ';' or something used for injecting commands.

luis4a0: Wait, thinking again here. We should use option 3. Multipass should behave as the shell would do. If we do `sudo adduser 1000 --force-badname`, then `getent passwd 1000` would yield `ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash` but `~1000` will evaluate to `/home/1000`. I think the user would expect Multipass to behave like the shell, no matter what getent does. We definitely should let the shell evaluate, and this would imply forgetting about numerical IDs for mounting.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Why not `echo ~`? It expands to ${HOME}, which is by default the value given by `getent`. Should the user change ${HOME}, I guess he would expect ~ to expand to the new value.

luis4a0:
Maybe late, I discovered that I was incorrectly mocking the server answers in the tests. While the ssh commands have always a final '\', the mock didn't. Adding the final newline to the mock answers eliminates the need for `mp::utils::trim_end`, because we know that there is only one extra character at the end of the return string.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Another thing: why not replacing all by `run_cmd(session, fmt::format("echo {}", target))`?

src/sshfs_mount/sshfs_mount.cpp
Saviq: If we go for bash expanding it, we don't need the "~" or "~user" check at all.

src/sshfs_mount/sshfs_mount.cpp
Saviq: `readlink` doesn't expand `~` itself, and if we go for `bash` expanding it, we don't need any echo, just remove the quotes from `P="{}"…` below.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right, it also simplifies the code

src/sshfs_mount/sshfs_mount.cpp
Saviq: Yeah, the argument against that is the "it's unsafe" consideration. But as I said before, they have `root` on the instance already…

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -161,8 +168,9 @@ void set_owner_for(mp::SSHSession& session, const std::string& root, const std::
     mp::utils::trim_end(vm_user);
     mp::utils::trim_end(vm_group);
 
-    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
-                                 relative_target.substr(0, relative_target.find_first_of('/'))));
+    if (!relative_target.empty())
+        run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
+                                     relative_target.substr(0, relative_target.find_first_of('/'))));
```
Saviq, line 173: I have a feeling calling this with an empty `relative_target` should be an error, not a NOOP…

tests/test_sshfsmount.cpp
```diff
@@ -113,7 +114,7 @@ struct SshfsMount : public mp::test::SftpServerTest
                     {
                         if (cmd == it->first)
                         {
-                            output = it->second;
+                            output = it->second + '\n';
```
Saviq, line 117: `default_cmds` are missing this treatment.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -122,27 +121,35 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
...
     {
-        absolute = complete_path.path();
+        if ('/' != target[0])
+        {
+            complete_path = "~/" + mp::utils::escape_all(target);
+        }
+        else
+        {
+            complete_path = mp::utils::escape_all(target);
+        }
```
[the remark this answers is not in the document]
luis4a0: Good, you're right.

About the `else if`, I think it will be even clearer with a `switch(target[0])`.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -122,27 +121,35 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
...
-            .trimmed();
+    std::string existing =
+        run_cmd(session, fmt::format("sudo /bin/bash -c 'P={}; while [ ! -d $P/ ]; do P=${{P%/*}}; done; echo $P/'",
+                                     complete_path));
+    if ('\n' == existing.back())
+        existing.pop_back();
+
+    std::string missing = run_cmd(session, fmt::format("echo {} | cut -c {}-", complete_path, existing.size() + 1));
+    if ('\n' == missing.back())
+        missing.pop_back();
```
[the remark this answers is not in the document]
luis4a0: Yes, that's indeed right. But the check is there because of the tests. Thinking a bit more, it's worth to pay a bit in the test code to make the code clearer here. (This check happens in a couple of places.)

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -122,27 +121,35 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
...
 
-    return {existing.toStdString(), QDir(existing).relativeFilePath(absolute).toStdString()};
+    return {existing, missing};
```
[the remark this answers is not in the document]
luis4a0: Right. But note in https://doc.qt.io/qt-5/qdir.html#isRelative that 
> Under Unix a path is relative if it does not start with a "/"
so I maintain my preference for `switch(target[0])` in place of `if` and `else if`.

I normally prefer to try a `case` block instead of nested `if`'s. That said, are there any reasons I'm missing supporting your approach over mine?

src/sshfs_mount/sshfs_mount.cpp
luis4a0: You are right, and it is this way because of the tests. Again, we can remove it here and add some lines in the tests.

tests/test_sshfsmount.cpp
luis4a0: Right, I'll investigate this. Maybe we should explicitly add `\n` in the mock returns because the real thing always answers with a newline at the end.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
Saviq: Sure, make it a `switch`.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
Saviq: That, or maybe have a `trim_end` utility function that asserts it is actually a newline before popping off the back…

I would not like a bare `.pop_back()` here any more than this verbose approach… GAH.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
Saviq: +1

src/sshfs_mount/sshfs_mount.cpp
Saviq: I would rather, yes - and if anything, we should assert here it's not empty…

tests/test_sshfsmount.cpp
Saviq: Yup, either the commands should have explicit '\n', or `make_channel_read_return` should add it there.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
luis4a0: What disturbs me from creating an utility function or checking as I did is that the command will not end in newline in two cases: (i) it is a test, or (ii) there is an error somewhere. I can add a function for it because it is safer, but I vote also for adding the explicit `\n` at the end of each mock anster.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right, assertion seems to be the correct thing to do.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
Saviq: Sure, which is why I suggested to actually `assert` that the trailing character _is_ a newline, it would mean the test would assert in Debug build mode (in CI), and fail if assertions are compiled out. And yes, all mocked answers should have a newline one way (explicit) or another (in `make_channel_read_return`).

tests/test_sshfsmount.cpp
```diff
@@ -93,43 +93,77 @@ struct SshfsMount : public mp::test::SftpServerTest
...
                 output = it->second;
                 remaining = output.size();
                 invoked = true;
             }
+            else
+            {
+                output = "\n";
+                remaining = output.size();
+                invoked = true;
+            }
```
Saviq, line 137: How about making `make_channel_read_return` always append / return a newline, rather than being explicit about it in all the return strings?

@townsend2010 could use your opinion here… I could go with either…

tests/test_sshfsmount.cpp
```diff
@@ -152,6 +186,32 @@ struct SshfsMount : public mp::test::SftpServerTest
...
 
+    // Check that a given command is invoked and that it throws when failing.
+    void test_failed_invocation(const std::string& fail_cmd)
+    {
+        CommandVector commands = {
+            {"echo $PWD/target", "/home/ubuntu/target\n"},
+            {"sudo /bin/bash -c 'P=/home/ubuntu/target; while [ ! -d $P/ ]; do P=${P%/*}; done; echo $P/'",
+             "/home/ubuntu/\n"},
+            {"id -u", "1000\n"},
+            {"id -g", "1000\n"}};
```
Saviq, line 197: Shouldn't these go into `default_cmds`? In this test, you don't want to verify that the command ordering is correct, but rather that the failure of `fail_cmd` is bubbled up correctly.

tests/test_sshfsmount.cpp
luis4a0: That will certainly make the test a bit clearer in the sense that we won't have those `\n` in every answer. We can do it.
In case we need an answer without the final newline, we can just write a return function by hand.

tests/test_sshfsmount.cpp
luis4a0: Right.

include/multipass/utils.h
```diff
@@ -72,6 +72,7 @@ std::string run_cmd_for_output(const QString& cmd, const QStringList& args, cons
 std::string& trim_end(std::string& s);
 std::string& trim_newline(std::string& s);
 std::string escape_char(const std::string& s, char c);
+std::string escape_all(const std::string& s);
```
Saviq, line 75: Should this be more explicit about what it does? `escape_for_shell` or some such?

Saviq, line 75: Should this be more explicit about what it does? `escape_for_shell` or something?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -122,26 +124,30 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
...
     {
-        std::string home = run_cmd(session, "pwd");
-        mp::utils::trim_newline(home);
-        absolute = QString::fromStdString(home) + '/' + complete_path.path();
-    }
-    else
-    {
-        absolute = complete_path.path();
+    case '~':
+        absolute = run_cmd(session, fmt::format("echo {}", target));
```
Saviq, line 132: We need to escape here already, no?

include/multipass/utils.h
luis4a0: good!

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Absolutely, here and in the `default:` case. And we need to keep the escape later, because the shell answers with unescaped characters.

Another detail is that `target[0]` must not be escaped when it equals `~`, so the shell expands it.

luis4a0:
Also added coverage for the `~` test.

tests/test_sshfsmount.cpp
```diff
@@ -299,57 +299,71 @@ TEST_P(SshfsMountExecuteThrowRuntErr, test_runtime_error_when_executing)
...
+    {"sudo /bin/bash -c 'P=/home/ubuntu/target; while [ ! -d $P/ ]; do P=${P%/*}; done; echo $P/'", "/home/ubuntu/\n"},
     {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", "\n"},
     {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R 1000:1000 \"target\"'", "\n"}};
 
 // Commands to check that it works for a nonexisting path.
 CommandVector nonexisting_path_cmds = {
-    {"sudo /bin/bash -c 'P=\"/nonexisting/path\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'", "/\n"},
+    {"sudo /bin/bash -c 'P=/nonexisting/path; while [ ! -d $P/ ]; do P=${P%/*}; done; echo $P/'", "/\n"},
     {"sudo /bin/bash -c 'cd \"/\" && mkdir -p \"nonexisting/path\"'", "\n"},
     {"sudo /bin/bash -c 'cd \"/\" && chown -R 1000:1000 \"nonexisting\"'", "\n"}};
```
Saviq, line 352: You don't need to have the `mkdir` and `chown` lines everywhere - it's probably enough that just one test verifies those, and the rest work with the defaults.

What you're not testing for, though, is the escaping. We need at least one test case that verifies the escaping is done properly.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -122,26 +124,31 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
...
+        mp::utils::trim_newline(absolute);
+        break;
     }
 
-    std::string existing = run_cmd(
-        session,
-        fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'", absolute));
+    std::string existing =
+        run_cmd(session, fmt::format("sudo /bin/bash -c 'P={}; while [ ! -d $P/ ]; do P=${{P%/*}}; done; echo $P/'",
+                                     mp::utils::escape_for_shell(absolute)));
```
Saviq, line 147: We do need the quotes around all the instances of `$P`, as it will be unescaped.

tests/test_sshfsmount.cpp
```diff
@@ -204,13 +204,13 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-        {"pwd", "/home/ubuntu\n"},
-        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+        {"echo $PWD/target", "/home/ubuntu/target\n"},
+        {"sudo /bin/bash -c 'P=/home/ubuntu/target; while [ ! -d $P/ ]; do P=${P%/*}; done; echo $P/'",
          "/home/ubuntu/\n"},
         {"id -u", "1000\n"},
         {"id -g", "1000\n"},
         {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -o slave -o transform_symlinks -o allow_other :\"source\" "
-         "\"target\"",
+         "\"/home/ubuntu/target\"",
```
Saviq, line 213: It's not clear to me why you needed this change?

tests/test_sshfsmount.cpp
luis4a0: OK, I don't have a strong opinion here. I think testing those doesn't hurt, but you're right in the sense that these lines are almost trivial and will work iff the first one does.

I left the first `mkdir` and `chown`'s.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Oh, right... there is no ~ there, so we can safely return to using quotes.

I also added a test for escaping.

tests/test_sshfsmount.cpp
luis4a0: Oops, that survived from some past change :(

I reverted it.

luis4a0:
Yeah, we'd need a `run_cmd` which escapes. I'll keep this in mind.

luis4a0:
It's definitely a bug, but NOT a regression. With multipass 1.1.0, one obtains exactly the same output.

I propose to report this as a separate issue. We must also run a script testing all the escaped characters (which are a lot, note that almost all the ASCII characters must be escaped to work with the shell).

Saviq:
OK you're right. Tilde handling here works fine.

bors r+

PR #1441, opened 2020-03-23
[settings] Accept const ref in get helper

ricab:
Split into separate PR on review request.

### DIFF
--- a/src/utils/settings.cpp
+++ b/src/utils/settings.cpp
@@ -97,7 +97,7 @@ void check_status(const QSettings& settings, const QString& attempted_operation)
                                      : QStringLiteral("access error (consider running with an administrative role)")};
 }
 
-QString checked_get(QSettings& settings, const QString& key, const QString& fallback, std::mutex& mutex)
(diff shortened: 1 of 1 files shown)


ricab:
@Saviq Hey, here is the separate PR for the const ref. I'll rebase the other PR on this.

ricab:
Actually, since this was the first commit in the winterm branch, I could just make the branch point to it.

Saviq:
@ricab should we have a test accompany this change?

ricab:
@Saviq as before, testing this would require mocking `QSettings`. We have `MockSettings`, but that mocks our own `Settings` class to test other things in multipass (get and set cmds, etc.). Since `QSettings` is outside of our control, I'd need an additional level of indirection. If you want, I can try to do that now?

Saviq:
No, it's fine. I actually thought of a compile-time check, but the actual code will do it instead…

bors r+

Saviq:
Hmm?

bors r+

PR #1443, opened 2020-03-24
[sshfs] use numeric IDs in set_owner_for(...)

Saviq:


Saviq:
Sorry @townsend2010, there's a bit more to it (that will be simplified with your upcoming changes, actually).

Saviq:
Let's see if bors is better, too…

bors try

PR #1445, opened 2020-03-25
Windows Terminal profile for primary - take2

ricab:
New approach, following change of plans requested in #1427. 

The setting is stored and retrieved in a regular fashion, but has platform-dependent side effects upon switching.


ricab:
Superseded by #1451

Issue #1449, opened 2020-03-27
Support adding a primary profile to Windows Terminal


ricab:
The user will be able to (de)activate this via a new setting, with the key: ~~`client.windows-terminal.add-profiles`~~ `client.apps.windows-terminal.profiles` (see [this](https://github.com/canonical/multipass/issues/756#issuecomment-621127639)):
- `client.apps.windows-terminal.profiles=primary` to enable
- `client.apps.windows-terminal.profiles=none` to disable

We'll interface with Windows Terminal via its configuration file: `profiles.json`. We'll look for it in its expected location, equating its presence to whether Windows Terminal is installed. We'll then try to add/enable or hide the primary profile according to the setting.

However, the success of the set operations above will not depend on this file. Instead, the multipass CLI (GUI?) client will keep trying to honor the current setting whenever it quits, logging errors or warnings when it can't (including right after the set operation):
  - a warning will be logged if the setting is enabled but the file isn't found
  - if the file exists but is unreadable or has bad contents...
    * ... an INFO will be logged if the state is "none"
    * ... an ERROR will be logged otherwise.

As the procedure to enforce the setting itself, the error handling above will be independent of what command is executed (right?).

This scheme can be extended in the future to have profiles for all instances, with a value "all".

This setting shall only be available on Windows.

ricab:
After many discussions, this describes what I understand is the design goal. @Saviq, could you please confirm if this matches what you have in mind? 

Also, WDYT about the points with question marks? 

In particular, on the question of when to sync. Notice that:
  - on quit &rarr; fits nicely into all commands, so nothing else needed for set, but state updating delayed for longer
  - on start &rarr; special handling needed in set (and launch, etc. in the future)
  - both &rarr; duplicated warnings/errors

And on the question of the errors/warnings being the same everywhere, that would allow us to just log in place. Otherwise, if you want different handling in different commands, we'd have to throw specific exceptions for the different cases and handle them in the commands.

Saviq:
> an error will be logged when the file exists but is unreadable or has bad contents, irrespective of the current setting.

I'm on the fence whether we should suppress that when the setting is "none". I'm thinking of the scenario when Terminal changes its schema in an incompatible way. Disabling our integration should suppress the error I'm thinking. So if `none`, no errors / warnings logged. At least we should reduce log severity, so it doesn't show up by default.

> As the procedure to enforce the setting itself, the error handling above will be independent of what command is executed (right?).

[...]

PR #1450, opened 2020-03-27
Winterm integration take3

ricab:
Yet another approach. Addresses #1449 . 

The value is checked upon `set`, but applied separately.


ricab:
Sorry for spam, forgot to make draft.

PR #1451, opened 2020-03-27
Windows Terminal integration

ricab:
Yet another approach. Addresses #1449 .

The setting is checked upon set, but applied separately.

Depends on #1467. Diff against that: https://github.com/canonical/multipass/compare/mockable-standard-paths...winterm-integration-take3

### DIFF
--- a/include/multipass/cli/client_common.h
+++ b/include/multipass/cli/client_common.h
@@ -53,6 +53,7 @@ std::unique_ptr<SSLCertProvider> get_cert_provider();
 void set_logger();
 void set_logger(multipass::logging::Level verbosity); // full param qualification makes sure msvc is happy
 void preliminary_setup();
+void final_adjustments();
 }
 } // namespace multipass
 #endif // MULTIPASS_CLIENT_COMMON_H
--- a/include/multipass/cli/client_common.h
+++ b/include/multipass/cli/client_common.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2019-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/include/multipass/cli/client_common.h
+++ b/include/multipass/cli/client_common.h
@@ -52,8 +52,8 @@ std::string get_server_address();
 std::unique_ptr<SSLCertProvider> get_cert_provider();
 void set_logger();
 void set_logger(multipass::logging::Level verbosity); // full param qualification makes sure msvc is happy
-void preliminary_setup();
-void final_adjustments();
+void pre_setup();
+void post_setup();
 }
 } // namespace multipass
 #endif // MULTIPASS_CLIENT_COMMON_H
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -39,6 +39,7 @@ namespace platform
 std::map<QString, QString> extra_settings_defaults();
 
 QString interpret_winterm_integration(const QString& val);
+void sync_winterm_profiles();
 
 QString autostart_test_data(); // returns a platform-specific string, for testing purposes
 void setup_gui_autostart_prerequisites();
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -36,6 +36,7 @@ namespace multipass
 {
 namespace platform
 {
+QString interpret_winterm_integration(const QString& val);
 QString autostart_test_data(); // returns a platform-specific string, for testing purposes
 void setup_gui_autostart_prerequisites();
 std::string default_server_address();
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -36,14 +36,21 @@ namespace multipass
 {
 namespace platform
 {
+std::map<QString, QString> extra_settings_defaults();
+
 QString interpret_winterm_integration(const QString& val);
+
 QString autostart_test_data(); // returns a platform-specific string, for testing purposes
 void setup_gui_autostart_prerequisites();
+
 std::string default_server_address();
 QString default_driver();
+
 QString daemon_config_home();                      // temporary
+
 bool is_backend_supported(const QString& backend); // temporary
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
+
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
 std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& config);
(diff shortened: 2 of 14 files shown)


include/multipass/cli/client_common.h
```diff
@@ -53,6 +53,7 @@ std::unique_ptr<SSLCertProvider> get_cert_provider();
 void set_logger();
 void set_logger(multipass::logging::Level verbosity); // full param qualification makes sure msvc is happy
 void preliminary_setup();
+void final_adjustments();
```
Saviq, line 56: This one sounds weird :)

Something like `post_setup`, `finally`, `before_exit`… "adjustments" say a bit too much about what is supposed to happen inside…

include/multipass/platform.h
```diff
@@ -36,13 +36,22 @@ namespace multipass
 {
 namespace platform
 {
+std::map<QString, QString> extra_settings_defaults();
+
+QString interpret_winterm_integration(const QString& val);
+void sync_winterm_profiles();
```
Saviq, line 42: This interface should be agnostic to the fact that a platform may decide to do Windows Terminal integration.

```suggestion
QString interpret_setting(const QString& key, const QString& val);
void perform_final_adjustments();
```

include/multipass/cli/client_common.h
ricab: OK, sure, I'll change that.

include/multipass/platform.h
ricab: Agreed, that makes it more general.

include/multipass/cli/client_common.h
```diff
@@ -53,6 +53,7 @@ std::unique_ptr<SSLCertProvider> get_cert_provider();
 void set_logger();
 void set_logger(multipass::logging::Level verbosity); // full param qualification makes sure msvc is happy
 void preliminary_setup();
+void concluding_setup();
```
Saviq, line 56: Having slept on it a bit… how about this?
```suggestion
void final_setup();
```

include/multipass/platform.h
```diff
@@ -36,13 +36,22 @@ namespace multipass
 {
 namespace platform
 {
+std::map<QString, QString> extra_settings_defaults();
+
+QString interpret_setting(const QString& key, const QString& val);
+void sync_winterm_profiles();
```
Saviq, line 42: This should also be `client_final_setup` or so?

ricab: I think it is preferable to be more specific about the operations being provided at this level. That leaves it to the CLI client (or any other calling code) to decide when to call. And it increases modularity, allowing client code to call this bit without that bit, and it is more coherent with the current granularity: we have things like `make_logger`, `make_update_prompt`, `setup_gui_autostart_prerequisites`, rather than just `client_setup`. The platform code is concerned with the implementation of specific operations in different platforms, the client code is concerned with organizing those operations.

include/multipass/cli/client_common.h
ricab: I thought concluding was more precise that final: this setup concludes execution, rather than being the last setup. As suggested in IRC, another option could be `prefix_setup` and `postfix_setup`. A bit more change, but probably makes the symmetry clearer for the reader. Let me know.

ricab: Going with `pre_setup` and `post_setup` (as requested in IRC).

ricab: Done

include/multipass/platform.h
Saviq: Yeah, but suddenly you have winterm-specific code in all platforms, even if it's just NOOP.

If you look at https://github.com/canonical/multipass/issues/756#issuecomment-621127639 - our longer-term plan for "integrations" - to follow what you're doing, you'd need an explicit call to each of the enabled integrations here.

IMO the client should have entry points for the integrations (pre_, post_setup, modify_request), but it should / need not have the context to decide whether to run them or not.

ricab: So you envision more "syncs" needed for other app integrations? I am not sure what the best approach will be for them, but I would prefer to decide separately. Creating general entry points here means assuming, in this PR, future syncs at specific points. Instead, we could generalize this then, as needed, when designing/implementing those features.

Also, the way I see it, things like pre/post make more sense at the client level. The implication of chronology or precedence is meaningful in the CLI as relative to a sequence of tasks being carried out there. Less so in the GUI (we could think of them as relative to the main loop, but that would be too frequent a period). So I guess that'd be something to improve when implementing the WT sync on the GUI side. At the platform level, however, there is no inherent sequence to speak of, so introducing it there feels like kind of polluting it somewhat.

ricab: How about:

1) In this PR, we rename this as `app_integration_sync` and only call the winterm sync on windows, (separate function residing in the unnamed namespace)
2) In future PRs, we refine as needed

:grey_question:

Saviq: (not sure you get notified about the :+1: above, so…)

:+1:

ricab: Hehe, cool! :ok_hand:

ricab: And yes, I was only notified of the comment.

include/multipass/cli/client_common.h
ricab: done

include/multipass/platform.h
ricab: done

include/multipass/platform.h
ricab: For (my own) future reference: this would pose problems with testing, unless we kept the granularity at a separate level. After discussing on IRC, we decided such refactoring now would probably be thrown away soon and not worth it. In the future, we should avoid providing NOOP syncs at the platform level, for instance, with some plugin structure providing hooks for different sync "moments".

tests/mock_logger.h
```diff
@@ -32,7 +36,32 @@ class MockLogger : public multipass::logging::Logger
     MockLogger() = default;
     MOCK_CONST_METHOD3(log, void(multipass::logging::Level level, multipass::logging::CString category,
                                  multipass::logging::CString message));
+
+    template <typename Matcher>
+    static auto make_cstring_matcher(const Matcher& matcher)
+    {
+        return Property(&multipass::logging::CString::c_str, matcher);
+    }
```
Saviq, line 44: There's another one of those here:
https://github.com/canonical/multipass/blob/dfaa8492602c96a099d98225baec0845fe0d1f0c/tests/mock_logger.h#L40-L44

ricab: Good catch. Fixed.

Saviq:
Yeah, it's working nicely :)

bors merge

ricab:
bors retry

Saviq:
@ricab something's wrong on macOS:
```
Undefined symbols for architecture x86_64:
  "multipass::platform::interpret_setting(QString const&, QString const&)", referenced from:
      multipass::Settings::set_aux(QString const&, QString) in libutils.a(settings.cpp.o)
  "multipass::platform::sync_winterm_profiles()", referenced from:
      multipass::client::post_setup() in libclient_common.a(client_common.cpp.o)
  "multipass::platform::extra_settings_defaults()", referenced from:
      (anonymous namespace)::make_defaults() in libutils.a(settings.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

ricab:
Darn :/ That's not in the travis build anywhere, is it? I am looking into it.

ricab:
@Saviq it built fine here (I have only some linker warnings on 3rd-party). Where did that error show up?

ricab:
Ah in the bors build, I found it.  

I'll be back

ricab:
For the record, this is just because staging builds don't pull the private counterpart.

PR #1454, opened 2020-03-30
[snap] skip RelWithDebInfo if we don't have secure vars

Saviq:
The tests get run in Coverage, and RelWithDebInfo tries to publish after a successful build, for which we don't have credentials.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,10 @@ jobs:
     - <<: *snapcraft-build
       name: RelWithDebInfo
       env: BUILD_TYPE=RelWithDebInfo
+      if: &secure_condition |
+        repo = canonical/multipass
+               AND (type != pull_request
+                    OR head_repo = repo)
 
       addons:
         snaps:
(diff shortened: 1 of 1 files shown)


ricab:
Hey, just had a look at the code, but could you explain what is going on in there a bit? Thank you.

Saviq:
@ricab I've modified this to be a simpler change - but to explain:

We tried to decrypt our credentials, which is only possible when the PR comes from the same repository. That's why #1420 failed - so this change disables that one job altogether. We run the tests in the Coverage job anyway, and bors will then run through the whole suite on try / merge.

Saviq:
[This](https://travis-ci.org/github/canonical/multipass/builds/668683526) is wrong… it should have ran RelWithDebInfo, too… digging.

ricab:
Indeed, weird. I guess the var is not being established.

Saviq:
bors try

Saviq:
bors try

Please.

Saviq:
bors try

PR #1467, opened 2020-04-03
Mockable standard paths

ricab:
Adds a mockable `StandardPaths` singleton, as a thin wrapper around `QStandardPaths`. Implements the corresponding mock with the same approach as `MockSettings`, that is, via:
1. a test environment, to run custom code during test setup and tear down
2. a test event listener, to verify and clear mock expectations on the singleton at the end of each test
3. default operations delegating on the parent

To achieve that, this PR also extracts common stuff into a `MockSingletonHelper`.

This approach enforces accounting of mock expectations in the respective test. It also avoids singleton state leaking from test to test and affecting the results based on execution order. Finally, it ensures that tests only deal with the mock. They use only parent operations that are deemed safe for testing, indirectly, via the mock itself.

### DIFF
--- a/include/multipass/standard_paths.h
+++ b/include/multipass/standard_paths.h
@@ -20,12 +20,23 @@
 
 #include "singleton.h"
 
+#include <QStandardPaths>
+
 namespace multipass
 {
 class StandardPaths : public Singleton<StandardPaths>
 {
 public:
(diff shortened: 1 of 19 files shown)


ricab:
Thanks @townsend2010 !

PR #1473, opened 2020-04-07
Refactor SSH tests

luis4a0:
1. refactor `make_exec_to_check_commands()`, using `std::find_if()`;
2. add `\n` to all answers of the sshfs mock, because the real server always returns the newline and we were missing this behavior in the mock;
3. enforce newline checks in sshfs server code; and
4. use parameterized tests.

### DIFF
--- a/src/sshfs_mount/sshfs_mount.cpp
+++ b/src/sshfs_mount/sshfs_mount.cpp
@@ -147,8 +147,7 @@ std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, cons
 // Create a directory on a given root folder.
 void make_target_dir(mp::SSHSession& session, const std::string& root, const std::string& relative_target)
 {
-    if (!relative_target.empty())
-        run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && mkdir -p \"{}\"'", root, relative_target));
+    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && mkdir -p \"{}\"'", root, relative_target));
 }
 
 // Set ownership of all directories on a path starting on a given root.
@@ -160,7 +159,7 @@ void set_owner_for(mp::SSHSession& session, const std::string& root, const std::
     mp::utils::trim_newline(vm_user);
     mp::utils::trim_newline(vm_group);
 
-    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
+    run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} \"{}\"'", root, vm_user, vm_group,
                                  relative_target.substr(0, relative_target.find_first_of('/'))));
 }
 
@@ -177,8 +176,11 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
 
     // We need to create the part of the path which does not still exist,
     // and set then the correct ownership.
-    make_target_dir(session, leading, missing);
-    set_owner_for(session, leading, missing);
+    if (missing != ".")
+    {
+        make_target_dir(session, leading, missing);
+        set_owner_for(session, leading, missing);
+    }
 
     auto output = run_cmd(session, "id -u");
     mpl::log(mpl::Level::debug, category,
--- a/src/sshfs_mount/sshfs_mount.cpp
+++ b/src/sshfs_mount/sshfs_mount.cpp
@@ -127,22 +127,21 @@ std::pair<std::string, std::string> get_path_split(mp::SSHSession& session, cons
 
     if (complete_path.isRelative())
     {
-        QString home = QString::fromStdString(run_cmd(session, "pwd")).trimmed();
-        absolute = home + '/' + complete_path.path();
+        std::string home = run_cmd(session, "pwd");
+        mp::utils::trim_newline(home);
+        absolute = QString::fromStdString(home) + '/' + complete_path.path();
     }
     else
     {
         absolute = complete_path.path();
     }
 
-    QString existing =
-        QString::fromStdString(
-            run_cmd(session,
-                    fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'",
-                                absolute)))
-            .trimmed();
+    std::string existing = run_cmd(
+        session,
+        fmt::format("sudo /bin/bash -c 'P=\"{}\"; while [ ! -d \"$P/\" ]; do P=${{P%/*}}; done; echo $P/'", absolute));
+    mp::utils::trim_newline(existing);
 
-    return {existing.toStdString(), QDir(existing).relativeFilePath(absolute).toStdString()};
+    return {existing, QDir(QString::fromStdString(existing)).relativeFilePath(absolute).toStdString()};
 }
 
 // Create a directory on a given root folder.
@@ -158,8 +157,8 @@ void set_owner_for(mp::SSHSession& session, const std::string& root, const std::
 {
     auto vm_user = run_cmd(session, "id -u");
     auto vm_group = run_cmd(session, "id -g");
-    mp::utils::trim_end(vm_user);
-    mp::utils::trim_end(vm_group);
+    mp::utils::trim_newline(vm_user);
+    mp::utils::trim_newline(vm_group);
 
     run_cmd(session, fmt::format("sudo /bin/bash -c 'cd \"{}\" && chown -R {}:{} {}'", root, vm_user, vm_group,
                                  relative_target.substr(0, relative_target.find_first_of('/'))));
@@ -185,6 +184,7 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
     mpl::log(mpl::Level::debug, category,
              fmt::format("{}:{} {}(): `id -u` = {}", __FILE__, __LINE__, __FUNCTION__, output));
     auto default_uid = std::stoi(output);
+
     output = run_cmd(session, "id -g");
     mpl::log(mpl::Level::debug, category,
              fmt::format("{}:{} {}(): `id -g` = {}", __FILE__, __LINE__, __FUNCTION__, output));
--- a/tests/test_sshfsmount.cpp
+++ b/tests/test_sshfsmount.cpp
@@ -86,21 +86,22 @@ struct SshfsMount : public mp::test::SftpServerTest
     // cooperate better, i.e., make the reader read only when a command was issued.
     auto make_exec_to_check_commands(const CommandVector& commands, std::string::size_type& remaining,
                                      CommandVector::const_iterator& next_expected_cmd, std::string& output,
-                                     bool& invoked, mp::optional<std::string>& fail_cmd, mp::optional<bool>& fail_bool)
+                                     bool& invoked, mp::optional<std::string>& fail_cmd,
+                                     mp::optional<bool>& fail_invoked)
     {
-        *fail_bool = false;
+        *fail_invoked = false;
 
         auto request_exec = [this, &commands, &remaining, &next_expected_cmd, &output, &invoked, &fail_cmd,
-                             &fail_bool](ssh_channel, const char* raw_cmd) {
+                             &fail_invoked](ssh_channel, const char* raw_cmd) {
             invoked = false;
 
             std::string cmd{raw_cmd};
 
             if (fail_cmd && cmd.find(*fail_cmd) != std::string::npos)
             {
-                if (fail_bool)
+                if (fail_invoked)
                 {
-                    *fail_bool = true;
+                    *fail_invoked = true;
                 }
                 exit_status_mock.return_exit_code(SSH_ERROR);
             }
@@ -165,7 +166,7 @@ struct SshfsMount : public mp::test::SftpServerTest
 
     void test_command_execution(const CommandVector& commands, mp::optional<std::string> target = mp::nullopt,
                                 mp::optional<std::string> fail_cmd = mp::nullopt,
-                                mp::optional<bool> fail_bool = mp::nullopt)
+                                mp::optional<bool> fail_invoked = mp::nullopt)
     {
         bool invoked{false};
         std::string output;
@@ -175,8 +176,8 @@ struct SshfsMount : public mp::test::SftpServerTest
         auto channel_read = make_channel_read_return(output, remaining, invoked);
         REPLACE(ssh_channel_read_timeout, channel_read);
 
-        auto request_exec =
-            make_exec_to_check_commands(commands, remaining, next_expected_cmd, output, invoked, fail_cmd, fail_bool);
+        auto request_exec = make_exec_to_check_commands(commands, remaining, next_expected_cmd, output, invoked,
+                                                        fail_cmd, fail_invoked);
         REPLACE(ssh_channel_request_exec, request_exec);
 
         make_sshfsmount(target.value_or(default_target));
@@ -224,7 +225,7 @@ struct SshfsMountExecute : public SshfsMount, public testing::WithParamInterface
 {
 };
 
-struct SshfsMountExecuteAndFail
+struct SshfsMountExecuteAndNoFail
     : public SshfsMount,
       public testing::WithParamInterface<std::tuple<std::string, CommandVector, std::string>>
 {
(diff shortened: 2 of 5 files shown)


src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -182,10 +181,13 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
     set_owner_for(session, leading, missing);
 
     auto output = run_cmd(session, "id -u");
+    mp::utils::trim_newline(output);
```
Saviq, line 184: This isn't necessary, `std::stoi` deals with whitespace on its own.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -182,10 +181,13 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
...
     mpl::log(mpl::Level::debug, category,
              fmt::format("{}:{} {}(): `id -u` = {}", __FILE__, __LINE__, __FUNCTION__, output));
     auto default_uid = std::stoi(output);
+
     output = run_cmd(session, "id -g");
+    mp::utils::trim_newline(output);
```
Saviq, line 190: Here, too.

tests/test_sshfsmount.cpp
```diff
@@ -93,43 +94,93 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+        };
+
+        return request_exec;
+    }
+
+    // Mock that fails after executing some commands.
+    auto make_exec_that_executes_and_fails(const CommandVector& commands, const std::string& fail_cmd,
+                                           std::string::size_type& remaining,
+                                           CommandVector::const_iterator& next_expected_cmd, std::string& output,
+                                           bool& invoked_cmd, bool& invoked_fail)
```
Saviq, line 150: Do we really need `commands` here? If `default_cmds` had all the commands necessary to go through successfully, and we'd fail on `fail_cmd`, this would get simplified a lot.

tests/test_sshfsmount.cpp
```diff
@@ -93,43 +94,93 @@ struct SshfsMount : public mp::test::SftpServerTest
...
                 output = it->second;
                 remaining = output.size();
                 invoked = true;
             }
+            else
+            {
+                output = "\n";
+                remaining = output.size();
+                invoked = true;
+            }
```
Saviq, line 138: I think we should get rid of that case. If we didn't match in either `commands` or `default_cmds`, the output should be empty.

tests/test_sshfsmount.cpp
```diff
@@ -188,161 +239,152 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+TEST_P(SshfsMountFail, test_failed_invocation)
 {
-    bool invoked{false};
-    auto request_exec = make_exec_that_fails_for({"chown"}, invoked);
+    CommandVector commands = {
+        {"snap run multipass-sshfs.env", "LD_LIBRARY_PATH=/foo/bar\nSNAP=/baz\n"},
+        {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 3.0.0\n"},
+        {"pwd", "/home/ubuntu\n"},
+        {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+         "/home/ubuntu/\n"}};
```
Saviq, line 280: These should all go into `default_cmds` - we don't want `make_exec_that_executes_and_fails` to be checking the command ordering, do we?

I mean, _if_ we had a case where one command failing would just mean a different branch being taken, we'd need for each command in this vector to have a "fails" boolean member. But we don't have a case like that today.

Saviq:
Also, I think `invalid_fuse_version_throws` could make its way into the parameterized test, if the exception thrown was another param.

And unblocks_when_sftpserver_exits could make use of `test_command_execution` with an empty `commands` vector?

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Right. In fact, I put there to check for the newline, but I agree it's not needed and the name of the function can be misleading in the sense that it does not specify that it checks for `\n`. I agree to remove it.

src/sshfs_mount/sshfs_mount.cpp
luis4a0: Also fixed.

tests/test_sshfsmount.cpp
luis4a0: Oh, you're right!

tests/test_sshfsmount.cpp
luis4a0: Right, it would mimic better the real server.

tests/test_sshfsmount.cpp
luis4a0: Right. In any case, if we really want to trace the execution of commands, we won't use `default_cmds`, but a `CommandVector` containing at least the ordering. Or, we can think on a better data structure to hold booleans or whatever for each command being executed. In any case, not using the default set of commands in `default_cmds`.

luis4a0:
Two good insights, I implemented both.

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +219,126 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+//
+
+INSTANTIATE_TEST_SUITE_P(SshfsMountThrowWhenError, SshfsMountFail,
+                         testing::Values("mkdir", "chown", "id -u", "id -g", "cd", "pwd"));
+
+// Commands to check that a version of FUSE smaller that 3 gives a correct answer.
+CommandVector fuse_cmds = {{"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 2.9.0\n"},
+                           {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -o slave -o transform_symlinks -o "
+                            "allow_other -o nonempty :\"source\" \"target\"",
+                            "don't care\n"}};
```
Saviq, line 297: We don't have an explicit test for when the version is 3 or higher, it is just tested via `default_cmds`. Please think whether we should add a version of the above with version: 3.0.0 and higher?

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +219,126 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+
+// Commands to check that the server correctly executes commands.
```
Saviq, line 299:
```suggestion
// Commands to check that the server correctly creates the mount target.
```

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +219,126 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+CommandVector exec_cmds = {
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+     "/home/ubuntu/\n"},
+    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", "\n"},
+    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R 1000:1000 target'", "\n"}};
+
+// Commands to check that the server works if an absolute path is given.
+CommandVector absolute_cmds = {
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+     "/home/ubuntu/\n"}};
```
Saviq, line 309: Here, too, I think like we're missing a couple checks. We should verify what happens on different return values of this command:
1. full path already existing
2. none of the path existing

Ideally we'd check that neither `mkdir` or `chown` are executed unless necessary. If you made the change to add an optional `fail_cmd` I suggested above, you could `ASSERT_NO_THROW` with a `CommandVector{…}` returning the value of `target` and failing on either `mkdir` or `chown`.

tests/test_sshfsmount.cpp
```diff
@@ -137,6 +138,36 @@ struct SshfsMount : public mp::test::SftpServerTest
         return request_exec;
     }
 
+    // Mock that fails after executing some commands.
+    auto make_exec_that_executes_and_fails(const std::string& fail_cmd, std::string::size_type& remaining,
+                                           std::string& output, bool& invoked_cmd, bool& invoked_fail)
```
Saviq, line 143: Now I look at it again… Couldn't an `mp::optional<std::string> fail_cmd` be added to `make_exec_to_check_commands`?

tests/test_sshfsmount.cpp
```diff
@@ -93,36 +95,35 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+                    ADD_FAILURE() << "\"" << (found_cmd->first) << "\" executed out of order; expected \""
+                                  << next_expected_cmd->first << "\"";
+
+                    return SSH_OK;
                 }
             }
 
+            // If the command list was entirely checked or if the executed command is not on the list, check the
+            // default commands to see if there is an answer to the current command. If not, answer with a
+            // newline, because all strings returned by the server we are mocking end in newline.
```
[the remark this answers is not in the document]
luis4a0: Oops, forgot the remove the comment on newline.

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +219,126 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-    auto request_exec = make_exec_that_fails_for({"mkdir"}, invoked);
-    REPLACE(ssh_channel_request_exec, request_exec);
-
-    EXPECT_THROW(make_sshfsmount(), std::runtime_error);
-    EXPECT_TRUE(invoked);
-}
+};
 
-TEST_F(SshfsMount, throws_when_unable_to_chown)
+// Mocks the server raising std::invalid_argument exception on the execution of given commands.
```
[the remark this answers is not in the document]
luis4a0: Changed

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +219,126 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-    auto channel_read = make_channel_read_return(output, remaining, invoked);
-    REPLACE(ssh_channel_read_timeout, channel_read);
-
-    EXPECT_THROW(make_sshfsmount(), std::invalid_argument);
-    EXPECT_TRUE(invoked);
+    test_command_execution(commands, target);
 }
 
-TEST_F(SshfsMount, throws_when_unable_to_obtain_gid)
+TEST_P(SshfsMountExecuteThrow, test_invalid_arg_when_executing)
```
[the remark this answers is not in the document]
luis4a0: Changed

tests/test_sshfsmount.cpp
luis4a0: Good catch. Added the test.

tests/test_sshfsmount.cpp
luis4a0: Changed.

tests/test_sshfsmount.cpp
luis4a0: That is a neat idea. In fact, I am just discovering that, when no directory must be created, `mkdir "."` and `chown -R 1000:1000 "."` are executed. At least in an old version of the code. That should be corrected when merging the `~` handling.

tests/test_sshfsmount.cpp
luis4a0: Of course :)

Saviq: I meant that this is basically `make_exec_to_check_commands` with empty `commands` now, so this can be dropped?

`make_exec_that_fails_for` is also just a special case of the above now? You could either make them both overloads, or drop completely :)

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +227,173 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+                                         std::make_pair("/nonexisting/path", nonexisting_path_cmds)));
+
+// Commands to test that when a mount path already exists, no mkdir nor chown is ran.
+CommandVector execute_no_mkdir_cmds = {
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+     "/home/ubuntu/target/\n"}};
+
+INSTANTIATE_TEST_SUITE_P(SshfsMountSuccessAndAvoidCommands, SshfsMountExecuteAndFail,
+                         testing::Values(std::make_tuple("target", execute_no_mkdir_cmds, "mkdir"),
+                                         std::make_tuple("target", execute_no_mkdir_cmds, "chown")));
```
Saviq, line 368: So this is the case that you said needs fixing? I don't think we should be adding a test that tests for wrong behaviour :)

It'd be good to fix in this PR, IMO.

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +227,173 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+CommandVector exec_cmds = {
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+     "/home/ubuntu/\n"},
+    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && mkdir -p \"target\"'", "\n"},
+    {"sudo /bin/bash -c 'cd \"/home/ubuntu/\" && chown -R 1000:1000 \"target\"'", "\n"}};
+
+// Commands to check that the server works if an absolute path is given.
+CommandVector absolute_cmds = {
+    {"sudo /bin/bash -c 'P=\"/home/ubuntu/target\"; while [ ! -d \"$P/\" ]; do P=${P%/*}; done; echo $P/'",
+     "/home/ubuntu/\n"}};
```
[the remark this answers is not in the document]
luis4a0: Right

tests/test_sshfsmount.cpp
luis4a0: Yes, `make_exec_that_fails` is a special case, I refactored it.

On fixing the `mkdir`, `chown` issue on this PR, yes, it makes sense to fix it here. In fact, the problem was that when no directory must be created, the 'missing' part of the path is `.`. This makes sense because it's the output of `QDir::relativeFilePath()`. I added so far a check in `src/sshfs_mount/sshfs_mount.cpp`. Now I moved it to the correct place: if missing is `.`, then no folder must be created nor changed its owner.

tests/test_sshfsmount.cpp
```diff
@@ -188,167 +202,186 @@ struct SshfsMount : public mp::test::SftpServerTest
...
-    EXPECT_TRUE(invoked);
-}
+// Mocks correct execution of a given vector of commands/answers. The first string specifies the parameter with
+// which make_sshfsmount mock is called.
+struct SshfsMountExecute : public SshfsMount, public testing::WithParamInterface<std::pair<std::string, CommandVector>>
+{
+};
 
-TEST_F(SshfsMount, throws_when_unable_to_make_target_dir)
+struct SshfsMountExecuteAndFail
```
Saviq, line 227:
```suggestion
struct SshfsMountExecuteAndNoFail
```
We actually want these to _not_ fail, right? Or not throw?

tests/test_sshfsmount.cpp
```diff
@@ -83,46 +86,54 @@ struct SshfsMount : public mp::test::SftpServerTest
...
 
-            if (next_expected_cmd != commands.end())
+            if (fail_cmd && cmd.find(*fail_cmd) != std::string::npos)
+            {
+                if (fail_bool)
+                {
+                    *fail_bool = true;
+                }
+                exit_status_mock.return_exit_code(SSH_ERROR);
+            }
```
Saviq, line 106: Use `fail_invoked` rather than `fail_bool` for consistency?

tests/test_sshfsmount.cpp
luis4a0: Oops, right.

tests/test_sshfsmount.cpp
luis4a0: ok

Saviq:
Yes. This makes it for a perfect test harness for the future, even if it doesn't change coverage straight up.

bors r+

luis4a0:
bors r=Saviq

Saviq:
Once more, with feeling…

bors merge

luis4a0:
bors retry

PR #1478, opened 2020-04-08
Refactor many `id` calls in sshfsmount

luis4a0:
The commands `id -u` and `id -g` are called two times each, consecutively, in sshfsmount. This can be refactored so each one is called only once.


luis4a0:
Oops, sorry for this. In any case, I'll force push again later, after merging #1473

luis4a0:
Since the PR is approved:

bors r=Saviq

PR #1494, opened 2020-04-17
[travis] run clang-format on pull requests

Saviq:
---
[Here's a previous build](https://travis-ci.org/github/canonical/multipass/builds/676103727) that had a formatting issue.


ricab:
As per Saviq's advice, using bors to check what happens when `type != pull_request`.

bors try

ricab:
OK, confirmed the new stage is just skipped then.

bors r+

Issue #1461, opened 2020-04-02
cloud-init url


hemna:
**What are you trying to do?**
I would like to be able to use a cloud-init.yml from a url while launching a new vm 

```
multipass launch -n test -c 2 --cloud-init http://my-home.com/multipass/cloud-init.yml
```

**What's your proposed solution?**
--cloud-init should support http:// files

**Additional context**

Saviq:
Thanks @hemna, that's something we could do indeed.

ianbtr:
Does the URLDownloader class support HTTPS URLs as well?

nwneisen:
It looks like this change died out due to needing tests written. I'd like to finish it up if that's alright.

Saviq:
Sure @nwneisen, you're welcome to it :)

krisstern:
If no one else is currently working on this issue I would like to pick this up.

krisstern:
@Saviq Would that be okay? I don't see any activity regarding this issue since June 2021.

Saviq:
Hey @krisstern, that would be great! You could rebase #1499, or start from scratch altogether.

krisstern:
Yeah, cool @Saviq Let me rebase #1499 and add tests rather than starting from scratch.

krisstern:
I see these lines in the `multipass/tests/test_image_vault.cpp` file that wasn't there when #1499 was first submitted:
https://github.com/canonical/multipass/blob/2f43d15d9963caeee3134dfafba55cb27c147f90/tests/test_image_vault.cpp#L68-L93

Will these be potentially useful for this issue @Saviq?

Saviq:
@krisstern you'd probably be best served using the `MockUrlDownloader`:

https://github.com/canonical/multipass/blob/2f43d15d9963caeee3134dfafba55cb27c147f90/tests/mock_url_downloader.h#L27-L34

krisstern:
Okay thanks for the tip!

PR #1499, opened 2020-04-20
Implements #1461

ianbtr:
Implements #1461.

This appears to work when using [this file](https://raw.githubusercontent.com/ianbtr/multipass/test_cloudinit_downloads/cloud_init_files/test.yml), like so:

`multipass launch --cloud-init https://raw.githubusercontent.com/ianbtr/multipass/test_cloudinit_downloads/cloud_init_files/test.yml`

Presently I'm unsure of how to add an appropriate test, is there a facility that mocks the URLDownloader?

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -26,11 +26,14 @@
 
 #include <multipass/format.h>
 
+#include <multipass/url_downloader.h>
+
 #include <yaml-cpp/yaml.h>
 
 #include <QDir>
 #include <QFileInfo>
 #include <QTimeZone>
+#include <QUrl>
 
 #include <regex>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


Saviq:
Hi @ianbtr thanks for this!

[tests/test_image_vault.cpp](https://github.com/canonical/multipass/blob/master/tests/test_image_vault.cpp#L90) has a handful of different mocked downloaders. The problem is that there currently isn't a way to pass it down from the test to the `cmd::Launch::parse_args()` method.

Looking at it it may be the right thing to pass the `mp::ClientConfig` instance to `cmd::Launch::run`, you would have to move parts of `::parse_args` (like actually reading the file in and downloading the file) into `::run`. You could then add a `URLDownloader` member to `mp::ClientConfig` and use it in `::run` - the test can then set a mock one in `send_command` and you will be able to check how it's used.

For a bigger contribution like that, I'll need to ask you to sign [the CLA](https://ubuntu.com/legal/contributors) please, if you have not done it yet :)

Saviq:
Hi @ianbtr are you able to finish this up with tests?

ianbtr:
Unfortunately, I will not be able to finish writing tests.

townsend2010:
We incorporated this in #2593 and is now merged!  Thanks and I'll close this PR!

Issue #1139, opened 2019-10-23
Skip image resize if the default is too small


paride:
Following the instructions in [1] I can successfully launch Fedora cloud images distributed at [2] using multipass:
```
$ multipass launch -n paride-fedora file:///home/paride/Fedora-Cloud-Base-30-1.2.x86_64.qcow2
Launched: paride-fedora

$ multipass exec paride-fedora cat /etc/os-release
NAME=Fedora
VERSION="30 (Cloud Edition)"
[...]
```
however I can't launch CentOS 7 instances using the CentOS cloud images [3]:
```
$ multipass launch -vvv -n paride-centos file:///home/paride/CentOS-7-x86_64-GenericCloud.qcow2
launch failed: Cannot resize instance image
```
This while running the multipass snap 0.8.1/1230 on an up-to-date amd64 Eoan system. Let me know if you need any other detail. Thanks!

[1] https://discourse.ubuntu.com/t/new-way-to-launch-images-other-than-cloud-images-from-simplestreams/6282
[2] https://alt.fedoraproject.org/cloud/
[3] https://cloud.centos.org/centos/7/images/

Saviq:
Hi @paride do the CentOS images have `cloud-init` set up? There's no other way we can inject ourselves into the VM.

ricab:
Hi @paride, this error arises because the default disk size for instances (5G) is smaller than the minimum required with that image (8G). You can make it work by passing `--disk 8G` in the launch command.

There is a PR improving this error msg, so in a future release you'd see something like: `launch failed: Requested disk (5368709120 bytes) below minimum for this image (8589934592 bytes)`.

Perhaps we should consider deriving the default automatically in such cases...

Saviq:
D'oh, should've read closer.

Indeed this works fine:
```
$ multipass launch --disk 10G https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2
Launched: frisky-chicken
```

I will make this bug about the minimum disk image handling.

paride:
I can confirm the workaround works. Thanks!

PR #1504, opened 2020-04-21
Skip image resize

luis4a0:
If the VM has a disk size bigger than the default, do not try to resize at launch it if no size was specified.

Fixes https://github.com/canonical/multipass/issues/1139

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -643,8 +643,10 @@ mp::MemorySize get_image_size(const mp::VMImage& image)
 // the size of the image.
 mp::MemorySize compute_final_image_size(const mp::VMImage& image, mp::optional<mp::MemorySize> command_line_value)
 {
-    mp::MemorySize image_size = get_image_size(image);
     mp::MemorySize disk_space{};
+
+    // TODO: the virtual image size should come from the vault.
+    mp::MemorySize image_size = get_image_size(image);
 
     if (!command_line_value)
     {


luis4a0:
Oops, forgot to update the tests. First thing tomorrow morning :)

luis4a0:
I should still add a test for the default disk size; the coverage will be complete then.

src/client/cli/cmd/launch.cpp
```diff
@@ -107,9 +107,10 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     QCommandLineOption diskOption(
         {"d", "disk"},
         QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
-                                           "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
+                                           "bytes, or with K, M, G suffix.\nMinimum: {}, default: the bigger of {} "
+                                           " and the image size.",
```
Saviq, line 111:
```suggestion
                                           "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}, or the default "
                                           "image size, if larger.",
```

Not entirely happy with this wording, either…

src/client/cli/cmd/launch.cpp
```diff
@@ -107,9 +107,10 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
                                            min_disk_size, default_disk_size)),
-        "disk", QString::fromUtf8(default_disk_size));
+        "disk", QString());
```
Saviq, line 113:
```suggestion
        "disk");
```

src/daemon/daemon.cpp
```diff
@@ -321,15 +321,15 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
...
     if (opt_mem_size && *opt_mem_size >= min_mem)
         mem_size = *opt_mem_size;
     else
         option_errors.add_error_codes(mp::LaunchError::INVALID_MEM_SIZE);
 
-    if (opt_disk_space && *opt_disk_space >= min_disk)
+    if (opt_disk_space && (*opt_disk_space >= min_disk || 0 == opt_disk_space->in_bytes()))
         disk_space = *opt_disk_space;
     else
         option_errors.add_error_codes(mp::LaunchError::INVALID_DISK_SIZE);
```
Saviq, line 335: I would refactor this to switch on `disk_space_str.empty()`, no need to `try_mem_size()` on it, if it is empty. The "0" is artificial here.

src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -103,15 +104,31 @@ void check_min_img_size(const mp::MemorySize& requested_size, const mp::Path& im
...
+    const auto match = re.match(img_info);
+
+    if (!match.hasMatch())
+        throw std::runtime_error{fmt::format("Could not obtain image's virtual size")};
+
+    const auto min_size = mp::MemorySize(match.captured("size").toStdString());
+
+    // If the size is zero at this point, it means that the user did not specify a size.
+    // We need to resize only if the default size is bigger than the image size.
+    if (0 == requested_size.in_bytes())
```
Saviq, line 116: It would be cleaner if it was an `mp::optional`, I suppose?

tests/test_daemon.cpp
```diff
@@ -685,6 +685,6 @@ INSTANTIATE_TEST_SUITE_P(Daemon, MinSpaceRespectedSuite,
                                  Values("1024m", "2Gb", "987654321")));
 INSTANTIATE_TEST_SUITE_P(Daemon, MinSpaceViolatedSuite,
                          Combine(Values("test_create", "launch"), Values("--mem", "--disk"),
-                                 Values("0", "0B", "0GB", "123B", "42kb", "100")));
+                                 Values("123B", "42kb", "100")));
```
Saviq, line 688: This actually shows why you shouldn't use "0" as the empty value :)

src/client/cli/cmd/launch.cpp
luis4a0: Is another option to leave it as it was?

src/client/cli/cmd/launch.cpp
luis4a0: Yeah, better.

src/daemon/daemon.cpp
luis4a0: Good.

src/platform/backends/shared/linux/backend_utils.cpp
luis4a0: Yeah, `mp::optional` will solve our problems.

tests/test_daemon.cpp
luis4a0: You're right in the sense that we must leave those zeroes. We leave the test untouched then.

luis4a0:
I closed the PR accidentally. It seems no build of the last push will be done, so...

bors try

luis4a0:
bors try

src/daemon/daemon.cpp
```diff
@@ -321,18 +321,28 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
...
     else
         option_errors.add_error_codes(mp::LaunchError::INVALID_MEM_SIZE);
 
-    if (opt_disk_space && *opt_disk_space >= min_disk)
-        disk_space = *opt_disk_space;
-    else
-        option_errors.add_error_codes(mp::LaunchError::INVALID_DISK_SIZE);
+    // If the user did not specify a disk size, then 0 will be passed down. Otherwise, the specified size will be
+    // checked.
+    mp::MemorySize disk_space{}; // Zero by default.
```
Saviq, line 333: Didn't we say an `optional<mp::MemorySize>` would be cleaner than assigning a special meaning to "0"?

src/daemon/daemon.cpp
```diff
@@ -1964,7 +2006,27 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
+                if (checked_args.disk_space.in_bytes() == 0)
+                {
+                    auto default_disk_size_as_struct = mp::MemorySize(mp::default_disk_size);
+                    disk_space = image_size < default_disk_size_as_struct ? default_disk_size_as_struct : image_size;
+                }
+                else if (checked_args.disk_space < image_size)
+                {
+                    throw std::runtime_error(fmt::format("requested image size ({} bytes) is smaller than image "
+                                                         "size ({} bytes)",
+                                                         checked_args.disk_space.in_bytes(), image_size.in_bytes()));
```
Saviq, line 2022: Could this be done earlier? With how it is now, we copy the image just to then throw that the requested size was too small.

src/daemon/daemon.cpp
```diff
@@ -1964,7 +2006,27 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
+                mp::MemorySize disk_space{};
+
+                if (checked_args.disk_space.in_bytes() == 0)
+                {
+                    auto default_disk_size_as_struct = mp::MemorySize(mp::default_disk_size);
+                    disk_space = image_size < default_disk_size_as_struct ? default_disk_size_as_struct : image_size;
+                }
+                else if (checked_args.disk_space < image_size)
+                {
+                    throw std::runtime_error(fmt::format("requested image size ({} bytes) is smaller than image "
```
Saviq, line 2020: Why not use the same text that you remove below?

include/multipass/platform.h
```diff
@@ -56,6 +56,7 @@ VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
 std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& config);
+std::unique_ptr<Process> make_qemuimg_process(const QStringList& parameters);
```
Saviq, line 59: Couldn't this be generic instead?
```suggestion
std::unique_ptr<Process> make_process(const ProcessSpec& spec, const QStringList& parameters);
```

(or an appropriate smart pointer version of the above)

src/platform/backends/shared/qemuimg_process_spec.cpp
```diff
@@ -17,6 +17,8 @@
 
 #include "qemuimg_process_spec.h"
 
+#include <QString>
+#include <QStringList>
```
Saviq, line 21: Shouldn't these, strictly, go to the header instead?

src/daemon/daemon.cpp
luis4a0: Right, I only replaced it to pass down the command-line argument. Fixed.

src/daemon/daemon.cpp
luis4a0: As discussed, doing it earlier would imply a major refactor.

src/daemon/daemon.cpp
luis4a0: Right. Changed.

include/multipass/platform.h
luis4a0: Good point, done.

src/platform/backends/shared/qemuimg_process_spec.cpp
luis4a0: Right, moved to the header.

tests/test_daemon.cpp
```diff
@@ -643,6 +643,14 @@ TEST_P(DaemonCreateLaunchTestSuite, default_cloud_init_grows_root_fs)
     auto mock_factory = use_a_mock_vm_factory();
     mp::Daemon daemon{config_builder.build()};
 
+    auto mock_factory_scope = mpt::MockProcessFactory::Inject();
+
+    mock_factory_scope->register_callback([&](mpt::MockProcess* process) {
+        const mp::ProcessState qemuimg_exit_status{0, mp::nullopt};
+        const QByteArray qemuimg_output(fake_img_info(mp::MemorySize{("1048577")}));
+        simulate_qemuimg_info(process, qemuimg_exit_status, qemuimg_output);
+    });
```
Saviq, line 652: Could this be refactored into something shared between the tests here?

src/daemon/daemon.cpp
```diff
@@ -1840,7 +1840,7 @@ mp::MemorySize mp::Daemon::get_image_size(const mp::VMImage& image)
     }
 
     const auto img_info = QString{qemuimg_process->read_all_standard_output()};
-    const auto pattern = QStringLiteral("^virtual size: .+ \\((?<size>\\d+) bytes\\)$");
+    const auto pattern = QStringLiteral("^virtual size: .+ \\((?<size>\\d+) bytes\\)");
```
Saviq, line 1843: Could we be more strict with something like `\r?$` instead?

tests/test_daemon.cpp
luis4a0: Good idea, done.

src/daemon/daemon.cpp
luis4a0: Right, done.

src/daemon/daemon.h
```diff
@@ -138,6 +138,8 @@ public slots:
     void release_resources(const std::string& instance);
     std::string check_instance_operational(const std::string& instance_name) const;
     std::string check_instance_exists(const std::string& instance_name) const;
+    MemorySize get_image_size(const VMImage& image);
+    MemorySize compute_final_image_size(const VMImage& image, multipass::optional<MemorySize> memory_argument);
```
Saviq, line 142: Couldn't these be in anonymous namespace instead? They're utility functions that I don't believe have any dependency on the Daemon object?

tests/linux/test_backend_utils.cpp
```diff
@@ -129,25 +100,14 @@ void simulate_qemuimg_convert(const mpt::MockProcess* process, const QString& im
...
-            auto msg = QByteArray{qemuimg_info_output};
-            if (msg.isEmpty())
-                msg = fake_img_info(img_virtual_size);
-
-            simulate_qemuimg_info(process, img, qemuimg_info_result, msg);
-        }
-        else
         {
             simulate_qemuimg_resize(process, img, requested_size, qemuimg_resize_result);
         }
```
Saviq, line 113: The `if` is spurious with the `ASSERT_LE` above.

luis4a0:
About your question on passing `mp::nullopt`, that can be done. However, if we pass `mp::nullopt` we'll be loosing the actual size of the disk in the VM description. Don't you think better to pass a new `bool` argument to `prepare_instance_image()`?

tests/linux/test_backend_utils.cpp
luis4a0: Right, removed.

src/daemon/daemon.h
luis4a0: Right, they could be in namespace{}.

Saviq:
FWIW this got merged at ba87d910d02e4dde4ba632220114556e8aae1706.

PR #1519, opened 2020-05-01
[network] Implement Qt classes to communicate with a Unix socket

townsend2010:
This is needed to communicate with the LXD Unix socket.
    
This is a minimal implementation and can be modified later for more/better functionality.

---
We should be able to replace anywhere we use `QNetworkAccessManager` with this class and have just one implementation.

### DIFF
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -18,6 +18,7 @@
 #include "local_socket_reply.h"
 
 #include <multipass/format.h>
+#include <multipass/version.h>
 
 namespace mp = multipass;
 
@@ -25,12 +26,14 @@ namespace
 {
 constexpr int len = 65536;
 
-// The following is mostly "borrowed" from Qt src/network/access/qhttpthreaddelegate.cpp
+// Status code mapping based on
+// https://github.com/qt/qtbase/blob/dev/src/network/access/qhttpthreaddelegate.cpp
 QNetworkReply::NetworkError statusCodeFromHttp(int httpStatusCode)
 {
     QNetworkReply::NetworkError code;
 
-    // Only switch on the HTTP errors LXD will respond with for now
+    // Only switch on the LXD HTTP errors:
+    // https://lxd.readthedocs.io/en/latest/rest-api/#error
     switch (httpStatusCode)
     {
     case 400: // Bad Request
@@ -78,7 +81,7 @@ mp::LocalSocketReply::LocalSocketReply(const QString& socket_path, const QNetwor
                                        QIODevice* outgoingData)
     : QNetworkReply(), local_socket{std::make_unique<QLocalSocket>(this)}, reply_data{QByteArray(len, '\0')}
 {
-    QIODevice::open(QIODevice::ReadOnly);
+    open(QIODevice::ReadOnly);
 
     local_socket->connectToServer(socket_path);
 
@@ -99,6 +102,12 @@ mp::LocalSocketReply::~LocalSocketReply()
 
 void mp::LocalSocketReply::abort()
 {
+    close();
+
+    setError(OperationCanceledError, "Operation canceled");
+    emit error(OperationCanceledError);
+
+    emit finished();
 }
 
 qint64 mp::LocalSocketReply::readData(char* data, qint64 maxSize)
@@ -133,12 +142,9 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
     http_data += "Host: multipass\r\n";
 
     // Build the HTTP User-Agent header
-    // We'll just use what Qt uses
-    http_data += "User-Agent: Mozilla/5.0\r\n";
-
-    // Build the HTTP Accept header
-    // Default to accept everything
-    http_data += "Accept: */*\r\n";
+    http_data += "User-Agent: Multipass/";
+    http_data += mp::version_string;
+    http_data += "\r\n";
 
     if (op == "POST" || op == "PUT")
     {
@@ -163,7 +169,15 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
 
 void mp::LocalSocketReply::read_reply()
 {
-    local_socket->read(reply_data.data(), len);
+    auto data_ptr = reply_data.data();
+    int bytes_read{0};
+
+    do
+    {
+        bytes_read = local_socket->read(reply_data.data(), len);
+
+        data_ptr += bytes_read;
+    } while (bytes_read > 0);
 
     parse_reply();
 
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -20,6 +20,9 @@
 #include <multipass/format.h>
 #include <multipass/version.h>
 
+#include <QRegularExpression>
+#include <QRegularExpressionMatch>
+
 namespace mp = multipass;
 
 namespace
@@ -210,16 +213,12 @@ void mp::LocalSocketReply::parse_reply()
     }
 }
 
-// Some of this logic is taken from Qt src/network/access/qhttpnetworkreply.cpp
 void mp::LocalSocketReply::parse_status(const QByteArray& status)
 {
-    const int minLength = 11;
-    const int dotPos = 6;
-    const int spacePos = 8;
-    const char httpMagic[] = "HTTP/";
+    QRegularExpression http_status_regex{"^HTTP/\\d\\.\\d (?P<status>\\d{3})\\ (?P<message>.*)$"};
+    QRegularExpressionMatch http_status_match = http_status_regex.match(QString::fromLatin1(status));
 
-    if (status.length() < minLength || !status.startsWith(httpMagic) || status.at(dotPos) != '.' ||
-        status.at(spacePos) != ' ')
+    if (!http_status_match.hasMatch())
     {
         setError(QNetworkReply::ProtocolFailure, "Malformed HTTP response from server");
 
@@ -228,18 +227,14 @@ void mp::LocalSocketReply::parse_status(const QByteArray& status)
         return;
     }
 
-    int i = spacePos;
-    int j = status.indexOf(' ', i + 1); // j == -1 || at(j) == ' ' so j+1 == 0 && j+1 <= length()
-    const QByteArray code = status.mid(i + 1, j - i - 1);
-
     bool ok;
-    auto statusCode = code.toInt(&ok);
+    auto statusCode = http_status_match.captured("status").toInt(&ok);
 
     if (statusCode >= 400)
     {
         auto error_code = statusCodeFromHttp(statusCode);
 
-        setError(error_code, QString::fromLatin1(status.constData() + j + 1));
+        setError(error_code, http_status_match.captured("message"));
 
         emit error(error_code);
     }
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -1,0 +1,232 @@
+/*
+ * Copyright (C) 2020 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "local_socket_reply.h"
+
+#include <multipass/format.h>
+
+namespace mp = multipass;
+
+namespace
+{
+constexpr int len = 65536;
+
+// The following is mostly "borrowed" from Qt src/network/access/qhttpthreaddelegate.cpp
+QNetworkReply::NetworkError statusCodeFromHttp(int httpStatusCode)
+{
+    QNetworkReply::NetworkError code;
+
+    // Only switch on the HTTP errors LXD will respond with for now
+    switch (httpStatusCode)
+    {
+    case 400: // Bad Request
+        code = QNetworkReply::ProtocolInvalidOperationError;
+        break;
+
+    case 401: // Authorization required
+        code = QNetworkReply::AuthenticationRequiredError;
+        break;
+
+    case 403: // Access denied
+        code = QNetworkReply::ContentAccessDenied;
+        break;
+
+    case 404: // Not Found
+        code = QNetworkReply::ContentNotFoundError;
+        break;
+
+    case 409: // Resource Conflict
+        code = QNetworkReply::ContentConflictError;
+        break;
+
+    case 500: // Internal Server Error
+        code = QNetworkReply::InternalServerError;
+        break;
+
+    default:
+        if (httpStatusCode > 500)
+        {
+            // some kind of server error
+            code = QNetworkReply::UnknownServerError;
+        }
+        else
+        {
+            // content error we did not handle above
+            code = QNetworkReply::UnknownContentError;
+        }
+    }
+
+    return code;
+}
+} // namespace
+
+mp::LocalSocketReply::LocalSocketReply(const QString& socket_path, const QNetworkRequest& request,
+                                       QIODevice* outgoingData)
+    : QNetworkReply(), local_socket{std::make_unique<QLocalSocket>(this)}, reply_data{QByteArray(len, '\0')}
+{
+    QIODevice::open(QIODevice::ReadOnly);
+
+    local_socket->connectToServer(socket_path);
+
+    if (!local_socket->waitForConnected(5000))
+    {
+        throw std::runtime_error(fmt::format("Cannot connect to {}: {}", socket_path, local_socket->error()));
+    }
+
+    QObject::connect(local_socket.get(), &QLocalSocket::readyRead, this, &LocalSocketReply::read_reply);
+
+    send_request(request, outgoingData);
+}
+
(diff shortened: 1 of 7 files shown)


src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+    http_data += request.url().path();
+    http_data += " HTTP/1.1\r\n";
+
+    // Build the HTTP Host header
+    // Host can be anything, so we'll use 'multipass'
+    http_data += "Host: multipass\r\n";
+
+    // Build the HTTP User-Agent header
+    // We'll just use what Qt uses
+    http_data += "User-Agent: Mozilla/5.0\r\n";
```
Saviq, line 137: It may actually be interesting to put `Multipass/<version>` here, so LXD can potentially use that data for metrics.

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    // Build the HTTP Accept header
+    // Default to accept everything
+    http_data += "Accept: */*\r\n";
```
Saviq, line 141: Does LXD require it?

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        if (outgoingData)
+        {
+            outgoingData->open(QIODevice::ReadOnly);
+
+            http_data += "Content-Length: ";
+            http_data += QByteArray::number(outgoingData->size());
+            http_data += "\r\n\r\n";
+            http_data += outgoingData->readAll();
+        }
```
Saviq, line 155: We'll have to reimplement this when streaming images in, right?

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+void mp::LocalSocketReply::read_reply()
+{
+    local_socket->read(reply_data.data(), len);
+
+    parse_reply();
+
+    emit finished();
+}
```
Saviq, line 171: This assumes that the reply will never be > 64k long, right? Add a comment?

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    if (statusCode >= 400)
+    {
+        auto error_code = statusCodeFromHttp(statusCode);
+
+        setError(error_code, QString::fromLatin1(status.constData() + j + 1));
+
+        emit error(error_code);
+    }
+}
```
Saviq, line 232: Couldn't we just regex this?…

src/network/local_socket_reply.h
```diff
@@ -0,0 +1,59 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_LOCAL_SOCKET_REPLY_H
+#define MULTIPASS_LOCAL_SOCKET_REPLY_H
+
```
Saviq, line 20:
```cpp
#include <QByteArray>
```

src/network/network_access_manager.cpp
```diff
@@ -0,0 +1,60 @@
...
+QNetworkReply* mp::NetworkAccessManager::createRequest(QNetworkAccessManager::Operation operation,
+                                                       const QNetworkRequest& orig_request, QIODevice* device)
+{
+    auto scheme = orig_request.url().scheme();
+
+    // To support http requests over Unix sockets, the initial URL needs to be in the form of:
+    // unix:///path/to/unix_socket@path/in/server (or 'local' instead of 'unix')
+    //
+    // For example, to get the general LXD configuration when LXD is installed as a snap:
+    // unix:////var/snap/lxd/common/lxd/unix.socket@1.0
```
Saviq, line 37: Yeah, that works - if anyone actually wants '@' in their URL, they can always encode it, right?

tests/test_local_network_access_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+        QObject::connect(&test_server, &QLocalServer::newConnection, [&] {
+            auto client_connection = test_server.nextPendingConnection();
+
+            client_connection->waitForReadyRead();
+            auto data = client_connection->readAll();
+
+            if (!expected_data.isEmpty())
+            {
+                EXPECT_THAT(data, Eq(expected_data));
```
Saviq, line 58:
```suggestion
                EXPECT_EQ(data, expected_data);
```

tests/test_local_network_access_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+    QByteArray http_response;
+    http_response += "HTTP/1.1 200 OK\r\n";
+    http_response += "Content-Type: text/plain\r\n";
+    http_response += "\r\n";
+
+    local_socket_server_handler(http_response);
+
+    auto reply = handle_request(base_url, "GET");
+
+    EXPECT_THAT(reply->error(), Eq(QNetworkReply::NoError));
```
Saviq, line 118:
```suggestion
    EXPECT_EQ(reply->error(), QNetworkReply::NoError);
```

tests/test_local_network_access_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+    http_response += "Content-Length: 5\r\n";
+    http_response += "\r\n";
+    http_response += reply_data;
+    http_response += "\r\n";
+
+    local_socket_server_handler(http_response);
+
+    auto reply = handle_request(base_url, "GET");
+
+    ASSERT_THAT(reply->error(), Eq(QNetworkReply::NoError));
```
Saviq, line 137:
```suggestion
    ASSERT_EQ(reply->error(), QNetworkReply::NoError);
```

tests/test_local_network_access_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    auto data = reply->readAll();
+
+    EXPECT_THAT(data, Eq(reply_data));
```
Saviq, line 141:
```suggestion
    EXPECT_EQ(data, reply_data);
```

tests/test_local_network_access_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    QByteArray reply_data{"What's up?"};
+
+    QByteArray http_response;
+    http_response += "HTTP/1.1 200 OK\r\n";
+    http_response += "Content-Type: text/plain\r\n";
+    http_response += "Content-Length: 5\r\n";
+    http_response += "Transfer-Encoding: chunked\r\n";
+    http_response += "\r\n";
+    http_response += "5\r\n";
```
Saviq, line 154: This is 53, no? So your reply is a bit malformed? ;P

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+#include <multipass/format.h>
+
+namespace mp = multipass;
+
+namespace
+{
+constexpr int len = 65536;
+
+// The following is mostly "borrowed" from Qt src/network/access/qhttpthreaddelegate.cpp
```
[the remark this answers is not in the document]
townsend2010: Right, I just put the comment here to let anyone reading the code know where it came from.  Of course it's not necessary to put that, but I thought it would make it more clear.  Any additional opinions based on what I said or are you still on the notion to whack it?

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+QNetworkReply::NetworkError statusCodeFromHttp(int httpStatusCode)
+{
+    QNetworkReply::NetworkError code;
+
+    // Only switch on the HTTP errors LXD will respond with for now
```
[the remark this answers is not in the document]
townsend2010: Sure, I can add a link for where the values come from.

Regarding error `412`, there isn't an equivalent `QNetworkReply::NetworkError`, so I'm just letting the default case handle that.

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    return code;
+}
+} // namespace
+
+mp::LocalSocketReply::LocalSocketReply(const QString& socket_path, const QNetworkRequest& request,
+                                       QIODevice* outgoingData)
+    : QNetworkReply(), local_socket{std::make_unique<QLocalSocket>(this)}, reply_data{QByteArray(len, '\0')}
```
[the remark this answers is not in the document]
townsend2010: I ended up having to do this because the data we get from LXD is not terminated by `\0` and rather than trying to figure out where to add `\0` in the reply data, it was cleaner just setting the whole `QByteArray` to `\0` upon initialization.

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+    QIODevice::open(QIODevice::ReadOnly);
```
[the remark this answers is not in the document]
townsend2010: I can't remember if I tried that or not...I can certainly see, but I do think in the end, it's the same static function that is called.

src/network/local_socket_reply.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+mp::LocalSocketReply::~LocalSocketReply()
+{
+    local_socket->disconnectFromServer();
+}
+
+void mp::LocalSocketReply::abort()
+{
+}
```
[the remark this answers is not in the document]
townsend2010: Hmm, I think I forgot to see about whether we need to implement this or not.  I'll follow up on that and either implement it or respond back here.

src/network/local_socket_reply.cpp
townsend2010: Sure, whatever should work.

src/network/local_socket_reply.cpp
townsend2010: Not sure, I was just following what `curl` does.  I can certainly try it without it and remove it if it works.

src/network/local_socket_reply.cpp
townsend2010: Yes, if it requires us "uploading" an image to LXD's 'local' remote.  I kind of had in the back of my mind what we'd have to do that, but I think that should be for another iteration of this.

src/network/local_socket_reply.cpp
townsend2010: For right now, yes.  It should really loop there until `-1` or `0` is returned from `read()`.  Perhaps I should go ahead and fix that.

src/network/local_socket_reply.cpp
townsend2010: Sure, many ways to do this.  I thought the fastest way was to just copy what Qt already does in its implementation :shrug:  Didn't feel the need to reinvent the wheel.

src/network/local_socket_reply.h
townsend2010: Yep, should be explicit about includes.

src/network/network_access_manager.cpp
townsend2010: It would be that someone has an `@` in their path to the local socket.  I'm not sure what the chances of that happening, particularly pertaining to the LXD socket since that is really all we care about now.  Since this is an internal way of parsing local sockets and REST paths, then we can always change it later and it should still work.

tests/test_local_network_access_manager.cpp
townsend2010: Ok, it's cleaner to use `EXPECT_EQ`  so I'll switch them all up.  I'll just change it in the next push instead of committing all of these suggestions if that is ok.

tests/test_local_network_access_manager.cpp
townsend2010: I'm not really following you.  The data is 5 bytes (`Hello`), so the chunked data size is `5` here.

townsend2010:
@Saviq,

Thanks for the review!  Regarding removing `Content-*` in the tests, I'll do that.

As for the one missing coverage line, it's actually calling the real `QNetworkAccessManager`, so no way to easily mock that.  Maybe I can give it a bad `http` url and check that the reply errors out in an expected way?  I'll try it and see.

tests/test_local_network_access_manager.cpp
Saviq: Of course.

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
Saviq: Then just give a URL to the file, saying something like "Status code mapping based on <url>".

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
Saviq: Yeah was meaning to remove the 412 comment…

src/network/local_socket_reply.cpp
Saviq: Yeah, didn't mean to ask that you do that now.

src/network/local_socket_reply.cpp
Saviq: It's just that someone needs to review it and maintain ;)

tests/test_local_network_access_manager.cpp
Saviq: It's actually "What's up?", no? And I misunderstood what you meant by "in hex" above, so this should actually be 9, right?

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Ack, will do.

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Ok

tests/test_local_network_access_manager.cpp
townsend2010: Oh, I was looking at the wrong test :grin:   Will fix.

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Hmm, I'm not sure what you mean by "remove the 412 comment"?  Do you mean add a comment explaining why 412 is not a case?

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Ok, just calling `open()` will do the trick.

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
Saviq: _I_ was meaning to remove the 412 comment ;) I'll just clean this mess up now ;P

src/network/local_socket_reply.cpp
townsend2010: @Saviq, hey, your regex foo is much better than mine, so I'm asking for some help on this if you'd rather this be a regex :grin:  Otherwise, this works and I'd rather spend my time getting this PR good to go and getting #1533 done.

Saviq: Oh sure, https://regex101.com/r/FJVUf8/1 :)

Saviq: Or well, https://regex101.com/r/FJVUf8/2/ to be exactly as the above, but I don't think the status code should be anything else than 3 digits?

townsend2010: Man, your good:)  Yes, afaik, http status codes should be 3 digits, so I think it's safe to delineate on that or otherwise it's a malformed status code.  Thanks!

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/network/local_socket_reply.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/network/local_socket_reply.cpp
townsend2010: Done

src/network/local_socket_reply.cpp
townsend2010: Done

src/network/local_socket_reply.cpp
townsend2010: Done, accept I didn't add an explicit `-1` case, but instead wrapped it all in a `> 0` case to keep looping.

src/network/local_socket_reply.cpp
townsend2010: Done

src/network/local_socket_reply.h
townsend2010: Done

tests/test_local_network_access_manager.cpp
townsend2010: Done

tests/test_local_network_access_manager.cpp
townsend2010: Done

tests/test_local_network_access_manager.cpp
townsend2010: Done

tests/test_local_network_access_manager.cpp
townsend2010: All done :smiley:

tests/test_local_network_access_manager.cpp
townsend2010: Fixed

townsend2010:
@Saviq,

I believe I addressed all of your feedback.  I did add a test for checking if `QNetworkAccessManager` proper is called too.

PR #1520, opened 2020-05-02
[custom] add snapcraft:core20 minimal image (Fixes #1417)

### LINKED ISSUE #1417: How to use focal images?
focal images already exist in daily streams, but not yet in release streams.

It seems to me, that once they do land in the release stream "multipass will not just work" because of the hardcoded mappings at https://github.com/canonical/multipass/blob/master/src/daemon/custom_image_host.cpp#L56

Could you please update those mappings for core20 / 20.04 / focal pointing at the non-existent release streams? Such that when they will be published there, "multipass just starts working" with them?

It would be also nice to have ability to point multipass at the daily streams, but imho that is a separate request which i have filed at https://github.com/canonical/multipass/issues/1416

Saviq:
While we're working on the buildd images, `core20` can now be supported using the `minimal` images.


townsend2010:
Bah...

bors r+

PR #1530, opened 2020-05-12
[args] Enable debug verbosity on more than 3 vs

ricab:
Multipass accepted more than 3 vs in `-v` but kept the verbosity as error. This sets it to debug on 3 or more vs.

I couldn't find tests for arg parsing or verbosity. I am not sure if this is the PR to add them, so I just tested manually by editing code to print the configured verbosity.

### DIFF
--- a/src/client/cli/argparser.cpp
+++ b/src/client/cli/argparser.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Canonical, Ltd.
+ * Copyright (C) 2017-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


ricab:
I'd be fine with that too. I think it should be one or the other, just not accepting multiple vs but keeping default verbosity.

townsend2010:
@Saviq, any opinion on whether this should accept 3 or more `-v`'s for `debug` verbosity or error out if more than 3 `-v`'s are given?

Saviq:
I don't think there's a point in erroring out - but we have a `trace` level now, so that should probably be handled, too.

townsend2010:
Ok, that's fine.  My only concern is that the `help` says `repeat up to three times for more detail`, so if a user repeats more than 3 (or 4 now), then they may have an expectation that more detail may be given.  But meh, it's really splitting hairs at this point.

That is a good point about differentiating 3 times and 4 (or more) times and thus, fixing the help output as well.

ricab:
@townsend2010 How about now? I fixed the help and also added tests.

PR #1533, opened 2020-05-13
Support LXD as a backend driver

townsend2010:
This allows Multipass to communicate with the LXD daemon in order to create and mange instances.

---

List of TODO items needed for the next iteration:
- Additional test coverage.
- See if it's possible to mock `NetworkAccessManager` and `LocalSocketReply` instead of using `MockLocalSocketServer`.
- Figure out a way to use a single `NetworkAccessManager` across `LXDVirtualMachineFactory` and `LXDVMImageVault`.
- Overhaul `LXDVirtualMachineFactory::hypervisor_health_check()` including detecting if LXD socket is not available, etc.
- Determine how best to handle the cloud-init YAML config.  Do we store it somewhere and re-populate upon daemon start so `VirtualMachineDesription` contains the values?  Do we just create it once and let `LXDVirtualMachineFactory` configure it and then forget about it since LXD has it stored?  
- Re-add back in `lxd-request()` the ability to wait on operations to complete.
- Fix `LXDVirtualMachine` to wait on an instance to be created instead of the clunky `while` loop w/ `sleep()`.
- Determine if LXD itself can properly update and prune source images and potentially refactor the `VMImageVault`'s `update_images()` and `prune_expired_images` and associated timer.
- Determine if the `source` part of the LXD initialization can be modified to not explicitly call out the remote since the image should already be in the `local` repository.
- Allow importing images from the "custom" image host and http/file based launches.


Saviq:
```
In file included from /root/parts/multipass/src/src/platform/backends/lxd/lxd_vm_image_vault.cpp:18:
/root/parts/multipass/src/src/platform/backends/lxd/lxd_vm_image_vault.h:51:26: error: private field 'url_downloader' is not used [-Werror,-Wunused-private-field]
    URLDownloader* const url_downloader;
                         ^
1 error generated.
```

townsend2010:
@Saviq, yep, saw that yesterday and will fix today along with some other things :grinning:

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -0,0 +1,386 @@
...
+                        std::this_thread::sleep_for(5s);
+                    }
+                }
+                // Implies the task is finished
+                catch (const LXDNotFoundException& e)
+                {
+                    break;
+                }
+            }
+        }
```
Saviq, line 199: Should this go into lxd_request?

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -0,0 +1,386 @@
...
+                    }
+
+                    auto status_code = task_reply["metadata"].toObject()["status_code"].toInt(-1);
+                    if (status_code == 200)
+                    {
+                        break;
+                    }
+                    else
+                    {
+                        std::this_thread::sleep_for(5s);
```
Saviq, line 190: This a long sleep, `1s` would be rare enough, IMO.

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -0,0 +1,386 @@
...
+    if (present_state == State::running || present_state == State::delayed_shutdown)
+    {
+        auto state_task = request_state("stop");
+        if (state_task["metadata"].toObject()["class"] == QStringLiteral("task") &&
+            state_task["status_code"].toInt(-1) == 100)
+        {
+            QUrl task_url(QString("%1/operations/%2/wait")
+                              .arg(base_url.toString())
+                              .arg(state_task["metadata"].toObject()["id"].toString()));
+            lxd_request(manager, "GET", task_url);
```
Saviq, line 240: Couldn't use that instead of the above poll? Or will it time out?

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -0,0 +1,386 @@
...
+            if (ip)
+            {
+                return mpu::TimeoutAction::done;
+            }
+            else
+            {
+                return mpu::TimeoutAction::retry;
+            }
+        };
+        auto on_timeout = [] { return std::runtime_error("failed to determine IP address"); };
```
Saviq, line 337: `throw`!

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void mp::LXDVirtualMachineFactory::remove_resources_for(const std::string& name)
+{
+}
+
+mp::FetchType mp::LXDVirtualMachineFactory::fetch_type()
+{
+    return mp::FetchType::None;
+}
```
Saviq, line 55: This would be back at "Image".

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -0,0 +1,84 @@
...
+mp::VMImage mp::LXDVirtualMachineFactory::prepare_source_image(const mp::VMImage& /* source_image */)
+{
+    throw std::runtime_error("prepare_source_image unimplemented");
+}
+
+void mp::LXDVirtualMachineFactory::prepare_instance_image(const mp::VMImage& /* instance_image */,
+                                                          const VirtualMachineDescription& /* desc */)
+{
+    throw std::runtime_error("prepare_instance_image unimplemented");
+}
```
Saviq, line 66: These should be NOOPs instead.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+void mp::LXDVirtualMachineFactory::configure(const std::string& /* name */, YAML::Node& /* meta_config */,
+                                             YAML::Node& /* user_config */)
+{
+}
```
Saviq, line 71: Hmm should we actually configure cloud init through here rather than shipping it with the vm description?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <thread>
+
+namespace mp = multipass;
+namespace mpl = multipass::logging;
+
+using namespace std::literals::chrono_literals;
+
+namespace
+{
+constexpr auto category = "image vault";
```
Saviq, line 40:
```suggestion
constexpr auto category = "lxd image vault";
```

include/multipass/fetch_type.h
```diff
@@ -23,6 +23,7 @@ namespace multipass
 {
 enum class FetchType
 {
+    None,
```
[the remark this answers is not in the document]
townsend2010: Agreed, I plan on changing this to a `LXD` fetch type and add it to the bottom of the `enum`.

src/daemon/daemon.cpp
```diff
@@ -1968,7 +1967,8 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                                                config->ssh_username, vm_image, meta_data_cloud_init_config,
                                                user_data_cloud_init_config, vendor_data_cloud_init_config);
 
-                config->factory->prepare_instance_image(vm_image, vm_desc);
+                if (fetch_type != mp::FetchType::None)
+                    config->factory->prepare_instance_image(vm_image, vm_desc);
```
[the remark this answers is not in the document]
townsend2010: Yep

src/daemon/default_vm_image_vault.cpp
```diff
@@ -621,6 +626,12 @@ mp::VMImage mp::DefaultVMImageVault::download_and_prepare_source_image(
 
     DeleteOnException image_file{source_image.image_path};
 
+    if (fetch_type == FetchType::None)
+    {
+        source_image.image_path = "";
+        return source_image;
+    }
+
```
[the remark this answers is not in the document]
townsend2010: Oh, right, I forgot about this one.  Will remove.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -693,16 +704,22 @@ mp::VMImage mp::DefaultVMImageVault::image_instance_from(const std::string& inst
...
+        return {copy(prepared_image.image_path, output_dir),
+                copy(prepared_image.kernel_path, output_dir),
+                copy(prepared_image.initrd_path, output_dir),
+                prepared_image.id,
+                prepared_image.stream_location,
+                prepared_image.original_release,
+                prepared_image.current_release,
+                prepared_image.release_date,
+                {}};
+    }
```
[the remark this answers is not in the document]
townsend2010: Yep

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -0,0 +1,87 @@
...
+    download_timeout.setInterval(timeout);
+
+    QNetworkRequest request{url};
+
+    auto verb = QByteArray::fromStdString(method);
+    QByteArray data;
+    if (json_data)
+    {
+        data = QJsonDocument(*json_data).toJson();
+        mpl::log(mpl::Level::debug, request_category, fmt::format("Sending data: {}", data));
```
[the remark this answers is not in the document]
townsend2010: Agreed that most of these will become `trace`, I just haven't done it yet;)

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: I kind of think the caller of `lxd_request()` should determine how it waits on an operation to finish.  Some callers may use the `operation/blah/wait` method while others may use a polling (and later websockets) method.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -0,0 +1,216 @@
...
+    {
+        return match.captured(1).toInt();
+    }
+
+    return -1;
+}
+} // namespace
+
+mp::LXDVMImageVault::LXDVMImageVault(std::vector<VMImageHost*> image_hosts, const QUrl& base_url)
+    : image_hosts{image_hosts}, base_url{base_url}, manager{std::make_unique<mp::NetworkAccessManager>()}
```
Saviq, line 58: {C,Sh}ould the manager not be shared?

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Yep, I planned on shortening this, but good to point it out so I don't forget:)

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: I'm not sure if you're referring to using `wait` here or the polling method.

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: OMG, how did I let this in!!!!!

src/daemon/daemon.cpp
```diff
@@ -644,8 +634,17 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
...
+                                              spec.mem_size,
+                                              spec.disk_space,
+                                              name,
+                                              mac_addr,
+                                              spec.ssh_username,
+                                              vm_image,
+                                              cloud_init_iso,
+                                              {},
+                                              {},
+                                              {}};
```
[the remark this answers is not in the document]
townsend2010: Maybe, but as you just mentioned, perhaps the image vault should configure it?  I'll need to think about it some.

src/platform/backends/lxd/lxd_virtual_machine.cpp
Saviq: Yes, using `wait`, why not use it above?

src/platform/backends/lxd/lxd_virtual_machine.cpp
Saviq: Sure, it would have to be the caller's decision whether to let `lxd_request()` handle the async operation or not. I'm kinda thinking `::waitForFinished()` style. Except it's not an object, I know ;)

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Sorry, I understand this needs to change, but I'm not sure I follow what you mean here.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Right...not sure why I did that :shrug:

include/multipass/fetch_type.h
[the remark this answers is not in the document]
Saviq: Hmm why a new enum? I know it kinda only makes sense for the lxd backend to be used with the lxd vault… but I don't think that enum should be the one to ensure that.

Should the image vault be a member of the factory, then?

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Hmm, I'll look into that.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    for (const auto& image_host : image_hosts)
+    {
+        for (const auto& remote : image_host->supported_remotes())
+        {
+            remote_image_host_map[remote] = image_host;
+        }
+    }
+}
```
Saviq, line 67: Should we filter out those without a stream URL here?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -0,0 +1,216 @@
...
+                        auto download_progress = parse_percent_as_int(
+                            task_reply["metadata"].toObject()["metadata"].toObject()["download_progress"].toString());
+
+                        if (!monitor(LaunchProgress::IMAGE, download_progress))
+                        {
+                            lxd_request(manager.get(), "DELETE", task_url);
+                            throw mp::AbortedDownloadException{"Download aborted"};
+                        }
+
+                        std::this_thread::sleep_for(5s);
```
Saviq, line 137:
```suggestion
                        std::this_thread::sleep_for(1s);
```

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+void mp::LXDVMImageVault::prune_expired_images()
+{
+}
+
+void mp::LXDVMImageVault::update_images(const FetchType& fetch_type, const PrepareAction& prepare,
+                                        const ProgressMonitor& monitor)
+{
+}
```
Saviq, line 187: These are TODO, right? We should definitely prune old images, and {c,sh}ould we drive the update some way?

src/platform/platform_linux.cpp
```diff
@@ -117,10 +120,24 @@ mp::VirtualMachineFactory::UPtr mp::platform::vm_backend(const mp::Path& data_di
...
+                                                                 mp::URLDownloader* downloader, mp::Path cache_dir_path,
+                                                                 mp::Path data_dir_path, mp::days days_to_expire)
+{
+    const auto& driver = utils::get_driver_str();
+    if (driver == QStringLiteral("lxd"))
+        return std::make_unique<LXDVMImageVault>(image_host);
+    else
+        return std::make_unique<DefaultVMImageVault>(image_host, downloader, cache_dir_path, data_dir_path,
+                                                     days_to_expire);
+}
```
Saviq, line 139:
```suggestion
std::unique_ptr<mp::VMImageVault> mp::platform::make_image_vault(std::vector<mp::VMImageHost*> image_hosts,
                                                                 mp::URLDownloader* downloader, mp::Path cache_dir_path,
                                                                 mp::Path data_dir_path, mp::days days_to_expire)
{
    const auto& driver = utils::get_driver_str();
    if (driver == QStringLiteral("lxd"))
        return std::make_unique<LXDVMImageVault>(image_hosts);
    else
        return std::make_unique<DefaultVMImageVault>(image_hosts, downloader, cache_dir_path, data_dir_path,
                                                     days_to_expire);
}
```

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: I think the only way to make that work and still be generic enough for all image vault & backend driver cases would be to refactor/expand `URLDownloader` to use the new `NetworkManager` class.

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -113,21 +116,40 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(c
...
-            const auto prefix = derive_unpacked_file_path_prefix_from(image_location);
-            const auto kernel_location = prefix + "-vmlinuz-generic";
-            const auto initrd_location = prefix + "-initrd-generic";
+            const auto& driver = utils::get_driver_str();
+
+            QJsonObject image;
+            QString sha256, image_location, kernel_location, initrd_location;
+            int size = -1;
+
+            if (driver == "lxd")
```
Saviq, line 125:
```suggestion
            // TODO: make this a VM factory call with a preference list
            if (driver == "lxd")
```

townsend2010:
@Saviq,

Thanks for the first review.

> Need a lxd plug, don't we?

Yep

> Need to prune multipass find.

I'm pretty sure I already did that for cloud images.  I know I have not accounted for our custom repository yet.  Is this what you are seeing?

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -113,21 +116,40 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(c
...
+            {
+                image = items["lxd.tar.xz"].toObject();
+                sha256 = image["combined_disk1-img_sha256"].toString();
+
+                if (sha256.isEmpty())
+                {
+                    continue;
+                }
```
Saviq, line 133: I think this?
```suggestion
                image = items["disk-kvm.img"].toObject();
                sha256 = image["sha256"].toString();
```

tests/lxd/test_lxd_backend.cpp
```diff
@@ -0,0 +1,602 @@
...
+
+namespace
+{
+using LXDInstanceStatusParamType = std::pair<QByteArray, mp::VirtualMachine::State>;
+
+struct LXDBackend : public Test
+{
+    LXDBackend()
+        : socket_path{QString("%1/test_socket").arg(data_dir.path())},
+          test_server{mpt::MockLocalSocketServer(socket_path)},
```
Saviq, line 52: Should we not test this on the access manager level?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: > I'm pretty sure I already did that for cloud images. I know I have not accounted for our custom repository yet. Is this what you are seeing?

This is probably what could achieve this?

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Hmm, right, I think using a `wait` with no or a long-ish timeout above would be just as effective.  I observed that LXD can take quite a long time uncompressing an image for the first time.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
Saviq: I mean that this should be `mp::FetchType::ImageOnly`. As I said elsewhere, I don't think we should encode anything LXD here, since you re-implemented the image vault.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
Saviq: Ekhm… [_you_ didn't](https://github.com/canonical/multipass/blame/c86c49003ab7dc9728306a20b21fb7a131ea9420/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp#L57-L66)…

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: They _might_ be TODO.  LXD can also handle this in and of itself, so should we config LXD to do this for us or we drive it?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: I just mean that, ideally, we'd have a shared manager for everything LXD… Not saying that URLDownloader should, too.

My rationale being that we could then set it up once (especially with SSL and authentication) and share it through all LXD components. Also saving memories.

src/simplestreams/simple_streams_manifest.cpp
townsend2010: Hmm, the `disk-kvm.img` is not available for 18.04 which is why I chose the `lxd.tar.xz`type.  And is LXD able to use that type at all?  I'm pretty sure LXD will not like it, but I can try.

tests/lxd/test_lxd_backend.cpp
townsend2010: We already unit test the access manager itself in other tests, so I think mocking it as high up as possible is the best way to go for testing the backend.  I sense you disagree with this approach, so am I missing some benefits if we use the access manger as well?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: The problem I'm seeing here is that `DaemonConfig` is responsible for setting up the image vault and setting up the backend driver and does that as two separate steps that are only coupled by `URLDownloader`.  In order for `LXDVMImageVault` and `LXDVirtualMachineFactory` to share a `NetworkAccessManager` object would be to somehow relate those two, but do it in a way that `DaemonConfig` does not have to account for backend and image host differences.  I think we want `DaemonConfig` to be a generic as possible and we achieve that via the `platform` calls, but still, those platform calls do not couple the image vault and the backed driver.

If we don't go the `URLDownloader` way, then we'd need to refactor the `VirtualMachineFactory` to accept the `VMImageVault` object (at least a pointer to it) and then each factory and image vault can determine their relationship to one another.  Some would just ignore each other like now and we'd basically be forcing it hold at least a pointer to the vault and others like LXD would have a fairly coupled relationship like sharing the `NetworkAcessManager`, etc.

I'll think about it some more, but I don't think it will be trivial.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Oh...lol

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Right

include/multipass/fetch_type.h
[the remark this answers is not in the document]
townsend2010: Right, right, no need for anything new here.

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Well, still, the caller should determine _how_ it waits on an operation to complete.  I suppose we could have a default wait logic in `lxd_request` and if the caller doesn't want or need to use that, then it can pass in `0` for the timeout and `lxd_request` will just return the json response from LXD.  WDYT?

include/multipass/fetch_type.h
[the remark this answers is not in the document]
townsend2010: Done

src/daemon/daemon.cpp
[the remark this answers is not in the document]

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/daemon/default_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/platform/backends/lxd/lxd_request.cpp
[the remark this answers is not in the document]
townsend2010: Done

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: done

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Changed to `ImageOnly`

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Basic NOOPs now:)

src/platform/backends/lxd/lxd_virtual_machine.cpp
Saviq: That's roughly what I had in mind. Maybe not inferred from `timeout == 0` but rather something explicit like `bool wait`. Ideally something that resembles something already existing :)

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+mp::FetchType mp::LXDVirtualMachineFactory::fetch_type()
+{
+    return mp::FetchType::ImageOnly;
+}
+
+mp::VMImage mp::LXDVirtualMachineFactory::prepare_source_image(const mp::VMImage& /* source_image */)
+{
+    return VMImage{};
```
Saviq, line 59: Not `return source_image`?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: Bear in mind none of the comments I made here are my requirement, or stuff to be fixed in this PR. It's just driving my thinking about the architecture, since this is pushing what we had in place.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: If we can make it DTRT without our involvement, probably best.

src/simplestreams/simple_streams_manifest.cpp
Saviq: `lxd.tar.xz` is for containers, not VMs. I've asked what image types they're looking for.

tests/lxd/test_lxd_backend.cpp
Saviq: I mean to _mock_ the access manager, rather than talking through a test socket (which was the correct thing to do for testing the manager itself).

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Hmm, ok, I'll basically put it back to how you originally had it :grin:

src/simplestreams/simple_streams_manifest.cpp
townsend2010: Hmm, I think both are valid for VM's.  Looking through the LXD code, it looks like they prefer the `disk-kvm.img` product, but if that is not available, it looks for `combined_disk1-img_sha256` under the `lxd.tar.xz` product.  I will do the same:)

tests/lxd/test_lxd_backend.cpp
townsend2010: Duh on my end.  I will do that to simplify this.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Ok, then I will look into that and failing being able to have LXD control all of that, I'll implement it here.

src/simplestreams/simple_streams_manifest.cpp
Saviq: I stand corrected, they seem to be doing special stuff.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Sure, and I totally agree with the idea of being able to couple the image vault and vm factory when needed.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Yep, that is what I'll do for now until we can support the custom repository and http/file based launches.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Right, fixed

tests/lxd/test_lxd_backend.cpp
townsend2010: Looking at this a bit more, what will need to happen is to make `LocalAccessManager` a non-`final` class and then create a mock `LocalAccessManager` class and have it respond how we want.  Then we'll need to modify `LXDVirtualMachine` and `LXDVMImageVault` to accept a `LocalAccessManager` object (ie, the mocked object) and not create one in the ctor.

Then we can get rid of the `MockLocalSocketServer` and just put that logic back into the `test_local_network_access_manager.cpp` file.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: After looking at this, the image host itself has no knowledge if `stream_location` is empty or now.  It would have to iterate through each image that the image host knows about to see if `stream_location` is known.  I can do this, but it's just going to be a hack for now.

townsend2010: In order to match what other platforms do, I implemented this in `mp::platform::is_remote_supported()` and that takes care of the remotes part.  I then added logic to throw in `fetch_image()` for URL based launches.

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Done

townsend2010:
@Saviq,

Hey, I've updated this to include the `lxd` plug, filter the supported remotes, and to create the LXD "multipass" project along with our own network bridge.

I know there are a number of your comments and other things still to address, but this just might be ready for experimental release:)  I can then follow up addressing those things along with more test coverage.

tests/lxd/test_lxd_backend.cpp
townsend2010: Even some more diving into this and really, doing it the current way where we talk over a local socket to a "mocked" server is the most straightforward way to go.  If we went the route of mocking `NetworkAccessManager` itself, we'll basically need to reimplement that along with `LocalSocketReply` and that doesn't really make much sense.  I think leaving it the way it is is the right thing to do in order not to introduce much more complexity in testing for the sake of not using a temporary local socket.

src/platform/platform_linux.cpp
```diff
@@ -148,10 +165,18 @@ bool mp::platform::is_alias_supported(const std::string& alias, const std::strin
 
 bool mp::platform::is_remote_supported(const std::string& remote)
 {
+    const auto& driver = utils::get_driver_str();
+    // TODO: Remove this constraint once we support other image remotes for hte LXD backend
+    if (driver == QStringLiteral("lxd") && (remote != "release" && remote != "daily"))
```
Saviq, line 170: Shouldn't this also test for `remote.isEmpty()`?

townsend2010: Yep, will fix.

Saviq:
We'll need to handle this on startup:
```console
cze 01 12:48:46 michal-laptop multipassd[767819]: Requesting LXD: GET unix:///var/snap/lxd/common/lxd/unix.socket@1.0/projects/multipass
cze 01 12:48:46 michal-laptop multipassd[767819]: Caught an unhandled exception: Cannot connect to /var/snap/lxd/common/lxd/unix.socket: 0
cze 01 12:48:46 michal-laptop systemd[1]: snap.multipass.multipassd.service: Main process exited, code=exited, status=1/FAILURE
```

townsend2010:
@Saviq,

> We'll need to handle this on startup:

Yes, the health check needs a complete overhaul.  Is that something that needs to be in a alpha/preview/experimental release of this?

Saviq:
> Is that something that needs to be in a alpha/preview/experimental release of this?

It's probably ok for now.

townsend2010:
@Saviq,

Ack, then I'll put this on my list of TODO's for the next release of this :grin:

src/platform/platform_linux.cpp
Saviq: Yeah, it shouldn't be ;)

Saviq:
I say:

bors merge

PR #1537, opened 2020-05-15
[detail] Use multi-thread-capable signal masking

### LINKED ISSUE #1538: SSHFS server only blocks signals in main thread
The sshfs server wants to [handle signals in the main thread](https://github.com/canonical/multipass/blob/7a148b009878f9af7a9ea6d78c4186d8f7b398ba/src/sshfs_mount/sshfs_server.cpp#L99). However, the mask is applied in one of two running threads. Until #1539 is fixed, it is unspecified what actually happens (here and elsewhere). After, the signals are not blocked if delivered to the sshfs thread.

This should be done in separate steps:
  1. While there is only one thread, apply signal mask (to have it applied in ensuing threads)
  2. Wait for the signals in the main thread, only after it is done creating other threads

Since this is platform dependent, an approach would be for `mpp::wait_for_quit_signals` to return a std::function to wait for signals, rather than waiting on them directly. The caller could then choose when and where to wait.

### LINKED ISSUE #1539: Signal masking relies on "unspecified behavior"
From the man page:

> The use of sigprocmask() is unspecified in a multithreaded process; see pthread_sigmask(3).

ricab:
Fixes #1539. Fixes #1538.

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -28,6 +28,7 @@
 
 #include <QString>
 
+#include <functional>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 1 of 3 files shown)


ricab:
Noticed this while browsing code.

ricab:
#1538 is a related but separate issue

ricab:
Well, I'll actually just deal with both issues here.

PR #1559, opened 2020-05-26
[daemon] prefer the `wheel` group

Saviq:


abitrolly:
I confirm this fixes `multipass socket access denied` error on Fedora 32.

townsend2010:
Welp, this messes up socket permission on Mac, ie, no worky :frowning_face:

townsend2010:
I should add that the `wheel` group does exist on MacOS, but the user is not a member of that group.  I think this should be split out into a platform independent method of obtaining the preferred group.

PR #1564, opened 2020-05-28
[platform] Use a platform-independent way of getting supported groups

### LINKED ISSUE #1563: [osx] Permission denied when trying to access the multipass socket
After merging in #1559, Multipass on Mac no gets a "permission denied" error when trying to access the multipass socket.  This is due to the `wheel` group existing on OSX, but users are not part of that group by default.

We should have a platform independent way of determining the group the socket will be owned by.

townsend2010:
- Add simple test for coverage.
- Also, some "drive by" header clean up. Sorry:)

Fixes #1563


townsend2010:
Closing this as #1565 is a less complicated fix and should take care of it.

PR #1566, opened 2020-05-29
[snap] pass daemon arguments in `launch-multipassd`

Saviq:
This is so that you can override `--verbosity` with a systemd `.override` file.


luis4a0:
bors +r

PR #1571, opened 2020-06-03
Hotkey config

### LINKED ISSUE #1311: Keyboard shortcut conflicting with Chrome "View Source" under macOS
> "Use a keyboard shortcut (Ctrl+Alt+U or ⌘⌥U) to open a shell in your primary instance" 

introduced in 1.0.0. This is basically a very nice feature, but it should not be hardcoded but configurable.

Web developers tend to use "View Source" in Chrome, which has the exact same shortcut and is now conflicting. Multipass has now precedence and the shortcut in Chrome is no longer working.

ricab:
Fixes #1311. Also introduces `get --raw`.

### DIFF
--- a/src/platform/platform_shared.cpp
+++ b/src/platform/platform_shared.cpp
@@ -30,6 +30,18 @@ QString multipass::platform::default_hotkey()
     return QStringLiteral("Ctrl+Alt+U");
 }
 
+// A few notes on this:
+// 1) Some shortcuts may feel counter-intuitive. For example in a keyboard where pressing "shift+-" produces an
+// underscore, "_" is still interpreted the same as "-". IOW, "shift+-" == "shift+_" != "_" (just like "u" is the same
+// as "U").
+// 2) QHotKey fails to register some of the shortcuts we accept here (e.g. "Media Play").
+// 3) QKeySequence seems to have problems with AtlGr. QKeySequence("AltGr").toString() prints rubbish (that it does not
+// interpret back to mean the same thing). Unfortunately it is not enough to specify "ú" when that's what the layout
+// produces for AltGr+U. The Sequence "ú" is accepted and QHotKey registers it, but is gets triggered on "U" and not
+// "AltGr+U".
+// 4) There does not seem to be a way to specify numpad keys in QKeySequence (with or without numlock).
+// 5) meta only seems to work with other modifiers (e.g. ctrl+meta+x works, but meta+x doesn't even though it is
+// accepted with no warning)
 QString mpp::interpret_general_hotkey(const QString& val)
 {
     auto sequence = QKeySequence{val};
--- a/src/platform/platform_shared.cpp
+++ b/src/platform/platform_shared.cpp
@@ -27,7 +27,7 @@ namespace mpp = multipass::platform;
 
 QString multipass::platform::default_hotkey()
 {
-    return QStringLiteral("Ctrl+Alt+U");
+    return QKeySequence{QStringLiteral("Ctrl+Alt+U")}.toString(QKeySequence::NativeText); // Cmd+Opt+U on mac
 }
 
 // A few notes on this:
@@ -45,7 +45,7 @@ QString multipass::platform::default_hotkey()
 QString mpp::interpret_general_hotkey(const QString& val)
 {
     auto sequence = QKeySequence{val};
-    auto ret = sequence.toString();
+    auto ret = sequence.toString(QKeySequence::NativeText);
 
     if (ret.isEmpty() && !sequence.isEmpty())
         throw InvalidSettingsException(mp::hotkey_key, val, "Invalid key sequence");
--- a/src/platform/platform_shared.cpp
+++ b/src/platform/platform_shared.cpp
@@ -36,10 +36,9 @@ QString mpp::interpret_general_hotkey(const QString& val)
     auto ret = sequence.toString();
 
     if (ret.isEmpty() && !sequence.isEmpty())
-        throw InvalidSettingsException(mp::hotkey_key, val, "Invalid key sequence"); // TODO@ricardo test
+        throw InvalidSettingsException(mp::hotkey_key, val, "Invalid key sequence");
     if (sequence.count() > 1)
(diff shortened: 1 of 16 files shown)


ricab:
@Saviq what do you think about this? I am going a bit further than we discussed by tentatively placing things at the platform level, so that we can handle the mac case properly, rather than tell users they need to use ctrl for cmd and meta for ctrl. 

Refactoring needed to use current linux approach on win.

Saviq:
How do you think they'll be inputting Command and Option? Cmd and Opt?

ricab:
I was thinking something like that. Not sure about Opt. Here's what they use: https://support.apple.com/en-us/HT201236

ricab:
But I suppose we can accept both opt and alt.

ricab:
I have found that both QKeySequence and QHotKey have [a few rough corners](https://github.com/canonical/multipass/blob/a0cd5ff8bae5fe0020ceaa3bb485392d087b405f/src/platform/platform_shared.cpp#L33-L44). The altgr and meta issues are particularly unfortunate, because they don't work but there is no error whatsoever. The numpad issue too. 

The more I think about this, the more it feels like we should restrict the values to a subset that we know we can support well. Perhaps something like:

- one of [F1-F12], or
- one or more of [ctrl, alt, shift, meta], but not meta alone, plus any non-modifier key, excluding numpad, media keys, Fn, altgr, backspace, enter, caps lock, ... oh dear, how do we even document? We can't say [a-z0-9F1-F12] because that still excludes a bunch of symbols that work well but depend on keyboard layout (e.g. '.').

ricab:
OK this should be about done now. Only need a final test on windows and probably some rebasing.

src/platform/platform_shared.cpp
```diff
@@ -0,0 +1,51 @@
...
+// accepted with no warning)
+QString mpp::interpret_general_hotkey(const QString& val)
+{
+    auto sequence = QKeySequence{val};
+    auto ret = sequence.toString(QKeySequence::NativeText);
+
+    if (ret.isEmpty() && !sequence.isEmpty())
+        throw InvalidSettingsException(mp::hotkey_key, val, "Invalid key sequence");
+    if (sequence.count() > 1)
+        throw InvalidSettingsException(mp::hotkey_key, val, "Multiple key sequences are not supported");
```
Saviq, line 48: Should we not try and compare `val` to `ret`? Or are they not guaranteed to be the same?

src/platform/platform_shared.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#ifndef MULTIPASS_PLATFORM_SHARED_H
+#define MULTIPASS_PLATFORM_SHARED_H
+
+#include <QString>
+
+namespace multipass::platform
+{
+QString interpret_general_hotkey(const QString& val);
```
Saviq, line 25: Why `_general_`?

src/utils/settings.cpp
```diff
@@ -42,11 +44,17 @@ const auto client_root = QStringLiteral("client");
 const auto petenv_name = QStringLiteral("primary");
 const auto autostart_default = QStringLiteral("true");
 
+QString default_hotkey()
+{
+    return QKeySequence{mp::hotkey_default}.toString(QKeySequence::NativeText); // outcome depends on platform
```
Saviq, line 49: Oh so will this translate to "Cmd" and "Opt" on macOS?

Saviq:
I wonder if we should display the hotkey in the GUI, then?

src/platform/platform_shared.cpp
```diff
@@ -0,0 +1,51 @@
...
+// underscore, "_" is still interpreted the same as "-". IOW, "shift+-" == "shift+_" != "_" (just like "u" is the same
+// as "U").
+// 2) QHotKey fails to register some of the shortcuts we accept here (e.g. "Media Play").
+// 3) QKeySequence seems to have problems with AtlGr. QKeySequence("AltGr").toString() prints rubbish (that it does not
+// interpret back to mean the same thing). Unfortunately it is not enough to specify "ú" when that's what the layout
+// produces for AltGr+U. The Sequence "ú" is accepted and QHotKey registers it, but is gets triggered on "U" and not
+// "AltGr+U".
+// 4) There does not seem to be a way to specify numpad keys in QKeySequence (with or without numlock).
+// 5) meta only seems to work with other modifiers (e.g. ctrl+meta+x works, but meta+x doesn't even though it is
+// accepted with no warning)
```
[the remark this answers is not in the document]
ricab: I am not sure TBH, those were conclusions from testing on Linux. I got some of those same results on the others, but I did not check each one on every platform. The items referring to `Meta` and `AltGr` have doubtful applicability on macOS (at least with the default keyboard).

src/platform/platform_shared.cpp
ricab: They're not guaranteed to be the same, "alt+ctrl+u" is a valid input that gets interpreted as "Ctrl+Alt+U".

src/platform/platform_shared.h
ricab: LOL, yeah indeed :smile: I'll remove that. 

When I started this was not in a separate header and was called by the actual platform functions to interpret the hotkey, so it was a way to distinguish.

src/utils/settings.cpp
ricab: Yup :wink: Well, actually to ⌘⌥U

ricab:
> I wonder if we should display the hotkey in the GUI, then?

Makes sense indeed.

src/platform/platform_shared.h
ricab: Done.

Saviq:
Alright, nice!

bors merge

Saviq:
Conflict on macOS:

```
CONFLICT (content): Merge conflict in src/platform/CMakeLists.txt
```

ricab:
> Conflict on macOS:
> 
> ```
> CONFLICT (content): Merge conflict in src/platform/CMakeLists.txt
> ```

Hi, that's because bors does not grab the matching branch when staging, right?

PR #1574, opened 2020-06-04
Additional LXD driver fixes and more tests

townsend2010:
Adds some LXD driver backend fixes along with support for tests.
More tests to increase coverage.


townsend2010:
Making this a draft again as I'd like to propose a change first that will allow `LXDVirtualMachineFactory` and `LXDVMImageVault` to use one `NetworkAccessManager` and that will affect this PR.

townsend2010:
I have #1587 up now.  Let's get that in before proceeding with this PR.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -109,6 +76,39 @@ void mp::LXDVirtualMachineFactory::hypervisor_health_check()
...
+    {
+        lxd_request(manager.get(), "GET",
+                    QUrl(QString("%1/networks/%2").arg(base_url.toString()).arg(multipass_bridge_name)));
+    }
+    catch (const LXDNotFoundException&)
+    {
+        QJsonObject network{{"name", multipass_bridge_name}, {"description", "Network bridge for Multipass"}};
+
+        lxd_request(manager.get(), "POST", QUrl(QString("%1/networks").arg(base_url.toString())), network);
+    }
```
Saviq, line 111: This feels weird here… Should we instead have a `VirtualMachineFactory::initialize` that would only be called after the health checks pass?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -81,7 +81,7 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
     {
         auto instance_info = lxd_request(
             manager, "GET",
-            QUrl(QString("%1/virtual_machines/%2").arg(base_url.toString()).arg(QString::fromStdString(query.name))));
+            QUrl(QString("%1/virtual-machines/%2").arg(base_url.toString()).arg(QString::fromStdString(query.name))));
```
Saviq, line 84: How did that work? o.O

Or didn't it?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -226,11 +228,13 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
 
 void mp::LXDVMImageVault::prune_expired_images()
 {
+    mpl::log(mpl::Level::trace, category, "Pruning expired images not implemented");
 }
 
 void mp::LXDVMImageVault::update_images(const FetchType& fetch_type, const PrepareAction& prepare,
                                         const ProgressMonitor& monitor)
 {
+    mpl::log(mpl::Level::trace, category, "Updating images not implemented");
```
Saviq, line 237: I think I'd put these at `debug`, but :shrug:.

tests/lxd/test_lxd_image_vault.cpp
```diff
@@ -0,0 +1,392 @@
...
+        {
+            if (url.contains("1.0/images/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"))
+            {
+                return new mpt::MockLocalSocketReply(mpt::image_info_data);
+            }
+        }
+        else if (op == "POST" && url.contains("1.0/images"))
+        {
+            // It should not try to request an image download
+            ADD_FAILURE();
```
Saviq, line 197: I believe this works:

```suggestion
            ADD_FAILURE() << "Image download shouldn't be requested";
```

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: I believe making sure the project and the network bridge exist during the health check is exactly what the health check is for.  My assertion is what happens if a user deletes the project or network while multipass is running?  We'd want to check for that and create those if missing, right?  Otherwise, we can just pass an error up to the user to either recreate whatever is missing or tell them to restart `multipassd` and it will make sure they are there on start up.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: It didn't :frowning_face:

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: #1594 implements this, so I think messing with this in this PR is unnecessary.

tests/lxd/test_lxd_image_vault.cpp
townsend2010: Sure, adding a helpful message in case this **does** fail is good.

townsend2010:
> Overall I feel like EXPECT_THROW(…, std::runtime_error) is a bit broad, not dissimilar to catching.

Yep, will either use new exception types and/or match on the error string where appropriate.

townsend2010:
I've rebased this on top of https://github.com/canonical/multipass/tree/disable-suspend-support, so we'll need to get that in first.

Saviq:
Not sure if you're aware:
```bash
$ multipass launch
# Ctrl+C
$ multipass delete -p --all
# wait for it…
delete failed: unix:///var/snap/lxd/common/lxd/unix.socket@1.0/operations/30b3847a-1e5f-41bc-ad07-fd8ac688654b/wait?project=multipass: Operation canceled
```

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -81,42 +81,37 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
             }
+
+            return source_image;
         }
     }
     catch (const std::exception&)
     {
-        // Instance doesn't exist so move on
+        // Instance or image doesn't exist so move on
     }
```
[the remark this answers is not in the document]
townsend2010: Hmm, I could.  'lxd_request()` will throw `LXDNotFoundException` which I could catch (and ignore) separately.  However, I would have to modify `UbuntuVMImageHost::info_for_full_hash_impl()` to throw a new type if an image is not found.  Since that affects all other backends, is this the place to add that?

tests/lxd/test_lxd_backend.cpp
```diff
@@ -129,11 +138,59 @@ TEST_F(LXDBackend, creates_project_and_network_on_first_run)
 
     mp::LXDVirtualMachineFactory backend{std::move(mock_network_access_manager), data_dir.path(), base_url};
 
+    backend.hypervisor_health_check();
+
     EXPECT_TRUE(project_created);
     EXPECT_TRUE(profile_updated);
     EXPECT_TRUE(network_created);
```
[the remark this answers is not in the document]
townsend2010: Yeah, it would make for a better test.  I will add that.

townsend2010:
It looks like `LXDBackend.shutdown_while_starting_throws_and_sets_correct_state` may be a bit racy as it hung in ci, but works on my machine.  I'll need to tweak that some.

townsend2010:
> Not sure if you're aware:

Nope, I will look into that and see what's going on.

tests/lxd/test_lxd_backend.cpp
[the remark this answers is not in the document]
townsend2010: Ok, done.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: Actually, if I do add a different exception type when an image isn't found, I would have to modify `DefaultVMImageVault` instead of "all other backends".  Still, is this PR the place to add this since I would be touching an unrelated file?  I would rather add the different exception type in a follow up PR.  That said, I can add another `catch` here for `LXDNotFoundException` in anticipation of this.

townsend2010:
Hmm, regarding the operation time out, I tried it and did not see that.  However, I do think there is a race going on with the race being multipass trying to delete and purge an instance that is in the middle of being initialized by LXD.  The LXD operation that is creating the instance is not cancellable, so I think the best we can do in this situation is to fail the delete for any instances in the creation stage.  I don't think this is a regression in the PR, so I think that should be handled in a follow up.  I can add that to my ever growing list of LXD tasks :wink:

src/platform/backends/lxd/lxd_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: @Saviq 

I added the `LXDNotFoundException` `catch` block.  Let me know what you think about my proposal for the new exception type.

Saviq:
> The LXD operation that is creating the instance is not cancellable, so I think the best we can do in this situation is to fail the delete for any instances in the creation stage.

FWIW in my case it waited for the instance to come up before barfing.

townsend2010:
> FWIW in my case it waited for the instance to come up before barfing.

Meh, it just seems to depend on which stage of launch things are in.  When I did it, the `delete` "worked" in that it returned quickly and `list` showed no instances.  However, behind the scenes, LXD was busy still initializing the instance and once it was done, a subsequent `list` showed the instance up and running. :man_shrugging:

Saviq:
A'ighty!

bors merge

PR #1580, opened 2020-06-08
[streams] don't die on missing products

Saviq:
If products are missing, don't fail, but log info instead.

It may be a valid situation on LXD at the moment.

### DIFF
--- a/src/daemon/ubuntu_image_host.cpp
+++ b/src/daemon/ubuntu_image_host.cpp
@@ -22,6 +22,7 @@
 #include <multipass/url_downloader.h>
 
 #include <multipass/exceptions/download_exception.h>
+#include <multipass/exceptions/empty_manifest_exception.h>
 #include <multipass/exceptions/unsupported_image_exception.h>
 
 #include <multipass/format.h>
@@ -229,7 +230,11 @@ void mp::UbuntuVMImageHost::fetch_manifests()
             manifests.emplace_back(
                 std::make_pair(remote.first, download_manifest(QString::fromStdString(remote.second), url_downloader)));
         }
-        catch (mp::DownloadException& e)
+        catch (mp::EmptyManifestException& /* e */)
+        {
+            on_manifest_empty(fmt::format("Did not find any supported products in \"{}\"", remote.first));
+        }
+        catch (std::runtime_error& e)
         {
             on_manifest_update_failure(e.what());
         }
--- a/src/daemon/ubuntu_image_host.cpp
+++ b/src/daemon/ubuntu_image_host.cpp
@@ -22,7 +22,7 @@
 #include <multipass/url_downloader.h>
 
 #include <multipass/exceptions/download_exception.h>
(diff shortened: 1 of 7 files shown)


include/multipass/exceptions/empty_stream_exception.h
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_EMPTY_STREAM_EXCEPTION_H
+#define MULTIPASS_EMPTY_STREAM_EXCEPTION_H
+
+#include <stdexcept>
+
+#include <multipass/format.h>
```
townsend2010, line 23: I don't see a reason to include this here.

include/multipass/exceptions/empty_stream_exception.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace multipass
+{
+class EmptyStreamException : public std::runtime_error
```
townsend2010, line 27: Just wondering if we should name this `EmptyManifestException` as we have tended to refer to the data as a "manifest" in other places?

src/daemon/ubuntu_image_host.cpp
```diff
@@ -229,7 +230,11 @@ void mp::UbuntuVMImageHost::fetch_manifests()
             manifests.emplace_back(
                 std::make_pair(remote.first, download_manifest(QString::fromStdString(remote.second), url_downloader)));
         }
-        catch (mp::DownloadException& e)
+        catch (mp::EmptyManifestException& /* e */)
+        {
+            on_manifest_empty(fmt::format("Did not find any supported products in \"{}\"", remote.first));
+        }
+        catch (std::runtime_error& e)
```
ricab, line 237: Hi, shouldn't this be DownloadException?

townsend2010: I saw this, but thought the intent was to catch all other errors instead with the change.  Let's see what @Saviq says :grin:

Saviq: No, the intent was to not bail on any of the errors (one broken remote shouldn't cause all the other ones).

The EmptyManifestException was needed to not warn about that particular case.

ricab: Uh, I must be looking at this wrong, but isn't a `DownloadException` a `runtime_error`? So `DownloadException`s will still be treated as before, but now so will any other `runtime_error`s? Or was the error you wanted to continue on not a `DownloadException` in the first place? In that case, it should be another specific error, so that we're not catching undesired stuff here, right?

Saviq: Sure, I agree catching `std::runtime_error` is dangerously broad. It's just that's what `SimpleStreamsManifest::fromJson` is throwing. I'll fix.

bors cancel

PR #1587, opened 2020-06-11
[lxd] Use a single NetworkAccessManager object

townsend2010:
- All LXD backend classes will share a single NetworkAccessManager object owned by LXDVirtualMachineFactory.

### DIFF
--- a/src/daemon/daemon_config.cpp
+++ b/src/daemon/daemon_config.cpp
@@ -132,7 +132,7 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
             hosts.push_back(image.get());
         }
 
-        vault = platform::make_image_vault(
+        vault = factory->create_image_vault(
             hosts, url_downloader.get(),
             mp::utils::backend_directory_path(cache_directory, factory->get_backend_directory_name()),
             mp::utils::backend_directory_path(data_directory, factory->get_backend_directory_name()), days_to_expire);
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -55,8 +55,6 @@ QString daemon_config_home();                      // temporary
 
 bool is_backend_supported(const QString& backend); // temporary
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
-std::unique_ptr<VMImageVault> make_image_vault(std::vector<VMImageHost*> image_host, URLDownloader* downloader,
-                                               Path cache_dir_path, Path data_dir_path, days days_to_expire);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
 std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& config);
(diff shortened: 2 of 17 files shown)


src/daemon/daemon_config.cpp
```diff
@@ -132,10 +133,19 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
+            vault = platform::make_image_vault(factory.get(), driver, hosts);
+        }
+        else
+        {
+            vault = std::make_unique<DefaultVMImageVault>(
+                hosts, url_downloader.get(),
+                mp::utils::backend_directory_path(cache_directory, factory->get_backend_directory_name()),
+                mp::utils::backend_directory_path(data_directory, factory->get_backend_directory_name()),
+                days_to_expire);
+        }
```
Saviq, line 148: I wonder if the Factory itself could provide the VMImageVault? To avoid the `if (driver…)` bit?

townsend2010: @Saviq,

I thought about that, but then we'd have to modify *all* 6 supported platform factories in order for them to provide the `VMImageVault`.  That seemed like a bunch of work just to make the LXD situation a little cleaner.

Saviq: Well, yeah, but the alternative being special-casing the lxd driver in a bunch of places…

I long suggested we should have `Base*` default implementations of things like the VM factory, maybe this is the right time to introduce it?

townsend2010: I, too, am a fan of defining a base implementation with each factory able to override any parts of it as it sees fit.  Of course, this goes against the original philosophy of how the various classes define the interface and then the platform itself defines the implementation.  But given that we keep extending this further and further with new backends, etc., I think that doesn't make sense for the cases of code duplication that each implementation shares.

That said, we do have an example in the code where we do break from the original philosophy where we have a base `CommomVMImageHost` class (based off of `VMImageHost`) and `UbuntuVMImageHost` and `CustomVMImageHost` derive from that.

Since this a fairly significant break with (most of) the current code philosophy, I'd like ask @ricab and @luis4a0 their opinions on this. (I know @luis4a0 is on holiday, so his response may be late to the game :grinning: )

Saviq: TBH I don't see this as a departure from the philosophy. We still *only* define the interface first. Any shared implementation comes later, either in the form of a base class, or maybe mixins.

townsend2010: Ok, then I'm not sure why we haven't done this sooner :grin: At any rate, then sure, I'm all for defining a shared base class implementation and going from there.  I will see if it makes more sense to do that in a separate PR or do it here and rename the branch, title, etc.

townsend2010: This is fixed now.  The first part of this is in #1609 and this PR has been updated to commonize creating the image vault.

townsend2010:
Until #1609 is merged, here is the diff of this branch: https://github.com/canonical/multipass/compare/add-base-vm-factory-class...use-one-network-access-manager

townsend2010:
This is ready for another round of review :smile:

Saviq:
Lookin' good.

bors merge

PR #1588, opened 2020-06-11
[lxd] Add support for suspending instances

townsend2010:


townsend2010:
@Saviq,

> This doesn't survive reboots.

You mean if you "suspend", ie, freeze a VM, LXD doesn't remember the last state on host reboot?

> It really only does what it says on the tin - freezes the VM, doesn't store anything to disk.

That's a deficiency in LXD.  I plan to follow up with the LXD team, but this is the best LXD gives us for this.

Saviq:
> > 
> 
> That's a deficiency in LXD. I plan to follow up with the LXD team, but this is the best LXD gives us for this.

Sure, and we should error out on `multipass suspend` until such a time that it actually works with LXD.

townsend2010:
> we should error out on multipass suspend until such a time that it actually works with LXD

Oh, ok, I thought partial support was better than erroring out, but I'm fine with that.  I'll fix up this PR to strip out the suspend logic and throw the error instead.

townsend2010:
This is replaced by #1604.

PR #1599, opened 2020-06-16
[lxd] Fix slow startup of daemon when getting image info

townsend2010:
- Use LXD's store image info instead of querying SimpleStreams.
- Make getting SimpleStreams manifest updates run in a separate thread.


townsend2010:
This is superseded by #3153.

PR #1603, opened 2020-06-17
[cosmetic] Reorder util functions into small lumps

ricab:
Hopefully easier to sift through this way.


ricab:
@townsend2010 how about now?

ricab:
Gahh we shouldn't have such differences outside of platform

ricab:
@townsend2010 I aligned the utils with the other side: brought extra ones from there and also added a missing log message. They are not used here ATM but they could be, so I think this makes more sense than tying to any platform.

townsend2010:
@ricab,

I totally agree that the utils shouldn't be platform/branch dependent, thanks for fixing that- it's been bugging me for some time now and just never did it myself :smiley:

townsend2010:
bors merge

PR #1604, opened 2020-06-17
[lxd] Disable suspend support

townsend2010:


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -204,10 +204,6 @@ mp::LXDVirtualMachine::LXDVirtualMachine(const VirtualMachineDescription& desc,
 
 mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
-    update_suspend_status = false;
-
-    if (current_state() == State::running)
-        suspend();
```
Saviq, line 210: (I know I answered a question about this differently before… sorry…)

Wouldn't shutting it down be more consistent? The only other platform that can't suspend (hyperkit) does that. I know, because we're the parent anyway - but leaving VMs running feels wrong.

townsend2010: @Saviq,

Sure, I have no problem doing that.

townsend2010: @Saviq,

Actually, having thought about it some more on a walk, I don't think stopping the instance on `multipassd` shutdown is the right thing to do.  If `snapd` decides to update Multipass (or any other reason the daemon may shutdown), I don't want whatever a user is doing in (an) instance(s) to be lost, so either we should leave it running or freeze it.

Saviq: IMO "`snapd` deciding to update" is much different than "any other reason"… The other reason for a clean shutdown would be `snap stop multipass`. We don't want VMs to stay on.

But for refresh, we can have a refresh hook that stores a file in `$SNAP_DATA` and we can NOOP on shutdown based on that. The launch script should then clear that file.

townsend2010: Ok, I see what you mean.  I guess you're most concerned about leaving any `qemu` processes running when the intention is to shut down Multipass for some period of time, right?

It's difficult, from the daemon's perspective, to understand why it's being shut down.  The refresh stuff you suggest does help with that case.  However, I think that should be for a different PR and this PR will unilaterally shut down any running instances when the daemon exits.

townsend2010: This is done.

tests/lxd/test_lxd_backend.cpp
```diff
@@ -256,6 +260,57 @@ TEST_F(LXDBackend, machine_persists_and_sets_state_on_shutdown)
...
+            outgoingData->open(QIODevice::ReadOnly);
+            auto data = outgoingData->readAll();
+            auto op = request.attribute(QNetworkRequest::CustomVerbAttribute).toString();
+            auto url = request.url().toString();
+
+            if (op == "GET")
+            {
+                if (url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164"))
+                {
+                    vm_shutdown = true;
```
Saviq, line 280: Is this actually `vm_shutdown`? And not… `waiting`?

tests/lxd/test_lxd_backend.cpp
```diff
@@ -256,6 +260,57 @@ TEST_F(LXDBackend, machine_persists_and_sets_state_on_shutdown)
...
+                    else
+                    {
+                        return new mpt::MockLocalSocketReply(mpt::vm_state_fully_running_data);
+                    }
+                }
+            }
+            else if (op == "PUT" && url.contains("1.0/virtual-machines/pied-piper-valley/state") &&
+                     data.contains("stop"))
+            {
+                return new mpt::MockLocalSocketReply(mpt::stop_vm_data);
```
Saviq, line 298: Shouldn't we track `stop_requested` here?

tests/lxd/test_lxd_backend.cpp
```diff
@@ -531,32 +602,36 @@ TEST_F(LXDBackend, lxd_request_wrong_json_throws)
...
+        });
 
     mp::LXDVirtualMachine machine{default_description, stub_monitor, mock_network_access_manager.get(), base_url};
 
-    ASSERT_EQ(machine.current_state(), mp::VirtualMachine::State::stopped);
-
-    machine.suspend();
-
-    EXPECT_EQ(machine.current_state(), mp::VirtualMachine::State::stopped);
+    EXPECT_THROW(machine.suspend(), std::runtime_error);
```
Saviq, line 631: Don't want to be a bit more selective? Check the message?

townsend2010: You can't (easily) check the message when using `EXPECT_THROW`.  You can only check the type.  The only way to more selective to use a different exception type and expect that.

tests/lxd/test_lxd_backend.cpp
townsend2010: So this is just the `bool` to say that the client requested the `GET` of the `operations` URL and that we dictated   the VM did indeed fully shut down.  It's basically used to indicate that the client went through all of its logic to know that the vm is fully shut down.

tests/lxd/test_lxd_backend.cpp
townsend2010: Yeah, we probably should check for that.

tests/lxd/test_lxd_backend.cpp
Saviq: Oh right, the return below says the operation completed, which means the VM is shut down. My bad.

tests/lxd/test_lxd_backend.cpp
Saviq: Yeah, which I commented a bit more about in #1604 review. I feel like this isn't dissimilar to catching a `std::runtime_error` - a bit broad…

We could use the mock logger to check the message, though, couldn't we?

townsend2010: Actually, as @ricab mentioned in chat, there are the exception + message checks in `extra_assertions.h`.  I will use that.

tests/lxd/test_lxd_backend.cpp
townsend2010: Done

tests/lxd/test_lxd_backend.cpp
townsend2010: Ok, I think keeping it a `std::runtime_error` is fine here, but the test now checks the string to make sure it's the expected message since that is really what we care about.

tests/lxd/test_lxd_backend.cpp
```diff
@@ -628,7 +631,10 @@ TEST_F(LXDBackend, unsupported_suspend_throws)
 
     mp::LXDVirtualMachine machine{default_description, stub_monitor, mock_network_access_manager.get(), base_url};
 
-    EXPECT_THROW(machine.suspend(), std::runtime_error);
+    const auto throw_msg_matcher = mp::make_optional(StrEq("suspend is currently not supported"));
```
Saviq, line 634: Why `optional`?

townsend2010: lol, actually, no reason other than I copied that from other code and didn't think about it :smiley:   Will fix.

townsend2010: Ok, fixed this.

PR #1613, opened 2020-06-22
[virtual_machine_factory] Introduce new base class common implementation

townsend2010:
This allows for the various virtual machine factories to use a base class implementation
for defaults to cut down on code reuse.


Saviq:
This is fine.

bors merge

PR #1627, opened 2020-07-01
[travis] use clang-10

Saviq:


.travis.yml
```diff
@@ -41,10 +41,10 @@ jobs:
 
       addons:
         apt:
-          packages: [clang-format-9]
+          packages: [clang-format-10]
```
ricab, line 44: How about we make the dependency explicit?

```suggestion
          packages: [clang-10, clang-format-10]
```

Saviq: No, this bit only does the formatting, it doesn't need the compiler.

Saviq:
> Thank you, this is quite welcome, but the build is bound to fail... We can fix in separate PRs though. One such fix is in #1626 already.

Sure, I actually wanted to ensure it does, indeed, fail.

.travis.yml
ricab: Ah sorry, clang-10 is installed in the other one. I suppose this is just for conformity.

Saviq:
Green!

bors merge a=ricab a=Saviq r=Saviq

PR #1629, opened 2020-07-01
Fix warning-errors in clang build

ricab:


ricab:
Stacked on top of #1627. Changed base to that to check if GH has improved in that regard lately.

Issue #1004, opened 2019-08-28
Use xterm-256color


Paris:
`multipass shell` should set the `$TERM` variable to `xterm-256color` or preferrably the same value of the variable set by the environment where it was invoked, at the moment it's just `xterm` which isn't ideal.

jsilvermoslearn:
is there a workaround for this? a way to export shell env vars that multipass passes to the login shell?

thanks

flying-sausages:
@jsilvermoslearn you can SSH into your VM which will work the way I assume you'd expect

uishon:
Any chance to solve this issue on Windows as well? Just tried it out on 1.4.0 and when I `multipass shell` on Windows, `TERM` is still set to `xterm` and not `xterm-256color`. If I `ssh` to that same multipass instance, `TERM` is set up correctly. But, I prefer using `multipass shell`...

PR #1630, opened 2020-07-02
Switch to ssh_channel_request_pty_size to set TERM explicitly

davidkrauser:
Previously, all Multipass sessions used the default TERM value "xterm". With this change, we pull TERM in from the environment and set it explicitly.

Cleaned up the code that updates the pty size. This makes it simpler to pull in geometry values for the initial call to request a channel.

Also removed unused setup_environment method in Console. I can revert this commit if we want this to stick around.

This fixes: #1004

### DIFF
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -21,32 +21,41 @@
 
 #include <sys/ioctl.h>
 
+#include <cstdlib>
+
 namespace mp = multipass;
 
 namespace
 {
-mp::Console::ConsoleGeometry last_geometry{0, 0};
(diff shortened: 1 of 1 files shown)


Saviq:
Thanks @davidkrauser, we'll check it out and get it in asap :)

Saviq:
bors try

davidkrauser:
Hi @ricab - thanks for the review. I dropped the conflicting commit and pushed up a bugfix for an edge case I missed if TERM is unset :-)

ricab:
Alright, thanks again for this.

bors try

PR #1639, opened 2020-07-13
[lxd/snap] Do not stop instances during snap refresh

townsend2010:


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -179,7 +180,7 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
     update_shutdown_status = false;
 
-    if (state == State::running)
+    if (state == State::running && !QFileInfo::exists(mp::utils::snap_common_dir() + "/snap_refresh"))
```
Saviq, line 183: Shouldn't we `if (mp::utils::is_snap())` first?

townsend2010: I don't think it's necessary since the check here will determine if it's a snap as well.  If it's not a snap, `SNAP_COMMON` won't exist.  I just think adding an `is_snap()` call is just adding an extra call for no benefit, particularly in the case when it is a snap.

Saviq: Actually this would then check for `$CWD/snap_refresh` (I think?), since `snap_common_dir()` doesn't check if `SNAP_COMMON` is empty or not? Sounds like unexpected behaviour to me…

townsend2010: Well, if `$SNAP_COMMON` doesn't exist, ie, is empty, then it's going to look for `/snap_refresh`.  Maybe the better test here is to check if `$SNAP_COMMON` exists and if so, then check if `$SNAP_COMMON/snap_refresh` exists.

townsend2010: I should also add that for what I just proposed, I don't think `utils::is_snap()` is enough.  We should actually check if `utils::snap_common_dir()` returns an empty string or not.

Saviq: IMO `utils::snap*` should error out if we're not running in the snap / the variables are not set.

townsend2010: > IMO utils::snap* should error out if we're not running in the snap / the variables are not set.

Agreed, they should throw some NotASnapException or something (I'll make a better name :grin:), but modifying those utils should be in a different PR.  I'll work on that.

townsend2010: Ok, the snap-throws-on-not-finding-vars PR is now merged.  This PR has been updated to reflect those changes.

PR #1640, opened 2020-07-13
[daemon] skip image computation with LXD

Saviq:


luis4a0:
Passed Travis.

bors r+

PR #1642, opened 2020-07-13
[snap_utils] Throw exception when snap dir env vars are not found

townsend2010:
- When a snap-related directory environment variable is not found, throw the new SnapEnvironmentException exception.

### DIFF
--- a/src/platform/backends/shared/linux/apparmor.cpp
+++ b/src/platform/backends/shared/linux/apparmor.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2019-2020 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -19,6 +19,7 @@
 
 #include "apparmor.h"
 
+#include <multipass/exceptions/snap_environment_exception.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/snap_utils.h>
@@ -48,7 +49,7 @@ void throw_if_binary_fails(const char* binary_name, const QStringList& arguments
 
 QStringList generate_extra_apparmor_args()
 {
-    if (mp::utils::is_snap())
+    try
     {
         QString apparmor_cache_dir = mp::utils::snap_common_dir() + "/apparmor.d/cache/multipass";
         QDir cache_dir;
@@ -61,6 +62,11 @@ QStringList generate_extra_apparmor_args()
             mpl::log(mpl::Level::debug, "daemon", "Failed to create cache directory for AppArmor - disabling caching");
         }
     }
+    catch (const mp::SnapEnvironmentException&)
+    {
+        // Ignore
+    }
+
     return {"-W"};
 }
 
--- a/src/platform/backends/shared/linux/apparmor.cpp
+++ b/src/platform/backends/shared/linux/apparmor.cpp
@@ -75,7 +75,7 @@ QStringList generate_extra_apparmor_args()
 mp::AppArmor::AppArmor() : apparmor_args{generate_extra_apparmor_args()}
 {
     int ret = aa_is_enabled();
-    if (ret < 0)
+    if (ret <= 0)
     {
         throw mp::AppArmorException("AppArmor is not enabled");
     }
--- a/src/platform/backends/shared/linux/process_factory.cpp
+++ b/src/platform/backends/shared/linux/process_factory.cpp
@@ -64,7 +64,7 @@ private:
 
 mp::optional<mp::AppArmor> create_apparmor()
 {
-    if (!mp::utils::is_snap() && qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
+    if (!qEnvironmentVariableIsSet("SNAP") && qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
     {
         mpl::log(mpl::Level::warning, "apparmor", "AppArmor disabled by environment variable");
         return mp::nullopt;
--- a/src/platform/backends/shared/linux/process_factory.cpp
+++ b/src/platform/backends/shared/linux/process_factory.cpp
@@ -16,13 +16,14 @@
  */
 
 #include "process_factory.h"
+
+#include <multipass/exceptions/snap_environment_exception.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/process/basic_process.h>
 #include <multipass/process/process_spec.h>
 #include <multipass/process/simple_process_spec.h>
 #include <multipass/snap_utils.h>
-#include <multipass/utils.h>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
@@ -64,23 +65,28 @@ private:
 
 mp::optional<mp::AppArmor> create_apparmor()
 {
-    if (!qEnvironmentVariableIsSet("SNAP") && qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
+    try
     {
-        mpl::log(mpl::Level::warning, "apparmor", "AppArmor disabled by environment variable");
-        return mp::nullopt;
+        mp::utils::snap_dir();
     }
-    else
+    catch (const mp::SnapEnvironmentException&)
     {
-        try
+        if (qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
         {
-            mpl::log(mpl::Level::info, "apparmor", "Using AppArmor support");
-            return mp::AppArmor{};
-        }
-        catch (mp::AppArmorException& e)
-        {
-            mpl::log(mpl::Level::warning, "apparmor", fmt::format("Failed to enable AppArmor: {}", e.what()));
+            mpl::log(mpl::Level::warning, "apparmor", "AppArmor disabled by environment variable");
             return mp::nullopt;
         }
+    }
+
+    try
+    {
+        mpl::log(mpl::Level::info, "apparmor", "Using AppArmor support");
+        return mp::AppArmor{};
+    }
+    catch (mp::AppArmorException& e)
+    {
+        mpl::log(mpl::Level::warning, "apparmor", fmt::format("Failed to enable AppArmor: {}", e.what()));
+        return mp::nullopt;
     }
 }
 } // namespace
--- a/src/platform/platform_linux.cpp
+++ b/src/platform/platform_linux.cpp
@@ -18,6 +18,7 @@
 #include <multipass/constants.h>
 #include <multipass/exceptions/autostart_setup_exception.h>
 #include <multipass/exceptions/settings_exceptions.h>
+#include <multipass/exceptions/snap_environment_exception.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
@@ -81,12 +82,12 @@ std::string mp::platform::default_server_address()
 {
     std::string base_dir;
 
-    if (mu::is_snap())
+    try
     {
         // if Snap, client and daemon can both access $SNAP_COMMON so can put socket there
         base_dir = mu::snap_common_dir().toStdString();
     }
-    else
+    catch (const mp::SnapEnvironmentException&)
     {
         base_dir = "/run";
     }
--- a/src/process/qemuimg_process_spec.cpp
+++ b/src/process/qemuimg_process_spec.cpp
@@ -15,6 +15,7 @@
  *
  */
 
+#include <multipass/exceptions/snap_environment_exception.h>
 #include <multipass/process/qemuimg_process_spec.h>
 #include <multipass/snap_utils.h>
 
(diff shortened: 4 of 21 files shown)


tests/linux/test_snap_utils.cpp
```diff
@@ -34,14 +35,14 @@ TEST(SnapUtils, test_is_confined_when_snap_dir_set)
 {
     mpt::SetEnvScope env("SNAP", "/tmp");
 
-    EXPECT_TRUE(mu::is_snap());
+    EXPECT_NO_THROW(mu::snap_dir());
```
ricab, line 38: Missing the corres

tests/linux/test_snap_utils.cpp
```diff
@@ -34,14 +35,14 @@ TEST(SnapUtils, test_is_confined_when_snap_dir_set)
 {
     mpt::SetEnvScope env("SNAP", "/tmp");
```
ricab, line 36:
```suggestion
    mpt::SetEnvScope env("SNAP", "/tmp");
    mpt::SetEnvScope env("SNAP_COMMON", "/tmp");
```

src/platform/backends/shared/linux/process_factory.cpp
```diff
@@ -64,7 +64,7 @@ class AppArmoredProcess : public mp::BasicProcess
 
 mp::optional<mp::AppArmor> create_apparmor()
 {
-    if (!mp::utils::is_snap() && qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
+    if (!qEnvironmentVariableIsSet("SNAP") && qEnvironmentVariableIsSet("DISABLE_APPARMOR"))
```
ricab, line 67: Why a different approach here? I mean, it could also be something like:

```c++
try
{
    mu::snap_dir();
    if (qEnvironmentVariableIsSet("DISABLE_APPARMOR")
    {
        ...
    }
}
catch(...)
...
```

tests/linux/test_snap_utils.cpp
townsend2010: Although I know it's harmless, I'm not sure why we need to set `SNAP_COMMON` for this test.  Could you please explain why you suggested this?

src/utils/snap_utils.cpp
```diff
@@ -15,27 +15,32 @@
...
+
     return QFileInfo(snap_dir).canonicalFilePath().toUtf8(); // To resolve any symlinks
 }
 
 QByteArray mu::snap_common_dir()
 {
     auto snap_common = qgetenv("SNAP_COMMON");                  // Inside snap, this can be trusted
+    if (snap_common.isEmpty())
+    {
+        throw mp::SnapEnvironmentException("SNAP_COMMON");
```
ricab, line 42: I suppose both `SNAP` and `SNAP_COMMON` should be guaranteed when running in a snap (right?), otherwise, it would be quite easy to disable apparmor profiles... But the opposite isn't true: we can have any of those defined when not in a snap.

By treating `SNAP` and `SNAP_COMMON` separately, we're relying on them being either both defined or both not defined. If we run in an environment where one of them is defined but not the other, we'll be doing some things as in a snap and some others as not in a snap. The situation was slightly different before this PR, because we would always gate on `SNAP` being defined (via `is_snap`), even when reading `SNAP_COMMON`.

src/process/qemuimg_process_spec.cpp
```diff
@@ -65,13 +66,13 @@ profile %1 flags=(attach_disconnected) {
     QString extra_capabilities;
     QString signal_peer; // who can send kill signal to qemu-img
 
-    if (mu::is_snap())
+    try
     {
         root_dir = mu::snap_dir();
         image_dir = mu::snap_common_dir();         // FIXME - am guessing we work inside this directory
```
ricab, line 72: This is a particular case of my earlier comment. If execution reaches this point and throws, we use a mix of snap and non-snap dirs in the profile.

src/platform/platform_linux.cpp
```diff
@@ -81,12 +82,12 @@ std::string mp::platform::default_server_address()
 {
     std::string base_dir;
 
-    if (mu::is_snap())
+    try
     {
         // if Snap, client and daemon can both access $SNAP_COMMON so can put socket there
         base_dir = mu::snap_common_dir().toStdString();
```
ricab, line 88: Another particular case of the behavior change I mentioned.

tests/linux/test_snap_utils.cpp
ricab: Sorry, this was actually an oversight on my part. I noticed there was no analogous test checking `mu::snap_common_dir` didn't throw and it seemed that adding that var here would do it, but it won't because that function is not being called. So my suggestion would instead be to have that analogous test.

townsend2010: Right, I can either add the second call _with_ your suggestion applied or just add a second test.

src/utils/snap_utils.cpp
townsend2010: > But the opposite isn't true: we can have any of those defined when not in a snap.

Why do we care about this scenario?  This implies that someone is running from source (like us on the project) or has packaged Multipass in some other format **and** set `SNAP` and/or `SNAP_COMMON` when `multipassd` is running.  IMO, that is not a supported scenario and the user is on their own and we shouldn't add complexity to guard against that.

src/platform/backends/shared/linux/process_factory.cpp
townsend2010: Ok, I agree with you that is a more consistent way of handling this.  Thanks!

tests/linux/test_snap_utils.cpp
ricab: Yes, exactly.

src/utils/snap_utils.cpp
ricab: I am not sure we need to do anything, but I wanted to point it out and get feedback. 

I was thinking that users could have vars with the same name in their environment, even on other platforms. That is unlikely in the daemon, but not impossible. And utils can be used in clients too, and in cross-platform code, even if that is not the case today. Of course such variables would already cause problems without this PR, but now there is a slight difference which I wanted to make sure of.

I suppose we can say that we don't support those variables outside of snaps and that we require both to be either defined or undefined simultaneously. Ideally that would involve some mention in documentation.

Saviq:
Will reopen to trigger GH

tests/linux/test_snap_utils.cpp
townsend2010: Done.  I made another test for this.

src/platform/backends/shared/linux/process_factory.cpp
townsend2010: I redid this logic taking the exception into account now.

src/utils/snap_utils.cpp
townsend2010: Your points here are duly noted :grin:  I think we should keep it in mind, but leave it as is for now.

src/process/qemuimg_process_spec.cpp
townsend2010: Noted

src/platform/platform_linux.cpp
townsend2010: Yep

src/utils/snap_utils.cpp
townsend2010: Hey @ricab,

After thinking about this some more, I took your feedback and added a new function in the `snap_utils` to verify that `SNAP_NAME` == `multipass` before trying to get either snap directory environment variables.  This should at least add another level of checks on making sure we are running as a snap or not.  Let me know what you think.

ricab: Thanks, yes I think that makes sense, as it adds another way in which we catch bad snap-like environments early, so certainly an improvement.

One way to enforce all three vars would be a class that initialized everything in the ctor and threw when any of them was not defined. Something like:

```cpp

class SnapInfo
{
public:
  SnapInfo() 
    : snap_dir{checked_getenv(...)}
    , snap_common(checked_getenv(...)}
  {
    checked_getenv("SNAP_NAME");
  }
  
  const string& snap_dir() const;
  const string& snap_common() const;

private:
  string checked_getenv(const string& var_name)
  {
    if (string val = getenv(var_name); val)
      return val;
    throw ...;
  }

  string snap_dir;
  string snap_common;
}

// client would do
try
{
  auto snap_dir = mu::SnapInfo{}.snap_dir().
  // or
  auto snap_dir = mu::SnapInfo::instance().snap_dir().
}
catch...
```

We'd be reading 3 vars each time though, instead of the current 2. The extra reads could be saved by making it a singleton. It wouldn't require mocking since we can mock everything we need at the environment level (and the singleton can be reset). 

But existing tests would need updating, so it may be more work than it's worth. I leave it up to you.

townsend2010: Hey @ricab,

Thanks for the feedback!  I may just go the singleton route here since I think the calls to `getenv()` every time is redundant and can be streamlined in a singleton class as you say.

townsend2010: Ok, I've been working on a singleton approach and it is not a trivial change, particularly around the tests.  *Any* test that uses the snap stuff will have to account for resetting the singleton in its test fixture, which many of the tests don't have one.

Although I think the singleton approach is the cleanest, it adds quite a bit of complexity for little gain.  I think leaving it how it is in this PR is the best use of time.

ricab: Yes, I suspected that. OK, fine by me.

ricab: Hi @townsend2010, I started looking at the code again and realized that what you have now still requires repeated guards in each test.

```c++
    QTemporaryDir snap_dir;
    mpt::SetEnvScope env("SNAP_COMMON", snap_dir.path().toUtf8());
    mpt::SetEnvScope env2("SNAP_NAME", snap_name);
```

This code is repeated each time and IMO it should be replaced with a fixture setup anyway...

townsend2010: We do different things with the env vars in each test, so I don't think a test fixture would help here.

ricab: IIUC, outside snap-utils we need only 2 situations: snap and non-snap. IOW: 

```c++
EXPECT_CALL(mock_snap_detector, get_[common_]dir).WillOnce(Return(...)|Throw(...));
```

Only snap detection tests themselves need further granularity. A singleton would rid us of a lot of boiler plate in tests, separate them into better *units*, read the environment O(1) times, instead of O(n), and enforce consistency among all 3 vars.

That said, I agree it would be a significant additional step that we can leave for a future PR.

src/platform/backends/shared/linux/apparmor.cpp
```diff
@@ -48,7 +49,7 @@ void throw_if_binary_fails(const char* binary_name, const QStringList& arguments
 
 QStringList generate_extra_apparmor_args()
 {
-    if (mp::utils::is_snap())
+    try
```
ricab, line 52: Two questions about making apparmor dependent on the environment vars. I expect we are safe on both, but just to be absolutely sure:
  - are the 3 vars guaranteed to be defined when in a strict snap? is there any known way to unset/overwrite them? (it would be a preexisting issue, but a pressing one)
  - would `SNAP_NAME` or `SNAP_COMMON` be any easier to unset/overwrite than `SNAP` - if so, that would be an issue with this PR.

tests/linux/mock_aa_syscalls.h
```diff
@@ -0,0 +1,27 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_AA_IS_ENABLED_H
+#define MULTIPASS_AA_IS_ENABLED_H
```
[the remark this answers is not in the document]
townsend2010: Oh, right, good catch.  Thanks!

src/platform/backends/shared/linux/apparmor.cpp
townsend2010: > are the 3 vars guaranteed to be defined when in a strict snap?

Yes, and not just strict, but classic as well.

> is there any known way to unset/overwrite them?

Since the snap vars are set by snapd and *only* seen by the running snap, the only way would be for the snap itself to change them.  If something _outside_ of the snap can unset/overwrite them, then that is a serious failing of the whole snap architecture.

> would SNAP_NAME or SNAP_COMMON be any easier to unset/overwrite than SNAP - if so, that would be an issue with this PR.

See my answer above.  Short answer, it would be a serious failing of the whole snap architecture.

tests/linux/mock_aa_syscalls.h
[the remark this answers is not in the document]
townsend2010: Ok, fixed.

src/platform/backends/shared/linux/apparmor.cpp
ricab: OK, good to know.

PR #1646, opened 2020-07-14
Log the command-line of the command being executed

luis4a0:
This logging is implemented in BasicProcess, thus it will work for all processes launched as a ProcessSpec.

The feature was discussed in https://github.com/canonical/multipass-private/pull/273

### DIFF
--- a/src/process/basic_process.cpp
+++ b/src/process/basic_process.cpp
@@ -181,7 +181,7 @@ mp::ProcessState mp::BasicProcess::execute(const int timeout)
     start();
 
     mpl::log(mpl::Level::debug, "basic_process",
-             fmt::format("executing {} {}", qUtf8Printable(process_spec->program()),
+             fmt::format("executing: {} {}", qUtf8Printable(process_spec->program()),
                          qUtf8Printable(process_spec->arguments().join(' '))));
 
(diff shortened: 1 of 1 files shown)


src/process/basic_process.cpp
```diff
@@ -178,6 +180,10 @@ mp::ProcessState mp::BasicProcess::execute(const int timeout)
     mp::ProcessState exit_state;
     start();
 
+    mpl::log(mpl::Level::debug, "basic_process",
+             fmt::format("executing {} {}", qUtf8Printable(process_spec->program()),
```
ricab, line 184:
```suggestion
             fmt::format("executing '{} {}'", qUtf8Printable(process_spec->program()),
```

ricab:
hmm 

bors d+

src/process/basic_process.cpp
luis4a0: I thought about that, but didn't use quotes in the log because in case the command has some sort of quotes it may be confusing.

luis4a0: Anyway, having quotes separates the command from the rest, and the line can be parsed even with quotes. I accept the suggestion.

luis4a0:
bors r=ricab

PR #1647, opened 2020-07-15
Move process logging to start

ricab:
So that direct calls to `BasicProcess::start` are also logged.

### DIFF
--- a/src/process/basic_process.cpp
+++ b/src/process/basic_process.cpp
@@ -106,6 +106,9 @@ QProcessEnvironment mp::BasicProcess::process_environment() const
 
 void mp::BasicProcess::start()
 {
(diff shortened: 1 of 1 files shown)


ricab:
Hey @luis4a0, I noticed that there are [some](https://github.com/canonical/multipass/blob/08ffbedc3a030d7213dd82cd6de77e0d56668e03/src/sshfs_mount/sshfs_mounts.cpp#L51) [cases](https://github.com/canonical/multipass/blob/6ec67ca0a8ca56e5761c7515d5853ef3dfb6933a/src/platform/backends/qemu/qemu_virtual_machine.cpp#L259) that call start directly. Since execute also calls start, moving logging there should cover all cases.

I also placed it *before* starting. That way, if anything wrong happens during start, we know what it's about.

PR #1657, opened 2020-07-23
[grpc] Changes needed to support updating gRPC to version 1.30.2

townsend2010:
- boringssl-with-bazel
  - boringssl itself has been dropped.
  - Point openssl includes to boringssl-with-bazel directory.
  - Drop libssh link to decrepit library is this is no longer built.
  - Explicitly link the cert library to the ssl library.
- protobuf
  - Drop '-Wno-extended-offsetof' check as this is no longer used by protobuf.
  - Add '-Wno-error' for rpc builds since new protobuf uses the zero byte array extension
    and ISO C++ forbids causing -pendantic to issue a warning that gets treated like an error.

### DIFF
--- a/src/rpc/CMakeLists.txt
+++ b/src/rpc/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright © 2017 Canonical Ltd.
+# Copyright © 2017-2020 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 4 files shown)


src/rpc/CMakeLists.txt
```diff
@@ -11,13 +11,13 @@
...
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-# Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
 
 set(MULTIPASS_PROTOCOL_SPEC multipass.proto)
 set(GRPC_GENERATED_SOURCE_DIR ${MULTIPASS_GENERATED_SOURCE_DIR}/multipass/rpc)
 file(MAKE_DIRECTORY ${GRPC_GENERATED_SOURCE_DIR})
 
+add_compile_options(-Wno-error)
```
ricab, line 19:
```suggestion
add_compile_options(-Wno-error=pedantic)
```

townsend2010: Ah, cool, didn't realize that it could be that specific.

townsend2010:
Hey @ricab,

Thanks for the review and suggestion.

> a question about a change in the grpc repo itself: why this line?

Honestly, I don't know why that is there.  I'll remove that from our grpc branch and update this branch to point to the new commit.

ricab:
Thank you @townsend2010 ! 

I thought you had added the `memset` this time. Digging a bit more, it was there in upstream in the past and it was dropped in the same [commit that added the ctor](https://github.com/albaguirre/grpc/commit/f5a71f73139cb74bc8703bfb89a5203a12b182ca). That line must have slipped us in some merge ¿? 

Could I ask that we add the new change in a new commit instead, please? So that the history is clear(er) next time. Thanks in advance!

PR #1663, opened 2020-07-28
Control network interfaces

luis4a0:
This branch adds control to the interfaces in the VM. On one hand, it adds the command `list-networks`, which shows the hardware, bridge and switch interfaces on the system. This functionality uses data from the backend, as well as data gathered from the platform. On the other hand, it adds the parameter `--networks` to the `launch` command, which adds a network interface to the created VM and is backend-specific.


luis4a0:
Replaced by https://github.com/canonical/multipass/pull/1745

PR #1667, opened 2020-07-29
Update hyperkit (note: no longer no-op)

ricab:
Build mac package with updated hyperkit.


Saviq (approved):
#1751 confirms that this works in our new and improved CI setup.

PR #1668, opened 2020-07-31
Update deps

ricab:
This updates our 3rd-party where there were tagged updates available, except libssh (to avoid collision with the ssl move), and premock (as discussed).

### DIFF
--- a/3rd-party/submodule_info.md
+++ b/3rd-party/submodule_info.md
@@ -8,43 +8,43 @@ info on where to find release information about the submodule.
 
 ### gRPC
 Version: 1.25.0 (+[our patches](https://github.com/CanonicalLtd/grpc/compare/v1.25.0..3efaade))
-[https://github.com/CanonicalLtd/grpc.git](https://github.com/CanonicalLtd/grpc.git)
-[https://github.com/grpc/grpc/releases](https://github.com/grpc/grpc/releases)
+<https://github.com/CanonicalLtd/grpc.git>
+<https://github.com/grpc/grpc/releases>
 
 ### libssh
 Version: 0.9.3 (+[our patches](https://github.com/CanonicalLtd/libssh/compare/libssh-0.9.3..2608224))
-[https://github.com/CanonicalLtd/libssh.git](https://github.com/CanonicalLtd/libssh.git)
-[https://www.libssh.org/](https://www.libssh.org/)
+<https://github.com/CanonicalLtd/libssh.git>
+<https://www.libssh.org/>
 
 ### fmt
 Version: 7.0.2
-[https://github.com/fmtlib/fmt.git](https://github.com/fmtlib/fmt.git)
-[https://github.com/fmtlib/fmt/releases](https://github.com/fmtlib/fmt/releases)
+<https://github.com/fmtlib/fmt.git>
+<https://github.com/fmtlib/fmt/releases>
 
 ### QHotKey
 Version: 1.4.1
-[https://github.com/Skycoder42/QHotkey.git](https://github.com/Skycoder42/QHotkey.git)
-[https://github.com/Skycoder42/QHotkey/releases](https://github.com/Skycoder42/QHotkey/releases)
+<https://github.com/Skycoder42/QHotkey.git>
+<https://github.com/Skycoder42/QHotkey/releases>
 
 ### scope_guard
 Version: 0.2.3 (+[upstream patches](https://github.com/ricab/scope_guard/compare/v0.2.3..760de0a))
-[https://github.com/ricab/scope_guard.git](https://github.com/ricab/scope_guard.git)
-[https://github.com/ricab/scope_guard/releases](https://github.com/ricab/scope_guard/releases)
+<https://github.com/ricab/scope_guard.git>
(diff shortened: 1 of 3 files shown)


3rd-party/submodule_info.md
```diff
@@ -7,41 +7,51 @@ info on where to find release information about the submodule.
 **This file needs updating when a submodule is updated!**
 
 ### gRPC
-Version: 1.25.0 (+[our patches](https://github.com/CanonicalLtd/grpc/compare/v1.25.0..3efaade))
-[https://github.com/CanonicalLtd/grpc.git](https://github.com/CanonicalLtd/grpc.git)
-[https://github.com/grpc/grpc/releases](https://github.com/grpc/grpc/releases)
+Version: 1.25.0 (+[our patches](https://github.com/CanonicalLtd/grpc/compare/v1.25.0..3efaade)) |
```
townsend2010, line 10: This file wasn't in place when we updated gRPC recently.  Could you please update this to reflect that we are using version 1.30.2 plus our patches?  Thanks!

ricab: Of course, will do!

ricab: Done. I had to push the tags to our fork.

3rd-party/submodule_info.md
```diff
@@ -7,41 +7,51 @@ info on where to find release information about the submodule.
...
 
 ### yaml-cpp
-Version: 0.6.3
-[https://github.com/jbeder/yaml-cpp.git](https://github.com/jbeder/yaml-cpp.git)
-[https://github.com/jbeder/yaml-cpp/releases](https://github.com/jbeder/yaml-cpp/releases)
+Version: 0.6.3 |
+<https://github.com/jbeder/yaml-cpp.git> |
+<https://github.com/jbeder/yaml-cpp/releases>
+
+### JsonCpp
```
townsend2010, line 49: I don't think we should reference this here and should only be in the private branch since it's *only* used there.

3rd-party/submodule_info.md
```diff
@@ -7,41 +7,51 @@ info on where to find release information about the submodule.
...
+Version: 1.9.3 |
+<https://github.com/open-source-parsers/jsoncpp.git> |
+<https://github.com/open-source-parsers/jsoncpp/releases>
+
+### HyperKit
```
townsend2010, line 54: Same as above for HyperKit.

3rd-party/submodule_info.md
ricab: I thought the goal of this file was to inform of our dependencies, and we have discussed that the private branch is for things that are considered IP (and that they should be contained in the platform dirs as much as possible).

For the rest, my understanding is that it is better to have them on both sides, to reduce diffs and allow more changes on master to be reflected on full without intervention. An example: when we later go and update libssh, if this file differs, we'd need to fix a conflict to merge on full. If the file was the same, the change would apply transparently.

Admittedly, that comes at the expense of sometimes requiring, on the public side, commits that would otherwise go in a single private branch. But we have to do that a lot already, because most code is shared anyway, so I am not sure this approach would introduce many such cases. And the changes are still only done once, on one side, and applied cleanly on the other.

I guess the more common ground we have, the more we can do only once. OTH, if the two branches accumulate differences, at the limit all changes requires specific handling on both sides.

Anyway, whatever we choose, I think we should clear up the direction we want to go as a team, so that we all know which way to row.

townsend2010: IMO, `#ifdef`s in the code and `cmake` platform handling is acceptable, but this is referring to submodules that the `master` branch simply does not pull in, compile, and use.  It is not accurate and I can see it being confusing for someone who may be trying to contribute to Multipass.

If the goal is to bring as much as we can into `master` (of which I'm definitely not opposed), then we should also add the submodules in `master`, but just not build them (I know, cloning time that is for nil, but it makes the `master`/`full` diff less).   At least there is a real reference to them that would match this file.

@Saviq & @luis4a0, any opinions?

luis4a0: To make a decision on this, I think we must take an orthogonal view. We must consider how do we want to organize the repos in the future. If we decide to implement backends as plug-ins, then we have to keep in the full branch the submodules which impact only one backend (such as Hyperkit), but we can optionally leave in master the submodules used by many backends.

Saviq: I do think we should have a single place to store this. Splitting it between this and `full` just adds confusion and complexity.

That said, maybe it simply shouldn't be here, but rather in the wiki? While having it with code has some appeal (and so that it updates along with the actual update), it's only informative (and at that, mostly to ourselves), and changed manually, so there's no advantage of having a single source of truth…

townsend2010: Ok, I also think referencing submodules that are not referenced in the branch also adds confusion, but certainly alleviates any complexity.  That said, since at least the complexity is resolved by leaving it here, then we should just leave as it is.

Regarding adding to a wiki, I have doubts that we would update the wiki when we update the submodules.  I think having it here is probably best.

PR #1670, opened 2020-08-04
[detail] Add missing param name for clarity

ricab:


ricab:
I knew you'd complain if it was a drive-by change of another PR :stuck_out_tongue:

PR #1679, opened 2020-08-12
[lxd] Supporting downloading and importing images into LXD image store

townsend2010:


townsend2010:
I'm marking this ready for review.  The tests basically cover the "go path" part of this and doesn't take failures into account.  WE can get better coverage later as it's time to move on to other things :smile:

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -364,3 +461,88 @@ void mp::LXDVMImageVault::poll_download_operation(const QJsonObject& json_reply,
...
+        QUrl task_url(QString("%1/operations/%2/wait")
+                          .arg(base_url.toString())
+                          .arg(json_reply["metadata"].toObject()["id"].toString()));
+
+        auto task_reply = lxd_request(manager, "GET", task_url, mp::nullopt, 300000);
+
+        return task_reply["metadata"].toObject()["metadata"].toObject()["fingerprint"].toString().toStdString();
+    }
+
+    return {};
```
Saviq, line 507: We'd end up with empty `source_image.id`? Shouldn't this be an error?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -136,29 +212,31 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
-            source_object.insert("fingerprint", id);
+            source_image.id = lxd_image_hash;
         }
-        else
+        else if (!info.stream_location.isEmpty())
         {
-            source_object.insert("alias", QString::fromStdString(query.release));
+            lxd_download_image(id, info.stream_location, QString::fromStdString(query.release), monitor);
         }
+        else
```
Saviq, line 224: Feels weird as a fall-through, should we be asserting something here?

src/platform/platform_linux.cpp
```diff
@@ -159,11 +159,6 @@ bool mp::platform::is_alias_supported(const std::string& alias, const std::strin
 
 bool mp::platform::is_remote_supported(const std::string& remote)
 {
-    const auto& driver = utils::get_driver_str();
-    // TODO: Remove this constraint once we support other image remotes for the LXD backend
-    if (driver == QStringLiteral("lxd") && (remote != "release" && remote != "daily"))
-        return false;
-
```
Saviq, line 166: We still can't support `appliance:`, right?

Saviq (changes requested):
NB: I removed the top "debug test" commit.

This didn't work for me:
```
$ multipass launch snapcraft:core20 
launch failed: Cannot read image format: qemu-img failed (Process returned exit code: 1) with output:
qemu-img: Could not open '/tmp/multipassd-Xcs3UE/ubuntu-20.04-minimal-cloudimg-amd64.img': Could not open '/tmp/multipassd-Xcs3UE/ubuntu-20.04-minimal-cloudimg-amd64.img': Permission denied
```

src/daemon/default_vm_image_vault.cpp
```diff
@@ -601,16 +549,16 @@ mp::VMImage mp::DefaultVMImageVault::download_and_prepare_source_image(
...
 
     if (existing_source_image)
     {
         source_image = *existing_source_image;
     }
     else
     {
-        source_image.id = id;
-        source_image.image_path = image_dir.filePath(filename_for(info.image_location));
+        source_image.id = id.toStdString();
```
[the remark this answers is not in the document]
townsend2010: Yes, I want to keep `id` as a `QString` since it is used like that in the new code.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -56,11 +72,71 @@ auto parse_percent_as_int(const QString& progress_string)
...
+{
+    QString new_image_path{image_path};
+
+    if (image_path.endsWith(".xz"))
+    {
+        new_image_path = mp::vault::extract_image(image_path, monitor, true);
+    }
+
+    QString original_image_path{new_image_path};
+    new_image_path = mp::backend::convert_to_qcow_if_necessary(new_image_path);
```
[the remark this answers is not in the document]
townsend2010: You know, I never tried just passing the image as-is.  I can certainly try it with `lxc` and see what happens.  If it works, great!

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -56,11 +72,71 @@ auto parse_percent_as_int(const QString& progress_string)
...
+    emitter << metadata_node << YAML::Newline;
+
+    metadata_yaml_file.open(QIODevice::WriteOnly);
+    metadata_yaml_file.write(emitter.c_str());
+    metadata_yaml_file.close();
+
+    const auto metadata_tarball_path = lxd_import_dir.filePath("metadata.tar");
+    auto process = MP_PROCFACTORY.create_process(
+        "tar", QStringList() << "-cf" << metadata_tarball_path << "-C" << lxd_import_dir.path()
+                             << QFileInfo(metadata_yaml_file.fileName()).fileName());
```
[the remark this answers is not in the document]
townsend2010: There is and I looked through it and figured it was just "faster" to do this than introduce a new library and add all of the code to do what I can do here in basically one line.  `tar` is included in Core, so I thought just calling the executable was the quicker choice to bring it to market :smile:  That said, I prefer libraries as well, but thought it was overkill in this instance.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Yeah, it should probably throw here instead.  Will fix.

src/platform/platform_linux.cpp
townsend2010: Right, but the `appliance` remote is streams based, so it's smart enough to figure out there are no LXD images available, so they won't show up in `multipass find` or be available for `launch`.  So once those images are there, it should just start working without any code intervention.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Hmm, I'm not sure what we should be asserting.  This is the last way to try to get an image and we know that it's a legitimate image since we got it from a manifest.  Otherwise, we would have already failed before getting here.

src/platform/platform_linux.cpp
Saviq: D'uh, I implemented that myself ;P

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: Sure, if we assume that we're using the vault correctly, nothing below will fail, but maybe due dilligence to assert `info.image_location` isn't empty? Your call.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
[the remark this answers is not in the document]
townsend2010: After some experimenting, `lxc` will import the `xz` image, but treats it as if it were a compressed `rootfs` and not a VM image.  I see no way to force it otherwise.  Also, if we leave it raw, it is a **huge** upload that takes more time than having `qemu-img` convert it.  I think we are stuck with this.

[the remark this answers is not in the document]
townsend2010: I believe we can safely resolve this since Mr. LXD said it only supports qcow2 :smiley:

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Fixed

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Ok, I changed this to account for `info.image_location` and `throw` if none of the conditions can be met.

townsend2010:
The download failure using the snap should be fixed now.

Saviq (changes requested):
This breaks the `qemu` backend now:
```plain
wrz 21 20:08:40 michal-laptop multipassd[2878898]: QDir::mkpath: Empty or null file name                    
…
wrz 21 20:08:40 michal-laptop multipassd[2878898]: Caught an unhandled exception: unable to create directory '/var/snap/multipass/common/cache/multipassd'
```

townsend2010:
> This breaks the qemu backend now:

Fixed now, thanks!

Saviq:
> Fixed now, thanks!

-ish:

```plain
 [ RUN      ] Utils.unable_to_make_dir_throws
/root/parts/multipass/src/tests/test_utils.cpp:470: Failure
Expected: mp::utils::make_dir(QDir(temp_dir.path()), new_dir) throws an exception of type std::runtime_error.
  Actual: it throws nothing.
[  FAILED  ] Utils.unable_to_make_dir_throws (0 ms)
```

townsend2010:
> -ish:

Grrr, freakin' tests that behave differently on different machines...

townsend2010:
Ok, I had to drop the test for catching the exception since the tests run as `root` and there really isn't a trivial way to keep `root` from being able to create directories no matter the file permissions or where the directory resides.

Saviq (approved):
W00t!
```plain
 multipass launch core18 & mp launch core
…
Launched: oriented-basset                                                      
Launched: relieved-panther
 multipass ls
oriented-basset         Running           10.152.94.34     Not Available
relieved-panther        Running           10.152.94.126    Not Available
 lxc ls --project multipass
+------------------+---------+----------------------+----------------------------------------------+-----------------+-----------+
|       NAME       |  STATE  |         IPV4         |                     IPV6                     |      TYPE       | SNAPSHOTS |
+------------------+---------+----------------------+----------------------------------------------+-----------------+-----------+
| oriented-basset  | RUNNING | 10.152.94.34 (eth0)  | fd42:254b:731a:5118:5054:ff:fe57:f4cc (eth0) | VIRTUAL-MACHINE | 0         |
+------------------+---------+----------------------+----------------------------------------------+-----------------+-----------+
| relieved-panther | RUNNING | 10.152.94.126 (eth0) | fd42:254b:731a:5118:5054:ff:fe0a:7f8c (eth0) | VIRTUAL-MACHINE | 0         |
+------------------+---------+----------------------+----------------------------------------------+-----------------+-----------+
```

bors merge

PR #1682, opened 2020-08-13
[lxd] Do not send unnecessary source info on instance creation

townsend2010:
The image is downloaded separately, so there is no need to send source info that tells LXD where to download an image during instance creation.


Saviq:
ACK

bors merge

PR #1683, opened 2020-08-13
[test] Improve the MockLogger and adapt tests accordingly

### LINKED ISSUE #1676: MockLogger not reset in some tests
Some tests register a mock logger globally, but fail to reset it. Since the logger's lifetime is managed with a shared_ptr, it will stick around for ensuing tests. Also, call expectations on the mock logger are only verified on the following reset (different test).

Some examples I noticed:
https://github.com/canonical/multipass/blob/834eca128704bd0a149bc8a8852db4912a83f15f/tests/lxd/test_lxd_backend.cpp#L58
https://github.com/canonical/multipass/blob/0d74fa3510e7f57c443cb9a4264fa4a7893b70aa/tests/test_base_virtual_machine_factory.cpp#L51

ricab:
This tries to make the `MockLogger` interface safer and more intuitive. Introduces some functionality that will be needed in upcoming tests, and employs it already where applicable, simplifying setup code.

Fixes #1676.


src/logging/log.cpp
```diff
@@ -71,3 +71,8 @@ void mpl::set_logger(std::shared_ptr<Logger> logger)
     global_logger = std::move(logger);
     qInstallMessageHandler(qt_message_handler);
 }
+
+auto mpl::get_logger() -> Logger* // for tests, don't rely on it lasting
```
townsend2010, line 75: I'm curious why you use the `auto blah -> Logger*` format here instead of just a straight up `Logger* blah` like in the function definition?  I know they are treated one and the same and I'm not going to block this at all, but I'm just wondering if there is some sort of optimization I don't know about :smile:

tests/mock_logger.cpp
```diff
@@ -0,0 +1,57 @@
...
+namespace mp = multipass;
+namespace mpl = multipass::logging;
+namespace mpt = multipass::test;
+using namespace testing;
+
+mpt::MockLogger::MockLogger(const PrivatePass&)
+{
+}
+
+auto mpt::MockLogger::inject() -> Scope
```
townsend2010, line 29: Same question as above...

src/logging/log.cpp
ricab: It's just that the trailing return type is automatically looked up in the function's scope (like arguments). So you can say `auto ns::f() -> foo` whereas the namespace is needed if `foo` comes first: `ns::foo ns::f()`. It feels like a little redundancy that can be economized, so I usually do it like that. When the ns is small it may be more typing because of the added auto.

tests/mock_logger.cpp
ricab: The advantage is clearer here, the alternative would be: `mpt:MockLogger::Scope mpt::MockLogger::inject()`.

ricab:
Thanks for the review and for fixing conflicts @townsend2010! Let me know if there's anything else you need from my side.

src/logging/log.cpp
townsend2010: Ah, cool, yeah, that certainly makes it easier to code for without sacrificing readability.

tests/mock_logger.cpp
townsend2010: Yep, I see.  Nice!

townsend2010:
Good to go!  Thanks!

bors merge

PR #1685, opened 2020-08-13
[lxd] Drop saving stream_location as this is only needed for LXDVMImageVault

townsend2010:
The LXDVMImageVault is the only class that needs stream_location as provide by the ImageHost. There is no need to save this or make it available for LXDVirtualMachine.


Saviq:
Looks legit!

bors merge

PR #1686, opened 2020-08-14
Move ps to proc factory

ricab:
Public pre-requisites to move PS to `mp::Process`.


luis4a0 (approved):
Looks good, tests cover the new code and works on the metal.

PR #1688, opened 2020-08-14
[lxd] Query network leases for IP address instead of using state

### LINKED ISSUE #1684: [lxd] Sometimes unable to determine IP address
I have noticed lately that `multipassd` is unable to determine the IP address of the LXD VM.  This is due to LXD advertising the interface name as something other than `eth0` (like `enp5s0`).  The current code looks specifically for an interface named `eth0`.  I have not dug into as to why the interface name is this and it appears a subsequent boot of the instance does have the `eth0` name.

However, since the name is not consistent, we should instead query the `1.0/networks/mpbr0/leases` endpoint and find the IP address based on the hostname.

townsend2010:
Previously, eth0 was keyed on for the network interface name, but this name is not stable and can be named something else from time to time. Instead, query network leases for the mpbr0 bridge and match hostname to get the IP address.

Fixes #1684


townsend2010:
Hey @Saviq,

I updated this to now use our generated MAC address and then key on this when using the LXD `leases` API. I think it's as robust as we can get.

Also, I will need to follow up in another PR to remove the default template's network setup and in that PR, I will need a way to update templates on daemon start if & when we make changes to the template.

Saviq:
> I updated this to now use our generated MAC address and then key on this when using the LXD `leases` API. I think it's as robust as we can get.

Can you say why you've opted to go for the leases API rather than state? You could match on `hwaddr` in `/state`, too?

townsend2010:
@Saviq,

> Can you say why you've opted to go for the leases API rather than state? You could match on `hwaddr` in `/state`, too?

Yes, I thought of that and I'm on the fence.  Network interfaces in `/state` are just a collection of json objects with each object representing an interface, so iterating is not quite so straight forward like if it had been an array of objects, but still doable of course.  Another factor is that I already had coded getting the IP address via `/leases` and felt that it still served it's purpose of getting the IP address.  However, `/leases` will still report the IP address of stopped instances, so when starting a stopped instance, it will immediately detect the IP address and begin trying the ssh connection, so that is something to consider.

src/platform/backends/lxd/lxd_virtual_machine.h
```diff
@@ -59,11 +59,14 @@ class LXDVirtualMachine final : public VirtualMachine
     VMStatusMonitor* monitor;
     bool update_shutdown_status{true};
     multipass::optional<IPAddress> ip;
-    const QUrl base_url;
     NetworkAccessManager* manager;
+    const QUrl base_url;
```
Saviq, line 63: And here? Was it clang that reordered these? I fail to see a pattern :)

Saviq:
bors delegate+

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -112,13 +110,16 @@ mp::MemorySize get_minimum_disk_size(const mp::MemorySize& requested_disk_size)
...
-                                         NetworkAccessManager* manager, const QUrl& base_url)
+                                         NetworkAccessManager* manager, const QUrl& base_url,
+                                         const QString& bridge_name)
     : VirtualMachine{desc.vm_name},
       name{QString::fromStdString(desc.vm_name)},
       username{desc.ssh_username},
       monitor{&monitor},
+      manager{manager},
       base_url{base_url},
-      manager{manager}
```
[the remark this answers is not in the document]
townsend2010: Well, it's mainly me wanting to keep things in the same order of the variables being passed into the ctor.  Obviously not necessary, but just looked cleaner as I was adding more variables here.

src/platform/backends/lxd/lxd_virtual_machine.h
townsend2010: See above :smiley:

townsend2010:
@Saviq,

> I'll let you decide :)

I'm gonna think about it some and weight the options.

townsend2010:
Looking at LXD code, it also uses the dhcp leases of `dnsmasq`, so we're really not doing anything different here, so I'm going to leave it as-is.

bors r=Saviq

PR #1690, opened 2020-08-14
[lxd] Add try/catch around updating and pruning images

### LINKED ISSUE #1689: [lxd] daemon exits with "unhandled exception: LXD object not found" when starting with no multipass project defined
The daemon will bail on start up when there is no `multipass` project defined yet.

townsend2010:
Previously, when the "multipass" project didn't exist on daemon start, there would be an unhandled exception and the daemon would exit.

Fixes #1689

### DIFF
--- a/src/platform/backends/lxd/lxd_vm_image_vault.cpp
+++ b/src/platform/backends/lxd/lxd_vm_image_vault.cpp
@@ -193,62 +193,74 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
 
 void mp::LXDVMImageVault::prune_expired_images()
 {
-    auto json_reply = lxd_request(manager, "GET", QUrl(QString("%1/images?recursion=1").arg(base_url.toString())));
-
-    auto images = json_reply["metadata"].toArray();
-
-    for (const auto image : images)
+    try
     {
-        auto image_info = image.toObject();
-        auto last_used = std::chrono::system_clock::time_point(std::chrono::milliseconds(
-            QDateTime::fromString(image_info["last_used_at"].toString(), Qt::ISODateWithMs).toMSecsSinceEpoch()));
+        auto json_reply = lxd_request(manager, "GET", QUrl(QString("%1/images?recursion=1").arg(base_url.toString())));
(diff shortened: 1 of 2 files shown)


src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -195,63 +195,75 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
+                lxd_request(
+                    manager, "DELETE",
+                    QUrl(QString("%1/images/%2").arg(base_url.toString()).arg(image_info["fingerprint"].toString())));
+            }
         }
     }
+    catch (const LXDNotFoundException&)
+    {
+        return;
+    }
```
Saviq, line 224: This feels a bit broad… If the `DELETE` request returns 404, should we still just ignore it?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -195,63 +195,75 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
 
-            poll_download_operation(json_reply, monitor, task_complete);
+                poll_download_operation(json_reply, monitor, task_complete);
+            }
         }
     }
+    catch (const LXDNotFoundException&)
+    {
+        return;
+    }
```
Saviq, line 266: Ditto? There's a lot more requests than just the update one that you're catching here, and not all of them should be ignored, I'd say?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Right, I was too focused on the first `lxd_request()`.  Will work on a better solution.

townsend2010: Ok, this should be better now.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Should be better now.

PR #1696, opened 2020-08-17
[network_manager] Generalize requests and support chunked uploads

townsend2010:
Don't set "Multipass" specific data in the NetworkAccessManager and LocalSocketReply.
Support chunked transfers for uploading data.

### DIFF
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -143,25 +143,20 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
     auto op = request.attribute(QNetworkRequest::CustomVerbAttribute).toByteArray();
 
     // Build the HTTP method part
-    http_data += op;
-    http_data += ' ';
-    http_data += request.url().path();
+    http_data += op + ' ' + request.url().path();
 
     if (request.url().hasQuery())
     {
-        http_data += "?";
-        http_data += request.url().query();
+        http_data += "?" + request.url().query();
     }
 
     http_data += " HTTP/1.1\r\n";
 
     // Build the HTTP Host header
-    auto host = request.url().host();
+    auto host = request.url().host().toUtf8();
     if (!host.isEmpty())
     {
-        http_data += "Host: ";
-        http_data += host;
-        http_data += "\r\n";
+        http_data += "Host: " + host + "\r\n";
     }
     else
     {
@@ -172,9 +167,7 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
     auto user_agent = request.header(QNetworkRequest::UserAgentHeader).toByteArray();
     if (!user_agent.isEmpty())
     {
-        http_data += "User-Agent: ";
-        http_data += user_agent;
-        http_data += "\r\n";
+        http_data += "User-Agent: " + user_agent + "\r\n";
     }
 
     if (!local_socket_write(http_data))
@@ -182,14 +175,13 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
 
     if (op == "POST" || op == "PUT")
     {
-        http_data = "Content-Type: ";
-        http_data += request.header(QNetworkRequest::ContentTypeHeader).toByteArray();
-        http_data += "\r\n";
+        http_data = "Content-Type: " + request.header(QNetworkRequest::ContentTypeHeader).toByteArray() + "\r\n";
 
         if (outgoingData)
         {
             auto content_length = request.header(QNetworkRequest::ContentLengthHeader).toByteArray();
             auto transfer_encoding = request.rawHeader("Transfer-Encoding");
+            bool is_chunked{false};
 
             if (!content_length.isEmpty() && !transfer_encoding.isEmpty())
             {
@@ -203,20 +195,15 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
             }
             else if (!content_length.isEmpty())
             {
-                http_data += "Content-Length: ";
-                http_data += content_length;
-                http_data += "\r\n";
+                http_data += "Content-Length: " + content_length + "\r\n";
             }
             else
             {
-                http_data += "Transfer-Encoding: ";
-                http_data += transfer_encoding;
-                http_data += "\r\n";
+                http_data += "Transfer-Encoding: " + transfer_encoding + "\r\n";
+                is_chunked = true;
             }
 
-            http_data += "\r\n";
-
-            if (!local_socket_write(http_data))
+            if (!local_socket_write(http_data + "\r\n"))
                 return;
 
             local_socket->flush();
@@ -237,26 +224,22 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
                 if (bytes_read == 0)
                     break;
 
-                if (transfer_encoding == "chunked")
-                {
-                    QByteArray http_chunk_size{QByteArray::number(bytes_read, 16)};
-                    http_chunk_size += "\r\n";
-                    if (!local_socket_write(http_chunk_size))
-                        return;
-                }
-
-                if (!local_socket_write(QByteArray::fromRawData(data_buffer.data(), bytes_read)))
+                if (is_chunked && !local_socket_write(QByteArray::number(bytes_read, 16) + "\r\n"))
                     return;
 
-                if (!local_socket_write("\r\n"))
+                if (!local_socket_write(QByteArray::fromRawData(data_buffer.data(), bytes_read) + "\r\n"))
                     return;
 
                 local_socket->waitForBytesWritten();
             }
+
+            // Trailer part for chunked data
+            if (is_chunked && !local_socket_write("0\r\n"))
+                return;
         }
     }
 
-    if (!local_socket_write("\r\n\r\n"))
+    if (!local_socket_write("\r\n"))
         return;
 
     local_socket->flush();
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -18,6 +18,9 @@
 #include "local_socket_reply.h"
 
 #include <multipass/exceptions/http_local_socket_exception.h>
+#include <multipass/format.h>
+
+#include <vector>
 
 #include <QRegularExpression>
 
@@ -26,6 +29,7 @@ namespace mp = multipass;
 namespace
 {
 constexpr int len = 65536;
+constexpr int max_bytes = 32768;
 
 // Status code mapping based on
 // https://github.com/qt/qtbase/blob/dev/src/network/access/qhttpthreaddelegate.cpp
@@ -215,48 +219,42 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
             local_socket->flush();
 
             outgoingData->open(QIODevice::ReadOnly);
-            QByteArray data_buffer;
-            const auto max_bytes{32768};
+            std::vector<char> data_buffer;
+            data_buffer.reserve(max_bytes);
 
             while (true)
             {
-                auto bytes_available = outgoingData->bytesAvailable();
-                if (bytes_available == 0)
+                auto bytes_read = outgoingData->read(data_buffer.data(), max_bytes);
+                if (bytes_read < 0)
+                {
+                    throw mp::HttpLocalSocketException(
+                        fmt::format("Cannot read data to send to socket: {}", outgoingData->errorString()));
+                }
+
+                if (bytes_read == 0)
                     break;
 
-                data_buffer = outgoingData->read(bytes_available);
-
-                auto data_ptr = data_buffer.data();
-
-                while (bytes_available > 0)
+                if (transfer_encoding == "chunked")
                 {
-                    auto bytes_to_write = bytes_available < max_bytes ? bytes_available : max_bytes;
-
-                    if (transfer_encoding == "chunked")
-                    {
-                        QByteArray http_chunk_size{QByteArray::number(bytes_to_write, 16)};
-                        http_chunk_size += "\r\n";
-                        local_socket->write(http_chunk_size);
-                    }
-
-                    auto bytes_written = local_socket->write(data_ptr, bytes_to_write);
-                    if (bytes_written < 0)
-                    {
-                        setError(QNetworkReply::InternalServerError, local_socket->errorString());
-                        emit error(QNetworkReply::InternalServerError);
-
-                        return;
-                    }
-
-                    if (bytes_written == 0)
-                        break;
-
-                    local_socket->write("\r\n");
-                    local_socket->waitForBytesWritten();
-
-                    bytes_available -= bytes_written;
-                    data_ptr += bytes_written;
+                    QByteArray http_chunk_size{QByteArray::number(bytes_read, 16)};
+                    http_chunk_size += "\r\n";
+                    local_socket->write(http_chunk_size);
                 }
+
+                auto bytes_written = local_socket->write(QByteArray::fromRawData(data_buffer.data(), bytes_read));
+                if (bytes_written < 0)
+                {
+                    setError(QNetworkReply::InternalServerError, local_socket->errorString());
+                    emit error(QNetworkReply::InternalServerError);
+
+                    return;
+                }
+
+                if (bytes_written == 0)
+                    break;
+
+                local_socket->write("\r\n");
+                local_socket->waitForBytesWritten();
             }
         }
     }
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -17,8 +17,6 @@
 
 #include "local_socket_reply.h"
 
-#include <multipass/version.h>
-
 #include <QRegularExpression>
 
 namespace mp = multipass;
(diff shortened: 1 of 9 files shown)


src/network/local_socket_reply.cpp
```diff
@@ -141,36 +139,113 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+        http_data += request.url().query();
+    }
+
     http_data += " HTTP/1.1\r\n";
 
     // Build the HTTP Host header
-    // Host can be anything, so we'll use 'multipass'
-    http_data += "Host: multipass\r\n";
+    auto host = request.url().host();
+    if (!host.isEmpty())
```
Saviq, line 154: Is a request that doesn't have a host valid?

src/network/local_socket_reply.cpp
```diff
@@ -141,36 +139,113 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+    {
+        http_data += "Host: ";
+        http_data += host;
+        http_data += "\r\n";
```
Saviq, line 158: If you can think of a more concise way of doing this, could be nice. There's a _lot_ of `… += "\r\n"` below…

src/network/local_socket_reply.cpp
```diff
@@ -141,36 +139,113 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+                if (bytes_available == 0)
+                    break;
+
+                data_buffer = outgoingData->read(bytes_available);
+
+                auto data_ptr = data_buffer.data();
+
+                while (bytes_available > 0)
+                {
+                    auto bytes_to_write = bytes_available < max_bytes ? bytes_available : max_bytes;
```
Saviq, line 217: This could be a bit more concise / standard with [`qint64 QIODevice::read(char *data, qint64 maxSize)`](https://doc.qt.io/qt-5.9/qiodevice.html#read), no? You wouldn't need two loops, either?

Also feels weird to `while (true)` and then `break` just a couple lines below.

src/network/local_socket_reply.cpp
```diff
@@ -141,36 +139,113 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+                        emit error(QNetworkReply::InternalServerError);
+
+                        return;
+                    }
+
+                    if (bytes_written == 0)
+                        break;
+
+                    local_socket->write("\r\n");
+                    local_socket->waitForBytesWritten();
```
Saviq, line 239: No error handling?

src/network/local_socket_reply.cpp
townsend2010: No, it's not :wink:   Will either throw an exception or emit an error in `LocalSocketReply`.

src/network/local_socket_reply.cpp
townsend2010: Well, this is how Qt does it in their network related code, so I'm following their style.  That said, I can use `fmt::format()` for building the string and then use `QByteArray::fromStdString()` to convert it.  Or a `QString().arg()` and `QString.toUtf8()` combo.

src/network/local_socket_reply.cpp
townsend2010: Again, I'm just following how Qt does this in their internal code with some tweaks for our purposes.  Also, in `QHttpMultiPart`, it's possible for a part to be backed by a `QFile`, so the ensuing `QIODevice` here may have much more data that can be read at any one time than can be written to the local socket.  I suppose as you say we can just read and then write `max_bytes` at a time, but that will hurt performance.

src/network/local_socket_reply.cpp
townsend2010: I *suppose* an error could occur writing those 2 bytes, so I should account for that. :smile:

townsend2010:
@Saviq,

Thanks for the review.

> Could improve coverage, too:

Right, I forgot to cherry-pick the chunked test from the other branch.  Will add that.  Of course, I can't test the error handling since I can't mock the `write()` calls there, so the coverage will still be lacking in that part.

src/network/local_socket_reply.cpp
Saviq: Your call if you think it worth it at all.

src/network/local_socket_reply.cpp
Saviq: The file thing is an argument indeed, but then what about memory? Will a file-backed `QIODevice` have the whole file in `::bytes_available()`? We don't want to read it all into memory, do we? So we should still have a max read?

Don't we have a overflow here, btw? If `bytesAvailable()` can be more than the size of `data_buffer`?

src/network/local_socket_reply.cpp
Saviq: Wouldn't it actually cover the writes above, too?

src/network/local_socket_reply.cpp
townsend2010: >  Will a file-backed `QIODevice` have the whole file in `::bytes_available()`?

From my experience, yes.

> We don't want to read it all into memory, do we?

It will be temporary and most of the images are in the 100's of MB's, so I don't think it would impact users in such a way that they would notice, but if cloud images grow to GB's, then that could be a problem.

> So we should still have a max read?

Need to weigh the performance vs. memory usage as mentioned above.  I don't see it as a problem now, but can't guarantee the future.  Perhaps trying to find a `max_bytes` size like 100 MB or something will keep performance and still avoid memory pressure in the future.

> Don't we have overflow here, btw? If `bytesAvailable()` can be more than the size of `data_buffer`?

We are assigning the `QByteArray` `data_buffer` to the `QByteArray` returned by `read()`, so the `read` is doing all of the `QByteArray` allocation.  If we do indeed do this, I should probably ensure the size of `data_buffer` does match `bytes_available` so it is the expected size.

src/network/local_socket_reply.cpp
townsend2010: Hmm, right....

src/network/local_socket_reply.cpp
Saviq: > It will be temporary and most of the images are in the 100's of MB's, so I don't think it would impact users in such a way that they would notice, but if cloud images grow to GB's, then that could be a problem.

That can easily be the case for images, so I do believe this to be a problem.



> Need to weigh the performance vs. memory usage as mentioned above. I don't see it as a problem now, but can't guarantee the future. Perhaps trying to find a `max_bytes` size like 100 MB or something will keep performance and still avoid memory pressure in the future.

I really don't think the additional loops will hurt performance here… it's I/O that's going to be the bottleneck here. I'm not saying we should be reading in 16kB chunks, but even 100MB seems excessive. There's just no point in keeping that much in memory for any period of time.

> We are assigning the `QByteArray` `data_buffer` to the `QByteArray` returned by `read()`, so the `read` is doing all of the `QByteArray` allocation. If we do indeed do this, I should probably ensure the size of `data_buffer` does match `bytes_available` so it is the expected size.

Right, sorry, was totally thinking `data_buffer` is of `max_bytes` size… :roll_eyes:

src/network/local_socket_reply.cpp
townsend2010: I made a helper method for all of the various `write()`'s, so this should be taken care of now.

src/network/local_socket_reply.cpp
townsend2010: Reworked all of this and should be better and more straightforward now.

src/network/local_socket_reply.cpp
townsend2010: I kept the `QByteArray`'s, but I tightened up the construction of the `char` data and should look much better now.

src/network/local_socket_reply.cpp
```diff
@@ -139,38 +143,105 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+    {
+        http_data += "?" + request.url().query();
+    }
+
     http_data += " HTTP/1.1\r\n";
 
     // Build the HTTP Host header
-    // Host can be anything, so we'll use 'multipass'
-    http_data += "Host: multipass\r\n";
+    auto host = request.url().host().toUtf8();
```
Saviq, line 156: Hmm `toUtf8()` feels wrong here?

src/network/local_socket_reply.cpp
```diff
@@ -139,38 +143,105 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
         {
+            auto content_length = request.header(QNetworkRequest::ContentLengthHeader).toByteArray();
+            auto transfer_encoding = request.rawHeader("Transfer-Encoding");
+            bool is_chunked{false};
+
+            if (!content_length.isEmpty() && !transfer_encoding.isEmpty())
+            {
+                throw mp::HttpLocalSocketException(
+                    "Both \'Content-Length\' and \'Transfer-Encoding\' headers cannot be set at the same time");
+            }
```
Saviq, line 190: I don't think that's true… You can't have it `chunked` and provide `Content-Length`, but you could have [any other value of `Transfer-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding).

Unless you mean that it's not supported, in which case reword the exception? :)

src/network/local_socket_reply.cpp
```diff
@@ -139,38 +143,105 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+            else if (!content_length.isEmpty())
+            {
+                http_data += "Content-Length: " + content_length + "\r\n";
+            }
+            else
+            {
+                http_data += "Transfer-Encoding: " + transfer_encoding + "\r\n";
+                is_chunked = true;
+            }
```
Saviq, line 204: Strictly, this should be `if "chunked" in transfer_encoding`, right?

src/network/local_socket_reply.cpp
```diff
@@ -139,38 +143,105 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+            {
+                auto bytes_read = outgoingData->read(data_buffer.data(), max_bytes);
+                if (bytes_read < 0)
+                {
+                    throw mp::HttpLocalSocketException(
+                        fmt::format("Cannot read data to send to socket: {}", outgoingData->errorString()));
+                }
+
+                if (bytes_read == 0)
+                    break;
```
Saviq, line 225: I'd probably make it a:

```cpp
int bytes_read;
while (bytes_read = outgoingData->read(...) >= 0)
{
    ...
}

if (bytes_read < 0)
{
    throw ...
}
```

I suppose I'm allergic to `while(true)`, will leave it up to you :)

src/network/local_socket_reply.cpp
```diff
@@ -139,38 +143,105 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
+
+                if (is_chunked && !local_socket_write(QByteArray::number(bytes_read, 16) + "\r\n"))
+                    return;
+
+                if (!local_socket_write(QByteArray::fromRawData(data_buffer.data(), bytes_read) + "\r\n"))
```
Saviq, line 230: Is trailing `\r\n` expected in non-chunked transfers? How is the server supposed to know it's not part of the data? o.O

I mean, if we exceed `max_bytes` in non-chunked transfers, we'll be putting `\r\n` in the middle of the data?

tests/linux/test_local_network_access_manager.cpp
```diff
@@ -244,6 +287,77 @@ TEST_F(LocalNetworkAccessManager, query_in_url_is_preserved)
...
+    auto server_response = [&http_response, &random_data](auto data) {
+        QByteArray first_part = random_data.mid(0, 32768);
+        QByteArray second_part = random_data.mid(32768);
+
+        // Implies the data was split, ie, chunked
+        EXPECT_FALSE(data.contains(random_data));
+
+        // Implies that the correct data was still received
+        EXPECT_TRUE(data.contains(first_part));
+        EXPECT_TRUE(data.contains(second_part));
```
Saviq, line 304: `.contains` feels like a weak test, taking into account the `\r\n` doubt I have above?

Saviq (changes requested):
Could use a test for non-chunked transfer above `max_bytes`, WDYT?

src/network/local_socket_reply.cpp
townsend2010: I'm getting `host` as a `QByteArray` from a `QString`.  There are only 3 ways that I know of to do that: `QString::toLatin1()`, `QString::toLocal8Bit() `, and `QString::toUtf8()`.  I've tended to use `toUtf8()` when needing to do this in the past.  Is there a better way?

src/network/local_socket_reply.cpp
townsend2010: Ah, ok, I misread the specs.  Then this check needs to be for `chunked` instead of `Transfer-Encoding` and I'll reword the error message.

src/network/local_socket_reply.cpp
townsend2010: Right, as mentioned above, I misread the spec :smile:

src/network/local_socket_reply.cpp
townsend2010: It all accomplishes the same thing and I suppose is a coder's preference.  I'll think about it some and look at it from a readability perspective.

src/network/local_socket_reply.cpp
townsend2010: Hmm, yeah, I'll look at this more closely and figure out what the implication is in the case you presented.

src/network/local_socket_reply.cpp
Saviq: Sorry, I just meant that hostname can't be UTF-8, so `::toLatin1()` feels like a better choice :)

tests/linux/test_local_network_access_manager.cpp
townsend2010: I kind of thought this test is not the best, but the alternative is having to implement a "server" that can parse different lengths of data transfers, etc. and I feel that is starting to take away from the actual implementation of what we are trying to accomplish.  That said, I'll look more into this and see if there is a better way.

townsend2010:
> Could use a test for non-chunked transfer above max_bytes, WDYT?

I'll look into that.

src/network/local_socket_reply.cpp
townsend2010: After discussion in chat, will change it to `toLatin1()` to look a little "weird" even though `Utf8()` works just as well.

townsend2010:
After giving it more thought, I'm reluctant to add an HTTP "validitor" myself to the unit tests.  My concern is that the very code that I'm validating is written by me, the same (mis)understanding will be brought into the HTTP validator as well.  What I suggest is finding some sort of external validator/linter that can do this.  I know Go has the ability to do HTTP over local sockets, so maybe finding some way to utilize that as an external process that the tests connect to would be better.  But that will take time and maybe that will be for later.

src/network/local_socket_reply.cpp
townsend2010: Done

src/network/local_socket_reply.cpp
townsend2010: Done.

src/network/local_socket_reply.cpp
townsend2010: Yep, done.

src/network/local_socket_reply.cpp
townsend2010: Ok, I "caved" and did it your way :laughing: But it does make it more readable.

src/network/local_socket_reply.cpp
townsend2010: Fixed

tests/linux/test_local_network_access_manager.cpp
townsend2010: As mentioned below, I really don't think I should be the one validating the very HTTP that I'm creating in `LocalSocketReply::send_request()` as any misunderstandings of the spec will likely transfer over into the unit test validation.

For now, I'm going to add a TODO to figure out a better way to validate HTTP in the long term and in the short term just make it so coverage is satisfied.  WDYT?

tests/linux/test_local_network_access_manager.cpp
```diff
@@ -123,7 +123,7 @@ struct LocalSocketWriteErrorTestSuite : LocalNetworkAccessManager, WithParamInte
 {
 };
 
-const std::vector<int> local_socket_write_error_suite_inputs{0, 1, 2, 3, 4, 5};
+const std::vector<int> local_socket_write_error_suite_inputs{0, 1, 2, 3, 4, 5, 6};
```
Saviq: This feels like we'll forget to update it :)

[`Range(begin, end [, step])`](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-value-parameterized-tests), too :)

Saviq (approved):
Let's go!

bors merge

tests/linux/test_local_network_access_manager.cpp
townsend2010: I agree, but it needs to match the number of `local_socket_write()` calls in `LocalSocketReply::send_request()`.  And sure, it could be a range as well, but this was quick to implement :smiley:

PR #1710, opened 2020-08-27
[ci] fix snap publishing

Saviq:


Saviq:
For reference: I failed to actually _use_ the inputs I defined in the refactor:

https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action#creating-an-action-metadata-file

PR #1730, opened 2020-09-10
Add common ensure_vm_is_running_for() function

townsend2010:
Also, instead of changing the instance state to detect if an instance is shutdown while starting, use a flag.


luis4a0 (approved):
Looks good. Apart from the tests, I tested it.

luis4a0:
Private also works, so
bors r+

PR #1736, opened 2020-09-15
[lxd] Disable secure boot for all instances

townsend2010:
- This is necessary because some architectures and images that LXD supports does not work with secure boot.


Saviq (approved):
Yup.

bors merge

PR #1737, opened 2020-09-16
[detail] Fix a couple of shadowing vars

ricab:
CLion was complaining about these.


townsend2010 (approved):
Thanks, LGTM!

bors merge

Issue #1475, opened 2020-04-08
QEMU backend fails to realize dnsmasq died (and doesn't restart it)


markshuttle:
On my Ubuntu 20.04 laptop I am unable to launch new instances:

mark@thrifty:~$ multipass launch ubuntu
launch failed: The following errors occurred:                                   
memorable-topminnow: timed out waiting for response
mark@thrifty:~$ multipass launch ubuntu                                         
launch failed: The following errors occurred:                                   
arriving-rabbit: timed out waiting for response

Interestingly, qemu is still running for both attempts:

mark@thrifty:~$ ps ax | grep qemu                                               
1904243 ?        Sl     0:21 /snap/multipass/1784/usr/bin/qemu-system-x86_64 --enable-kvm -device virtio-scsi-pci,id=scsi0 -drive file=/var/snap/multipass/common/data/multipassd/vault/instances/memorable-topminnow/ubuntu-18.04-server-cloudimg-amd64.img,if=none,format=qcow2,discard=unmap,id=hda -device scsi-hd,drive=hda,bus=scsi0.0 -smp 1 -m 1024M -device virtio-net-pci,netdev=hostnet0,id=net0,mac=52:54:00:e0:74:c4 -netdev tap,id=hostnet0,ifname=tap-97e3bbd7047,script=no,downscript=no -qmp stdio -cpu host -chardev null,id=char0 -serial chardev:char0 -nographic -cdrom /var/snap/multipass/common/data/multipassd/vault/instances/memorable-topminnow/cloud-init-config.iso
1904495 ?        Sl     0:20 /snap/multipass/1784/usr/bin/qemu-system-x86_64 --enable-kvm -device virtio-scsi-pci,id=scsi0 -drive file=/var/snap/multipass/common/data/multipassd/vault/instances/arriving-rabbit/ubuntu-18.04-server-cloudimg-amd64.img,if=none,format=qcow2,discard=unmap,id=hda -device scsi-hd,drive=hda,bus=scsi0.0 -smp 1 -m 1024M -device virtio-net-pci,netdev=hostnet0,id=net0,mac=52:54:00:af:05:38 -netdev tap,id=hostnet0,ifname=tap-c854facc299,script=no,downscript=no -qmp stdio -cpu host -chardev null,id=char0 -serial chardev:char0 -nographic -cdrom /var/snap/multipass/common/data/multipassd/vault/instances/arriving-rabbit/cloud-init-config.iso

**Logs**
[...]

PR #1744, opened 2020-09-18
Drop dnsmasq orphan adoption

### LINKED ISSUE #1741: Stale dnsmasq.pid can prevent IP allocation due to PID clash
**Describe the bug**
If a stale `dnsmasq.pid` exists from a previous execution and another running process has the PID, instances fail to boot properly with the error:

```
failed to determine IP address
```

PID allocation on systems that don't reboot often means this is unlikely, but if you reboot daily [ for instance ] , it becomes much more likely.  I have seen it happen , so it is at least possible :) 

**To Reproduce**
Reboot a system
If you are unlucky, and one of the running PIDs matches the contents of `dnsmasq.pid`, then dnsmasq will not start instances fail to launch.

To reliably reproduce:
Stop multipassd
echo {PID_OF_A_RUNNING_PROCESS} > /var/snap/multipass/common/data/multipassd/network/dnsmasq.pid
Start multipassd
Try and launch an instance

**Expected behavior**
`dnsmasq.pid` should either be purge on `multipassd` stop, or additional logic added to check for a process signature that matches dnsmaq fore assuming dnsmaqk is running.

Instances should still be able to launch.  A non-technical user will find it hard to debug this issue, and it could the cuase of some other tickets I've seen:
https://github.com/canonical/multipass/issues/1653
https://github.com/canonical/multipass/issues/1584

**Logs**
Also see this in the service logs:
```
Sep 18 11:40:02 hostname multipassd[27492]: Looking for dnsmasq
Sep 18 11:40:02 hostname multipassd[27492]: Read pid "27405" from file "/var/snap/multipass/common/data/multipassd/network/dnsmasq.pid"
Sep 18 11:40:02 hostname multipassd[27492]: existing dnsmasq found with pid 27405
```

**Additional info**
Ubuntu 18.04.5

multipass  1.4.0
multipassd 1.4.0

ricab:
Fixes #1741 (and possibly #1475).

### DIFF
--- a/src/platform/backends/qemu/dnsmasq_process_spec.cpp
+++ b/src/platform/backends/qemu/dnsmasq_process_spec.cpp
@@ -25,13 +25,8 @@ namespace mp = multipass;
 namespace mu = multipass::utils;
 
 mp::DNSMasqProcessSpec::DNSMasqProcessSpec(const mp::Path& data_dir, const QString& bridge_name,
-                                           const QString& pid_file_path, const std::string& subnet,
-                                           const QString& conf_file_path)
-    : data_dir(data_dir),
-      bridge_name(bridge_name),
-      pid_file_path{pid_file_path},
-      subnet{subnet},
-      conf_file_path{conf_file_path}
+                                           const std::string& subnet, const QString& conf_file_path)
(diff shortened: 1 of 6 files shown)


ricab:
Still need to fix fake dnsmasq processes now being left over by tests.

PR #1752, opened 2020-09-24
[lxd] Detect connection failures with the LXD socket

townsend2010:
Throw new LocalSocketConnectionException type and handle accordingly.

### DIFF
--- a/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
@@ -19,6 +19,7 @@
 #include "lxd_virtual_machine.h"
 #include "lxd_vm_image_vault.h"
 
+#include <multipass/exceptions/local_socket_connection_exception.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/utils.h>
@@ -67,7 +68,28 @@ void mp::LXDVirtualMachineFactory::prepare_instance_image(const mp::VMImage& /* 
 
 void mp::LXDVirtualMachineFactory::hypervisor_health_check()
 {
-    auto reply = lxd_request(manager.get(), "GET", base_url);
+    QJsonObject reply;
+
+    try
+    {
+        reply = lxd_request(manager.get(), "GET", base_url);
+    }
+    catch (const LocalSocketConnectionException& e)
+    {
+        auto error_code = e.get_error();
+
+        if (error_code == QLocalSocket::ConnectionRefusedError)
+        {
+            throw std::runtime_error(
+                fmt::format("{}\nPlease ensure the LXD snap interface is connected via `snap connect "
+                            "multipass:lxd`\nand that the LXD snap is enabled and running.",
+                            e.what()));
+        }
+        else if (error_code == QLocalSocket::ServerNotFoundError)
+        {
+            throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and running.", e.what()));
+        }
+    }
 
     if (reply["metadata"].toObject()["auth"] != QStringLiteral("trusted"))
     {
--- a/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
@@ -76,19 +76,10 @@ void mp::LXDVirtualMachineFactory::hypervisor_health_check()
     }
     catch (const LocalSocketConnectionException& e)
     {
-        auto error_code = e.get_error();
 
-        if (error_code == QLocalSocket::ConnectionRefusedError)
-        {
-            throw std::runtime_error(
-                fmt::format("{}\nPlease ensure the LXD snap interface is connected via `snap connect "
-                            "multipass:lxd`\nand that the LXD snap is enabled and running.",
-                            e.what()));
-        }
-        else if (error_code == QLocalSocket::ServerNotFoundError)
-        {
-            throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and running.", e.what()));
-        }
+        throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and enabled. Also make sure\n"
+                                             "the LXD interface is connected via `snap connect multipass:lxd lxd`.",
+                                             e.what()));
     }
 
     if (reply["metadata"].toObject()["auth"] != QStringLiteral("trusted"))
--- a/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
@@ -77,9 +77,10 @@ void mp::LXDVirtualMachineFactory::hypervisor_health_check()
     catch (const LocalSocketConnectionException& e)
     {
 
-        throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and enabled. Also make sure\n"
-                                             "the LXD interface is connected via `snap connect multipass:lxd lxd`.",
-                                             e.what()));
+        throw std::runtime_error(
+            fmt::format("{}\n\nPlease ensure the LXD snap is installed and enabled. Also make sure\n"
+                        "the LXD interface is connected via `snap connect multipass:lxd lxd`.",
+                        e.what()));
     }
 
     if (reply["metadata"].toObject()["auth"] != QStringLiteral("trusted"))
--- a/src/platform/backends/lxd/lxd_vm_image_vault.cpp
+++ b/src/platform/backends/lxd/lxd_vm_image_vault.cpp
@@ -19,6 +19,7 @@
 #include "lxd_request.h"
 
 #include <multipass/exceptions/aborted_download_exception.h>
+#include <multipass/exceptions/local_socket_connection_exception.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/network_access_manager.h>
@@ -187,6 +188,11 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const 
     {
         // Instance doesn't exist, so move on
     }
+    catch (const LocalSocketConnectionException& e)
+    {
+        mpl::log(mpl::Level::warning, category, fmt::format("{} - returning blank image info", e.what()));
+        return VMImage{};
+    }
     catch (const std::exception&)
     {
         // Image doesn't exist, so move on
@@ -274,6 +280,13 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
     catch (const LXDNotFoundException&)
     {
         return false;
+    }
+    catch (const LocalSocketConnectionException& e)
+    {
+        mpl::log(mpl::Level::warning, category,
+                 fmt::format("{} - Unable to determine if \'{}\' exists", e.what(), name));
+        // Assume instance exists until it knows for sure
+        return true;
     }
 }
 
@@ -550,6 +563,10 @@ QJsonArray mp::LXDVMImageVault::retrieve_image_list()
     catch (const LXDNotFoundException&)
     {
         // ignore exception
+    }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::warning, category, e.what());
     }
 
     return image_list;
--- a/src/platform/backends/lxd/lxd_vm_image_vault.cpp
+++ b/src/platform/backends/lxd/lxd_vm_image_vault.cpp
@@ -564,7 +564,7 @@ QJsonArray mp::LXDVMImageVault::retrieve_image_list()
     {
         // ignore exception
     }
-    catch (const std::exception& e)
+    catch (const LocalSocketConnectionException& e)
     {
         mpl::log(mpl::Level::warning, category, e.what());
     }
--- a/include/multipass/exceptions/local_socket_connection_exception.h
+++ b/include/multipass/exceptions/local_socket_connection_exception.h
@@ -18,8 +18,6 @@
 #ifndef MULTIPASS_LOCAL_SOCKET_CONNECTION_EXCEPTION_H
 #define MULTIPASS_LOCAL_SOCKET_CONNECTION_EXCEPTION_H
 
-#include <QLocalSocket>
-
 #include <stdexcept>
 
 namespace multipass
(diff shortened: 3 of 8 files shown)


townsend2010:
Converting to draft while I add some tests.

townsend2010:
Ok, this is ready for review now.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -67,7 +68,28 @@ void mp::LXDVirtualMachineFactory::prepare_instance_image(const mp::VMImage& /*
...
+    }
+    catch (const LocalSocketConnectionException& e)
+    {
+        auto error_code = e.get_error();
+
+        if (error_code == QLocalSocket::ConnectionRefusedError)
+        {
+            throw std::runtime_error(
+                fmt::format("{}\nPlease ensure the LXD snap interface is connected via `snap connect "
+                            "multipass:lxd`\nand that the LXD snap is enabled and running.",
```
Saviq, line 85: This needs to be `… lxd`, otherwise:
```
error: snap "core" has no "lxd" interface slots
```

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -67,7 +68,28 @@ void mp::LXDVirtualMachineFactory::prepare_instance_image(const mp::VMImage& /*
...
+                            e.what()));
+        }
+        else if (error_code == QLocalSocket::ServerNotFoundError)
+        {
+            throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and running.", e.what()));
+        }
```
Saviq, line 91: I wonder if we should differentiate the message here? Sure, we can say what the underlying problem was, but the "solution", maybe we can just say the same across the board - "make sure the LXD snap is installed and that the `lxd` interface is connected" (sprinkle words to taste)?

As for "running"… LXD is socket-activated, so the socket should always be there, as long as it's installed. But maybe that's going into too much tech detail.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -275,6 +281,13 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
     {
         return false;
     }
+    catch (const LocalSocketConnectionException& e)
+    {
+        mpl::log(mpl::Level::warning, category,
+                 fmt::format("{} - Unable to determine if \'{}\' exists", e.what(), name));
+        // Assume instance exists until it knows for sure
+        return true;
```
Saviq, line 289: This looks weird, shouldn't we instead raise the error further to avoid what `return false` would result in?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -551,6 +564,10 @@ QJsonArray mp::LXDVMImageVault::retrieve_image_list()
     {
         // ignore exception
     }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::warning, category, e.what());
+    }
```
Saviq, line 570: Catch-all on purpose? With the richer exceptions now, couldn't we get rid of the catch-all further up, too?

Saviq (changes requested):
Having disconnected the lxd interface, I'm getting:
```
[2020-09-29T13:42:38.125] [debug] [lxd factory] Failed to authenticate to LXD:
[2020-09-29T13:42:38.125] [debug] [lxd factory] unix:///var/snap/lxd/common/lxd/unix.socket@1.0: {"metadata":null}
launch failed: Failed to authenticate to LXD.
```

Would be useful to give more info.
```

**Comment by townsend2010** [comment]
@Saviq,

I've had different results when disconnecting the lxd interface.  I just tried it again and this is what I get:
```
Sep 29 08:33:50 Ghost multipassd[2596088]: Requesting LXD: GET unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0/virtual-machines/test-detect/state?project=multipass
Sep 29 08:33:50 Ghost audit[2596088]: AVC apparmor="DENIED" operation="connect" profile="snap.multipass.multipassd" name="/var/snap/lxd/common/lxd/unix.socket" pid=2596088 comm="multipassd" requested_mask="wr" denied_mask="wr" fsuid=0 ouid=0
Sep 29 08:33:50 Ghost audit[2596088]: AVC apparmor="DENIED" operation="connect" profile="snap.multipass.multipassd" name="/var/snap/lxd/common/lxd/unix.socket" pid=2596088 comm="multipassd" requested_mask="wr" denied_mask="wr" fsuid=0 ouid=0
Sep 29 08:33:50 Ghost multipassd[2596088]: Cannot connect to /var/snap/lxd/common/lxd/unix.socket: QLocalSocket::connectToServer: Socket access error
Sep 29 08:33:50 Ghost multipassd[2596088]: Requesting LXD: GET unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0/virtual-machines/test-detect?project=multipass
Sep 29 08:33:50 Ghost multipassd[2596088]: Cannot connect to /var/snap/lxd/common/lxd/unix.socket: QLocalSocket::connectToServer: Socket access error - returning blank image info
```
What steps are doing exactly to get what you see and how are you getting that output?

File: src/platform/backends/lxd/lxd_virtual_machine_factory.cpp (hunk 1 above)
Comment (townsend2010) [reply src/platform/backends/lxd/lxd_virtual_machine_factory.cpp:85]: Right, will fix.

File: src/platform/backends/lxd/lxd_virtual_machine_factory.cpp (hunk 2 above)
Comment (townsend2010) [reply src/platform/backends/lxd/lxd_virtual_machine_factory.cpp:91]: Well, we can detect connection refused as opposed to the socket just not existing, so I thought it would be better to give a user a better clue as to how to fix their issue if there is permission denied, ie, the interface is not connected.  Also, if the LXD snap is disabled, it still leaves the socket hanging around, but nothing listening on it, so we'll get connection refused for that as well hence the "enabled and running" part of that message.

Server not found mainly implies LXD is not installed hence the shorter help message.

File: src/platform/backends/lxd/lxd_vm_image_vault.cpp (hunk 3 above)
Comment (townsend2010) [reply src/platform/backends/lxd/lxd_vm_image_vault.cpp:289]: I considered that, but then it will require quite a bit of change in the `Daemon` class' ctor and felt it was way too much effort to basically accomplish what this does.  In the end, we know there is supposed to be an instance, but we cannot determine it's state, so we set it `UNKNOWN` and it's not startable.

File: src/platform/backends/lxd/lxd_vm_image_vault.cpp (hunk 4 above)
Comment (townsend2010) [reply src/platform/backends/lxd/lxd_vm_image_vault.cpp:570]: I suppose this could be a `LocalSocketConnectionException` and let other exception types bubble up.  I'm unsure what catch-all further up you are referring to.

File: src/platform/backends/lxd/lxd_vm_image_vault.cpp [hunk 5]
```diff
@@ -187,6 +188,11 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
     {
         // Instance doesn't exist, so move on
     }
+    catch (const LocalSocketConnectionException& e)
+    {
+        mpl::log(mpl::Level::warning, category, fmt::format("{} - returning blank image info", e.what()));
+        return VMImage{};
+    }
```
Saviq, line 195: > I'm unsure what catch-all further up you are referring to.

I meant the catch-all just below here, but your other answer in the other PR explains why that. We should definitely fix these to be more specific in the other image hosts at some point.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
Saviq: Yeah, the problem is that it may well say the interface isn't connected after they install.

I just want to avoid this:
```plain
$ multipass launch
Please ensure LXD installed…
$ snap install lxd
$ multipass launch
Please ensure LXD interface connected…
```
:rage:

If it's uninstalled, let's tell them to install, run and connect; If it's not running, tell them to run and connect… You catch my drift :)

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: Right, sounds like a bigger refactoring with a common `mp::ShrugException`. Shall we have a TODO or an issue?

Saviq (commented):
> What steps are doing exactly to get what you see and how are you getting that output?

I disconnected the interface… But now can't reproduce, but am getting a similarly unhelpful:
```
$ multipass sh
launch failed: Cannot connect to /var/snap/lxd/common/lxd/unix.socket: 3
```

It seems we're still not resilient enough on startup, though:

```bash
$ snap disconnect multipass:lxd
$ snap restart multipass
# in the logs
wrz 29 15:19:34 michal-laptop systemd[1]: Started Service for snap application multipass.multipassd.
wrz 29 15:19:34 michal-laptop multipassd[900010]: propsReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.328\" (uid=0 pid=900010 comm=\"/snap/multipass/2644/bin/multipassd --verbosity d>
wrz 29 15:19:34 michal-laptop multipassd[900010]: nmReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.328\" (uid=0 pid=900010 comm=\"/snap/multipass/2644/bin/multipassd --verbosity de\">
wrz 29 15:19:34 michal-laptop multipassd[900010]: "Object path cannot be empty"
wrz 29 15:19:34 michal-laptop multipassd[900010]: gRPC listening on unix:/var/snap/multipass/common/multipass_socket, SSL:on
wrz 29 15:19:34 michal-laptop multipassd[900010]: Requesting LXD: GET unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0/images?recursion=1&project=multipass
wrz 29 15:19:34 michal-laptop multipass.multipassd[900010]: [error] [daemon] Caught an unhandled exception: Cannot connect to /var/snap/lxd/common/lxd/unix.socket: 3
wrz 29 15:19:34 michal-laptop systemd[1]: snap.multipass.multipassd.service: Main process exited, code=exited, status=1/FAILURE
```

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Ah, ok, I do catch your drift.  So yeah, One Message to Rule Them All!

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: I honestly don't think it's necessary to refactor as in the end, it will accomplish exactly the same thing as far as the daemon is concerned.  All we are avoiding here is for the daemon to remove the instance from it's internal map along with the removing it from the JSON.  We don't want it to do that until it knows for sure what the status of the instance is from LXD.

Saviq:
> > What steps are doing exactly to get what you see and how are you getting that output?
> 
> I disconnected the interface… But now can't reproduce, but am getting a similarly unhelpful:
> …
> It seems we're still not resilient enough on startup, though:

Sorry. Wrong snap.

But yes, on `multipass shell` I get the authentication error, if disconnected.

townsend2010:
Ok, yes, with `multipass shell`, I do see this.  Will work on fixing that.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: The method name is `has_record_for()`. Returning `true` when we don't actually know is just not something I'm comfortable with in the long run :)

Renaming the function to `should_keep_instance()`, I could maybe be convinced with, but it feels like moving too much of the logic into the vault.

I'd be happiest with introducing an exception type common to all vaults, converting the error here to that one, or a child exception type, and interpreting that in the daemon. That doesn't need to happen in this PR, though.

townsend2010: Thinking about it, perhaps we really don't need this function at all for either vault.  I think that is what should be investigated first in the future.  Not sure a TODO is the right thing for this since it affects multiple places in the code.  I guess a bug would be better even though it's quite technical in nature.

Saviq: #1766 then.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Ok, fixed.  Let me know if you want it tweaked (or propose what you want :smiley: )

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
townsend2010: Ok, there is now only one message for all socket connection failure types.  Note that the actual error string is prepended so there is still a clue about what the failure is.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Changed to `LocalSocketConnectionException`.

townsend2010:
@Saviq,

The `shell` (and `launch`) failure when the interface is not connected should be fixed now.

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -67,7 +68,19 @@ void mp::LXDVirtualMachineFactory::prepare_instance_image(const mp::VMImage& /*
...
+
+    try
+    {
+        reply = lxd_request(manager.get(), "GET", base_url);
+    }
+    catch (const LocalSocketConnectionException& e)
+    {
+
+        throw std::runtime_error(fmt::format("{}\nPlease ensure the LXD snap is installed and enabled. Also make sure\n"
+                                             "the LXD interface is connected via `snap connect multipass:lxd lxd`.",
```
Saviq, line 81:
```suggestion
        throw std::runtime_error(fmt::format("{}\n\nPlease ensure the LXD snap is installed and enabled. Also make sure\n"
                                             "the LXD interface is connected via `snap connect multipass:lxd lxd`.",
```

Saviq (approved):
Just a tiny tweak on the message and get it in, please.

Saviq:
bors delegate+

townsend2010:
Ugh, bad rebase :frowning_face: 

Will fix in a while.

Saviq:
> Ugh, bad rebase :frowning_face:

At least you verified the CTest output ;)

townsend2010:
Not a bad rebase, it's that I didn't change a couple of tests to reflect the new `\n` in the help message.

> At least you verified the CTest output ;)

Indeed!

townsend2010:
Ok, let's do it!

bors r=Saviq

townsend2010:
Glad the cmake test output on failure is now working.  It seems `DNSMasqServer.dnsmasq_fails_and_throws` is the test that is flaky (not related to this PR).

bors r=Saviq

townsend2010:
bors, why you bein' so sensitive on this test??????  Try it one more time...

bors r=Saviq

PR #1760, opened 2020-09-28
[gha] do not allow coverage to fail

Saviq:


townsend2010 (approved):
There was an error in the coverage run of the tests and it failed this time, so this does what it's supposed too.

bors merge

PR #1762, opened 2020-09-28
[lxd_image_vault] Get correct info for instance with custom image

townsend2010:


src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -163,24 +163,44 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
+                }
 
-            for (const auto& alias : info.aliases)
+                return source_image;
+            }
+            catch (const std::exception&)
             {
-                source_image.aliases.push_back(alias.toStdString());
+                // Ignore
             }
```
Saviq, line 203: Could we be stricter here?

Saviq (changes requested):
Since we can… I wonder if we should just store the metadata with the instance image? This way, even when we remove the data from the custom image host, it'd just be there in LXD?

Don't we do the same in the default vault? Copy all the image data into the instance vault?

townsend2010:
@Saviq,

>  I wonder if we should just store the metadata with the instance image?

We do and that is what I'm using here now to get the image info for any particular instance using a custom image.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Well, `info_for_full_hash()` will throw a `std::runtime_error` when it doesn't find a matching hash, but there are multiple image hosts, so we need to ignore that exception so we can move on in the loop.  So the only better granularity we can have here is to catch the `runtime_error` explicitly, but that is quite generic in and of itself.

Saviq: Add a TODO, please.

townsend2010: A TODO for what?  Change the exception type that `info_for_full_hash()` throws?  I don't think changing this to catch `std::runtime_error` is the right thing to do.

Saviq: Yes, for refactoring so that `info_for_full_hash` throws specific exceptions.

townsend2010: Ok, I added a TODO here and in the place in `ubuntu_image_host.cpp` where it really needs to be implemented.

Saviq:
Yups.

bors merge

PR #1767, opened 2020-09-29
[gha] trigger on "Ready for review"

Saviq:


ricab (commented):
Why on Linux? Does this make sense when it would have already run on drafts?

Saviq:
> Why on Linux? Does this make sense when it would have already run on drafts?

Because of this, which just landed:

https://github.com/canonical/multipass/blob/346644970903bac0fe4e7d15caf550374dca7d03/.github/workflows/linux.yml#L216-L231

We need the dispatch to know when it's out of draft.

ricab (approved):
Yep.

bors r+

ricab:
bors retry

PR #1768, opened 2020-09-29
[gha] enable verbose CTest on failure

Saviq:


townsend2010 (approved):
Makes sense...no real way to know if this does do what we want it to until there is a failure.

bors merge

Saviq:
> Makes sense...no real way to know if this does do what we want it to until there is a failure.

You can easily check with `CTEST_OUTPUT_ON_FAILURE=1 make test` locally.

townsend2010:
> You can easily check with CTEST_OUTPUT_ON_FAILURE=1 make test locally.

Yes, and I did that, but I have no local test failures, so the output is the same :smiley:

PR #1769, opened 2020-09-29
[gha] fix build params with multi-line commit messages

Saviq:
The message in context is full, we need just the subject, and that's not available in the context.


Saviq:
bors try

ricab (approved):
Makes sense.

bors merge

PR #1771, opened 2020-09-30
[gha] move CTest environment variable

Saviq:
The global variable does not transfer into the LXD environment.


townsend2010 (approved):
This should definitely do it.

bors merge

PR #1782, opened 2020-10-06
[lxd] Use image fingerprint instead of alias when downloading

townsend2010:
This also changes how Multipass updates and prunes images since it can no longer use the `refresh` endpoint since that doesn't work with fingerprint-based launches.


townsend2010:
Hmm, why is it complaining about no coverage uploaded for base (`master@d9a7264`) when it's right here: https://codecov.io/gh/canonical/multipass/commit/d9a72641a254a0613fe4344aa3cb93a67a6b055b

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -421,17 +432,26 @@ void mp::LXDVMImageVault::lxd_download_image(const QString& id, const QString& s
...
-    else
+    QJsonObject image_object{{"source", source_object}};
+
+    auto release = QString::fromStdString(query.release);
+
+    if (!id.startsWith(release))
     {
-        source_object.insert("alias", release);
-    }
+        QJsonObject properties_object{{"query.release", "default"},
```
Saviq, line 443: Shouldn't we use the the exact alias the user requested? So that when we update, we update for the requested alias, which could very well be "lts"?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -421,17 +432,26 @@ void mp::LXDVMImageVault::lxd_download_image(const QString& id, const QString& s
...
+                                      {"query.remote", QString::fromStdString(query.remote_name)}};
 
-    QJsonObject image_object{{"source", source_object}};
+        // Need to save the original image's last_used_at as a property since there is no way to modify the
+        // new image's last_used_at field.
+        if (!last_used.isEmpty())
+        {
+            properties_object.insert("last_used_at", last_used);
+        }
```
Saviq, line 451: Can't we rely on `last_used_at` being empty until it's used?

Saviq (commented):
The one problem I see with approach this is that the if someone launches via `default`, `lts`, `20.04` etc., it will not result in the update to DTRT. It may well be a problem with our default host implementation, would have to read in.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Good catch! This is cruft left over from testing. I need to switch it back to be `id`.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Which `last_used_at`?  The one LXD stores in the main image object or this one here I'm introducing?

townsend2010:
> The one problem I see with approach this is that the if someone launches via default, lts, 20.04 etc., it will not result in the update to DTRT.

How so?  I'm not sure I follow.  Could you give an example of what you see as being wrong?

Saviq:
> How so? I'm not sure I follow. Could you give an example of what you see as being wrong?

As discussed, when you launch once by `20.04`, once by `lts` — when `lts` moves to `22.04`, we will fall behind.

townsend2010:
> As discussed, when you launch once by 20.04, once by lts — when lts moves to 22.04, we will fall behind.

Right, this is systemic to both image vaults.  As we agreed, this will be fixed later.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Fixed

Saviq:
This seems to have gotten broken by #1790?

townsend2010:
> This seems to have gotten broken by #1790?

I reproduced the issue using the PR snap and the issue is actually caused by #1806 and will be fixed by #1809.

Saviq:
OK this was my LXD's state being weird:
```plain
"Create instance from image: Failed to run: zfs destroy rpool/lxd/images/2f456602cd86a6dc4d50102f028729421fece7e68f7c8aec6d8af5d9bbc681bf@readonly: cannot destroy 'rpool/lxd/images/2f456602cd86a6dc4d50102f028729421fece7e68f7c8aec6d8af5d9bbc681bf@readonly': snapshot has dependent clones\nuse '-R' to destroy the following datasets:\nrpool/lxd/virtual-machines/multipass_dazzling-hare\nrpool/lxd/virtual-machines/multipass_admirable-tick",
```

But we're not bubbling those errors up. So my 404 problem was that it tried to remove the instance after it failed to start - we should instead show the LXD error as above.

Any case, not this PR's problem.

bors merge

Saviq:
Once more, with feeling:

bors merge

PR #1784, opened 2020-10-06
[lxd] Wait on state operations to complete

townsend2010:
Fixes #1785

PR #1790, opened 2020-10-07
[lxd] Prefer lxd kvm images from SimpleStreams

townsend2010:


Saviq (changes requested):
From https://github.com/canonical/multipass/pull/1784#issuecomment-706141371:
> There seems to be more happening here… focal instances seem to reboot after first boot, so Multipass bails out, because the instance goes on→off→on…

This is actually expected in the kvm images, so I don't think we should land this until we deal with that.

Saviq (approved):
Yups.

bors merge

Issue #1515, opened 2020-04-29
Fix test failures on Windows and macOS


ricab:
These are the tests that fail on Windows ATTO~W~Edit:

```
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] ImageVault.custom_image_url_downloads
[  FAILED  ] ImageVault.invalid_image_alias_throw
[  FAILED  ] ImageVault.http_download_returns_expected_image_info
[  FAILED  ] TestQemuImgProcessSpec.apparmor_profile_running_as_symlinked_snap_correct
[  FAILED  ] HyperVBackend.creates_in_off_state
```

And on macOS, as of yesterday (copied from a travis build):

```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] Client.launch_cmd_custom_image_file_ok
[  FAILED  ] Client.launch_cmd_custom_image_http_ok
[  FAILED  ] ImageVault.custom_image_url_downloads
[  FAILED  ] ImageVault.invalid_image_alias_throw
[  FAILED  ] ImageVault.http_download_returns_expected_image_info
[  FAILED  ] HyperkitBackend.creates_in_off_state
```

ricab:
We've got new failures on windows:

```
[  FAILED  ] TestQemuImgProcessSpec.apparmor_profile_running_as_symlinked_snap_correct
[  FAILED  ] SftpServer.handles_readlink
[  FAILED  ] SftpServer.handles_symlink
[  FAILED  ] SftpServer.broken_symlink_does_not_fail
[  FAILED  ] SftpServer/Stat.handles/message_SFTP_LSTAT, where GetParam() = '\a' (7)
[  FAILED  ] SftpServer/Stat.handles/message_SFTP_STAT, where GetParam() = '\x11' (17)
```

Updating description.

ricab:
One failure gone: `Utils.trim_newline_assertion_works`. Updating description accordingly.

ricab:
And I noticed that the `Sftp*` failures weren't happening in GHA, so I [fiddled](https://community.perforce.com/s/article/3472) with [symlinks](https://dingyuliang.me/windows-symbolic-link-introduction/) and they started passing after a reboot :grin: 

Updating description.

sharder996:
Closing as these test are no longer causing failures. We no longer use Travis CI, so this is no longer an issue.

PR #1796, opened 2020-10-12
Disable failing tests

ricab:
Disable failing tests on Windows and macOS. This is meant as a temporary measure, so that we can otherwise fail CI on failing tests, thereby avoiding pile up of further test failures.

Addresses #1515.


Saviq:
Yups!

bors merge

PR #1805, opened 2020-10-20
Control network proposed tweaks

ricab:
A few proposed fixes and simplifications for the public bridging PR.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -153,7 +153,7 @@ auto make_cloud_init_network_config(const mp::NetworkInterface& default_interfac
     {
         if (extra_interfaces[i].auto_mode)
         {
-            std::string name = "extra" + std::to_string(i);
+            name = "extra" + std::to_string(i);
             network_data["ethernets"][name]["match"]["macaddress"] = extra_interfaces[i].mac_address;
             network_data["ethernets"][name]["dhcp4"] = true;
             network_data["ethernets"][name]["wakeonlan"] = "true";
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -766,15 +766,16 @@ std::unordered_set<std::string> mac_set_from(const mp::VMSpecs& spec)
     return macs;
 }
 
-// Move the contents of t to s. If both sets are disjoint, then s becomes sUt, t becomes undefined (for efficiency
-// reasons) and true is returned. If they are not disjoint, then s and t remain untouched and false is returned.
+// Move the contents of t to s. If both sets are disjoint, then s becomes sUt, t becomes valid but unspecified (for
+// efficiency reasons) and true is returned. If they are not disjoint, then s and t remain untouched and false is
+// returned.
 bool merge_if_disjoint(std::unordered_set<std::string>& s, std::unordered_set<std::string>& t)
 {
     for (const auto& elt : s)
         if (t.find(elt) != t.end())
             return false;
 
-    for (const auto& elt : t)
+    for (auto& elt : t)
         s.insert(std::move(elt));
 
     return true;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2172,7 +2172,7 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
         });
 
     prepare_future_watcher->setFuture(QtConcurrent::run([this, server, request, name,
-                                                         checked_args]() -> VirtualMachineDescription {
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -2217,40 +2217,32 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
-            // interfaces in a new vector.
-            std::vector<mp::NetworkInterface> extra_interfaces;
-            for (const auto& iface : checked_args.extra_interfaces)
-            {
-                extra_interfaces.push_back(mp::NetworkInterface{
-                    config->factory->interface_id(iface.id),
-                    iface.mac_address.empty() ? generate_unused_mac_address(added_mac_addresses) : iface.mac_address,
-                    iface.auto_mode});
-            }
-
```
luis4a0: There is a reason to check first the specified MAC address and generating addresses afterwards: all the addresses specified by the user must be added to `added_mac_addresses` first, so they are taken into account when generating new addresses. I'd like to keep that behavior.

OTOH, the non-const version is better, we can avoid copying the vector.

src/daemon/daemon.cpp
```diff
@@ -784,18 +784,12 @@ bool merge_if_disjoint(std::unordered_set<std::string>& s, std::unordered_set<st
...
-    {
-        mac_address = mp::utils::generate_mac_address();
-    }
-
-    s.insert(mac_address);
+    while (true)
+        if (auto [it, success] = s.insert(mp::utils::generate_mac_address()); success)
+            return *it;
 
-    return mac_address;
```
luis4a0: Sweet :)

src/daemon/daemon.cpp
```diff
@@ -786,10 +786,14 @@ std::string generate_unused_mac_address(std::unordered_set<std::string>& s)
...
     // should also check in the ARP table.
-    while (true)
+    static constexpr auto max_tries = 5;
+    for (auto i = 0; i < max_tries; ++i)
         if (auto [it, success] = s.insert(mp::utils::generate_mac_address()); success)
             return *it;
 
+    throw std::runtime_error{
+        fmt::format("Failed to generate an unique mac address after {} attempts. Number of mac addresses in use: {}",
+                    max_tries, s.size())};
```
luis4a0: Given that the MAC generator always generate addresses started in `52:54:00:`, we have 16 million possible MAC addresses. For this, I think we'll never generate even one repeated MAC. I think that error would indicate something very weird is happening. I'm ok with the change, although I don't foresee a possibility to fail.

luis4a0 (changes requested):
Great @ricab, thanks for this! I just left a couple comments, and think there is only one thing to be changed.

src/daemon/daemon.cpp
ricab: Exactly. I agree with what you say and that was exactly why I chose to add it with a relatively small number of retries. Chance of collision is very low if everything is correct, so we never expect 5 straight collisions. That would signal problems that would be difficult to understand otherwise.

Some thoughts to explain my reasoning. With a uniform mac distribution, if we have `n` allocated macs, the probability would be `(n/16^6)^5`. With `n=100` allocated macs, that would be on the order of `10^-27`. However, a small bug can radically change that. Suppose we have some sort of leak such that `n` reaches 10 million: the probability is now ~8% (that's why I included the number of used macs in the exception msg). Or suppose the generation is very biased and hits the same macs for some reason. Without this, it could just take longer to generate, which would be difficult to identify/understand.

Strictly speaking, exceptions are meant to deal with unexpected conditions, not to signal bugs - asserts should be used for that. However, we could conceivably hit an external condition causing repetition (like something wrong with the source of entropy in the OS or something). So, when I wrote the `while(true)` I got goosebumps and thought it was better this way.

src/daemon/daemon.cpp
ricab: Right, good point! I missed that. I would still prefer to condense the code somehow... Let me see if I can come up with something.

ricab: @luis4a0 I have just pushed a fix, WDYT?

Indeed, I don't see a way to escape the two loops. I thought to [partition](https://en.cppreference.com/w/cpp/algorithm/partition), but I didn't want to change the order that the user requested (and of course that internally would also loop).

luis4a0: Yes, partitioning is a possibility but it's better to keep the users order. The code you pushed works and it's clear enough, let's leave it like that.

src/daemon/daemon.cpp
luis4a0: I agree.

PR #1807, opened 2020-10-21
[snapcraft] use buildd images for core18 and core20

Saviq:


src/daemon/custom_image_host.cpp
```diff
@@ -68,16 +68,16 @@ const QMap<QString, CustomImageInfo> multipass_image_info{
 const QMap<QString, CustomImageInfo> snapcraft_image_info{
     {{"ubuntu-16.04-minimal-cloudimg-amd64-disk1.img"},
      {"https://cloud-images.ubuntu.com/minimal/releases/xenial/release/",
-      {"core", "core16"},
```
sergiusens, line 71: Please do not remove core, it is a valid use case:

- as a trigger for Snapcraft 3.0 functionality
- as the default when using multipass without a base (e.g.; OS X).

It is also documented here https://snapcraft.io/docs/base-snaps

sergiusens (changes requested):
multipass snap works fine, when using base core20 and core18, it does not though for base core.

This:

```yaml
name: play-core
base: core
version: '0.1'
summary: test using core as base
description: |
  filler...
  ...

grade: devel
confinement: devmode

parts:
  my-part:
    plugin: nil
```

results in:

```
$ snapcraft
Launching a VM.
launch failed: Unable to find an image matching "core"                          
An error occurred with the instance when trying to launch with 'multipass': returned exit code 2.
```

sergiusens:
Considering the linked documentation and the fact that core16 is not ready, perhaps drop that instead.

src/daemon/custom_image_host.cpp
Saviq: Of course. I confused the two… guess why :P

Fixed.

Saviq:
@sergiusens I took your YAML and ran this on qemu (they don't boot yet on LXD):

```bash
$ bash -euo pipefail -c '
for base in core{20,18,}; do
  sed -i "s/^base:.*/base: $base/" snapcraft.yaml
  snapcraft
  multipass delete --purge --all
done'
```

The same for `core{18,20}` on macOS on hyperkit and VirtualBox. Also launched those two on Windows on Hyper-V and VirtualBox.

All good from my POV.

src/platform/platform_linux.cpp
```diff
@@ -154,6 +154,12 @@ bool mp::platform::link(const char* target, const char* link)
 
 bool mp::platform::is_alias_supported(const std::string& alias, const std::string& remote)
 {
+    auto driver = utils::get_driver_str();
+
+    // snapcraft:core{18,20} images don't work on LXD yet
```
townsend2010, line 159: What is the status of these images working with LXD and why don't they work?

Saviq: They don't have working UEFI booting. It's on their radar. I know this somewhat regresses the LXD backend, but the alternative is to keep `minimal` images for LXD, and use `buildd` for others… I'd rather not.

I was also wondering whether we should disallow the `snapcraft:` remote on LXD altogether.

townsend2010: Ok, this was all taken care of on chat.

townsend2010 (changes requested):
I think this is good once LXD support for the `snapcraft:` remote is disabled.

townsend2010 (changes requested):
Test coverage dropped in the last commit.  Probably need to add back in something like https://github.com/canonical/multipass/pull/1679/files#diff-7ef9608cecf497ebab4b6ca882daa3f62d9438804c5072ee37257e08806acf54.

Saviq:
> Test coverage dropped in the last commit.

Fixed.

townsend2010 (approved):
LGTM

bors merge

PR #1809, opened 2020-10-22
Use image vault for determining default disk size

### LINKED ISSUE #1806: [lxd] Default disk size on launch is undefined
When using the LXD backend, if the client does not set a disk size on `launch`, the resulting default disk size is undefined and can potentially lead to a wildly large requested disk size.

The reason is that the disk `MemorySize` `optional` is set to `nullopt`, but when using the LXD backend, the `optional` is simply dereferenced without checking if it's actually a `nullopt` or not.  If the resulting junk value is larger than `10737418240` bytes, that junk disk size will be requested.

The quick fix here is to check if the disk size is a `nullopt`, but there is a `TODO` stating that the image vault should be responsible for setting this disk size.  I agree that using the image vault is the correct way to fix this and the resulting PR will implement this.

townsend2010:
Fixes #1806


tests/test_image_vault.cpp
```diff
@@ -117,6 +120,42 @@ struct ImageVault : public testing::Test
         hosts.push_back(&host);
     }
 
+    QByteArray fake_img_info(const mp::MemorySize& size)
```
luis4a0, line 123: Yeah, moving size check tests here seems to be the right thing.

src/daemon/daemon.cpp
```diff
@@ -2010,17 +1973,8 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
-                if (driver == "lxd")
-                {
-                    disk_space = *checked_args.disk_space;
-                }
-                else
-                {
-                    disk_space = compute_final_image_size(vm_image, checked_args.disk_space);
-                }
+                const auto image_size = config->vault->minimum_image_size_for(vm_image.id);
+                const auto disk_space = compute_final_image_size(image_size, checked_args.disk_space);
```
luis4a0, line 1977: Good, getting rid of backend-dependent code.

luis4a0 (approved):
I tested, this works. All tests pass. I like some refactors on the code.

luis4a0:
bors +r

luis4a0:
Yes, I did mean "r+"

luis4a0:
Ok, you want me to write the full thing:
bors r+

Saviq:
bors r=luis4a0

PR #1811, opened 2020-10-23
Linux platform functions to get network interfaces

luis4a0:
This PR adds the functions needed in Linux to get information about network interfaces present in the system. This will be needed to implement bridging in some backends.
Namely, this add the functions `platform::get_network_interfaces_info()` to get information about all the interfaces on the system.
This code works confined, which complicates the detection of some interfaces.


ricab:
Hey @luis4a0, I think you brought one commit too many: the first one, titled "Add platform-dependent network info interface". 

I thought this would be stacked on the bridging branch. This would replace the throwing platform_linux implementation and would be rebased on top of public/control. It doesn't achieve much without that anyway, right?

To test bridging with vbox on Linux, one would checkout the private/control branch and merge with this one locally. Does that make sense?

luis4a0:
Right, stacking on top of the `control-...` branch makes sense, and more knowing that GH will move the base once it gets merged. I'll do that.

ricab:
I see you stacked this and dropped the repeated commit, great!

luis4a0:
Yes, this should work now.

ricab:
Hey @luis4a0, I am revisiting this for the LXD implementation :slightly_smiling_face: 

The first commit in this branch appears bogus &ndash; description doesn't seem to match the diff, which is later reverted. I have a rebase on master that drops that commit. If you agree, I can push?

luis4a0:
Go ahead!

ricab:
This is now outdated.

PR #1817, opened 2020-10-28
[dnsmasq_server] Increase wait timeout to 100ms

townsend2010:
There are still some occasional CI test failures due to the timeout being a little short.


ricab (approved):
Yep, thanks!

bors r+

PR #1820, opened 2020-10-28
Validate network names

luis4a0:
This PR includes code to validate the network names given as argument to `--networks`.

This address one of the first concerns in the review of https://github.com/canonical/multipass/pull/1745 and implements what is described in https://github.com/canonical/multipass/projects/12#card-47746768

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -24,6 +24,7 @@
 #include <multipass/exceptions/create_image_exception.h>
 #include <multipass/exceptions/exitless_sshprocess_exception.h>
 #include <multipass/exceptions/invalid_memory_size_exception.h>
+#include <multipass/exceptions/not_implemented_on_this_backend_exception.h>
 #include <multipass/exceptions/sshfs_missing_error.h>
 #include <multipass/exceptions/start_exception.h>
 #include <multipass/logging/client_logger.h>
@@ -55,6 +56,7 @@
 #include <QSysInfo>
 #include <QtConcurrent/QtConcurrent>
 
+#include <algorithm>
 #include <cassert>
 #include <functional>
 #include <stdexcept>
@@ -430,6 +432,23 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
         mp::LaunchError option_errors;
     } ret{mem_size, disk_space, instance_name, extra_interfaces, option_errors};
     return ret;
+}
+
+// Checks that all the interfaces in arg_ifaces are contained in platform_ifaces. If not, log and store the error.
+void validate_network_ids(std::vector<mp::NetworkInterface>& arg_ifaces,
+                          const std::vector<mp::NetworkInterfaceInfo>& platform_ifaces, mp::LaunchError* errors)
+{
+    for (mp::NetworkInterface& arg_iface : arg_ifaces)
+    {
+        auto pred = [arg_iface](const mp::NetworkInterfaceInfo& info) { return info.id == arg_iface.id; };
+        const auto& result = std::find_if(platform_ifaces.cbegin(), platform_ifaces.cend(), pred);
+
+        if (result == platform_ifaces.cend())
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("Invalid network id \"{}\"", arg_iface.id));
+            errors->add_error_codes(mp::LaunchError::INVALID_NETWORK);
+        }
+    }
 }
 
 auto grpc_status_for_mount_error(const std::string& instance_name)
@@ -2039,6 +2058,22 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
                            std::promise<grpc::Status>* status_promise, bool start)
 {
     auto checked_args = validate_create_arguments(request);
+
+    // Validate the names of the networks to connect to. Check here, because this function is inside the daemon,
+    // thus it can call functions from the factory.
+    if (!checked_args.extra_interfaces.empty())
+    {
+        try
+        {
+            validate_network_ids(checked_args.extra_interfaces, config->factory->list_networks(),
+                                 &checked_args.option_errors);
+        }
+        catch (const mp::NotImplementedOnThisBackendException&)
+        {
+            // If list-networks is not implemented, we should report that --network is not implemented on this backend.
+            throw mp::NotImplementedOnThisBackendException("--network");
+        }
+    }
 
     if (!checked_args.option_errors.error_codes().empty())
     {
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -394,7 +394,7 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
 
         if (result == factory_networks.cend())
         {
-            mpl::log(mpl::Level::debug, category, fmt::format("Invalid network id \"{}\"", net.id()));
+            mpl::log(mpl::Level::warning, category, fmt::format("Invalid network id \"{}\"", net.id()));
             option_errors.add_error_codes(mp::LaunchError::INVALID_NETWORK);
         }
 
@@ -405,7 +405,7 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
                 mp::NetworkInterface{net.id(), mac, net.mode() != multipass::LaunchRequest_NetworkOptions_Mode_MANUAL});
         else
         {
-            mpl::log(mpl::Level::debug, category, fmt::format("Repeated MAC address \"{}\"", mac));
+            mpl::log(mpl::Level::warning, category, fmt::format("Repeated MAC address \"{}\"", mac));
             option_errors.add_error_codes(mp::LaunchError::INVALID_NETWORK);
         }
     }
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -366,24 +366,40 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
 }
 
 std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequest* request,
+                                                            std::vector<mp::NetworkInterfaceInfo> factory_networks,
                                                             mp::LaunchError& option_errors)
 {
-    // The associated bool indicates whether the interface mode is 'auto', i.e., needs to be configured by cloud-init.
     std::vector<mp::NetworkInterface> interfaces;
 
     for (const auto& net : request->network_options())
     {
+        // Check that the id the user specified is valid.
+        auto pred = [net](const mp::NetworkInterfaceInfo& info) { return info.id == net.id(); };
+        const auto& result = std::find_if(factory_networks.cbegin(), factory_networks.cend(), pred);
+
+        if (result == factory_networks.cend())
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("Invalid network id \"{}\"", net.id()));
+            option_errors.add_error_codes(mp::LaunchError::INVALID_NETWORK);
+        }
+
+        // Check that the MAC the user specified is not repeated.
         if (const auto& mac = QString::fromStdString(net.mac_address()).toLower().toStdString();
             mac.empty() || mpu::valid_mac_address(mac))
             interfaces.push_back(
                 mp::NetworkInterface{net.id(), mac, net.mode() != multipass::LaunchRequest_NetworkOptions_Mode_MANUAL});
         else
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("Repeated MAC address \"{}\"", mac));
             option_errors.add_error_codes(mp::LaunchError::INVALID_NETWORK);
+        }
     }
+
     return interfaces;
 }
 
-auto validate_create_arguments(const mp::LaunchRequest* request)
+auto validate_create_arguments(const mp::LaunchRequest* request,
+                               const std::vector<mp::NetworkInterfaceInfo>& factory_networks)
 {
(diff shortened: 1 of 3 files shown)


src/daemon/daemon.cpp
```diff
@@ -2040,6 +2059,22 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
 {
     auto checked_args = validate_create_arguments(request);
 
+    // Validate the names of the networks to connect to. Check here, because this function is inside the daemon,
```
ricab, line 2062: There is a call to `validate_create_arguments` just above, so I don't think it's appropriate to `validate_network_ids` just after. If you need the factory to check the ids, you can pass it on as a parameter. 

`validate_create_arguments` in turn calls `validate_extra_interfaces`, which is where the check should be. That would keep it tidy and free another 15 LOC in this already big function. The factory would just need to be forwarded there. 

(Alternatively, you could make those helpers private methods, but I don't think that's necessary for this.)

tests/test_daemon.cpp
```diff
@@ -988,4 +992,38 @@ TEST_F(Daemon, reads_mac_addresses_from_json)
...
+{
+    mpt::MockVirtualMachineFactory* mock_factory = use_a_mock_vm_factory();
+
+    mp::Daemon daemon{config_builder.build()};
+
+    EXPECT_CALL(*mock_factory, list_networks());
+
+    std::stringstream err_stream;
+    send_command({"launch", "--network", "eth2"}, std::cout, err_stream);
+    EXPECT_THAT(err_stream.str(), HasSubstr("Invalid network options supplied"));
```
ricab, line 1016: Of course... thanks for reminding my sleepy self of this today, I had forgotten to suggest it in this other PR.

src/daemon/daemon.cpp
```diff
@@ -2040,6 +2059,22 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
+    {
+        try
+        {
+            validate_network_ids(checked_args.extra_interfaces, config->factory->list_networks(),
+                                 &checked_args.option_errors);
+        }
+        catch (const mp::NotImplementedOnThisBackendException&)
+        {
+            // If list-networks is not implemented, we should report that --network is not implemented on this backend.
+            throw mp::NotImplementedOnThisBackendException("--network");
```
ricab, line 2074: Indeed, good catch.

ricab (commented):
The strategy and code looks good, but I think a part should be relocated (to avoid yet another re-check of the `checked_args`).

src/daemon/daemon.cpp
luis4a0: Ouch, right. I could save a lot of spreading by just passing a parameter. I did pass a `factory_networks` parameter to `validate_create_arguments`, which is itself passed to `validate_extra_interfaces` and everything is checked in the latter function.

src/daemon/daemon.cpp
```diff
@@ -432,6 +434,23 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
     return ret;
 }
 
+// Checks that all the interfaces in arg_ifaces are contained in platform_ifaces. If not, log and store the error.
+void validate_network_ids(std::vector<mp::NetworkInterface>& arg_ifaces,
+                          const std::vector<mp::NetworkInterfaceInfo>& platform_ifaces, mp::LaunchError* errors)
```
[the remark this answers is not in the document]
luis4a0: Yeah, the name is a bit misleading. However, this function was removed due to a refactoring, so now no rename is needed.

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,40 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
...
 
     for (const auto& net : request->network_options())
     {
+        // Check that the id the user specified is valid.
+        auto pred = [net](const mp::NetworkInterfaceInfo& info) { return info.id == net.id(); };
+        const auto& result = std::find_if(factory_networks.cbegin(), factory_networks.cend(), pred);
+
+        if (result == factory_networks.cend())
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("Invalid network id \"{}\"", net.id()));
```
ricab, line 382: I would raise this to warning, so that the user gets it back with a single `-v`. Ideally they shouldn't need to increase verbosity to see the reason for the error, so I think it would even be ok to issue an error. The user should be getting this back without increasing verbosity. But the way it shows is not the cleanest:

```console
$ multipass launch --network asdf
[2020-12-07T13:07:03.321] [error] [daemon] Invalid network id "asdf"

launch failed: Invalid arguments supplied
Invalid network options supplied
```
This may feel strange without a `-v`.

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,40 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
...
+        }
+
+        // Check that the MAC the user specified is not repeated.
         if (const auto& mac = QString::fromStdString(net.mac_address()).toLower().toStdString();
             mac.empty() || mpu::valid_mac_address(mac))
             interfaces.push_back(
                 mp::NetworkInterface{net.id(), mac, net.mode() != multipass::LaunchRequest_NetworkOptions_Mode_MANUAL});
         else
+        {
+            mpl::log(mpl::Level::debug, category, fmt::format("Repeated MAC address \"{}\"", mac));
```
ricab, line 393: Same reasoning would apply.

ricab (commented):
A few remaining points below.

src/daemon/daemon.cpp
```diff
@@ -2038,7 +2056,22 @@ std::string mp::Daemon::check_instance_exists(const std::string& instance_name)
...
+        try
+        {
+            factory_networks = config->factory->list_networks();
+        }
+        catch (const mp::NotImplementedOnThisBackendException&)
+        {
+            // If list-networks is not implemented, we should report that --network is not implemented on this backend.
+            throw mp::NotImplementedOnThisBackendException("--network");
+        }
+    }
```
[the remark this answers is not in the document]
luis4a0: Right, fixed. I misunderstood your comment (entirely my fault). Looks better and simplifies `create_vm`, which is already a monster in size.

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,40 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
 }
 
 std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequest* request,
+                                                            std::vector<mp::NetworkInterfaceInfo> factory_networks,
```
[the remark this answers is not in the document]
luis4a0: Passed the factory as argument, so this is not anymore relevant.

src/daemon/daemon.cpp
luis4a0: In order to improve the error message, we can do two things:
1. throw a `runtime_error` (as we do later in `create_vm`);
2. improve logging in the RPC, by adding a message to `LaunchError`.
What do you think?

luis4a0: Wait, we can improve logging without changing the RPC, here: https://github.com/canonical/multipass/blob/validate-network-names/src/daemon/daemon.cpp#L2078

It's a matter of parsing the errors there and improve reporting. Does it make sense to you?

tests/stub_virtual_machine_factory.h
```diff
@@ -74,6 +74,11 @@ struct StubVirtualMachineFactory : public multipass::BaseVirtualMachineFactory
     {
         return std::make_unique<StubVMImageVault>();
     }
+
+    std::vector<multipass::NetworkInterfaceInfo> list_networks() const
+    {
+        throw multipass::NotImplementedOnThisBackendException("list-networks");
+    }
```
ricab, line 81: Why did you need this here? Isn't it the same implementation in the base class. I must be missing something, cause compilation should have failed on the missing `override` (because of -Werror).

tests/test_daemon.cpp
```diff
@@ -266,6 +266,10 @@ struct Daemon : public Test
 
         ON_CALL(*mock_factory_ptr, get_backend_version_string()).WillByDefault(Return("mock-1234"));
 
+        ON_CALL(*mock_factory, list_networks())
```
ricab, line 269: Tiny default, could you use the same as other calls in here? Not that dereferencing a `unique_ptr` has any significant impact, but just for consistency, so that the reader does not go wonder what is what.

ricab (commented):
Hey @luis4a0, this pretty much good to land, thanks.

I have one main question, I think I am missing something on the change to the stub factory. Otherwise just a couple of details if you're willing.

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,55 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
 }
 
 std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequest* request,
+                                                            const std::unique_ptr<mp::VirtualMachineFactory>& factory,
```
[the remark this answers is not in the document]
luis4a0: Right, makes sense. I feel better now, passing a reference instead of a pointer :)

tests/stub_virtual_machine_factory.h
luis4a0: Right, everything works the same way without redefining this function.

tests/test_daemon.cpp
luis4a0: Oops, I didn't note this. Maybe it was an auto-complete. Fixed.

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,54 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
...
+        auto pred = [net](const mp::NetworkInterfaceInfo& info) { return info.id == net.id(); };
+        const auto& result = std::find_if(factory_networks.cbegin(), factory_networks.cend(), pred);
+
+        if (result == factory_networks.cend())
+        {
+            mpl::log(mpl::Level::warning, category, fmt::format("Invalid network id \"{}\"", net.id()));
+            option_errors.add_error_codes(mp::LaunchError::INVALID_NETWORK);
+        }
+
+        // Check that the MAC the user specified is not repeated.
```
ricab, line 401: Humm, that's not what this is doing, that happens elsewhere. This is just validating syntax.

ricab (commented):
@luis4a0 I am afraid I detected a misplaced comment, otherwise all is good.

src/daemon/daemon.cpp
luis4a0: Oops, this comment must be a leftover of some previous implementation. Corrected.

ricab: Ah, because it was copy pasted from below? There's a case that tying the comment to the code would have prevented :wink:

src/daemon/daemon.cpp
```diff
@@ -364,24 +366,54 @@ auto try_mem_size(const std::string& val) -> mp::optional<mp::MemorySize>
...
+        }
+
+        // In case the user specified a MAC address, check it is valid.
         if (const auto& mac = QString::fromStdString(net.mac_address()).toLower().toStdString();
             mac.empty() || mpu::valid_mac_address(mac))
             interfaces.push_back(
                 mp::NetworkInterface{net.id(), mac, net.mode() != multipass::LaunchRequest_NetworkOptions_Mode_MANUAL});
         else
+        {
+            mpl::log(mpl::Level::warning, category, fmt::format("Repeated MAC address \"{}\"", mac));
```
ricab, line 408: Actually, this is also mistaken.
```suggestion
            mpl::log(mpl::Level::warning, category, fmt::format("Invalid MAC address \"{}\"", mac));
```

ricab (changes requested):
Hey @luis4a0, I am afraid the warning you added is also not right. The mac address is not repeated in that case, just invalid. It highlights the lack of a test for that... 

You may prefer to separate that bit into a new PR, this one doesn't need to be concerned with mac syntax.

src/daemon/daemon.cpp
luis4a0: You are indisputably right :S

src/daemon/daemon.cpp
luis4a0: Right, let me check all comments and warnings. This piece of code suffered lots of changes, so these can be outdated.

luis4a0: Done. I understand this PR is not about MAC checking, but the comment and warning we corrected were in the diff, so I consider OK to fix it here.

ricab: Yes, that makes sense. This PR involved changing the severity in a network check and the mac check can be regarded as a necessary piece for consistency. However, it is an untested piece... Care to add a test akin to `refuses_launch_with_invalid_network_interface` but checking mac validity (so `refuses_launch_with_invalid_mac_address` ? Thanks!

ricab: Hey Luis, I chanced upon daemon tests and realized that the missing test here would not be easy after all. I thought it would be the same as the tests you already added, only with a bad MAC replacing the bad network ID. But that's not the case.

It is true that the repeated MAC check derives from having 2 interfaces at different levels (CLI and RPC). Although both should normally be tested, we have currently no way to test the daemon directly via RPC without going through a (test) client. But that test client reuses most of the real client, and that includes the mac check. 

So we have no way to send a bad MAC to the daemon, and I hadn't realized that. It could be done, but it would require re-implementing a launch cmd or otherwise hooking to the RPC. That is an open problem and not something that should be done for this. So sorry about the trouble.

ricab (approved):
Alright, thank you for the PR, this is good to go!

bors merge

PR #1824, opened 2020-11-02
[lxd] Remove 10 GB minimum disk size

townsend2010:


ricab (commented):
This looks good, but unfortunately it didn't work for me:

```sh
$ multipass launch groovy
launch failed: LXD object not found
```

I reproduced the same thing with a direct build from source. The corresponding trace log: https://paste.ubuntu.com/p/Gwjc5CgGRn/

We agreed to leave this pending for now, until the issue is understood.

townsend2010:
Ok, after some discussion with @stgraber, the difference is that I'm using a dir-based storage pool and @ricab is using a zfs-based storage pool.  The zfs-based pool still has the 10GB minimum image size limit unless someone specifically changes it using something like:
`lxc storage set default volume.size 4GiB`

I think the proper way to handle this is to check which storage driver is being used.  If it's `dir`, then we use the logic presented in this PR (at the time I wrote this).  If it's any other storage driver type, then I get the config of the storage pool and if `volume.size` is set, use that.  Otherwise, use the original 10GB minimum.

ricab:
IIUC, there is a plan to change this in LXD so that the img size is what's enforced as minimum in all cases. What sort of timescale would that take? I wonder if it's worth having the special case in the meantime.

townsend2010:
Yes, they acknowledged that it's a bug on their end, but they gave no timeline of when it may be fixed.  The other problem is that they would have to unpack the whole image in this case, which takes time.  Not sure how much time, but we would have to weigh the user experience for such a thing as our default image size is less than the zfs pool default image size.

ricab:
I see, so you think we'd better keep the minimum to volume.size even when LXD can go below it? Is it noticeably slow(er) when using dir? Since the same unpacking happens in that case... 

I'm wondering what's worst in terms of UX: refusing to go below 10G or taking longer. I often want small instances and I might be willing to pay a small time penalty for that. But I agree that a silent penalty by default would be bad if it is more than a couple seconds.

townsend2010:
I'm basing my assertion on what @stgraber said in chat:
> ...but size requests below volume.size will likely be quite a bit slower due to us having to unpack the image from scratch just for that one instance.

I really have no way to quantify that with zfs since it doesn't work today.  For dir-based, it *always* unpacks the image, so I can't compare that either.

stgraber:
That'd make ZFS about as slow as dir, possibly even slower.
Here creating a new instance from an existing image on ZFS takes less than a second vs over a minute when creating on dir.

townsend2010:
@stgraber, thanks for the additional info. Should I enter a LXD bug about the 10GB minimum on zfs pools?

I think what we should take what I proposed a step further once the LXD bug for zfs is fixed.  If `volume.size` is missing and the user did not provide a disk size, then we default to 10GB in order to have the fast instance creation.  If the user does provide a disk size < 10GB, then we confirm it's not too small and then allow it and if user's complain about long instance creation times, we can explain what the issue is.

stgraber:
Yeah, going with a minimum of `10GB` unless the user has set `volume.size` to a lower value seems quite reasonable.
As for the bug, I'm pretty sure @tomponline is tracking it and will take very soon but feel free to file an issue on our side, the reproducer I posted on Mattermost is pretty straightforward.

ricab:
Hi @townsend2010, I was mainly after your feeling from the dir-based experience (like how long you had to wait), didn't know if you'd compared to zfs. But @stgraber clarified, thanks! Yeah, +1 on avoiding one more minute by default.

So, I think what you propose makes sense, but could you please confirm my understanding: in the long run, like the minimum image size, the default image size would become a function of the backend, right? Basically `max(img.size, volume.size)` in LXD with non-dir storages, correct? 

Maybe our lxd-specific code could then issue a warning when `min <= requested_size < default`. So that the user saw  (with `-v`) something like "use `--disk` larger than `<default>` for a faster experience".

tomponline:
@stgraber yep tracking the fix at https://github.com/lxc/lxd/issues/8114

townsend2010:
Hey @ricab,

> in the long run, like the minimum image size, the default image size would become a function of the backend, right?

I think that is the right thing to do since this is now dependent on the backend and would come from the `VMImageVault` implementation.

> Basically max(img.size, volume.size, default_for_other_backends_aka_5G) in LXD with non-dir storages, correct?

Did you mean non-**zfs** storage pools here?  zfs will be the only one we treat differently when all defaults are being used.

> Maybe our lxd-specific code could then issue a warning when `min <= requested_size < default`.

Yes, that is a good idea.  I also saw your bug about the help for disk size, so I'll account for that as well.

ricab:
OK sounds like a plan :+1: 

> Did you mean non-zfs storage pools here? zfs will be the only one we treat differently when all defaults are being used.

I singled out directory storage because I understood that it was the only one that didn't have a volume.size, but I really don't know. Reading [here](https://linuxcontainers.org/lxd/docs/master/storage) it seems only "Currently valid for loop based pools and zfs". So I suppose that would exclude btrfs and lvm. I also don't know if the perf impact of going below volume.size would be the same in all storage types supporting that field.

townsend2010:
I'm going to close this PR since the proper way to handle this will be quite different.

PR #1834, opened 2020-11-09
[lxd] increase first boot timeout to 20s

Saviq:
This impacts only the "Instance shutdown during start" situation on LXD. I verified 5s is not enough on a Pi4, and the worst case is that this will delay `launch` failing when someone did indeed stop the instance being started.


townsend2010 (approved):
Sure, makes sense on slower machines.  +1

bors merge

PR #1835, opened 2020-11-10
[lxd] Handle LXD errors better

townsend2010:
This will bubble up LXD errors and present them to the client and log them.

### DIFF
--- a/src/platform/backends/lxd/lxd_request.cpp
+++ b/src/platform/backends/lxd/lxd_request.cpp
@@ -18,7 +18,6 @@
 #include "lxd_request.h"
 
 #include <multipass/format.h>
-#include <multipass/logging/log.h>
 #include <multipass/network_access_manager.h>
 #include <multipass/version.h>
 
@@ -32,15 +31,6 @@ namespace mpl = multipass::logging;
 
 namespace
 {
-constexpr auto request_category = "lxd request";
-
-void log_and_throw_error(const std::string& message, mpl::Level level)
-{
-    mpl::log(level, request_category, message);
-
-    throw std::runtime_error(message);
-}
-
 template <typename Callable>
 const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int timeout, Callable&& handle_request)
 {
@@ -63,7 +53,7 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
         url.setQuery(project_query_string);
     }
 
-    mpl::log(mpl::Level::trace, request_category, fmt::format("Requesting LXD: {} {}", method, url.toString()));
+    mpl::log(mpl::Level::trace, mp::request_category, fmt::format("Requesting LXD: {} {}", method, url.toString()));
     QNetworkRequest request{url};
 
     request.setHeader(QNetworkRequest::UserAgentHeader, QString("Multipass/%1").arg(mp::version_string));
@@ -88,34 +78,35 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
         throw mp::LXDNotFoundException();
 
     if (reply->error() == QNetworkReply::OperationCanceledError)
-        log_and_throw_error(fmt::format("Timeout getting response for {} operation on {}", method, url.toString()),
-                            mpl::Level::warning);
+        throw mp::LXDRuntimeError(
+            fmt::format("Timeout getting response for {} operation on {}", method, url.toString()),
+            mpl::Level::warning);
 
     auto bytearray_reply = reply->readAll();
     reply->deleteLater();
 
     if (bytearray_reply.isEmpty())
-        log_and_throw_error(fmt::format("Empty reply received for {} operation on {}", method, url.toString()),
-                            mpl::Level::error);
+        throw mp::LXDRuntimeError(fmt::format("Empty reply received for {} operation on {}", method, url.toString()),
+                                  mpl::Level::error);
 
     QJsonParseError json_error;
     auto json_reply = QJsonDocument::fromJson(bytearray_reply, &json_error);
 
     if (json_error.error != QJsonParseError::NoError)
-        log_and_throw_error(fmt::format("Error parsing JSON response for {}: {}\n{}", url.toString(),
-                                        json_error.errorString(), bytearray_reply),
-                            mpl::Level::error);
+        throw mp::LXDRuntimeError(fmt::format("Error parsing JSON response for {}: {}\n{}", url.toString(),
+                                              json_error.errorString(), bytearray_reply),
+                                  mpl::Level::error);
 
     if (json_reply.isNull() || !json_reply.isObject())
-        log_and_throw_error(fmt::format("Invalid LXD response for {}: {}", url.toString(), bytearray_reply),
-                            mpl::Level::error);
+        throw mp::LXDRuntimeError(fmt::format("Invalid LXD response for {}: {}", url.toString(), bytearray_reply),
+                                  mpl::Level::error);
 
-    mpl::log(mpl::Level::trace, request_category, fmt::format("Got reply: {}", QJsonDocument(json_reply).toJson()));
+    mpl::log(mpl::Level::trace, mp::request_category, fmt::format("Got reply: {}", QJsonDocument(json_reply).toJson()));
 
     if (reply->error() != QNetworkReply::NoError)
-        log_and_throw_error(fmt::format("Network error for {}: {} - {}", url.toString(), reply->errorString(),
-                                        json_reply.object()["error"].toString()),
-                            mpl::Level::error);
+        throw mp::LXDRuntimeError(fmt::format("Network error for {}: {} - {}", url.toString(), reply->errorString(),
+                                              json_reply.object()["error"].toString()),
+                                  mpl::Level::error);
 
     return json_reply.object();
 }
@@ -133,7 +124,7 @@ const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::
             request.setHeader(QNetworkRequest::ContentTypeHeader, "application/x-www-form-urlencoded");
             request.setHeader(QNetworkRequest::ContentLengthHeader, QByteArray::number(data.size()));
 
-            mpl::log(mpl::Level::trace, request_category, fmt::format("Sending data: {}", data));
+            mpl::log(mpl::Level::trace, mp::request_category, fmt::format("Sending data: {}", data));
         }
 
         return manager->sendCustomRequest(request, verb, data);
@@ -170,17 +161,17 @@ const QJsonObject mp::lxd_wait(mp::NetworkAccessManager* manager, const QUrl& ba
 
         if (task_reply["error_code"].toInt() >= 400)
         {
-            log_and_throw_error(fmt::format("Error waiting on operation: {}", task_reply["error"].toString()),
-                                mpl::Level::error);
+            throw mp::LXDRuntimeError(fmt::format("Error waiting on operation: {}", task_reply["error"].toString()),
+                                      mpl::Level::error);
         }
         else if (task_reply["status_code"].toInt() == 400)
         {
-            log_and_throw_error(fmt::format("Failure waiting on operation: {}", task_reply["status"].toString()),
-                                mpl::Level::error);
+            throw mp::LXDRuntimeError(fmt::format("Failure waiting on operation: {}", task_reply["status"].toString()),
+                                      mpl::Level::error);
         }
         else if (task_reply["metadata"].toObject()["status_code"].toInt() == 400)
         {
-            log_and_throw_error(
+            throw mp::LXDRuntimeError(
                 fmt::format("Operation completed with error: {}", task_reply["metadata"].toObject()["err"].toString()),
                 mpl::Level::error);
         }
--- a/src/platform/backends/lxd/lxd_request.cpp
+++ b/src/platform/backends/lxd/lxd_request.cpp
@@ -180,9 +180,9 @@ const QJsonObject mp::lxd_wait(mp::NetworkAccessManager* manager, const QUrl& ba
         }
         else if (task_reply["metadata"].toObject()["status_code"].toInt() == 400)
         {
-            log_and_throw_error(fmt::format("Operation completed with error: {}",
-                                            task_reply["metadata"].toObject()["err"].toString()),
-                                mpl::Level::error);
+            log_and_throw_error(
+                fmt::format("Operation completed with error: {}", task_reply["metadata"].toObject()["err"].toString()),
+                mpl::Level::error);
         }
     }
 
--- a/src/platform/backends/lxd/lxd_request.cpp
+++ b/src/platform/backends/lxd/lxd_request.cpp
@@ -92,6 +92,9 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
     auto bytearray_reply = reply->readAll();
     reply->deleteLater();
 
+    if (bytearray_reply.isEmpty())
+        throw std::runtime_error(fmt::format("Empty reply received for {} operation on {}", method, url.toString()));
+
     QJsonParseError json_error;
     auto json_reply = QJsonDocument::fromJson(bytearray_reply, &json_error);
 
--- a/src/platform/backends/lxd/lxd_request.cpp
+++ b/src/platform/backends/lxd/lxd_request.cpp
@@ -82,6 +82,10 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
     if (reply->error() == QNetworkReply::ContentNotFoundError)
         throw mp::LXDNotFoundException();
 
+    if (reply->error() == QNetworkReply::OperationCanceledError)
+        throw std::runtime_error(
+            fmt::format("Timeout getting response for {} operation on {}", method, url.toString()));
+
     if (reply->error() != QNetworkReply::NoError)
         throw std::runtime_error(fmt::format("{}: {}", url.toString(), reply->errorString()));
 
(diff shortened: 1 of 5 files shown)


src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -34,6 +34,13 @@ namespace
 {
 constexpr auto request_category = "lxd request";
 
+void log_and_throw_error(const std::string& message, mpl::Level level)
+{
+    mpl::log(level, request_category, message);
+
+    throw std::runtime_error(message);
+}
```
Saviq, line 42: This feels weird to be here in `lxd_request.cpp` only. Shouldn't this be a general thing?

Or really, shouldn't whatever catches the `runtime_error` log it, if it doesn't bubble it up? I know you'd lose control over the logging level, but if needed, that could be done with different exception types?

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -135,3 +153,38 @@ const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::
...
+                                mpl::Level::error);
+        }
+        else if (task_reply["status_code"].toInt() == 400)
+        {
+            log_and_throw_error(fmt::format("Failure waiting on operation: {}", task_reply["status"].toString()),
+                                mpl::Level::error);
+        }
+        else if (task_reply["metadata"].toObject()["status_code"].toInt() == 400)
+        {
+            log_and_throw_error(fmt::format("Operation completed but with error: {}",
```
Saviq, line 183:
```suggestion
            log_and_throw_error(fmt::format("Operation completed with error: {}",
```

tests/lxd/test_lxd_backend.cpp
```diff
@@ -874,13 +875,39 @@ TEST_F(LXDBackend, lxd_request_timeout_aborts_and_throws)
...
-                         std::runtime_error,
-                         Property(&std::runtime_error::what, AllOf(HasSubstr(base_url.toString().toStdString()),
-                                                                   HasSubstr("Operation canceled"))));
+
+    const std::string op{"GET"};
+    const std::string error_string{
+        fmt::format("Timeout getting response for {} operation on {}", op, base_url.toString().toStdString())};
+
+    MP_EXPECT_THROW_THAT(mp::lxd_request(mock_network_access_manager.get(), op, base_url, mp::nullopt, 3),
+                         std::runtime_error, Property(&std::runtime_error::what, AllOf(HasSubstr(error_string))));
```
Saviq, line 884:
```suggestion
                         std::runtime_error, Property(&std::runtime_error::what, HasSubstr(error_string)));
```

tests/lxd/test_lxd_backend.cpp
```diff
@@ -909,11 +942,220 @@ TEST_F(LXDBackend, lxd_request_wrong_json_throws)
...
+
+TEST_F(LXDBackend, lxd_wait_error_returned_throws_and_logs)
+{
+    ON_CALL(*mock_network_access_manager.get(), createRequest(_, _, _)).WillByDefault([](auto, auto request, auto) {
+        auto op = request.attribute(QNetworkRequest::CustomVerbAttribute).toString();
+        auto url = request.url().toString();
+
+        if (op == "GET")
+        {
+            if (url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```
Saviq, line 990:
```suggestion
        if (op == "GET" && url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```

tests/lxd/test_lxd_backend.cpp
```diff
@@ -909,11 +942,220 @@ TEST_F(LXDBackend, lxd_request_wrong_json_throws)
...
+
+TEST_F(LXDBackend, lxd_wait_status_code_failure_returned_throws_and_logs)
+{
+    ON_CALL(*mock_network_access_manager.get(), createRequest(_, _, _)).WillByDefault([](auto, auto request, auto) {
+        auto op = request.attribute(QNetworkRequest::CustomVerbAttribute).toString();
+        auto url = request.url().toString();
+
+        if (op == "GET")
+        {
+            if (url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```
Saviq, line 1049:
```suggestion
        if (op == "GET" && url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```

tests/lxd/test_lxd_backend.cpp
```diff
@@ -909,11 +942,220 @@ TEST_F(LXDBackend, lxd_request_wrong_json_throws)
...
+
+TEST_F(LXDBackend, lxd_wait_metadata_status_code_failure_returned_throws_and_logs)
+{
+    ON_CALL(*mock_network_access_manager.get(), createRequest(_, _, _)).WillByDefault([](auto, auto request, auto) {
+        auto op = request.attribute(QNetworkRequest::CustomVerbAttribute).toString();
+        auto url = request.url().toString();
+
+        if (op == "GET")
+        {
+            if (url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```
Saviq, line 1108:
```suggestion
        if (op == "GET" && url.contains("1.0/operations/b043d632-5c48-44b3-983c-a25660d61164/wait"))
```

Saviq (changes requested):

Saviq:
Got some way to show this in action, btw?

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Hmm, yeah, I was trying to combine redundant code, but yeah, after reading your feedback, it should up to the caller to log anything as it sees fit.  The callee here should just throw the error.

townsend2010:
> Got some way to show this in action, btw?

Good question.  The way I was able to test *some* of it was to pass in some ridiculously small timeout into `lxd_request()` wherever it may be called in `lxd_virtual_machine.cpp`.  But other than that, there really is no "easy" way to show in action.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Ok, I've been working on this and all I end up doing is instead of having one place that logs a message for our purposes, I'm spreading the same thing over and over to different parts of the code, making it redundant.  If you don't like `log_and_throw_error()` here, I can move it to `utils.cpp`, but I think this function commonizes something we need.

townsend2010: Ok, I removed this function and made a generic function in `utils.h`.  If we want to log file and line numbers and not get the function in utils, then the message that is passed in should contain the file and line and that is what will be logged.

townsend2010:
Hey @Saviq!  Any more feedback on this?

src/platform/backends/lxd/lxd_request.cpp
townsend2010: I removed this function and also got rid of the templated function introduced in the previous now defunct commit.  We also discussed using a large `try`/`catch` encapsulating `lxd_request()` and `lxd_wait()` with the `catch` logging the message and then rethrowing.  However, I found that if we want logging level control, a new exception type will need to be introduced to accept the logging level.  When I did this, it became clear that the exception itself should just do the logging which is what I have now.

ricab: Admittedly I didn't follow this discussion or read the code closely, so apologies if something here is not applicable :slightly_smiling_face: but I am intrigued. Normally it should be up to some caller up in the stack to decide how to handle an exception, including whether it can be recovered from, whether it should be logged, with what severity, etc. 

Exceptions are a way for the code where the problem occurs to abort what it's doing but delegate handling to callers, providing any relevant specifics in the exception itself, but letting "someone above", where all the context is available, judge how to react. As far as the thrower knows, proper handling could be switching backends, trying a different server, showing a GUI popup, reporting to a remote entity, retrying... 

Assuming an exception is an error at the throw site disables code reuse and approaches of "asking for forgiveness rather than permission". So logging in the exception constructor sounds strange to me, especially if it's logged with more severity then `debug`. [Jon Kalb](http://www.exceptionsafecode.com/) is a good source of material about this, his talks on error handling are elucidative.

townsend2010: Hey @ricab,

Thanks for the feedback!  It is definitely still applicable:smile:

So here's the problem I'm *trying* to solve.  In `lxd_request_common()` and `lxd_wait()`, there are a number of places where an "issue" can occur in the request of an API endpoint in the LXD daemon as can be seen in all the places that `throw`.  Right now, all but one "issue" I want to log at the `error` severity, so I deemed it necessary for the code to need to set the severity level for logging when throwing the exception.  Of course, there are a few ways to handle this.  My approach is to try to keep code redundancy to a minimum which is why I've gone the way of using common `log_and_throw`-type functions or the `LXDRuntimeError` which just encapsulates the loggig in itself.

That said, since this awkward, I could instead keep the `LXDRuntimeError` exception type, and any caller of functions that throw this type can just assume it's an error and choose to log.  Really, there is only one place where I set `warning` instead of `error` and I could argue the `warning` should be an `error` as well.

@saviq suggested a `try`/`catch`/(re)`throw` around the functions much like we do with the daemon command functions like `launch`.  I think I might try that approach coupled with just making the callers log any caught `LXDRuntimeError`'s at the `error` level.

ricab: OK, I think I understand your point. Looking from afar, that seems like a sensible option (single `error` severity from a handler up in the stack). If that meant giving up too much, you could introduce separate exception types. But if you find that the simpler code is worth giving up the distinction, better still. 

Anyway, I just wanted to provide some feedback, but I'm leaving this to you guys.

townsend2010: Ok, after some more thinking and discussion with @ricab, I've decided to just use the `try`/`catch`/(re)`throw` methodology to log messages.

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -82,29 +80,39 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
...
     if (json_error.error != QJsonParseError::NoError)
-        throw std::runtime_error(fmt::format("{}: {}", url.toString(), json_error.errorString()));
+        throw mp::LXDRuntimeError(fmt::format("Error parsing JSON response for {}: {}\n{}", url.toString(),
+                                              json_error.errorString(), bytearray_reply));
 
     if (json_reply.isNull() || !json_reply.isObject())
-        throw std::runtime_error(fmt::format("Invalid LXD response for url {}: {}", url.toString(), bytearray_reply));
+        throw mp::LXDRuntimeError(fmt::format("Invalid LXD response for {}: {}", url.toString(), bytearray_reply));
 
     mpl::log(mpl::Level::trace, request_category, fmt::format("Got reply: {}", QJsonDocument(json_reply).toJson()));
```
Saviq, line 103: We won't see this if JSON fails to parse. Should we move this up (and skip the parse + dump loop)?

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -82,29 +80,39 @@ const QJsonObject lxd_request_common(const std::string& method, QUrl& url, int t
...
+                                              json_reply.object()["error"].toString()));
+
     return json_reply.object();
 }
 } // namespace
 
 const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::string& method, QUrl url,
-                                  const mp::optional<QJsonObject>& json_data, int timeout)
+                                  const mp::optional<QJsonObject>& json_data, int timeout) // clang-format off
+try // clang-format on
```
Saviq, line 115: Why these? Do we need to move to newer clang to avoid?

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -123,9 +131,16 @@ const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::
...
+{
+    mpl::log(mpl::Level::error, request_category, e.what());
+
+    throw;
+}
 
 const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::string& method, QUrl url,
-                                  QHttpMultiPart& multi_part, int timeout)
+                                  QHttpMultiPart& multi_part, int timeout) // clang-format off
+try // clang-format on
```
Saviq, line 143: ditto

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -135,3 +150,48 @@ const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::
...
+catch (const LXDRuntimeError& e)
+{
+    mpl::log(mpl::Level::error, request_category, e.what());
+
+    throw;
+}
+
+const QJsonObject mp::lxd_wait(mp::NetworkAccessManager* manager, const QUrl& base_url, const QJsonObject& task_data,
+                               int timeout) // clang-format off
+try // clang-format on
```
Saviq, line 162: ditto

src/platform/backends/lxd/lxd_request.cpp
```diff
@@ -135,3 +150,48 @@ const QJsonObject mp::lxd_request(mp::NetworkAccessManager* manager, const std::
...
+        }
+        else if (task_reply["status_code"].toInt() == 400)
+        {
+            throw mp::LXDRuntimeError(fmt::format("Failure waiting on operation: {}", task_reply["status"].toString()));
+        }
+        else if (task_reply["metadata"].toObject()["status_code"].toInt() == 400)
+        {
+            throw mp::LXDRuntimeError(
+                fmt::format("Operation completed with error: {}", task_reply["metadata"].toObject()["err"].toString()));
+        }
```
Saviq, line 187: Should all these not be `!= 200`? Like, It's an error any time the returned status is not `200`? I fail to see documentation on specific status codes returned by `/wait`…

Saviq (changes requested):

src/platform/backends/lxd/lxd_request.cpp
townsend2010: I did this because it's used elsewhere in many parts of the code, so I thought it was necessary.  I can try it and if it works, then great and then we can remove it elsewhere in a separate PR.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Ugh...so LXD has like three different ways to report errors and status. 

So if there is an HTTP error, then `task_reply["error_code"]` is going to be set and per the documentation will "be one of of 400, 401, 403, 404, 409, 412 or 500."

As for the API call itself, there is also a status code which is obtained via `task_reply["status_code"]`.  According to the docs, a failure is **only** indicated by code 400.

Then there is the associated operation of the API call which is obtained via `task_reply["metadata"].toObject()["status_code"]` and again, a failure is only indicated by code 400.

I got these values via https://lxd.readthedocs.io/en/latest/rest-api/#return-values and https://lxd.readthedocs.io/en/latest/rest-api/#status-codes.

Also, no we don't want it to be `!= 200`.  For example, if an operation is running, it will have a status code of `103` and we don't want to fail on that.  To expand that even further, we don't want to fail on any status code in the 100's or 200.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Um, well, if the JSON fails to parse, we won't get anything from `QJsonDocument(json_reply).toJson()` anyways, so I don't see why we need to move this up....unless I misunderstand what you mean here.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Marking this as resolved since I'm using the `try`/`catch` method.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Ok, I removed it and `clang-format` seems happy.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Removed.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: Removed

src/platform/backends/lxd/lxd_request.cpp
Saviq: I just worry LXD might introduce a new error/status code and we'll miss it. Should everywhere be `>= 400`, then, as LXD tightly follows HTTP status codes?

src/platform/backends/lxd/lxd_request.cpp
Saviq: I mean that when we fail to parse the JSON further up, we won't log the contents of the HTTP reply. Which, when it fails to parse, will likely be interesting.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: > I just worry LXD might introduce a new error/status code and we'll miss it.

Well, I think we're going to have to trust the docs here.  If they deviate from what that says, then they have broken the API and if they do things properly, we should be made aware of that.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: If it fails to parse the JSON further up, then `QJsonDocument(json_reply).toJson()` here isn't going to work, thus is won't log what we expect.

src/platform/backends/lxd/lxd_request.cpp
townsend2010: As discussed, I now have `>= 400` for the two conditions and now also log the error/status code.

Saviq (approved):
Yup, thanks!

PR #1836, opened 2020-11-12
Control platform update

ricab:
Proposed updates for the "control" branch, public side.

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -72,8 +72,7 @@ bool is_image_url_supported();
 
 std::function<int()> make_quit_watchdog(); // call while single-threaded; call result later, in dedicated thread
 
-// Get information of one given network interface, or about all of them.
-NetworkInterfaceInfo get_network_interface_info(const std::string& iface_name);
+// Get information on the network interfaces that are seen by the platform, indexed by name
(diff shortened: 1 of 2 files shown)


include/multipass/platform.h
```diff
@@ -72,9 +72,9 @@ bool is_image_url_supported();
 
 std::function<int()> make_quit_watchdog(); // call while single-threaded; call result later, in dedicated thread
 
-// Get information of one given network interface, or about all of them.
-NetworkInterfaceInfo get_network_interface_info(const std::string& iface_name);
+// Get information on the network interfaces that are seen by the platform, indexed by name
 std::map<std::string, NetworkInterfaceInfo> get_network_interfaces_info();
+std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
```
luis4a0, line 77: Excellent comment, "give the chance" is what it does, and it's much better than explaining all the situation.

luis4a0 (approved):

include/multipass/platform.h
ricab: Thanks, I couldn't find a better name in the end, so felt it needed a comment.

PR #1843, opened 2020-11-17
[lxd] Handle state requests better

townsend2010:


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -209,12 +211,19 @@ void mp::LXDVirtualMachine::stop()
...
     }
 
-    request_state("stop");
+    if (present_state == State::starting)
+    {
+        stop_options.insert("force", true);
+        forced_shutdown = true;
+    }
+
+    request_state("stop", stop_options);
```
Saviq, line 220: Add a TODO relating to #1909, please?

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -344,9 +353,14 @@ const QUrl mp::LXDVirtualMachine::network_leases_url()
...
-void mp::LXDVirtualMachine::request_state(const QString& new_state)
+void mp::LXDVirtualMachine::request_state(const QString& new_state, const QJsonObject& state_options)
 {
-    const QJsonObject state_json{{"action", new_state}};
+    QJsonObject state_json{{"action", new_state}};
+
+    for (auto option = state_options.constBegin(); option != state_options.constEnd(); ++option)
+    {
+        state_json.insert(option.key(), option.value());
+    }
```
Saviq, line 363: QJsonObject feels like it's playing against you here, wouldn't a simpler container be easier to deal with?

Saviq (changes requested):

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Hmm, this particular issue seems mostly orthogonal to #1909.  The problem I'm trying to fix here is that LXD will ignore a `stop` on a VM if it's not completely booted, ie, in the `starting` state unless a `force` option is used.

It looks to me that #1909 is trying to solve the issue of being able to shut down a "stuck" VM, regardless of backend used.

That said, fixing #1909 shouldn't have any bearing on this change here and so we shouldn't add that TODO.

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Well, yeah, except I'm having to send a `QJsonObject` to LXD here and found that just using a `QJsonObject` from the beginning was the simplest way to go.  Otherwise, we'd still need to convert from simpler container to `QJsonObject` here via some iterator.  The sad thing is that `QJsonObject` just doesn't have a good way to concatenate 2 `QJsonObject`s :frowning:

townsend2010: Just after posting the above, I thought of a way to simplify this some.  I can do something like:
```
QJsonObject state_json{state_options};

state_json.insert("action", new_state);
```

That should avoid the iterator and still accomplish the same thing.

src/platform/backends/lxd/lxd_virtual_machine.cpp
Saviq: Well, yeah, but when we introduce a `force_stop()` or so on the backend interface, this could call that instead, rather than reimplementing it?

townsend2010: Ah, ok, I see your point there.  Ok, I'm convinced :wink:

townsend2010: Done

src/platform/backends/lxd/lxd_virtual_machine.cpp
townsend2010: Ok, I changed to the way I proposed above.

Saviq (approved):
Yup. Thanks!

townsend2010:
After some discussion, the better way to handle this is to fail `shutdown` and `delete` operations if an instance is in the `starting` state and this goes for all backends.  Then, in the future, the proper way to handle this is the queue the `delete` or `shutdown` in the daemon and once the instance goes into the `running` or `unknown` state, then the operation will run.  The client by default will just return if the daemon accepts the operation and we'll add a `--wait` (or whatever we name it) in case a user wants to really wait on the operation to complete.

That said, closing this PR.

PR #1846, opened 2020-11-19
Work on Qt 5.15 by default

ricab:
Allow deprecation warnings (don't make them errors) on Linux and macOS (they were already allowed on Windows). This is meant to avoid multiplying preprocessor directives with code depending on Qt version (which is not currently exercised in CI or in our local build environments).

This takes advantage of backward compatibility within the same major Qt version (5), so that upcoming code builds normally. Changes will only be needed for Qt 6 and they are very simple. Ifdefs will only be needed then if we decide to simultaneously support versions >= 6 and < 5.15.


CMakeLists.txt
```diff
@@ -148,18 +153,35 @@ message(STATUS "Setting version to: ${MULTIPASS_VERSION}")
...
+      add_compile_options(-Wno-expansion-to-defined)
+    endif()
+
+    add_definitions(-DMULTIPASS_PLATFORM_LINUX)
+  endif()
 
-CHECK_CXX_COMPILER_FLAG("-Wno-expansion-to-defined" COMPILER_SUPPORTS_NO_EXP_TO_DEFINED)
-if(COMPILER_SUPPORTS_NO_EXP_TO_DEFINED)
-  add_compile_options(-Wno-expansion-to-defined)
+  add_compile_options(-Wno-error=deprecated-declarations)
```
Saviq, line 183: Could this, and the `add_` further up, be combined, or do they need to come in order?

ricab: I wasn't sure. I see this in the man:

       You can mix options and other arguments.  For the most part, the order you use doesn't
       matter.  Order does matter when you use several options of the same kind; for example, if
       you specify -L more than once, the directories are searched in the order specified.  Also,
       the placement of the -l option is significant.

Let me have a better look.

ricab: You're right, `-Wno-error=deprecated-declarations` can even appear before `-Werror`. Fix coming.

ricab: Done, amended the last commit.

Saviq:
LGTM

bors r+

Issue #938, opened 2019-08-01
Check disk space before creating instances


Saviq:
We should check whether the instance we'll be trying to launch will fit in the available disk space.

katie-knister:
Hi @Saviq! I'm interested in helping out with this. This is my first time contributing to an open source project. Do you have any tips on starting this task and/or notes on which parts of the code are relevant?

Saviq:
Hi @katie-knister, welcome on the FOSS train, you're in for a ride :)

This is where the disk size gets validated:

https://github.com/canonical/multipass/blob/4f7ae5cac81ff379f5452a34a20a11e434d2089f/src/daemon/daemon.cpp#L317L331

Should be easy to check whether there's enough space, too.

katie-knister:
Thank you so much for your quick and helpful response! I'll be working on this task in the next few days. Excited to get started!

katie-knister:
Hi @Saviq , I created PR #1854 to address this issue. It uses the std::filesystem library to check for available space in the user’s directory. Do you think you could take a look and see if it’s on the right track? I am also wondering why the build check fails with including filesystem since it successfully builds locally. I’m also interested in what recommendations you have for testing this. Thank you!

ricab:
Fixed by #1854.

PR #1789, opened 2020-10-07
[daemon] add MULTIPASS_STORAGE support (Fixes #1215)

Saviq:

PR #1854, opened 2020-11-25
Add check for available disk space on launch

katie-knister:
To address #938 I used the std::filesystem library to check for available space in the user’s directory.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -51,11 +51,12 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QJsonParseError>
+#include <QStorageInfo>
+#include <QString>
 #include <QSysInfo>
 #include <QtConcurrent/QtConcurrent>
 
 #include <cassert>
-#include <filesystem>
 #include <functional>
 #include <stdexcept>
 #include <utility>
@@ -321,10 +322,7 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
     if (!disk_space_str.empty())
     {
         auto opt_disk_space = try_mem_size(disk_space_str);
-        auto available_disk_space_uint = std::filesystem::space(std::filesystem::current_path()).available;
-        auto available_disk_space = try_mem_size(std::to_string(available_disk_space_uint));
-        if (opt_disk_space && *opt_disk_space >= min_disk && available_disk_space &&
-            *opt_disk_space <= *available_disk_space)
+        if (opt_disk_space && *opt_disk_space >= min_disk)
         {
             disk_space = opt_disk_space;
         }
@@ -627,7 +625,16 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
     {
         disk_space = *command_line_value;
     }
-
+    
+    auto available_bytes = QStorageInfo::root().bytesAvailable();
+    std::string available_bytes_str = QString::number(available_bytes).toStdString();
+    auto available_disk_space = mp::MemorySize(available_bytes_str + "B");
+    
+    if (available_disk_space < disk_space)
+    {
+    	throw std::runtime_error(fmt::format("Available disk ({} bytes) below requested/default size ({} bytes)", available_disk_space.in_bytes(), disk_space.in_bytes()));
+    }
+    
     return disk_space;
 }
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -625,16 +625,17 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
     {
         disk_space = *command_line_value;
     }
-    
+
     auto available_bytes = QStorageInfo::root().bytesAvailable();
     std::string available_bytes_str = QString::number(available_bytes).toStdString();
     auto available_disk_space = mp::MemorySize(available_bytes_str + "B");
-    
+
     if (available_disk_space < disk_space)
     {
-    	throw std::runtime_error(fmt::format("Available disk ({} bytes) below requested/default size ({} bytes)", available_disk_space.in_bytes(), disk_space.in_bytes()));
+        throw std::runtime_error(fmt::format("Available disk ({} bytes) below requested/default size ({} bytes)",
+                                             available_disk_space.in_bytes(), disk_space.in_bytes()));
     }
-    
+
     return disk_space;
 }
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -55,6 +55,7 @@
 #include <QtConcurrent/QtConcurrent>
 
 #include <cassert>
+#include <filesystem>
 #include <functional>
 #include <stdexcept>
 #include <utility>
@@ -320,7 +321,9 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
     if (!disk_space_str.empty())
     {
         auto opt_disk_space = try_mem_size(disk_space_str);
-        if (opt_disk_space && *opt_disk_space >= min_disk)
+        auto available_disk_space_uint = std::filesystem::space(std::filesystem::current_path()).available;
+	auto available_disk_space = try_mem_size(std::to_string(available_disk_space_uint));
+	if (opt_disk_space && *opt_disk_space >= min_disk && available_disk_space && *opt_disk_space <=  *available_disk_space)
         {
             disk_space = opt_disk_space;
         }
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -322,8 +322,9 @@ auto validate_create_arguments(const mp::LaunchRequest* request)
     {
         auto opt_disk_space = try_mem_size(disk_space_str);
         auto available_disk_space_uint = std::filesystem::space(std::filesystem::current_path()).available;
-	auto available_disk_space = try_mem_size(std::to_string(available_disk_space_uint));
-	if (opt_disk_space && *opt_disk_space >= min_disk && available_disk_space && *opt_disk_space <=  *available_disk_space)
+        auto available_disk_space = try_mem_size(std::to_string(available_disk_space_uint));
+        if (opt_disk_space && *opt_disk_space >= min_disk && available_disk_space &&
+            *opt_disk_space <= *available_disk_space)
         {
             disk_space = opt_disk_space;
         }
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -607,7 +607,8 @@ mp::InstanceStatus::Status grpc_instance_status_for(const mp::VirtualMachine::St
 // Computes the final size of an image, but also checks if the value given by the user is bigger than or equal than
 // the size of the image.
 mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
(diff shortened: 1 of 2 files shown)


Saviq:
Hi @katie-knister, unfortunately to be able to support older systems (like earlier macOS versions), we can't be on the edge of C++, where `<filesystem>` got introduced.

For our cross-platform needs, we decided on Qt to abstract the underlying system where possible, and [`QStorageInfo::bytesAvailable`](https://doc.qt.io/qt-5/qstorageinfo.html#bytesAvailable) should do for this use case.

Saviq:
@katie-knister I also had a look at your code and saw that if the user did not pass a disk size, the default would be used, and that wouldn't be checked for available space.

A better place to put this check would be here:

https://github.com/canonical/multipass/blob/88618b06b15c95456643ef400e452f484c9aadb9/src/daemon/daemon.cpp#L609-L632

katie-knister:
Hi @Saviq, thanks for the suggestions! I have updated the PR to change where the check is done and use QStorageInfo instead. The build now succeeds and passes 1238/1239 tests in BuildAndTest (Coverage). It fails DNSMasqServer.dnsmasq_fails_and_throws, where a runtime_error is expected but nothing is actually thrown. This is weird because the test seems unrelated to what I changed. Since my change is pretty straightforward and doesn’t involve suppressing any exceptions, I’m unsure what could have caused this, and am wondering if you have any ideas of how to resolve/debug this.

ricab:
Hi @katie-knister , yeah this test keeps coming back to bite us, I don't think it's related to your changes. The test is dependent on a timeout that is apparently still not enough...

Saviq:
bors try

src/daemon/daemon.cpp
```diff
@@ -624,6 +626,16 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
         disk_space = *command_line_value;
     }
 
+    auto available_bytes = QStorageInfo::root().bytesAvailable();
```
ricab, line 629: I see two issues with this:
1. we may be storing images elsewhere, either because multipass is installed in a different filesystem or because [`MULTIPASS_STORAGE`](https://github.com/canonical/multipass/pull/1789) was used
1. this does not currently work under snap confinement; you can try it yourself by using the PR channel which is built in CI: `multipass install/refresh channel=edge/pr1854 multipass`; but we're looking for a way to overcome this and will write back when we find it

ricab (changes requested):
Hi @katie-knister, thanks again for the PR :slightly_smiling_face: 

I think your approach is good overall. Using `MemorySize` for comparison and throwing `runtime_error` makes sense to me, clear error message too. Unfortunately, locations and snap confinement make it a bit trickier, so I don't think this will work just yet. Details inline.

katie-knister:
@ricab Thank you for your review! It sounds like this is blocked for now (if that is not the case, please let me know). I plan to get started on a different “good first issue” in the meantime. Please let me know if this issue becomes unblocked or if you have any recommendations for any other “good first issues” that could be particularly helpful to finish now.

Saviq:
Hi @ricab @katie-knister, I just tried this with `--devmode`, meaning confinement is disabled for the snap, and the issue persists.

I think in reality, the problem is what @ricab highlighted - you need to check the space where Multipass will actually be writing the image, not on the root filesystem, which, in case of running as a snap, is a read-only mount of the `core18` base snap:

```bash
 snap run --shell multipass -c "df /"
Filesystem     1K-blocks  Used Available Use% Mounted on
/dev/loop15        56704 56704         0 100% /
```

So indeed when you check for the root's available space, it's going to be 0.

src/daemon/daemon.cpp
Saviq: You also need to cater for [`QStorageInfo::bytesAvailable`](https://doc.qt.io/qt-5/qstorageinfo.html#bytesAvailable) returning `-1` in case of error.

katie-knister: @ricab @Saviq Thank you for reviewing this! I changed from checking the root's available space to using `DaemonConfig::data_directory` as @Saviq suggested. I also added a check for if `QStorageInfo::bytesAvailable` returns `-1`. The "BuildAndTest" checks don't seem to be running (nothing has updated after over an hour), so I thought I should just go ahead and let you know about these updates to see what you think.

ricab:
bors try

ricab (commented):
Hi @katie-knister, there seems to be a merge conflict now, so we don't get the snap and I couldn't test yet, but the code looks good to me. 

One further thing I'd like to request is for tests to the new functionality you're adding (we try to add only tested code as much as possible). Tests for this should probably go around [here](https://github.com/canonical/multipass/blob/7b1c2db9e682a8d0c0be22fcbbe5c4a6bf5752ad/tests/test_daemon.cpp#L869). 

You can't easily mock Qt stuff, but you wouldn't have to: you could check the available space in the test itself, then request just a bit more than that and confirm that the daemon throws the new `runtime_error`. [This](https://github.com/canonical/multipass/blob/fe4761285b51533de18a017594a353726f9fa7c8/tests/extra_assertions.h#L27) may be useful for that. One other test would be the -1 case (you'd have to overwrite the `data_directory` that is set [here](https://github.com/canonical/multipass/blob/7b1c2db9e682a8d0c0be22fcbbe5c4a6bf5752ad/tests/test_daemon.cpp#L232-L231) with something bogus).

katie-knister:
Hi @ricab, thank you for the information on testing! I resolved the merge conflict and added tests. They are all passing. I added another check to the `launches_with_correct_disk_size` test in a way that should fit with the current structure. However, I made a new test for the `-1` check since it required overwriting the `data_directory`, which is not something we would want to do for every combination of inputs passed to `launches_with_correct_disk_size`.

tests/test_daemon.cpp
```diff
@@ -860,13 +870,48 @@ TEST_P(LaunchImgSizeSuite, launches_with_correct_disk_size)
...
+
+    std::vector<std::string> all_parameters{first_command_line_parameter};
+    for (const auto& p : other_command_line_parameters)
+        all_parameters.push_back(p);
+
+    std::stringstream stream;
+    EXPECT_CALL(*mock_factory, create_virtual_machine(_, _)).Times(0);
+    send_command(all_parameters, trash_stream, stream);
+    EXPECT_THAT(stream.str(), HasSubstr("Failed to determine information about the volume containing"));
+}
```
ricab, line 913: This test can be a lot simpler than the one you used as inspiration. For instance, you don't need all the parameters, since you use only one value; then you don't need to construct a complex command either; and neither do you need all the mocking. Consider this:

```c++
TEST_P(InvalidDataDirectorySuite, launch_fails_with_invalid_data_directory)
{
    config_builder.data_directory = QString("invalid_data_directory");

    std::stringstream stream;
    EXPECT_CALL(*mock_factory, create_virtual_machine(_, _)).Times(0);
    send_command({GetParam()}, trash_stream, stream);
    EXPECT_THAT(stream.str(), HasSubstr("Failed to determine information about the volume containing"));
}
```
You'd need to change the param interface to receive only one string. You could then do something very similar for the other test, adding just `--disk ...` to the command.

tests/test_daemon.cpp
```diff
@@ -876,8 +921,12 @@ INSTANTIATE_TEST_SUITE_P(Daemon, MinSpaceViolatedSuite,
                                  Values("0", "0B", "0GB", "123B", "42kb", "100")));
 INSTANTIATE_TEST_SUITE_P(Daemon, LaunchImgSizeSuite,
                          Combine(Values("test_create", "launch"),
-                                 Values(std::vector<std::string>{}, std::vector<std::string>{"--disk", "4G"}),
+                                 Values(std::vector<std::string>{}, std::vector<std::string>{"--disk", "4G"},
+                                        std::vector<std::string>{"--disk", "9999G"}),
```
ricab, line 925: Good point, you can request a very large disk size and assume it's not acceptable. That will allow you to drop the QStorageInfo stuff. 

However, 10 terabytes may not be so uncommon in a few years. MemorySize currently uses a `long long`, which is guaranteed to be at least 64 bits. It's currently signed (something we could change), so we need something very large but below `2^63`. Perhaps `"999999999G"`. When 1 exabyte is not enough we need to change the underlying representation anyway :wink:

ricab (commented):
Hi @katie-knister, thanks for the tests. I'm proposing a few simplifications inline.

You did well to separate the `-1` check. Actually, I think the other one should be in a separate test too. The idea is that each test should check one property of the intended behavior.

Let me know if you need any help with it.

tests/test_daemon.cpp
```diff
@@ -860,13 +870,48 @@ TEST_P(LaunchImgSizeSuite, launches_with_correct_disk_size)
...
         EXPECT_THAT(stream.str(), AllOf(HasSubstr("Requested disk"), HasSubstr("below minimum for this image")));
     }
+    else if (other_command_line_parameters.size() > 0 &&
+             mp::MemorySize(other_command_line_parameters[1]) > mp::MemorySize(available_bytes_str + "B"))
+    {
+        std::stringstream stream;
+        EXPECT_CALL(*mock_factory, create_virtual_machine(_, _)).Times(0);
+        send_command(all_parameters, trash_stream, stream);
+        EXPECT_THAT(stream.str(), AllOf(HasSubstr("Available disk"), HasSubstr("below requested/default size")));
+    }
```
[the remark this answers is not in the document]
katie-knister: Makes sense, I was wondering about that. I switched it to its own test and removed the parts that I added to this test.

tests/test_daemon.cpp
katie-knister: Thank you for this information! I was wondering about simplifying this test, but I didn't want to accidentally remove parts that were actually necessary -- so I appreciate you taking the time to discuss how to simplify it. Now I've made the changes you suggested.

tests/test_daemon.cpp
katie-knister: Oh, interesting! Thank you, I've made this change.

katie-knister:
Hi @ricab, thank you for your feedback on the tests. I just made the changes you suggested.

tests/test_daemon.cpp
```diff
@@ -878,6 +921,10 @@ INSTANTIATE_TEST_SUITE_P(Daemon, LaunchImgSizeSuite,
                          Combine(Values("test_create", "launch"),
                                  Values(std::vector<std::string>{}, std::vector<std::string>{"--disk", "4G"}),
                                  Values("1G", mp::default_disk_size, "10G")));
+INSTANTIATE_TEST_SUITE_P(Daemon, DiskSpaceRequestedExceedsAvailableSuite,
+                         Combine(Values("test_create", "launch"),
+                                 Values(std::vector<std::string>{"--disk", "999999999G"})));
```
ricab, line 926: This is the only value ever used by the test (the vector with the 2 args), so there is no point having it as a parameter.

ricab (commented):
Hi @katie-knister, I think there is some confusion regarding the parameterized tests. The principle of your tests is correct, but some parts come from other tests and are not really applicable. I created a [draft PR](https://github.com/katie-knister/multipass/pull/1/files) with what I mean because it is simpler and more precise than words.

ricab (approved):
OK, this should be ready to merge.

bors merge

PR #1860, opened 2020-11-27
[daemon] add `multipass` group support

Saviq:


townsend2010:
Hey @Saviq,

What is the motivation for this change?  I thought a long time ago, we agreed not to support a `multipass` group...

Saviq:
The motivation is that if someone wants to, and has created the group, we'll use it.

We don't want to go about managing the group, but if it's there…

Driver was KDE folks building snaps in CI - they'd rather not use `sudo`.

townsend2010:
Ok, I see no harm in that.  Will approve them :smiley:

townsend2010 (approved):
Ok, makes sense if someone adds a `multipass` group.

bors merge

Saviq:
_Actually_ I thought of it more and felt it's giving a false sense of security. You could mount and edit the host's `/etc/sudoers` and gain root this way.

Issue #531, opened 2018-12-13
Suggest `--` when `exec` gets an unrecognised option


Saviq:
When trying to `exec` a command, it's likely to happen that someone forgets to separate options with a `--`. It would be helpful if we suggested its use:

```
$ multipass exec triumphant-gopher lsb_release -a
Unknown option 'a'.
```

Something like:
```
$ multipass exec triumphant-gopher lsb_release -a
Unknown option 'a'.

You need to separate the options to be passed to the command with "--", like this:
multipass exec triumphant-gopher -- lsb_release -a
```

dlbeck:
@Saviq  I am interested in taking on this issue.  I am new to Multipass (and open source in general) and am wondering if you have advice of how I should get start and where to look in the code. I am also open to contribute to a different issue if you think there is one that is more approachable. Thanks!

Saviq:
Hi @dlbeck, welcome to the FOSS train ;)

It should be easy to modify this part of the code to include a note about `--`:

https://github.com/canonical/multipass/blob/4f7ae5cac81ff379f5452a34a20a11e434d2089f/src/client/cli/cmd/exec.cpp#L30L34

dlbeck:
Hi @Saviq I have created a PR for this issue here: #1863 . It fails the `DNSMasqServer.dnsmasq_fails_and_throws` test which seems unrelated to my task. It looks like @katie-knister is having the same issue in PR #1854 so I'm wondering if the issue could be something else. Thanks!

ricab:
Hi @dlbeck , yeah, we keep finding that the timeout that this test depends on is too short for CI. I don't think it's related to your changes. Thanks for your PR!

ricab:
So we discussed a bit, and we think that the message should be a bit simpler than in this issue's description:

```
$ multipass exec triumphant-gopher lsb_release -a
Unknown option 'a'.

Options to the inner command should come after "--", like this:
multipass exec <instance> -- <command> <arguments>
```

dlbeck:
Thanks, that makes sense @ricab . I have made the changes in the PR when you have a chance to review it.

Saviq:
Fixed in #1863

PR #1863, opened 2020-11-29
Add suggested command when 'exec' gets an unrecognized option

dlbeck:
Addresses #531 . When a user forgets to separate options with `--` when running the `exec` command, provides a suggested command that includes the `--`. Below is a screenshot of the behavior with the change.

multipass_error_msg

### DIFF
--- a/src/client/cli/argparser.cpp
+++ b/src/client/cli/argparser.cpp
@@ -156,6 +156,18 @@ mp::ParseCode mp::ArgParser::commandParse(cmd::Command* command)
     if (!parsedOk)
     {
         cerr << qUtf8Printable(parser.errorText()) << '\n';
+
+        if (parser.errorText().contains("Unknown option"))
+        {
+            cerr << "\nYou need to separate the options to be passed to the command with \"--\", like this: \n";
+            QStringList suggestedCommand(arguments);
+            suggestedCommand.insert(3, "--");
+            foreach (auto& argument, suggestedCommand)
+            {
+                cerr << argument.toStdString() << " ";
+            }
+            cerr << '\n';
+        }
         return ParseCode::CommandLineError;
     }
 
--- a/src/client/cli/argparser.cpp
+++ b/src/client/cli/argparser.cpp
@@ -156,18 +156,6 @@ mp::ParseCode mp::ArgParser::commandParse(cmd::Command* command)
     if (!parsedOk)
     {
         cerr << qUtf8Printable(parser.errorText()) << '\n';
-
-        if (parser.errorText().contains("Unknown option"))
-        {
-            cerr << "\nYou need to separate the options to be passed to the command with \"--\", like this: \n";
-            QStringList suggestedCommand(arguments);
-            suggestedCommand.insert(3, "--");
-            foreach (auto& argument, suggestedCommand)
-            {
-                cerr << argument.toStdString() << " ";
-            }
-            cerr << '\n';
-        }
         return ParseCode::CommandLineError;
     }
 
@@ -351,6 +339,11 @@ QStringList mp::ArgParser::positionalArguments() const
     return positionalArguments;
 }
 
+QStringList mp::ArgParser::unknownOptionNames() const
+{
+    return parser.unknownOptionNames();
+}
+
 void mp::ArgParser::setVerbosityLevel(int verbosity)
 {
     if (verbosity > 4 || verbosity < 0)
@@ -366,4 +359,9 @@ void mp::ArgParser::setVerbosityLevel(int verbosity)
 int mp::ArgParser::verbosityLevel() const
 {
     return verbosity_level;
+}
+
+bool mp::ArgParser::containsArgument(const QString& argument) const
+{
+    return arguments.contains(argument);
 }
--- a/tests/test_cli_client.cpp
+++ b/tests/test_cli_client.cpp
@@ -718,35 +718,31 @@ TEST_F(Client, exec_cmd_no_double_dash_unknown_option_fails_print_suggested_comm
     std::stringstream cerr_stream;
     EXPECT_THAT(send_command({"exec", "foo", "cmd", "--unknownOption"}, trash_stream, cerr_stream),
                 Eq(mp::ReturnCode::CommandLineError));
-    EXPECT_NE(std::string::npos,
-              cerr_stream.str().find(
-                  "Options to the inner command should come after \"--\", like this:\nmultipass exec <instance> -- "
-                  "<command> <arguments>\n"))
-        << "cerr has: " << cerr_stream.str();
+    EXPECT_THAT(
+        cerr_stream.str(),
+        HasSubstr("Options to the inner command should come after \"--\", like this:\nmultipass exec <instance> -- "
+                  "<command> <arguments>\n"));
 }
 
 TEST_F(Client, exec_cmd_double_dash_unknown_option_fails_does_not_print_suggested_command)
 {
     std::stringstream cerr_stream;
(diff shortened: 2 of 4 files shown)


Saviq:
bors try

src/client/cli/argparser.cpp
```diff
@@ -156,6 +156,18 @@ mp::ParseCode mp::ArgParser::commandParse(cmd::Command* command)
     if (!parsedOk)
     {
         cerr << qUtf8Printable(parser.errorText()) << '\n';
+
+        if (parser.errorText().contains("Unknown option"))
+        {
+            cerr << "\nYou need to separate the options to be passed to the command with \"--\", like this: \n";
+            QStringList suggestedCommand(arguments);
+            suggestedCommand.insert(3, "--");
```
ricab, line 164: Getting the right command in all cases would be quite tricky. A few examples of difficulties: 
  * the input could require the `--` to be in a different place (e.g. `multipass -vvv exec ...`)
  * bad options might be mistakes on genuine multipass options (e.g. `multipass exec --helpp`)
  * there could be more than one problem (e.g. `multipass --nonsense exec ls -la`)
  * If arguments had special characters, when printing the amended command back, we'd have to at least quote those

For these reasons, I think it's best to just say that the right usage is `multipass exec <instance> -- <command> <arguments>`.

src/client/cli/argparser.cpp
```diff
@@ -156,6 +156,18 @@ mp::ParseCode mp::ArgParser::commandParse(cmd::Command* command)
     if (!parsedOk)
     {
         cerr << qUtf8Printable(parser.errorText()) << '\n';
+
+        if (parser.errorText().contains("Unknown option"))
+        {
+            cerr << "\nYou need to separate the options to be passed to the command with \"--\", like this: \n";
```
ricab, line 162: I think we could use a nicer message. Something like: "Options to the inner command should come after \"--\", like this:"

ricab (changes requested):
Hi @dlbeck. Thanks again for your PR :slightly_smiling_face: I am requesting a few changes inline. I know that you followed the issue description closely, but we discussed a bit now and there are a couple of aspects that we think should be simplified. I will comment on the issue as well, for coherence.

dlbeck:
Thanks for the review @ricab . I have addressed those changes!

ricab:
bors try

ricab:
bors try-

ricab:
bors try

ricab:
grrr 

bors try

ricab (commented):
Hi @dlbeck, The code looks good to me now, thanks. We're having some trouble with CI currently and I wasn't able to test yet, will do that later.

However, we do try to add only tested code as much as possible, so I would like to request that you add tests for the new functionality you add in this PR. That would involve some simple [tests for the `ArgParser`](https://github.com/canonical/multipass/blob/33a351d419f8f755fa222c68f85717e50ba3dbf3/tests/test_argparser.cpp#L57) and something checking the error msg [in the exec command](https://github.com/canonical/multipass/blob/0d1588c6ef197ea078fd48c095d27c5ca4840a48/tests/test_cli_client.cpp#L721). Here's an example of [how to check what you get in the cerr stream](https://github.com/canonical/multipass/blob/0d1588c6ef197ea078fd48c095d27c5ca4840a48/tests/test_cli_client.cpp#L588).

ricab:
Once again

bors try

dlbeck:
@ricab Thanks for pointing me in the right direction on testing. I have added 3 tests

1. If a command has an unknown option and doesn't contain `--`, the new error msg prints.
2. If a command has an unknown option and does contain `--`, the new error msg doesn't print.
3. If a command does not have any unknown options and does not contain `--`, the new error msg doesn't print

Right now, 1 passes but 2 and 3 fail. I'm pretty confident these fail because I don't use the testing framework correctly. Test 2 requires a `--` to be in the `send_command` . However,  I believe the `--` causes `send_command` to fail because when I remove it, the `send_command` failure in the screenshot below no longer occurs. Test 3 seems to fail because we don't pass in a correct "known" option.  When I tried passing in a known option like `--help`, I got the `send_command` error (see screenshot). This is interesting because when I use an unknown option, the `send_command` error no longer occurs.

Any advice you have on resolving these issues or properly using the exec testing framework would be appreciated. Thanks!
image

tests/test_cli_client.cpp
```diff
@@ -713,6 +713,42 @@ TEST_F(Client, exec_cmd_help_ok)
...
+              cerr_stream.str().find(
+                  "Options to the inner command should come after \"--\", like this:\nmultipass exec <instance> -- "
+                  "<command> <arguments>\n"))
+        << "cerr has: " << cerr_stream.str();
+}
+
+TEST_F(Client, exec_cmd_double_dash_unknown_option_fails_does_not_print_suggested_command)
+{
+    std::stringstream cerr_stream;
+    EXPECT_THAT(send_command({"exec", "foo", "--", "cmd", "--unknownOption"}, trash_stream, cerr_stream),
```
ricab, line 731: Hey @dlbeck , this looks well structured, so Multipass wouldn't fail the command. I believe you want `--unknownOption` to come before `--`.

tests/test_cli_client.cpp
```diff
@@ -713,6 +713,42 @@ TEST_F(Client, exec_cmd_help_ok)
...
+                  "Options to the inner command should come after \"--\", like this:\nmultipass exec <instance> -- "
+                  "<command> <arguments>\n"))
+        << "cerr has: " << cerr_stream.str();
+}
+
+TEST_F(Client, exec_cmd_no_double_dash_no_unknown_option_fails_does_not_print_suggested_command)
+{
+    std::stringstream cerr_stream;
+    EXPECT_THAT(send_command({"exec", "foo", "cmd", "--help"}, trash_stream, cerr_stream),
+                Eq(mp::ReturnCode::CommandLineError));
```
ricab, line 744: Here, `--help` is accepted as a request for help to the `exec` command, so it doesn't fail either. I think you want to expect a success in this case.

ricab (commented):
Hi @dlbeck, you're very close. Only a couple of simple changes and it should work.

tests/test_cli_client.cpp
```diff
@@ -713,6 +713,42 @@ TEST_F(Client, exec_cmd_help_ok)
...
+TEST_F(Client, exec_cmd_no_double_dash_unknown_option_fails_print_suggested_command)
+{
+    std::stringstream cerr_stream;
+    EXPECT_THAT(send_command({"exec", "foo", "cmd", "--unknownOption"}, trash_stream, cerr_stream),
+                Eq(mp::ReturnCode::CommandLineError));
+    EXPECT_NE(std::string::npos,
+              cerr_stream.str().find(
+                  "Options to the inner command should come after \"--\", like this:\nmultipass exec <instance> -- "
+                  "<command> <arguments>\n"))
+        << "cerr has: " << cerr_stream.str();
```
ricab, line 725: Oh, one more thing. You can use a higher-level testing predicate, like this:

```c++
    EXPECT_THAT(cerr_stream.str(), HasSubstr("expected substr"));
```

It's simpler, more expressive, and leaves it to the testing framework to print the values that didn't match.

dlbeck:
Thanks @ricab! I modified the tests and they are all passing now.

ricab:
bors try

ricab:
bors retry

dlbeck:
@ricab Is there anything else that needs to be done on this PR before it's landed?

ricab (approved):
I finally tested and everything seems to work as it should. Code and tests look good. Thanks again @dlbeck 

bors r+

Saviq:
bors r=ricab

PR #1868, opened 2020-12-01
Fix lambda captures

luis4a0:
Clang complains about capturing not needed variables. In particular, we are capturing a non-mutable constexpr, which does not need to be captured. See [here](https://en.cppreference.com/w/cpp/language/lambda).


townsend2010 (approved):
Yeps.

bors merge

townsend2010:
Ok, seems GHA is back up.

bors merge

PR #1869, opened 2020-12-03
[networking] Drop unused traits in VM description

ricab:
We had left this for after the main PRs were merged. 

Revert to mac address only instead of a full network interface struct in VM description. This keeps the code aligned with persistent records and avoids failure to honor the extra properties (id and mode).

### DIFF
--- a/src/platform/backends/qemu/qemu_vm_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vm_process_spec.cpp
@@ -38,27 +38,27 @@ QStringList initial_qemu_arguments(const mp::VirtualMachineDescription& desc, co
     auto mem_size = QString::number(desc.mem_size.in_megabytes()) + 'M'; /* flooring here; format documented in
     `man qemu-system`, under `-m` option; including suffix to avoid relying on default unit */
 
-    QStringList args{"--enable-kvm",
-                     "-hda",
-                     desc.image.image_path,
-                     "-smp",
-                     QString::number(desc.num_cores),
-                     "-m",
-                     mem_size,
-                     "-device",
-                     QString("virtio-net-pci,netdev=hostnet0,id=net0,mac=%1")
-                         .arg(QString::fromStdString(desc.default_interface.mac_address)),
-                     "-netdev",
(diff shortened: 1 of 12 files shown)


src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -38,27 +38,27 @@ QStringList initial_qemu_arguments(const mp::VirtualMachineDescription& desc, co
...
+        QString("tap,id=hostnet0,ifname=%1,script=no,downscript=no").arg(tap_device_name),
+        "-qmp",
+        "stdio",
+        "-cpu",
+        "host",
+        "-chardev",
+        "null,id=char0",
+        "-serial",
+        "chardev:char0",
+        "-nographic"};
```
luis4a0, line 61: Did you force this reformatting? It seems to me that clang-format did a mistake.

luis4a0 (approved):
Yes, this is something needed. AFAIK, the name of the record in the JSON was `mac_address` and the change to `default_mac_address` makes complete sense: we have many MAC addresses now. So let's merge this after clarifying the issue with formatting!

ricab:
Hi @luis4a0, thanks for the review

> we have many MAC addresses now

Yes, that's what I thought too :)

src/platform/backends/qemu/qemu_vm_process_spec.cpp
ricab: I have a git hook to run clang-format on my commits and refuse the commit if changes are needed. That hook signaled this, and running manually, this was the result. 

Clang-format's decisions are not always the most pleasing, but it's just following rules. I think here it prefers to save one level of indentation (in line 50).

ricab: If this bothers you, I can create a separate var for the big virtio string.

luis4a0: No, it doesn't bother me at all. Just commenting.

luis4a0:
Will merge manually, because bors doesn't take into account the private counterpart so the private build fails.

PR #1871, opened 2020-12-04
[gha] only dispatch on PRs when within the same repository

Saviq:


.github/workflows/linux.yml
```diff
@@ -203,6 +203,9 @@ jobs:
   Dispatch:
     needs: BuildAndTest
 
+    # Only dispatch if we have access to secrets.
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
```
ricab, line 207: Could you please remind me what this means? Shouldn't this be an `&&`? Or is push for bors and the rest assumed to be a pull_request? Are there no other possible event types?

Also, we needed `always()` in other cases, but I am not sure why anymore. Was it to continue even if there was a failure? We don't want that here, do we? Just checking.

ricab (commented):
This GHA stuff is pretty obscure to me, so I asked some questions inline (posted separately by mistake).

.github/workflows/linux.yml
Saviq: > Could you please remind me what this means?

This means that it's either a push, or a PR from the same repository. On PRs from "outside" [secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) are not available, and we wouldn't want to dispatch without a review anyway (it could be malicious).

> Shouldn't this be an `&&`? Or is push for bors and the rest assumed to be a pull_request? Are there no other possible event types?

[There are](https://docs.github.com/en/actions/reference/events-that-trigger-workflows), but _this_ workflow is only triggered on `push`es and `pull_request`s:

https://github.com/canonical/multipass/blob/acdc1351df7fc968509922272865e5a5b33309ac/.github/workflows/linux.yml#L3-L13

> Also, we needed `always()` in other cases, but I am not sure why anymore. Was it to continue even if there was a failure? We don't want that here, do we? Just checking.

Correct, we only want to dispatch when everything went well.

ricab: Thanks, makes sense.

ricab (approved):
Thank you for the explanations. All looks sensible to me now.

bors r+

Saviq:
bors r=ricab

PR #1876, opened 2020-12-08
Rename list-networks command

luis4a0:
This reflects a change in the Multipass CLI UX document.

The private bits of this PR are in https://github.com/canonical/multipass-private/pull/335.


Saviq (commented):
I'm not sure we even should have the short command for this. There's not an obvious acronym (I was thinking `nws` for it being plural).

It's not a command people will need to run often.

I vote for dropping `ln`/`lsn` altogether.

townsend2010:
I agree, there shouldn't be a need for an alias for `networks` as there is no common CLI command equivalent that makes sense to use for muscle memory.

luis4a0:
Thanks for your comments! I used the command alias only because the other commands have one. But I agree it's not a very used command. I thus removed the alias.

ricab (approved):
All looks good, thanks!

bors r+

ricab:
bors single on
bors merge

PR #1878, opened 2020-12-09
Fix direct use of `std::cin` in launch cmd

ricab:
It needs to use the input stream in its `Terminal` attribute instead.


townsend2010 (approved):
Yep

bors merge

PR #1880, opened 2020-12-11
[daemon] Remove network ID reinterpretation

ricab:
Stop reinterpreting network IDs at the daemon, and remove the corresponding interface from the VM factories. Network ID reinterpretation can be treated as an implementation detail of the VirtualBox backend.

### DIFF
--- a/include/multipass/virtual_machine_factory.h
+++ b/include/multipass/virtual_machine_factory.h
@@ -65,9 +65,6 @@ public:
     // List all the network interfaces seen by the backend.
     virtual std::vector<NetworkInterfaceInfo> networks() const = 0;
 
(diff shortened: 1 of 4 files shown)


luis4a0 (approved):
This makes complete sense, we don't want to add many layers (daemon, factory) to this reinterpretation, since it's only a detail on the VirtualBox backend. The code looks good.

ricab:
Thanks @luis4a0, but I wanted to get your PR in first to make sure mine are OK on the other platforms. We'll then need to merge this one manually as well.

luis4a0:
Oh, sorry, I misunderstood what you said :(

PR #1889, opened 2020-12-16
[tests] Add disk size check util and mock

townsend2010:
This fixes the reliance on the amount of disk space available during tests which can be non-deterministic.

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -86,7 +86,7 @@ void PrintTo(const YAML::Node& node, ::std::ostream* os)
 
 namespace
 {
-const qint64 default_total_bytes{16106127360};
+const qint64 default_total_bytes{16'106'127'360}; // 16G
 
 template<typename R>
   bool is_ready(std::future<R> const& f)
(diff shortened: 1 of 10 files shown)


tests/test_daemon.cpp
```diff
@@ -83,6 +86,8 @@ void PrintTo(const YAML::Node& node, ::std::ostream* os)
 
 namespace
 {
+const qint64 default_total_bytes{16106127360};
```
ricab, line 89:
```suggestion
const qint64 default_total_bytes{16'106'127'360}; // 16G
```

ricab (approved):
Thanks for this, it works nicely :) 

We may have to mock the function at a larger scope if we ever increase the default disk, but I guess that would be easy. I have a minor suggestion inline, but approving already.

bors delegate+

tests/test_daemon.cpp
townsend2010: Oh, cool, I didn't know one could add `'` to make the number more readable.

townsend2010:
> We may have to mock the function at a larger scope if we ever increase the default disk, but I guess that would be easy.

Yeah, it would be real easy to do that by adding a default mock return in the `MockDaemon` which would affect every test that didn't `REPLACE` the function.

townsend2010:
bors r=ricab

tests/test_daemon.cpp
ricab: Yeah introduced in C++14

PR #1893, opened 2020-12-18
Tiny mockenv change

ricab:
Just a couple of mends to `SetEnvScope` (which I happened to have a close look at when investigating a test failure when running within CLion).

### DIFF
--- a/tests/mock_environment_helpers.h
+++ b/tests/mock_environment_helpers.h
@@ -28,12 +28,12 @@ namespace test
 class SetEnvScope
 {
 public:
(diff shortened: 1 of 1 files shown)


luis4a0 (approved):
Yeah, it took me a while to realize what `SetEnvScope` was doing but now I understand and see how your changes polish the code.

Will merge after the Windows test succeeds (it didn't at first because there was an issue when downloading some prerequisite).

luis4a0:
Yeah, everything seems to be fixed now.

bors r+

PR #1894, opened 2020-12-18
[cmake] Fix a missing dependency for tests

ricab:
Another tiny fix that I found when trying to better integrate with CLion. Without this, tests fail if built on their own (e.g. `make covreport`).


luis4a0 (approved):
Yeah, good catch :)

Will merge after checking that the Windows build succeeds.

PR #1895, opened 2020-12-18
Fix testing in foreign snap

ricab:
Use utility to obtain home for auto-mount. This fixes a failure when running with an environment inherited from another snap (e.g. CLion). The test in question would change `HOME` but still get an auto-mount of `SNAP_REAL_HOME`, if this was defined.


luis4a0 (approved):
This is very useful, indeed. Moreover, tests are complete and include a nice refactoring.

luis4a0:
Now the CI should work. Let's give it a try.

bors r+

PR #1906, opened 2021-01-06
[snap] Use core20 base for building and running

townsend2010:


.github/workflows/linux.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       run: |
         COVERAGE_BUILD=${HOME}/build
         mkdir --parents ${COVERAGE_BUILD}
-        sg lxd -c "/snap/bin/lxc profile device add default build disk source=${COVERAGE_BUILD} path=/root/parts/multipass/build"
+        sg lxd -c "/snap/bin/lxc profile device add default build disk source=${COVERAGE_BUILD} path=/root/parts/multipass"
         echo "::set-output name=build::${COVERAGE_BUILD}"
```
Saviq, line 109: You want to add `/build` to this output, otherwise codecov will scan much more than it needs to.

.github/workflows/linux.yml
```diff
@@ -132,9 +132,17 @@ jobs:
 
     - name: Measure coverage
       if: ${{ matrix.build-type == 'Coverage' }}
-      timeout-minutes: 5
+      timeout-minutes: 10
+      shell: /bin/bash -xeuo pipefail {0}
```
Saviq, line 136: Can revert that now.

.github/workflows/linux.yml
```diff
@@ -132,9 +132,17 @@ jobs:
...
       run: |
         sg lxd -c '/snap/bin/lxc start snapcraft-multipass'
+        timeout 10 sg lxd -c \
+          'while [ "$( /snap/bin/lxc exec snapcraft-multipass -- \
+             systemctl show --property=ActiveState snapd )" != "ActiveState=active" ]; do sleep 1; done'
+        sg lxd -c \
+          '/snap/bin/lxc exec snapcraft-multipass -- \
+             sh -c "sudo sed -i \"s/use JSON/use JSON::XS/\" \`which geninfo\`"'
+        sg lxd -c '/snap/bin/lxc exec snapcraft-multipass -- apt-get -y install libjson-xs-perl'
```
Saviq, line 145: Add a comment why all this?

Saviq (changes requested):
After a quick look on the coverage changes, I think it's just down to how it's measured now. For example, all the `on_failure` lambda misses suggest they were counted as hits before just due to the lambda being declared, rather than called.

Looks good overall!

townsend2010:
Hey @Saviq,

> After a quick look on the coverage changes, I think it's just down to how it's measured now.

Yeah, this is the same conclusion I drew on Friday after looking through it a little closer.

.github/workflows/linux.yml
townsend2010: Ok, in iterating over this, I had it there, removed it, added it back, etc. and it never seemed to make a difference.  But if adding is The Right Thing to Do, then I'll put it there.

.github/workflows/linux.yml
townsend2010: Oops, right.

.github/workflows/linux.yml
townsend2010: Will do.

.github/workflows/linux.yml
townsend2010: Ok, done and rebased into a previous related commit.

.github/workflows/linux.yml
townsend2010: Ok, done and rebased into a previous related commit.

.github/workflows/linux.yml
townsend2010: Ok, done and rebased into a previous related commit.

Saviq:
Let's do it.

bors r+

Saviq:
Wait.

bors r-

.github/workflows/linux.yml
```diff
@@ -105,8 +105,8 @@ jobs:
       run: |
         COVERAGE_BUILD=${HOME}/build
         mkdir --parents ${COVERAGE_BUILD}
-        sg lxd -c "/snap/bin/lxc profile device add default build disk source=${COVERAGE_BUILD} path=/root/parts/multipass/build"
-        echo "::set-output name=build::${COVERAGE_BUILD}"
+        sg lxd -c "/snap/bin/lxc profile device add default build disk source=${COVERAGE_BUILD} path=/root/parts/multipass"
+        echo "::set-output name=build::${COVERAGE_BUILD}/build"
```
Saviq, line 109: We end up with `${HOME}/build/build` now… While that works, we could probably tweak `COVERAGE_BUILD` to be `${HOME}/coverage` or so.

Saviq (approved):
OK cleaned it up.

bors r+

Issue #1904, opened 2021-01-05
Xenial images fail to launch


Saviq:
Since we introduced bridging on `edge`, xenial images don't work any more:
```plain
 mp launch xenial
launch failed: The following errors occurred:                                   
failed to determine IP address
```
Most likely the network config we started providing is incompatible with the old version of cloud-init in those images.

PR #1913, opened 2021-01-08
Don't use cloud-init to configure network in old images

luis4a0:
We use to configure the network interfaces using cloud-init. Specifically, we use version 2 of the network configuration. However, old images, like xenial, don't support this version of the network configuration file.

In this PR, we do the following to overcome the situation:
1. don't allow old images to use bridging (which needed configuration via cloud-init);
2. don't use cloud-init at all to configure network interfaces in the case on which bridging is not used.

This addresses issue https://github.com/canonical/multipass/issues/1904

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -150,8 +150,9 @@ auto make_cloud_init_network_config(const std::string default_mac_addr,
 {
     YAML::Node network_data;
 
-    // If there exists only the default interface in the instance, then we don't need cloud-init.
-    if (!extra_interfaces.empty())
+    // Generate the cloud-init file only if there is at least one extra interface needing auto configuration.
+    if (std::find_if(extra_interfaces.begin(), extra_interfaces.end(),
+                     [](const auto& iface) { return iface.auto_mode; }) != extra_interfaces.end())
     {
         network_data["version"] = "2";
 
@@ -388,7 +389,10 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
         std::string full_name =
             (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
 
-        if (no_bridging_images.find(full_name) != no_bridging_images.end())
+        if (no_bridging_images.find(full_name) != no_bridging_images.end() &&
+            std::find_if(request->network_options().begin(), request->network_options().end(), [](const auto& net) {
+                return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
+            }) != request->network_options().end())
             throw std::runtime_error(fmt::format("--network not implemented for {}", full_name));
 
         try
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -85,6 +85,7 @@ constexpr auto cloud_init_timeout = 5min;
 constexpr auto stop_ssh_cmd = "sudo systemctl stop ssh";
 const std::string sshfs_error_template = "Error enabling mount support in '{}'"
                                          "\n\nPlease install the 'multipass-sshfs' snap manually inside the instance.";
+const std::unordered_set<std::string> no_bridging_images = {"release:16.04", "release:xenial", "snapcraft:core"};
 
 mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
 {
@@ -379,8 +380,11 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
 
     if (!request->network_options().empty())
     {
-        if (request->image() == "xenial")
-            throw std::runtime_error("--network not implemented for xenial");
+        std::string full_name =
+            (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+
+        if (no_bridging_images.find(full_name) != no_bridging_images.end())
+            throw std::runtime_error(fmt::format("--network not implemented for {}", full_name));
 
         try
         {
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -394,7 +394,8 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
                 return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
             }) != request->network_options().end())
         {
-            throw std::runtime_error(fmt::format("Automatic bridge configuration not implemented for {}", full_name));
+            throw std::runtime_error(fmt::format(
+                "Automatic network configuration not available for {}. Consider using manual mode.", full_name));
         }
 
         try
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -382,41 +382,40 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
                                                             const mp::VirtualMachineFactory& factory,
                                                             mp::LaunchError& option_errors)
 {
-    std::vector<mp::NetworkInterfaceInfo> factory_networks;
+    std::vector<mp::NetworkInterface> interfaces;
 
-    if (!request->network_options().empty())
+    mp::optional<std::vector<mp::NetworkInterfaceInfo>> factory_networks = mp::nullopt;
+
+    std::string full_name =
+        (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+
+    bool dont_allow_auto = no_bridging_images.find(full_name) != no_bridging_images.end();
+
+    for (const auto& net : request->network_options())
     {
-        std::string full_name =
-            (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+        if (!factory_networks)
+        {
+            try
+            {
+                factory_networks = factory.networks();
+            }
+            catch (const mp::NotImplementedOnThisBackendException&)
+            {
+                throw mp::NotImplementedOnThisBackendException("bridging");
+            }
+        }
 
-        if (no_bridging_images.find(full_name) != no_bridging_images.end() &&
-            std::find_if(request->network_options().begin(), request->network_options().end(), [](const auto& net) {
-                return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
-            }) != request->network_options().end())
+        if (dont_allow_auto && net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO)
         {
             throw std::runtime_error(fmt::format(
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -373,6 +385,12 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
 
     if (!request->network_options().empty())
     {
+        std::string full_name =
+            (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+
+        if (no_bridging_images.find(full_name) != no_bridging_images.end())
+            throw std::runtime_error(fmt::format("--network not implemented for {}", full_name));
```
ricab, line 392: I just realized we should not refer to a CLI specific thing like `--network` from the daemon, since it may be meaningless in other clients (e.g. future GUI or some other thing using the GRPC).

Would you mind replacing with "extra networks" or similar? Here and below (line 401). I think this is one of those cases where it makes sense to change in the same PR, for consistency, but if you prefer, we can do it in a separate PR.

src/daemon/daemon.cpp
```diff
@@ -178,8 +188,10 @@ auto make_cloud_init_image(const std::string& name, const QDir& instance_dir, YA
     mp::CloudInitIso iso;
     iso.add_file("meta-data", mpu::emit_cloud_config(meta_data_config));
     iso.add_file("vendor-data", mpu::emit_cloud_config(vendor_data_config));
-    iso.add_file("network-config", mpu::emit_cloud_config(network_data_config));
     iso.add_file("user-data", mpu::emit_cloud_config(user_data_config));
+    if (!network_data_config.IsNull())
+        iso.add_file("network-config", mpu::emit_cloud_config(network_data_config));
```
ricab, line 193: Care to test this one too? :slightly_smiling_face:

ricab (commented):
Thanks @luis4a0, this looks good! Only a couple things inline.

Saviq:
I just realized something. It's not `--network` that we can't do on those images. It's `--network mode=auto` — we can still allow manually configured interfaces!

luis4a0:
Right, that escaped from my sight. Fixed, I should add some test for it though.

src/daemon/daemon.cpp
luis4a0: I prefer to do it here, if @Saviq is ok with that. Not the cleanest, but we don't want to delay releasing this anymore.

src/daemon/daemon.cpp
luis4a0: This line seems to be tested now: https://codecov.io/gh/canonical/multipass/pull/1913/src/src/daemon/daemon.cpp?before=src/daemon/daemon.cpp

In fact, I am surprised by it not being covered at the moment you made the comment (I checked that it was indeed uncovered at that point).

src/daemon/daemon.cpp
luis4a0: Done.

src/daemon/daemon.cpp
```diff
@@ -373,14 +386,24 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
...
+            (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+
+        if (no_bridging_images.find(full_name) != no_bridging_images.end() &&
+            std::find_if(request->network_options().begin(), request->network_options().end(), [](const auto& net) {
+                return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
+            }) != request->network_options().end())
+        {
+            throw std::runtime_error(fmt::format("Automatic bridge configuration not implemented for {}", full_name));
+        }
+
```
ricab, line 399: I understand you're doing this here to avoid wasting wasting cycles in network listing. However, if I try this in a backend that does not support bridging, I now get:
```
$ multipass launch --network en0 xenial
launch failed: Automatic bridge configuration not implemented for release:xenial
```

I wonder if it's worth wasting one or two seconds and check the mode only in the loop below, along with other validations. That would also make the code cleaner. You could drop the initial `if` entirely and get the networks in the first iteration of the loop, like so (pseudocode):

```C++
auto host_networks = [&factory](){
  static auto networks = factory.networks(); // called only once
  return networks;
};

for (...)
{
  auto networks = host_networks();
  ...
} 
```

WDYT? Worth it? Separate PR?

src/daemon/daemon.cpp
```diff
@@ -373,14 +386,24 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
...
     {
+        std::string full_name =
+            (request->remote_name().empty() ? "release" : request->remote_name()) + ':' + request->image();
+
+        if (no_bridging_images.find(full_name) != no_bridging_images.end() &&
+            std::find_if(request->network_options().begin(), request->network_options().end(), [](const auto& net) {
+                return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
+            }) != request->network_options().end())
+        {
+            throw std::runtime_error(fmt::format("Automatic bridge configuration not implemented for {}", full_name));
```
ricab, line 397: I wonder if we could guide people to `mode=manual` when reading this message. I would also replace "bridge" with "network" and "implemented" with "available". So, perhaps something like `Automatic network configuration not available for {}. Consider using manual mode.`. WDYT?

ricab (commented):
This is mostly ready. The code looks good and behaves well in testing. I have only a couple of minor suggestions for extending the tests, as well as a couple of proposals for the network validation, depending on what you think.

tests/test_daemon.cpp
```diff
@@ -1061,7 +1086,16 @@ TEST_F(Daemon, refuses_launch_because_bridging_is_not_implemented)
...
+    EXPECT_THAT(err_stream.str(), HasSubstr("The bridging feature is not implemented on this backend"));
+}
+
+TEST_F(Daemon, refuses_bridging_in_old_image)
+{
+    mp::Daemon daemon{config_builder.build()};
+
+    std::stringstream err_stream;
+    send_command({"launch", "xenial", "--network", "eth0"}, std::cout, err_stream);
+    EXPECT_THAT(err_stream.str(), HasSubstr("Automatic bridge configuration not implemented for release:xenial"));
```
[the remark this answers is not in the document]
luis4a0: Sure, done.

tests/test_daemon.cpp
```diff
@@ -746,6 +750,28 @@ TEST_P(DaemonCreateLaunchTestSuite, adds_pollinate_user_agent_to_cloud_init_conf
...
+
+    send_command(launch_args);
+}
+
+INSTANTIATE_TEST_SUITE_P(Daemon, LaunchWithNoNetworkCloudInit,
+                         Values(std::vector<std::string>{"launch"}, std::vector<std::string>{"launch", "xenial"},
+                                std::vector<std::string>{"launch", "groovy"},
+                                std::vector<std::string>{"launch", "xenial", "--network", "id=eth0,mode=manual"},
+                                std::vector<std::string>{"launch", "groovy", "--network", "id=eth0,mode=manual"},
+                                std::vector<std::string>{"launch", "--network", "id=eth0,mode=manual"}));
```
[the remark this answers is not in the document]
luis4a0: Done.

src/daemon/daemon.cpp
luis4a0: Yeah, I think it's worth it. But I'll separate the code in a new commit.

src/daemon/daemon.cpp
luis4a0: Makes complete sense. Changed.

src/daemon/daemon.cpp
ricab: Actually, scratch that approach using static, my mistake. I was thinking that would be local to the lambda object, but no, it's local to the lambda's `operator()`, so the static will stay for the remainder of the program. (Even if such caching was desired, it should be done centrally, not here). 

Perhaps the simplest here is just an optional (to distinguish from the case where no networks are found). Something like:

```C++
optional<decltype(factory.networks())> networks = nullopt;
for (...)
{
  if (!networks)
    networks = factory.networks();
  ...
}
```

luis4a0: About caching, I don't think it's a good idea. There is always the possibility of hotplugging devices, what will make caching useless. Unless we implement another mechanism to trigger an update on such events. In any case, out of the scope of this PR.

I was in doubt whether the scope of the `static` would be this function or global, too. Let me try the new approach, without the static thing.

luis4a0: Done. The code looks better, indeed.

ricab: > About caching, I don't think it's a good idea. There is always the possibility of hotplugging devices, what will make caching useless. Unless we implement another mechanism to trigger an update on such events. In any case, out of the scope of this PR.

Sure, I was not proposing, actually excluding it here :) I meant that this wouldn't be the way or place to achieve it, *even* if we wanted it. It's something we can discuss in the future, but disregard now.

> I was in doubt whether the scope of the static would be this function or global, too.

The scope is the function, but it still has static duration. Meaning it will be initialized in the first function call (`operator()`) and stay there, no matter what different objects (lambda instances) are used to call it. 

I also considered `[networks = factory.networks()] { return networks; }`, but then it would be called when defining the lambda, so outside the loop, which is what we want to avoid.

src/daemon/daemon.cpp
```diff
@@ -369,30 +382,45 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
...
 
-    std::vector<mp::NetworkInterface> interfaces;
+            if (no_bridging_images.find(*full_name) != no_bridging_images.end() &&
+                std::find_if(request->network_options().begin(), request->network_options().end(), [](const auto& net) {
+                    return net.mode() == multipass::LaunchRequest_NetworkOptions_Mode_AUTO;
+                }) != request->network_options().end())
+            {
+                throw std::runtime_error(fmt::format(
+                    "Automatic network configuration not available for {}. Consider using manual mode.", *full_name));
+            }
```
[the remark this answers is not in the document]
luis4a0: Right, that escaped me :(

Corrected, you can take a look now.

Saviq:
macOS build available: [multipass-1.7.0-dev.525.pr1913+g91b77178.mac-Darwin.pkg](https://multipass-ci.s3.amazonaws.com/pr1913/multipass-1.7.0-dev.525.pr1913%2Bg91b77178.mac-Darwin.pkg)
Snap build available: `snap refresh multipass --channel edge/pr1913`
Windows build available: [multipass-1.7.0-dev.525.pr1913+g91b77178.win-win64.exe](https://multipass-ci.s3.amazonaws.com/pr1913/multipass-1.7.0-dev.525.pr1913%2Bg91b77178.win-win64.exe)

PR #1912, opened 2021-01-08
Implement `networks` in LXD backend

ricab:
This implements network listing from the LXD backend. As discussed, only bridges are returned. 

If available, descriptions from LXD are used. Otherwise, we fall back to "Network bridge". 

Example:
```
$ multipass networks
Name    Type    Description
lxdbr0  bridge  Network bridge
mpbr0   bridge  Network bridge for Multipass
virbr0  bridge  Network bridge
```

Bridge members are left for another PR, as well as `launch --network`.

PR #1917, opened 2021-01-13
LXD test improvements

ricab:
LXD test improvements extracted from #1912: provide an empty but non-null reply from the mock network mgr by default, to avoid crashing on unexpected mock calls. Expect calls explicitly. Remove superfluous calls to `unique_ptr::get`

### DIFF
--- a/tests/lxd/mock_network_access_manager.h
+++ b/tests/lxd/mock_network_access_manager.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Canonical, Ltd.
+ * Copyright (C) 2020-2021 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -20,6 +20,8 @@
 
 #include <multipass/network_access_manager.h>
 
+#include "mock_local_socket_reply.h"
+
 #include <gmock/gmock.h>
 
 using namespace testing;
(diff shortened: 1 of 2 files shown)


townsend2010 (approved):
LGTM.  The only thing is that:
https://github.com/canonical/multipass/blob/74e6e1c8c0bf673195d4cf720991a554a64d5b8b/tests/lxd/test_lxd_backend.cpp#L22
is not really necessary anymore since it's now included in `mock_network_access_manager.h`.

Will approve and let you decide if you want to change that or just merge.

townsend2010:
bors d+

ricab:
Thank you @townsend2010. I think the best approach for includes is *include what you use*. Borrowing the explanation from https://include-what-you-use.org/ : 

*"Include what you use" means this: for every symbol (type, function variable, or macro) that you use in foo.cc, either foo.cc or foo.h should #include a .h file that exports the declaration of that symbol.*

So, we assume a coupling of `foo.{h,cpp}` and rely on includes there, but not on other indirect includes. The way things are with C++, there is no way to say if includes in a header are meant as interface or implementation. So the best is not to depend on them being interface. In practice, this means that if `mock_network_access_manager.h` stops including `mock_local_socket_reply.h`, `test_lxd_backend.cpp` does not have to change.

ricab:
OK, so I am going ahead and merging. Let me know if you want to discuss or revisit our approach with includes.

bors r=townsend2010

PR #1928, opened 2021-01-18
Rework disk space check. Fixes #1926

### LINKED ISSUE #1926: Snapcraft using multipass from latest/edge results in "below requested/default size" error
Hi,

Building snap using snapcraft with multipass from latest/edge (latest/edge: 1.7.0-dev.328+g8823c927)

results in

```
$ snapcraft
Launching a VM.
Build environment is in unknown state, cleaning first.
info failed: The following errors occurred:
instance "snapcraft-ubuntu-package-status" does not exist
launch failed: Available disk (221285634048 bytes) below requested/default size (274877906944 bytes)
An error occurred with the instance when trying to launch with 'multipass': returned exit code 2.
```

Reverting to latest/stable

```
sudo snap refresh multipass --channel=latest/stable
```

Allows snapcraft to proceed. I spoke to saviq on insternal chat who suspects this is due to the new checks that the requested space is available.

Saviq:
Allow over-committing disk space, as that's a feature, not a bug. Still fail on disk space lower than image size, and warn on over-committing.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -789,7 +789,8 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
     if (available_disk_space < disk_space)
     {
         mpl::log(mpl::Level::warning, category,
-                 fmt::format("Reserving more disk space than available ({} bytes)", available_disk_space.in_bytes()));
+                 fmt::format("Reserving more disk space ({} bytes) than available ({} bytes)", disk_space.in_bytes(),
+                             available_disk_space.in_bytes()));
     }
 
     return disk_space;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -788,8 +788,8 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
 
     if (available_disk_space < disk_space)
     {
-        throw std::runtime_error(fmt::format("Available disk ({} bytes) below requested/default size ({} bytes)",
-                                             available_disk_space.in_bytes(), disk_space.in_bytes()));
+        mpl::log(mpl::Level::warning, category,
+                 fmt::format("Reserving more disk space than available ({} bytes)", available_disk_space.in_bytes()));
     }
 
     return disk_space;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -780,6 +780,12 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
     std::string available_bytes_str = QString::number(available_bytes).toStdString();
     auto available_disk_space = mp::MemorySize(available_bytes_str + "B");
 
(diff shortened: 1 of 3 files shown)


src/daemon/daemon.cpp
```diff
@@ -788,8 +788,8 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
 
     if (available_disk_space < disk_space)
     {
-        throw std::runtime_error(fmt::format("Available disk ({} bytes) below requested/default size ({} bytes)",
-                                             available_disk_space.in_bytes(), disk_space.in_bytes()));
+        mpl::log(mpl::Level::warning, category,
+                 fmt::format("Reserving more disk space than available ({} bytes)", available_disk_space.in_bytes()));
```
ricab, line 792: When the user did not request a size explicitly, they get the default and may not know how much that is. It could be useful to include it in the message, to gauge how much more space is needed.

ricab (approved):
Alright, I reviewed your commits and they look good. I have a suggestion regarding the log message, but I'll leave it up to you. Delegating so that you can merge if you approve of my part.

bors delegate+

ricab:
\o/

src/daemon/daemon.cpp
Saviq: Done.

ricab (approved):
While I would like to see `launches_with_correct_disk_size` split into separate unit tests, exercising different conditions, that test comes from before. This PR makes it more complex insomuch as the if-else chain gets more complex in the code being tested. Ideally, the if-else chain wouldn't be reflected in the test, but I think that can be regarded as a separate task. 

So, approving again and keeping the earlier bors delegation.

ricab (approved):
The tests look neat now. Curiosity: why did you go for 0 and 1 bytes?

Note: I tried this with `launch -v` to see if I got the warning and it doesn't show. It's in the log, but not going over to the client. I thought we'd fixed that, but apparently not (I checked in `--stable` and no added msgs with `-vvvv` (on `launch`, but it does show on `stop`))

tests/test_daemon.cpp
```diff
@@ -86,7 +89,7 @@ void PrintTo(const YAML::Node& node, ::std::ostream* os)
 
 namespace
 {
-const qint64 default_total_bytes{16'106'127'360}; // 16G
+const qint64 default_total_bytes{16'106'127'360}; // 15G
```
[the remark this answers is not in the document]
Saviq: > Curiosity: why did you go for 0 and 1 bytes?

They're just as arbitrary as anything else… didn't think it made sense to come up with any other values.

Saviq:
> Curiosity: why did you go for 0 and 1 bytes?

They're just as arbitrary as anything else… didn't think it made sense to come up with any other values.

Saviq (commented):
ACK, then!

bors r=ricab,Saviq

PR #1931, opened 2021-01-20
[LXD] Support HTTP & file based launches

townsend2010:


Saviq (changes requested):
Couldn't get a local image to launch:

```plain
$ multipass launch file://...
launch failed: failed to open file for writing during cloud-init generation
```

[launch.log](https://github.com/canonical/multipass/files/5983041/launch.log)

townsend2010:
> Couldn't get a local image to launch:

I've no idea what changed, but this worked when I put the PR up.  I wonder if something with the bridged networking is not playing nice.

townsend2010:
@Saviq,

Could you please try this again?  I rebased on the prerequisite branch, installed the snap that ci built, and then tried it and it worked.

townsend2010:
BTW, the macOS & Windows builds will fail until the `refactor-image-host-and-vault` branches are merged into their respective main branches.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -207,18 +207,46 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
+    }
+    else
+    {
+        QUrl image_url(QString::fromStdString(query.release));
+        QDateTime last_modified;
+
+        if (query.query_type == Query::Type::HttpDownload)
+        {
+            // Generate a sha256 hash based on the URL and use that for the id
+            id = QString(QCryptographicHash::hash(query.release.c_str(), QCryptographicHash::Sha256).toHex());
```
Saviq, line 232: Will LXD be OK with the id being different than a hash of the image itself?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -207,18 +207,46 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
+            if (!QFile::exists(image_url.path()))
+                throw std::runtime_error(fmt::format("Custom image `{}` does not exist.", image_url.path()));
 
-    source_image.id = id.toStdString();
-    source_image.original_release = info.release_title.toStdString();
-    source_image.release_date = info.version.toStdString();
-    source_image.aliases = copy_aliases(info.aliases);
+            source_image.image_path = image_url.path();
+            id = mp::vault::compute_image_hash(source_image.image_path);
+            last_modified = QDateTime::currentDateTime();
```
Saviq, line 242: So we're hashing the whole file? Is there value in that, vs doing what you do for http above?

Shouldn't this be file modification time? Or would it prevent us from re-fetching the image?

src/utils/vm_image_vault_utils.cpp
```diff
@@ -32,6 +33,21 @@ QString mp::vault::filename_for(const mp::Path& path)
     return file_info.fileName();
 }
 
+QString mp::vault::copy(const QString& file_name, const QDir& output_dir)
+{
+    if (file_name.isEmpty())
+        return {};
```
Saviq, line 39: Shouldn't this be an error? What's the use case for calling copy with an empty source? :man_shrugging:

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -237,12 +265,20 @@ mp::VMImage mp::LXDVMImageVault::fetch_image(const FetchType& fetch_type, const
...
+            {
+                // TODO: Need to make this async like in DefaultVMImageVault
+                image_path = lxd_import_dir.filePath(mp::vault::filename_for(info.image_location));
 
-            url_download_image(info, image_path, monitor);
+                url_download_image(info, image_path, monitor);
+            }
+            else
+            {
+                image_path = mp::vault::copy(source_image.image_path, lxd_import_dir.path());
```
Saviq, line 280: Couldn't we avoid this? And just use the original file?

Saviq (changes requested):

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: Yes, LXD generates it's own hash when an image is imported into it- we just create this hash for our internal use and is mapped to the LXD hash once we get that back when it imported.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: What I'm trying to accomplish here (and [here](https://github.com/canonical/multipass/blob/2fc3442ad1f4d622040cde34733969006b02cc26/src/daemon/default_vm_image_vault.cpp#L309)) is that we know the hash of the image itself so in the case that someone moved the image file on their disk, we would still know it's the same image and thus not do another import in the image vault.

As far as `last_modified`, this is really only used for the `release_date` field in `VMImageInfo` and is unused at this time.

src/utils/vm_image_vault_utils.cpp
townsend2010: This is actually a wholesale copy of https://github.com/canonical/multipass/blob/main/src/daemon/default_vm_image_vault.cpp#L172-L185 and I'll have a follow up PR to remove it from that file.

That said, yeah, I suppose it should throw here, but perhaps taking care of all of that should be in the follow up PR.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
townsend2010: I suppose we could, but I have two reasons for doing it this way.
1. One is if the file is a compressed `xz` file, the `post_process_downloaded_image()` function takes care of uncompressing it and will delete the original file.
2. Following on `1.`, it allows commonizing most of the logic between both HTTP & file launches.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: Right. I remember now.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
Saviq: I suppose it's something we should make common for all the backends - decompress / image conversion / copy should be NOOP if unnecessary.

tests/lxd/test_lxd_image_vault.cpp
```diff
@@ -930,3 +914,142 @@ TEST_F(LXDImageVault, minimum_image_size_throws_when_not_found)
                          Property(&std::runtime_error::what,
                                   AllOf(HasSubstr(fmt::format("Cannot retrieve info for image with id \'{}\'", id)))));
 }
+
+TEST_F(LXDImageVault, http_based_image_downloads_and_creates_correct_upload)
+{
+    const std::string content{"This is a fake image!"};
+    mpt::TrackingURLDownloader url_downloader{content};
```
Saviq, line 921: Nothing to `EXPECT` on the downloader?

townsend2010: Yeah, I think you're right since we do want to verify the downloader was called.  I'll add some expectation.

Saviq (approved):
And yup.

bors merge

PR #1932, opened 2021-01-20
[default_image_vault] Use new common copy util

townsend2010:
---

There was a question in https://github.com/canonical/multipass/pull/1931#discussion_r595008464 about whether returning an empty path in `vault::copy()` is the right thing to do.  This is done because of https://github.com/canonical/multipass/blob/afbea256e07cde725f32807adb11fb08782e1b97/src/daemon/default_vm_image_vault.cpp#L656-L657 where there may not be a kernel and initrd to copy.  It basically makes it a no-op on platforms that don't need that and is the easiest way to handle all cases.

### DIFF
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -169,21 +169,6 @@ std::unordered_map<std::string, mp::VaultRecord> load_db(const QString& db_name)
     return reconstructed_records;
 }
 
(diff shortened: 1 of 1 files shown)


luis4a0 (approved):
Yeah, looks like the correct thing to do. Thanks @townsend2010!

luis4a0:
bors retry

luis4a0:
bors retry

PR #1940, opened 2021-01-26
[local_socket_reply] Handle multiple chunked transfers

### LINKED ISSUE #1938: [lxd] Some LXD reply data is cut off
As seen in #1927, the reply data from LXD is not being fully read, leading to "illegal" JSON.  We need to see if there is more data to be read in the reply before moving on.

townsend2010:
Fixes #1938

### DIFF
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -93,7 +93,7 @@ QList<QByteArray> split_http_lines(const QByteArray& reply_data)
         start = end + 2;
     }
 
-    list.append(reply_data.mid(start));
+    list.append(reply_data.mid(start).split('\n').front());
 
     return list;
 }
@@ -187,6 +187,10 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
         http_data += "User-Agent: " + user_agent + "\r\n";
     }
 
+    // Workaround weird issue in LXD's use of Go where Go's HTTP handler thinks there is more
+    // data and "sees" and "empty" query and reponds with an unexpected 400 error
+    http_data += "Connection: close\r\n";
+
     if (!local_socket_write(http_data))
         return;
 
@@ -270,7 +274,7 @@ void mp::LocalSocketReply::read_reply()
 
     do
     {
-        bytes_read = local_socket->read(reply_data.data(), len);
+        bytes_read = local_socket->read(data_ptr, len);
 
         data_ptr += bytes_read;
     } while (bytes_read > 0);
@@ -286,8 +290,6 @@ void mp::LocalSocketReply::parse_reply()
     bool chunked_transfer_encoding{false};
     auto reply_lines = split_http_lines(reply_data);
     auto it = reply_lines.constBegin();
-
-    content_data = "\0";
 
     parse_status(*it);
 
@@ -312,26 +314,28 @@ void mp::LocalSocketReply::parse_reply()
 
             if (chunked_transfer_encoding)
             {
+                QByteArray chunked_data;
+
                 while (true)
                 {
                     data_len = (*it).toLongLong(&ok, 16);
 
                     if (data_len == 0)
                     {
+                        content_data = chunked_data;
                         return;
                     }
 
                     ++it;
 
-                    content_data.append((*it).constData(), data_len);
+                    chunked_data.append((*it).constData(), data_len);
 
                     ++it;
                 }
             }
             else
             {
-                content_data.append((*it).constData(), data_len);
-
+                content_data = (*it).trimmed();
                 return;
             }
         }
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Canonical, Ltd.
+ * Copyright (C) 2020-2021 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


src/network/local_socket_reply.cpp
```diff
@@ -80,6 +80,23 @@ QNetworkReply::NetworkError statusCodeFromHttp(int httpStatusCode)
...
+    while ((end = reply_data.indexOf("\r\n", start)) != -1)
+    {
+        list.append(reply_data.mid(start, end - start));
+        start = end + 2;
+    }
+
+    list.append(reply_data.mid(start).split('\n').front());
+
+    return list;
+}
```
Saviq, line 99: Why not [`QByteArray::split()`](https://doc.qt.io/qt-5/qbytearray.html#split)?

src/network/local_socket_reply.cpp
```diff
@@ -170,6 +187,10 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
         http_data += "User-Agent: " + user_agent + "\r\n";
     }
 
+    // Workaround weird issue in LXD's use of Go where Go's HTTP handler thinks there is more
+    // data and "sees" and "empty" query and reponds with an unexpected 400 error
+    http_data += "Connection: close\r\n";
```
Saviq, line 192: Is there a LXD bug we could reference?

src/network/local_socket_reply.cpp
```diff
@@ -266,26 +287,57 @@ void mp::LocalSocketReply::read_reply()
...
 
-            break;
+                    chunked_data.append((*it).constData(), data_len);
+
+                    ++it;
+                }
+            }
+            else
+            {
+                content_data = (*it).trimmed();
```
Saviq, line 338: Not using `data_len` here? You read it from the `Content-Length` header above?

Saviq (changes requested):

Saviq:
Another thing, shouldn't this work?

```diff
diff --git a/tests/linux/test_local_network_access_manager.cpp b/tests/linux/test_local_network_access_manager.cpp
index aad9cf2a..44d0b467 100644
--- a/tests/linux/test_local_network_access_manager.cpp
+++ b/tests/linux/test_local_network_access_manager.cpp
@@ -222,6 +222,30 @@ TEST_F(LocalNetworkAccessManager, reads_expected_data_multi_chunked)
     EXPECT_EQ(data, "This is chunked data.");
 }
 
+TEST_F(LocalNetworkAccessManager, reads_expected_data_multi_chunked_newlines)
+{
+    QByteArray http_response;
+    http_response += "HTTP/1.1 200 OK\r\n";
+    http_response += "Transfer-Encoding: chunked\r\n";
+    http_response += "\r\n";
+    http_response += "a\r\n";
+    http_response += "This is \r\n\r\n";
+    http_response += "d\r\n";
+    http_response += "chunked data.\r\n";
+    http_response += "0\r\n\r\n";
+
+    auto server_response = [&http_response](auto...) { return http_response; };
+    test_server.local_socket_server_handler(server_response);
+
+    auto reply = handle_request(base_url, "GET");
+
+    ASSERT_EQ(reply->error(), QNetworkReply::NoError);
+
+    auto data = reply->readAll();
+
+    EXPECT_EQ(data, "This is \r\nchunked data.");
+}
+
 TEST_F(LocalNetworkAccessManager, client_posts_correct_data)
 {
     QByteArray expected_data{"POST /1.0 HTTP/1.1\r\n"
```

src/network/local_socket_reply.cpp
townsend2010: @stgraber dug in to this and found it's not a LXD bug, but either a bug in Go or some weird expected behavior in it.  I don't know if he opened a bug anywhere for this and felt it wasn't necessary to follow up on either since we have a way to avoid it.

src/network/local_socket_reply.cpp
townsend2010: Because we need to split on `\r\n` first and `QByteArray::split()` can only split on a single character.

townsend2010:
> Another thing, shouldn't this work?

Not sure what you mean...the test does work AFAIK.

Saviq:
> Not sure what you mean...the test does work AFAIK.

That's not your test. I modified it to have `\r\n` in the content.

src/network/local_socket_reply.cpp
Saviq: Oh indeed. I looked at `QString::split()` first and it had `QString sep`, and no mention that it's meant to be a single character.

But yeah, `QByteArray::split()` has `char sep`…

townsend2010:
> That's not your test. I modified it to have \r\n in the content.

Oh....hmm, yeah, that should work and I believe the splitting of the HTTP lines is broken.

src/network/local_socket_reply.cpp
townsend2010: Hmm, I suppose instead of just taking the content wholesale, it would be safer (and more correct) to only copy the length specified in the `Content-Length` header.

townsend2010: Again, this is a Good Enough™ type of thing in that it's assumed the rest of the body is the content data.  I'll add a comment here explaining that.

townsend2010:
> That's not your test. I modified it to have \r\n in the content.

As discussed in chat, this is just for communicating with LXD and not for general purpose.  It's highly unlikely that LXD will send a `\r\n` in the actual content data, so what is already here should work.

Saviq (approved):
Yupsies.

bors merge

PR #1942, opened 2021-01-26
Permit empty default MAC on startup

luis4a0:
When introducing bridging, we altered the behavior of reading the MAC addresses. Before bridging, empty MAC addresses were allowed and the daemon generated a new MAC. When bridging, since we hardcode MAC's on the network cloud-init, we disallowed empty MAC's in the JSON database.

In this PR, we are enabling having empty default MAC addresses when there are no bridged interfaces in the instance, in order to mimic the old startup.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -263,7 +263,7 @@ std::vector<mp::NetworkInterface> read_extra_interfaces(const QJsonObject& recor
             auto mac_address = entry.toObject()["mac_address"].toString().toStdString();
             if (!mpu::valid_mac_address(mac_address))
             {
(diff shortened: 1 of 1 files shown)


Saviq:
> we can safely generate a new MAC in the future.

That's actually not true, as cloud-init will encode the MAC as given on first boot:
```yaml
```

To have this work, we'd need to trigger cloud-init to run again, by, for example, changing the instance id we put in cloud-init metadata.

Saviq:
We now realized what the issue is, and decided this is not how we should handle this. An empty MAC does indeed suggest a bigger problem, so bailing is a better choice than writing over stuff.

PR #1946, opened 2021-01-28
Add `--force` argument to the `stop` command

### LINKED ISSUE #1909: Add a `--force` option to `multipass stop` and `restart`
When something unexpected happens and we lose the ability to control the VM, it may be necessary to forcefully stop the instance by terminating, and ultimately killing the instance.

Also a `multipass delete --purge` could benefit from this as there's no point in waiting for the instance to gracefully shut down when we're deleting it right after.

### LINKED ISSUE #2492: Multipass can't delete LXD instances that go into Error state
**Describe the bug**
When using the LXD backend, after a failed launch where the instance gets into an error state (112 from LXD, translated to unknown in multipass), multipass is unable to delete the instance:

```
$ multipass delete --purge --all
[2022-03-24T17:29:10.900] [error] [lxd request] Operation completed with error: (400) The instance cannot be cleanly shutdown as in Error status
delete failed: Operation completed with error: (400) The instance cannot be cleanly shutdown as in Error status
```

We would need to give LXD the equivalent of the `--force` flag in such cases. 

**To Reproduce**
Because of https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1935880 , a pretty reliable way to get a LXD vm in error state is just
1. `multipass launch --cpus 2`
    - Until the remainder of #2250 is fixed, this will loop until timeout, leaving the instance in unknown state
1. `multipass delete --purge <instance>` then fails as above

**Expected behavior**
Multipass would be able to delete the instance.

luis4a0:
Add a new parameter to force the shutdown instances (i.e., power down them).

Fixes #1909, Fixes #2492

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -63,7 +63,7 @@ public:
 
     virtual ~VirtualMachine() = default;
     virtual void start() = 0;
-    virtual void shutdown(bool force = false) = 0;
+    virtual void shutdown(const bool force = false) = 0;
     virtual void suspend() = 0;
     virtual State current_state() = 0;
     virtual int ssh_port() = 0;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -63,7 +63,7 @@ public:
 
     virtual ~VirtualMachine() = default;
     virtual void start() = 0;
-    virtual void shutdown() = 0;
+    virtual void shutdown(bool force = false) = 0;
     virtual void suspend() = 0;
     virtual State current_state() = 0;
     virtual int ssh_port() = 0;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -63,7 +63,7 @@ public:
 
     virtual ~VirtualMachine() = default;
     virtual void start() = 0;
-    virtual void shutdown(const bool force = false) = 0;
+    virtual void shutdown(bool force = false) = 0;
     virtual void suspend() = 0;
     virtual State current_state() = 0;
     virtual int ssh_port() = 0;
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -84,7 +84,7 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    QCommandLineOption force_option({"f", "force"}, "Force switching the instance off");
+    QCommandLineOption force_option("force", "Force switching the instance off");
     parser->addOptions({all_option, time_option, cancel_option, force_option});
 
     auto status = parser->commandParse(this);
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -84,7 +84,8 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    parser->addOptions({all_option, time_option, cancel_option});
+    QCommandLineOption force_option({"f", "force"}, "Force switching the instance off");
+    parser->addOptions({all_option, time_option, cancel_option, force_option});
 
     auto status = parser->commandParse(this);
     if (status != ParseCode::Ok)
@@ -102,6 +103,12 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     if (parser->isSet(time_option) && parser->isSet(cancel_option))
     {
         cerr << "Cannot set \'time\' and \'cancel\' options at the same time\n";
+        return ParseCode::CommandLineError;
+    }
+
+    if (parser->isSet(force_option) && (parser->isSet(time_option) || parser->isSet(cancel_option)))
+    {
+        cerr << "Cannot set \'force\' along with \'time\' or \'cancel\' options at the same time\n";
         return ParseCode::CommandLineError;
     }
 
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -112,6 +112,8 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
         return ParseCode::CommandLineError;
     }
 
+    request.set_force_stop(parser->isSet(force_option));
+
     auto time = parser->value(time_option);
 
     if (time.startsWith("+"))
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -84,7 +84,10 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    QCommandLineOption force_option("force", "Force switching the instance off");
+    QCommandLineOption force_option(
+        "force",
+        "Immediately force switching the instance off. This could potentially corrupt a running"
+        " instance, so use with care.");
     parser->addOptions({all_option, time_option, cancel_option, force_option});
 
     auto status = parser->commandParse(this);
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -62,8 +62,8 @@ QString cmd::Stop::short_help() const
 
 QString cmd::Stop::description() const
 {
-    return QStringLiteral("Stop the named instances, if running. Exits with\n"
-                          "return code 0 if successful.");
+    return QStringLiteral("Stop the named instances. Exits with return code 0 \n"
+                          "if successful.");
 }
 
 mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -84,10 +84,9 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    QCommandLineOption force_option(
-        "force",
-        "Immediately force switching the instance off. This could potentially corrupt a running"
-        " instance, so use with care.");
+    QCommandLineOption force_option("force",
+                                    "Force the instance to shut down immediately. Warning: This could potentially "
+                                    "corrupt a running instance, so use with caution.");
     parser->addOptions({all_option, time_option, cancel_option, force_option});
 
     auto status = parser->commandParse(this);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2134,7 +2134,10 @@ try // clang-format on
         if (request->cancel_shutdown())
             operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
         else if (request->force_stop())
-            operation = std::bind(&Daemon::switch_off_vm, this, std::placeholders::_1);
+        {
+            auto adapted_switch_off_vm = [this](auto&& arg) -> grpc::Status { return this->switch_off_vm(arg); };
+            operation = adapted_switch_off_vm;
+        }
         else
             operation = std::bind(&Daemon::shutdown_vm, this, std::placeholders::_1,
                                   std::chrono::minutes(request->time_minutes()));
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3121,12 +3121,14 @@ grpc::Status mp::Daemon::reboot_vm(VirtualMachine& vm)
 grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::milliseconds delay)
 {
     const auto& name = vm.vm_name;
-    const auto& state = vm.current_state();
+    const auto& current_state = vm.current_state();
 
     using St = VirtualMachine::State;
     const auto skip_states = {St::off, St::stopped, St::suspended};
 
-    if (std::none_of(cbegin(skip_states), cend(skip_states), [&state](const auto& st) { return state == st; }))
+    if (std::none_of(cbegin(skip_states), cend(skip_states), [&current_state](const auto& skip_state) {
+            return current_state == skip_state;
+        }))
     {
         delayed_shutdown_instances.erase(name);
 
@@ -3148,12 +3150,14 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
 grpc::Status mp::Daemon::switch_off_vm(VirtualMachine& vm)
 {
     const auto& name = vm.vm_name;
-    const auto& state = vm.current_state();
+    const auto& current_state = vm.current_state();
 
     using St = VirtualMachine::State;
     const auto skip_states = {St::off, St::stopped};
 
-    if (std::none_of(cbegin(skip_states), cend(skip_states), [&state](const auto& st) { return state == st; }))
+    if (std::none_of(cbegin(skip_states), cend(skip_states), [&current_state](const auto& skip_state) {
+            return current_state == skip_state;
+        }))
     {
         delayed_shutdown_instances.erase(name);
 
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2133,6 +2133,8 @@ try // clang-format on
         std::function<grpc::Status(VirtualMachine&)> operation;
         if (request->cancel_shutdown())
             operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
+        else if (request->force_stop())
+            operation = std::bind(&Daemon::switch_off_vm, this, std::placeholders::_1);
         else
             operation = std::bind(&Daemon::shutdown_vm, this, std::placeholders::_1,
                                   std::chrono::minutes(request->time_minutes()));
@@ -3071,7 +3073,7 @@ bool mp::Daemon::delete_vm(InstanceTable::iterator vm_it, bool purge, DeleteRepl
             delayed_shutdown_instances.erase(name);
 
         mounts[name].clear();
-        instance->shutdown();
+        instance->shutdown(purge);
 
         if (!purge)
         {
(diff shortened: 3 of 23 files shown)


townsend2010:
Hey @luis4a0,

I know this is still in draft, but I always envisioned just adding a `bool force` flag to `shutdown()` and `stop()` (it also can be argued that either `shutdown()` or `stop()` can be dropped altogether, but that's for a different day).  That way, you are not having to duplicate any special handling or logic that you are currently doing if some of the `force_shutdown()` implementations.

luis4a0:
Hey @townsend2010, thanks a lot for your comment! What you propose makes complete sense, I'll finish testing if the commands I used indeed work to turn off VM's and then rework the code.

sharder996:
Added a few small changes; mostly refactoring with the current state of `main` and some wiring that got missed.

Fixes #2492 

and 

#2693 can be closed as its no longer needed

ricab:
Once this is in, I think we should also change `delete` such that running instances get force-stopped if a regular `stop` is not enough. I don't think it makes sense for deleted instances to keep running...

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -343,9 +343,24 @@ void mp::QemuVirtualMachine::start()
...
 }
 
-void mp::QemuVirtualMachine::shutdown()
+void mp::QemuVirtualMachine::shutdown(const bool force)
 {
-    if (state == State::suspended)
+    if (force)
+    {
+        state = State::off;
+        mpl::log(mpl::Level::info, vm_name, "Forced shutdown");
```
ricab, line 351: Could this say instead

```suggestion
        mpl::log(mpl::Level::info, vm_name, "Forcing shutdown");
```

Just to be clear that the shutdown did not occur just yet. I think it would make more sense given the order of the logs, and just in case the daemon crashes when attempting to kill (or if there is an apparmor denial or something).

src/client/cli/cmd/stop.cpp
```diff
@@ -84,7 +84,8 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    parser->addOptions({all_option, time_option, cancel_option});
+    QCommandLineOption force_option("force", "Force switching the instance off");
```
ricab, line 87: I would prefer if we added a little more explanation to let users know that this equates to pulling the plug on the machine and that may result in corruption of the VM.

ricab (commented):
Hey @townsend2010, thanks for the PR! Not a proper review, but I was looking in order to base my work on top, so leaving a couple of comments.

src/client/cli/cmd/stop.cpp
townsend2010: Agreed, will modify the wording here.  Thanks!

src/daemon/daemon.cpp
```diff
@@ -3183,6 +3185,26 @@ grpc::Status mp::Daemon::shutdown_vm(VirtualMachine& vm, const std::chrono::mill
     return grpc::Status::OK;
 }
 
+grpc::Status mp::Daemon::switch_off_vm(VirtualMachine& vm)
+{
+    const auto& name = vm.vm_name;
+    const auto& state = vm.current_state();
+
+    using St = VirtualMachine::State;
+    const auto skip_states = {St::off, St::stopped};
```
ricab, line 3194: Shouldn't this skip `suspended` too?

ricab: No, that makes total sense to me, I think that's the best. The code isn't doing that yet though, is it?

townsend2010: > The code isn't doing that yet though, is it?

It does on the `hyperv` backend automatically, but not on the `qemu` backend.  I still have to confirm the behavior on the `virtualbox` and `libvirt` backends.

townsend2010: I should add it's still on my TODO to implement the suspend image deletion on the backends that support suspend and it doesn't work yet.

ricab: OK, makes sense, thanks!

src/client/cli/cmd/stop.cpp
townsend2010: Ok, I modified this.  Please let me know what you think.  Thanks!

ricab: Yep, much better. Thank you!

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
andrei-toterman, line 1: libvirt does not remove suspended state on force stop

andrei-toterman (changes requested):
Okay, the code looks pretty good IMO. It went through many iterations already after all.
In terms of functionality:
- The main issue is QEMU deadlocking.
- The other smaller issue is libvirt not removing the suspended state on force stop. But given that we're talking more and more about removing libvirt, I'm not sure if it's worth spending too much time on this issue.

georgeliao:
@andrei-toterman 
The comments are addressed. Feel free to give an another scan. I also have tested on qemu and lxd, they both worked fine. I did not touch libvirt at all based on the decision of our team discussion.

include/multipass/virtual_machine.h
```diff
@@ -63,7 +63,7 @@ class VirtualMachine : private DisabledCopyMove
 
     virtual ~VirtualMachine() = default;
     virtual void start() = 0;
-    virtual void shutdown() = 0;
+    virtual void shutdown(const bool force = false) = 0;
```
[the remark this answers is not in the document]
ricab: (Here and elsewhere)

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -343,31 +344,57 @@ void mp::QemuVirtualMachine::start()
...
-        if ((state == State::running || state == State::delayed_shutdown || state == State::unknown) && vm_process &&
-            vm_process->running())
+        if (vm_process)
         {
-            vm_process->write(qmp_execute_json("system_powerdown"));
-            vm_process->wait_for_finished(timeout);
+            mpl::log(mpl::Level::info, vm_name, "Killing process");
+            lock.unlock();
+            vm_process->kill();
+            if (vm_process != nullptr && !vm_process->wait_for_finished(timeout))
```
ricab, line 370: I don't think the same timeout should apply here as in normal shutdown/suspend situations. Note that this is intentionally high (5 min) to give QEMU a long time to quit properly and avoid corrupting the image. But here the intention is to exit immediately regardless of image damage. 

`QProcess::kill()` (which is what the above ends up calling) tries to send a SIGKILL. If the process does not quit immediately, it is unlikely to quit anytime. I don't think there is any point in waiting for any longer than 5s, just in case the machine is very busy/slow.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -343,31 +344,57 @@ void mp::QemuVirtualMachine::start()
...
+            {
+                mpl::log(mpl::Level::error, vm_name, "Killing qemu process could not finish.");
```
ricab, line 372: Hmm, why a log? This is an error and should I think it should fail the request, so I would expect this to throw an exception, as usual.

Also, could the message say instead "The QEMU process did not finish within x seconds"?

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -407,9 +407,13 @@ const QUrl mp::LXDVirtualMachine::network_leases_url()
     return base_url.toString() + "/networks/" + bridge_name + "/leases";
 }
 
-void mp::LXDVirtualMachine::request_state(const QString& new_state)
+void mp::LXDVirtualMachine::request_state(const QString& new_state, const QJsonObject args)
```
ricab, line 410: Couldn't the new param be a const reference instead? Otherwise, move contents in the implementation perhaps? Not sure if that's supported in (our version of) QJson stuff.

src/platform/backends/lxd/lxd_virtual_machine.h
```diff
@@ -85,7 +86,7 @@ class LXDVirtualMachine : public BaseVirtualMachine
     const QUrl url() const;
     const QUrl state_url();
     const QUrl network_leases_url();
-    void request_state(const QString& new_state);
+    void request_state(const QString& new_state, const QJsonObject args = {});
```
ricab, line 89: Idem: top-level const

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
```diff
@@ -345,41 +346,50 @@ void mp::LibVirtVirtualMachine::start()
...
-                    fmt::format("Cannot shutdown '{}': {}", vm_name, libvirt_wrapper->virGetLastErrorMessage())};
-                mpl::log(mpl::Level::warning, vm_name, warning_string);
-                throw std::runtime_error(warning_string);
-            }
-
-            state = State::off;
-            update_state();
-        }
-        else if (state == State::starting)
+    if (force)
```
ricab, line 366:
```suggestion
    if (force) // TODO delete suspension state if it exists
```

ricab (changes requested):
Alright, this not only implements a new feature, it also improves the whole stopping scheme, so thank you to all the participants in the most transferred PR in Multipass's history :smile_cat: 

@georgeliao, I have a few requests but nothing major and they shouldn't be hard. Good job!

src/client/cli/cmd/stop.cpp
```diff
@@ -84,7 +84,11 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
     QCommandLineOption time_option({"t", "time"}, "Time from now, in minutes, to delay shutdown of the instance",
                                    "time", "0");
     QCommandLineOption cancel_option({"c", "cancel"}, "Cancel a pending delayed shutdown");
-    parser->addOptions({all_option, time_option, cancel_option});
+    QCommandLineOption force_option(
+        "force",
+        "Immediately force switching the instance off. This could potentially corrupt a running"
```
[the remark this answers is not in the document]
georgeliao: Our [documentation](https://multipass.run/docs/use-an-instance#heading--Stop-an-instance-forcefully) says more about this. 
Force stop is used when the state is non-responsive, unknown or suspended. Do we just want to mentioned suspended state here?

include/multipass/virtual_machine.h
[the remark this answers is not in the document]
georgeliao: Actually, I was aware of the const value on the function declaration thing. Resharper on visual studio also suggests what clang-tidy suggests. 

However, in practice, that introduces more function parameter list inconsistency between the function declaration and definition. It also does not signal the reader that the function body does not modify the variable by looking at the signature. At last, if we want to apply this, we also should try to do it throughout the whole code base. 

What do you think?

src/client/cli/cmd/stop.cpp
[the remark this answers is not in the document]
georgeliao: BTw, now the new help text is `Force the instance to shut down immediately. Warning: This could potentially corrupt a running instance, so use with caution.` after the iteration with technical author.

include/multipass/virtual_machine.h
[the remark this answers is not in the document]
ricab: Hmm, I am not sure I follow. Let's discuss this in the sync?

src/daemon/daemon.cpp
```diff
@@ -2133,6 +2133,11 @@ try // clang-format on
         std::function<grpc::Status(VirtualMachine&)> operation;
         if (request->cancel_shutdown())
             operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
+        else if (request->force_stop())
+        {
+            auto adapted_switch_off_vm = [this](auto&& arg) -> grpc::Status { return this->switch_off_vm(arg); };
```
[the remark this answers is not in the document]
georgeliao: Well, both `std::bind` and lambda adapt the function interface. I would say lambda is definitely a preferred option due to C++ modernity, readability, and performance. 

About performance, https://www.youtube.com/watch?v=uzF4u9KgUWI&t=3121s 45:35 also gave some benchmark results on the compile time and run time cost of `std::bind`. 

About the universal reference parameter. Well, that `(auto&& arg)` is just a general idiom for using lambda as an adapter.

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
georgeliao: About the libvirt code, do we still want to bother to add the TODO given the fact that we will remove libvirt code soon?

src/platform/backends/lxd/lxd_virtual_machine.cpp
georgeliao: Nice catch, I overlooked this when I took over the code. The calling code only passed temporary objects into the `request_state` function. Likely, that is why Chris made a const value. But indeed, but general use case, it is better to have it as a const ref, I will change it.

src/platform/backends/qemu/qemu_virtual_machine.cpp
georgeliao: Indeed, 5s is a more reasonable timeout value for this particular case.

src/platform/backends/qemu/qemu_virtual_machine.cpp
georgeliao: Indeed, an exception is more appropriate here. About the text message, what about "The QEMU process **kill** did not finish within x seconds"? Which is more informative.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
ricab: In item 34 of Scott Meyers's book, the argument to prefer lambdas to `std::bind` rests largely on lambdas being "more readable, more expressive, and may be more efficient than using `std::bind`". The argument in that talk is also mainly efficiency (compile-time and runtime) + the appeal to Scott Meyers's authority.

IMO, "more readable" and "more expressive" is unfair in many cases. This is one such example: the bind version is shorter and more obvious. Errata comment "6/15/14 tyk All      Item 34" expresses that idea too. I don't think there was a second edition of this book, but the fact that he mentioned this in there is indication that he would probably express a different view today. When I read the book I had this errata already and so I always took this item with a grain of salt.

As for efficiency, 15% overhead to compile this line plus twice the function-call time is entirely negligible in here. Definitely a price I would pay for the more obvious code.

Note that I am not advocating using bind instead of lambdas everywhere. At all, I default to lambdas myself. But in some situations, `std::bind` is clearer.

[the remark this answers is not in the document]
ricab: > About the universal reference parameter. Well, that (auto&& arg) is just a general idiom for using lambda as an adapter. I think it signals better that lambda is being used as an adapter.

I don't think that general idiom makes sense without a `std::forward`. When you consider adding that, along with consistency with the other if-branches, it becomes clear which one is more readable/expressive. 

I mean, just have a look at the alternatives:

```cpp
    if (status.ok())
    {
        assert(instance_selection.deleted_selection.empty());
        assert(instance_selection.missing_instances.empty());

        std::function<grpc::Status(VirtualMachine&)> operation;
        if (request->cancel_shutdown())
            operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
        else if (request->force_stop())
            operation = std::bind(&Daemon::adapted_switch_off_vm, this, std::placeholders::_1);
        else
            operation = std::bind(&Daemon::shutdown_vm, this, std::placeholders::_1,
                                  std::chrono::minutes(request->time_minutes()));

        status = cmd_vms(instance_selection.operative_selection, operation);
    }
```

or 

```cpp
    if (status.ok())
    {
        assert(instance_selection.deleted_selection.empty());
        assert(instance_selection.missing_instances.empty());

        std::function<grpc::Status(VirtualMachine&)> operation;
        if (request->cancel_shutdown())
            operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
        else if (request->force_stop())
        {
            auto adapted_switch_off_vm = [this](auto&& arg) -> grpc::Status {
                return this->switch_off_vm(std::forward<decltype(arg)>(arg)); 
            };
            operation = adapted_switch_off_vm;
        }
        else
            operation = std::bind(&Daemon::shutdown_vm, this, std::placeholders::_1,
                                  std::chrono::minutes(request->time_minutes()));

        status = cmd_vms(instance_selection.operative_selection, operation);
    }
```

Which of those is clearer?

src/client/cli/cmd/stop.cpp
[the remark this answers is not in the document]
ricab: I would be fine with replacing that version in.

src/platform/backends/qemu/qemu_virtual_machine.cpp
ricab: Hmm, sending the "kill" signal itself finishes, this would be the process (or OS) not reacting on time. How about "The QEMU process did not finish within x seconds after being killed"?

src/platform/backends/libvirt/libvirt_virtual_machine.cpp
ricab: It is not a "fact" yet ;)

src/platform/backends/qemu/qemu_virtual_machine.cpp
georgeliao: This text "The QEMU process did not finish within x seconds after being killed" is fine by me as well.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
georgeliao: One concern of mine is that the run-time and compile-time cost might add up if we continue to use `std::bind`. My opinion here is that the new one should start with lambda. 

You mentioned the verboseness of the lambda syntax, it can be compressed into one-liner easily              `operation = [this](VirtualMachine& vm) { return this->switch_off_vm(vm); };` by removing the trailing return, functor name, and converting the universal reference to the exact parameter `switch_off_vm` needs. 

You also touched upon the consistency of the other two branches, then I think we'd better change the three of them.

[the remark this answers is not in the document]
georgeliao: All in all, I think a fair comparison should be 
```
        std::function<grpc::Status(VirtualMachine&)> operation;
        if (request->cancel_shutdown())
            operation = std::bind(&Daemon::cancel_vm_shutdown, this, std::placeholders::_1);
        else if (request->force_stop())
            operation = std::bind(&Daemon::switch_off_vm, this, std::placeholders::_1);
        else
            operation = std::bind(&Daemon::shutdown_vm, this, std::placeholders::_1,
                                  std::chrono::minutes(request->time_minutes()));
```
vs
```
        std::function<grpc::Status(VirtualMachine&)> operation;
        if (request->cancel_shutdown())
            operation = [this](const VirtualMachine& vm) { return this->cancel_vm_shutdown(vm); };
        else if (request->force_stop())
            operation = [this](VirtualMachine& vm) { return this->switch_off_vm(vm); };
        else
            operation = [this, delay_minutes = std::chrono::minutes(request->time_minutes())](VirtualMachine& vm) {
                return this->shutdown_vm(vm, delay_minutes);
            };
```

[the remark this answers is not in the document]
ricab: Yes, I agree that lambdas should be the default tool. Repeating what I said above *I default to lambdas myself*. It is just that in some cases, bind still makes it more obvious. And I wonder if the penalties are still the same with today's compilers as they were 8+ years ago. Even then, it is probably a matter of an extra pointer dereference...

But OK, the second version you have above is fine. I still find the first one marginally more obvious, personally, but I am perfectly fine with the other one. Let's change all of them then.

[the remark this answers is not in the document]
georgeliao: >  I still find the first one marginally more obvious.
I kind of agree that `std::bind` does make the adapter thing clear if you are familiar with the `std::placeholders` syntax. 

> And I wonder if the penalties are still the same with today's compilers as they were 8+ years ago
It is likely that std::bind must handle various types of argument forwarding and handle placeholders dynamically. The actually runtime cost penalty probably is negligible. Compile-time wise, lambda does not have TMP involved, so there will be some gain on this.

ricab (approved):
LGTM, thanks all!

andrei-toterman (approved):
Alright, everything seems to be working. Thanks, @georgeliao!

georgeliao:
@ricab we might need to manually merge again.

PR #1952, opened 2021-02-02
[daemon] disallow `--network ...,mode=auto` on Ubuntu Core 16

Saviq:
`netplan` in the image is too old to handle our config:

```plain
Stderr: Error in network definition //etc/netplan/50-cloud-init.yaml line 12 column 12: unknown key dhcp4-overrides
```

---

For reference: https://bugs.launchpad.net/netplan/+bug/1759014


luis4a0 (approved):
Yeah, this is the correct thing to do.

townsend2010 (approved):
This is fine for now, but the proper fix is to address #1953.

bors r=luis4a0,townsend2010

PR #1958, opened 2021-02-08
Rename 'id' to 'name' in launch command

luis4a0:
`multipass launch --network` accepts `id=` as the key to the interface name argument. This must be changed to `name=` in order to avoid confusion with the network listing, on which interface names are listed using 'name' as field name.


ricab (approved):
Alright!

bors r+

PR #1960, opened 2021-02-08
Bridge members lxd

ricab:
Include members in bridge descriptions, by implementing bridge inspection on the Linux platform. Example output:

```
$ multipass networks
Name    Type    Description
lxdbr0  bridge  Empty network bridge
mpbr0   bridge  Network bridge for Multipass
mybr0   bridge  Network bridge with enxe4bf7a032432
virbr0  bridge  Network bridge with virbr0-nic
```

### DIFF
--- a/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
@@ -44,7 +44,7 @@ std::string describe_bridge(const std::string& id, const QString& lxd_descriptio
 
     if (!lxd_description.isEmpty())
         return lxd_description.toStdString();
-    else if (auto it = platform_networks.find(id); it != platform_networks.end())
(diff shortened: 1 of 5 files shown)


ricab:
Only bridges are returned by the platform for the time being. My plan is to add inspection for other network types when they are used (as part of the work for automatic bridge creation). But I can also do it here if people prefer.

ricab:
Note that this requires #1963 and corresponding private bits to be merged, which is why it currently fails on the mac.

luis4a0:
> Only bridges are returned by the platform for the time being. My plan is to add inspection for other network types when they are used (as part of the work for automatic bridge creation). But I can also do it here if people prefer.

I think it is better not to do it here. Let's keep things separated for the time being.

luis4a0 (approved):
I like the approach in the linux platform and in the LXD backend of not banning interfaces of certain types, for it will be easy in the future to add new network types for reporting. Moreover, testing looks neat.

luis4a0:
Since #1963 is in, we are all set.

bors r+

PR #1961, opened 2021-02-09
[3rd-party] refresh submodules

Saviq:
---

Better ideas about how to handle libssh welcome :)

Also, grpc is a mess, so not included - @townsend2010 will have a look.


townsend2010 (approved):
Ok, this looks good now.

bors r+

PR #1963, opened 2021-02-10
Platform singleton

ricab:
Create a `Platform` singleton to facilitate mocking. Move `get_network_interfaces_info()` into it. Provide a `MockPlatform` for it.

Note: I kept it in the `multipass::platform` namespace for now, for consistency with the free functions. When/if we move all free functions to the class, we should probably get rid of the `platform` namespace altogether, to reduce verbosity in things like `using multipass::platform::Platform::Platform` :)

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -48,7 +48,7 @@ namespace platform
 class Platform : public Singleton<Platform>
 {
 public:
-    Platform(const Singleton<Platform>::PrivatePass&);
+    Platform(const Singleton::PrivatePass&);
     // Get information on the network interfaces that are seen by the platform, indexed by name
     virtual std::map<std::string, NetworkInterfaceInfo> get_network_interfaces_info() const;
 };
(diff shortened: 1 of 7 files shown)


townsend2010 (commented):
This looks good to me.  I'm not going to block on it, but would you be able to just add a simple test in `tests/linux/test_platform_linux.cpp` that calls `get_network_interfaces_info()` and expects that exception type?  It will help with coverage and should be easy to add.  I'll see what you say before doing the final approval.

ricab:
Sure, I can do that. It will be removed in #1960, but that way this PR is fully independent.

townsend2010:
Oh, then it doesn't seem to make sense to add that test since that will just be removed in the child PR.

townsend2010 (approved):
I like it!  Thanks!

bors r+

townsend2010:
Running bors again to update `staging`.

bors r+

townsend2010:
The Mac and Windows build will fail, so merging manually.

PR #1965, opened 2021-02-11
Don't ask for IP when querying the daemon from the GUI

luis4a0:
The GUI issues one `list` request to the daemon per second. Since now `list` asks the instances for the IP's it's using, this consumes lots of resources.

This PR fixes that, by adding a hidden parameter to the `list` command, telling the daemon whether it should ask the instance for the IP's.


townsend2010 (approved):
Works and LGTM!  Thanks!

bors r+

PR #1969, opened 2021-02-12
Add --no-ipv4 hidden parameter to list command

luis4a0:
So far we added in the RPC the ability to list instances without asking them for the IP. This saves precious time in situations on which the IP is not needed. In this PR, we add an option to the CLI to do this.

### DIFF
--- a/tests/test_cli_client.cpp
+++ b/tests/test_cli_client.cpp
@@ -848,7 +848,7 @@ TEST_F(Client, info_cmd_fails_with_names_and_all)
 // list cli tests
 TEST_F(Client, list_cmd_ok_no_args)
 {
-    EXPECT_CALL(mock_daemon, list(_, _, _));
+    EXPECT_CALL(mock_daemon, list(_, Property(&mp::ListRequest::request_ipv4, IsTrue()), _));
     EXPECT_THAT(send_command({"list"}), Eq(mp::ReturnCode::Ok));
 }
 
@@ -860,6 +860,12 @@ TEST_F(Client, list_cmd_fails_with_args)
 TEST_F(Client, list_cmd_help_ok)
 {
     EXPECT_THAT(send_command({"list", "-h"}), Eq(mp::ReturnCode::Ok));
+}
+
+TEST_F(Client, list_cmd_no_ipv4_ok)
+{
+    EXPECT_CALL(mock_daemon, list(_, Property(&mp::ListRequest::request_ipv4, IsFalse()), _));
+    EXPECT_THAT(send_command({"list", "--no-ipv4"}), Eq(mp::ReturnCode::Ok));
 }
 
 // mount cli tests
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -39,6 +39,7 @@
 #include "mock_logger.h"
 #include "mock_process_factory.h"
 #include "mock_standard_paths.h"
+#include "mock_virtual_machine.h"
 #include "mock_virtual_machine_factory.h"
 #include "mock_vm_image_vault.h"
 #include "stub_cert_store.h"
@@ -498,6 +499,12 @@ struct LaunchWithNoNetworkCloudInit : public Daemon, public WithParamInterface<s
 };
 
 struct RefuseBridging : public Daemon, public WithParamInterface<std::tuple<std::string, std::string>>
+{
+};
+
+struct ListIP : public Daemon,
+                public WithParamInterface<
+                    std::tuple<mp::VirtualMachine::State, std::vector<std::string>, std::vector<std::string>>>
 {
 };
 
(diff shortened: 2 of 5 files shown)


tests/test_daemon.cpp
```diff
@@ -1217,6 +1222,46 @@ TEST_F(Daemon, skips_over_instance_ghosts_in_db) // which will have been sometim
...
+    mp::Daemon daemon{config_builder.build()};
+
+    auto [cmds, strs] = GetParam();
+
+    std::stringstream stream;
+
+    for (const auto& c : cmds)
+    {
+        // send_command({"start", "lovely-unicorn"});
+        // send_command({"list"}, stream);
```
ricab, line 1244: leftovers

tests/test_daemon.cpp
```diff
@@ -1217,6 +1222,46 @@ TEST_F(Daemon, skips_over_instance_ghosts_in_db) // which will have been sometim
...
+        send_command(c, stream);
+    }
+
+    for (const auto& s : strs)
+    {
+        // EXPECT_THAT(stream.str(), HasSubstr("192.168.2.123"));
```
ricab, line 1250: leftover

ricab (commented):
Oh yeah, great PR :) 

Only minor comments in tests.

tests/test_cli_client.cpp
```diff
@@ -848,7 +854,7 @@ TEST_F(Client, info_cmd_fails_with_names_and_all)
 // list cli tests
 TEST_F(Client, list_cmd_ok_no_args)
 {
-    EXPECT_CALL(mock_daemon, list(_, _, _));
+    EXPECT_CALL(mock_daemon, list(_, make_list_request_ipv4_matcher<mp::ListRequest>(true), _));
```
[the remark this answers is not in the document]
luis4a0: Right, no point in having a simple function there. Changed.

tests/test_daemon.cpp
luis4a0: Oops, removed.

tests/test_daemon.cpp
luis4a0: Also removed.

tests/test_cli_client.cpp
```diff
@@ -862,6 +868,12 @@ TEST_F(Client, list_cmd_help_ok)
     EXPECT_THAT(send_command({"list", "-h"}), Eq(mp::ReturnCode::Ok));
 }
 
+TEST_F(Client, list_cmd_no_ipv4_ok)
+{
+    EXPECT_CALL(mock_daemon, list(_, make_list_request_ipv4_matcher<mp::ListRequest>(false), _));
```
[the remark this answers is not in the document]
luis4a0: Also changed.

tests/mock_virtual_machine.h
```diff
@@ -29,14 +29,23 @@ struct MockVirtualMachine : public multipass::VirtualMachine
 {
     MockVirtualMachine(const std::string vm_name) : VirtualMachine{vm_name}
     {
-        ON_CALL(*this, current_state()).WillByDefault(Return(multipass::VirtualMachine::State::off));
+        ON_CALL(*this, start()).WillByDefault([this]() { fake_state = multipass::VirtualMachine::State::running; });
+        ON_CALL(*this, stop()).WillByDefault([this]() { fake_state = multipass::VirtualMachine::State::off; });
+        ON_CALL(*this, shutdown()).WillByDefault([this]() { fake_state = multipass::VirtualMachine::State::off; });
+        ON_CALL(*this, suspend()).WillByDefault([this]() { fake_state = multipass::VirtualMachine::State::suspended; });
+        ON_CALL(*this, current_state()).WillByDefault([this]() { return fake_state; });
```
ricab, line 36: So you moved to mock after all :) But then you shouldn't need any of this. The advantage of the mock is that you can tell it what you want of it in the spot. For instance:
```cpp
EXPECT_CALL(mock, current_state())
    .WillOnce(Return(<desired_state>))
    .WillOnce(Return(<another_state>));
EXPECT_CALL(mock, ensure_vm_is_running()).WillRepeatedly(Throw(std::runtime_error{"Not running"}));
```

luis4a0: Good to know, thanks! The problem here is that, since we are dealing with a real daemon, thus AFAICS we cannot obtain the instance object (which is the mocked VM).

luis4a0:
Hey, bors had to rebase this from master!

tests/test_daemon.cpp
```diff
@@ -1217,6 +1224,42 @@ TEST_F(Daemon, skips_over_instance_ghosts_in_db) // which will have been sometim
     mp::Daemon daemon{config_builder.build()};
 }
 
+TEST_P(ListIP, lists_with_ip)
+{
+    [[maybe_unused]] auto mock_factory = use_a_mock_vm_factory();
```
ricab, line 1229: Why `[[maybe_unused]]`? It's used in line 1235.

tests/mock_virtual_machine.h
```diff
@@ -37,6 +37,11 @@ struct MockVirtualMachine : public multipass::VirtualMachine
         ON_CALL(*this, management_ipv4()).WillByDefault(Return("0.0.0.0"));
         ON_CALL(*this, get_all_ipv4(_)).WillByDefault(Return(std::vector<std::string>{"192.168.2.123"}));
         ON_CALL(*this, ipv6()).WillByDefault(Return("::/0"));
+        ON_CALL(*this, ensure_vm_is_running()).WillByDefault(Invoke([]() { throw std::runtime_error("Not running"); }));
+    }
+
+    MockVirtualMachine() : MockVirtualMachine{""}
```
ricab, line 43: Why add this?

tests/mock_virtual_machine.h
```diff
@@ -37,6 +37,11 @@ struct MockVirtualMachine : public multipass::VirtualMachine
         ON_CALL(*this, management_ipv4()).WillByDefault(Return("0.0.0.0"));
         ON_CALL(*this, get_all_ipv4(_)).WillByDefault(Return(std::vector<std::string>{"192.168.2.123"}));
         ON_CALL(*this, ipv6()).WillByDefault(Return("::/0"));
+        ON_CALL(*this, ensure_vm_is_running()).WillByDefault(Invoke([]() { throw std::runtime_error("Not running"); }));
```
ricab, line 40: Did you forget this one? Or is it needed?

luis4a0: I added it because it is what the stub does. If I remove it, the list tests fail.

I investigated why it is needed. It happens that the function `ensure_vm_is_running()` is called many times when executing things in the instance. In particular, when we use `send_commands()` the test client hangs up waiting for something from the VM, which in fact is a mock and will return from an execution on the terminal.

I added an expectation in the body of the test function, so as to not add things to the mock.

tests/mock_virtual_machine.h
luis4a0: I added it to make the interface the same as the stub. But right, it's not needed. Removed.

tests/test_daemon.cpp
luis4a0: Right, it was unused in an old version of the code. Removed.

tests/test_daemon.cpp
```diff
@@ -1217,6 +1224,45 @@ TEST_F(Daemon, skips_over_instance_ghosts_in_db) // which will have been sometim
...
+    EXPECT_CALL(*mock_factory, create_virtual_machine).WillRepeatedly([&instance_ptr](const auto&, auto&) {
+        return std::move(instance_ptr);
+    });
+
+    auto [state, cmd, strs] = GetParam();
+
+    EXPECT_CALL(*instance_ptr, current_state()).WillRepeatedly(Return(state));
+    EXPECT_CALL(*instance_ptr, ensure_vm_is_running()).WillRepeatedly(Invoke([]() {
+        throw std::runtime_error("Not running");
+    }));
```
ricab, line 1244:
```suggestion
    EXPECT_CALL(*instance_ptr, ensure_vm_is_running()).WillRepeatedly(Throw(std::runtime_error("Not running")));
```

ricab (approved):
Minor tweak inline, but otherwise good to go. Thanks for the daemon test, codecov should say 200% of the diff covered :wink: 

bors d+

PR #1974, opened 2021-02-15
Remove useless SSH session in the daemon

luis4a0:
In our first implementation of IP listing in the `list` command, we used to create a SSH session in the daemon to communicate with the instance. We moved later creating that session to the VirtualMachine class, but we forgot to remove the code in the daemon to create such session. As a result, we created two sessions each time we wanted to ask an instance for its IP's.

This PR just removes the spurious SSH session in the daemon.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1422,8 +1422,6 @@ try // clang-format on
 
         if (request->request_ipv4() && mp::utils::is_running(present_state))
         {
-            auto vm_specs = vm_instance_specs[name];
-
             std::string management_ip = vm->management_ipv4();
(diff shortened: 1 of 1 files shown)


ricab (commented):
Good catch @luis4a0, thanks! 

`vm_specs` in 1425 is now unused and can be removed too.

luis4a0:
Right, line 1425 was removed also. BTW, I don't know why the compiler didn't complain about it (although I'm sure it optimized and removed that line).

ricab (approved):
Yeah, not sure why it didn't complain either. Anyway, it's good now, thanks.

bors r+

luis4a0:
Just restored the branch so I could rebase https://github.com/canonical/multipass/pull/1969. Then, closed again the PR.

PR #1975, opened 2021-02-16
[utils] untangle a circular dependency with logger

Saviq:


src/utils/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2017-2020 Canonical Ltd.
+# Copyright © 2017-2021 Canonical Ltd.
```
ricab, line 1: This isn't wrong, since once you change it, it's been changed :) But not necessary either.

ricab (commented):
LGTM thanks. Just one comment inline.

src/utils/CMakeLists.txt
ricab: That said, I don't mind if you prefer to leave it like that. I wouldn't mind even doing a commit changing it in all files each year. It would add to each file history, but that wouldn't be a problem IMO. Anyway, as you prefer here.

Saviq: Yeah it was a leftover, fixed :)

ricab (approved):
Yup, thanks for this!

bors r+

PR #1980, opened 2021-02-22
[bors] Do not delete any child branches if parent branch is deleted

townsend2010:
---
According to [this](https://bors.tech/documentation/), setting this supposedly keeps any child branches from being deleted when merging and deleting parent branches:
```
If turned on, and if delete_merged_branches is also turned on, then when a pull request is merged and its
base branch is about to be deleted, any other pull requests made against the base branch will be fixed up.
```

Thought we should give it a try.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -2,4 +2,5 @@ status = [ "Linux", "Windows", "macOS" ]
 block-labels = [ "no-merge" ]
 timeout-sec = 12000
 delete-merged-branches = true
+update_base_for_deletes = true
 cut_body_after = "---"


ricab (approved):
Nice, thanks! I suppose we need to merge before testing, so approving but delegating the merge, so that you can choose how best to handle it.

bors d+

townsend2010:
Hey @ricab,

Thanks for the review!  I think I'm going to wait on @Saviq's feedback as well since he's most familiar with `bors` and perhaps he's tried this before and it didn't work or there was some other issue with it.

Saviq (approved):
Oh, oh, yes!

Weird that GitHub did it on its own at some point, but not any more…

PR #1986, opened 2021-02-26
Add command aliases to the CLI client

luis4a0:
This PR adds the ability to create aliases to be executed in determined instances.

Private bits are in https://github.com/canonical/multipass-private/pull/386


ricab (commented):
Hi @luis4a0, I did a high-level, superficial review pass. It looks very nice overall! :clap:

The one thing that concerns me a bit, from an architectural point of view, is the multiple `AliasDict` objects. They are meant to always represent the same mapping, but the multiple copies open a space for conflicts. They need to be loaded and saved multiple times, which increases the potential for concurrency issues. Admittedly I have only just considered this, but it seems to me that it would be better to have a single, central object owned by the client, that the relevant commands operated on via a reference. At that point, I am not sure the separate type is beneficial (relatively to a map). Then again, perhaps it's worth keeping it so that it could be reused in the GUI client. But please let me know if there are other things I am not considering.

I suspect you went with the multiple objects because the `Command` objects are created with `Client::add_command`, which takes no arguments, while a central dict would have to be passed only to a few commands. But you can modify `Client::add_command` to receive and pass on whatever arguments each `Command`'s ctor takes, with a technique called perfect forwarding. It's what things like `emplace` methods or `make_unique` use, and it can even move rvalues as appropriate. I will submit a PR that you can use if you want. (Done: #2032.)

I didn't quite get the `ClientFormatter` separate from the existing `Formatter`, but I confess I didn't look closely. I think we discussed something related to that, but I forget what the issue was. We can discuss further if you want.

luis4a0:
Hey @ricab, thanks a lot for the review!

About the multiple `AliasDict` objects, my intention was to read the file exactly once per session, and to write to it at most once. I thought an alternative when having a many objects would be making it a singleton, like done for `Settings`. OTOH, the solution you propose to add taking parameters to `add_command` is very elegant. I'll think a bit more on this.

About formatters, we already discussed on the topic. The current formatters work with RPC objects, while the new formatter work with a CLI-only thing, which is the alias dictionary. Implementing the dict formatters inside the current formatters would imply to link against RPC all things which use formatters, which is a big refactor. I opted to separate both, i.e., adding a "client" formatter, which formats data structures defined in the client only.

ricab:
Hey @luis4a0, about `AliasDict`, I'll let you think about it, but let me explain my reasoning a bit better. My take is that it is better to have a direct correspondence between resource ownership and an object to represent it. It's harder to guarantee invariants with shared ownership. Classes should be designed to be reusable (e.g. to employ in a full-blown GUI), but multiple objects representing the same resource would be harder to synchronize. Assuming that only one client will use the class at a time is also not very self contained. 

With a single `AliasDict` representing a persistent set of aliases, we'd get a single place to encapsulate synchronization and preserve invariants. Entities that only need reading would only get read access by default (e.g. with a const reference), write access being restricted to select entities (for now only the `alias` cmd, I think). New `AliasDict`s could be created if we ever needed separate alias sets.

A singleton would not allow that read/write separation. It's advantages would not be very useful here: not something to be used in arbitrary parts of the code (like settings or std paths); not something that avoids deep parameter passing all over (`AliasDict` param passed only in one level of the call tree); not a wrapper of otherwise free functions that we need to mock. The one advantage here would be enforcing a single object, but I don't think that's worth the "glorified global".

For the formatters, right, the dependency on the RPC... thanks for reminding me. I wish there was a way to isolate the dependency and otherwise homogenize things, but I don't have a proposal right now. I guess we could always revisit.

luis4a0:
Thanks for the explanation @ricab ! Now it makes complete sense to me what you proposed.

luis4a0:
Sorry, wrong button :X

src/client/common/client_formatter.cpp
```diff
@@ -0,0 +1,165 @@
...
+std::vector<std::string> mp::ClientFormatter::escape_args(const std::vector<std::string>& args) const
+{
+    std::vector<std::string> escaped_args;
+
+    auto replace = [](const std::string& s) { return QString::fromStdString(s).replace(" ", "\\ ").toStdString(); };
+
+    std::transform(args.cbegin(), args.cend(), std::back_inserter<std::vector<std::string>>(escaped_args), replace);
+
+    return escaped_args;
+}
```
Saviq, line 65: Just escaping space is going to bite us hard. What if there's semicolons or exclamation marks in the alias?

If anything, we'd need to quote as appropriate and escape quotes themselves.

Saviq (changes requested):
Hey, looks good overall! A handful of comments inline.

src/client/cli/cmd/alias.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    auto status = parser->commandParse(this);
+    if (status != ParseCode::Ok)
+        return status;
+
+    if (parser->positionalArguments().count() != 2)
+    {
+        cerr << "Wrong number of arguments given\n";
+        return ParseCode::CommandLineError;
+    }
```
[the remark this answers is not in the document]
townsend2010: Yeah, totally agree about not splitting args on spaces and I think option 2 is the best choice.

src/client/cli/argparser.cpp
```diff
@@ -144,8 +144,20 @@ mp::ParseCode mp::ArgParser::parse()
...
+    if (execute_alias)
+    {
+        chosen_command = findCommand("exec");
+
+        return ParseCode::Ok;
+    }
+
     // Fall through
-    cout << "Error: Unknown Command '" << qUtf8Printable(requested_command) << "' (try \"multipass help\")\n";
+    cout << "Error: Unknown command or alias '" << qUtf8Printable(requested_command) << "' (try \"multipass help\")\n";
```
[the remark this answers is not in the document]
luis4a0: Right, added.

include/multipass/cli/format_utils.h
```diff
@@ -41,6 +41,14 @@ template <typename Instances>
...
 
 void filter_aliases(google::protobuf::RepeatedPtrField<multipass::FindReply_AliasInfo>& aliases);
+
+// Computes the column width needed to display all the elements of a range [begin, end). get_width is a function
+// which takes as input the element in the range and returns its width in columns.
+static constexpr auto column_width = [](const auto begin, const auto end, const auto get_width, int minimum_width = 0) {
+    auto max_width =
+        std::max_element(begin, end, [&get_width](auto& lhs, auto& rhs) { return get_width(lhs) < get_width(rhs); });
+    return std::max(get_width(*max_width) + 2, minimum_width);
+};
```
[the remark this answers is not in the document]
luis4a0: Definetely. This PR is not the place to do it, we can keep it in mind for later.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,207 @@
...
+        else
+        {
+            QDir config_path = QFileInfo(config_file_name).absoluteDir();
+            if (!config_path.exists())
+            {
+                config_path.mkpath(config_path.absolutePath());
+            }
+        }
+
+        QFile::copy(temp_file_name, config_file_name);
```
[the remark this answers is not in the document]
luis4a0: You're right. Solved. Again, we must keep in mind improving the utils for file operations.

src/client/cli/cmd/alias.cpp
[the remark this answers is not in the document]
luis4a0: Thanks! About option 2: is `[--]` optional? Shouldn't it be `<name> <instance> <command> [-- <args> ...]`?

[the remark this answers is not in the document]
Saviq: It is optional, you only need it if any of the `<args>` starts with a `-`. All positional arguments beyond `<command>` are arguments to the alias, so there's no need for it.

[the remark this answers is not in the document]
ricab: +1 for no spaces in cmd, preferring no colon, and matching `exec` — this problem was already solved for that cmd, which works quite well. The only thing that surprised me is the placement of the `[--]`. Shouldn't it come before `<command>`?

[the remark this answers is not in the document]
Saviq: > The only thing that surprised me is the placement of the [--]. Shouldn't it come before <command>?

Probably should, yes:
```plain
$ multipass help exec
Usage: multipass exec [options] <name> [--] <command>
…
```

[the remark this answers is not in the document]
Saviq: The only problem I've with option 2 is that we now have 3 positional arguments. It becomes unwieldy and unclear whether `<name>` comes first, or does `<command>`? The most common use will be just a plain:
```plain
multipass alias cowsay instance cowsay
```

We should probably support a short-hand:
```plain
multipass alias cowsay instance
```

The order is awkward, too (the common "target", "second target" problem of `ln`). I think the most intuitive would be to do (and that's the order `snap` has it:
```plain
multipass alias instance:cowsay cowsay
# shorthand
multipass alias instance:cowsay
```

But then the handling of arguments becomes tricky:
```
# this is likely wrong
multipass alias instance:cowsay arg1 arg2
# as is this
multipass alias instance:cowsay -- arg1 arg2
# these would be the correct ones
multipass alias instance:cowsay cowsay arg1 arg2
multipass alias instance:cowsay -- cowsay arg1 arg2
```

I'm starting to think we should skip arguments altogether for now and see if we get feature requests for that.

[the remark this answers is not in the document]
Saviq: So I think I'm changing my vote to option 3:

3. `multipass alias <instance>:<command> [<alias>]`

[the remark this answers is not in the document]
luis4a0: And in case we skip arguments and stick to 3, who wants arguments can create a script on the instance. But in case we got requests to add arguments to aliases, I think we'll have to add a new syntax to define an alias with commands, thus complicating the interface.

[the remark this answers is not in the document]
ricab: Hmm, not allowing arguments and requiring a script/alias/fn inside the instance would be quite limiting. But the arguments would still be allowed if quoted with option 3, right? So the mismatch with `exec` worries me more.

The awkwardness in `ln` isn't intrinsic, is it? Other commands have the target first (e.g. `tar`). It's often the case when an equals sign is implicit (e.g. `git config --global alias.co checkout`). With `ln`, I think it's rather the mismatch with `cp` and `mv` that makes it awkward. I fear we could be causing the same feel for someone coming from `exec`:

```plain
# if you're used to 
multipass -v exec foo -- ip -br -c a
# it's perhaps less awkward to say
multipass -v alias fooip foo -- ip -br -c -a
# then
multipass -v alias foo:"ip -br -c -a" fooip
# idk
```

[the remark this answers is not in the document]
Saviq: Let me summarize what we talked about:

- `ssh` (the client, not the protocol) is weird (aka broken) in that:
  ```bash
  $ ssh localhost mkdir foo\ bar
  $ ssh localhost ls
  foo
  bar
  $ ssh localhost echo \$HOME
  /home/ubuntu
  ```
  I.e. it always has a shell interpret the command line. So the user has to be very careful on how to quote things to become separate arguments.
- `multipass exec` is not using a shell:
  ```bash
  $ multipass exec primary mkdir foo\ bar
  $ multipass exec primary ls
  'foo bar'
  $ multipass exec primary echo \$HOME
  $HOME
  ```
  We definitely want to maintain that in the `alias` feature, that exactly the arguments provided are passed through to the `<command>` executable in the instance. If we support arguments, you could still have a shell interpret the command with `sh -c "..."`. If `<command>` includes special characters, the executable has to be called exactly that.
- `alias` from `exec` is different in that `alias` is creating an entity (the alias), and it's prevalent in CLIs that the thing you create is last on the command line:
  ```bash
  $ cp <src> <tgt>
  $ mv <src> [<src> ...] <tgt>
  $ ln <target> [<link>]
  $ snap alias <snap.app> [<alias>]
  ```
- While awkward, it's still possible to allow, with time, the addition of arguments, possibly like so:
  ```plain
  multipass alias <instance>:<command> [<alias>] [-- <arg> …]
  ```
- There will still be even more complex use cases that are probably out of scope, e.g. templating/placeholders.
- The most common use case of aliases will be exporting a command from inside the instance to the host, and so I'd like to optimize for this:
  ```plain
  multipass alias <instance>:<command> [<alias>]
  ```
- Note this still supports arguments when _invoking_ the alias:
  ```plain
  multipass <alias> [<arg> ...]
  <alias> [<arg> ...]
  ```
  Equivalent to:
  ```plain
  multipass exec <instance> -- <command> [<arg> ...]
  ```

I vote for this last incantation for this.

src/client/common/alias_dict.cpp
```diff
@@ -175,33 +175,31 @@ void mp::AliasDict::save_dict()
         aliases_json.insert(name, definition);
     }
 
-    QTemporaryFile temp_file;
+    auto config_file_name = QString::fromStdString(aliases_file);
+
+    QDir config_path = QFileInfo(config_file_name).absoluteDir();
+    if (!config_path.exists())
+        config_path.mkpath(config_path.absolutePath());
```
Saviq, line 182: Shame `QTemporaryFile` doesn't take care of it…

src/client/common/alias_dict.cpp
```diff
@@ -175,33 +175,31 @@ void mp::AliasDict::save_dict()
...
-            if (!config_path.exists())
-            {
-                config_path.mkpath(config_path.absolutePath());
-            }
+            QFile::rename(config_file_name, backup_file_name);
         }
 
-        QFile::copy(temp_file_name, config_file_name);
+        QFile::rename(temp_file_name, config_file_name);
     }
```
Saviq, line 204: Should probably handle errors here?

src/client/common/alias_dict.cpp
luis4a0: Yeah, we could wrap all this in `utils`, and do something close to `tests/temp_file.h`.

src/client/common/alias_dict.cpp
luis4a0: Good idea. But... what to do if we can't rename? Error out?

In fact, I just fixed a bug in windows here. The `QTemporaryFile` refused to be renamed using `QFile::rename()`; instead, I had to use the non-static version of `rename`.

src/client/cli/cmd/alias.cpp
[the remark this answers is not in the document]
luis4a0: Hey, thanks for the comments! I've been considering all this. Now I'm convinced that, in terms of usability and simplicity, `multipass alias <instance>:<command> [<aliasname>]` would be the best choice at the moment. I don't think adding parameters after `--` would be a big issue. In fact, since we have three positional arguments being one of them optional, we can't avoid using `--`. Circumventing the use of `--` would imply to either make one argument mandatory, thus losing the simplicity of the approach.

src/client/cli/cmd/alias.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    if (aliases.get_alias(alias_name))
+    {
+        cerr << fmt::format("Alias '{}' already exists\n", alias_name);
+        return ParseCode::CommandLineError;
+    }
+
+    std::string instance = definition.left(colon_pos).toStdString();
+    QString full_command = definition.right(definition.size() - colon_pos - 1);
+    QStringList full_command_split = full_command.split(' ', QString::SkipEmptyParts);
```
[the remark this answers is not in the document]
luis4a0: Yeah, fire in the hole! This was now removed because we removed arguments from the interface.

src/client/common/client_formatter.cpp
luis4a0: Again, we removed arguments. For the time being, this is not anymore relevant.

src/client/cli/cmd/aliases.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+mp::ParseCode cmd::Aliases::set_formatter(mp::ArgParser* parser)
+{
+    std::string user_arg = parser->value(format_option_name).toStdString();
+
+    if (user_arg != "csv" && user_arg != "json" && user_arg != "table" && user_arg != "yaml")
+    {
+        fmt::print(cerr, "Invalid format type given.\n");
+        return ParseCode::CommandLineError;
+    }
```
townsend2010, line 62: I'm not sure if you know this, but there is already a function in client/cli/cmd/common_cli.cpp called handle_format_option() that basically does all of this.

src/client/cli/cmd/unalias.cpp
```diff
@@ -0,0 +1,87 @@
...
+        return parser->returnCodeFrom(ret);
+    }
+
+    if (aliases.remove_alias(alias_to_remove))
+    {
+        try
+        {
+            MP_PLATFORM.remove_alias_script(alias_to_remove);
+        }
+        catch (std::runtime_error&)
```
townsend2010, line 41: As mentioned before, it would be good to get the error message here and print it out as well.

src/client/common/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 add_library(client_common STATIC
+  alias_dict.cpp
+  client_formatter.cpp
```
townsend2010, line 17: A little nitpick...please put `client_formatter.cpp` *after* `client_common.cpp` to be alphabetical.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+        auto name = QString::fromStdString(record.first);
+        auto definition = alias_to_json(record.second);
+
+        aliases_json.insert(name, definition);
+    }
+
+    auto config_file_name = QString::fromStdString(aliases_file);
+
+    QDir config_path = QFileInfo(config_file_name).absoluteDir();
```
townsend2010, line 188: Could make this `auto`.

src/client/common/alias_dict.cpp
townsend2010: Yeah, if renaming has an error, it should error out since the config file won't be consistent at this point.

src/platform/platform_linux.cpp
```diff
@@ -152,6 +157,13 @@ void update_bridges(std::map<std::string, mp::NetworkInterfaceInfo>& networks)
         }
     }
 }
+
+std::string get_alias_script_path(const std::string& alias)
+{
+    QDir aliases_folder = MP_PLATFORM.get_alias_scripts_folder();
```
townsend2010, line 163: Perhaps `auto`?

src/platform/platform_linux.cpp
```diff
@@ -181,6 +193,57 @@ bool mp::platform::Platform::link(const char* target, const char* link)
...
+    std::string multipass_exec = mu::in_multipass_snap() ? "exec /usr/bin/snap run multipass"
+                                                         : QCoreApplication::applicationFilePath().toStdString();
+
+    try
+    {
+        std::string script = "#!/bin/sh\n\n" + multipass_exec + " " + alias + "\n";
+
+        MP_UTILS.make_file_with_content(file_path, script);
+    }
+    catch (const std::runtime_error&)
```
townsend2010, line 226: Again, shouldn't we be more verbose on what the actual error is?

src/platform/platform_linux.cpp
```diff
@@ -181,6 +193,57 @@ bool mp::platform::Platform::link(const char* target, const char* link)
...
+    auto permissions =
+        MP_FILEOPS.permissions(file) | QFileDevice::ExeOwner | QFileDevice::ExeGroup | QFileDevice::ExeOther;
+
+    if (!MP_FILEOPS.setPermissions(file, permissions))
+        throw std::runtime_error(fmt::format("error setting permissions to alias script '{}'", file_path));
+}
+
+void mp::platform::Platform::remove_alias_script(const std::string& alias)
+{
+    std::string file_path = get_alias_script_path(alias);
```
townsend2010, line 241: Could do `auto` here...

src/utils/file_ops.cpp
```diff
@@ -28,16 +28,31 @@ bool mp::FileOps::isReadable(QDir& dir)
...
+{
+    return dir.mkpath(dirName);
+}
+
 bool mp::FileOps::rmdir(QDir& dir, const QString& dirName)
 {
     return dir.rmdir(dirName);
 }
 
+bool mp::FileOps::exists(QFile& file)
```
townsend2010, line 41: Should be able to do this:

```suggestion
bool mp::FileOps::exists(const QFile& file)
```

src/utils/file_ops.cpp
```diff
@@ -28,16 +28,31 @@ bool mp::FileOps::isReadable(QDir& dir)
     return dir.isReadable();
 }
 
+bool mp::FileOps::mkpath(QDir& dir, const QString& dirName)
```
townsend2010, line 31: I *think* `QDir` should be able to be `const` here.

tests/file_operations.cpp
```diff
@@ -55,6 +55,10 @@ qint64 mpt::make_file_with_content(const QString& file_name, const std::string&
     if (!parent_dir.mkpath(".")) // true if directory already exists
         throw std::runtime_error(fmt::format("failed to create test dir: '{}'", parent_dir.path()));
 
+    QDir file_dir = QFileInfo(file).absoluteDir();
```
townsend2010, line 58: Could be `auto`.

townsend2010 (changes requested):
Hey @luis4a0,

Thanks!  Great code & test coverage!  I've still yet to try it out, but I did a preliminary pass of the code itself.  I'll do another pass of the code once I try this as I think I'll better understand what this all does :smiley: 

I'm also still trying to understand why the need for the `ClientFormatter` class as opposed to what we already have.  Is this just to improve upon what is already there and in a follow up PR, convert other commands to use it?

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+#include <QDir>
+#include <QFile>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonParseError>
+#include <QTemporaryFile>
+
+#include "../../daemon/json_writer.h"
```
townsend2010, line 30: I wonder if `json_writer` should just be moved into utils since it's now needed in more places than the daemon.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (aliases.erase(alias) > 0)
+    {
+        modified = true;
+        return true;
+    }
+
+    return false;
+}
+
+mp::AliasDict::size_type mp::AliasDict::remove_aliases_for_instance(const std::string& instance)
```
townsend2010, line 70: I'm not sure why this returns a size as it is not used by any of the callers.  You could then also tighten up the code below by getting rid of `erased_elements`, setting `modified = true` in the loop (assigning is a boolean is less expensive than a `++` operation, not that we really care about that :wink: ), and get rid of the `if (erased_elements > 0)` logic.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+    return aliases.size();
+}
+
+void mp::AliasDict::load_dict()
+{
+    QFile db_file{QString::fromStdString(aliases_file)};
+
+    aliases.clear();
+
+    if (!db_file.open(QIODevice::ReadOnly))
```
townsend2010, line 139: I wonder if there should be more of an error here is the dictionary database cannot be opened?

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (doc.isNull())
+        return;
+
+    QJsonObject records = doc.object();
+    if (records.isEmpty())
+        return;
+
+    for (auto it = records.constBegin(); it != records.constEnd(); ++it)
+    {
+        std::string alias = it.key().toStdString();
```
townsend2010, line 153: `auto` should be fine here.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,213 @@
...
+        QJsonObject record = it.value().toObject();
+        if (record.isEmpty())
+            break;
+
+        std::string instance = record["instance"].toString().toStdString();
+        std::string command = record["command"].toString().toStdString();
```
townsend2010, line 159: idem, `auto`

src/client/common/client_formatter.cpp
```diff
@@ -0,0 +1,140 @@
...
+        formatted_output = format_csv(aliases);
+    else if (preferred_format == "json")
+        formatted_output = format_json(aliases);
+    else if (preferred_format == "table")
+        formatted_output = format_table(aliases);
+    else if (preferred_format == "yaml")
+        formatted_output = format_yaml(aliases);
+
+    return formatted_output;
+}
```
townsend2010, line 49: I'm not a fan of all of the `if`/`else if`'s here.  In `format_utils.cpp`, it creates a map of the various formatters and then lookup is easy based on the preferred formatter.  I wonder if something like that would be better here.

townsend2010 (changes requested):
Here's another round of review :slightly_smiling_face: 

I don't know if this has been discussed already, but should we do some validation when creating an alias to ensure the given command exists in the instance?

Also, there are some inconsistencies in the various outputs:

- Both the `table` and `csv` outputs reference the alias name as `Alias` whereas both the `json` and `yaml` reference them as `name`.  I think it should be `alias` instead of `name`.
- The `yaml` output is quite different than the `json` output.  The structure of the `json` output is what I'd expect.  For the `yaml`, I'd expect it to look something like:
```yaml
aliases:
- command: ls
  instance: test
  name: test-ls
- command: free
  instance: test
  name: test-free
- command: top
  instance: test
  name: test-top
```

townsend2010:
BTW, my question about the `ClientFormatter` class, I see you answered @ricab already, so that makes sense. I would like in the future to be able to combine these as I think there is some code duplication.

Lastly, there is a merge conflict that needs fixing.

src/utils/file_ops.cpp
```diff
@@ -38,7 +38,7 @@ bool mp::FileOps::rmdir(QDir& dir, const QString& dirName)
     return dir.rmdir(dirName);
 }
 
-bool mp::FileOps::exists(QFile& file)
+bool mp::FileOps::exists(const QFile& file)
```
townsend2010, line 41: Need to change it in the header too :wink:

include/multipass/cli/alias_dict.h
```diff
@@ -0,0 +1,62 @@
...
+#define MULTIPASS_ALIAS_DICT_H
+
+#include <multipass/cli/alias_definition.h>
+#include <multipass/optional.h>
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace mp = multipass;
```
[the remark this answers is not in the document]
luis4a0: OK. I understand that those includes are superfluous and that we shouldn't use the namespace shortcut. Right?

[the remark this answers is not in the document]
townsend2010: Hey @luis4a0, Sorry, I'm not sure why there are 4 lines shown for this comment, but I was only meaning to reference `namespace mp = multipass` here (https://github.com/canonical/multipass/pull/1986/files#r646866219 shows it correctly :shrug:),  And yes, we shouldn't use the namespace shortcut in headers.

include/multipass/cli/alias_definition.h
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_ALIAS_DEFINITION_H
+#define MULTIPASS_ALIAS_DEFINITION_H
+
+#include <string>
+#include <vector>
```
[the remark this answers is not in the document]
luis4a0: Yes, removed `<vector>`.

include/multipass/cli/alias_dict.h
```diff
@@ -0,0 +1,62 @@
...
+
+#ifndef MULTIPASS_ALIAS_DICT_H
+#define MULTIPASS_ALIAS_DICT_H
+
+#include <multipass/cli/alias_definition.h>
+#include <multipass/optional.h>
+
+#include <string>
+#include <unordered_map>
+#include <vector>
```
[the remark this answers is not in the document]
luis4a0: Right. This one and the previous one were leftovers from the first implementation, on which we stored the arguments of the alias in a vector. Removed.

include/multipass/cli/argparser.h
```diff
@@ -39,7 +42,7 @@ class ArgParser
 
     void addPositionalArgument(const QString& name, const QString& description, const QString& syntax = QString());
 
-    ParseCode parse();
+    ParseCode parse(const multipass::optional<multipass::AliasDict>& aliases = std::nullopt);
```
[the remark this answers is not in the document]
luis4a0: OK, removed.

include/multipass/cli/argparser.h
```diff
@@ -67,13 +70,20 @@ class ArgParser
 
     bool containsArgument(const QString& argument) const;
 
+    multipass::optional<AliasDefinition> executeAlias()
```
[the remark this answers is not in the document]
luis4a0: OK, also removed.

include/multipass/cli/alias_dict.h
```diff
@@ -0,0 +1,62 @@
...
+    typedef typename DictType::key_type key_type;
+    typedef typename DictType::mapped_type mapped_type;
+    typedef typename DictType::size_type size_type;
+
+    AliasDict();
+    ~AliasDict();
+    void add_alias(const std::string& alias, const AliasDefinition& command);
+    bool remove_alias(const std::string& alias);
+    size_type remove_aliases_for_instance(const std::string& instance);
+    mp::optional<AliasDefinition> get_alias(const std::string& alias) const;
```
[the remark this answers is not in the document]
luis4a0: Right, removed.

include/multipass/cli/argparser.h
```diff
@@ -67,13 +70,20 @@ class ArgParser
...
+
 private:
     QString generalHelpText();
     QString helpText(cmd::Command* command);
+    ParseCode prepare_alias_execution();
 
     const QStringList& arguments;
     const std::vector<cmd::Command::UPtr>& commands;
     cmd::Command* chosen_command;
+    multipass::optional<AliasDefinition> execute_alias;
```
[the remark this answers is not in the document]
luis4a0: Removed.

include/multipass/utils.h
```diff
@@ -21,6 +21,7 @@
 #include <multipass/logging/level.h>
 #include <multipass/path.h>
 #include <multipass/singleton.h>
+#include <multipass/ssh/ssh_key_provider.h>
```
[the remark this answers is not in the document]
luis4a0: Right, `SSHKeyProvider` is forward-declared.

src/client/common/CMakeLists.txt
luis4a0: Good catch :)

src/utils/file_ops.cpp
luis4a0: Indeed! Added `const`.

src/utils/file_ops.cpp
luis4a0: Done :)

src/client/cli/argparser.cpp
```diff
@@ -94,7 +93,23 @@ mp::ArgParser::ArgParser(const QStringList& arguments, const std::vector<cmd::Co
 {
 }
 
-mp::ParseCode mp::ArgParser::parse()
+mp::ParseCode mp::ArgParser::prepare_alias_execution()
+{
+    if (parser.positionalArguments().size() > 1)
+    {
+        cerr << "Aliases admit no arguments\n";
```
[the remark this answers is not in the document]
luis4a0: Done.

include/multipass/cli/alias_dict.h
```diff
@@ -21,11 +21,7 @@
 #include <multipass/cli/alias_definition.h>
 #include <multipass/optional.h>
 
-#include <string>
```
townsend2010: I don't think it is necessary to remove this as `std::string` is used in the header.

src/client/cli/argparser.cpp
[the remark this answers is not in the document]
townsend2010: I don't see the change.  Has this not been pushed yet?

src/client/cli/cmd/aliases.cpp
townsend2010: Ah, right, it will return a different type.  But yeah, if it's not *too* much work to factor something out that can be shared between the two, that would be great.

src/client/cli/argparser.cpp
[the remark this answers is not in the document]
luis4a0: Oh, yes, sorry. Just pushed.

src/client/cli/cmd/aliases.cpp
luis4a0: I did the refactor, finally. Code looks *much* better now, thanks for the suggestion!

src/client/common/alias_dict.cpp
luis4a0: Moved.

src/client/common/client_formatter.cpp
luis4a0: This is outdated, since the formatter was refactored.

src/client/common/alias_dict.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: Done.

tests/file_operations.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: Done.

src/platform/platform_linux.cpp
luis4a0: Done.

src/platform/platform_linux.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: I thought it'll be useful to know how many aliases were removed but you're right: it's useless. Changed as you suggested: made the function `void` because the only thing that matters is the `modified` flag.

src/client/cli/cmd/alias.cpp
```diff
@@ -0,0 +1,125 @@
...
+        return ParseCode::CommandLineError;
+    }
+    if (colon_pos == 0)
+    {
+        cerr << "No instance name given\n";
+        return ParseCode::CommandLineError;
+    }
+
+    std::string instance = definition.left(colon_pos).toStdString();
+    std::string command = definition.right(definition.size() - colon_pos - 1).toStdString();
```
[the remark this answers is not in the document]
luis4a0: Done.

include/multipass/cli/alias_dict.h
luis4a0: Right. `<string>` is nevertheless used in `alias_definition.h` but it doesn't harm to have it here. Added it again.

src/client/common/alias_dict.cpp
luis4a0: Done, added `utils::create_temp_file_with_path()`.

src/client/cli/cmd/alias.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+
+    try
+    {
+        MP_PLATFORM.create_alias_script(alias_name, alias_definition);
+    }
+    catch (std::runtime_error&)
```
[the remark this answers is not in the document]
luis4a0: Done. Not catching here lets the exception from `create_alias_script` propagate.

src/client/cli/cmd/unalias.cpp
luis4a0: `remove_alias_script` calls `unlink`. Should we inspect `ERRNO` to see the reason in case of failure? Just letting the exception propagate will not add much verbosity here...

src/platform/platform_linux.cpp
luis4a0: Done. Not catching here lets the exception from `create_alias_script` propagate.

src/client/common/alias_dict.cpp
luis4a0: If the database file does not exist, we can safely `return` because there is nothing to do within the function. I added an error in case the file exists and cannot be opened, which I didn't consider before your comment.

src/client/cli/cmd/unalias.cpp
townsend2010: Ok, but what I'm saying is that `remove_alias_script()` throws a `runtime_error` with a message and that message is completely ignored here.  Why not at least combine that error message with the following error message so as not to basically duplicate part of the string.  For example, this is something like what I'm thinking:
```
catch (std::runtime_error& e)
{
    cerr << fmt::format("Warning: {} for '{}'\n", e.what(), alias_to_remove);
}
```
and perhaps change the message that `remove_alias_script()` sets to not have "error" in the wording.  Also, since `remove_alias_script` is a platform function, it may be useful to capture `errno` and add that to the error message.

luis4a0: Excellent, done.

src/client/common/alias_dict.cpp
luis4a0: Right, done.

townsend2010:
Hey @luis4a0, @Saviq, and @ricab,

Would like your opinions on how much should be validated when creating an alias.  Right now, really, no validation is done in that I can create an alias for a non-existent instance or for a command that doesn't exist inside the instance.  Should that be the case or should that be validated?

Saviq:
> Would like your opinions on how much should be validated when creating an alias. Right now, really, no validation is done in that I can create an alias for a non-existent instance or for a command that doesn't exist inside the instance. Should that be the case or should that be validated?

IMO instance should be validated, not the command itself.

luis4a0:
Hi,

I think validation at alias creation forces us to maintain consistency later. As example, say that we validate that a command exists inside an instance and we check it when creating an alias and, later, the command is removed from the instance.

OTOH, validating instance name is another story, it would be consistent with the automatic removal of aliases when purging the instance, already implemented.

townsend2010:
Yeah, IMO, validating if an instance exists should be done.  I was on the fence about validating the command inside, but since both of you suggest not to, I'm on your side of the fence :smiley:

ricab:
Your arguments make sense to me :)

src/client/cli/cmd/alias.cpp
```diff
@@ -110,6 +111,37 @@ mp::ParseCode cmd::Alias::parse_args(mp::ArgParser* parser)
...
+                instance_exists = true;
+                break;
+            }
+
+        return ReturnCode::Ok;
+    };
+
+    auto on_failure = [this](grpc::Status& status) { return ReturnCode::CommandLineError; };
+
+    if (dispatch(&RpcMethod::list, list_request, on_success, on_failure) == ReturnCode::CommandLineError)
```
townsend2010, line 133: Instead of using the `list` method and then iterating over all of the entries returned, it might be better to use the `info` method on the given instance name and see if returns successfully.

townsend2010 (commented):
I have a suggestion inline.

src/client/cli/cmd/alias.cpp
luis4a0: Good thought, done!

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,188 @@
...
+void mp::AliasDict::load_dict()
+{
+    QFile db_file{QString::fromStdString(aliases_file)};
+
+    aliases.clear();
+
+    if (db_file.exists())
+    {
+        if (!db_file.open(QIODevice::ReadOnly))
+            throw std::runtime_error(fmt::format("Error opening file '{}'", db_file.fileName()));
```
townsend2010, line 110: You should be able to make a unit test for this if you change this to use the `FileOps` singleton class.

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        temp_file.close();
+
+        if (MP_FILEOPS.exists(QFile{config_file_name}))
+        {
+            auto backup_file_name = config_file_name + ".bak";
+            QFile backup_file(backup_file_name);
+
+            if (MP_FILEOPS.exists(backup_file) && !MP_FILEOPS.remove(backup_file))
+                throw std::runtime_error(fmt::format("cannot remove old aliases backup file {}", backup_file_name));
```
townsend2010, line 178: Should be able to cover this in a unit test, right?

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+            QFile config_file(config_file_name);
+            if (!MP_FILEOPS.rename(config_file, backup_file_name))
+                throw std::runtime_error(fmt::format("cannot rename aliases config to {}", backup_file_name));
```
townsend2010, line 182: Same here?

src/client/common/alias_dict.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        if (!MP_FILEOPS.rename(temp_file, config_file_name))
+            throw std::runtime_error(fmt::format("cannot create aliases config file {}", config_file_name));
```
townsend2010, line 186: And here?

townsend2010 (commented):
There is a little missing unit test coverage I mentioned below.  I don't see that there really is a test for those and codecov is lying, but please correct me if I'm wrong.

src/client/common/alias_dict.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: Done.

src/client/common/alias_dict.cpp
luis4a0: Done.

townsend2010 (approved):
Ok, let's do this!  Whoop! Whoop!

bors merge

PR #1959, opened 2021-02-08
[sftp_server] Add better logging and handle failure conditions

townsend2010:
- Handle when NULL is returned when trying to allocate a handle in libssh
- Add much more logging for error conditions to account for the SFTP protocol's weak error reporting.

Fixes #1933, Fixes #1950

PR #1990, opened 2021-03-02
[libssh] Update libssh version to use the increased file handles

### LINKED ISSUE #1950: Mounts can't handle more than 256 open files
**Describe the bug**
Trying to open more than 256 files from a Multipass mount fails, leaving the opening process stuck inside the instance and the mount otherwise unusable. Described in detail [here](https://github.com/canonical/multipass/issues/1933#issuecomment-770837840).

**To Reproduce**
How, and what happened?
1. `multipass start` (to get primary with a Home mount)
1. use the script attached to try to open more than 256 files, e.g. `python3 open-files.py .../multipass/3rd-party`
1. observe that it gets stuck after the 256th (sometimes 255th) file
    - to escape, do `multipass umount primary` on the host

**Expected behavior**
Multipass would support more then 256 simultaneous files. Failing that, a clear error would be produced.

**Additional info**
 - OS: both macOS and Ubuntu tried
 - Version 1.5

**Additional context**
  Found when investigating #1933. Lots more detail in there.

townsend2010:
Fixes #1950

---
This is used in conjunction with #1959 to address all of the issues Multipass has with SFTP open file handles.

The diff of the `libssh` change is found at https://github.com/CanonicalLtd/libssh/commit/8de975d702a6b045c475a19627623c215ed50365.


townsend2010:
I have found a flaw in the `libssh` code I changed.  Converting to a draft for now.

townsend2010:
Ok, I think I have a working solution now.

Saviq:
If we hit the (I know, much higher) limit, will it hang still? Is there a way to drop an error instead?

Saviq:
macOS build available: [multipass-1.7.0-dev.344.pr1990+g628cb0d3.mac-Darwin.pkg](https://multipass-ci.s3.amazonaws.com/pr1944/multipass-1.7.0-dev.344.pr1990%2Bg628cb0d3.mac-Darwin.pkg)

townsend2010:
@Saviq
> If we hit the (I know, much higher) limit, will it hang still? Is there a way to drop an error instead?

It will until #1959 is merged (that PR has the fixes for your questions) in which the hang will not occur and the error will be logged in `journald`.

Saviq (approved):
### Before
```python
ubuntu@primary:~$ python3 <<EOF
STASH = []

for i in range(512):
  print(f"{i}...")
  STASH.append(open(f"{i}.tmp", "w"))
EOF
# ...
256...
^C^C^C
# stuck
```

### After
```plain
# ...
511...
ubuntu@primary:~$ 
```

:+1:

bors merge

townsend2010:
I'm guessing bors got a little confused due to the `main` branch name change.  Let's try again.

bors r=Saviq

townsend2010:
Coverage build failing since Codecov has no coverage for the `main` branch.  Merged manually.

Issue #1992, opened 2021-03-04
`list` fails when interrogating instances fails


Saviq:
With the `list` command now trying to interrogate instances for IP numbers, whenever ssh to the instance fails, the `list` command does, too:

```plain
$ multipass ls
list failed: ssh connection failed: 'Timeout connecting to 127.0.0.1'
$ multipass ls
list failed: [ssh proc] failed to open session channel: ''
$ multipass ls
# worked this time
```

luis4a0:
Hey @Saviq, should we just return an empty list of IP's and log in this case?

Saviq:
> Hey @Saviq, should we just return an empty list of IP's and log in this case?

If we know the management IP, show that, otherwise yes - empty.

PR #1993, opened 2021-03-04
Fix IP showing in the `list` command

luis4a0:
Fixes https://github.com/canonical/multipass/issues/1992.


src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -17,30 +17,44 @@
...
-            all_ipv4.push_back(ip);
+            while (ip_it.hasNext())
+            {
+                auto ip_match = ip_it.next();
+                auto ip = ip_match.captured(1).toStdString();
+
+                all_ipv4.push_back(ip);
+            }
+        }
+        catch (const std::exception& e)
```
ricab, line 55: I don't think we should sink all exceptions here. We're OK with SSH errors, but we'd want to know about `YouPCIsOnFireException` :wink: This would mean throwing a specific exception [here](https://github.com/canonical/multipass/blob/e1ac3898263ff735512177580e1a899d73a11aaf/include/multipass/ssh/throw_on_error.h#L39), [here](https://github.com/canonical/multipass/blob/e7913ee116068eaefa726336d8537b901a3ab37d/src/ssh/ssh_process.cpp#L67), and [here](https://github.com/canonical/multipass/blob/e7913ee116068eaefa726336d8537b901a3ab37d/src/ssh/ssh_process.cpp#L146). It could be a very general `SSHException` type. `ExitlessSSHProcessException` should probably inherit from it (we'd want to catch it here). I would just put both in the same header (e.g. ssh_exceptions.h).

There could be other things we'd want to ignore, but that's not the same as ignoring all cases. We'd have to revisit to catch them as needed. False positives are better than false negatives :smirk:

tests/mock_virtual_machine.h
```diff
@@ -19,24 +19,18 @@
...
+
+#include "platform/backends/shared/base_virtual_machine.h"
 
 namespace multipass
 {
 namespace test
 {
-struct MockVirtualMachine : public multipass::VirtualMachine
+// This mocks a virtual machine, but not the methods of the base virtual machine.
+struct MockNonBaseVirtualMachine : public multipass::BaseVirtualMachine
```
ricab, line 30: OK, now I understand what you were asking about. Your issue steams from the fact that you're really testing `BaseVirtualMachine`, not the daemon. These being unit tests, the daemon code itself could be perfectly tested with a mocked `get_all_ipv4`.

So, I think what you want is a `test_base_virtual_machine.h` to test `get_all_ipv4()` in. In that file (only useful there) you would have a special `StubBaseVirtualMachine : public multipass::BaseVirtualMachine`, with stubs for all the methods that need an implementation. Mocks are only useful if you want behavior to change (e.g. if you want `current_state` to return different things). Otherwise a stub is enough. `mock_virtual_machine.h` shouldn't have to change.

ricab (commented):
A couple of asks.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -17,30 +17,44 @@
...
+
 namespace multipass
 {
 
 std::vector<std::string> BaseVirtualMachine::get_all_ipv4(const SSHKeyProvider& key_provider)
 {
     std::vector<std::string> all_ipv4;
 
-    if (state == State::running)
+    if (current_state() == State::running)
```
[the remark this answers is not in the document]
luis4a0: Right, I created a separate commit for this.

src/platform/backends/shared/base_virtual_machine.cpp
luis4a0: Right. I created mp::SSHException and used it where necessary.

tests/mock_virtual_machine.h
luis4a0: Absolutely. I removed the tests from the daemon and created a new `BaseVMTest`. This way it was muuuuuch easier to test `get_all_ipv4()`.

tests/test_base_virtual_machine.cpp
```diff
@@ -0,0 +1,68 @@
...
+    REPLACE(ssh_channel_open_session, [](auto...) { return SSH_OK; });
+    REPLACE(ssh_channel_request_exec, [](auto...) { return SSH_ERROR; });
+    REPLACE(ssh_userauth_publickey, [](auto...) { return SSH_OK; });
+
+    // Check that it indeed throws at execution.
+    mp::SSHSession session{"host", 42};
+    EXPECT_THROW(session.exec("dummy"), mp::SSHException);
+
+    auto ipv4_count = base_vm.get_all_ipv4(key_provider);
+    EXPECT_EQ(ipv4_count.size(), (size_t)0);
```
ricab, line 67:
```suggestion
    EXPECT_EQ(ipv4_count.size(), 0u);
```

ricab (approved):
O'right, that's the ticket! 

I'd have stuck the stub into the cpp just 'cause it's used nowhere else, but header is also fine. Tiny suggestion only.

bors d+

tests/test_base_virtual_machine.cpp
luis4a0: Good, thanks. Done.

luis4a0:
bors r=ricab

PR #2000, opened 2021-03-09
[build] Add vcpkg submodule and use protobuf

townsend2010:


.github/workflows/linux.yml
```diff
@@ -80,6 +80,15 @@ jobs:
         [ ! -f ${PATCH_PREFIX}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}.patch
         [ ! -f ${PATCH_PREFIX}-${{ matrix.build-type }}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}-${{ matrix.build-type }}.patch
 
+    - name: Set up vcpkg
+      uses: lukka/run-vcpkg@v7.1
+      with:
+        setupOnly: true
+        vcpkgDirectory: '${{ github.workspace }}/3rd-party/vcpkg'
+        appendedCacheKey: ${{ hashFiles( '**/vcpkg_manifest/vcpkg.json' ) }}
+        vcpkgTriplet: ${{ matrix.triplet }}
```
Saviq, line 89: This ends up empty, as is unset in the matrix above.

.github/workflows/linux.yml
```diff
@@ -80,6 +80,15 @@ jobs:
         [ ! -f ${PATCH_PREFIX}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}.patch
         [ ! -f ${PATCH_PREFIX}-${{ matrix.build-type }}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}-${{ matrix.build-type }}.patch
 
+    - name: Set up vcpkg
+      uses: lukka/run-vcpkg@v7.1
+      with:
+        setupOnly: true
+        vcpkgDirectory: '${{ github.workspace }}/3rd-party/vcpkg'
+        appendedCacheKey: ${{ hashFiles( '**/vcpkg_manifest/vcpkg.json' ) }}
```
Saviq, line 88: This should probably be `vcpkg.json`, as we have no `vcpkg_manifest`?

.github/workflows/linux.yml
```diff
@@ -80,6 +81,20 @@ jobs:
         [ ! -f ${PATCH_PREFIX}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}.patch
         [ ! -f ${PATCH_PREFIX}-${{ matrix.build-type }}.patch ] || patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}-${{ matrix.build-type }}.patch
 
+    - name: Set up vcpkg archive cache
+      id: setup-vcpkg-cache
```
Saviq, line 85:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -80,6 +81,20 @@ jobs:
...
+      run: |
+        mkdir --parents ${{ env.VCPKG_INSTALL_CACHE }}
+        sg lxd -c '/snap/bin/lxc profile device add default vcpkg disk source=${{ env.VCPKG_INSTALL_CACHE }} path=/root/parts/multipass/build/vcpkg_installed'
```
Saviq, line 88:
```suggestion
        sg lxd -c '/snap/bin/lxc profile device add default vcpkg disk source=${VCPKG_INSTALL_CACHE} path=/root/parts/multipass/build/vcpkg_installed'
```

.github/workflows/linux.yml
```diff
@@ -80,6 +81,20 @@ jobs:
...
+
+    - name: Set up vcpkg
+      uses: lukka/run-vcpkg@v7.1
+      with:
+        setupOnly: true
+        vcpkgDirectory: '${{ github.workspace }}/3rd-party/vcpkg'
+        appendedCacheKey: ${{ runner.os }}-${{ matrix.build-type }}-${{ hashFiles( '**/vcpkg.json' ) }}
+        additionalCachedPaths: ${{ env.VCPKG_INSTALL_CACHE }}
```
Saviq, line 96: Does this make sense? Do we need vcpkg outside of the LXD container?

If not, you can use the cache action directly:
```yaml
    - name: CCache
      uses: actions/cache@v2
      with:
        key: vcpkg-${{ runner.os }}-${{ matrix.build-type }}-${{ hashFiles( '**/vcpkg.json' ) }}
        path: ${{ env.VCPKG_INSTALL_CACHE }}
```

snap/snapcraft.yaml
```diff
@@ -94,6 +94,10 @@ parts:
     - pkg-config
     - qtbase5-dev
     - qtbase5-dev-tools
+    - curl
+    - zip
+    - unzip
+    - tar
```
Saviq, line 100: vcpkg needs all these? Please sort.

Saviq (changes requested):

.github/workflows/linux.yml
townsend2010: I'm not I understand why the environment variable itself can be used here, but the previous `mkdir` command needs the context expansion?

snap/snapcraft.yaml
townsend2010: Yes and will do.

.github/workflows/linux.yml
townsend2010: As discussed in chat, the `run-vcpkg` action uses the `cache` action internally.

.github/workflows/linux.yml
Saviq: It doesn't need it, I could've updated it, too.

`${{ }}` are processed before it ever gets to BASH.

.github/workflows/linux.yml
Saviq: More importantly, the `vcpkg` executable gets built in this step, and is shipped along with the source into the snap. Which I dislike a lot ;)

If there's a way to change where vcpkg gets built, in a way that makes it not pollute the source, but rather it getting shipped along with `vcpkg_installed`, that would be nice.

.github/workflows/linux.yml
townsend2010: Ok, good to know.  I'll fix these up.

.github/workflows/linux.yml
townsend2010: Ok, I see your point and will see if there is a way to massage `vcpkg` to not build directly in its submodule directory.  If so, then I may be able to dump the `run-vcpkg` action and just use the `cache action` along with any of our own logic.

.github/workflows/linux.yml
townsend2010: Done

townsend2010:
Ok, got this in a state where I can come back to it later, but going to shelve it for now.

townsend2010:
Killing this as it's not worth the effort.

PR #2011, opened 2021-03-16
[daemon] switch over to HTTPS for cdimage.u.c

Saviq:
It has HTTPS now.


ricab (approved):
Worked well with both qemu and lxd. I suppose we don't want to test configuration, so approving.

bors r+

PR #2021, opened 2021-03-22
[lxd] revert chunked reads

Saviq:
This didn't fix the problem in the end, and introduced another.

Kept this commit, though: 43748d8fd3a28e71c641888899aa59a4057bca87


townsend2010 (approved):
Yeah, unfortunately this caused more issues than it "tried" to fix :frowning_face: 

bors merge

PR #2024, opened 2021-03-23
[snap] clean up plugs.json

Saviq:


ricab (approved):
Thanks, this makes much more sense to me now. 

bors r+

PR #2025, opened 2021-03-23
[lxd] read all data from the socket

Saviq:
We've only been reading once, when sometimes the socket could be buffered.

### DIFF
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Canonical, Ltd.
+ * Copyright (C) 2020-2021 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/network/local_socket_reply.cpp
+++ b/src/network/local_socket_reply.cpp
@@ -259,6 +259,14 @@ void mp::LocalSocketReply::read_reply()
 
     do
     {
+        if (reply_offset + local_socket->bytesAvailable() > len)
+        {
+            setError(UnknownContentError, "Response buffer overflow");
+            emit error(UnknownContentError);
+            setFinished(true);
+            emit finished();
+            return;
+        }
         bytes_read = local_socket->read(data_ptr, len);
 
         reply_offset += bytes_read;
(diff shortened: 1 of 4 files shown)


src/network/local_socket_reply.cpp
```diff
@@ -252,17 +254,33 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
 void mp::LocalSocketReply::read_reply()
 {
-    auto data_ptr = reply_data.data();
+    auto data_ptr = reply_data.data() + reply_offset;
     int bytes_read{0};
 
     do
     {
-        bytes_read = local_socket->read(reply_data.data(), len);
+        if (reply_offset + local_socket->bytesAvailable() > len)
```
townsend2010, line 262: Instead of an error, we could resize the `QByteArray` instead?  `len` is really just an arbitrary size.

townsend2010 (changes requested):
Thanks a looks good overall.  One request and an inline question.

Could you please update the copyright date in `local_socket_reply.[h|cpp]` and `local_socket_server_test_fixture.h`.

src/network/local_socket_reply.cpp
Saviq: Done.

Saviq (commented):
> Could you please update the copyright date in `local_socket_reply.[h|cpp]` and `local_socket_server_test_fixture.h`.

Done.

src/network/local_socket_reply.cpp
```diff
@@ -252,17 +254,31 @@ void mp::LocalSocketReply::send_request(const QNetworkRequest& request, QIODevic
...
 
+        reply_offset += bytes_read;
         data_ptr += bytes_read;
     } while (bytes_read > 0);
+}
+
+void mp::LocalSocketReply::read_finish()
+{
+    if (local_socket->bytesAvailable())
+        read_reply();
```
townsend2010, line 278: Any way to write a test to cover this line?  If it's too much work, then I say forget since we have plans for this to go away.

Saviq: No, it's internal to Qt, since we're talking to it over a socket, we've no control of when these signals get fired.

townsend2010: Ack

townsend2010 (approved):
LGTM, thanks!

bors merge

Saviq:
bors retry

PR #2028, opened 2021-03-31
Add helper types to singleton mocks

ricab:
Add and use singleton-mock public types, to facilitate injecting into test fixture classes (whose attributes need a fixed type).

### DIFF
--- a/tests/mock_singleton_helpers.h
+++ b/tests/mock_singleton_helpers.h
@@ -26,20 +26,32 @@
 #include <utility>
 
 #define MP_MOCK_SINGLETON_INSTANCE(mock_class)                                                                         \
+public:                                                                                                                \
     static mock_class& mock_instance()                                                                                 \
     {                                                                                                                  \
         return dynamic_cast<mock_class&>(instance());                                                                  \
     }
 
+#define MP_MOCK_SINGLETON_HELPER_TYPES(mock_class, parent_class)                                                       \
+private:                                                                                                               \
+    static constexpr auto academy = [] { return sg::make_scope_guard([] { parent_class::reset(); }); }; /*             \
+                                    use a lambda so that the type gets deduced before decltype below */                \
+                                                                                                                       \
+public:                                                                                                                \
+    using Guard = decltype(academy());                                                                                 \
+    using GuardedMock = std::pair<mock_class*, Guard>;
+
 #define MP_MOCK_SINGLETON_INJECT(mock_class, parent_class)                                                             \
-    [[nodiscard]] static auto inject()                                                                                 \
+public:                                                                                                                \
+    [[nodiscard]] static GuardedMock inject()                                                                          \
     {                                                                                                                  \
         parent_class::reset();                                                                                         \
         parent_class::mock<mock_class>();                                                                              \
-        return std::make_pair(&mock_instance(), sg::make_scope_guard([]() { parent_class::reset(); }));                \
+        return std::make_pair(&mock_instance(), academy());                                                            \
     } // one at a time, please!
 
 #define MP_MOCK_SINGLETON_BOILERPLATE(mock_class, parent_class)                                                        \
+    MP_MOCK_SINGLETON_HELPER_TYPES(mock_class, parent_class)                                                           \
     MP_MOCK_SINGLETON_INSTANCE(mock_class)                                                                             \
     MP_MOCK_SINGLETON_INJECT(mock_class, parent_class)
 
--- a/tests/mock_singleton_helpers.h
+++ b/tests/mock_singleton_helpers.h
@@ -58,7 +58,7 @@ public:                                                                         
 namespace multipass::test
 {
 
-template <typename ConcreteMock, template <typename MockClass> typename MockCharacter = ::testing::NaggyMock>
+template <typename ConcreteMock, template <typename /*MockClass*/> typename MockCharacter = ::testing::NaggyMock>
 class MockSingletonHelper : public ::testing::Environment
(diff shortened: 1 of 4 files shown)


tests/mock_singleton_helpers.h
```diff
@@ -26,27 +26,39 @@
 #include <utility>
 
 #define MP_MOCK_SINGLETON_INSTANCE(mock_class)                                                                         \
+public:                                                                                                                \
```
townsend2010, line 29: For my knowledge, could you please explain why adding `public` here is necessary?  Is it that it defaults to `private` if not explicit even though this is free static function?

tests/mock_singleton_helpers.h
```diff
@@ -26,27 +26,39 @@
...
     static mock_class& mock_instance()                                                                                 \
     {                                                                                                                  \
         return dynamic_cast<mock_class&>(instance());                                                                  \
     }
 
+#define MP_MOCK_SINGLETON_HELPER_TYPES(mock_class, parent_class)                                                       \
+private:                                                                                                               \
+    static constexpr auto academy = [] { return sg::make_scope_guard([] { parent_class::reset(); }); }; /*             \
```
townsend2010, line 37: Interesting name for a lambda.  Just curious how you arrived at this name :smile:

tests/mock_singleton_helpers.h
```diff
@@ -26,27 +26,39 @@
...
 #define MP_MOCK_SINGLETON_BOILERPLATE(mock_class, parent_class)                                                        \
+    MP_MOCK_SINGLETON_HELPER_TYPES(mock_class, parent_class)                                                           \
     MP_MOCK_SINGLETON_INSTANCE(mock_class)                                                                             \
     MP_MOCK_SINGLETON_INJECT(mock_class, parent_class)
 
 namespace multipass::test
 {
 
-template <typename ConcreteMock, template <typename MockClass> typename MockCharacter = ::testing::NaggyMock>
+template <typename ConcreteMock, template <typename /*MockClass*/> typename MockCharacter = ::testing::NaggyMock>
```
townsend2010, line 61: I see the `MockClass` template parameter has been commented out here and elsewhere.  Is having the template parameter still necessary or can it be removed altogether?  I'm not seeing it being used anywhere else but I'm probably missing it somewhere.

townsend2010 (commented):
Hey @ricab, thanks for this!

I just have a few inline questions before approving this.

tests/mock_singleton_helpers.h
ricab: With the order this is used in below, and since "includers" of this file currently use only the composite `BOILERPLATE` macro, this has no effect. However, these more elementary macros are as available as the composite one. For instance, we could use `MP_MOCK_SINGLETON_INSTANCE` in the `MockProcessFactory` today without the rest of the boilerplate.

Before this, the macros left it to the client to choose what access section to place them in. But that `academy` thing is really just an implementation detail which we should not leak into the caller's public members. Yet once we make it private, there is no way to revert to the previous access class (unless the macros received it as a parameter, which might be excessive). 

IOW, with the `private:` in line 36, the macro now potentially changed the access section at the call site. We know the new types are needed publicly, so I added that `public:` in line 40, but we can't leave the choice of access class to the caller anymore (we'd still be changing from `private:` or `protected:`). A consistent interface in all the macros requires the client to buy into the access class when using the macros, so I made that explicit by adding a public in those that remained. 

Another `public:` in the composite macro is probably in order as well, so I'll just add that in another commit, but let me know if this makes sense, or if you think I should pass `public` to the macro, or something else.

tests/mock_singleton_helpers.h
ricab: Hehe, I just wanted the name of the entity that produces "guards" and that was the best I could think of. I also thought a little humor wouldn't hurt :smile: I had "barracks" before even. I also thought of "squad" or "squadron" but went with this.

However, if it's unclear I'm ofc fine with changing. Would you have any suggestions?

tests/mock_singleton_helpers.h
ricab: That inner `template <typename>` is needed for the MockCharacter because Naggy, Nice, and Strict are themselves templates. However, the name of the parameter itself is unused, and the IDE was pointing it out, so I just commented it out.

ricab (commented):
Hey @townsend2010, thanks for the review! Here are my replies.

tests/mock_singleton_helpers.h
townsend2010: Ah, ok, that makes sense.  Thanks!

tests/mock_singleton_helpers.h
townsend2010: Ah, I see, like a military academy.  Very clever :smiley:  I think the name is fine given that I'm clued into the humor, but for other readers who may not "get it", perhaps just a little comment above saying something like `Produces "guards"` would help?

tests/mock_singleton_helpers.h
townsend2010: Thanks for the detailed explanation.  That makes perfect sense and I'm good with this approach.

tests/mock_singleton_helpers.h
ricab: Sure, I will add that.

townsend2010: Thanks!

townsend2010 (approved):
Great, thanks!

bors merge

PR #2029, opened 2021-03-31
[test] Extract property matching for exceptions

ricab:
Another ancillary PR that I am using for bridging tests, providing a helper function to match against runtime_error messages, often needed with the extra assertions.


townsend2010 (approved):
LGTM, thanks!

bors merge

PR #2032, opened 2021-04-01
[cli] Allow different params in Command ctors

ricab:
Perfect forward whatever arguments are needed to Command constructors. Could be useful for passing a client-owned `AliasDict` ref in #1986.


ricab:
Example usage: https://github.com/canonical/multipass/compare/perfect-fwd-cmd...perfect-fwd-cmd-demo

PR #2034, opened 2021-04-01
[cmake] Catch base classes with non-virtual dtors

ricab:
The whole `-Weffc++` doesn't work for us because:

> When selecting this option, be aware that the standard library headers do not obey all of these guidelines; use grep -v to filter out those warnings.

Puzzling why `-Wnon-virtual-dtor` is not part of `-Wall` or even `-Wextra`. It built fine here and we could remove it if it started failing on external headers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright © 2017-2020 Canonical Ltd.
+# Copyright © 2017-2021 Canonical Ltd.
 #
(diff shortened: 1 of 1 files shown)


ricab:
Thanks @AlanGriffiths!

Saviq:
bors merge

Saviq:
bors retry

Saviq:
Grrrh.

bors retry

PR #2040, opened 2021-04-05
[feature] Add Multipass Workflows

townsend2010:
---

TODO:

- [x] Check for and retrieve new workflows zip file
- [ ] Change workflows URL to whatever our official Github repository will be
- [x] Add environment variable override for URL
- [x] Add error handling and state if URL or zip file is unavailable
- [x] Add daemon unit tests
- [x] Add private side (TODO's for this are being tracked in that PR)
- [x] Add Poco libraries to snap package
- [x] ~~Optimize `URLDownloader` to check for `etag` header~~ This is to be a separate PR independent of Workflows.


Saviq (changes requested):
Having tried a `file:///...` URL for the override, I got this error:

```plain
find failed: bad conversion
```

To be more future-proof, let's tweak the `cloud-init` bit to be:
```yaml
cloud-init:
  user-data: |
    #cloud-config
    packages: [cowsay]
```

We'll need to integrate with #2046, otherwise:
```
$ multipass launch minikube -n minikube
```

We should name the launched instance after the workflow by default.

But then, ultimately:
```
$ multipass exec minikube -- minikube status
minikube
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured
```

So \o/.

src/utils/memory_size.cpp
```diff
@@ -31,6 +31,9 @@ constexpr auto giga = mega * kilo;
 
 long long in_bytes(const std::string& mem_value)
 {
+    if (mem_value.empty())
+        return 0LL;
```
Saviq, line 35: Hmm unsure about that. Shouldn't it be an error? Like… `""` is more `NULL` than `0` :P

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,263 @@
...
+const QString workflow_dir_version{"v1"};
+constexpr auto category = "workflow provider";
+
+auto workflows_map_for(const std::string& archive_file_path)
+{
+    std::map<std::string, std::string> workflows_map;
+    std::ifstream zip_stream{archive_file_path, std::ios::binary};
+    Poco::Zip::ZipArchive zip_archive{zip_stream};
+
+    for (auto it = zip_archive.headerBegin(); it != zip_archive.headerEnd(); ++it)
```
Saviq, line 55: Should we do any error handling here to see if the zip is bad? There's a [`Poco::Zip::ZipException`](https://pocoproject.org/docs/Poco.Zip.ZipException.html)?

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+        if (it->second.isFile())
+        {
+            auto file_name = it->second.getFileName();
+            QFileInfo file_info{QString::fromStdString(file_name)};
+
+            if (file_info.path().contains(workflow_dir_version) && file_info.suffix() == "yaml")
```
Saviq, line 62: Could also allow `yml` here.

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,263 @@
...
+            {
+                Poco::Zip::ZipInputStream zip_input_stream{zip_stream, it->second};
+                std::ostringstream out(std::ios::binary);
+                Poco::StreamCopier::copyStream(zip_input_stream, out);
+                workflows_map[file_info.baseName().toStdString()] = out.str();
```
Saviq, line 67: Should we do any validation on the basename?

Saviq (changes requested):
I had a problem on boot, where `multipass find` didn't show the workflows. I had to restart it twice (and wait?) for them to show up. And no errors in the logs, like it never decided to go and fetch them.

Some comments inline, moar tomorrow o/

include/multipass/exceptions/workflow_exceptions.h
```diff
@@ -0,0 +1,45 @@
...
+    {
+    }
+};
+
+class InvalidWorkflowException : public std::runtime_error
+{
+public:
+    InvalidWorkflowException(const std::string& error_msg) : runtime_error(error_msg)
+    {
+    }
```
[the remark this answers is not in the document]
townsend2010: Ah, right, use the inherited class's ctor.

src/utils/memory_size.cpp
townsend2010: I just knew this would generate some controversy :wink: I do understand your reluctance here, but I'll try to explain.

I found it necessary to know if memory options were given in the cli when launching a workflow or not.  As can be seen here (lines 2147 & 2148 specifically):
 
https://github.com/canonical/multipass/blob/b3fa1a1452f0001f3284590fbfc0ddf6fd7e7941/src/daemon/daemon.cpp#L2145-L2159

it creates memory entries of 0 bytes which infers that the user did not pass a memory argument.

I don't see any negative side effects when using a `MemorySize` class that is initialized with an empty string to say that it holds 0 bytes as its value.

That said if you (or anyone else on the team) really thinks this is a bad idea, then I can certainly revert this and change the aforementioned lines to look something like:
`MemorySize{request->mem_size().empty() ? "0b" : request->mem_size()};`
which accomplishes the same thing at the expense of some readability.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Yeah, I think it would be prudent to do that.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ack

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: I'm not sure I follow.  Like what would we be validating?

townsend2010:
Hey @Saviq,

Thanks for the quick first pass.

> Having tried a `file:///...` URL for the override, I got this error:

Ok, will look into that.
> To be more future-proof, let's tweak the `cloud-init` bit to be:

Sure, makes sense.
> We'll need to integrate with #2046, otherwise:

Maybe something like a `timeout` field in the yaml definition?
> We should also add a `version` field, as this looks a bit sad :)

Agreed, but wasn't part of the original MVP :stuck_out_tongue_winking_eye: I'll add it though.
> I had a problem on boot, where `multipass find` didn't show the workflows. I had to restart it twice (and wait?) for them to show up. And no errors in the logs, like it never decided to go and fetch them.

Weird, I've not seen that behavior.  Will look into it though.

src/utils/memory_size.cpp
Saviq: Well, maybe instead we should have an empty `MemorySize`?

src/workflow_provider/default_vm_workflow_provider.cpp
Saviq: Basically it should be a valid hostname, since we want to use it as the instance name (and by extension, its hostname).

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,263 @@
...
+}
+
+mp::Query mp::DefaultVMWorkflowProvider::fetch_workflow_for(const std::string& workflow_name,
+                                                            VirtualMachineDescription& vm_desc)
+{
+    update_workflows();
+
+    Query query{"", "", false, "", Query::Type::Alias};
+    auto& config = workflow_map.at(workflow_name);
+    auto workflow_config = YAML::Load(config);
```
Saviq, line 100: Feels weird to be parsing it JIT, there's a lot of `YAML::Load(config)` below. I suppose it depends what we want to optimize for - memory or CPU usage…

What we should do, though, is handle any errors in the YAML. Everywhere below you assume the YAML is correct with our schema. But what if it's not? Some more tests for that would be in order, too.

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    if (workflow_instance["limits"]["min-mem"])
+    {
+        auto min_mem_size_str{workflow_instance["limits"]["min-mem"].as<std::string>()};
+
+        try
+        {
+            MemorySize min_mem_size{min_mem_size_str};
+
+            if (vm_desc.mem_size.in_bytes() == 0)
```
Saviq, line 155: So yeah, I'd be happier with `.empty()` or so here.

tests/test_daemon.cpp
```diff
@@ -526,6 +573,83 @@ TEST_P(DaemonCreateLaunchTestSuite, adds_pollinate_user_agent_to_cloud_init_conf
...
+        .WillOnce([&mem_size, &disk_space](const mp::VirtualMachineDescription& vm_desc, auto&) {
+            EXPECT_EQ(vm_desc.num_cores, num_cores);
+            EXPECT_EQ(vm_desc.mem_size, mem_size);
+            EXPECT_EQ(vm_desc.disk_space, disk_space);
+
+            return std::make_unique<mpt::StubVirtualMachine>();
+        });
+
+    EXPECT_CALL(*mock_image_vault, fetch_image(_, _, _, _))
+        .WillOnce(Invoke([&release, &remote](const mp::FetchType& fetch_type, const mp::Query& query,
```
Saviq, line 598: Do you need the `Invoke()` here (and further down) for some reason? I thought just putting a lambda in `WillOnce()` would work?

ref.: https://google.github.io/googletest/gmock_cheat_sheet.html#using-a-function-functor-or-lambda-as-an-action

Saviq (changes requested):
> Maybe something like a timeout field in the yaml definition?

Yup.

> To be more future-proof, let's tweak the cloud-init bit to be:
> 
> ```yaml
> cloud-init:
>   user-data: |
>     #cloud-config
>     packages: [cowsay]

> Sure, makes sense.

Actually, it doesn't - it should be `vendor-data` :)

Some more comments inline.

tests/test_daemon.cpp
townsend2010: You're right, I think I got on a copy & pasta kick and just threw the `Invoke` in these places without prejudice :wink: Will remove these.

townsend2010:
> Actually, it doesn't - it should be `vendor-data`

Right :)

src/utils/memory_size.cpp
townsend2010: Ok, that makes sense and will do that instead.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ah, ok, got it and yes, will do that.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Sure, it's certainly possible that the map stores the YAML node instead.

And yes, we should catch YAML exceptions when converting to a type, log it and invalidate the particular workflow due to the error.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ack

src/utils/memory_size.cpp
townsend2010: Looking at this again a bit more, the `MemorySize` class is not really meant to be "empty" in the sense of not having any value.  Further, the `MemorySize()` constructor (without an argument) defaults to 0 bytes:
https://github.com/canonical/multipass/blob/b3fa1a1452f0001f3284590fbfc0ddf6fd7e7941/src/utils/memory_size.cpp#L68-L70

The functions that are called to get memory sizes in different units all return a `long long` so there is no concept of empty with that unless we specify something like `-1` is considered empty.  So I see these as our options:

- Revert it back to the original where a passed in empty string string is an error and throws an exception and I just do the `MemorySize{request->mem_size().empty() ? "0b" : request->mem_size()};` for my needs.
- Keep it as I have it now in this PR.
- Use `-1` to indicate empty when an empty string is passed in.

I really don't have a strong opinion in any of these options since in the end, they all basically accomplish the same thing for my needs.

townsend2010: Ok, will do, but we'll have to decide what an "empty" state is for something that returns `long long` in all of its public functions.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: So given that I'm reverting the `MemorySize` change back to the original, then we'll have to leave this be in order to determine if a user provided a memory size option on the `launch` cli.  If a user really did specify 0 bytes on the cli, that will be caught like it is currently.

.gitmodules
```diff
@@ -24,3 +24,6 @@
 [submodule "3rd-party/fmt"]
 	path = 3rd-party/fmt
 	url = https://github.com/fmtlib/fmt.git
+[submodule "3rd-party/poco"]
+	path = 3rd-party/poco
+	url = https://github.com/pocoproject/poco.git
```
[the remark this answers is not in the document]
townsend2010: Done

tests/test_daemon.cpp
townsend2010: Ok, these have been removed for this PR.

src/utils/memory_size.cpp
townsend2010: This has been reverted back to its original behavior.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Done

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: I will note that POCO is quite heavy handed on how it handles some error conditions and will call `std::abort()` if an assertion fails.  There is really no way for us to recover from such a case either.  For example, say a zip file is corrupted and is missing some expected info in the zip archive, POCO will fail an assertion and call `std::abort()` and we will bail.  Can't say I really like that approach from POCO as I think it should throw an exception instead and let the consumer of the library deal with it as it sees fit.

townsend2010: Here's an example in POCO about what I'm describing above:
https://github.com/pocoproject/poco/blob/3fc3e5f5b8462f7666952b43381383a79b8b5d92/Zip/src/ZipArchive.cpp#L78
and this is what `poco_assert()` does:
https://github.com/pocoproject/poco/blob/3fc3e5f5b8462f7666952b43381383a79b8b5d92/Foundation/include/Poco/Bugcheck.h#L95

townsend2010:
> Having tried a `file:///...` URL for the override, I got this error:

I'm unable to reproduce this.  What steps did you do?

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Scratch the last two things I posted.  The assertions with `std::abort()` shouldn't be too much of a problem.  I do see where it should throw an exception when the zip is not correct.  However, I'm directly modifying a zip file to put in a bad header and POCO doesn't seem to notice, so I'll look more into that.

townsend2010: Ok, doing some debugging in POCO, it has a bug in how it handles bad headers, etc. in a zip file.  I'll probably see if I can fix that sometime and upstream it.  I'll also see if there is an existing bug and if not, enter one in the POCO project.

That said, there won't be a good way to test zip exceptions other than put the zip stuff under control of a singleton and inject mocks to simulate exceptions.  I'll see about that.

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -50,24 +50,36 @@ auto workflows_map_for(const std::string& archive_file_path)
...
         }
     }
+    catch (const Poco::Exception& e)
+    {
+        mpl::log(mpl::Level::error, category, fmt::format("Error extracting Workflows zip file: {}", e.displayText()));
+    }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::error, category, fmt::format("Error generating Workflows map: {}", e.what()));
+    }
```
Saviq, line 82: This feels like too wide a net? Should we not propagate unexpected exceptions up?

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -50,24 +50,36 @@ auto workflows_map_for(const std::string& archive_file_path)
...
             {
-                Poco::Zip::ZipInputStream zip_input_stream{zip_stream, it->second};
-                std::ostringstream out(std::ios::binary);
-                Poco::StreamCopier::copyStream(zip_input_stream, out);
-                workflows_map[file_info.baseName().toStdString()] = out.str();
+                auto file_name = it->second.getFileName();
+                QFileInfo file_info{QString::fromStdString(file_name)};
+
+                if (file_info.path().contains(workflow_dir_version) &&
+                    (file_info.suffix() == "yaml" || file_info.suffix() == "yml"))
```
Saviq, line 65: `badv1yaml.yaml` would be considered OK here. Should be strict about `^v1/([a-z0-9-]+).(yaml|yml)$` or so.

tests/test_workflow_provider.cpp
```diff
@@ -307,3 +310,35 @@ TEST_F(VMWorkflowProvider, downloadFailureDuringUpdateLogsErrorAndDoesNotThrow)
...
+
+    EXPECT_CALL(*mock_poco_zip_utils, zip_archive_for(_)).WillOnce(Throw(std::runtime_error(error_msg)));
+
+    auto logger_scope = mpt::MockLogger::inject();
+    logger_scope.mock_logger->screen_logs(mpl::Level::error);
+    logger_scope.mock_logger->expect_log(mpl::Level::error,
+                                         fmt::format("Error generating Workflows map: {}", error_msg));
+
+    EXPECT_NO_THROW(mp::DefaultVMWorkflowProvider(workflows_zip_url, &url_downloader, cache_dir.path(),
+                                                  std::chrono::milliseconds(0)));
```
Saviq, line 343: Yeah, unsure about that…

Saviq (changes requested):

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: I still have yet to implement the validation of the file name as you suggested previously.  This is **only** to allow the `yaml` and `yml` suffixes.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Hmm, yeah, I think you're right.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: But yeah, I see what you mean about validating the whole path a little better :smiley:

Saviq:
> > Having tried a `file:///...` URL for the override, I got this error:
> 
> I'm unable to reproduce this. What steps did you do?

Can't reproduce any more. Works with a local file, must've done something wrong.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: This should be addressed now.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: I've (hopefully) addressed the handling of errors now.

Regarding the current JIT approach, I'm wondering if it should store the loaded `YAML::Node` in the map or actually parse the YAML and store the associated data structures (like `VirtualMachineDescription`, etc.) in the map?  WDYT?

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ok, got rid of this.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ok, looking a this some more, sure, `badv1yaml.yaml` would indeed be considered OK here as long as the **path** (not just the file name) has `v1` in it based on [`QFileInfo::path()`](https://doc.qt.io/qt-5/qfileinfo.html#path).  I could a bit better here to make sure the parent directory of the file is indeed `v1`, so instead of `QFileInfo::path().contains()`, I should use `QFileInfo::dir().dirName() == workflow_dir_version`.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Looking at this some more, I think it would be better to validate the key name we use under `instances` in the yaml rather than the file's base name.  We could allow a bit more flexibility in the file name, but the instance name in the yaml will need to conform to a valid host name.  WDYT?

townsend2010:
Ok, there is been a lot of activity on here, so this is to summarize what I believe is left at the time of this writing.

- `multipass find` not showing any workflows for a period time, particularly seen in CI
- Adding support to base instance name on workflow name.  [There is a question](https://github.com/canonical/multipass/pull/2040#discussion_r621561761) about where the instance name should come from (file name or workflow name under the `instances`.
- Add `timeout` key under instance configuration.  Really need #2046 merged first before adding the full support for this.
- [Answer question](https://github.com/canonical/multipass/pull/2040#discussion_r621404981) about how better to cut down on JIT parsing of YAML.
- Address review in #2055 since it's the prerequisite branch.
- ????

src/workflow_provider/default_vm_workflow_provider.cpp
Saviq: Either way works… No preference, really.

src/workflow_provider/default_vm_workflow_provider.cpp
Saviq: I don't think it makes much sense to give that flexibility. If you `multipass launch minikube`, and we want the hostnames to be `minikube[-$suffix]`, no point introducing more names.

Saviq (commented):
>     * Add `timeout` key under instance configuration.  Really need #2046 merged first before adding the full support for this.

That one landed. The order of precedence should be:
1. user-given `--timeout`,
2. workflow declared `timeout:` (it will have a different meaning, though),
3. default.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ok, after a brief chat, I understand what we need to do here.  Thanks!

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ok, pushed a commit to just store the `YAML::Node` in the map.  We can optimize further later if need be.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Ok, file name-> host name validation and using the workflow name as the instance name by default is done.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: This is done.

townsend2010:
> I had a problem on boot, where multipass find didn't show the workflows. I had to restart it twice (and wait?) for them to show up. And no errors in the logs, like it never decided to go and fetch them.

I've tried and tried to reproduce this, going as far as purging the Multipass snap, installing it, and trying `find` and it Just Works.  I'm not how else to debug this.

Saviq (changes requested):
Some more things inline

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,376 @@
...
+            query.release = tokens[0];
+        }
+        else
+        {
+            needs_update = true;
+            throw InvalidWorkflowException("Unsupported image scheme in Workflow");
+        }
+    }
+
+    if (workflow_instance["limits"]["min-cpu"])
```
Saviq, line 142: What if `["limits"]` isn't there?

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,376 @@
...
+        {
+            vm_desc.vendor_data_config = YAML::Load(workflow_instance["cloud-init"]["vendor-data"].as<std::string>());
+        }
+        catch (const YAML::BadConversion&)
+        {
+            needs_update = true;
+            throw InvalidWorkflowException(
+                fmt::format("Cannot convert cloud-init data for the {} workflow", workflow_name));
+        }
+    }
```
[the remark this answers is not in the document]
townsend2010: Oh, right, I'm overwriting our stuff since it's prepopulated and need to merge this instead.  Will fix.

src/daemon/daemon.cpp
```diff
@@ -2012,15 +2056,14 @@ void mp::Daemon::create_vm(const CreateRequest* request, grpc::ServerWriter<Crea
...
-
     if (!checked_args.option_errors.error_codes().empty())
     {
         return status_promise->set_value(grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, "Invalid arguments supplied",
                                                       checked_args.option_errors.SerializeAsString()));
     }
 
-    auto name = name_from(checked_args.instance_name, *config->name_generator, vm_instances);
+    auto name = name_from(checked_args.instance_name, config->workflow_provider->name_from_workflow(request->image()),
+                          *config->name_generator, vm_instances);
```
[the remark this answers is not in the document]
townsend2010: Yes, I went round and round about this and `timeout`, but that will take a non-trivial amount of refactoring.  Instead, it was much more straightforward to add some accessors for these bits of data.  I will add a TODO at this point since you are OK with that :wink:

include/multipass/utils.h
```diff
@@ -154,7 +156,7 @@ void multipass::utils::try_action_for(OnTimeoutCallable&& on_timeout, std::chron
         if (try_action(std::forward<Args>(args)...) == TimeoutAction::done)
             return;
 
-        std::this_thread::sleep_for(1s);
+        std::this_thread::sleep_for(timeout < 1s ? timeout : 1s);
```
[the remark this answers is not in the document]
townsend2010: Ok, will add a note.

src/daemon/daemon.cpp
```diff
@@ -812,6 +820,17 @@ void add_aliases(mp::FindReply& response, const std::string& remote_name, const
     }
 }
 
+auto timeout_for(const int requested_timeout, const int workflow_timeout)
+{
+    if (requested_timeout > 0)
+        return std::chrono::seconds(requested_timeout);
```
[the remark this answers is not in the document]
townsend2010: Yep, planned on adding a parametrized test for all this timeout logic to ensure it's covered correctly.

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,376 @@
...
+      workflows_ttl{workflows_ttl}
+{
+    try
+    {
+        update_workflows();
+    }
+    catch (const std::exception& e)
+    {
+        mpl::log(mpl::Level::error, category, fmt::format("Error on workflows start up: {}", e.what()));
+    }
```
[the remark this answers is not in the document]
townsend2010: Since this is the ctor, if we don't catch everything, then it will cause daemon start up to bail and `multipassd` won't be running.  Is that what we want?

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    try
+    {
+        image_info.version = QString::fromStdString(workflow_config["version"].as<std::string>());
+    }
+    catch (const YAML::BadConversion&)
+    {
+        needs_update = true;
+        throw InvalidWorkflowException(fmt::format(bad_conversion_template, version_key, workflow_name));
+    }
```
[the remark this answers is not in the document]
townsend2010: Yes, I thought of that, but I think that is something to optimize for after this PR gets in.  I'll add a `TODO` if you are good with that.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: I'm pretty sure it behaves exactly how it does now, but I'll make sure and add a test workflow that does not define a `limits` section.

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
Saviq: We could avoid calling `update_workflows` in the c'tor then? We do have the `top_catch_all` bit to deal with at least some of that.

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
Saviq: Yeah fine, no point spending time on abstracting something we want to get rid of anyway.

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
townsend2010: I can certainly remove the preloading of workflows on daemon startup, but I don't understand why we would want to do that.  The way `DaemonConfig()` (which calls the `Workflow` ctor amongst many others) is currently architected, any exceptions are not caught and the daemon will bail.  I think the idea was to let the various classes determine if exceptions are catastrophic and let the daemon bail.  For `DefaultVMWorkflowProvider`, I don't see anything that should be catastrophic- the daemon should run if Workflows are unavailable for any reason which I why I have this in the ctor.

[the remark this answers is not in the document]
Saviq: The problem is, as I see it, you're left with a potentially broken workflow provider object. I'd be OK catching network, YAML parsing, ZIP errors, etc. Those we can recover from. But can we recover from _any_ exception?

If we caught "unexpected" exceptions in `DaemonConfig()`, we could NULL the provider and just not use it. But we shouldn't catch them all and ignore - and continue to use a potentially broken provider.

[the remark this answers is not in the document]
townsend2010: Ok, I'll narrow the net to known exceptions, but I don't think this is the place to fix how `DaemonConfig()` doesn't handle exceptions that get bubbled up.

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
townsend2010: Should be fixed now.

include/multipass/default_vm_workflow_provider.h
```diff
@@ -0,0 +1,67 @@
...
+#include <QString>
+#include <QUrl>
+
+#include <chrono>
+#include <map>
+
+namespace multipass
+{
+// TODO: Change to correct Workflows URL once the official repo is set up
+const QString default_workflow_url{"https://codeload.github.com/townsend2010/multipass-workflows/zip/refs/heads/main"};
```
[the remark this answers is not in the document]
townsend2010: Ok, I pushed this as a commit in my branch to remove the comment above as well.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
townsend2010: Added `TODO`'s here and for timeout.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
townsend2010: Added a test that should better cover this.

include/multipass/utils.h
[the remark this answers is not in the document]
townsend2010: Done

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
townsend2010: Added a `TODO` at the top.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Yeah, if `limits` is not defined, this still picks that up.  I added a test workflow to cover this particular case.

src/workflow_provider/default_vm_workflow_provider.cpp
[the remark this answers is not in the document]
townsend2010: I just removed the `try`/`catch` here and fixed the associated test.  `update_workflows()` catches the known exceptions, so any others will just go up to `DaemonConfig::build()`.

src/workflow_provider/default_vm_workflow_provider.cpp
```diff
@@ -0,0 +1,378 @@
...
+        {
+            auto cloud_init_config = YAML::Load(workflow_instance["cloud-init"]["vendor-data"].as<std::string>());
+
+            for (const auto& config : cloud_init_config)
+            {
+                if (config.first.IsScalar())
+                {
+                    vm_desc.vendor_data_config[config.first.Scalar()] = config.second;
+                }
+            }
```
Saviq, line 218: This may need to be smarter at some point. Users may need to be part of the workflow's vendor config.

Saviq (approved):
Looking good :)

I'm running Integration here: https://github.com/canonical/multipass/actions/runs/811268224

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: > Users may need to be part of the workflow's vendor config.

I'm not sure I understand what you mean here.  Could you please elaborate?

Saviq:
Hmm on macOS:

```plain
$ multipass launch minikube --timeout 2400
Creating minikube /
launch failed: '' is not a supported alias. Please use `multipass find` for supported image aliases.
```

townsend2010:
> Hmm on macOS:

I think this is on all platforms actually.  I see in the `minikube` workflow you pushed to the official repo, there is not an `image` key and I think I assumed there would be.  So will work on that tomorrow to make it be the `default` image when that is not defined.

src/workflow_provider/default_vm_workflow_provider.cpp
Saviq: I just mean we'll need to be smarter about _adding_ a user, rather than replacing the whole `users:` bit, and such. But later :)

Saviq:
> I think this is on all platforms actually. I see in the `minikube` workflow you pushed to the official repo, there is not an `image` key and I think I assumed there would be.

I was going off of this :)

https://github.com/canonical/multipass/blob/2e9620e7972505c90450a554f5e430ab755f1c62/src/workflow_provider/default_vm_workflow_provider.cpp#L113

> So will work on that tomorrow to make it be the `default` image when that is not defined.

Yup, cool.

Saviq:
w00t!

https://github.com/canonical/multipass/runs/2507994331

bors merge

Saviq:
bors r-

Needs a merge on the other side.

townsend2010:
> I was going off of this :)

Right, it was just a bad assumption on my part is all.

src/workflow_provider/default_vm_workflow_provider.cpp
townsend2010: Oh, right, add to a list, etc.

Saviq (approved):
And yup!

CI +1 here: https://github.com/canonical/multipass/actions/runs/813769521
And here needed a LXD re-run 'cause it got too busy…: https://github.com/canonical/multipass/actions/runs/813876303

bors merge

Saviq:
bors r-

There's a conflict still :/

retr0h:
@townsend2010 have you thought of adding a command section to the workflow?  I've written a tool which behaves similarly to a workflow, with the exception of command execution sections. 

An example workflow:

* Boot multipass VM
* Install k3s via cloud init
* Copy kubeconfig to system invoking multipass `multipass transfer <instance>:/etc/rancher/k3s/k3s.yaml /local/path/k3s.yaml`
* Enable regcred acces to docker instance `multipass exec <instance> sudo docker login <registry_server> -u $ENV[user] -p ENV[pass]`
* Create k8s regcred secret `sudo kubectl create secret generic regcred --from-file=.dockerconfigjson=/root/.docker/config.json --type=kubernetes.io/dockerconfigjson`

The point being there may be lightweight orchestration events that would be helpful to run on create and outside cloud-init.

```
description: <string>      # * a short description of the workflow ("tagline")
version: <string>          # * a version string

instances:
  <name>:                  # * equal to the workflow name
    image: <base image>    # a valid image alias, see `multipass find` for available values
    limits:
      min-cpu: <int>       # the minimum number of CPUs this workflow can work with
      min-mem: <string>    # the minimum amount of memory (can use G/K/M/B suffixes)
      min-disk: <string>   # and the minimum disk size (as above)
    timeout: <int>         # maximum time for the instance to launch, and separately for cloud-init to complete
    cloud-init:
      vendor-data: |       # cloud-init vendor data
        <string>
    command:
      <string>                # string of shell commands to be executed locally by the user invoking multipass.
```

PR #2065, opened 2021-04-20
[logging] Fix some small UI issues in client logging

townsend2010:


src/process/basic_process.cpp
```diff
@@ -214,10 +214,9 @@ void mp::BasicProcess::setup_child_process()
 void mp::BasicProcess::handle_started()
 {
     pid = process.processId(); // save this, so we know it even after finished
-    const auto& program = qUtf8Printable(process_spec->program());
-
+    const auto program = process_spec->program().toStdString();
```
ricab, line 217: This is ok, but the ref wasn't wrong either: const refs extend the lifetime of temporaries. That's why we can do `f("literal")` when we have `voic f(const string&)`. The same is true for rvalue refs (lvalue refs can't be bound to temporaries). A couple of references: [GotW88](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/) & [cppref](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary).

src/process/basic_process.cpp
```diff
@@ -214,10 +214,9 @@ void mp::BasicProcess::setup_child_process()
...
     mpl::log(mpl::Level::debug, program,
-             fmt::format("[{}] started: {} {}", pid, program, qUtf8Printable(process_spec->arguments().join(' '))));
+             fmt::format("[{}] started: {} {}", pid, program, process_spec->arguments().join(' ')));
```
ricab, line 219: Why remove `qUtf8Printable`s?

ricab (approved):
Thanks, this looks good and I no longer get the SSL corruption. I have a comment and a question inline, but I am good with it as it is.

src/process/basic_process.cpp
townsend2010: Sure, I can (and should) leave the ref to the `std::string` now.  I'll explain below why I removed the `qtf8Printable`s.

src/process/basic_process.cpp
townsend2010: Hey @ricab, so, the symptom was that the `program` variable was not printing out expected values and instead, had *some* data that was in `process_spec->arguments().join(' ')`.  This really had me scratching my head as to how in the world this was occurring.  I did some digging and found that since `program` was set by the `qUtf8Printable(process_spec->program())` call previously, the next `qUtf8Printable(process_spec->arguments().join(' ')` call would modify the `program` pointer.  I'm not really sure if this intended behavior in Qt or a bug, but I found the best way to get around this is just to use `std::string`s instead and let the `CString` class handle that.

townsend2010:
Hey @ricab,

Thanks for the review!  There were certainly some strange issues that I've (hopefully) addressed in this PR.   I meant to mention in the description, but this PR depends on #2055, so we'll wait on that PR to get in better getting this in.

src/process/basic_process.cpp
townsend2010: Ok, fixed this.  Thanks!

src/process/basic_process.cpp
ricab: OK, bit of an enigma I see. Anyway, I think it's good like this.

ricab:
Hey @townsend2010, yeah I saw the stacked PRs. I will leave merging to you when you're ready, if that's ok.

bors d+

townsend2010:
@ricab
> I will leave merging to you when you're ready, if that's ok.

Perfect, thanks :smiley:

townsend2010:
bors r=ricab

PR #2066, opened 2021-04-21
[daemon] fix invalid Json reference

Saviq:
---

```gcc
multipass/src/daemon/daemon.cpp:218:26: error: loop variable 'entry' is always a copy because the range of type 'QJsonArray' does not return a reference [-Werror,-Wrange-loop-analysis]
        for (const auto& entry : record["extra_interfaces"].toArray())
                         ^
multipass/src/daemon/daemon.cpp:218:14: note: use non-reference type 'QJsonValueRef'                                                                                                                     
        for (const auto& entry : record["extra_interfaces"].toArray())
             ^~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -215,7 +215,7 @@ std::vector<mp::NetworkInterface> read_extra_interfaces(const QJsonObject& recor
 
     if (record.contains("extra_interfaces"))
     {
-        for (const auto& entry : record["extra_interfaces"].toArray())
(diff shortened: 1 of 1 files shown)


ricab (approved):
Ohh interesting. Thanks. Why didn't we see this before?

Saviq:
> Ohh interesting. Thanks. Why didn't we see this before?

We did: #1798 :)

It's only new macOS (10.15) Clang that complains.

PR #2074, opened 2021-04-26
[launch] add support for `--bridged` network

Saviq:
---

For context: the other options we were considering (`--public`, `--exposed`) would not be intuitive to anyone. Even if "bridging" isn't strictly correct, e.g. on Hyper-V, that is the term other hypervisors use and people that know what it is will likely know from the get go what it's about. Even ourselves we refer to the feature as "bridging", because that's the prevalent term.

I decided against validating on `multipass set local.bridged-interface` as the given interface may not exist at that point in time for whatever reason, and on `launch` we'll communicate well what the problem is, if any.

### DIFF
--- a/include/multipass/constants.h
+++ b/include/multipass/constants.h
@@ -43,6 +43,8 @@ constexpr auto winterm_profile_guid =
 
 constexpr auto petenv_key = "client.primary-name";     // This will eventually be moved to some dynamic settings schema
 constexpr auto driver_key = "local.driver";            // idem
+constexpr auto bridged_interface_key = "local.bridged-interface"; // idem
+constexpr auto bridged_network_name = "bridged";
 constexpr auto autostart_key = "client.gui.autostart"; // idem
 constexpr auto winterm_key = "client.apps.windows-terminal.profiles"; // idem
 constexpr auto hotkey_key = "client.gui.hotkey";                      // idem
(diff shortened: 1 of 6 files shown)


luis4a0:
Thanks for this @Saviq! The name `--bridged` sounds the best option to me.

The code looks good, I only have a minor concern with respect to the functionality: what happens if the host has an interface named 'bridged'? I think we should provide a way to bridge it. I know this is unlikely, but it's still possible.

One possibility I see to address this corner case is to allow `--network id=bridged`. WDYT?

Saviq:
> One possibility I see to address this corner case is to allow `--network id=bridged`. WDYT?

That would require us to change the RPC and CLI (`--bridged` and `--network bridged` would no longer be shortcuts for `id=bridged`) for tiny gain, IMO not worth it, but let me know if you think otherwise.

luis4a0:
Well, I don't think this is strictly necessary at that cost. It would be nice to have if easier to implement. Let's leave it like it is now and if some user complains at some point, we can think again about it.

Saviq:
@luisp I added the tweak to only restart in the daemon if the driver changed. Unfortunately I was unable to add coverage of the two cases yet, but I didn't want to block on that. I did add a test for the case when we _do_ want to restart, which wasn't there before.

luis4a0:
Looks good, thanks!

Saviq:
@luis4a0 I reverted the recent changes, went back to just adding the setting and `--bridged` option. I'll work on preventing the unnecessary restart in a separate PR.

Saviq:
bors retry

Saviq:
bors r=luis4a0,Saviq

Saviq:
bors retry

Saviq:
bors r=luis4a0,Saviq

Saviq:
bors retry

Saviq:
One last time.

bors retry

PR #2076, opened 2021-04-27
[util] Fix eager timeouts

ricab:
Avoid timing out up to 1 second early. Prefer erring on the long side.

Found this out when reviewing #2046.


Saviq:
+1, yeah.

bors merge

PR #2078, opened 2021-04-29
Create and use bridge

ricab:
Creates bridges if needed, uses existing otherwise.

### DIFF
--- a/src/platform/backends/shared/base_virtual_machine_factory.h
+++ b/src/platform/backends/shared/base_virtual_machine_factory.h
@@ -44,6 +44,10 @@ public:
         return {};
     };
 
+    void prepare_networking(std::vector<NetworkInterface>& extra_interfaces) override
+    {
+    }
+
     VMImageVault::UPtr create_image_vault(std::vector<VMImageHost*> image_hosts, URLDownloader* downloader,
                                           const Path& cache_dir_path, const Path& data_dir_path,
                                           const days& days_to_expire) override
--- a/src/platform/backends/shared/base_virtual_machine_factory.h
+++ b/src/platform/backends/shared/base_virtual_machine_factory.h
@@ -44,7 +44,7 @@ public:
         return {};
     };
 
-    void prepare_networking(std::vector<NetworkInterface>& extra_interfaces) override
+    void prepare_networking(std::vector<NetworkInterface>& /*extra_interfaces*/) override
     {
     }
 
--- a/src/platform/backends/shared/base_virtual_machine_factory.h
+++ b/src/platform/backends/shared/base_virtual_machine_factory.h
@@ -46,6 +46,7 @@ public:
 
     void prepare_networking(std::vector<NetworkInterface>& /*extra_interfaces*/) override
     {
+        // only certain backends need to do anything to prepare networking
     }
 
     VMImageVault::UPtr create_image_vault(std::vector<VMImageHost*> image_hosts, URLDownloader* downloader,
--- a/src/platform/platform_linux.cpp
+++ b/src/platform/platform_linux.cpp
@@ -115,10 +115,16 @@ mp::NetworkInterfaceInfo get_network(const QDir& net_dir)
     static const auto bridge_fname = QStringLiteral("brif");
 
     std::string type, description;
+    std::vector<std::string> links;
     if (auto bridge = "bridge"; net_dir.exists(bridge))
     {
         type = bridge;
         QStringList bridge_members = QDir{net_dir.filePath(bridge_fname)}.entryList(QDir::NoDotAndDotDot | QDir::Dirs);
+
+        links.reserve(bridge_members.size());
+        std::transform(bridge_members.cbegin(), bridge_members.cend(), std::back_inserter(links),
+                       [](const QString& interface) { return interface.toStdString(); });
+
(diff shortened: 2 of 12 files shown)


ricab:
Done, apart from tests.

src/platform/backends/shared/base_virtual_machine_factory.h
```diff
@@ -44,6 +44,10 @@ class BaseVirtualMachineFactory : public VirtualMachineFactory
         return {};
     };
 
+    void prepare_networking(std::vector<NetworkInterface>& extra_interfaces) override
+    {
+    }
+
```
Saviq, line 50: Do we want this to be in the base? I wonder if we should be explicit about this being empty on a given platform?

src/platform/platform_linux.cpp
```diff
@@ -115,10 +115,16 @@ mp::NetworkInterfaceInfo get_network(const QDir& net_dir)
...
     std::string type, description;
+    std::vector<std::string> links;
     if (auto bridge = "bridge"; net_dir.exists(bridge))
     {
         type = bridge;
         QStringList bridge_members = QDir{net_dir.filePath(bridge_fname)}.entryList(QDir::NoDotAndDotDot | QDir::Dirs);
+
+        links.reserve(bridge_members.size());
+        std::transform(bridge_members.cbegin(), bridge_members.cend(), std::back_inserter(links),
+                       [](const QString& interface) { return interface.toStdString(); });
```
Saviq, line 126: This will become unwieldy quickly:
```plain
$ multipass networks
Name             Type      Description
br-enx00133b490  bridge    Network bridge with enx00133b4909cd, tap3b37457c, tap82f64e6f
```

For each instance we'll add a `tap` interface. Maybe we should filter virtual interfaces out? I think for now I'd rather not have any interfaces listed than all of them.

Saviq (changes requested):

src/platform/platform_linux.cpp
ricab: Righhht, my bad, forgot about those. Let me fix that.

src/platform/backends/shared/base_virtual_machine_factory.h
ricab: I wondered the same, but ended up deciding against duplicating code. Figured that's what inheritance/base class implementations are for. Notice this is still separate from the interface (pure base) class.

I actually once asked the same (for [this](https://github.com/canonical/multipass/blob/6d86dc451ee89d90cb881653307bb2c7c5894682/src/platform/backends/shared/base_virtual_machine_factory.h#L61) from @luis4a0 in a review but ended up backtracking and apologizing :sweat_smile: 

WDYT? Would a comment like `// only certain backends need to do anything to prepare networking` help?

I do have to comment out that unused param in any case.

Saviq: I suppose it's just a "unimplemented" vs. "default behaviour" decision. Will leave it to you :)

src/platform/platform_linux.cpp
```diff
@@ -110,32 +110,50 @@ bool is_ethernet(const QDir& net_dir)
...
+            auto is_unknown = [&networks](const std::string& id) {
+                auto same_as = [&id](const auto& other) { return other.first == id; };
+                return std::find_if(networks.cbegin(), networks.cend(), same_as) == networks.cend();
+            };
+            links.erase(std::remove_if(links.begin(), links.end(), is_unknown),
+                        links.end()); // filter links to networks we don't recognize
+
+            assert(net.type == "bridge");
+            net.description =
+                links.empty() ? "Empty network bridge"
```
Saviq, line 153: This should probably just say "Network bridge" - it's not necessarily empty:

```plain
$ multipass networks
Name             Type      Description
enx00133b4909cd  ethernet  Ethernet device
lxdbr0           bridge    Empty network bridge
mpbr0            bridge    Network bridge for Multipass
virbr0           bridge    Empty network bridge

$ brctl show
bridge name     bridge id               STP enabled     interfaces
lxdbr0          8000.00163e092819       no              veth3fe7576d
mpbr0           8000.00163eba1a1b       no
virbr0          8000.525400f99bdb       yes             virbr0-nic
```

ricab: Hmm, right, now that we're filtering things we shouldn't be so specific in that description. I'll update that.

ricab: Done.

src/platform/platform_linux.cpp
ricab: Done.

Saviq:
w00t.

bors merge

ricab:
bors cancel

src/platform/backends/shared/base_virtual_machine_factory.h
ricab: I added the comment to hopefully make it a little clearer.

ricab:
Let's see if this works after cancel.

bors retry

ricab:
Doesn't look like it.

bors r=Saviq

ricab:
bors cancel

ricab:
alright, lets go again

bors r=Saviq

PR #2081, opened 2021-04-30
Client-daemon ping-pong to ask the user for bridging authorization

ricab:


ricab:
Yeyy I just made [CI hit](https://github.com/canonical/multipass/pull/2081/checks?check_run_id=2479864435#step:11:4769) a [bug in clang](https://bugs.llvm.org/show_bug.cgi?id=35450) 🎰

ricab:
OK, barring tests, this should be about ready to review.

src/client/cli/cmd/launch.cpp
```diff
@@ -471,3 +443,83 @@ mp::ReturnCode cmd::Launch::request_launch(const ArgParser* parser)
...
+
+    return OptInStatus::UNKNOWN;
+}
+
+bool cmd::Launch::ask_bridge_permission(multipass::LaunchReply& reply)
+{
+    static constexpr auto plural = "Multipass needs to create {} to connect to {}.\nThis will temporarily disrupt "
+                                   "connectivity on those interfaces.\n\nDo you want to continue (yes/no)?";
+    static constexpr auto singular = "Multipass needs to create a {} to connect to {}.\nThis will temporarily disrupt "
+                                     "connectivity on that interface.\n\nDo you want to continue (yes/no)?";
```
Saviq, line 499:
```suggestion
    static constexpr auto plural = "Multipass needs to create {} to connect to {}.\nThis will temporarily disrupt "
                                   "connectivity on those interfaces.\n\nDo you want to continue (yes/no)? ";
    static constexpr auto singular = "Multipass needs to create a {} to connect to {}.\nThis will temporarily disrupt "
                                     "connectivity on that interface.\n\nDo you want to continue (yes/no)? ";
```

Don't want to make `Yes` the default?

include/multipass/cli/command.h
```diff
@@ -148,13 +149,37 @@ class Command
...
-        static_assert(FailureCallableTraits::num_args == 1, "");
-        static_assert(std::is_same<typename SuccessCallableTraits::return_type, ReturnCode>::value, "");
-        static_assert(std::is_same<typename FailureCallableTraits::return_type, ReturnCode>::value, "");
-        static_assert(std::is_same<FailureCallableArgType, grpc::Status&>::value, "");
+        static_assert(std::is_same<typename SuccessCallableTraits::return_type, ReturnCode>::value);
+        static_assert(std::is_same<typename FailureCallableTraits::return_type, ReturnCode>::value);
+
+        static_assert(SuccessCallableTraits::num_args == 1);
+        static_assert(std::is_base_of_v<google::protobuf::Message, SuccessCallableArg0Type>,
+                      "`on_success` doesn't receive a Message");
```
Saviq, line 160: Would this be clearer?
```suggestion
                      "`on_success` should receive a Message");
```

include/multipass/cli/command.h
```diff
@@ -148,13 +149,37 @@ class Command
...
+
+        if constexpr (FailureCallableTraits::num_args != 1)
+        {
+            static_assert(FailureCallableTraits::num_args == 2, "`on_failure` needs to take either 1 or 2 parameters");
+            using FailureCallableArg1Type =
+                std::remove_reference_t<typename FailureCallableTraits::template arg<1>::type>;
+            static_assert(std::is_same_v<SuccessCallableArg0Type, FailureCallableArg1Type>,
+                          "`on_success` and `on_failure` handle different reply types");
```
Saviq, line 168: And this?
```suggestion
                          "`on_success` and `on_failure` should handle the same reply types");
```

src/client/cli/cmd/launch.cpp
```diff
@@ -401,7 +367,7 @@ mp::ReturnCode cmd::Launch::request_launch(const ArgParser* parser)
         return ReturnCode::Ok;
     };
 
-    auto on_failure = [this](grpc::Status& status) {
+    auto on_failure = [this, &parser](grpc::Status& status, mp::LaunchReply& reply) {
```
Saviq, line 370: You need to pause the timer here, too?

```plain
$ multipass launch --network enx00133b4909cd --timeout 1 
Multipass needs to create a bridge to connect to enx00133b4909cd.
This will temporarily disrupt connectivity on that interface.

Timed out waiting for instance launch.
```

Saviq (changes requested):
```plain
$ multipass launch --network enx00133b4909cd --network enx0023564c7940
Multipass needs to create bridges to connect to enx00133b4909cd, enx0023564c7940.
This will temporarily disrupt connectivity on those interfaces.

Do you want to continue (yes/no)?y
Launched: efficient-treefrog                                                    
$ multipass list
efficient-treefrog      Running           10.152.94.213    Ubuntu 20.04 LTS
                                          10.2.0.5
                                          10.2.0.21
```

include/multipass/cli/command.h
ricab: Sure, I'll change that.

include/multipass/cli/command.h
ricab: Yup

src/client/cli/cmd/launch.cpp
ricab: Ahh yes, good catch!

src/client/cli/cmd/launch.cpp
ricab: We could. Personally I thought the inconvenience was worth it here, to impress the (potential) weight of the decision on the user (who knows whatever is going on that depends on networking). But it's not a strong preference. 

Let me know if you think the default would be best. We can also ask others what they think.

ricab: Note that the other question doesn't have a default either. There I guess we could safely set it to "Later".

Saviq: Right, "Later" is not really a default, is it ;)

ricab: I am not sure I get what you mean. Syntactically, it is one of the possible replies. Semantically, "no decision yet"  is arguably the best default decision :)

ricab: > arguably the best default decision

I'm thinking: the one that assumes the least, so has the least compromise.

Saviq: Yeah, good as is.

src/client/cli/cmd/launch.cpp
ricab: Done.

ricab:
Just waiting for bors before a final rebase (but conflicts already sorted).

ricab:
Rebased on `staging`, which should be promoted to `main` soon.

Saviq (approved):
All good!

bors merge

Saviq:
You, too.

bors retry

PR #2084, opened 2021-05-03
[utils] try and reduce Timer tests runtime

Saviq:


tests/test_timer.cpp
```diff
@@ -118,22 +119,22 @@ TEST_F(Timer, restarts)
     int count = 0;
     std::mutex count_m;
 
-    mpu::Timer t{2s, [&count, &count_m]() {
+    mpu::Timer t{duration_cast<seconds>(1000ms), [&count, &count_m]() {
```
ricab, line 122: Why not just `1s`?

And does this really need to be 1000 times bigger than the other ones? Why the difference?

ricab (commented):
Nice, this is a very welcome improvement. I just didn't get the difference in restart.

tests/test_timer.cpp
Saviq: > Why not just `1s`?

Fixed.

> And does this really need to be 1000 times bigger than the other ones? Why the difference?

That's the lowest I found remotely reliable. The overhead of multiple consecutive `std::this_thread::sleep`s and `:;wait`s in the timer, context switching, makes it for a racy situation. As mentioned, the only way I can think of to make this fully reliable is to do away with any sleeps altogether and synchronizing between the timer and test manually.

ricab (approved):
Alrighty then, let's get it in!

bors r+

PR #2090, opened 2021-05-06
[daemon] don't restart on bridged network change

Saviq:


src/platform/backends/lxd/lxd_virtual_machine_factory.h
```diff
@@ -59,6 +63,7 @@ class LXDVirtualMachineFactory final : public BaseVirtualMachineFactory
     NetworkAccessManager::UPtr manager;
     const Path data_dir;
     const QUrl base_url;
+    static constexpr auto backend_name = "lxd";
```
ricab, line 66: :+1:

PR #2091, opened 2021-05-06
[cli] Recover newline after log

ricab:


ricab:
Abort. We found that this was intended, after all.

Issue #1781, opened 2020-10-06
[snap] need to reach out of $SNAP for host properties


Saviq:
All of our instances are reported as launched on `ubuntu-core-18` now, due to strict confinement and `base: core18`. We should try and find a way to reach out to get the actual host OS.

https://github.com/canonical/multipass/blob/d9a72641a254a0613fe4344aa3cb93a67a6b055b/src/daemon/daemon.cpp#L116-L117

townsend2010:
Qt reads `/etc/os-release` when it exists and uses that.  Of course, this is the Ubuntu Core info.  However, `/etc/lsb-release` within the snap does have actual host OS info.  Sadly, I don't see a way to force Qt to use this :frowning_face: so it looks like we'll need our own solution for the Linux case.

Saviq:
I wonder if it's a `snapd` bug, then, and it should actually expose host's `/etc/os-release`…

Saviq:
LXD has this:

https://github.com/snapcore/snapd/blob/5145669ea40e4b02f02605f41b456df540ecb5c9/interfaces/builtin/lxd_support.go#L45-L46

townsend2010:
Given that the output of `/etc/os-release` looks like this:
```
NAME="Ubuntu Core"
VERSION="18"
ID=ubuntu-core
PRETTY_NAME="Ubuntu Core 18"
VERSION_ID="18"
HOME_URL="https://snapcraft.io/"
BUG_REPORT_URL="http://bugs.launchpad.net/snappy/"
```
I would say that's pretty intentional.  That said, I agree that if it's not the actual Ubuntu Core distribution running (and not the core snap), then it should probably have what the host has in there.

Right about LXD, but that still doesn't help with Qt (it reads a hard-coded `/etc/os-release`) which is why I mentioned that we'd need our own Linux solution.

Saviq:
> I would say that's pretty intentional.

Well, that would be true on a Ubuntu Core 18 system, it is not, on a Classic system with the core18 snap, IMO. And then the discrepancy between lsb- and os-release…

townsend2010:
> Well, that would be true on a Ubuntu Core 18 system, it is not, on a Classic system with the core18 snap, IMO. And then the discrepancy between lsb- and os-release…

Right, what I was trying to say :smile:

townsend2010:
[...]

PR #2096, opened 2021-05-08
<lsb-release> reader #1781

surahman:
**_With respect to issue #1781:_**
I have developed tests and utility routines to read the `lsb-release` file on Ubuntu-based systems when applicable. When it is unavailable, or parsing fails, there will be a failover to the previous method of using `SysInfo::productType()` and `QSysInfo::productVersion()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,8 @@ parts
 prime
 stage
 snap/.snapcraft/state
+
+# build artifacts
+build/
+multipass-build-deps*
+
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -125,6 +125,8 @@ QString make_uuid();
 template <typename OnTimeoutCallable, typename TryAction, typename... Args>
 void try_action_for(OnTimeoutCallable&& on_timeout, std::chrono::milliseconds timeout, TryAction&& try_action,
                     Args&&... args);
+std::pair<QString, QString> read_LSB_Release(const QString& path = "/etc/lsb-release");
+std::pair<QString, QString> parse_LSB_Release(const QStringList& lsb_file_data, const char& delimiter = '=');
 
 } // namespace utils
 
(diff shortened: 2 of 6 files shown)


ricab:
Hi @surahman, thanks a lot for your work on this! 

Unfortunately, we've [discussed the disparity between `/etc/lsb-release` and `/etc/os-release`](https://forum.snapcraft.io/t/hosts-lsb-release-but-core-os-release/24351/) and it's not looking like we can rely on any of those. The fact that they're readable at all is misleading.

Instead, we should probably read from `/var/lib/snapd/hostfs/etc/os-release` and add the [`system-observe`](https://snapcraft.io/docs/system-observe-interface) plug to our snap. We'd then have to decide what to do if the user doesn't give us the plug. 

We're still discussing details, and we'll get back to you again, but we wanted to let you know so that you don't spend any more time.

ricab:
Hi again @surahman,

So we discussed a bit and confirmed `/var/lib/snapd/hostfs/etc/os-release` is the way to go, given how snaps are working... The implementation can just return `"unknown"` when unable to read that file (that's what `QSysInfo` defaults to as well).

Multipass needs to work on macOS/Windows too, so this would need to be a platform dependent implementation. That can be achieved by declaring a free function in `include/multipass/platform.h` and implementing it in `src/platform/platform_linux.cpp` (we'd take care of other platforms). The code you now have for parsing `lsb-release` would need to be adapted and moved to that compilation unit. The corresponding tests go in `tests/linux/test_platform_linux.cpp`.

Notice that the Linux implementation should still use `QSysInfo` when outide of a snap. So, something like

```c++
std::string multipass::platform::host_version() // returning single string probably enough for our purposes
{
  return mu::in_multipass_snap() ? 
    our_snap_impl_reading_os_release() :
    fmt::format("{}-{}", QSysInfo::productType(), QSysInfo::productVersion());
}
```

Let us know if you have questions and thanks again for your involvement!

surahman:
I am on it. I am happy to help and I am learning a lot from your codebase. These interactions and exercises are a great learning/practice tool and it is even better if it is of assistance to someone.

surahman:
I will go ahead and abandon this branch and PR to start on a new branch - it is the cleanest way to revert changes. I will cross-reference the two PR's once I have the new one up.

surahman:
I have created a new branch instead of doing a hard reset on this one: #2105

Saviq:
Thanks @surahman, I'll close this one then.

PR #2101, opened 2021-05-10
Fix eclipsing vms in ctor

### LINKED ISSUE #1658: Lost multipass instances on macOS
After a multipass crash and system restart the instances have disappeared.

[multipassd-last2000lines.log](https://github.com/canonical/multipass/files/4970444/multipassd-last2000lines.log)

**Additional info**
- OS: macOS 10.15.5
- multipass  1.3.0+mac

ricab:
Let exceptions that arise when creating a VM in the daemon constructor through. This avoids erasing instance data that could be recoverable and important for the user, but aborts the daemon since it can't honor its database, which would otherwise be out of sync. More robust approaches for particular causes can always added. Fixes #1658.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -901,6 +901,14 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         }
 
         auto vm_image = fetch_image_for(name, config->factory->fetch_type(), *config->vault);
+        if (!QFile::exists(vm_image.image_path))
+        {
+            mpl::log(mpl::Level::warning, category,
+                     fmt::format("Could not find image for '{}'. Expected location: {}", name, vm_image.image_path));
+            invalid_specs.push_back(name);
+            continue;
+        }
+
         const auto instance_dir = mp::utils::base_dir(vm_image.image_path);
         const auto cloud_init_iso = instance_dir.filePath("cloud-init-config.iso");
         mp::VirtualMachineDescription vm_desc{spec.num_cores,
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -953,6 +953,7 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
 
     for (const auto& bad_spec : invalid_specs)
     {
+        mpl::log(mpl::Level::warning, category, fmt::format("Removing invalid instance: {}", bad_spec));
         vm_instance_specs.erase(bad_spec);
     }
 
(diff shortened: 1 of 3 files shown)


townsend2010 (commented):
Although I think failing the daemon is better than deleting someone's instance wholesale, I'm still concerned that if the daemon can't start, then how does a user recover from this?

In the end, I'll approve this, but I'd like some feedback from the team here on how we plan to address a situation where the daemon just won't start due to hitting this.

Saviq:
> In the end, I'll approve this, but I'd like some feedback from the team here on how we plan to address a situation where the daemon just won't start due to hitting this.

Oh sure, this is a stop-gap. Long-term we should mark the instance "Broken" and skip over it. Another `multipass start` on it could recover it - if not, back to "Broken". Details TBD.

townsend2010:
> Oh sure, this is a stop-gap. Long-term we should mark the instance "Broken" and skip over it. Another `multipass start` on it could recover it - if not, back to "Broken". Details TBD.

Sure, I definitely understand this is a stop-gap and is better than just deleting the instance.  But we are trading deleting an instance wholesale with the potential of the daemon not starting at all (which is a better failure).  My question is, if/when users complain that Multipass is not starting at all, how will we deal with that?  Have users modify the instance DB to remove the problematic instance without deleting it in order to get the daemon to start?  Wing it and just cross that bridge when we get to it?

Again, I'm not saying this shouldn't go in, I just want us all to be on the same page about how we'll deal with the daemon not starting.  I guess where I'm going with this is we really need a full solution for this ASAP.

ricab:
There are two sorts of situations that this PR covers by aborting the daemon:

1. general problem preventing *any* instance from being launched (e.g. backend doesn't work)
2. a corrupt/buggy instance.

If I am not mistaken, most reports have been of the first kind. In such cases, marking all instances as BROKEN is not much better then quitting: if instances can't be launched, there isn't much that daemon could do anyway. That would be useful mainly for situation 2. But, depending on how frequent we find case 2 to be, maybe guiding users to remove/recover instances manually won't be that big a deal and that "full solution" may not be all that urgent. I do think it's less urgent than fixing permanently erased instances.

Of course, if we find specific measures to help on a case-by-case basis (e.g. a bigger timeout somewhere), we can implement them.

townsend2010 (approved):
Ok, this is better than deleting an instance.

bors merge

Saviq:
bors retry

Saviq:
bors retry

Saviq:
bors retry

Saviq:
bors retry

ricab:
This is getting funny :laughing:

townsend2010:
:facepalm:

Saviq:
Gah.

bors merge

Issue #1138, opened 2019-10-23
Allow disabling `primary` instance handling


Saviq:
I'm unlikely to ever want a `primary` instance as an Ubuntu user (if
there's something that I just want a pet Ubuntu instance for, I have my
desktop; anything more specific and I'm going to want to choose the
Ubuntu release I launch); is there any way I can configure multipass to
disable this behaviour?

_Originally posted by @OddBloke in https://github.com/CanonicalLtd/multipass/issues/1133#issuecomment-544975799_

Saviq:
I propose accepting `multipass set client.primary-name=` to set it to null and then handling that appropriately wherever it has special handling.

PR #2102, opened 2021-05-12
[cli] Toggle Primary Instance (petenv)

surahman:
**_Issue #1138:_**
I have added support to handle the `client.primary-name=` command. I am not sure about the mock tests I have set up and they would require some review. When stepping through the tests with a debugger I was unable to hit breakpoints on `set` and `set_aux` within the `utils/settings.cpp` file - which is typically an indicator of an issue. Manual argument passing tests work on the `cmdl`.


tests/test_mock_settings.cpp
```diff
@@ -47,4 +47,20 @@ TEST(Settings, can_have_get_mocked)
...
+TEST(Settings, set_aux_client_primary_name_null)
+{
+    EXPECT_NO_THROW(MP_SETTINGS.set("client.primary-name", ""));
+}
+
+TEST(Settings, set_aux_client_primary_name_not_null)
+{
+    EXPECT_NO_THROW(MP_SETTINGS.set("client.primary-name", "not null name"));
+}
+
```
ricab, line 65: The situation with `Settings` and tests is particular... We don't really have tests for the real `Settings` singleton, because it relies on `QSettings` which is not mockable. We'd have to provide another wrapper for `QSettings` which we could extend and mock and we never got around to that. 

OTH, we don't want tests to affect or be affected by real settings on the disk, so we make sure [all the tests run with a `MockSettings`](https://github.com/canonical/multipass/blob/c3d3afa0d62b9e3a766a769bd72d3d5f4ea84ac2/tests/main.cpp#L36) instead. This file has only a couple of very shallow tests for the mock itself, to show that it works when accessing the `Settings` singleton.

In short, what you wrote here only calls a mock and we don't have a way to actually test this ATM, so you can remove these entirely.

src/utils/settings.cpp
```diff
@@ -203,7 +203,12 @@ void multipass::Settings::set_aux(const QString& key, QString val) // work with
 {
     // TODO we should have handler callbacks instead
     if (key == petenv_key && !mp::utils::valid_hostname(val.toStdString()))
-        throw InvalidSettingsException{key, val, "Invalid hostname"};
+    {
+        if (val.isEmpty() && key != "client.primary-name") // Handle "multipass set client.primary-name=". Further checks required.
+        {
+            throw InvalidSettingsException{key, val, "Invalid hostname"};
+        }
+    }
```
ricab, line 211: Hmm, I don't think this is right. You want to accept empty values, so you should just add that condition to the first `if`:

```c++
if (key == petenv_key && !val.isEmpty() && !mp::utils::valid_hostname(val.toStdString()))
    throw ...
```

Note that `petenv_key` is the constant defining `"client.primary-name"` as the key we use for that.

src/client/cli/cmd/set.cpp
```diff
@@ -116,14 +116,16 @@ mp::ParseCode cmd::Set::parse_args(mp::ArgParser* parser)
...
         if (args.size() != 1)
         {
             cerr << "Need exactly one key-value pair (in <key>=<value> form).\n";
             status = ParseCode::CommandLineError;
         }
         else
         {
-            const auto keyval = args.at(0).split('=', QString::KeepEmptyParts);
+            const auto keyval = args.at(0).split('=', Qt::KeepEmptyParts);
+
```
ricab, line 128: It appears you added some stray newlines (in an otherwise empty commit).

ricab (changes requested):
Hi @surahman, thanks for the PR. I am afraid it doesn't cut it yet though. 

The logic for disabling primary would need to go in the commands that deal with it (along with the "if" condition below).

tests/test_mock_settings.cpp
surahman: Got it, thanks for the explanation!

src/utils/settings.cpp
surahman: Corrected, I had not looked up the `constants` prior.

src/client/cli/cmd/set.cpp
surahman: Cleaned up. The only intended change here was to silence a deprecated namespace warning.

surahman:
> The logic for disabling primary would need to go in the commands that deal with it (along with the "if" condition below).
I will correct the issue with the error being thrown about the missing settings key parameter.

ricab:
Hey @surahman, yeah I mean the places where the primary instance gets special treatment:

```
$ grep -r petenv_key src/client/ include/
src/client/gui/gui_cmd.cpp:    auto petenv_name = MP_SETTINGS.get(petenv_key).toStdString();
src/client/cli/cmd/restart.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
src/client/cli/cmd/stop.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
src/client/cli/cmd/suspend.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
src/client/cli/cmd/launch.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
src/client/cli/cmd/shell.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
src/client/cli/cmd/start.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
include/multipass/constants.h:constexpr auto petenv_key = "client.primary-name";     // This will eventually be moved to some dynamic settings schema
include/multipass/cli/format_utils.h:    const auto petenv_name = MP_SETTINGS.get(petenv_key).toStdString();
```

They would have to be modified to have the primary instance disabled. Otherwise this just sets the name to the empty string. 

(Note that, for historic reasons, the code often uses the term `petenv`, short for "pet environment", to designate the primary instance.)

src/client/cli/cmd/set.cpp
ricab: The problem is that doesn't work with earlier Qt versions:

```
./src/client/cli/cmd/set.cpp:126:59: error: ‘KeepEmptyParts’ is not a member of ‘Qt’
  126 |             const auto keyval = args.at(0).split('=', Qt::KeepEmptyParts);
      |                                                           ^~~~~~~~~~~~~~
```

So we decided to live with these warnings for now.

src/utils/settings.cpp
```diff
@@ -202,7 +202,7 @@ QString mp::Settings::get_client_settings_file_path() // idem
 void multipass::Settings::set_aux(const QString& key, QString val) // work with a copy of val
 {
     // TODO we should have handler callbacks instead
-    if (key == petenv_key && !mp::utils::valid_hostname(val.toStdString()))
+    if (key == petenv_key && !val.isEmpty() && !mp::utils::valid_hostname(val.toStdString()))
```
ricab, line 205: OK, so only this change for now. Would you mind squashing all the commits so far accordingly? Thanks.

surahman:
> ```
> $ grep -r petenv_key src/client/ include/
> src/client/gui/gui_cmd.cpp:    auto petenv_name = MP_SETTINGS.get(petenv_key).toStdString();
> src/client/cli/cmd/restart.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
> src/client/cli/cmd/stop.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
> src/client/cli/cmd/suspend.cpp:    const auto petenv_name = MP_SETTINGS.get(petenv_key);
> src/client/cli/cmd/launch.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
> src/client/cli/cmd/shell.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
> src/client/cli/cmd/start.cpp:    petenv_name = MP_SETTINGS.get(petenv_key);
> include/multipass/constants.h:constexpr auto petenv_key = "client.primary-name";     // This will eventually be moved to some dynamic settings schema
> include/multipass/cli/format_utils.h:    const auto petenv_name = MP_SETTINGS.get(petenv_key).toStdString();
> ```

@ricab thanks for the information. How would you like to have the `primary` is disabled situation dealt with? Is the correct of dealing with this to just return from the function with a `success` (potentially a silent error) or a `failure` (input command/request invalid) code?

ricab:
Hi @surahman, this is quite tricky. if the primary instance is disabled, multipass needs to disable any special treatment for it and error out in commands that need an instance but get none. So, if the user said for instance `multipass start`, without an instance name, multipass would indeed need to error out. Commands like start, stop, shell... would require at least one instance argument. Pretty much like a `multipass delete` would, but perhaps mentioning that the primary instance is disabled in the error message. The `help` messages would also need to be updated, where they mention primary today.

ricab:
BTW, I haven't looked at your new code yet. Please let me know when you feel it's ready for another review pass.

surahman:
@ricab I have gone ahead and updated the tests to include parameters. ATM `start` errors out if a command is passed without a parameter.

The other commands are also set to error out if the `petenv` is disabled. I gather that you would like a message to be sent to `cerr` and the command to return `ok` instead? Are there any example tests on how to capture the messages sent to the streams?

I will look through the `help` messages to locate instances of `primary`.

ricab:
Hmm, I don't quite follow, but primary is an instance that is automatically launched, started, stoped, etc. when you use the corresponding commands without arguments. And it's name can change, so that after `multipass set client.primary_name=heyo`, `multipass start # no args` launches an instance named `heyo`, `multipass stop` would stop it, etc.  BTW, have you seen https://multipass.run/docs/primary-instance ?

So, to disable it, `stop`, `start`, etc. should now error out, since they don't know what instance to work with. So that's what the tests should check, for instance:

```c++
    EXPECT_CALL(mock_settings, get(Eq(mp::petenv_key))).WillRepeatedly(Return(""));
    EXPECT_THAT(send_command({"start" /* no further args */ }), Eq(mp::ReturnCode::CommandFail));
```

while
```c++
    EXPECT_CALL(mock_settings, get(Eq(mp::petenv_key))).WillRepeatedly(Return("something"));
    EXPECT_THAT(send_command({"start" /* no further args */ }), Eq(mp::ReturnCode::CommandSuccess));
```

The `launch` command is different, because it expects an image name rather than an instance name. So `multipass launch # no args` is unrelated to primary.

BTW, the help messages are dynamically generated, so you should look for `petenv_key` rather than `primary`.

surahman:
@ricab I think it is ready for a review. The handling is mostly complete except for the cases where the `primary` is disabled and we have flags such as `--all`. As it stands right now, this will cause an error, even if there are other active instances with the disabled `primary`.

ricab:
> (Jarjar Binks is happy)

Great to hear that @surahman :wink: 

Thank you again for this! I will have to queue reviews for some days to advance on other fronts, but I want to keep answering your questions if you have any in the meantime. So keep asking away and I'll reply as I can.

> For the others, I will need to handle the disabled primary in the parse_args to check for additional passed parameter counts when necessary. Still drop an error message on cerr?

> [...] except for the cases where the primary is disabled and we have flags such as --all

I am afraid I still don't entirely understand the question, so I'll give a broader reply in the hope that it covers it. 

From the top of my head, the only particularity of `start` (along with `shell`), with respect to the primary instance, is that it launches it if it does not exist. This is not applicable to `restart` as things stand today because only running instances are restartable. IIRC the only way `launch` treats the primary instance differently is by mounting the home folder. 

Other than that, the primary instance is just a default argument when no instance and no `--all` is passed. To disable primary, we need to refuse `start`, `stop`, `suspend`, etc. with no arguments and no `--all`. 

Note that an instance with the name "primary" (and whatever `client.primary_name` was before disabling) can still exist, be launched, stopped, and otherwise function normally, like any other instance. That's no different from what happens today when you have an instance named "primary" and change the primary name to, say, "foo". If you say `multipass stop`, "foo" will be stopped, "primary" will be untouched. And `--all` just keeps targeting all instances. 

Here is an expected interaction (messages not accurate):

```
$ multipass list
No instances
$ multipass stop --all
[NOOP]
$ multipass start --all
[NOOP]
$ multipass start
[launches primary, auto-mounts]
$ multipass set client.primary-name=foo
$ multipass stop
error - foo does not exist
$ multipass stop --all
[stops primary]
$ multipass start
[launches foo]
$ multipass start --all
[starts primary]
$ multipass delete -p primary
[deletes primary]
$ multipass launch -n primary
[launches primary, no auto-mount]
$ multipass set client.primary-name=
$ multipass stop
error: primary instance disabled, need an instance name or --all
$ multipass start
error: primary instance disabled, need an instance name or --all
$ multipass stop primary
[stops primary]
$ multipass stop foo
[stops foo]
$ multipass start --all
[starts foo and primary]
$ multipass delete --all --purge
[deletes foo and primary]
$ multipass launch -n primary
[launches an instance named primary, no auto-mount]
$ multipass launch -n foo
[launches an instance named foo, no auto-mount]
$ multipass set client.primary-name=primary
$ multipass stop
[stops primary]
```

surahman:
Thanks, I am trying to understand the valid `CLI` parameters without any experience with the `CLI`, so I am relying on documentation and tests like these:
```c++
TEST_F(Client, start_cmd_considers_configured_petenv)
{
    const auto custom_petenv = "jarjar binks";
    EXPECT_CALL(mock_settings, get(Eq(mp::petenv_key))).WillRepeatedly(Return(custom_petenv));

    const auto petenv_matcher = make_instance_in_repeated_field_matcher<mp::StartRequest, 1>(custom_petenv);
    EXPECT_CALL(mock_daemon, start(_, petenv_matcher, _));
    EXPECT_THAT(send_command({"start"}), Eq(mp::ReturnCode::Ok));
}
```

surahman:
> $ multipass set client.primary-name=
> $ multipass start
> error: primary instance disabled, need an instance name or --all

@ricab focusing on `start` here, I have the checks for passing the `--all` flag working but I now need to employ a check to see if other instances are available. Do I need to send an IPC `dispatch` request for a `list` to obtain a list of available instances, or is there some pre-built routine available to that end?

surahman:
Sorry, just renamed the branch it closed the link to this PR.

ricab:
Hi @surahman, why would you need that "check to see if other instances are available"? If you did, then having a common function would make sense, but I don't see why you would. A start command with no arguments will be wrong independently of what instances are available at the moment.

There is actually a common function to check the validity of the command in this respect, so you could drop [this if](https://github.com/canonical/multipass/pull/2102/files#diff-2868af61f453cf60edb3c228b594b5121226ece5e07637e0d4a3b9abc804e6faR179) and instead edit the call just below:

```diff
diff --git a/src/client/cli/cmd/start.cpp b/src/client/cli/cmd/start.cpp
index 1c94ecb..7609f09 100644
--- a/src/client/cli/cmd/start.cpp
+++ b/src/client/cli/cmd/start.cpp
@@ -173,7 +173,7 @@ mp::ParseCode cmd::Start::parse_args(mp::ArgParser* parser)
     if (status != ParseCode::Ok)
         return status;
 
-    auto parse_code = check_for_name_and_all_option_conflict(parser, cerr, /*allow_empty=*/true);
+    auto parse_code = check_for_name_and_all_option_conflict(parser, cerr, /*allow_empty=*/ !petenv_name.isEmpty());
     if (parse_code != ParseCode::Ok)
         return parse_code;
```

surahman:
> Hi @surahman, why would you need that "check to see if other instances are available"?

Sorry, my modifications were erroring out `start -all` and I thought there needed to be a check to see if there are any available instances when the `--all` flag is passed which can actually be launched. Thanks for the clarification!

> ```diff
> diff --git a/src/client/cli/cmd/start.cpp b/src/client/cli/cmd/start.cpp
> index 1c94ecb..7609f09 100644
> --- a/src/client/cli/cmd/start.cpp
> +++ b/src/client/cli/cmd/start.cpp
> @@ -173,7 +173,7 @@ mp::ParseCode cmd::Start::parse_args(mp::ArgParser* parser)
>      if (status != ParseCode::Ok)
>          return status;
>  
> -    auto parse_code = check_for_name_and_all_option_conflict(parser, cerr, /*allow_empty=*/true);
> +    auto parse_code = check_for_name_and_all_option_conflict(parser, cerr, /*allow_empty=*/ !petenv_name.isEmpty());
>      if (parse_code != ParseCode::Ok)
>          return parse_code;
> ```

Thanks, this tightens up the code.

surahman:
I have made adjustments, all tests are passing locally. I am not sure if the failed build pipeline is due to the branch rename and subsequent closing.

surahman:
@townsend2010 thanks for the info :)

ricab (changes requested):
Alright, I finally managed to circle back to this, sorry for the delay here.

This is getting closer, but a few aspects still need improving and one important piece is still missing: the GUI. I had to skip the tests, leaving that for a later pass.

Most of my requests are repeated in the multiple affected commands. Note that, when I mentioned updating the help messages, I meant "where they mention primary today", so mainly `addPositionalArgument` AFAICS, not the general description.

include/multipass/cli/command.h
```diff
@@ -135,6 +135,27 @@ class Command
         });
     }
 
+    uint32_t request_instance_count(const int& verbosity_level = 0)
```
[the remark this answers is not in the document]
surahman: Thanks for the feedback, I will look into these issues as time permits. I have not spent all that much time reviewing this code, just scattered work on it when I had time.

surahman:
I hope I got everything from this review cycle 🤞🏽.

surahman:
Found some time and reviewed your comments again, please let me know if anything got overlooked.

surahman:
If the CLI updates are complete we can look at the GUI. What is supposed to be happening there?

I would like to complete the two PR I have open before moving on to others.

surahman:
A quick attempt at wiring in the GUI. This is the first time I am running the build on any of my machines and I am running into issues where the GUI does not seem to be able to open the shell, installing from snap does not have this issue. Not sure if I have broken something in the build or if I am missing instructions (followed readme.md).

surahman:
A quick attempt at wiring in the GUI in some spare time. This is the first time I am running the GUI of a build on any of my machines and I am running into issues where the GUI does not seem to be able to open the shell. Installing from snap does not have this issue. Not sure if I have broken something in the build or if I am missing instructions (following `README.md`).

When disabling the primary instance we may need to stop it, which I believe would need to be wired into the `create_actions`
here:
https://github.com/surahman/multipass/blob/fb4e4559c7a14c0ba354451cabb769ac3d8b289c/src/client/gui/gui_cmd.cpp#L179

ricab (commented):
Nice nice, this is getting close :slightly_smiling_face:

The CLI part is practically done. The GUI that I am less sure of. I think for the time being it would be better to just disable (grey out) the 3 buttons concerning primary. But I will discuss this with others and give you some more feedback. BTW, the start button is missing when launching the GUI with a disabled primary.

The [help for the name option in launch](https://github.com/canonical/multipass/blob/bd340e8d9703fafcf3349bdf05fe1804b7ec7e88/src/client/cli/cmd/launch.cpp#L199) also needs updating (it shouldn't mention primary or auto-mounting when disabled).

src/client/gui/gui_cmd.cpp
```diff
@@ -59,7 +60,8 @@ void set_input_state_for(QList<QAction*> actions, const mp::InstanceStatus& stat
     {
         start,
         open_shell,
-        stop
+        stop,
+        toggle
```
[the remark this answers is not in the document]
ricab: Or maybe just an enable button (but no disable). We don't need to worry about maintaining the previous name if it was reset elsewhere.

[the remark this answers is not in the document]
ricab: Hey @surahman, we have a release to ship and folks are on holidays, so we'll need to defer discussion until later and we'd better put the GUI part on hold.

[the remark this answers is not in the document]
surahman: > we have a release to ship and folks are on holidays

@ricab all work and no play make life dull 😉.

I have begun resolving the `CLI` issues and started refactoring and updating the tests.

tests/test_cli_client.cpp
```diff
@@ -531,6 +531,33 @@ TEST_F(Client, shell_cmd_fails_unknown_options)
...
+
+    EXPECT_THAT(send_command({"shell", "foo"}), Eq(mp::ReturnCode::Ok));
+    EXPECT_THAT(send_command({"shell", "-h"}), Eq(mp::ReturnCode::Ok));
+    EXPECT_THAT(send_command({"shell", "--all"}), Eq(mp::ReturnCode::CommandLineError));
+}
+
+TEST_F(Client, shell_cmd_all_fail)
+{
+    EXPECT_THAT(send_command({"shell", "--all"}), Eq(mp::ReturnCode::CommandLineError));
+}
```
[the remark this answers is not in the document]
surahman: The purpose of this test is to ensure the changes I have made do not break the `--all` option. I have removed it.

ricab (commented):
Looking good, thanks @surahman! 

We'll get back to you with a path for the GUI soon.

surahman:
@ricab no worries.

I would ask that you folks consider the following scenario:
Primary Instance is running.
User disables Primary Instance.
No checks or shutdown signal is sent to the Primary Instance, it continues to run.
User re-enables Primary Instance.

In the best case scenario, the Primary Instance name is stored and the user can issue commands to it again. Worst case scenario, if you assign a new random name using the functions in `src/petname/`, you have an orphaned Primary Instance. This is why I wanted to check for an active Primary Instance on the daemon via an RPC or some other method.

ricab:
Hey @surahman, I think that's fine. It's similar to changing the name to something else. Suppose: 

1. a primary instance exists, with name "primary"
2. the user does `multipass set client.primary-name=another`

The instance named "primary" remains, but it becomes a regular instance that happens to be named "primary". A `multipass start` at that point would launch "another". Setting to "" just makes it so that `multipass start` would instead fail.

So, in both cases:

3. `multipass set client.primary-name=primary`

Just makes the instance named "primary" get the special treatment again.

Notice we use the term "primary" as both instance property and instance name. Perhaps a little confusing, but just realize an instance named "primary" doesn't need to have the primary property, and vice-versa.

ricab:
@surahman for the GUI, we discussed a bit and we think the best for now is to just remove the entire primary section: buttons and separator. They would need to be added back when re-enabled (reflecting the current state).

Additionally, the [hotkey](https://github.com/canonical/multipass/blob/main/src/client/gui/gui_cmd.cpp#L107) should not open a shell anymore. That may be easier to achieve by just checking if primary is disabled inside the action callback.

surahman:
@ricab thanks for all the info - I will get started on this, just waiting on the formatter pull to be complete. The delta between the code means I have to discard all the build artifacts and do a clean build, and that takes a while...

Saviq:
> @ricab thanks for all the info - I will get started on this, just waiting on the formatter pull to be complete. The delta between the code means I have to discard all the build artifacts and do a clean build, and that takes a while...

That shouldn't be necessary? Just building again should rebuild just the needed bits. I can also recommend `ccache`, and well - you could always have multiple build dirs :)

surahman:
@Saviq thank you, I will change my workflows. I currently have a single `build` directory and I `stash` my `gitignore`, and then `stash` my build artefacts, then switch branches, then pop my `stash` and go on my merry way.

This time I jumped from the `cloud-init-iso` branch to this old branch and got a whole load of undefined reference errors for `adapter-something-something` during linking. My guess is `make` compared the timestamps of last modified on the compilation units against their artefacts and decided they did not need updating.

ricab:
Hey @surahman, you should be able to add your custom ignore stuff to `.git/info/exclude` in your local repo, instead of stashing `.gitignore`. You can also have your own global `.gitignore`, if that makes sense for your other projects.

I also just leave the build tree around, only clearing it if there are changes to CMake that require it (my IDE manages that sort of thing actually). And yeah, `ccache` is a must. Anyway, take your time of course :slightly_smiling_face:

surahman:
> Hey @surahman, you should be able to add your custom ignore stuff to `.git/info/exclude` in your local repo, instead of stashing `.gitignore`. You can also have your own global `.gitignore`, if that makes sense for your other projects.

@ricab thank you! This makes life a lot easier 😄, and I shall get `ccache` installed as well.

> just remove the entire primary section: buttons and separator. They would need to be added back when re-enabled (reflecting the current state).

I wanted to clarify that you want me to remove the complete `primary` section, except a button to re-enable it, when the `disable primary` is toggled on, and returned when it is toggled off?

ricab:
> @ricab thank you! This makes life a lot easier smile, and I shall get ccache installed as well.

Sure thing! Glad to help :)

> I wanted to clarify that you want me to remove the complete primary section, except a button to re-enable it, when the disable primary is toggled on, and returned when it is toggled off?

The internal verdict was to remove entirely when primary is disabled, not even leaving a button to re-enable it. Beside entries for existing instances, only the "about" and "quit" would remain. When `client.primary-name` becomes non-empty again, the buttons would return.

surahman:
@ricab I need some guidance on the GUI modifications. The decorator pattern is familiar because I have done GUI work in Java but I am not proficient with Qt.

> Additionally, the [hotkey](https://github.com/canonical/multipass/blob/main/src/client/gui/gui_cmd.cpp#L107) should not open a shell anymore. That may be easier to achieve by just checking if primary is disabled inside the action callback.

I believe I have handled this with a safety check in the `open_multipass_shell` routine for an empty instance name.

> The internal verdict was to remove entirely when primary is disabled, not even leaving a button to re-enable it. Beside entries for existing instances, only the "about" and "quit" would remain. When `client.primary-name` becomes non-empty again, the buttons would return.

I believe I have partially achieved this by tightening up the `handle_petenv_instance`. The only location I have found where the `petenv` related menu separator items are added is in the `create_actions` routine, and it is here that I have added a check. The tray icon is missing the `petenv` section on a clean launch now and having any instance will crash the GUI - something is awry.

src/client/gui/gui_cmd.cpp
```diff
@@ -165,6 +165,12 @@ void cmd::GuiCmd::create_actions()
     QObject::connect(&petenv_start_action, &QAction::triggered, [this] {
         future_synchronizer.addFuture(QtConcurrent::run(this, &GuiCmd::start_instance_for, current_petenv_name));
     });
+    QObject::connect(&petenv_disable_action, &QAction::triggered,
+                     [this]
+                     {
+                         current_petenv_name = "";
+                         MP_SETTINGS.set(petenv_key, QString::fromStdString(current_petenv_name));
+                     });
```
ricab, line 173: I am not sure if this was leftover from earlier, but I don't think we need a button to disable primary.

src/client/gui/gui_cmd.cpp
```diff
@@ -235,6 +241,13 @@ void cmd::GuiCmd::update_menu()
     {
         about_separator->setVisible(true);
     }
+
+    const bool petenv_visibility = MP_SETTINGS.get(petenv_key).isEmpty() ? false : true;
```
ricab, line 245:
```suggestion
    const bool petenv_visibility = !MP_SETTINGS.get(petenv_key).isEmpty();
```

src/platform/client/client_platform_linux.cpp
```diff
@@ -24,8 +24,9 @@ namespace mcp = multipass::cli::platform;
 
 void mcp::open_multipass_shell(const QString& instance_name)
 {
-    QProcess::startDetached(
-        "xterm", {"-title", instance_name, "-j", "-e", QString("multipass shell %1 || read").arg(instance_name)});
+    if (!instance_name.isEmpty())
```
ricab, line 27: Could this check be moved to the call site, please? 

I believe it's the job of the GUI and not so much this util function to avoid opening a shell when primary is disabled (this function is better regarded as having a(n undocumented) pre-condition that the instance name is valid). 

This can have an assert instead, if you like.

ricab (commented):
Hey @surahman, this is looking quite alright, I think you've got the right approach and the GUI seems to be doing all we need :smiley:

Only a couple of details inline, but +1 on both checking disabled primary for shell opening and updating button visibility.

ricab:
@surahman, I failed to notice that open_shell was being given an empty instance name to open the primary shell. So my talk about preconditions and asserts mustn't have made much sense. I just pushed what I meant: passing the actual name to open shell (giving us a proper window title), documenting the precondition in the util, and asserting (part of it) in the implementation.

Among other details, I also switched to using the cached setting, to avoid reading from disk multiple times. It shouldn't go stale for more than a second and better consider the same thing throughout anyway. Let me know what you think.

bors try

surahman:
This line caught me out initially, until I looked up how we can display a message for an `assert`, nice 😉 :
```c++
assert(!instance_name.isEmpty() && "Instance name cannot be empty");
```

It seems like what you have done is solid, but this line seems to still be passing in a `null` string on line 111:
```c++
mp::cli::platform::open_multipass_shell(QString());
```

ricab:
> This line caught me out initially, until I looked up how we can display a message for an `assert`, nice wink :

Yeah, asserts are weird in C++, no slot for a message, so need that trick :sunglasses: 

> It seems like what you have done is solid, but this line seems to still be passing in a null string on line 111 in src/client/gui/gui_cmd.cpp:

It looks alright [here](https://github.com/canonical/multipass/pull/2102/files#diff-e058fbe77c4ffdb2c0015d1339a6b6f9ede02d61b8921b1d8b7f01f9652fa660R111), are you perhaps looking at an old commit?

surahman:
> It looks alright [here](https://github.com/canonical/multipass/pull/2102/files#diff-e058fbe77c4ffdb2c0015d1339a6b6f9ede02d61b8921b1d8b7f01f9652fa660R111), are you perhaps looking at an old commit?

Yup, sorry, was looking at one of the earlier commits - looks good 👍🏽

ricab:
Hmm bors?

bors try

ricab:
Oh there's a conflict that needs fixing. @surahman would you be able to fix that? Thanks!

ricab:
bors cancel

ricab:
bors try

Saviq:
bors cancel

Saviq:
bors try-

Saviq:
bors try

ricab:
bors cancel

ricab:
bors try-

ricab:
bors try

ricab:
@surahman I believe you have dropped some test code when fixing the conflicts. I didn't want to force-push into your repo (probably couldn't anyway) so I have pushed a [fixed merge](https://github.com/canonical/multipass/commit/4cb066e4ac1d9c18cd74f383ef600148d28c6bc8) in the branch [primary-null-proposed](https://github.com/surahman/multipass/compare/set_client.primary-name=null...canonical:primary-null-proposed).

You should be able to move this PR to it with the following steps:

1. `git checkout set_client.primary-name=null`
2. `git fetch <canonicals_multipass_remote>`
3. `git reset --hard <canonicals_multipass_remote>/primary-null-proposed`
4. `git push --force` (after confirming it's pointing at the new merge commit)

surahman:
@ricab sorry about that, thanks for the fix 👍🏽.

ricab:
Alright, let's try again.

bors try

ricab (approved):
Alright, let's merge this. Yeyyy :partying_face: Thanks again for your effort getting this done @surahman 

bors merge

surahman:
@ricab you are most welcome! Thanks for all your patience, guidance, and assistance with this.

ricab:
bors retry

ricab:
bors retry

PR #2105, opened 2021-05-13
[platform] Read Snap Linux OS Release

surahman:
**_Issue #1781:_**
Creating changes requested in #2096 on a new branch instead of doing a hard reset on the original branch to maintain the changes in the original files whilst I work.

Tests have been added to the `PlatformLinux` fixture and an empty value in a field is treated as `unknown`:
```
NAME=""
...
VERSION_ID=""
```
I have not wired into `/src/daemon/daemon.cpp` string pollination because the suggested format of `distro_name-distro_rel` does not match the original.


src/platform/platform_linux.cpp
```diff
@@ -308,3 +309,56 @@ std::string mp::platform::reinterpret_interface_id(const std::string& ux_id)
...
+                distro_rel = val;
+        }
+    }
+
+    return std::pair(distro_id, distro_rel);
+}
+
+std::string mp::platform::read_os_release(const QString& path)
+{
+    QStringList lsb_file_data;
```
ricab, line 339: Idem: name update.

src/platform/platform_linux.cpp
```diff
@@ -308,3 +309,56 @@ std::string mp::platform::reinterpret_interface_id(const std::string& ux_id)
...
+
+    QString distro_id = "unknown";
+    QString distro_rel = "unknown";
+
+    for (const QString& line : lsb_file_data)
+    {
+        QStringList split = line.split(delimiter, Qt::KeepEmptyParts);
+        if (split.length() == 2 && split[1].length() > 2)
+        {
+            QString val = split[1].mid(1, split[1].length() - 2);
```
ricab, line 326: Could you please add a short comment clarifying that this removes the quotes?

include/multipass/platform.h
```diff
@@ -86,6 +86,12 @@ std::function<int()> make_quit_watchdog(); // call while single-threaded; call r
 
 std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
 
+std::string read_os_release(const QString& path = "/var/lib/snapd/hostfs/etc/os-release");
```
ricab, line 89: Reading of `os-release` should fall back to `(/var/lib/snapd/hostfs)/usr/lib/os-release` when `(...)/etc/os-release` can't be read. See `man 5 os-release`.

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +555,172 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+    auto expected = std::pair<std::string, std::string>("Ubuntu", "21.04");
+
+    auto output = mp::platform::parse_os_release(input);
+
+    EXPECT_EQ(expected.first, output.first.toStdString());
+    EXPECT_EQ(expected.second, output.second.toStdString());
+}
+
+TEST_F(PlatformLinux, parse_os_release_ubuntu2104lts_delimiter)
```
ricab, line 662: Idem: I don't think we need to support handling arbitrary delimiters. The least functionality we commit to, the less we have to maintain :smile:

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +555,172 @@ INSTANTIATE_TEST_SUITE_P(
...
+                         {"VERSION_CODENAME=hirsute"},
+                         {"UBUNTU_CODENAME=hirsute"}};
+
+    auto expected = std::pair<std::string, std::string>("Ubuntu", "21.04");
+
+    auto output = mp::platform::parse_os_release(input);
+
+    EXPECT_EQ(expected.first, output.first.toStdString());
+    EXPECT_EQ(expected.second, output.second.toStdString());
+}
```
ricab, line 637: These last three tests all repeat the same structure and could be parameterized. The parameter would be the pair you now set in `expected`, which you could render into `input` with [`QString::arg`](https://doc.qt.io/qt-5/qstring.html#arg). You can look for `TEST_P` in our code to see examples of how that works.

I guess the tests for the empty file and different order are fine as separate tests.

ricab (commented):
Hey @surahman, thanks for this, it is now very close to what we need! I have left some comments and requests on smaller aspects that can be improved, but the overall approach and code look good to me :slightly_smiling_face:

One other thing I would ask is to please run `git clang-format` on your changes, so that they follow [our format](https://github.com/canonical/multipass/blob/main/.clang-format).

> I have not wired into /src/daemon/daemon.cpp string pollination because the suggested format of distro_name-distro_rel does not match the original.

Could you please elaborate on this? Doesn't your code produce the same as `QSysInfo`?

tests/linux/test_platform_linux.cpp
surahman: Make sense, updated.

src/platform/platform_linux.cpp
```diff
@@ -308,3 +309,56 @@ std::string mp::platform::reinterpret_interface_id(const std::string& ux_id)
 {
     return ux_id;
 }
+
+std::pair<QString, QString> mp::platform::parse_os_release(const QStringList& lsb_file_data, const char& delimiter)
```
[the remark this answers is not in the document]
surahman: Removed.

surahman:
Thanks @ricab I am effecting changes. 

> Could you please elaborate on this? Doesn't your code produce the same as `QSysInfo`?

https://github.com/canonical/multipass/blob/89f73fb8debf41a496d1ab058308f9764cc038dc/src/daemon/daemon.cpp#L126-L127
The output from `read_or_release` is a single `std::string` of the form:
```
{}-{}
distro_name-distro_version
```

ricab:
Hey @surahman, right, so you could do something like this to plug into the daemon:

```c++
pollinate_user_agent_string += fmt::format("multipass/host/{} # written by Multipass\n", 
                                           platform::read_os_release()); 
```

But let me know if there's something else I am missing

surahman:
> ```c++
> pollinate_user_agent_string += fmt::format("multipass/host/{} # written by Multipass\n", 
>                                            platform::host_version()); 
> ```

@ricab my bad :man_facepalming:

> These are only useful for the implementation of `host_version()` on Linux. They're implementation details that clients of `platform.h` don't care about and which don't apply in other platforms.
> 
> So, could you please remove these declarations and move the corresponding functions to the unnamed namespace in `platform_linux.cpp`? Thanks.

Removing these from the header file makes it difficult to expose them for testing. Is there perhaps another header I can put them into to expose them specifically for testing?

ricab:
> > 
> 
> Removing these from the header file makes it difficult to expose them for testing. Is there perhaps another header I can put them into to expose them specifically for testing?

Yep, good point, that would be [here](https://github.com/canonical/multipass/blob/cee3ae95a298c32b674b33337560fe5d57d26540/src/platform/platform_linux_detail.h).

surahman:
I have added the header declarations to `src/platform/platform_linux_detail.h` but if memory serves correctly it is bad to declare anonymous namespaces in headers. Is it acceptable to incase the routines in the `multipass::platform::detail` namespace? I have not created the parameterized test fixture just yet.

ricab:
Yes, definitely, `multipass::platform::detail` is the right place.

src/platform/platform_linux.cpp
```diff
@@ -154,6 +155,59 @@ void update_bridges(std::map<std::string, mp::NetworkInterfaceInfo>& networks)
...
+}
+
+std::string mp::platform::Platform::read_os_release(const QString& os_release_path)
+{
+    QFile fd(os_release_path);
+    bool exists_os_rel_path = QFileInfo::exists(os_release_path); // Check once, going to disk is expensive.
+    if (!exists_os_rel_path && QFileInfo::exists("/var/lib/snapd/hostfs/etc/os-release"))
+        fd.setFileName("/var/lib/snapd/hostfs/usr/lib/os-release");
+    else if (!exists_os_rel_path && QFileInfo::exists("/var/lib/snapd/hostfs/usr/lib/os-release"))
+        fd.setFileName("/var/lib/snapd/hostfs/etc/os-release");
```
ricab, line 189: I have a couple of concerns about this:
1. The file could exist and still be unreadable. This is not likely, nor is it the situation mentioned in `man 5 os-release`, but we might as well fall back in that case as well. That would amount to going directly to `open` and check whether that succeeds, instead of checking if the file exists.
1. I don't think this function should receive the file to read from and then silently use something else. I understand you want the param for testing purposes, but then this probably needs another layer. How about this:

```cpp
std::unique_ptr<QFile> find_os_release()
{
    const std::array<QString, 3> options{QStringLiteral("/var/lib/snapd/hostfs/etc/os-release"), 
        QStringLiteral("/var/lib/snapd/hostfs/usr/lib/os-release"), 
        QStringLiteral("")};

    auto it = options.cbegin();
    auto ret = std::make_unique<QFile>(*it);
    while (++it != options.cend() && !MP_FILEOPS::open(*ret, QIODevice::ReadOnly | QIODevice::Text))
        ret->setFileName(*it);

    return ret; // need the unique_ptr because QFile is non-copyable
}
```

`MP_FILEOPS::open` replaces `QFile::open` to allow mocking. Then in `read_os_release` you'd receive `const QFile& os_release` and, if open, do you thing, otherwise return the default.

ricab (commented):
Hi @surahman, here are a few more comments.

There seem to be some formatting changes that are unrelated to the PR. Could you please drop them? The idea of using clang-format was to format only your changes (sorry if I wasn't more specific before). So, you need `git-clang-format` installed and then, each time before you commit, after you've staged your changes, you run `git clang-format`. That formats only your diff and leaves the rest alone.

src/platform/platform_linux.cpp
```diff
@@ -154,6 +155,59 @@ void update_bridges(std::map<std::string, mp::NetworkInterfaceInfo>& networks)
...
+    {
+        QStringList split = line.split('=', Qt::KeepEmptyParts);
+        if (split.length() == 2 && split[1].length() > 2) // Check for at least 1 char between quotes.
+        {
+            QString val = split[1].mid(1, split[1].length() - 2); // Removing quotes from input string.
+            if (split[0] == id_field)
+                distro_id = val;
+            else if (split[0] == version_field)
+                distro_rel = val;
+        }
```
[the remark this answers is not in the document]
surahman: Good review points, I have made the updates. I would have liked to use `std::move` instead of the pointers but that will not avoid unneeded string splits.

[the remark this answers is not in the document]
surahman: I have created a lambda to strip quotes, the name is explanatory enough to remove the comment. Copy elision optimization should take care of unneeded copying. Tests passing.

src/platform/platform_linux.cpp
surahman: Removed `unique_ptr` and switched to `QFile` on stackframe. `unique_ptr` usage necessitates a `dynamic_cast` or release of the `unique_ptr` to create the `QTextStream`. Figuring out tests.

src/platform/platform_linux.cpp
```diff
@@ -154,6 +155,69 @@ void update_bridges(std::map<std::string, mp::NetworkInterfaceInfo>& networks)
...
+
+    for (const QString& line : os_data)
+    {
+        QStringList split = line.split('=', Qt::KeepEmptyParts);
+        if (split.length() == 2 && split[1].length() > 2) // Check for at least 1 char between quotes.
+        {
+            if (split[0] == id_field)
+                distro_id = strip_quotes(split[1]);
+            else if (split[0] == version_field)
+                distro_rel = strip_quotes(split[1]);
```
ricab, line 190: :+1: the little DRY infringement is worth the added readability in this case.

src/platform/platform_linux.cpp
```diff
@@ -154,6 +155,69 @@ void update_bridges(std::map<std::string, mp::NetworkInterfaceInfo>& networks)
...
+        }
+    }
+
+    return std::pair(distro_id, distro_rel);
+}
+
+std::string multipass::platform::detail::read_os_release()
+{
+    QFile& os_release = *find_os_release().release();
```
ricab, line 199: Oh, I don't think you can release, it would leak :/ I guess you need `auto os_release = find_os_release();` and then deref with `*` or `->` as needed. Or you can add a reference var if you want, but keeping the `unique_ptr` around and owning the `QFile` so that it gets destroyed when it goes out of scope.

src/platform/platform_linux.cpp
ricab: Yeah, have a look here: https://github.com/canonical/multipass/blob/1e88479def51630cab9937804f47eee25f820cb0/tests/test_sftpserver.cpp#L1087-L1089

ricab (commented):
Just had a quick look. Feel free to ping me if there is anything I can help with or when you think it's ready for another look.

src/platform/platform_linux.cpp
surahman: Memory leak 🤦🏽‍♂️, corrected.

surahman:
@ricab thanks for all the help and feedback, I think the core logic is complete.

I have never done file system mocking and could use some advice and guidance on this. I have read through the [GTest cookbook](https://google.github.io/googletest/gmock_cook_book.html) on mocking.

Beginning with the `find_os_release` routine, my test, as it stands, simply checks to ensure that the `QFile.open` method is called in the loop in the routine.

In `read_os_release` the test checks for `QFile.open` are called but the returned filehandle will always be "bad" in and code within the `if` statement will never be executed, and thus never covered.

ricab (commented):
Shaping up nicely :smiley: 

I added a couple of hints for the tests, but if you have trouble leave it and I can pick it up and add the necessary tweaks, when I have the time.

The CLA check is still failing in CI. I didn't follow closely, but saw you needed to do something in the other PRs for it to pass. I suppose this one would need the same.

Saviq:
> The CLA check is still failing in CI. I didn't follow closely, but saw you needed to do something in the other PRs for it to pass. I suppose this one would need the same.

It's Lint that's failing, not CLA :)

ricab:
Oh right, I saw that wrong.

surahman:
I am trying to not drive up the infrastructure costs - those builds take a while. I made the mistake of not creating a parallel development branch and I am using the repo to move files between machines ;)

surahman:
CLA check seems to be failing now 😞

townsend2010:
Hey @surahman, it was a transient Ubuntu archive issue.  I re-ran it and it worked this time :smiley:

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +557,169 @@ INSTANTIATE_TEST_SUITE_P(
...
+TEST_F(PlatformLinux, find_os_release_dummy_file)
+{
+    const QString test_file_path = mpt::test_data_path() + "os-release_sample";
+    auto os_release_dummy = std::make_unique<QFile>(test_file_path);
+// FAIL LINT, STOP BUILD.
+    //    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();
+    //    EXPECT_CALL(*mock_file_ops, open(os_release_dummy, _)).Times(1);
+
+    auto output = multipass::platform::detail::find_os_release();
+    EXPECT_EQ(output->fileName(), test_file_path);
```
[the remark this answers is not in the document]
surahman: Running into some issues with the second file location test:
```c++
TEST_F(PlatformLinux, find_os_release_usr_lib)
{
    const auto expected_filename = QStringLiteral("/var/lib/snapd/hostfs/usr/lib/os-release");

    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();

    InSequence seq;
    EXPECT_CALL(*mock_file_ops,
                open(Property(&QFile::fileName, Eq(expected_filename)), QIODevice::ReadOnly | QIODevice::Text))
        .Times(2)
        .WillOnce(Return(false))
        .WillOnce(Return(true));
    EXPECT_CALL(*mock_file_ops, open).Times(0); // no other open attempts

    auto output = multipass::platform::detail::find_os_release();
    EXPECT_EQ(output->fileName(), expected_filename);
}
```
The `find_os_release` logic seems to be solid, but this setup seems to be failing. Tried resetting the test saturation and removing the `Times` expectation - no joy.

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +557,169 @@ INSTANTIATE_TEST_SUITE_P(
...
+}
+
+TEST_F(PlatformLinux, read_os_release_from_file)
+{
+    const std::string expected = "distribution_name-distribution_release";
+    const QString test_file_path = mpt::test_data_path() + "os-release_sample";
+    auto os_release_dummy = std::make_unique<QFile>(test_file_path);
+
+    //    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();
+    //    EXPECT_CALL(*mock_file_ops, open(os_release_dummy, _)).Times(2);
```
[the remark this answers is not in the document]
surahman: I have not tested the following but it seems like it should work:
```c++
std::string multipass::platform::detail::read_os_release()
{
    std::unique_ptr<QFile> os_release = find_os_release();

    QStringList os_info;
    if (os_release->isOpen())
    {
        const int buffer_size = 32;
        char buffer[buffer_size];
        QTextStream input;
        while (MP_FILEOPS.read(*os_release, buffer, sizeof(char) * buffer_size) != -1)
        {
            input << buffer;
        }

        QString line = input.readLine();
        while (!line.isNull())
        {
            os_info.append(line);
            line = input.readLine();
        }
        os_release->close();

        auto parsed = parse_os_release(os_info);

        return fmt::format("{}-{}", parsed.first, parsed.second);
    }

    return "unknown-unknown";
}
```

I have added the initial setup for this test in the suite. The above code has not been added. I feel like this routine is simple enough to be verified in code review.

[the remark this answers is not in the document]
ricab: Hey @surahman, don't worry, this is very close to what we need now and I can pick it up and add commits on top to sort out the remaining tests (authorship of your commits isn't affected). I am bound to fail anticipating new difficulties unless I get my hands on it. But tests are important to check future code as much as today's code. And this is so close, it would be a pity to drop :slightly_smiling_face: 

These google test/mock specifics may be beyond your interests but, if you want to know, I can think of two possible approaches:
1. Mock all the way, including `isOpen` and `readLine` (the latter to avoid the intermediate text stream in the implementation)
2. Go back to using a temporary file after all, and use [`SetArgReferee`](https://github.com/google/googletest/blob/master/docs/reference/actions.md#side-effects) in `open` to switch the arg to your earlier sample file, and combine with the return via [DoAll](https://github.com/google/googletest/blob/master/docs/reference/actions.md#composite-actions)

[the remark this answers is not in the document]
surahman: I agree with you on testing 100% of code, amongst other things it protects against unexpected behavior from changes in libraries and other dependant code. Mock testing has been the more interesting part of this exercise, but if this is a feature that needs to go out then please take over. If time is not an issue I will figure it out 😉.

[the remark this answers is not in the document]
surahman: @ricab Found some idle time and very quickly built up the mock hooks for `isOpen` and `readLine`. `isOpen` is working as expected but `readLine` only recovers a single line of input. I am unsure if the `readLine` issue is related to the way I have written the mock routine or the way the test fixture is set up. Thoughts?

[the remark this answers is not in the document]
ricab: Of course @surahman, please have at it! :smiley:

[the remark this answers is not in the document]
surahman: @ricab something appears to be broken in this test, it does not seem to want to try and open the file:
```c++
TEST_F(PlatformLinux, find_os_release_usr_lib)
{
    const auto expected_filename = QStringLiteral("/var/lib/snapd/hostfs/usr/lib/os-release");

    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();

    InSequence seq;
    EXPECT_CALL(*mock_file_ops,
                open(Property(&QFile::fileName, Eq(expected_filename)), QIODevice::ReadOnly | QIODevice::Text))
        .WillOnce(Return(false))
        .WillOnce(Return(true));
    EXPECT_CALL(*mock_file_ops, open).Times(0); // no other open attempts

    auto output = multipass::platform::detail::find_os_release();
    EXPECT_EQ(output->fileName(), expected_filename);
}
```

```
tests/linux/test_platform_linux.cpp:674: tried expectation #0: EXPECT_CALL(*mock_file_ops, open(Property(&QFile::fileName, Eq(expected_filename)), QIODevice::ReadOnly | QIODevice::Text))...(of type QString)
         Expected: to be called once
/home/saad/Multipass/tests/linux/test_platform_linux.cpp:678: tried expectation #1: EXPECT_CALL(*mock_file_ops, open)...
         Expected: all pre-requisites are satisfied
           Actual: the following immediate pre-requisites are not satisfied:
linux/test_platform_linux.cpp:674: pre-requisite #0
                   (end of pre-requisites)
         Expected: to be never called
           Actual: never called - saturated and active
```
Any ideas?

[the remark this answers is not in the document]
surahman: All builds and tests passing locally. I have pushed for review.

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +557,185 @@ INSTANTIATE_TEST_SUITE_P(
...
+    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();
+
+    InSequence seq;
+    EXPECT_CALL(*mock_file_ops, open).WillOnce(Return(true));
+    EXPECT_CALL(*mock_file_ops, open).Times(0); // no other open attempts
+    EXPECT_CALL(*mock_file_ops, isOpen).Times(1).WillOnce(Return(true));
+    for (size_t idx = 0; idx < os_rel_len; ++idx)
+    {
+        EXPECT_CALL(*mock_file_ops, readLine).WillOnce(Return(os_rel_file.readLine())).RetiresOnSaturation();
+    }
```
ricab, line 725: If you had `output` as a `QStringList` (as per suggestion above), you could use a range-for:

```
for (const auto& line : output)
  EXPECT_CALL(...).WillOnce(Return(line)).RetiresOnSaturation();
```

Or you can have a single expectation using an iterator and a lambda (since `ReturnRoundRobin` is not available in the gmock version we use). Something of the sort:

```
auto it = output.begin();
EXPECT_CALL(...).WillRepeatedly([&it, end = output.end()] {
    return it == end ? "" : *it++;
});

**Review by ricab — COMMENTED** [review]
Nice, nice, you got a hang of those tests :wink: 

This is nearly there, only a few minor adjustments inline.

File: tests/linux/test_platform_linux.cpp (hunk 13 above)
Comment (surahman) [reply tests/linux/test_platform_linux.cpp:723]: > f you had `output` as a `QStringList` (as per suggestion above), you could use a range-for:

This makes things a lot cleaner and simpler, good suggestion - thanks!

Comment (surahman) [reply tests/linux/test_platform_linux.cpp]: A bit of a conundrum as to why `read_line` seems to be called twice (over-saturated)
```c++
TEST_F(PlatformLinux, read_os_release_from_file)
{
    const std::string expected = "A-B";
    const QStringList& input = parse_os_release_single_char_fields.first;
    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();

    InSequence seq;
    EXPECT_CALL(*mock_file_ops, open).WillOnce(Return(true));
    EXPECT_CALL(*mock_file_ops, is_open).WillOnce(Return(true));
    for (const QString& line : input)
    {
        EXPECT_CALL(*mock_file_ops, read_line).WillOnce(Return(line));
    }

    auto output = multipass::platform::detail::read_os_release();

    EXPECT_EQ(expected, output);
}
```
Setting it to retire on saturation does not fix the issue, and I do not feel it should be required seeing as how the `input` variable is updated at the loop and not in the mock call.

ricab: Right, the code does `while (!line.isNull())`, so we need a null string at the end... I am trying something, I'll get back to you.

tests/linux/test_platform_linux.cpp
```diff
@@ -554,4 +557,183 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();
+
+    InSequence seq;
+    EXPECT_CALL(*mock_file_ops, open).WillOnce(Return(true));
+    EXPECT_CALL(*mock_file_ops, is_open).WillOnce(Return(true));
+    for (size_t idx = 0; idx < os_rel_len; ++idx)
+    {
+        EXPECT_CALL(*mock_file_ops, read_line).WillOnce(Return(os_rel_file.readLine())).RetiresOnSaturation();
+    }
```
ricab, line 723: OK, so this worked for me:
```suggestion
    const auto& [input, expected_pair] = parse_os_release_ubuntu2104lts;
    auto expected = fmt::format("{}-{}", expected_pair.first, expected_pair.second);

    auto [mock_file_ops, guard] = mpt::MockFileOps::inject();

    InSequence seq;
    EXPECT_CALL(*mock_file_ops, open).WillOnce(Return(true));
    EXPECT_CALL(*mock_file_ops, is_open).WillOnce(Return(true));

    for (const auto& line : input)
        EXPECT_CALL(*mock_file_ops, read_line).WillOnce(Return(line)).RetiresOnSaturation();
    EXPECT_CALL(*mock_file_ops, read_line).WillOnce(Return(QString{}));
```

surahman: Nice.

ricab (commented):
Hmm, I was about ready to merge, but I am afraid something's not right:

```
ubuntu@primary:~$ sudo mount /dev/cdrom /media
mount: /media: WARNING: device write-protected, mounted read-only.
ubuntu@primary:~$ grep "written by Multipass" /media/vendor-data
    content: "multipass/version/1.7.0-dev.890.pr2105+gfe88c157 # written by Multipass\nmultipass/driver/qemu-4.2.1 # written by Multipass\nmultipass/host/unknown-unknown # written by Multipass\n"
```

@surahman you can download your snap from the "Artifacts" drop-down in the [checks tab](https://github.com/canonical/multipass/pull/2105/checks), then unzip and install it with `snap install --dangerous <snap_filename>`, to see how it's working.

ricab:
By the way...

>  I am afraid something's not right

I meant the `unknown-unknown` part.

ricab:
Also this ends up in `/etc/pollinate/add-user-agent` inside the instance (thanks @Saviq):

```
ubuntu@primary:~$ cat /etc/pollinate/add-user-agent
multipass/version/1.7.0-dev.890.pr2105+gfe88c157 # written by Multipass
multipass/driver/qemu-4.2.1 # written by Multipass
multipass/host/unknown-unknown # written by Multipass
```

surahman:
> Also this ends up in `/etc/pollinate/add-user-agent` inside the instance (thanks @Saviq):

Thats strange, I am fairly certain I am not writting to anything, it is supposed to simply reads a file and returns what is reads... I will look into it.

ricab:
@surahman nevermind, it's on us - I forgot we need need another snap interface before this can work.

surahman:
So we are all good? I got a little worried there and broke out GDB to inspect the binary...

ricab:
Yeah, I did the same and meanwhile remembered :wink:

surahman:
Phew 😓... No worries, Multipass is a pretty vast and complex vertical.

ricab:
So this needs #2132. @surahman would you be able to rebase this on that branch, please? That would be good to bring this up to date with recent changes too. 

I'll put the remaining bits in place for the other platforms.

surahman:
> So this needs #2132.

Done. Local build and tests all passing, waiting for the Github CI build artifacts for testing.

ricab:
Thanks, I updated the base to match. It should get updated back automatically once add-system-observe gets merged.

ricab (approved):
Alright, I have lined up the remaining bits for this. Hopefully all goes well, but they'll also need to be reviewed. I'm going away for a few days, but hopefully we can merge this by next week.

Thanks again for this!

surahman:
No worries, stay safe.

ricab:
Ah, you'll need to `snap connect multipass:system-observe` to see things working with the snap :wave:

ricab:
bors try

ricab:
bors try

ricab:
bors retry

ricab:
And again...

bors retry

townsend2010:
Let's see if anything is better now:

bors retry

Saviq:
Maybe it just needed a good night's sleep...

bors retry

Saviq:
OK that last one's an expected failure, fixed by @ricab on the macOS / Windows side.

ricab:
Indeed, this is good now and the snap works fine. 

bors r+

ricab:
No need for bors anymore.

bors cancel

PR #2112, opened 2021-05-18
[3rd-party] Bump scope_guard to v1.0.0

ricab:
Needed to avoid warning in newer clangs:

```
`ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Werror,-Wdtor-name]
sg::detail::scope_guard<Callback>::~scope_guard() noexcept´
```


Saviq (approved):
Yup.

bors merge

Saviq:
bors retry

PR #2114, opened 2021-05-18
[utils] Pass timeout when getting ssh_hostname() in wait_until_ssh_up()

### LINKED ISSUE #2113: [daemon] `--timeout` not respected in IP determination
We're seeing both `lxd` and `qemu` backends failing to respect the `--timeout` passed:

```plain
Tue, 18 May 2021 13:28:25 GMT multipass start --timeout 1200
...
Tue, 18 May 2021 13:34:58 GMT launch failed: The following errors occurred:
Tue, 18 May 2021 13:35:08 GMT failed to determine IP address
```

townsend2010:
Fixes #2113

### DIFF
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -242,8 +242,7 @@ bool mp::utils::valid_mac_address(const std::string& mac)
 void mp::utils::wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
                                   std::function<void()> const& ensure_vm_is_running)
 {
-    mpl::log(mpl::Level::debug, virtual_machine->vm_name,
-             fmt::format("Trying SSH on {}:{}", virtual_machine->ssh_hostname(), virtual_machine->ssh_port()));
+    mpl::log(mpl::Level::debug, virtual_machine->vm_name, "Waiting for SSH to be up");
     auto action = [virtual_machine, &ensure_vm_is_running] {
         ensure_vm_is_running();
         try
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -46,6 +46,8 @@
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
+using namespace std::chrono_literals;
+
 namespace
 {
 constexpr auto category = "utils";
@@ -247,7 +249,7 @@ void mp::utils::wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::
         ensure_vm_is_running();
         try
         {
-            mp::SSHSession session{virtual_machine->ssh_hostname(), virtual_machine->ssh_port()};
+            mp::SSHSession session{virtual_machine->ssh_hostname(1ms), virtual_machine->ssh_port()};
 
             std::lock_guard<decltype(virtual_machine->state_mutex)> lock{virtual_machine->state_mutex};
             virtual_machine->state = VirtualMachine::State::running;


src/utils/utils.cpp
```diff
@@ -243,7 +243,7 @@ void mp::utils::wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::
                                   std::function<void()> const& ensure_vm_is_running)
 {
     mpl::log(mpl::Level::debug, virtual_machine->vm_name,
-             fmt::format("Trying SSH on {}:{}", virtual_machine->ssh_hostname(), virtual_machine->ssh_port()));
+             fmt::format("Trying SSH on {}:{}", virtual_machine->ssh_hostname(timeout), virtual_machine->ssh_port()));
```
Saviq, line 246: Shouldn't we just _not_ call `::ssh_hostname()` here? Like… we have the `try_action_for()` below that's meant to handle timeouts. Worst case we'll wait 2*timeout here.

townsend2010: What would we call instead?  This for *only* getting the hostname/IP address and not for seeing when SSH is up and running.

townsend2010: Or ditch the log message or at least remove the IP/port info?

townsend2010: Also, if not here, then the timeout will have to be added to
https://github.com/canonical/multipass/blob/d8d8b2149d8b72a1d9509929e0527d21ac7eddd8/src/utils/utils.cpp#L251
instead.

So, let me know how you prefer this to be handled.

Saviq: > Or ditch the log message or at least remove the IP/port info?

Yeah, we shouldn't block in this just to log it. Can log after we get the info in the action.

> Also, if not here, then the timeout will have to be added to

Will it? Isn't `try_action_for` dealing with that exactly?

townsend2010: Actually, having discovered the LXD instance creation timeout and fixing that by moving it into `ensure_vm_is_running()`, I think this message should not include the hostname/IP address & port info.

townsend2010: Ok, I modified this message to not get the hostname/port and just log a generic message.  I also modified the subsequent `ssh_hostname()` call to have a 1ms timeout so it will use this action instead of the one built into `ssh_hostname()`.

src/utils/utils.cpp
```diff
@@ -242,13 +244,12 @@ bool mp::utils::valid_mac_address(const std::string& mac)
...
 {
-    mpl::log(mpl::Level::debug, virtual_machine->vm_name,
-             fmt::format("Trying SSH on {}:{}", virtual_machine->ssh_hostname(), virtual_machine->ssh_port()));
+    mpl::log(mpl::Level::debug, virtual_machine->vm_name, "Waiting for SSH to be up");
     auto action = [virtual_machine, &ensure_vm_is_running] {
         ensure_vm_is_running();
         try
         {
-            mp::SSHSession session{virtual_machine->ssh_hostname(), virtual_machine->ssh_port()};
+            mp::SSHSession session{virtual_machine->ssh_hostname(1ms), virtual_machine->ssh_port()};
```
Saviq, line 252: Should we add a debug line after this saying something like "Found instance hostname: … port: …"?

Saviq (commented):
In all the other calls to `ssh_hostname()` we're OK not having a specific timeout? If not, worth filing an issue?

townsend2010:
> In all the other calls to ssh_hostname() we're OK not having a specific timeout?

This is the first time `ssh_hostname()` is called, so either we'll get an IP address here and all other `ssh_hostname()` calls will return quickly or they will have the default 2 minute timeout.  So the question is if we want all the others to timeout quickly or not.  I have not looked into those to see what they are doing.

src/utils/utils.cpp
townsend2010: I don't think it's necessary since this is just testing for an ssh connection, not getting the IP address.  If we want to log the IP address discovery, that should probably go in `ip_address_for()`, but that seems to be unrelated to this PR.

Saviq (approved):
> So the question is if we want all the others to timeout quickly or not.

IIRC we cache it, so as long as it succeeds once, the remaining calls will be immediate.

bors merge

Saviq:
Let the ball roll…

bors retry

Saviq:
bors retry

townsend2010:
bors retry

PR #2115, opened 2021-05-19
[url_downloader] Add timeout when retrieving HTTP headers

### LINKED ISSUE #2104: [network] No timeout set when retrieving LastModified header
If there are network connectivity issues and Multipass tries to get the `LastModified`, it will never timeout since we don't set a timeout for that operation and `multipassd` can hang on any other network operation.

townsend2010:
Also, some refactoring of getting HTTP headers and waiting for the reply.

Fixes #2104


luis4a0 (approved):
Yeah, it looks great!

luis4a0:
bors retry

PR #2118, opened 2021-05-20
Improve Timer test reliability part deux

townsend2010:
Adds a getter function for current Timer state for better testing Timer state expectations.
Relax timeouts in some tests to improve reliability.

### DIFF
--- a/include/multipass/timer.h
+++ b/include/multipass/timer.h
@@ -53,7 +53,7 @@ private:
     void main();
     const std::chrono::seconds timeout;
     const std::function<void()> callback;
-    TimerState state;
+    TimerState current_state;
     std::thread t;
     std::condition_variable cv;
     std::mutex cv_m;
--- a/include/multipass/timer.h
+++ b/include/multipass/timer.h
@@ -52,7 +52,7 @@ private:
     void main();
     const std::chrono::seconds timeout;
     const std::function<void()> callback;
-    multipass::utils::TimerState state;
+    TimerState state;
     std::thread t;
     std::condition_variable cv;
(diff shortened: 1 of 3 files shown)


src/utils/timer.cpp
```diff
@@ -94,4 +93,9 @@ void Timer::stop()
         t.join();
 }
 
-} // namespace multipass::utils
+mpu::TimerState mpu::Timer::current_state()
```
Saviq, line 96:
```suggestion
mpu::TimerState mpu::Timer::state()
```

include/multipass/timer.h
```diff
@@ -47,12 +47,13 @@ class Timer
     void pause();
     void resume();
     void stop();
+    TimerState current_state();
```
Saviq, line 50:
```suggestion
    TimerState state();
```

Saviq (commented):
I'd rename `current_state` to `state` on the public side of the interface and rename the private member to something else.

Otherwise, seems legit!

src/utils/timer.cpp
townsend2010: I committed in my branch to avoid the cla check failure.

include/multipass/timer.h
townsend2010: I committed in my branch to avoid the cla check failure.

townsend2010:
bors r=saviq,ricab

ricab (approved):
+0.5 (partial review: of the actual timer code, tests skipped).

townsend2010:
Oh, good grief.  Try again:

bors retry

PR #2119, opened 2021-05-20
Bash completions for the aliases features

luis4a0:
This PR adds completion for the Multipass commands `alias`, `unalias` and `aliases`.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -48,6 +48,28 @@ _multipass_complete()
         opts=$( \eval $cmd | \grep -Ev '(\+--|Name)' | \cut -d',' -f 1 )
     }
 
+    _multipass_instances_with_colon()
+    {
+        local opts_bak="${opts}"
+        opts=""
+
+        _multipass_instances
+        local instances="${opts}"
+        opts="$opts_bak"
+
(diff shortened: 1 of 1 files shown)


townsend2010 (changes requested):
So, if I'm in `multipass alias` and I'm tab completing the instance name and choose an instance name and hit tab again, it completes the instance name twice.  For example, if I have an instance named `test`:
`$ multipass alias te`
Hit tab twice and then I get:
`$ multipass alias test:test:`

luis4a0:
Hey @townsend2010! Thanks for the report. It's fixed now. I also added a check to allow only one instance name in the alias 
definition.

townsend2010 (changes requested):
Ok, the `alias` and `aliases` completions like fine. However, I think `unalias` should bash complete the alias names as well.

luis4a0:
Oh, I must have forgotten `unalias`!

Done, it's implemented now.

townsend2010 (approved):
Ok, LGTM.  Thanks!

bors merge

townsend2010:
Hmm, some weird build error.  Try again...

bors retry

townsend2010:
Urg, again...

bors retry

PR #2122, opened 2021-05-24
[daemon] Make sure there is a path before checking if it exists

### LINKED ISSUE #2121: [regression] LXD instances are deleted on daemon startup
Due to 725713a81, all LXD instances are deleted on daemon start up.  This is due to LXD instances not having a path and as such, the `QFile::exists()` will always return `false` and the daemon will delete it.

townsend2010:
Fixes #2121

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -901,7 +901,7 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         }
 
         auto vm_image = fetch_image_for(name, config->factory->fetch_type(), *config->vault);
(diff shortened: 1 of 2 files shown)


ricab (commented):
Yep, thanks for the fix. Tests need adapting though...

townsend2010:
Right, working on those now :smiley:

ricab (approved):
LGTM, thanks.

bors r+

townsend2010:
Ugh, seems there is some fragility in the qemu integration test around discovery of snap being installed in the instance.

bors retry

townsend2010:
Ugh, again, I had to cancel the Mac build due a test hanging.

bors retry

Issue #2045, opened 2021-04-09
[cli] missing input validation for `--cpus`


Saviq:
**Describe the bug**
We can `multipass launch --cpus many`. We shouldn't be able to.

**To Reproduce**
How, and what happened?
1. `multipass launch --cpus many`
1. `Launched: …`

**Expected behavior**
It should error out.

surahman:
I have added checks for alpha, alphanumeric, decimal, and zero-valued inputs.

PR #2125, opened 2021-05-29
[cli] Validate CPU Parameter in Launch

surahman:
**_Issue #2045:_**

I have added validation for alpha, alphanumeric, and natural number valued inputs for this option in the `launch` command.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -269,15 +269,15 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     if (parser->isSet(cpusOption))
     {
         bool conversion_pass;
-        const int cpu_count = parser->value(cpusOption).toInt(&conversion_pass); // Base 10 conversion with check.
+        const int cpu_text = parser->value(cpusOption).toInt(&conversion_pass); // Base 10 conversion with check.
 
-        if (!conversion_pass || cpu_count < 1)
+        if (!conversion_pass || cpu_text < 1)
         {
-            cerr << "error: Please enter a non-zero integer value for the CPU count.\n ";
+            fmt::print(cerr, "error: Invalid CPU count '{}', need a positive integer value.\n", cpu_text);
             return ParseCode::CommandLineError;
         }
 
-        request.set_num_cores(cpu_count);
+        request.set_num_cores(cpu_text);
     }
 
     if (parser->isSet(memOption))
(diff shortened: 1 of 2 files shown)


src/client/cli/cmd/launch.cpp
```diff
@@ -268,7 +268,16 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
 
     if (parser->isSet(cpusOption))
     {
-        request.set_num_cores(parser->value(cpusOption).toInt());
+        bool conversion_pass;
+        const int cpu_count = parser->value(cpusOption).toInt(&conversion_pass); // Base 10 conversion with check.
+
+        if (!conversion_pass || cpu_count < 1)
+        {
+            cerr << "error: Please enter a non-zero integer value for the CPU count.\n ";
```
ricab, line 276: Little tweak to match other error msgs and remove the space at the end:

```suggestion
        const auto& cpu_text = parser->value(cpusOption);
        const int cpu_count = cpu_text.toInt(&conversion_pass); // Base 10 conversion with check.

        if (!conversion_pass || cpu_count < 1)
        {
            fmt::print(cerr, "error: Invalid CPU count '{}', need a positive integer value.\n", cpu_text);
```

ricab (commented):
Nice, thanks! Just a minor tweak inline, but otherwise looks good :slightly_smiling_face:

src/client/cli/cmd/launch.cpp
ricab: Hey @surahman, notice that `cpu_text` was meant to refer to the text before converting to int, which is what needs to go in the error message (can't use the int from a failed conversion).

ricab: BTW, you should be able to "Commit suggestion" through GH directly.

surahman: > BTW, you should be able to "Commit suggestion" through GH directly.

Sorry, I did not know that - will keep it in mind for next time. Diff is out of date now 😞 .

> Hey @surahman, notice that `cpu_text` was meant to refer to the text before converting to in.

🤦🏽 corrected.

ricab (approved):
Alright, let's get this in. Thanks!

bors r+

src/client/cli/cmd/launch.cpp
ricab: No worries!

ricab:
bors retry

PR #2127, opened 2021-06-01
[timer] Make current_state enum atomic

townsend2010:
Make the current_state enum atomic and remove some of the locks.  This also makes the state() getter deterministic.


townsend2010:
Going to go a different route.

PR #2129, opened 2021-06-01
Singleton consts

ricab:
Add missing `const`s to singleton methods. I left a couple of conceptual exceptions untouched, like `Utils::exit` or `Settings::set`. The latter would allow caching without changing the interface.

### DIFF
--- a/include/multipass/file_ops.h
+++ b/include/multipass/file_ops.h
@@ -36,21 +36,21 @@ public:
     FileOps(const Singleton<FileOps>::PrivatePass&) noexcept;
 
     // QDir operations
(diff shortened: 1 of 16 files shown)


townsend2010 (approved):
This looks good, but conflicts need fixing.  Once that's done, we can get it in.

ricab:
Thank you @townsend2010, I'll leave that for next week.

ricab:
Done @townsend2010 (here and on the other side). Thanks.

townsend2010 (approved):
LGTM, thanks!

bors merge

PR #2130, opened 2021-06-01
[LXD] Handle storage pools better

townsend2010:
This will use the default storage pool if it exists. Otherwise, it will use a multipass storage pool and create it if it doesn't exist.

### DIFF
--- a/src/platform/backends/lxd/lxd_virtual_machine.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine.cpp
@@ -115,11 +115,12 @@ QJsonObject generate_base_vm_config(const multipass::VirtualMachineDescription& 
     return config;
 }
 
-QJsonObject generate_devices_config(const multipass::VirtualMachineDescription& desc, const QString& default_mac_addr)
+QJsonObject generate_devices_config(const multipass::VirtualMachineDescription& desc, const QString& default_mac_addr,
+                                    const QString& storage_pool)
 {
     QJsonObject devices{{"config", QJsonObject{{"source", "cloud-init:config"}, {"type", "disk"}}},
                         {"root", QJsonObject{{"path", "/"},
-                                             {"pool", "default"},
+                                             {"pool", storage_pool},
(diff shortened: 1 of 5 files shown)


Saviq:
The `multipass` pool should be used if it exists, shouldn't it? I mean:

```
if pool("multipass"): return pool("multipass")
elif pool("default"): return pool("default")
else create_pool("multipass"): return pool("multipass")
```

townsend2010:
> The `multipass` pool should be used if it exists, shouldn't it?

Hmm, I guess.  The advice was to use the `default` pool first, if it exists.  If it doesn't, then create/use a `multipass` pool.  But I suppose if there is already a `multipass` pool available, then at some point, there wasn't a `default` pool.  But I do see what you mean, if we did create a `multipass` pool and the user decided to create a `default` pool later, we should continue to use the `multipass` pool, so I'll fix that.

Saviq:
> But I do see what you mean, if we did create a `multipass` pool and the user decided to create a `default` pool later, we should continue to use the `multipass` pool, so I'll fix that.

We're missing a test case for when both of the pools exist, don't we? And we lost the profile update test?

townsend2010:
> We're missing a test case for when both of the pools exist, don't we?

Ok, right.

> And we lost the profile update test?

We no longer are updating the default profile with the PR, hence dropping the test for that.

townsend2010:
> We're missing a test case for when both of the pools exist, don't we?

Done.

Saviq:
> at some point, there wasn't a `default` pool.

That, or someone created the `multipass` pull manually.

I triggered CI to get new builds and will test soon.

Saviq (approved):
Works well!

I launched an instance with LXD not initialized, then another one in the `multipass` pool, then removed those and the pool, and launched one in the `default` pool after. And all that without restarting the daemon, so good!

Saviq:
bors merge

Saviq:
@townsend2010 this needs an update for clang-format-12 and some conflict resolution.

townsend2010:
bors r=Saviq

PR #2135, opened 2021-06-04
[tests] Fix Timer.resumes race in CI

townsend2010:


Saviq:
Looks legit!

bors merge

PR #2137, opened 2021-06-07
Preparing the ground to create switches in Hyper-V

ricab:


tests/lxd/test_lxd_backend.cpp
```diff
@@ -1897,3 +1898,38 @@ TEST_F(LXDBackend, posts_network_data_config_if_available)
...
+{
+    CustomLXDFactory backend{std::move(mock_network_access_manager), data_dir.path(), base_url};
+    std::vector<mp::NetworkInterface> extra_networks{{"netid", "mac", false}};
+
+    auto address = [](const auto& v) { return &v; }; // replace with Address matcher once available
+    EXPECT_CALL(backend, prepare_networking_guts(ResultOf(address, Eq(&extra_networks)), Eq("bridge")));
+    backend.prepare_networking(extra_networks);
+}
+
+TEST_F(LXDBackend, createsBridgesViaBackendUtils)
```
Saviq, line 1925:
```suggestion
TEST_F(LXDBackend, creates_bridges_via_backend_utils)
```

A bunch more below that are camelCase.

tests/test_base_virtual_machine_factory.cpp
```diff
@@ -121,3 +135,167 @@ TEST_F(BaseFactory, creates_cloud_init_iso_image)
...
+        for (size_t i = 0; i < maxim; ++i) // alternate to avoid testing only partitioned extra_nets
+            for (const auto& requests : {requested_bridged, requested_unbridged, requested_existing_bridges})
+                if (i < requests.size())
+                    ret.push_back(requests[i]);
+
+        return ret;
+    }
+};
+
+TEST_P(TestBridgePreparation, prepareNetworkingGutsCreatesBridgesAndReplacesIds)
```
Saviq, line 225: This test a bit unwieldy, could split in the three/four cases it handles?

Saviq (changes requested):

tests/lxd/test_lxd_backend.cpp
ricab: I asked the same thing to @townsend2010 recently and he reminded me that, as I had pointed out to him, [gtest advises against that](https://github.com/google/googletest/blob/master/docs/faq.md#why-should-test-suite-names-and-test-names-not-contain-underscore) :slightly_smiling_face: So he's been switching to camelCase recently and I thought it only made sense to start doing the same... WDYT?

tests/test_base_virtual_machine_factory.cpp
ricab: Yeah it is indeed, and it was really hard to get right! It was approx. a full day's worth of grind, going different tests, then single, then parameter, then back and forth again, getting the matchers to work...

The problem is that we need to match against a container, with multiple elements, corresponding to different scenarios. It would be much simpler to test single elements, but the code could break on multiple. And if I test multiple, even with all elements corresponding to a single scenario (say all existing bridges), I cover single. We'd then still need to test the heterogeneous case, which again I couldn't test without covering all single cases anyway. So I just went with that. 

The alternative would be to expose a more granular (mockable) interface only for the purpose of testing, i.e. `prepare_single_interface(...)` which would be unit tested, and mocked when testing the whole `prepare_networking()`. Maybe that's what I should have done now that I think about it from a distance, even though I dislike tailoring interfaces to tests... Is that what you think should happen? Otherwise how do you think I should approach this concretely?

tests/lxd/test_lxd_backend.cpp
Saviq: +1 on camelCase, then, as long as we try and be consistent...

tests/test_base_virtual_machine_factory.cpp
ricab: OK, I'll see what I can do. I don't think `prepare_networking_guts` can be any simpler, but perhaps I can expose a more elementary &ndash; and mockable &ndash; interface, for testing.

tests/test_base_virtual_machine_factory.cpp
```diff
@@ -121,3 +135,167 @@ TEST_F(BaseFactory, creates_cloud_init_iso_image)
...
+        {"eth0", "ethernet", "e0"}, {"eth1", "ethernet", "e1"},       {"wlan0", "wifi", "w0"},
+        {"wlan1", "wifi", "w1"},    {"br0", br_type, "b0", {"eth0"}}, {"br1", br_type, "b1", {"wlan1"}},
+        {"br2", br_type, "empty"}};
+    std::vector<mp::NetworkInterface> mix_extra_nets() const // mixes all requested
+    {
+        std::vector<mp::NetworkInterface> ret;
+        const auto& [requested_bridged, requested_unbridged, requested_existing_bridges] = GetParam();
+
+        auto maxim = std::max(requested_bridged.size(), requested_unbridged.size());
+        maxim = std::max(maxim, requested_existing_bridges.size());
```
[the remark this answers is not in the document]
ricab: No, that is much better. I missed that `initializer_list` overload, thanks!

tests/test_base_virtual_machine_factory.cpp
```diff
@@ -121,3 +135,167 @@ TEST_F(BaseFactory, creates_cloud_init_iso_image)
...
+    // Now all predicates and matchers that we need to check extra_nets...
+    auto is_same_bridge = [&same_id](const auto& a, const auto& b) { return same_id(a, b) && b.type == br_type; };
+    auto is_old_bridge = mpt::HasCorrespondentIn(host_nets, is_same_bridge);
+    auto is_unrecognized_network = Not(mpt::HasCorrespondentIn(host_nets, same_id));
+    auto is_old_bridge_or_unrecognized = AnyOf(is_old_bridge, is_unrecognized_network);
+    auto was_requested = AnyOf(mpt::ContainedIn(requested_unbridged), mpt::ContainedIn(requested_bridged),
+                               mpt::ContainedIn(requested_existing_bridges));
+
+    auto is_new_bridge = Field(&mp::NetworkInterface::id, EndsWith("br"));
+    auto starts_with_id = [](const auto& a, const auto& b) { return a.id.rfind(b.id) == 0; };
```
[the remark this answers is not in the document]
ricab: Oops, I meant to pass 0 as second argument but forgot. Good catch.

tests/test_base_virtual_machine_factory.cpp
ricab: Oh yes, simpler indeed in that it offloads the body of the `for`. I think I have a path for this.

tests/test_base_virtual_machine_factory.cpp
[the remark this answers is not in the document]
Saviq: Well, I mean if `a.id` is "foofoo" and `b.id` is "foo", this will return false, yet it _does_ "start with" "foo". Why not plain `find`?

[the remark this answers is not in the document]
ricab: It [wouldn't return false](https://godbolt.org/z/xeTc4hf3M). Notice that `rfind` itself returns 0 when it "starts with" (0 is the starting position of the substring). Plain `find` would work, but keep looking for substrings starting at all positions, which is unnecessary. This is a [customary approach](https://stackoverflow.com/questions/1878001/how-do-i-check-if-a-c-stdstring-starts-with-a-certain-string-and-convert-a) before `starts_with` is available (as in C++20).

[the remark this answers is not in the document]
Saviq: Oh indeed, with `pos=0`, it works as intended. I misinterpreted `pos` there.

tests/test_base_virtual_machine_factory.cpp
ricab: Done, let me know what you think!

Saviq: Yeah, much better, and worth the change in implementation :)

Saviq (approved):
bors merge

Saviq:
> Merge conflict.

Ah, you'll need a rebase :)

Saviq:
bors merge

Saviq:
bors retry

PR #2142, opened 2021-06-22
Teach GTest to print QStrings while avoiding breaking ODR

ricab:
The motivation for this is to teach Google Test to print `QStrings` and other types (e.g. when things fail). I had attempted QString a few times before with no luck. It turns out that the printers need to be visible to all files that declare tests, to [avoid breaking ODR](https://stackoverflow.com/a/36941270). 

So, this replaces the usual gmock/gtest includes with a single `common.h` include which includes those and defines the extra prints. It also provides the extra assertions and matchers. Note that this applies to all printers, namely ones we implement in the future and the ones for network structs introduced in #2137, although in that case things seemed to work locally (but still UB).

Since I needed to change the include order in a bunch of files, I went all the way and organized includes consistently and in [increasing order of specificity](https://www.accu.org/journals/overload/24/133/overload133.pdf#page=9), fixed ensuing compilation problems and dropped a few unused includes that CLion identified.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,6 +21,7 @@ set(MULTIPASS_GMOCK_DIR ${CMAKE_SOURCE_DIR}/3rd-party/grpc/third_party/googletes
 set(MULTIPASS_GTEST_DIR ${CMAKE_SOURCE_DIR}/3rd-party/grpc/third_party/googletest/googletest)
 
 add_executable(multipass_tests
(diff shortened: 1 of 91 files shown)


ricab:
No merge until the base PR is in.

townsend2010 (approved):
LGTM

bors merge

townsend2010:
Right, bors is gonna fail.  Will merge manually.

Issue #1953, opened 2021-02-02
Image filtering for disallowing auto mode --network should be cleaned up


townsend2010:
The image filtering for disallowing `auto` mode `--network` takes liberties with the remote.

For example, in #1952 the `core`/`core16` alias is not really part of a proper remote, yet the output is:
```
launch failed: Automatic network configuration not available for release:core. Consider using manual mode.
```

even though `release` was not specified at launch nor is `core` part of the `release` remote.

PR #2149, opened 2021-06-24
Fix non-bridging check for images

luis4a0:
This PR fixes https://github.com/canonical/multipass/issues/1953.

The check assumed the remote to be `release` when the user did not specify one. This is not true, because there are images which do not accept a remote name (such as `core`). This worked but it was confusing in the code. It was fixed to check separately such images (called 'remoteless images' in the code). The error message was corrected as well, not showing a remote name if the user did not specify one.

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -1071,8 +1071,11 @@ std::vector<std::string> old_releases{"10.04",   "lucid",  "11.10",   "oneiric",
                                       "14.10",   "utopic", "15.04",   "vivid",   "15.10", "wily",    "16.04",
                                       "xenial",  "16.10",  "yakkety", "17.04",   "zesty"};
 
+std::vector<std::string> old_remoteless_rels{"core", "core16"};
+
 INSTANTIATE_TEST_SUITE_P(DaemonRefuseRelease, RefuseBridging, Combine(Values("release", ""), ValuesIn(old_releases)));
 INSTANTIATE_TEST_SUITE_P(DaemonRefuseSnapcraft, RefuseBridging, Values(std::make_tuple("snapcraft", "core")));
+INSTANTIATE_TEST_SUITE_P(DaemonRefuseRemoteless, RefuseBridging, Combine(Values(""), ValuesIn(old_remoteless_rels)));
 
 constexpr auto ghost_template = R"(
 "{}": {{
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -1073,7 +1073,8 @@ std::vector<std::string> old_releases{"10.04",   "lucid",  "11.10",   "oneiric",
 
 std::vector<std::string> old_remoteless_rels{"core", "core16"};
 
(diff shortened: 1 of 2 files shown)


townsend2010 (commented):
The output is what I'd expect now.  Thinking about this though, I'm wondering if having the older releases in the map is necessary since the only backends that support this (correct me if I'm wrong) are LXD, Virtualbox, and Hyper-V and none of those support these older releases.

For example, if I do the following with the LXD driver:
`multipass launch --network foo wily`
I get an error that leads me to believe that `wily` will work, just not with `--network`, but that is not true.  `wily` is simply not supported with our LXD backend.

To me, this should first check if a backend supports the image and if it does, then the `--network` option should be validated.

I'm not approving nor requesting changes at this point, but rather, opening this up for discussion.

luis4a0:
Hey @townsend2010, thanks for the review!

I think you're right about validating first if the image can be launched on that specific backend.

OTOH, there are things to be discussed about not including in the list images which cannot be bridged:
1. does this check need to be a backend-specific feature?
2. we don't gain much by removing some elements from the list, but that can bite us if we implement bridging on a new backend and forget to add the old images to that list.

townsend2010:
Hey @luis4a0!

Ok, I don't have a problem leaving the set intact since it's not huge and running `find()` thorough it is negligible performance-wise.  However, if we do leave the whole set, we need to account for older `daily` builds as well.

So, I think we are fine if we first validate if the image is indeed supported for the current backend and if so, then validate if the image will work if `--network` is given.

tests/test_daemon.cpp
```diff
@@ -1078,6 +1078,32 @@ INSTANTIATE_TEST_SUITE_P(DaemonRefuseRelease, RefuseBridging,
...
+    std::vector<mp::VMImageHost*> hosts;
+    hosts.push_back(mock_image_host_ptr);
+
+    mp::URLDownloader downloader(std::chrono::milliseconds(1));
+    mp::DefaultVMImageVault default_vault(hosts, &downloader, mp::Path("/"), mp::Path("/"), mp::days(1));
+
+    auto mock_image_vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
+    auto mock_image_vault_ptr = mock_image_vault.get();
+
+    EXPECT_CALL(*mock_image_vault_ptr, all_info_for(_)).Times(1).WillOnce([&default_vault](const mp::Query& query) {
```
townsend2010, line 1095: A single `WillOnce` implies `Times(1)`, so you can drop the `Times` here if you'd like, although it's not necessary.

townsend2010 (commented):
Great, LGTM.  I left one little comment below.  It's not necessary, but I'll let you decide before I get this in.

tests/test_daemon.cpp
luis4a0: Oh, didn't know that. I removed `Times(1)` to make the code clearer. Thanks!

townsend2010 (approved):
Ok, good to go!  Thanks!

bors merge

townsend2010:
Ugh...

bors retry

Issue #1718, opened 2020-09-03
File/directory ownership unmapped if changed with `chown`


facundobatista:
```bash
$ multipass launch -n test
$ mkdir foo
$ ls -ld foo
drwxrwxr-x 2 michal michal 2 wrz  3 14:10 foo
# correct ownership on host

$ multipass mount foo test:foo --uid-map 1000:0 --gid-map 1000:0
$ multipass exec test -- ls -ld foo
drwxrwxr-x 2 root root 2 wrz  3 14:10 foo
# correct ownership inside the instance

$ multipass exec test -- sudo chown 0:0 foo
$ ls -ld foo
drwxrwxr-x 2 root root 2 wrz  3 14:10 foo
# this should be michal:michal still!
```

facundobatista:
Just found that if internally I copy the stuff by doing `cp -r --preserve=mode,timestamps otherdir/stuff/ somedir/`, internally I see them as `root`, but from outside everything is owned by `facundo`!!

Saviq:
> Just found that if internally I copy the stuff by doing `cp -r --preserve=mode,timestamps otherdir/stuff/ somedir/`, internally I see them as `root`, but from outside everything is owned by `facundo`!!

Yes, it all suggests we forget to make a reverse ownership mapping when `chown` happens.

BenWibking:
Is there any update on this?

Saviq:
@BenWibking thanks for the bump. We'll schedule this to fix, soon.

BenWibking:
Did this get fixed in version 1.7?

townsend2010:
Hey @BenWibking!

We started to work on it, but found that it had more implications that need consideration such as security, etc. and did not want to block the release for this.

That said, we are going to discuss internally how best to handle this and it should be addressed in the next release.

yingshaoxo:
I could suggest you guys use this to make sure all the permission is right in side of multipass:

```bash
sudo chown -R $(whoami):$(whoami) .
sudo chmod g+rw .
```

PR #2154, opened 2021-07-01
Fix `chown` in the host if uid/gid is reverse mapped

luis4a0:
Fixes https://github.com/canonical/multipass/issues/1718.

### DIFF
--- a/include/multipass/id_map.h
+++ b/include/multipass/id_map.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Canonical, Ltd.
+ * Copyright (C) 2021 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -15,29 +15,15 @@
  *
  */
 
-#ifndef MULTIPASS_SSHFS_SERVER_CONFIG_H
-#define MULTIPASS_SSHFS_SERVER_CONFIG_H
+#ifndef MULTIPASS_ID_MAP_H
+#define MULTIPASS_ID_MAP_H
 
-#include <multipass/id_map.h>
-
-#include <string>
-#include <unordered_map>
+#include <utility>
+#include <vector>
 
 namespace multipass
 {
+typedef typename std::vector<std::pair<int, int>> id_map;
+}
 
-struct SSHFSServerConfig
-{
-    std::string host;
-    int port;
-    std::string username;
-    std::string instance;
-    std::string private_key;
-    std::string source_path;
-    std::string target_path;
-    id_map gid_map;
-    id_map uid_map;
-};
-
-} // namespace multipass
-#endif // MULTIPASS_SSHFS_SERVER_CONFIG_H
+#endif // MULTIPASS_ID_MAP_H
--- a/include/multipass/id_relationship_list.h
+++ b/include/multipass/id_relationship_list.h
@@ -23,7 +23,7 @@
 
 namespace multipass
 {
-typedef typename std::vector<std::pair<int, int>> id_map;
+typedef typename std::vector<std::pair<int, int>> id_relationship_list;
 }
 
 #endif // MULTIPASS_ID_MAP_H
--- a/src/rpc/multipass.proto
+++ b/src/rpc/multipass.proto
@@ -176,14 +176,14 @@ message InfoRequest {
     int32 verbosity_level = 2;
 }
 
-message MapPair {
+message IdMap {
     int32 host_id = 1;
     int32 instance_id = 2;
 }
 
 message MountMaps {
-    repeated MapPair uid_mappings = 1;
-    repeated MapPair gid_mappings = 2;
+    repeated IdMap uid_mappings = 1;
+    repeated IdMap gid_mappings = 2;
 }
 
 message MountInfo {
--- a/src/rpc/multipass.proto
+++ b/src/rpc/multipass.proto
@@ -182,8 +182,8 @@ message MapPair {
 }
 
 message MountMaps {
-    repeated MapPair uid_map = 1;
-    repeated MapPair gid_map = 2;
+    repeated MapPair uid_mappings = 1;
+    repeated MapPair gid_mappings = 2;
 }
 
 message MountInfo {
--- a/src/rpc/multipass.proto
+++ b/src/rpc/multipass.proto
@@ -176,9 +176,14 @@ message InfoRequest {
     int32 verbosity_level = 2;
 }
 
+message MapPair {
+    int32 host_id = 1;
+    int32 instance_id = 2;
+}
+
 message MountMaps {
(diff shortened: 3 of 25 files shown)


src/sshfs_mount/sftp_server.cpp
```diff
@@ -334,6 +336,24 @@ int mp::SftpServer::mapped_gid_for(const int gid)
     return gid;
 }
 
+int mp::SftpServer::reverse_uid_for(const int uid, const int rev_uid_if_not_found)
+{
+    auto compare_uid_map_value = [uid](std::pair<int, int> p) { return uid == p.second; };
+
+    auto found = std::find_if(uid_map.cbegin(), uid_map.cend(), compare_uid_map_value);
```
townsend2010, line 343: Why not just construct the lambda directly in the `find_if` function like you've done elsewhere?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -334,6 +336,24 @@ int mp::SftpServer::mapped_gid_for(const int gid)
...
+
+    return found == uid_map.cend() ? rev_uid_if_not_found : found->first;
+}
+
+int mp::SftpServer::reverse_gid_for(const int gid, const int rev_gid_if_not_found)
+{
+    auto compare_gid_map_value = [gid](std::pair<int, int> p) { return gid == p.second; };
+
+    auto found = std::find_if(gid_map.cbegin(), gid_map.cend(), compare_gid_map_value);
```
townsend2010, line 352: idem - lambda directly in `find_if`.

townsend2010 (changes requested):
Hey @luis4a0,

Thanks for doing this.  The solution seems to work well for me.

Code-wise, I'm really not a fan of calling this a "map" anymore since it's ceasing to be a map with these changes.  The only place where we may possibly want to leave the word "map" in place is for the `MountMaps` message name and anywhere there is referenced since we don't really want to break API/ABI.

include/multipass/id_map.h
```diff
@@ -0,0 +1,29 @@
...
+
+#ifndef MULTIPASS_ID_MAP_H
+#define MULTIPASS_ID_MAP_H
+
+#include <utility>
+#include <vector>
+
+namespace multipass
+{
+typedef typename std::vector<std::pair<int, int>> id_map;
```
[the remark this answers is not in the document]
luis4a0: Done. I changed the name of the data structure, as you suggest. However, I didn't change the names `uid_map` and `gid_map` because I think those names make the interfaces easier to understand. WDYT?

src/sshfs_mount/sftp_server.cpp
luis4a0: Done.

src/sshfs_mount/sftp_server.cpp
luis4a0: Done.

include/multipass/id_relationship_list.h
```diff
@@ -0,0 +1,29 @@
...
+
+#ifndef MULTIPASS_ID_MAP_H
+#define MULTIPASS_ID_MAP_H
+
+#include <utility>
+#include <vector>
+
+namespace multipass
+{
+typedef typename std::vector<std::pair<int, int>> id_relationship_list;
```
townsend2010, line 26: Eh, that name I suggested is not good.  Since we are keeping the word "map" in the variable names, let's change it to `id_map_list`.  Don't forget to change the `#define` name to match.

townsend2010 (changes requested):
Ok, let's change up the typename and anywhere where a variable is defined with that type, let's change the name to `foo_map_list`.  For example, `const id_map_list gid_map` would become `const id_map_list gid_map_list`.

include/multipass/id_relationship_list.h
townsend2010: Actually, after some discussion with @Saviq, let's call this `id_mappings` to help differentiate what this actually is.  And then for the `gid_map` name, it would `gid_mappings` and similar for others.

townsend2010: Hey @luis4a0,

Sorry, I guess you missed my last comment to call it `id_mappings` and such :frowning_face:

luis4a0: Yes, indeed @townsend2010. GH does not update automatically in some cases and I missed that last message. Let me take care of that. Thanks!

src/rpc/multipass.proto
```diff
@@ -176,9 +176,14 @@ message InfoRequest {
     int32 verbosity_level = 2;
 }
 
+message MapPair {
```
townsend2010, line 179: To be consistent when you rename `id_map_list` to `id_mappings`, let's call this `IdMap`.

src/sshfs_mount/sftp_server.cpp
```diff
@@ -334,6 +338,22 @@ int mp::SftpServer::mapped_gid_for(const int gid)
...
+    return found == uid_map_list.cend() ? rev_uid_if_not_found : found->first;
+}
+
+int mp::SftpServer::reverse_gid_for(const int gid, const int rev_gid_if_not_found)
+{
+    auto found = std::find_if(gid_map_list.cbegin(), gid_map_list.cend(),
+                              [gid](std::pair<int, int> p) { return gid == p.second; });
+
+    return found == gid_map_list.cend() ? rev_gid_if_not_found : found->first;
+}
```
townsend2010, line 355: This code is pretty redundant and seems like it can be refactored into an anonymous function that accepts the map, id, and default id.  Something like:
```c++
namespace
{
int reverse_id_for(const id_mappings& id_maps, const int id, const int rev_id_if_not_found)
{
    auto found = std::find_if(id_maps.cbegin(), id_maps.cend(), [&id](std::pair<int, int> p) {
        return id == p.second;
    });

    return found == id_maps.cend() ? rev_id_if_not_found : found->first;
}
}
```

townsend2010 (changes requested):
A few more comments inline.

include/multipass/id_relationship_list.h
luis4a0: Done!

src/rpc/multipass.proto
luis4a0: Done!

src/sshfs_mount/sftp_server.cpp
luis4a0: Right, done. Refactored forward and reverse mapping functions.

luis4a0:
Oops, forgot `git clang-format` in some commit. Fixing...

luis4a0:
Done.

townsend2010 (approved):
Ok, LGTM.  Thanks!

bors merge

Saviq:
bors retry

Saviq:
Bup bup.

bors retry

Saviq:
Argh.

bors retry

PR #2169, opened 2021-07-09
[daemon] Fix memory reported in older ubuntu vms

### LINKED ISSUE #2168: Info reports ten-fold memory for older ubuntu instances
Memory in `multipass info` is wrong in some cases. For instance, on a trusty instance it claims 9.7G of memory:

```
ricab@ricab-xps13:~ $ multipass info prodigious-bullsnake
Name:           prodigious-bullsnake
State:          Running
IPv4:           10.81.153.78
Release:        Ubuntu 14.04.6 LTS
Image hash:     3c4ad0defbe7 (Ubuntu 14.04 LTS)
Load:           0.00 0.02 0.04
Disk usage:     807.1M out of 4.9G
Memory usage:   2.7G out of 9.7G
Mounts:         --
```

Which is 10x the actual memory:

```
ricab@ricab-xps13:~ $ multipass exec prodigious-bullsnake -- free -hm
             total       used       free     shared    buffers     cached
Mem:          993M       281M       712M       352K       7.9M       228M
-/+ buffers/cache:        44M       948M
Swap:           0B         0B         0B
```

The cause is an issue with the command we use to inspect memory: `free -b | sed '1d;3d' | awk '{printf $2}'`. That `sed` excludes the first and third lines, but that leaves the swap line there. `awk` then concatenates the two, so there is one 0 too much:

```
ubuntu@prodigious-bullsnake:~$ free -b | sed '1d;3d'
Mem:    1041985536  297934848  744050688     356352    8613888  240173056
Swap:            0          0          0
ubuntu@prodigious-bullsnake:~$ free -b | sed '1d;3d' | awk '{printf $2}'
10419855360ubuntu@prodigious-bullsnake:~$
```

ricab:
Fixes #2168.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1318,8 +1318,8 @@ try // clang-format on
                                    *config->ssh_key_provider};
 
             info->set_load(mpu::run_in_ssh_session(session, "cat /proc/loadavg | cut -d ' ' -f1-3"));
-            info->set_memory_usage(mpu::run_in_ssh_session(session, "free -b | sed '1d;3d' | awk '{printf $3}'"));
-            info->set_memory_total(mpu::run_in_ssh_session(session, "free -b | sed '1d;3d' | awk '{printf $2}'"));
+            info->set_memory_usage(mpu::run_in_ssh_session(session, "free -b | grep 'Mem:' | awk '{printf $3}'"));
+            info->set_memory_total(mpu::run_in_ssh_session(session, "free -b | grep 'Mem:' | awk '{printf $2}'"));
             info->set_disk_usage(mpu::run_in_ssh_session(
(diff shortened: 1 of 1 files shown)


ricab:
I tried to write a unit test for this, but it would involve mocking SSH-ing into a VM and execution of commands therein. While doable, it would take a while, both via mocking `mp::SSHSession` (would need a new factory) and via premocked ssh functions. I am not sure this is the right PR for that. Note that no other `Daemon::info` tests exist at all.

So, marking this ready. Let me know if you think otherwise.

townsend2010 (approved):
Yeah, it's fine that this has no tests.  We'll get them in when we do all the daemon testing :slightly_smiling_face: 

LGTM

bors merge

Saviq:
bors retry

PR #2171, opened 2021-07-09
[daemon] Fix regression in Workflows-based launches

### LINKED ISSUE #2170: [regression] Workflow based launches are broken
In the 1.7 RC, trying to launch a workflow fails:
```
$ multipass launch minikube
launch failed: Unable to find an image matching "minikube"
```

This is another regression due to https://github.com/canonical/multipass/pull/2149/commits/f91fe3c34561478795787e702d91f10ac85c885d.

townsend2010:
Fixes #2170

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -468,8 +468,11 @@ void validate_image(const mp::LaunchRequest* request, const mp::VMImageVault& va
     }
 }
 
-auto validate_create_arguments(const mp::LaunchRequest* request, const mp::VirtualMachineFactory& factory)
+auto validate_create_arguments(const mp::LaunchRequest* request, const mp::DaemonConfig* config)
 {
+    assert(config && config->factory && config->workflow_provider && config->vault && "null ptr somewhere...");
+    validate_image(request, *config->vault, *config->workflow_provider);
+
     static const auto min_mem = try_mem_size(mp::min_memory_size);
     static const auto min_disk = try_mem_size(mp::min_disk_size);
     assert(min_mem && min_disk);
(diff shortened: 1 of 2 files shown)


ricab (commented):
I am ok with this in general, but I would ask that the check be extracted into a `validate_image` auxiliary function (receiving the `vault` and the `workflow_provider`) and that that function be called from `validate_create_arguments`, which would need to receive the full config (or even be made a private method of the daemon). 

That would allow dropping the comment explaining that the purpose of the code is validating the image (clear from the function name) and avoid adding to an already very long function, where individual calls are easily missed.

ricab (approved):
Good by me.

bors d=townsend2010

src/daemon/daemon.cpp
```diff
@@ -451,8 +451,28 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequ
     return interfaces;
 }
 
-auto validate_create_arguments(const mp::LaunchRequest* request, const mp::VirtualMachineFactory& factory)
+void validate_image(const mp::LaunchRequest* request, const mp::VMImageVault& vault,
+                    mp::VMWorkflowProvider& workflow_provider)
```
townsend2010, line 455: Could be `const`, right?

townsend2010 (commented):
One thing inline.

src/daemon/daemon.cpp
ricab: No, I tried that first, but we call `info_for` which isn't `const` (it changes `needs_update`).

townsend2010: Ah, right, ok.

townsend2010:
bors r=ricab

townsend2010:
Sigh...

bors retry

Issue #1040, opened 2019-09-11
[launch] allow cloud-init file injection


Saviq:
Some cloud-init modules (e.g. [scripts](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#scripts-per-boot)) require files to be placed into the data source.

We need a `--cloud-init-files` or so that will allow the user to specify a file/directory tree to be included on the data source.

v1k0d3n:
would love to see this implemented. where does this currently fall on the release todo list @Saviq? i can see this being extremely useful for some of our use cases.

Saviq:
Hey @v1k0d3n I don't have good news about this, either ;)

It's not something on our immediate roadmap, but it should be relatively easy to add if you wanted to contribute.

The cloud-init ISO is created [here, in the `daemon.cpp` code](https://github.com/CanonicalLtd/multipass/blob/b22ff56651c225d2e9cb219703a99f8ac6a21d0c/src/daemon/daemon.cpp#L147) - it would be easy to feed more files in.

v1k0d3n:
@Saviq more bad news!? agh. :) j/k

when I have some free time, I'll see if it's something I can knock out. it's at least helpful to know that it's not on product's immediate radar, but I'm surprised more users aren't asking for the feature. this feature seems like a great way to differentiate multipass vs. containerized use cases.

surahman:
I am trying to setup a plan to resolve this, here is what I have:

1. Add position argument `--cloud-init-files` to the `launch` command.
2. Update the help message.
3. Add an argument with the `scripts` path (default empty string) value to: https://github.com/canonical/multipass/blob/b22ff56651c225d2e9cb219703a99f8ac6a21d0c/src/daemon/daemon.cpp#L147-L148

What should the record structure be for the `scripts` directory for the cloud-init YAML file? What structure/how should I be adding the scripts to the ISO?

Saviq:
Hi @surahman, my current idea for this would be a `--cloud-init-tree` argument that would take a path as the argument and that would be the entire file structure injected into the ISO.

It would be a double-edged sword as I think we wouldn't override `network-config` or `vendor-data` if it existed, so users could potentially break some of Multipass assumptions. I was thinking we'd have a way to generate a "pristine" tree for users to modify, with comments to what each of the entries are responsible for.

surahman:
> my current idea for this would be a `--cloud-init-tree` argument that would take a path as the argument and that would be the entire file structure injected into the ISO.

[...]

PR #2173, opened 2021-07-12
Cloud Init Tree

surahman:
**_Issue #1040:_**

#### Feature: `--cloud-init-tree`
Enhancement of the `launch` command on the `client` which takes a path as a parameter. The path points to a directory that contains a nested directory-file structure that will comprise the ISO's image to use for `cloud-init`.

##### Kanban Board:
Current: https://github.com/surahman/multipass/projects/2
Old: https://github.com/surahman/multipass/projects/1


surahman:
@saviq Is there a utility file I can place the `extract_iso_structure` routine in? I have tested the code outside of the multipass code base and it works as expected. I have not created and wired those tests up here yet, pending a more appropriate location for the routine.

Saviq:
Hi @surahman, thanks again for awesome contributions!

> @Saviq I have begun with this feature and I will try to take it as far as I can. I am attempting to offload as much from the backend daemon process to the client (SOP in distributed systems). Project [Kanban board](https://github.com/surahman/multipass/projects/1).

I don't think that's the right direction. It will mean extra onus on any additional clients that come up in the future. And creates a risky binding between the client and daemon for what goes into `vendor-data`. The client should transfer the file tree, not a "baked" ISO, and the daemon needs to decide what exactly to do with the files it receives.
 
>     * The [CloudInitIso](https://github.com/canonical/multipass/blob/4a4a77106c1417c76a8f2a1ef003fd40e059dcb8/include/multipass/cloud_init_iso.h) class appears to only have an `add_file` method which takes a `string` as input. Your plan was to have the files (potentially large) streamed over as byte streams via `gRPC`, and these would need to be written chunk-by-chunk as they arrive and not kept memory resident. I will require this functionality from the ISO write utility; is this available?

No, we've not had a need for anything like this so far. But your explanation of what needs to happen looks right.

> Aside from this, I need to know if I can extend the `LaunchRequest` with a boolean flag to switch to the `cloud-init-tree` handler on the backend daemon (already done, please see _files changed_).
> 
> I will also need to extend the `gRPC` protocol buffer to support the Client to Server `FileXfer` with a message(s) and a service. Please see the algorithms outlined below, the `Client` side of is already stubbed on the `dev` branch. The logline reply from the daemon will contain the path & filename, as well as the number of bytes written. Upon file xfer completion the file will be verified against its `sha256` hash for integrity. Protobuf IDL:

That sounds right, except I think it has to be part of the `Launch` request, using the streaming feature of gRPC. If the file transfer is a completely separate gRPC routine, how do you "bind" the transferred file to a launch request?

> The `LaunchRequest` message triggers the launch process on the daemon, and `request_launch` handles the communication with the client. To send a stream of ISO file data to the backend is not tricky, but handling the request is. You would need to ensure that either:
> 
>     1. The ISO arrives first, then `LaunchRequest` is triggered.
> 
>     2. The request to transfer the file happens mid `LaunchRequest`.
> 
>     3. The file transfer happens after the `LaunchRequest` and that it waits for the ISO to fully transfer.
> 
> 
> Not too sure about how to proceed with this because it has broad and far-reaching architectural implications. I am not completely familiar with the architecture or deeper code base. This task touches a lot of surfaces.
> 
> A contrived way of achieving this would be to nest the `LaunchRequest` into `FileXferRequest` and also allow that to trigger a launch.
> 
> My preferred solution would be to initiate the ISO file transfer in the `request_launch` routine and transfer over the entire ISO before allowing control flow to proceed to handle the actual `LaunchRequest` as it is now. The file transfer RPC would need to return an ISO name (random string generated on the daemon) and include it in the `LaunchRequest`.

Right, so that's exactly the problem I mention above. Conceptually, what I think needs to happen is:
1. client initiates a streaming `LaunchRequest`, signaling that there will be files streamed
2. daemon receives the request and waits for file transfer to complete
3. client transfers all the files (ideally multithreaded)
4. client signals to the daemon that all the files are sent
5. daemon continues with the launch

This way all the files getting transferred are bound to the request itself, and if anything goes wrong they can be easily cleared up.

The file transfer would ideally be encapsulated enough that it can later be used for transferring image files from the client, and even in `multipass transfer` so that we can stop relying on direct network communication between the client and instance.

surahman:
> That sounds right, except I think it has to be part of the `Launch` request, using the streaming feature of gRPC. If the file transfer is a completely separate gRPC routine, how do you "bind" the transferred file to a launch request?

This is a very good point and your comment which follows would make the whole process cleaner and smoother. I think this will then necessitate the restructuring or encapsulation of the core `LaunchRequest` into another request to trigger the launch? `LaunchRequest`, as it is, is not a streaming request but would need to be to facilitate a file transfer?  There would need to be a `oneof` message with options for the core `LaunchRequest`, and then ones for the file metadata and data blocks.

We could nest the gRPC requests so that we call the file transfer RPC request from within/during the `LaunchRequest` from the `Client`. I think there would then need to be some sort of nested request while the `Client` is executing/issued the `LaunchRequest`, then the `Daemon` would need to send an okay status to kick off the transfer. I believe that would be very clean solution-wise, assuming we can nest the RPCs.

surahman:
@Saviq I was thinking a little bit about the task at hand and have written up my thoughts:

I believe, without knowing what else is on the horizon, that it might be worthwhile performing the file transfers over raw `TCP/IP` rather than `gRPC`. `gRPC` operates on `TCP/IP` but is a layer above transport, so it is a degree of indirection. I think `gRPC` also expends a significant amount of time allocating/deallocating on the heap, which in itself is an extremely time-consuming process. If the objective is merely to transfer files between the `Client` and `Daemon` then perhaps `gRPC` is not required?

From my research on message sizes in `gRPC` it appears that the sweet spot varies in the KB range, and this could result in lots of messages/packets. Some people have conducted experiments on message payload sizes, and smaller payloads fare better. With `TCP/IP` you would need to generate the payload and hand it off to the transport layer directly. It, in turn, would deal with the window sizing and congestion control without having the `gRPC` framework above it.

It would be possible to have a Client/Server `TCP/IP`, potentially with `TLS` enabled, using `Qt` with minimal code to maintain. You would not need to update the `LaunchRequest` to use nested `oneof` messages. This would, in turn, mean not having to touch any of the current tested and functioning code for updates to the IDL. The `LaunchRequest` would go ahead as usual, and if a file transfer is required, there would be a handoff midway in the `LaunchRequest` to the Client/Server to handle the transfer. Once the transfer is completed, the `LaunchRequest` would continue and use the ISO.

I believe this would be a more efficient solution, as well as being easier to maintain. Again, I do not know what else is on the horizon.

Saviq:
@surahman thanks for the write-up. While your characterization isn't wrong, it does miss a number of advantages that gRPC gives us that you would have to implement yourself:
- encryption
- multiplexing (multiple clients accessing the daemon at the same time)
- chunking
- concurrency
- ...

Like... if frameworks like gRPC only introduced overhead, and no advantages, who would ever write or use them? There are likely more efficient solutions, but that they're easier to maintain is arguable. gRPC is already integrated into Multipass, introducing another mode of client-server communication would have to be well argumented. And why not replace gRPC altogether at that point?

Without seeing data that gRPC would be _significantly_ slower than other solutions (taking into account the expected file sizes, which at least in this case will be rather small), I'd stick to gRPC.

surahman:
@Saviq you make very good points, all of the aforementioned utilities are provided out of the box by `gRPC`. I also agree that `gRPC` does provide a lot of excellent utilities and functionality which makes life easier. In researching, I came across this blog, and I do note that this is just one person's experiments and are potentially no representative of the broader set of cases: https://ops.tips/blog/sending-files-via-grpc/

I defer to your experience on the matter with you having spent significantly more time with it. Just providing my 2-cents 😉.

surahman:
Hi @Saviq, I have an updated plan of action here: https://github.com/surahman/multipass/projects/2
The planned order I will follow is from the bottom-up of the `To do` list.

PR #2191, opened 2021-08-05
Fix compilation with Clang 12 on Linux.

luis4a0:
Clang is picky with lambda captures.

### DIFF
--- a/tests/qemu/test_firewall_config.cpp
+++ b/tests/qemu/test_firewall_config.cpp
@@ -192,16 +192,17 @@ TEST_F(FirewallConfig, dtorDeleteErrorLogsErrorAndContinues)
 
 TEST_P(FirewallToUseTestSuite, usesExpectedFirewall)
 {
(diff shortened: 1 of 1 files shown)


ricab (approved):
Yeah, it's a pain that we can't capture structured bindings (until C++20 I think). FWIW, you can still use regular vars, to conserve the names and avoid repeating `get<>` if you need them more than once:

```c++
    const auto& param = GetParam();
    const auto& expected_firewall = std::get<0>(param);
    ...
```

It's perhaps not too relevant here though, so accepting.

bors r+

Saviq:
bors retry

Saviq:
bors merge

Saviq:
bors r=ricab

ricab:
Oh dear.

bors r+

luis4a0:
Let's retry... there is a couple of combinations backend/platform which still didn't fail. Come on CI, go for it! :joy:

ricab:
Sure, here goes

bors r+

Issue #2188, opened 2021-08-02
Every time I ignore the multipass for windows update prompt, it queues an additional event handler. All of them fire at once when I actually click it


parktheredcar:
**Describe the bug**
On windows, multipass update notifier is opening the github releases page many times when I click the update notification prompt - one tab for every time I didn't click it previously.

**To Reproduce**
How, and what happened?
1.  Install multipass for windows
2. Wait for a new release and ignore the update prompt notification
3. Maybe ignore the notification when it pops up a few more times. Keep count of how many times you ignored it (N times)
4. Finally, click the update notification. It opens the same multipass github releases page in N duplicate tabs
5. Close all the extra tabs

**Expected behavior**
Clicking the update notification one time opens one single tab.

**Logs**
n/a
**Additional info**
Windows 10 20H2

**Additional context**
n/a

Saviq:
Hi @parktheredcar, you should never ignore it, then! ;)

But for real, I believe the issue is that we indeed connect the action multiple times, we need to disconnect first:

https://github.com/canonical/multipass/blob/2126b95a23af6d7c97457d0c08fd47392bfe6a56/src/client/gui/gui_cmd.cpp#L252-L253

surahman:
@Saviq I was poking around the documentation and found a method to [check if a signal is connect](https://doc.qt.io/qt-5/qobject.html#isSignalConnected):
```c++
        const auto connection_check = QMetaMethod::fromSignal(&QSystemTrayIcon::messageClicked);
        if (!isSignalConnected(connection_check))
        {
            QObject::connect(&tray_icon, &QSystemTrayIcon::messageClicked,
                             [this] { QDesktopServices::openUrl(QUrl(update_action.whatsThis())); });
        }
```
I am not sure if this will alleviate the issue.

Saviq:
@surahman thanks, that indeed looks like it will do the trick, can you put up a PR and we can take it from there?

PR #2193, opened 2021-08-06
[gui] Update available multiple msgs bugfix

surahman:
**_Issue #2188:_**
Introducing a check for signal connection to the GUI when an update is available.

We could also, potentially, do the following and move the `disconnect` from the `else` statement out to before the update check - which is what you suggested. It is the same code as the original, but just refactored:
```c++
void cmd::GuiCmd::update_about_menu()
{
    auto reply = version_future.result();

    about_client_version.setText("multipass version: " + QString::fromStdString(multipass::version_string));
    about_daemon_version.setText("multipassd version: " + QString::fromStdString(reply.version()));

    QObject::disconnect(&tray_icon, &QSystemTrayIcon::messageClicked, 0, 0);
    tray_icon_menu.removeAction(&update_action);

    if (update_available(reply.update_info()))
    {
        update_action.setIcon(QApplication::style()->standardIcon(QStyle::SP_MessageBoxInformation));

        QObject::connect(&tray_icon, &QSystemTrayIcon::messageClicked,
                         [this] { QDesktopServices::openUrl(QUrl(update_action.whatsThis())); });

        tray_icon.showMessage(QString::fromStdString(reply.update_info().title()),
                              QString("%1\n\nClick here for more information.")
                                  .arg(QString::fromStdString(reply.update_info().description())));
    }
}
```

My rationale for performing the check to see if there is a connected signal is because it might be more efficient than a potentially needless attempt to disconnect. I am open to suggestions and the above code is sitting on the `dev` branch waiting to be merged if required.

### DIFF
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -269,7 +269,7 @@ void cmd::GuiCmd::update_about_menu()
         tray_icon_menu.insertAction(about_menu.menuAction(), &update_action);
         tray_icon.showMessage(QString::fromStdString(reply.update_info().title()),
                               QString("%1\n\nClick here for more information.")
-                              .arg(QString::fromStdString(reply.update_info().description())));
+                                  .arg(QString::fromStdString(reply.update_info().description())));
     }
 }
 


ricab (changes requested):
Hi @surahman, unfortunately this doesn't work [because](https://doc.qt.io/qt-5/qobject.html#isSignalConnected):

> signal must be a signal member of this object, otherwise the behaviour is undefined.

Can't call it on `tray_icon` either because it's protected. But your alternative of always disconnecting should work. Want to commit it?

surahman:
Hi @ricab, sorry about the multiple commits, the refactored code has been pushed.

surahman:
Sorry, cleaned up the commit history.

ricab (approved):
Yep, that fixes it! Thanks a lot @surahman :+1: 

bors r+

surahman:
Thanks for reviewing @ricab!

PR #2205, opened 2021-08-19
[Proof-of-concept] Test daemon with mocked RPC

ricab:


ricab:
This needs some boiler plate for the promise-future dance, but we should be able to factor it out into some utility that tests can use.

PR #2210, opened 2021-08-21
[issue] A different default memory size

tomkwok:
The memory size of an instance is still non-configurable in multipass, and changing `multipassd-vm-instances.json` has no effect for me. As a current workaround I will need to change the memory value in `constants.h` and get a build with a different default memory size. I was not able to get Qt5 X11 properly configured when trying to build locally on macOS, so I am using the CI build system here to generate the binary. I don't know why multipass-ci that generates the binary for macOS cannot be installed in my fork, so I am now creating this pull request in order to trigger the build system that generates the build that I want.

### DIFF
--- a/include/multipass/constants.h
+++ b/include/multipass/constants.h
@@ -29,7 +29,7 @@ constexpr auto min_memory_size = "128M";
 constexpr auto min_disk_size = "512M";
 constexpr auto min_cpu_cores = "1";
 
-constexpr auto default_memory_size = "1G";
+constexpr auto default_memory_size = "512M";
 constexpr auto default_disk_size = "5G";
 constexpr auto default_cpu_cores = min_cpu_cores;
 constexpr auto default_timeout = std::chrono::seconds(300);


Saviq:
@tomkwok what do you mean "changing ...json" doesn't work for you? Also you say "different _default_ memory size", but that has nothing to do with the JSON file.

Why do you need a different _default_? You can always choose the amount at `multipass launch` time.

And yes, you won't be able to build a macOS package yourself, that part we had to keep proprietary.

I'll kick a CI run to generate the package for you, but that's not sustainable.

bors try

tomkwok:
@ricab @surahman maintainer please help approve running the workflow as a workaround solution for me.

Also would appreciate answers to: Why editing the json does not work? Why multipass-ci does not work on my fork?

tomkwok:
Thanks for your help. First thing I was not aware that I can change the memory size with `multipass launch -m`. Thanks for this information. However I want the memory size configuration to persist for a certain instance (say `primary`). This way when I launch the multipass GUI I would have the instance launching with my memory size configuration that is not 1G. What should I do to achieve this? Yes I understand changing the hard-coded value is not sustainable.

Saviq:
@tomkwok you can always `multipass launch -n primary -m 512M` manually. It's only when creating the instance is when the default is taken into account.

You'll need to fix this test for this to go through CI:

https://github.com/canonical/multipass/blob/6fecaaf6023622655d97a74f605b5caaf597302b/tests/test_daemon.cpp#L647

tomkwok:
Okay while I did not get a macos build here, my problem has been solved as I just realized I can `multipass launch -m 512M -n primary`. The memory config will persist for primary. I thought it wouldn't.

Saviq:
Cool!

PR #2214, opened 2021-08-26
Daemon testing with RPC mocking

ricab:
This continues with the approach in #2205 (replacing it) to enable `Daemon` testing by calling its public slots directly, providing a `MockServerWriter`. To allow that, slot signatures are changed to receive a `ServerWriterInterface` param in place of the concrete `ServerWriter` implementation.


townsend2010 (approved):
This is very nice!  Thanks!

bors merge

Saviq:
bors retry

PR #2218, opened 2021-08-27
Fix slow daemon tests on Windows

ricab:
This addresses the 1 second each daemon test has been taking on Windows. Basically that was how much time we were waiting in the check for an existing daemon. I'm not sure why we don't see the same on other platforms, but I suppose it fails immediately when using unix sockets.

BTW, GRPC itself already logs an error when trying to create the second server, mentioning:

> Only one usage of each socket address (protocol/network address/port) is normally permitted

I am not sure if that's a default message whenever it can't open the port, or if it somehow knows that it is being used for another server of its own. I looked for some sort of API to get that error info, but I didn't find any.

### DIFF
--- a/src/daemon/daemon_rpc.cpp
+++ b/src/daemon/daemon_rpc.cpp
@@ -38,7 +38,7 @@ void throw_if_server_exists(const std::string& address)
     auto stub = mp::Rpc::NewStub(channel);
 
     grpc::ClientContext context;
-    auto deadline = std::chrono::system_clock::now() + std::chrono::seconds(1);
+    auto deadline = std::chrono::system_clock::now() + std::chrono::milliseconds(100); // should be enough...
     context.set_deadline(deadline);
 
     mp::PingRequest request;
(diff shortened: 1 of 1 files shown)


src/daemon/daemon_rpc.cpp
```diff
@@ -32,28 +32,24 @@ namespace
 {
 constexpr auto category = "rpc";
 
-void throw_if_server_exists(const std::string& address)
+bool see_if_im_in_the_kitchen(const std::string& address)
```
townsend2010, line 35: Although I think the name of this function is funny, I think we should stick to something along the lines of `check_if_server_running()` or something.

townsend2010 (approved):
This is really good stuff.  I has been something vexing me for a looong time and I'm very glad you found the issue.

I've just one comment inline, but I'll definitely not block on it.

bors delegate

src/daemon/daemon_rpc.cpp
ricab: Of course, I will fix that  :sweat_smile:

ricab:
Thanks for the review @townsend2010! I'll drop the joke in the function name.

ricab:
bors r=townsend2010

townsend2010:
bors retry

PR #2221, opened 2021-08-30
[transfer] Improve error on arg w/ too many colons

### LINKED ISSUE #2220: Confusing error on transfer argument with wrong format
Originally reported on [discourse](https://discourse.ubuntu.com/t/working-with-multipass-instances/8422/4).

Multiple colons in transfer arguments result in a confusing error:
```
PS > $HOME
C:\Users\some_user
PS > multipass transfer -vvvv asdf beefy-constrictor:$HOME/asdf
An instance name is needed for either source or destination
```

ricab:
Fixes #2220.

### DIFF
--- a/src/platform/client/CMakeLists.txt
+++ b/src/platform/client/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright © 2018-2019 Canonical Ltd.
+# Copyright © 2018-2021 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -16,4 +16,6 @@ add_library(client_platform STATIC
   client_platform.cpp
   client_platform_linux.cpp)
 
-target_link_libraries(client_platform Qt5::Core)
+target_link_libraries(client_platform
+        fmt
+        Qt5::Core)
--- a/src/platform/client/client_platform.cpp
+++ b/src/platform/client/client_platform.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Canonical, Ltd.
+ * Copyright (C) 2018-2021 Canonical, Ltd.
  *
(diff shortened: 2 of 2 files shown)


Saviq:
Thing is... colons are valid filename characters, so rejecting them is probably not DTRT either.

ricab:
Yeah I considered that, but I believe it's mostly a separate issue. We don't accept file names with colons now either and this wasn't meant to deal with that. It just explains why the argument is being rejected.

I believe that supporting colons in filenames would require one of:
1. changing `transfer` in a backward incompatible way, so that `<instance>:<path>` was no longer valid;
2. extending the interface of `transfer` (e.g. adding dedicated options);
3. escaping colons.
4. supporting quotes
 
With the 2, 3, and 4, non-escaped colons in the current param would still be wrong. If we went for 1, that would indeed render this obsolete, but unless that's something we want to do right away, this fixes the error in the current scheme.

ricab:
I'll wait for further feedback before attempting to get this covered in tests.

townsend2010:
Ok, yeah, colons are allowed inside an instance for directory and file names.  I think I was too strict with this when I did this logic.  I think anything before the first colon should be treated as the instance name.  Anything after this colon is the path inside the instance regardless of potentially more colons.

ricab:
You're right, colons can't be used for instance names, so there's no ambiguity if we just do the split on the first one. In a case like the one originating this, we'd write a file `/home/ubuntu/C:\...`, but I guess that's alright.

townsend2010:
> In a case like the one originating this, we'd write a file `/home/ubuntu/C:\...`, but I guess that's alright.

Right, it will be up to the user to figure out if that is the path they want to use since `$HOME` would be expanded on the host.

ricab:
Yep, similar thing happens on Linux anyway (`/home/ubuntu/home/...`)

ricab:
Closing this now, since the feedback is that a different approach is preferable.

PR #2225, opened 2021-09-07
[qemu] Additional tweaks needed for QEMU platform support

townsend2010:


src/platform/backends/qemu/qemu_platform.h
```diff
@@ -42,7 +42,11 @@ class QemuPlatform
     virtual optional<IPAddress> get_ip_for(const std::string& hw_addr) = 0;
     virtual void remove_resources_for(const std::string&) = 0;
     virtual void platform_health_check() = 0;
-    virtual QStringList platform_args(const VirtualMachineDescription& vm_desc)
+    virtual QStringList base_platform_args()
+    {
+        return {};
+    };
+    virtual QStringList full_platform_args(const VirtualMachineDescription& vm_desc)
```
Saviq, line 49: As discussed, IMO we should be explicit about how the args returned are used. So have specific `vmstate_`, `start_` methods here, and anything common between them should be handled in the platform-specific bit.

townsend2010: I've renamed these `vmstate_platform_args()` and `vm_platform_args()`.

src/platform/backends/qemu/qemu_platform.h
```diff
@@ -42,7 +42,11 @@ class QemuPlatform
...
     virtual void platform_health_check() = 0;
-    virtual QStringList platform_args(const VirtualMachineDescription& vm_desc)
+    virtual QStringList vmstate_platform_args()
+    {
+        return {};
+    };
+    virtual QStringList vm_platform_args(const VirtualMachineDescription& vm_desc)
     {
         return {};
     };
```
Saviq, line 52: I wonder if these should be made pure virtual? To force descendant classes to return empty, but only if that's applicable?

townsend2010: Perhaps `vm_platform_args()` should be pure virtual since I would say any platform inheriting this will need to implement this, but I'm not so sure about `vmstate_platform_args()`.  If we really want this to be an interface definition, then *all* methods should be pure virtual.

Saviq: I'm just trying to avoid "forgetting" about those in descendants, is all. Your call :)

Saviq (approved):
+1 from me, regardless of virtuals or not.

bors d+

Saviq (approved):

townsend2010:
bors r=Saviq

Issue #2224, opened 2021-09-07
Wrong parameter parsing


luis4a0:
**Describe the bug**
Issuing `multipass launch -n asdf -mem 2048M` gives
```
launch failed: '2048M' is not a supported alias. Please use `multipass find` for supported image aliases.
```

**Expected behavior**
`-mem` should be recognized as a wrong parameter, _before_ checking the validity of `2048M` as an alias.

townsend2010:
Hey @luis4a0,

Thanks for entering this.  For the expected behavior, I'm not sure if this is what you meant, but `-mem` itself is itself a valid option (the `-m` part of it is the option), it's just that `em` is not a valid memory size and that is what needs to be validated.  Same goes for all other options.

For example, it's possible someone may enter `-cpus` and that would be parsed as `-c` with `pus` being the number of CPUs given at launch.

luis4a0:
The fact that `-mem` is a valid option escaped me. Thanks for pointing this out!

Now I see it's not a trivial fix. We need to parse in two stages (analogous to lex and yacc): first we generate tokens from all the options in the command line (that is, one for each unit separated by spaces) and then we parse the those tokens, combined in the order given. Maybe there is a simpler way to do it, I'll think about it.

townsend2010:
Hey @luis4a0, ok, cool.

It should actually be quite trivial to fix.  In `validate_create_arguments()`, instead of calling `validate_image()` early, call it after all of the other options have been validated. :wink:

ricab:
Perhaps this could be dealt with in the client: confirm it is a valid MemorySize before adding it to the request. That's already done for CPUs.

luis4a0:
Thanks for your insights @townsend2010 and @ricab! I implemented the latter, it proved to be easier. In fact, I didn't remove check code for the daemon because it is still needed to check that the requested size is bigger than the minimum.

Saviq:
> I didn't remove check code for the daemon

You also can't _trust_ the client to have done the validation :)

ricab:
> > I didn't remove check code for the daemon
> 
> You also can't _trust_ the client to have done the validation :)

Yep, the daemon of course needs to validate the contents of RPC requests (to be sure, I did not mean to suggest otherwise).

PR #2226, opened 2021-09-07
Parse memory, disk parameters in the client.

luis4a0:
Fixes https://github.com/canonical/multipass/issues/2224.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -281,15 +281,7 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     {
         auto arg_mem_size = parser->value(memOption).toStdString();
 
-        try
-        {
-            mp::MemorySize{arg_mem_size};
-        }
-        catch (mp::InvalidMemorySizeException&)
-        {
-            fmt::print(cerr, "Invalid memory size value supplied: {}.\n", arg_mem_size);
-            return ParseCode::CommandLineError;
-        }
+        mp::MemorySize{arg_mem_size}; // throw if bad
 
         request.set_mem_size(arg_mem_size);
     }
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -23,8 +23,10 @@
 #include <multipass/cli/client_platform.h>
 #include <multipass/constants.h>
 #include <multipass/exceptions/cmd_exceptions.h>
+#include <multipass/exceptions/invalid_memory_size_exception.h>
 #include <multipass/exceptions/snap_environment_exception.h>
 #include <multipass/format.h>
+#include <multipass/memory_size.h>
 #include <multipass/settings.h>
 #include <multipass/snap_utils.h>
 #include <multipass/utils.h>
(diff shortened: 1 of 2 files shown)


townsend2010:
I still think the logic in `validate_create_arguments()` should be changed such that `validate_image()` is called *after* the other options are validated.  As @Saviq said, the daemon can't trust that any client has validated the options first.

ricab:
To be clear, +1 to daemon validation. But we also have basic type verification in the client.  We need some in order to translate cmd line -> RPC request. Ideally, our RPC would have a `MemorySize` type to enforce it, but we go a lot beyond type checking for a number of other arguments (e.g. networks, timeouts, cpus...).

From the daemon's point of view, the RPC request is wrong in two ways: `2048M` image and `em` memory. Either of those is a good justification to reject the request. It's just that we'd like to detect bad cmd line constructions first, and that is a client job.

I'll wait for this matter to coalesce before going on with review.

luis4a0:
Hey @townsend2010, your argument is right. However, we already make checks in the client, for the number of CPU's and the parameter is not validated on the daemon.

In order to be consistent, we should also move that check to the daemon.

What do you think?

townsend2010:
@luis4a0,

Yeah, I think *all* parameters should be validated in the daemon with the alias/image validation being last to be validated to ensure the options are all good.

luis4a0:
@townsend2010 And do you think the checks (cpu's, memory and disk) should be removed from the client? They don't harm, but they can be seen as repeating things.

townsend2010:
@luis4a0,
> And do you think the checks (cpu's, memory and disk) should be removed from the client?

No, I think the client should validate as well.  But the daemon can't rely on what any particular client (there may be other clients out of our control) to validate and trust what it's given.  I know it seems a little redundant, but I think it's the right thing to do.

luis4a0:
Yes, it seems to be the right thing to do. Thanks @townsend2010!

@ricab So we are adding the checks in the client, as well as improving checking in the daemon. Are you OK with this?

ricab:
For the record, we're considering the approach in this PR complete (discussed elsewhere). There may be other improvements on the daemon side, but they appear unnecessary to deal with this issue and can be done separately. So I am queueing for review.

src/client/cli/cmd/launch.cpp
```diff
@@ -277,12 +279,36 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+
+        try
+        {
+            mp::MemorySize{arg_mem_size};
+        }
+        catch (mp::InvalidMemorySizeException&)
+        {
+            fmt::print(cerr, "Invalid memory size value supplied: {}.\n", arg_mem_size);
+            return ParseCode::CommandLineError;
+        }
```
ricab, line 292: Hmm, the exception already includes ~same info, and it is handled "above" if you just throw. So, this would be enough:

```suggestion
        mp::MemorySize{arg_mem_size}; // throw if bad
```

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e17-dont-try-to-catch-every-exception-in-every-function and https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e18-minimize-the-use-of-explicit-trycatch .

This yields:

```
[2021-09-10T11:53:57.777] [error] [client] Caught an unhandled exception: asdf is not a valid memory size
```

As opposed to:

```
Invalid disk size value supplied: asdf.
```

I think the simplicity is preferable here, but if you prefer the tailored message, I am also OK with that. Up to you.

ricab (commented):
Yeah, this is what we need. Just a couple of comments.

luis4a0:
Hey @ricab! Agreed, this PR is the fix we intended.

luis4a0:
Oops, sorry, I didn't see your latest messages :(

src/client/cli/cmd/launch.cpp
```diff
@@ -277,12 +279,28 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+
+        try
+        {
+            mp::MemorySize{arg_disk_size};
+        }
+        catch (mp::InvalidMemorySizeException&)
+        {
+            fmt::print(cerr, "Invalid disk size value supplied: {}.\n", arg_disk_size);
+            return ParseCode::CommandLineError;
+        }
```
ricab, line 301: This is the same, I think we'd better take the same approach in both :)

ricab (commented):
Hey @luis4a0, I didn't point it out specifically, but I meant both places that used the same code. (so either one approach or the other in both).

src/client/cli/cmd/launch.cpp
```diff
@@ -23,8 +23,10 @@
 #include <multipass/cli/client_platform.h>
 #include <multipass/constants.h>
 #include <multipass/exceptions/cmd_exceptions.h>
+#include <multipass/exceptions/invalid_memory_size_exception.h>
```
ricab, line 26: Final tiny detail... this can be dropped now

```suggestion
```

ricab (approved):
Perfect, thanks!

bors r+

ricab:
bors retry

Saviq:
bors retry

Saviq:
Will you?

bors retry

PR #2227, opened 2021-09-08
[daemon] Handle exceptions starting VMs in ctor

### LINKED ISSUE #2197: Daemon bails on start when starting an instance fails
When the daemon starts and there are instances to start, if the instance fails to start, Qt will complain about an exception being thrown in an even handler, and then the daemon complains about an unhandled exception and then bails.  This leaves the daemon in an unstartable state.

ricab:
Since we can't throw anything out of `QTimer` (qt slot), we need to catch all exceptions types. Fixes #2197

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -35,6 +35,7 @@
 #include <multipass/query.h>
 #include <multipass/settings.h>
 #include <multipass/ssh/ssh_session.h>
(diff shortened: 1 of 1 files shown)


Saviq (approved):
Yup.

bors merge

ricab:
Wow that was the fastest approval ever!

Issue #2229, opened 2021-09-08
Improve argument checking in daemon


luis4a0:
#2226 fixes parameter parsing (reported in #2224), but it does by checking parameters in the CLI client. This means that, if another client sends the daemon the same wrong parameters (`-mem 2048M`, in this case), the daemon will fail to detect the error.

While it is ok to check arguments in the client, we need to enforce checking this in the daemon. That ca be accomplished by moving instance validation _after_ the validation of the other parameters.

luis4a0:
A more serious work is needed here, as discussed in #2230.

tobe2098:
Issue is still relevant within its scope of considering alternative clients connecting to the daemon.

PR #2230, opened 2021-09-08
[daemon] Improve parsing of memory/disk size.

luis4a0:
Addresses #2229.


Saviq (commented):
As discussed, this doesn't change anything when assuming `multipass` as our client, and that's the only supported one at this point.

We agree that we could do better (report "one or more" errors to the client, validating in rising complexity order and having cut-offs at significant complexity jumps).

luis4a0:
Closing this, then.

PR #2239, opened 2021-09-15
[no-merge] Instance mod baseline prototype

ricab:
A hacked prototype of instance mod following a delete + recreate
baseline. Modifies an instance named 'asdf' to have 3 cores when
`multipass version` is called.


ricab:
OK this destroy + create approach (same as when daemon restarts) is not enough for instance modification in "managed" backends (e.g. libvirt, vbox, hyper-v), because:

- vm definitions are left behind on destruction
- @Saviq points out customization would be lost (including snapshots)

We'll need to augment our backend interfaces.

PR #2243, opened 2021-09-17
[build] Add definitions for backends and arch

townsend2010:


Saviq (approved):
Yup, I don't love that we need this, but OK.

bors merge

PR #2272, opened 2021-10-05
 [snap] Use Release cmake build type

townsend2010:
This will build the Multipass binaries without debug symbols and cut the Snap down ~240MB in size.

### DIFF
--- a/.github/workflows/linux-Coverage.patch
+++ b/.github/workflows/linux-Coverage.patch
@@ -12,7 +12,7 @@
      - dnsmasq-utils
      source: .
      cmake-parameters:
--    - -DCMAKE_BUILD_TYPE=RelWithDebInfo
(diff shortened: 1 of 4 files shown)


townsend2010:
Hmm, not sure why the `BuildAndTest (RelWithDebInfo)` check is showing up and is required.

townsend2010:
> Because ~~it's~~ it was required in branch protection rules:

Ah, ok, I wasn't aware of that.

At any rate, here is the lowdown on the change.

Size of snap in `edge`: 370MB
Size of snap in this PR: 134MB

I think that is a pretty big win :smile:

Saviq (approved):
Yeah, let's.

bors merge

Saviq:
bors retry

PR #2275, opened 2021-10-06
[workflows] drop unpacked test workflows

Saviq:
All testing happens on the `.zip` file, this just makes it confusing.`


townsend2010 (commented):
I don't believe we should remove these as this makes it apparent what changes have occurred wrt to these files via version control.  The zip file should be created from all of these files.

PR #2277, opened 2021-10-07
Disabled copy move

ricab:
Avoid having to declare copy/move ctors/assignement operators each time. That's error prone, not only because the signatures need to be correct in each occurrence, but also because there is a tendency to explicitly delete move operations as well, which I learned recently is apparently [not advisable](https://stackoverflow.com/a/38820178). 

`DisabledCopyMove` serves the same purpose as [`boost::noncopyable`](https://www.boost.org/doc/libs/1_77_0/libs/core/doc/html/core/noncopyable.html), and is used the same way. Just inherit privately:

```C++
class Foo : private DisabledCopyMove {};
```

### DIFF
--- a/include/multipass/cert_store.h
+++ b/include/multipass/cert_store.h
@@ -19,6 +19,7 @@
 #define MULTIPASS_CERT_STORE_H
 
 #include "disabled_copy_move.h"
+
 #include <string>
 
 namespace multipass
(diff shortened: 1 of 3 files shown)


ricab:
And fixed the author to pass the CLA check.

PR #2282, opened 2021-10-12
[tests] move to upstream googletest

Saviq:
Relying on gRPC's submodule does not give us the control over googletest
versions we'd like.

Let's "[Live at head](https://github.com/google/googletest#live-at-head)" instead, using CMake's FetchContent.

Based on [GTests's Quickstart: Building with CMake](https://google.github.io/googletest/quickstart-cmake.html).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -17,8 +17,17 @@ set (CMAKE_AUTOMOC ON)
 
 include(c_mock_defines.cmake)
 
-set(MULTIPASS_GMOCK_DIR ${CMAKE_SOURCE_DIR}/3rd-party/grpc/third_party/googletest/googlemock)
-set(MULTIPASS_GTEST_DIR ${CMAKE_SOURCE_DIR}/3rd-party/grpc/third_party/googletest/googletest)
+include(FetchContent)
+set(FETCHCONTENT_QUIET FALSE)
+
+FetchContent_Declare(googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG main
+  GIT_SHALLOW TRUE
+  GIT_PROGRESS TRUE
+)
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
 
 add_executable(multipass_tests
   common.cpp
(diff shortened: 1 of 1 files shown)


Saviq:
> * having to react to any bugs or backward incompatibility problems as they arise (rather than gating manually and updating when we want);

You'd hope that if they recommend "Living at head", there'd be little of that. We can still pin it through `GIT_TAG` if necessary. But at least we'll know there's a problem.

> * possible difficulties in IDEs (should be fine in IDEs that know CMake)

Since the checkout path is static, and done at configure time, that should work fine.

> * need network to build

Not different to submodules, and you can set [`FETCHCONTENT_FULLY_DISCONNECTED`](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare) if you are offline.

> * one more fancy step for CI to possibly break on

No different to submodules IMO.

> And a question: how would this play with ccache? Would unchanged google tests sources still result in cache hits?

Yes, should be fine - it's not any different to a submodule this way - a submodule is just a checkout, so is this.

> I don't know, updating would be good, but I am a little hesitant to jump to the crest of the wave. I guess we could try it and see how it goes (with the fallback of pinning if necessary). But this affects development and I would prefer to see what others think about it. WDYT @luis4a0 and @townsend2010?

To get at newer googletest, we'd need to, as you say, add another submodule. The advantage of this, IMO, is that you _can_ unpin it. You can't, a submodule. And a tarball actually may make sense.

ricab:
> Not different to submodules, and you can set FETCHCONTENT_FULLY_DISCONNECTED if you are offline.

Ah good to know there is the option (I did mean build and not checkout). 

> Yes, should be fine - it's not any different to a submodule this way - a submodule is just a checkout, so is this.

I mean, it will check out new stuff to the build directory more often (when there is an update?). I guess it's fine if files retain their modification dates...

> To get at newer googletest, we'd need to, as you say, add another submodule. The advantage of this, IMO, is that you can unpin it. You can't, a submodule. And a tarball actually may make sense.

OK, I guess I am fine with it but would just let it simmer a bit waiting for others if tht's ok.

luis4a0:
Hey @ricab!

I don't see big drawbacks, but a gain in features. And the time we may spend in changes must be spent at some point anyway.

So I'm +1 for switching to the cutting edge.

ricab:
> I just meant that without the …DISCONNECTED… var set to ON, it failed on reconfiguring. But built fine with it.

I see, it needs to reach out to reconfigure otherwise.

ricab:
I just needed this in two different ways again. Since there were no objections, I would merge. But @Saviq, did you get the chance to chase those CI issues?

Saviq:
> @Saviq, did you get the chance to chase those CI issues?

Just looked and somehow Windows fails to find the gtest headers, even though it has them in `-I` :/

Need a deeper look.

ricab (approved):
Alright, LGTM. Thanks again!

bors r+

Saviq:
bors retry

Saviq:
bors retry

Saviq:
bors retry

ricab:
bors retry

PR #2285, opened 2021-10-15
Test settings

ricab:
Finally get our `Settings` tested. At least the parts we know will have to remain (even if moved to handlers).

Depends on #2277.

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -55,6 +55,7 @@ public:
     virtual QString get_workflows_url_override() const;
     virtual bool is_alias_supported(const std::string& alias, const std::string& remote) const;
     virtual bool is_remote_supported(const std::string& remote) const;
+    virtual bool is_backend_supported(const QString& backend) const; // temporary (?)
     virtual int chown(const char* path, unsigned int uid, unsigned int gid) const;
     virtual bool link(const char* target, const char* link) const;
     virtual bool symlink(const char* target, const char* link, bool is_dir) const;
@@ -79,7 +80,6 @@ QString default_privileged_mounts();
 
 QString daemon_config_home(); // temporary
 
-bool is_backend_supported(const QString& backend); // temporary
 VirtualMachineFactory::UPtr vm_backend(const Path& data_dir);
 logging::Logger::UPtr make_logger(logging::Level level);
 UpdatePrompt::UPtr make_update_prompt();
--- a/include/multipass/settings.h
+++ b/include/multipass/settings.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2020 Canonical, Ltd.
+ * Copyright (C) 2019-2021 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/include/multipass/settings.h
+++ b/include/multipass/settings.h
@@ -41,6 +41,20 @@ public:
     virtual QString get(const QString& key) const;            // throws on unknown key
     virtual void set(const QString& key, const QString& val); // throws on unknown key or bad settings
 
+    /**
+     * Obtain a setting as a certain type
+     * @tparam T The type to obtain the setting as
+     * @param key The Key that identifies the setting
+     * @return The value that the setting converts to, when interpreted as type @p T. Follows the conversion rules used
+     * by @c QVariant.
+     * If the current @em value cannot be converted, a default @p T value is returned.
+     * @note QVariant's conversion rules may be surprising and not always correctly documented. For instance according
+     * to Qt's documentation at the time of writing, @c "False" would convert to @c bool as @c true, but it actually
+     * converts to @c false. However, @c "no" converts to @c true.
+     * @throws UnsupportedSettingValueType@<T@> If QVariant cannot convert from a @c QString to type @p T. Note that
+     * this is only thrown when the type itself can't be converted to. The actual value may still fail to convert, in
+     * which case a default @c T value is returned.
+     */
     template <typename T>
     T get_as(const QString& key) const;
 
@@ -62,7 +76,7 @@ template <typename T>
 T multipass::Settings::get_as(const QString& key) const
 {
     auto var = QVariant{get(key)};
-    if (var.canConvert<T>())
+    if (var.canConvert<T>()) // Note: this only checks if the types themselves are convertible, not the values
         return var.value<T>();
     throw UnsupportedSettingValueType<T>(key);
 }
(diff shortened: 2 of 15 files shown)


src/utils/qsettings_wrapper.h
```diff
@@ -0,0 +1,104 @@
...
+
+    virtual void setValue(const QString& key, const QVariant& value)
+    {
+        assert(qsettings);
+        qsettings->setValue(key, value);
+    }
+
+    QVariant value(const QString& key, const QVariant& default_value = QVariant()) const
+    {
+        return value_impl(key, default_value); // indirection avoids default args in virtual method
```
Saviq, line 66: Got a reference as to why you're avoiding that?

tests/test_settings.cpp
```diff
@@ -0,0 +1,428 @@
...
+public:
+    mpt::MockFileOps::GuardedMock mock_file_ops_injection = mpt::MockFileOps::inject<NiceMock>();
+    mpt::MockFileOps* mock_file_ops = mock_file_ops_injection.first;
+    MockQSettingsProvider::GuardedMock mock_qsettings_injection = MockQSettingsProvider::inject<StrictMock>(); /* this
+    is made strict to ensure that, other than explicitly injected, no QSettings are used; that's particularly important
+    when injecting real get and set behavior (don't want tests to be affected, nor themselves affect, disk state) */
+    MockQSettingsProvider* mock_qsettings_provider = mock_qsettings_injection.first;
+    std::unique_ptr<NiceMock<MockQSettings>> mock_qsettings = std::make_unique<NiceMock<MockQSettings>>();
+    mpt::MockSettings& mock_settings = mpt::MockSettings::mock_instance();
+    FILE fake_file;
```
Saviq, line 99: Add a couple newlines between the injections so it reads easier?

tests/test_settings.cpp
```diff
@@ -0,0 +1,428 @@
...
+    const auto key = "whatever";
+    for (const auto& [val, reprs] : setting_val_reprs<TypeParam>())
+    {
+        for (const auto& repr : reprs)
+        {
+            EXPECT_CALL(mpt::MockSettings::mock_instance(), get(Eq(key))).WillOnce(Return(repr));
+            EXPECT_EQ(MP_SETTINGS.get_as<TypeParam>(key), val);
+        }
+    }
+}
```
Saviq, line 393: Funky!

Saviq (changes requested):
Good work :)

Some minor comments inline.

include/multipass/platform.h
```diff
@@ -55,6 +55,7 @@ class Platform : public Singleton<Platform>
     virtual QString get_workflows_url_override() const;
     virtual bool is_alias_supported(const std::string& alias, const std::string& remote) const;
     virtual bool is_remote_supported(const std::string& remote) const;
+    virtual bool is_backend_supported(const QString& backend) const; // temporary (?)
```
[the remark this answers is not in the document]
ricab: I know I had added that, but I wasn't sure if it wouldn't remain useful after all. I am not sure if we'll ever want the daemon to change drivers "live". Until then, it will only really try to set the driver after a restart, but should still refuse to restart if invalid, so needs a dedicated check.

I didn't dig into the code much though, so just added the interrogation mark. But I can remove if you want?

include/multipass/settings.h
```diff
@@ -41,6 +41,20 @@ class Settings : public Singleton<Settings>
     virtual QString get(const QString& key) const;            // throws on unknown key
     virtual void set(const QString& key, const QString& val); // throws on unknown key or bad settings
 
+    /**
+     * Obtain a setting as a certain type
+     * @tparam T The type to obtain the setting as
+     * @param key The Key that identifies the setting
```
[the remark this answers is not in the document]
ricab: Oversight, will fix.

include/multipass/settings.h
```diff
@@ -41,6 +41,20 @@ class Settings : public Singleton<Settings>
...
+     * @return The value that the setting converts to, when interpreted as type @p T. Follows the conversion rules used
+     * by @c QVariant.
+     * If the current @em value cannot be converted, a default @p T value is returned.
+     * @note QVariant's conversion rules may be surprising and not always correctly documented. For instance according
+     * to Qt's documentation at the time of writing, @c "False" would convert to @c bool as @c true, but it actually
+     * converts to @c false. However, @c "no" converts to @c true.
```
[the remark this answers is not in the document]
ricab: Actually, I misread the docs. Here's the doc I was referring to: https://doc.qt.io/qt-5/qvariant.html#toBool

> Returns true if the variant has userType() QMetaType::Bool, QMetaType::QChar, QMetaType::Double, QMetaType::Int, QMetaType::LongLong, QMetaType::UInt, or QMetaType::ULongLong and the value is non-zero, or if the variant has type QMetaType::QString or QMetaType::QByteArray and its lower-case content is not one of the following: empty, "0" or "false"; otherwise returns false.

I missed "lower-case" in there. I'll remove the note.

tests/test_settings.cpp
ricab: Heh :)

Saviq (approved):
Yup, thanks a lot!

bors merge

Saviq:
bors retry

Saviq:
Hmm `multipass list` on LXD [resulted in](https://github.com/canonical/multipass/runs/4105139056?check_suite_focus=true#step:4:325):

> list failed: Unable to read settings: access error (consider running with an administrative role)

ricab:
`multipass list` reads settings to find out the primary name and place that first in the output. Removing the client's settings file reproduced the same problem as in the integration test, but only when in a snap (:man_shrugging:). 

I didn't quite get why that was happening with `fopen`, but I did miss `fclose`. Not sure if that could be affecting things somehow (e.g. causing `errno != ENOENT`). I didn't want to check for `EACCES` specifically because I remember this being fragile stuff that could yield different codes in different cases (other platforms, unreadable parent dir...).

I reverted back to using streams to [detect existing but unreadable files](https://github.com/canonical/multipass/blob/d44fcd1c45be733fc4868005c1125d7a2f221f09/src/utils/settings.cpp#L97). Now using a mockable `FileOps::open`, with `fstream` instead of `ifstream` since it can be used with any open mode (templating on stream type would be unhelpful for mocking).

`FileOps::open` is overloaded, so I had to specify matchers in a few unrelated `EXPECT_CALL`s.

Saviq:
bors try

Saviq (approved):
ACK

bors merge

(I know you'll fail due to conflicts, but at least you'll run Linux CI)

PR #2297, opened 2021-10-19
[alias] sanitize alias names

Saviq:


Saviq:
> I am afraid `baseName` is dropping the extension:

D'oh!

Fixed.

PR #2320, opened 2021-11-01
Compile with g++ 11

luis4a0:
Impish includes gcc 11, so we need to avoid warnings and errors generated by it.


src/cert/ssl_cert_provider.cpp
```diff
@@ -109,7 +109,14 @@ long random_long()
 {
     long out{0};
     std::array<uint8_t, 4> bytes;
+
+    // The array must be used uninitialized. Otherwise, there will be no randomness.
+#ifndef __clang__
```
ricab, line 114: This needs to account for MSVC and AppleClang, right? I believe we can do that by checking `__GNUC__` too. 

It seems that [other compilers define that too](https://stackoverflow.com/questions/28166565/detect-gcc-as-opposed-to-msvc-clang-with-macro#comment96299743_43205345). So ideally we'd set `-DMULTIPASS_COMPILER_GCC` in `CMakeLists.txt` and then check that here (similar to what we do for `MULTIPASS_PLATFORM_LINUX/WINDOWS/APPLE`). But if you want to check just `(defined(__GNUC__) && !defined(__clang__))`, I guess that'd be OK too.

ricab (commented):
Thanks @luis4a0 , the approach makes sense to me, but I think we need to account for other compilers (CI broke for other reasons, so not sure exactly how they react to those pragmas).

src/cert/ssl_cert_provider.cpp
luis4a0: Hey @ricab, thanks for reviewing this!

Yes, the issue here is that it must _only_ work with gcc and, as you mention, other compilers do emulate the `#pragma GCC`. The point is that the block `#ifndef __clang__`...`#endif` also contains code :man_facepalming:. I'd fix that first. Let me give it a try.

luis4a0: Done, let's see now...

ricab: Oh indeed, I missed that entirely.

CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
 string(TOLOWER "${CMAKE_BUILD_TYPE}" cmake_build_type_lower)
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  add_definitions(-DMULTIPASS_COMPILER_CLANG)
```
ricab, line 37: What about AppleClang?

Also, I would rather see this closer to where the `MULTIPASS_PLATFORM_...` vars are defined, to be sure they are coherent. So in this [if-else](https://github.com/canonical/multipass/blob/a11c10192a0192ec998431672a2975619addcf5b/CMakeLists.txt#L156). WDYT?

ricab (commented):
Thanks for following the idea of the dedicated vars.

CMakeLists.txt
luis4a0: Well, both `MULTIPASS_COMPILER_CLANG` and `MULTIPASS_PLATFORM_MACOS` defined will be equivalent to `MULTIPASS_COMPILER_APPLE_CLANG`. I my point of view, I'll add that flag if Clang and Apple Clang differ substantially, making important to have such flag. Do you think it's worth the effort of adding such a flag?

OTOH, answering to your question, maybe detecting compiler as part of the platform would simplify a bit the code. Let me try to do it.

ricab: `MULTIPASS_COMPILER_CLANG` + `MULTIPASS_PLATFORM_MACOS` is fine by me, but I am not sure that's what would be defined with that code? IOW, does `CMAKE_CXX_COMPILER_ID STREQUAL "Clang"` with AppleClang? Otherwise we'd say MSVC

luis4a0: Oh, good question. The answer is no, see [here](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html).

luis4a0: Fixed.

luis4a0: Fixed

ricab:
Hey @luis4a0, I am still getting this on Impish:

```
[ 94%] Building CXX object tests/CMakeFiles/multipass_tests.dir/__/3rd-party/grpc/third_party/googletest/googletest/src/gtest-all.cc.o
In file included from /home/ubuntu/multipass/3rd-party/grpc/third_party/googletest/googletest/src/gtest-all.cc:42:
/home/ubuntu/multipass/3rd-party/grpc/third_party/googletest/googletest/src/gtest-death-test.cc: In function ‘bool testing::internal::StackGrowsDown()’:
/home/ubuntu/multipass/3rd-party/grpc/third_party/googletest/googletest/src/gtest-death-test.cc:1301:24: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
 1301 |   StackLowerThanAddress(&dummy, &result);
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
/home/ubuntu/multipass/3rd-party/grpc/third_party/googletest/googletest/src/gtest-death-test.cc:1290:13: note: by argument 1 of type ‘const void*’ to ‘void testing::internal::StackLowerThanAddress(const void*, bool*)’ declared here
 1290 | static void StackLowerThanAddress(const void* ptr, bool* result) {
      |             ^~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/multipass/3rd-party/grpc/third_party/googletest/googletest/src/gtest-death-test.cc:1299:7: note: ‘dummy’ declared here
 1299 |   int dummy;
      |       ^~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [tests/CMakeFiles/multipass_tests.dir/build.make:966: tests/CMakeFiles/multipass_tests.dir/__/3rd-party/grpc/third_party/googletest/googletest/src/gtest-all.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:10311: tests/CMakeFiles/multipass_tests.dir/all] Error 2
make: *** [Makefile:182: all] Error 2
```

Meanwhile, CI did not appear to trigger Windows/Mac builds properly.

Saviq:
> Meanwhile, CI did not appear to trigger Windows/Mac builds properly.

It's because there's a conflict:

> Auto-merging CMakeLists.txt
> CONFLICT (content): Merge conflict in CMakeLists.txt
> Automatic merge failed; fix conflicts and then commit the result.

luis4a0:
> Hey @luis4a0, I am still getting this on Impish:

Hey @ricab, do you think we should correct compilation errors on third-party stuff? Maybe they'll fix them shortly. Maybe @Saviq can also answer this :)

ricab:
> Hey @ricab, do you think we should correct compilation errors on third-party stuff? Maybe they'll fix them shortly. Maybe @Saviq can also answer this :)

Not fix the errors, but maybe add `-Wno-error` (if it remains).  However, since we're getting a newer gtest with #2282, maybe we can wait for that one. I don't even mind letting this in as is, but we need to be aware that building with g++11 will still not work.

luis4a0:
I vote for waiting #2282. Unless it takes much time to land.

Saviq:
> > Hey @luis4a0, I am still getting this on Impish:
> 
> Hey @ricab, do you think we should correct compilation errors on third-party stuff? Maybe they'll fix them shortly. Maybe @Saviq can also answer this :)

Is it only googletest? Would be good to try with #2282 then. Otherwise, the overhead of fixing 3rd parties (that we don't _already_ carry patches for) is probably not worth it. Working around with flags would be my preference there, if possible.

ricab:
> Is it only googletest? 

It was the only one I observed, since the build stops.
Hey @ricab, do you think we should correct compilation errors on third-party stuff?

To be clear, I never meant to suggest touching 3rd party. I meant simply that this PR is titled `compile with g++11`, but it doesn't. 

Normally, the way to fix that would be to add the appropriate compile options in *our* CMakeLists.txt. But since we're moving gtest, that may well fix it for us.

ricab:
(sorry, closed by mistake)

luis4a0:
> It's because there's a conflict:

Fixed by creating a private PR.

luis4a0:
> Working around with flags would be my preference there, if possible.

I did in `CMakeLists.txt`. This finally fixes all the compilation issues.

CMakeLists.txt
```diff
@@ -67,6 +69,15 @@ find_package(OpenSSL REQUIRED)
 # Needs to be here before we set further compilation options
 add_subdirectory(3rd-party)
 
+# TODO: remove this when gRPC fixes the compilation with g++ 11.
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  string(REPLACE "." ";" MULTIPASS_GCC_VERSION ${CMAKE_CXX_COMPILER_VERSION})
+  list(GET MULTIPASS_GCC_VERSION 0 MULTIPASS_GCC_VERSION_MAJOR)
```
ricab, line 75: Clever, but why do you check the version? GCC supports `maybe-uninitialized` since at least version 4, so the following should work:

```
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  target_compile_options(gRPC INTERFACE "-Wno-error=maybe-uninitialized")
endif()
```

And I think this should be in the 3rd-party CMakeLists.txt instead. There is a target_compile_options for it there already.

luis4a0: Yeah, but it's weird it doesn't complain in our code in versions < 11. Maybe the `-Wmaybe-uninitialized` parameter is disabled by some other argument in older versions. Anyway, it doesn't harm to add it for all versions, and it simplifies the code. Changed.

ricab: Right, it makes an effort to detect such cases, but it offers no guarantees. I suppose they just detect more cases in version 11.

ricab (commented):
Alright, LGTM. Thank you!

bors r+

ricab (approved):

ricab:
(Bors will fail because of private conflicts, but we can then merge with staging if everything is otherwise alright)

ricab:
bors retry

src/cert/ssl_cert_provider.cpp
```diff
@@ -109,8 +109,19 @@ long random_long()
 {
     long out{0};
     std::array<uint8_t, 4> bytes;
+
+#ifdef MULTIPASS_COMPILER_GCC
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+
+    // The array must be used uninitialized. Otherwise, there will be no randomness.
```
ricab, line 118: I'm afraid this does not sound right. `RAND_bytes` is what generates the randomness. Relying on whatever was on memory would be undefined behavior, a poor randomness source, and a security hazard. The array is left uninitialized because it is written to before it is ever read from. The goal is just performance (avoid writing zeros just before overwriting them). 

Could you please remove?

ricab (changes requested):
I was just about to merge when I noticed something I hadn't seen before (I believe you added it in a rebase?)

ricab:
bors cancel

src/cert/ssl_cert_provider.cpp
luis4a0: Oops, right, the comment was entirely wrong. Removed.

ricab (approved):
:+1: 

bors r+

PR #2344, opened 2021-11-15
[backend/utils] Fix detecting KVM due to Snap confinement

### LINKED ISSUE #2336: Virtualisation check is not being run correctly.  launch failed: The check_kvm_support script failed to start. Ensure it is in multipassd's PATH.
***Description of the bug***
On systems with virtualisation disabled, the check_kvm script either does not run when launching a VM or the return code is misinterpreted. I would expect the check_kvm script to tell the user that the system needs virtualisation enabled.

***To Reproduce***
I tried this on both an on-prem box and AWS cloud instance running Ubuntu 20.04
I am running as root...

```plain
$ multipass launch -n test1
launch failed: The check_kvm_support script failed to start. Ensure it is in multipassd's PATH.
```

Adding the snap's bin folder to the path made no difference.
```plain
$ cat /etc/environment
PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/multipass/current/bin"
```

Adding the snap path to the systemd service made no difference.
```plain
$ cat /etc/systemd/system/multipath-tools.service
[Unit]
Description=Device-Mapper Multipath Device Controller
<snipped>

[Service]
<snipped>
Environment=PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/multipass/current/bin"
```

Running the command manually displays the correct output.
```plain
$ /sbin/check_kvm_support
KVM (vmx) is disabled by your BIOS.
Enter your BIOS setup and enable Virtualization Technology (VT).
```

Which matches the output from
```
$ kvm-ok
> INFO: /dev/kvm does not exist
> HINT: sudo modprobe kvm_intel
> INFO: Your CPU supports KVM extensions
> INFO: KVM (vmx) is disabled by your BIOS
> HINT: Enter your BIOS setup and enable Virtualization Technology (VT),
> and then hard poweroff/poweron your system
> KVM acceleration can NOT be used
```

Enabling virtualisation fixed the issue.
```plain
$ multipass launch -n test1
Launched: test1
$ kvm-ok
INFO: /dev/kvm exists
KVM acceleration can be used
```

***Expected behavior***
I would expect the VM to fail to launch with the error output from check_kvm_support when virtualisation is disabled. The error message provided is confusing.

***Logs***
Unfortunately the system was rebooted to enable virtualisation and I do not have logs from the time of the issue.

***Additional info***
```plain
$ uname -a
Linux  5.4.0-90-generic #101-Ubuntu SMP Fri Oct 15 20:00:55 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

$ multipass version
multipass   1.8.0
multipassd  1.8.0

$ multipass info --all
Name:           test1
State:          Running
IPv4:           xxx.xxx.xxx.xxx
Release:        Ubuntu 20.04.3 LTS
Image hash:     939be728cbc7 (Ubuntu 20.04 LTS)
Load:           0.12 0.07 0.09
Disk usage:     1.3G out of 4.7G
Memory usage:   130.5M out of 981.2M
Mounts:         --
```

townsend2010:
Fixes #2336


src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -260,24 +260,21 @@ std::string mp::Backend::get_subnet(const mp::Path& network_dir, const QString&
...
+    }
 
-        if (process_state.exit_code == 1)
-        {
-            throw std::runtime_error(check_kvm->read_all_standard_output().trimmed().toStdString());
-        }
+    if (!MP_FILEOPS.open(kvm_device, QIODevice::ReadWrite))
+    {
+        throw std::runtime_error("The KVM device cannot be opened for reading and writing.\n"
+                                 "Please ensure the Snap KVM interface is connected for Multipass.");
```
Saviq, line 277: This could use `snap connect multipass:kvm` spelled out.

Saviq (changes requested):

src/platform/backends/shared/linux/backend_utils.cpp
townsend2010: Well, it's an auto-connection, so the user would have most likely disconnected the connection themselves, so I would say they would know how to connect it.  I wanted to keep the verbosity down on the message, so I felt it unnecessary to explicitly spell it out for the aforementioned reason.

townsend2010: I'll add, if you can think of a reason where the connection is not connected other than user intervention, then I'm definitely open to spelling it out.

Saviq: It's only auto-connected when coming from our store. Say it comes through a brand store, it might lose that assertion.

Either way, it's not like we need to strive for verbosity here?

townsend2010: Ok, I'll add it.  By verbosity, I meant in line for consistency with the messages above where we don't spell out stuff on enabling KVM.

townsend2010: Ok, done.  Let me know what you think of the wording.

Saviq (approved):
Yups!

bors merge

Saviq:
bors retry

Saviq:
Bors please…

bors merge

carlospintojunior:
For information....
I reproduced this error with virtualization turned off.

> user@pc:~$ egrep -c '(vmx|svm)' /proc/cpuinfo
0

Turning on virtualization, everything worked normally.

PR #2345, opened 2021-11-16
[set] add interactive mode

Saviq:


include/multipass/exceptions/cmd_exceptions.h
```diff
@@ -28,6 +28,12 @@ class ValidationException : public std::runtime_error
     using std::runtime_error::runtime_error;
 };
 
+class ValueException : public std::runtime_error
```
ricab, line 31: Could this be made more specific to the prompter, like `PromptException`? That would clarify the distinction between troubles reading vs interpreting setting values. And it is not conceptually tied to a command: the `PlainPrompter` could in principle be used outside of CLI commands.

src/client/cli/cmd/set.cpp
```diff
@@ -124,16 +127,30 @@ mp::ParseCode cmd::Set::parse_args(mp::ArgParser* parser)
...
+                key = keyval.at(0);
+                try
+                {
+                    val = QString::fromStdString(prompter.prompt(key.toStdString()));
+                }
+                catch (const mp::ValueException& e)
+                {
+                    cerr << e.what() << std::endl;
+                    status = ParseCode::CommandLineError;
+                }
```
ricab, line 152: Extract into a `checked_prompt` or something? `status = checked_prompt(prompt, key);`

ricab (commented):
Nice, and should be easy to compose into `Settings`. I had a quick look and have only a couple of  suggestions.

src/client/common/prompters.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+
+std::string PlainPrompter::prompt(const std::string& text)
+{
+    term->cout() << text << ": ";
+
+    std::string value;
+    std::getline(term->cin(), value);
+
+    if (term->cin().eof())
```
ricab, line 33: Oh one more thing, I think this should check for `fail()` too/instead. The [table at the bottom here](https://en.cppreference.com/w/cpp/io/basic_ios/fail) is helpful. I am not sure if we want `eof()` but `!fail()` to be a valid way to express the empty string (e.g. when redirecting stdin).

include/multipass/exceptions/cmd_exceptions.h
Saviq: FWIW I didn't want to introduce another file (we have too many separate exception headers already…). Maybe it could be renamed to `cli_exceptions.h` - CLI does not necessarily mean just our `multipass` command, but generally the Command Line Interface - to which a Prompter, by design, is tied. But then `ValidationException` probably doesn't fit.

And then exception type proliferation… I feel like we could have some "generic" enough exceptions that apply in multiple places. Nothing in a Prompter could throw a Settings-related exception, so do we need a separate `PromptValueException` and `SettingsValueException`?

Don't get me wrong. I'm not saying what you're asking for is wrong. I'm just asking if there's maybe a middle ground that we should be going for?

src/client/cli/cmd/set.cpp
Saviq: I'd have to pass `cerr` in, too (unless it was a private member of `Set`)… Since this is only used once, does it really help?

src/client/common/prompters.cpp
Saviq: Indeed! I'm on the fence about `eof()`, though. This is meant to be interactive, so is redirection a use case we should optimize for? And if not, is EOF expected when interactive? If you close your terminal at the prompt, should that count as empty string input?

Saviq: I've gone for rejecting `!good()`.

include/multipass/exceptions/cmd_exceptions.h
ricab: OK, I see what you mean. I guess proliferating exception types is what we pay to avoid error-handling boilerplate in each function call (as you see in Go, eww). With exceptions, you don't need to `if (ret != success) return ret;` in every node of the call tree, but you need a way to distinguish problems arising at different nodes, *if they need different treatment*. Hence the multiple types. It's still a very worthy tradeoff IMO. We may also choose to introduce umbrella headers and revise exception hierarchies at some point.

In general, when programing a reusable unit, one doesn't know how it will be used and how its errors should be handled. One of the main points of exceptions is exactly to delegate error handling up the stack. Because the throwing unit doesn't know if its errors merit specific treatment, the generic approach is to throw a dedicated type. However, since we are the only consumer, we can take the lazy approach of throwing `runtime_error` until we know it needs special treatment. 

I understand you're after the same sort of thing, just a little more refined. I think that makes sense, but it feels to me like there is a naming issue. Perhaps `CLIException` or `TerminalException` would be better names? At face value, `ValueException` could easily arise in situations where the command line isn't at fault (e.g. reading a setting from disk). And yet, it's maybe not the best fit for a failing stream.

> Nothing in a Prompter could throw a Settings-related exception, so do we need a separate `PromptValueException` and `SettingsValueException`?

As with `runtime_error`, you can analyze the call tree and determine that the generic exception is only thrown in one situation today, but that's fragile. When we integrate Prompter and Settings (or whenever we descend into both from a single try block), we may want to distinguish those cases: one of them is not necessarily a cmd line error.

src/client/cli/cmd/set.cpp
ricab: Right, private. Reuse is not the only reason for small functional units. I think it would be better for readability, but I recognize it's not a big deal here.

I once watched a talk that really clicked with me and I really wish I still had the link to. The presenter asked (rephrasing from memory, but you get the idea): 

*When a friend asks you to describe your weekend, do you say "I got up, had breakfast, took a shower, chose my clothes, donned those clothes, chose my shoes, put them on, opened the door, exited the house, closed the door, inserted the key, turned the key 720° clockwise, went downstairs, entered the car..."? Or do you say "I went on a road trip, it was great", with perhaps a few other relevant details? It's the same with programming.*

It stuck with me. Not sure if I shared this yet. That's not a fair representation of the code above, but things tend to become more and more like that unless split out. I think `val = prompt(key)` would be more immediate for the reader, plus *code is read much more often than written*. But again, not a big deal here, just sharing where I am coming from.

src/client/common/prompters.cpp
ricab: Makes sense. Thanks!

src/client/cli/cmd/set.cpp
Saviq: Done

include/multipass/exceptions/cmd_exceptions.h
Saviq: OK went for a `PromptException` in a new `cli_exceptions.cpp` in the end.

townsend2010 (commented):
Hey @Saviq,

This looks good to me.  Just a question about usage of this interface.  So your intent is to put any and all prompters into the `prompters.{h|cpp}` files and not separate prompter files, right?

Saviq:
> This looks good to me. Just a question about usage of this interface. So your intent is to put any and all prompters into the `prompters.{h|cpp}` files and not separate prompter files, right?

Yeah I didn't think it made sense to split it all out. There would be more copyright header than meat in those…

townsend2010 (approved):
> Yeah I didn't think it made sense to split it all out.

Ok, sure, makes sense.

bors merge

Aidenmichael (commented):
learn

PR #2348, opened 2021-11-18
[tests] Split out cli client tests

townsend2010:
This is a first step to help organize the client cli tests and make them more manageable.


Saviq (changes requested):
This needs a rebase, and a `s/2021/2022/` in the headers :)

Noice otherwise :)

Saviq:
@townsend2010 bump?

townsend2010:
Due to the changes I made for the authentication stuff, a `MockPlatform` needs to be defined *before* 
`StrictMock<MockDaemonRpc> mock_daemon` in `ClientTestFixture` so the `set_server_socket_restrictions()` call in the`DaemonRpc` ctor can be mocked.  However, due to how the singleton boiler plate is currently done, this `MockPlatform` definition cannot be done in a header due to the following error:
```
In file included from /home/townsend/multipass/multipass/tests/client/test_cli_client.cpp:18:
/home/townsend/multipass/multipass/tests/client/client_test_fixture.h:81:8: error: ‘multipass::test::ClientTestFixture’ has a field ‘multipass::test::ClientTestFixture::platform_attr’ whose type uses the anonymous namespace [-Werror=subobject-linkage]
   81 | struct ClientTestFixture : public Test
      |        ^~~~~~~~~~~~~~~~~
```
One of the main points of this refactoring is to move the `MockDaemonRpc` object into a class (`ClientTestFixture`) in a shared header and removing this defeats the purpose of most of this refactoring.  Until this issue can be overcome, I think this should wait.

jibel:
@andrei-toterman to review what this PR is about and close it.

andrei-toterman:
I don't think these changes are still relevant today. Those changes are splitting a test file into multiple test files, but the current efforts are about splitting out the test executable itself into multiple test executables. While the changes themselves make sense, they are very out of date. So we can still split up a file into multiple files just for easier development, but I think it would be too difficult to build on top of this PR.

PR #2357, opened 2021-11-26
[snap] extend `stop-timeout` to 5 minutes

### LINKED ISSUE #2316: [snap] auto-refresh wreaking havoc if kill time out hits
All the vms became UNKNOW status.
image

I try to modify this file, reset the status of vms.
`sudo vi /var/snap/multipass/common/data/multipassd/multipassd-vm-instances.json`
and set "machine_type" to "pc-i440fx-bionic" for the broken instance, and then I start one of the instances.

1. start one instance
image

2. the log of multipassd, in the log is waiting for ssh to be up.

image

3.  through the log, we can get the IP of the instance `t-istio-edge-1`. 
image

4. But finally, the result is
```
t-istio-edge-2: timed out waiting for response

```
image

```

ubuntu@sdewan:~$ m shell primary-node-2
shell failed: Cannot retrieve credentials in unknown state

```
actually the VM is running

-- other logs 
image
image

Saviq:
Fixes #2316

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -50,6 +50,7 @@ apps:
       XDG_CONFIG_HOME: &daemon-config $SNAP_DATA/config
       DAEMON_CONFIG_HOME: *daemon-config # temporary
     daemon: simple
+    stop-timeout: 5m
     plugs:
       - all-home
       - firewall-control


luis4a0:
Hey @Saviq, thanks for this. I tested launching 20 VM's and it takes roughly 2'30" to stop Multipass on my computer. In this sense, the timeout works. However, when I restarted the snap, there are four VM's in unknown state. So I guess there must be something more. I am investigating.

luis4a0:
There is still a bug in suspending/unsuspending, but not related to timeouts. So this PR is good to go.

bors r+

PR #2375, opened 2021-12-17
[qemu] Fix starting from resume regression

### LINKED ISSUE #2374: [qemu] instances unreachable after daemon restart
```plain
$ mp launch
Launched: artistic-waterbear                                                    
$ mp exec artistic-waterbear -- uname -a
Linux artistic-waterbear 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ mp suspend artistic-waterbear
$ mp start artistic-waterbear
$ mp exec artistic-waterbear -- uname -a
Linux artistic-waterbear 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ snap stop multipass
Stopped.
$ snap start multipass
Started.
$ mp ls 
artistic-waterbear      Unknown           --               Ubuntu 20.04 LTS
```

And it's stuck like that.

The one reason that comes to mind is that we're releasing the IP on VM object destruction:
```plain
Dec 15 12:56:02 nuc dnsmasq-dhcp[53624]: DHCPRELEASE(mpqemubr0) 10.139.137.91 52:54:00:35:8d:1f               
```

So after the VM comes back, we don't know where it is.

[multipassd.log](https://github.com/canonical/multipass/files/7719496/multipassd.log)

townsend2010:
Fixes #2374

### DIFF
--- a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
+++ b/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
@@ -125,6 +125,12 @@ mp::QemuPlatformDetail::QemuPlatformDetail(const mp::Path& data_dir)
 
 mp::QemuPlatformDetail::~QemuPlatformDetail()
 {
+    for (const auto& it : name_to_net_device_map)
+    {
+        const auto& [tap_device_name, hw_addr] = it.second;
+        remove_tap_device(tap_device_name);
+    }
+
     delete_virtual_switch(bridge_name);
 }
 
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -213,8 +213,6 @@ mp::QemuVirtualMachine::~QemuVirtualMachine()
             shutdown();
         }
     }
-
-    qemu_platform->remove_resources_for(vm_name);
 }
 
 void mp::QemuVirtualMachine::start()
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -40,6 +40,8 @@
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
+using namespace std::chrono_literals;
+
 namespace
 {
 constexpr auto suspend_tag = "suspend";
@@ -187,13 +189,36 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
       qemu_platform{qemu_platform},
       monitor{&monitor}
 {
-    QObject::connect(this, &QemuVirtualMachine::on_delete_memory_snapshot, this,
-                     [this] {
-                         mpl::log(mpl::Level::debug, vm_name, fmt::format("Deleted memory snapshot"));
-                         vm_process->write(hmc_to_qmp_json("delvm " + QString::fromStdString(suspend_tag)));
-                         delete_memory_snapshot = false;
-                     },
-                     Qt::QueuedConnection);
+    QObject::connect(
+        this, &QemuVirtualMachine::on_delete_memory_snapshot, this,
+        [this] {
+            mpl::log(mpl::Level::debug, vm_name, fmt::format("Deleted memory snapshot"));
+            vm_process->write(hmc_to_qmp_json("delvm " + QString::fromStdString(suspend_tag)));
+            is_starting_from_suspend = false;
+        },
+        Qt::QueuedConnection);
+
+    // The following is the actual code to reset the network via QMP if an IP address is not obtained after
+    // starting from suspend.  This will probably be deprecated in the future.
+    QObject::connect(
+        this, &QemuVirtualMachine::on_reset_network, this,
+        [this] {
+            mpl::log(mpl::Level::debug, vm_name, fmt::format("Resetting the network"));
+
+            auto qmp = QJsonDocument::fromJson(qmp_execute_json("set_link")).object();
+            QJsonObject args;
+            args.insert("name", "virtio-net-pci.0");
+            args.insert("up", false);
+            qmp.insert("arguments", args);
+
+            vm_process->write(QJsonDocument(qmp).toJson());
+
+            args["up"] = true;
+            qmp["arguments"] = args;
+
+            vm_process->write(QJsonDocument(qmp).toJson());
+        },
+        Qt::QueuedConnection);
 }
 
 mp::QemuVirtualMachine::~QemuVirtualMachine()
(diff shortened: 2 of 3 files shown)


src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -125,6 +125,12 @@ mp::QemuPlatformDetail::QemuPlatformDetail(const mp::Path& data_dir)
 
 mp::QemuPlatformDetail::~QemuPlatformDetail()
 {
+    for (const auto& it : name_to_net_device_map)
+    {
+        const auto& [tap_device_name, hw_addr] = it.second;
+        remove_tap_device(tap_device_name);
+    }
```
Saviq, line 132: This we should've done unrelated to the bug, IIUC?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -213,8 +238,6 @@ mp::QemuVirtualMachine::~QemuVirtualMachine()
             shutdown();
         }
     }
-
-    qemu_platform->remove_resources_for(vm_name);
```
Saviq, line 217: We still need this on purge, don't we?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -386,6 +410,20 @@ void mp::QemuVirtualMachine::on_restart()
 
 void mp::QemuVirtualMachine::ensure_vm_is_running()
 {
+    if (is_starting_from_suspend)
+    {
+        // Due to https://github.com/canonical/multipass/issues/2374, the DHCP address is removed from
+        // the dnsmasq leases file, so if the daemon restarts while an instance is suspended and then
+        // starts the instance, the daemon won't be able to reach the instance since the instance
+        // won't refresh it's IP address.  The following will force the instance to refresh by resetting
```
Saviq, line 418:
```suggestion
        // won't refresh its IP address.  The following will force the instance to refresh by resetting
```

Saviq (changes requested):
Any tests we could add?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -187,13 +189,36 @@ mp::QemuVirtualMachine::QemuVirtualMachine(const VirtualMachineDescription& desc
...
+        this, &QemuVirtualMachine::on_delete_memory_snapshot, this,
+        [this] {
+            mpl::log(mpl::Level::debug, vm_name, fmt::format("Deleted memory snapshot"));
+            vm_process->write(hmc_to_qmp_json("delvm " + QString::fromStdString(suspend_tag)));
+            is_starting_from_suspend = false;
+        },
+        Qt::QueuedConnection);
+
+    // The following is the actual code to reset the network via QMP if an IP address is not obtained after
+    // starting from suspend.  This will probably be deprecated in the future.
```
Saviq, line 202:
```suggestion
    // starting from suspend. TODO: remove this in the future.
```

src/platform/backends/qemu/qemu_virtual_machine.cpp
townsend2010: Not here.  Purge in the daemon already calls this directly.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
townsend2010: No, because this is the whole reason why I called `remove_resources_for()` in the `QemuVirtualMachine` dtor to begin with.  Now that there is a bug actually calling that, the logic needs to be moved here.

townsend2010:
> Any tests we could add?

Yes, if we want this to take longer and take away from other things.  We'd need to add a non-trivial amount of testing code to emulate this failure and such.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -386,6 +410,20 @@ void mp::QemuVirtualMachine::on_restart()
...
 void mp::QemuVirtualMachine::ensure_vm_is_running()
 {
+    if (is_starting_from_suspend)
+    {
+        // Due to https://github.com/canonical/multipass/issues/2374, the DHCP address is removed from
+        // the dnsmasq leases file, so if the daemon restarts while an instance is suspended and then
+        // starts the instance, the daemon won't be able to reach the instance since the instance
+        // won't refresh its IP address.  The following will force the instance to refresh by resetting
+        // the network every 30 seconds until the start timoeut is reached.
+        if (std::chrono::steady_clock::now() > current_resume_start_time + 30s)
```
Saviq, line 420: Should we do this sooner after resume? Normally resume takes mere seconds, with this it will only work after 30s. Maybe we could do first at 5s, then at 30s intervals?

townsend2010: It's just an arbitrary time, so I guess 5s would be enough.  I think how fast an instance resumes really depends on the load of the system, but sure I can do 5s and then 30s thereafter.

Saviq: Of course, this will mean, though, that best case scenario is reduced to 5s, worst case is only delayed by 16.3(3)% :)

townsend2010: Ok, done.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -232,7 +255,9 @@ void mp::QemuVirtualMachine::start()
         mpl::log(mpl::Level::info, vm_name, fmt::format("Resuming from a suspended state"));
 
         update_shutdown_status = true;
-        delete_memory_snapshot = true;
+        is_starting_from_suspend = true;
+        current_resume_start_time = std::chrono::steady_clock::now();
+        network_reset_wait_time = 5s;
```
Saviq, line 260: This could be a bit simpler with a `network_deadline = std::chrono::steady_clock::now() + 5s;` and then `+30s` below, no?

townsend2010: I think what you added is better and should work as intended.  :+1:

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Canonical, Ltd.
+ * Copyright (C) 2022 Canonical, Ltd.
```
townsend2010, line 2: This should be:
```
* Copyright (C) 2021-2022  Canonical, Ltd.
```

Saviq (approved):
Yups.

bors merge

Saviq:
Runners went on holiday…

bors retry

Saviq:
Hmm!

bors merge

PR #2411, opened 2022-01-25
Add bash completion for `get` and `set` commands

nicolasbock:
This change adds completion for the `get` and `set` commands.

Signed-off-by: Nicolas Bock <nicolas.bock@canonical.com>

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -165,6 +165,26 @@ _multipass_complete()
         if [[ ${mode_found} == 0 ]]; then opts="${opts} mode="; fi
     }
 
+    # Get the setting keys.
+    _multipass_setting_keys()
+    {
+        local append_string=""
+        local -a settings
+
+        if [[ -v 1 ]]; then
+            append_string=$1
+        fi
+
+        readarray -t settings < <(multipass get --help | \
+            awk 'BEGIN { p = 0; } /^Keys:/ { p = 1; next } /^ *$/ { if (p) { p = 0; }} {if (p) { print }}')
+        for key in ${settings[@]}; do
+            if [[ "${COMP_WORDS[*]}" =~ ${key} ]]; then
+                return
+            fi
+        done
+        opts="${opts} $(printf "%s${append_string}" "${settings[@]}")"
+    }
+
     # Add comma and equals sign to the list of word separators if they are not already there, useful to separate
     # options of the form "opt1=val1,opt2=val2".
     if [[ ! "${COMP_WORDBREAKS}" =~ "," ]]; then
@@ -226,6 +246,11 @@ _multipass_complete()
         "unalias")
             _multipass_aliases
             opts="${opts} ${multipass_aliases}"
+        ;;
+        "get")
+            if [[ ! ${COMP_WORDS[*]} =~ --raw ]]; then
+                opts="${opts} --raw"
+            fi
         ;;
     esac
 
@@ -337,6 +362,13 @@ _multipass_complete()
             "alias")
                 _unspecified_alias_vars
                 # TODO: Do use spaces after the completion of an option starting with "--".
+                compopt -o nospace
+            ;;
+            "get")
+                _multipass_setting_keys
+            ;;
+            "set")
+                _multipass_setting_keys "="
                 compopt -o nospace
             ;;
             "help")


ricab:
Hi @nicolasbock. Thanks for this! I have been longing for this completion for a while, but didn't have time to get to it myself. I haven't tried it yet, but it's very welcome.

There are some other changes in the works that will affect that help text, but we can blend. Adding to my TODO list to review (if no one else does first).

nicolasbock:
Hi @ricab ! Thanks for the comments! If you point me towards these other changes I can modify the completions here to make them work with those changes.

ricab:
Hey @nicolasbock, have a look at #2352. That's a little outdated now, I mean to push again soon. 

The thing is, we're going through an overhaul of the settings, with the client now leaving daemon settings to the daemon. The client will no longer know or care what settings the daemon understands (some of them will soon be dynamic). Instead, it will simply ship get/set/keys requests over as needed. But we have yet to settle on the best way to interrogate settings keys. For now, when the daemon can't be reached, I just have a placeholder for its settings (`'local.*`) in the help text you're parsing. 

I will take your PR into account when considering the best approach and I'd otherwise let it simmer for a little :slightly_smiling_face: But I won't forget it. Hope that's alright!

nicolasbock:
Thanks for the additional info @ricab !

ricab:
bors try

completions/bash/multipass
```diff
@@ -227,6 +242,9 @@ _multipass_complete()
             _multipass_aliases
             opts="${opts} ${multipass_aliases}"
         ;;
+        "get")
+            opts="${opts} --raw"
```
ricab, line 246: Ideally, this would only be added if not already used. But we can probably let it stick, as that's what we're doing for the likes of `--help` already. And it is not technically wrong to say `--raw` multiple times.

completions/bash/multipass
```diff
@@ -339,6 +357,13 @@ _multipass_complete()
                 # TODO: Do use spaces after the completion of an option starting with "--".
                 compopt -o nospace
             ;;
+            "get")
+                _multipass_setting_keys
```
ricab, line 361: Both `get` and `set` both take a single key argument (for the time being, at least). It would be nice to complete that argument only if not yet used (so that we're not suggesting something that doesn't work):

```
$ multipass get client.primary-name local.driver
Need exactly one setting key.
```

ricab (commented):
Hey @nicolasbock, circling back to this as it makes sense to include for the next release (even if it could need updating later).

Nice code :+1:  only a few details inline.

completions/bash/multipass
```diff
@@ -165,6 +165,21 @@ _multipass_complete()
         if [[ ${mode_found} == 0 ]]; then opts="${opts} mode="; fi
     }
 
+    # Get the setting keys.
+    _multipass_setting_keys()
+    {
+        local append_string=""
+        local -a settings
+
+        if [[ -v 1 ]]; then
```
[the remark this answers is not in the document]
nicolasbock: This is working for me. When I type

```console
 multipass set client.primary-na
```

and hit the `tab` key it completes to

```console
multipass set client.primary-name=
```

completions/bash/multipass
nicolasbock: Fixed with

```suggestion
            if [[ ! ${COMP_WORDS[*]} =~ --raw ]]; then
                opts="${opts} --raw"
            fi
```

completions/bash/multipass
nicolasbock: Fixed.

completions/bash/multipass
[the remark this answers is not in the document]
ricab: Weird, it doesn't here: https://asciinema.org/a/ibIom1Ff3YNs2IrLvrPOfVSrr

I wonder what's affecting it. Trying with a plain function doesn't work either:

```
elsa@ricab-xps13:~$ test(){
> if [[ -v 1 ]]; then
>   echo yes
> else
>   echo no
> fi
> }
elsa@ricab-xps13:~$ test
no
elsa@ricab-xps13:~$ test asdf
```

This works though:
```
elsa@ricab-xps13:~$ test(){ if [[ -v toto ]]; then   echo yes; else   echo no; fi; }
elsa@ricab-xps13:~$ test
no
elsa@ricab-xps13:~$ toto=asfd
elsa@ricab-xps13:~$ test
yes
```

Perhaps `-v` for argument numbers only works on very recent bash? I have 5.0.17.

("elsa" is just a sandbox account I use)

sharder996:
Closing in favour of https://github.com/canonical/multipass/pull/3550 which utilizes the `multipass get --keys` flag.

PR #2414, opened 2022-01-26
[certs] Add some trace logging for client cert detection

townsend2010:


src/client/common/client_common.cpp
```diff
@@ -140,11 +141,13 @@ std::shared_ptr<grpc::Channel> mp::client::make_channel(const std::string& serve
         const std::vector<QString> cert_dirs{data_location + gui_client_cert_dir, data_location + cli_client_cert_dir};
         for (const auto& cert_dir : cert_dirs)
         {
+            mpl::log(mpl::Level::trace, "client", fmt::format("Searching for client certs in {}…", cert_dir));
             if (client_certs_exist(cert_dir))
             {
                 if (auto rpc_channel{
                         create_channel_and_validate(server_address, get_ssl_credentials_opts_from(cert_dir))})
                 {
+                    mpl::log(mpl::Level::trace, "client", "found!");
```
Saviq, line 150: If this is to stay, it should probably be more expressive.

Saviq (changes requested):

townsend2010:
Ok, when I originally put this up, it was just for debugging some issue, but if you prefer it to actually be in the main code, then I'll fix up the messages more as you suggested.

Saviq:
> Ok, when I originally put this up, it was just for debugging some issue, but if you prefer it to actually be in the main code, then I'll fix up the messages more as you suggested.

Your call, it's just this PR wasn't a draft or `no-merge` :)

townsend2010:
> Your call, it's just this PR wasn't a draft or no-merge

Sure.  It would probably be beneficial though to have the logging, so I'll fix it.

src/client/common/client_common.cpp
townsend2010: Looking at what I added here again, this will print out after this line I added: https://github.com/canonical/multipass/blob/623e8fe3bdb598292fc480c13a1496db3eee63e1/src/client/common/client_common.cpp#L144

I guess I can make it more verbose, but basically it is just confirming that it was found for the previous log message.

Saviq: Sure, problem is when these two lines get separated by some other concurrent logging, it's difficult to correlate. There's more cases like that of course, but we should try and avoid adding more :)

townsend2010: Ok, done.

townsend2010:
@Saviq, this should be ready now.

townsend2010:
Hmm, now that Github is all green, maybe bors will get picked up now...

bors r=Saviq

townsend2010:
Ok, everything is green again.  Let' see if this takes this time.

bors r=Saviq

townsend2010:
bors merge=Saviq

Saviq:
bors retry

Saviq:
My there was a lot of bad luck on this. One last try.

bors retry

Issue #2317, opened 2021-10-28
[alias] Error message has wrong usage when invoking alias through 'multipass'


townsend2010:
```
$ multipass mp-ls -la
Unknown options: l, a.

Options to the inner command should come after "--", like this:
multipass <alias> -- <command> <arguments>
```
The `<command>` part in the hint in incorrect.  This is what works:
```
$ multipass mp-ls -- -la
total 28
drwxr-xr-x 4 ubuntu ubuntu 4096 Oct 27 09:54 .
drwxr-xr-x 3 root   root   4096 Oct 27 09:54 ..
-rw-r--r-- 1 ubuntu ubuntu  220 Feb 25  2020 .bash_logout
-rw-r--r-- 1 ubuntu ubuntu 3771 Feb 25  2020 .bashrc
drwx------ 2 ubuntu ubuntu 4096 Oct 27 09:54 .cache
-rw-r--r-- 1 ubuntu ubuntu  807 Feb 25  2020 .profile
drwx------ 2 ubuntu ubuntu 4096 Oct 27 09:54 .ssh
```
So no "command" is used, but rather just arguments that are particular to the alias.

_Issue created from a [Mattermost message](https://chat.canonical.com/_redirect/pl/ahybhbhf73gx8caw74hcuj7amh) *by @townsend2010*._

PR #2427, opened 2022-02-04
Fix error message when not using `--` for alias arguments

luis4a0:
The error message when launching an alias using `multipass` was incorrect. This PR fixes https://github.com/canonical/multipass/issues/2317 by improving the error message.


Saviq (changes requested):
The `exec` message was correct, please don't change that.

While the position of `--` doesn't actually matter, as long as it's before any options, it's customary for it to be placed before the command, and `multipass help exec` says as much, too.

Saviq (approved):
Good!

bors merge

Saviq:
Fixed a CI problem.

bors merge

Saviq:
GH having a bad day…

PR #2434, opened 2022-02-08
[settings] Do not generate a hash for an empty passphrase

### LINKED ISSUE #2410: [set] fix handling of empty passphrases
```plain
$ sudo multipass set local.passphrase=
# or
$ sudo multipass set local.passphrase  # pass empty twice
Please enter passphrase:
Please re-enter passphrase:
$ sudo multipass get local.passphrase  # bug https://github.com/canonical/multipass/issues/2398
d72c87d0f077c7766f2985dfab30e8955c373a13a1e93d315203939f542ff86e73ee37c31f4c4b571f4719fa8e3589f12db8dcb57ea9f56764bb7d58f64cf705
$ multipass register ""
```

This shouldn't work. Setting an empty passphrase should set the internal representation to empty, and so it should never work.

Not sure if interactive `set` should ever allow empties.

townsend2010:
This will instead unset the passphrase.

Fixes #2410

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2072,7 +2072,7 @@ try
 
     auto stored_hash = MP_SETTINGS.get(mp::passphrase_key);
 
-    if (stored_hash.isNull())
+    if (stored_hash.isNull() || stored_hash.isEmpty())
     {
         return status_promise->set_value(
             grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
--- a/src/utils/settings.cpp
+++ b/src/utils/settings.cpp
@@ -215,7 +215,10 @@ void multipass::Settings::set_aux(const QString& key, QString val) // work with 
     else if (key == winterm_key || key == hotkey_key)
         val = mp::platform::interpret_setting(key, val);
     else if (key == passphrase_key)
-        val = MP_UTILS.generate_scrypt_hash_for(val);
+    {
+        if (!val.isEmpty())
+            val = MP_UTILS.generate_scrypt_hash_for(val);
+    }
 
     auto settings = persistent_settings(key);
     checked_set(*settings, key, val, mutex);


Saviq (commented):
When first installed:
```plain
$ multipass register foo
register failed: Passphrase is not set. Please `multipass set local.passphrase` with a trusted client.
```

But after setting the passphrase to empty, it just says it's not correct:
```plain
$ sudo multipass set local.passphrase=""
$ multipass register foo
register failed: Passphrase is not correct. Please try again.
```

I think we need to say the same in both cases. Which of the two, I'm not sold on…

Saviq:
> There is no way to authenticate the client if the passphrase is not set, so the user should be told as such. I'll fix it to be this way barring any counterarguments to what I've said 🙂

The only reason why I'm on the fence is that you're telling a potential attacker that it's not set, rather than plainly "it's wrong". In theory it could change the attack vector. Something like saying "there is (no )such user" on a password reminder form. Suddenly an attacker knows they can focus on a particular account. Just a gut feeling is all, though :)

townsend2010:
Hmm, well, the info tells the user (attacker?) that they have to use something that is (will be) already documented.  Also, if the attacker knows that it is an incorrect passphrase, they could simply switch to the aforementioned attack vector and try setting the passphrase to a different passphrase.  Since we are more for a "nice" user experience, I don't think explicitly saying that the passphrase is not set is opening this up for malicious attacks.  Just my $0.02.

townsend2010:
Just to add to this, I'm mainly concerned about the user experience of when a passphrase is not set vs. just entering the wrong passphrase.  If the perceived security of just saying the passphrase is wrong no matter what is better in the eyes of users and security experts, then I'm for that.

townsend2010:
Ok, I fixed it such that the "Passphrase is not set" message is displayed when the passphrase is blank.  We'll go from here.

Saviq (approved):
Yups.

bors merge

PR #2441, opened 2022-02-11
[settings] Do not return hash when getting local.passphrase

### LINKED ISSUE #2398: Don't show the hash on `multipass get local.passphrase`
We shouldn't show the passphrase:

```plain
$ sudo mp get local.passphrase
<empty>
$ sudo mp set local.passphrase=foobar
$ sudo mp get local.passphrase
839dcbcdef8576ebf3511eb9ebfca9a1d7f876d62f0340b5a9a6fdcf7f49d6e7c77f2478e68eb85dbc3850ce00f9c5d249154f2e6dbe236745ed6e66aa3466f0
```

townsend2010:
Instead return "true" or "false" depending on if it's set or not.

Fixes #2398

### DIFF
--- a/src/utils/settings.cpp
+++ b/src/utils/settings.cpp
@@ -124,6 +124,10 @@ QString checked_get(const mp::WrappedQSettings& settings, const QString& key, co
     auto ret = settings.value(key, fallback).toString();
 
     check_status(settings, QStringLiteral("read"));
+
+    if (key == mp::passphrase_key)
+        ret = ret.isEmpty() ? QStringLiteral("false") : QStringLiteral("true");
+
     return ret;
 }
 


Saviq (changes requested):
We say `<empty>` when things are, should probably do the same here rather than `false`. And also wrap `true` in `<>` for consistency.

https://github.com/canonical/multipass/blob/b382c093a3b276c9e5a01f9577a465a70349d3aa/src/client/cli/cmd/get.cpp#L38-L39

This is done in `get` above, but that means actually pushing the values over. And I'm not sure it's the right approach for the passphrase hash.

@ricab since this will all change with your settings handlers, care to weigh in?

townsend2010:
Hmm, I originally had `<empty>` and `<true>`, but those two terms don't seem to, uh, mix well for a UI.  In my mind, when one sees `true` for something, the opposite is `false`.  When `<empty>` is seen, I would expect the opposite to be the actual value if it's not empty.  After not liking the `empty`/`true` output, I just ended up following what LXD does with this.

> This is done in `get` above, but that means actually pushing the values over the RPC. And I'm not sure it's the right approach for the passphrase hash.

I don't believe this is correct in the current implementation.  IIUC, the client reading directly from file in this case using the `Settings` singleton which does not touch the RPC. I know this is all going to change, but I put up this PR since we really need to get a release out soon.

Saviq (approved):
OK, I can get behind marshalling a passphrase value returning a boolean after all.

bors merge

PR #2462, opened 2022-03-01
[qemu/platform] Add plumbing for getting networks

townsend2010:


Saviq:
@Saviq ok. so can confirm the current 1.9 dev build isn't ready for this yet. Attempting to start a "simple" instance using

`multipass launch --network en11 --name networktest`

results in a timeout, the VM being set to an `Unknown` status, and appears to take down the `multipassd` service, as subsequent attempts cannot connect to the multipass socket.

I also noticed that my driver is set to `qemu`, and that it says invalid driver when I try change it to `hyperkit`.

_Originally posted by @mlb5000 in https://github.com/canonical/multipass/issues/1857#issuecomment-1063261680_

townsend2010:
Hey @mlb5000,

Are you using an M1 or an Intel Mac?  What version of macOS do you have?

Thanks!

Saviq:
@mlb5000 that timeout we just found, yes (it wasn't timing out before, I promise you). We're digging into it. `hyperkit`'s only available on Intel CPUs, FWIW.

Saviq:
bors merge

PR #2466, opened 2022-03-02
[metrics_provider] Remove implementation

### LINKED ISSUE #767: Send usage data question blocks run when running in a batch script
I was wondering why a batch script was taking a while to complete and noticed that multipass was blocked on asking me the following question:

"Send usage data (yes/no/Later)? no"

Tad annoying to have a tool that blocks like this when running the tool inside a python script and it blocks waiting for input when I'm trying to get automated multipass scripts to work. Is there a way around this?

Also the question has a lower case "yes" and "no" but an upper case "Later" which is rather awkward.

### LINKED ISSUE #1162: Disable interactivity when reading content from stdin
> The problems you refer to were around the metrics question being asked, which should never be shown if we're not on a tty, or when there is anything to be read from stdin, because how do we differentiate between cloud-init being read and the answer to the question. Reading stdin and interactivity are conflicting. 

Which we don't ensure today:
```
$ multipass launch --cloud-init -
#cloud-config
One quick question before we launch … Would you like to help                    
the Multipass developers, by sending anonymous usage data?
This includes your operating system, which images you use,
the number of instances, their properties and how long you use them.
We’d also like to measure Multipass’s speed.

Creating guaranteed-goat -^C
```

We need a generic mechanism of changing the terminal mode to non-interactive if we're to read any data in from it.

_Originally posted by @Saviq in https://github.com/CanonicalLtd/multipass/pull/1161/review_comment/create_

### LINKED ISSUE #2465: Remove the metrics provider infrastructure
The original intent to upload Multipass-based metrics to the internal metrics service has been deprecated.  In such, the whole `MetricsProvider` implementation can be removed along with the metrics related question shown during `launch` (for now).

townsend2010:
The metrics service we planned on using has been deprecated, thus removing the need
for this code.

Fixes #767, Fixes #1162, Fixes #2465


Saviq (approved):
Yup, thanks!

bors merge

Saviq:
bors retry

townsend2010:
One more time...

bors retry

PR #2481, opened 2022-03-16
Get `--keys` and prepare the ground for wildcards

ricab:
Expand the `get` command in a couple of ways:

1. Expand the command's syntax to
    - accept multiple keys;
    - accommodate wildcards instead of concrete keys.
2. Add a `--keys` option, to obtain available settings at any given time;
    1. restrict the output to the specified single key, if given.

As discussed, parts of the implementation are postponed for now (so we can prioritize instance mod). Step 1 above opens the command for future support of multiple keys and wildcards (which are meant to work with and without the `--keys` flag) but stops short of implementing that support for now. Step 2 is implemented in a way that will accommodate the same wildcard arg(s) for `--keys`, but interprets only a single key arg at the moment. That's a particular case of wildcard matching which, although not very useful for the time being, is consistent with the end goal.

### DIFF
--- a/src/client/cli/cmd/get.cpp
+++ b/src/client/cli/cmd/get.cpp
@@ -96,10 +96,10 @@ mp::ParseCode cmd::Get::parse_args(mp::ArgParser* parser)
         {
             arg = args.at(0);
         }
-        else if (!keys_opt)
+        else if (!keys_opt || args.count()) // support 0 or 1 positional arg when --keys is given
         {
-            cerr << "Multiple settings values not implemented yet. Please try again with one single settings key for "
-                    "now.\n";
+            cerr << "Multiple settings not implemented yet. Please try again with either a single key or just the "
+                    "`--keys` flag for now.\n";
             status = ParseCode::CommandLineError;
         }
     }
--- a/src/client/cli/cmd/get.cpp
+++ b/src/client/cli/cmd/get.cpp
@@ -120,7 +120,7 @@ void cmd::Get::print_settings() const
 
 void multipass::cmd::Get::print_keys() const
 {
-    const auto keys = MP_SETTINGS.keys(); // TODO@ricab update this to remove fuzziness (i.e. local.*)
+    const auto keys = MP_SETTINGS.keys();
     const auto format = "{}\n";
 
     if (arg.isEmpty())
--- a/src/client/cli/cmd/get.cpp
+++ b/src/client/cli/cmd/get.cpp
@@ -38,10 +38,10 @@ mp::ReturnCode cmd::Get::run(mp::ArgParser* parser)
         {
             if (keys)
                 print_keys();
-            else if (!key.isEmpty())
+            else if (!arg.isEmpty())
             {
-                if (const auto val = MP_SETTINGS.get(key);
-                    key == mp::passphrase_key) // TODO integrate into setting specs
+                if (const auto val = MP_SETTINGS.get(arg);
+                    arg == mp::passphrase_key) // TODO integrate into setting specs
                     cout << (val.isEmpty() ? "false" : "true");
                 else if (val.isEmpty() && !raw)
                     cout << "<empty>";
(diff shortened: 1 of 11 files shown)


src/client/cli/cmd/get.cpp
```diff
@@ -67,34 +65,71 @@ QString cmd::Get::short_help() const
...
-    parser->addPositionalArgument("key", "Path to the setting whose configured value should be obtained.", "<key>");
+    parser->addPositionalArgument(
+        "arg", "Setting(s) key(s), i.e. path(s) to the setting(s) whose configured value(s) should be obtained.",
+        "[<arg> ...]");
 
     QCommandLineOption raw_option("raw", "Output in raw format. For now, this affects only the representation of empty "
                                          "values (i.e. \"\" instead of \"<empty>\").");
+    QCommandLineOption keys_option("keys", "List available settings keys. This outputs the whole list of currently "
+                                           "available settings keys, or a matching subset when <arg>... is(are) "
+                                           "provided");
```
Saviq, line 83:
```suggestion
                                           "available settings keys, or a matching subset when any <arg> are "
                                           "provided");
```

src/client/cli/cmd/get.cpp
```diff
@@ -67,34 +65,71 @@ QString cmd::Get::short_help() const
...
         {
-            key = args.at(0);
+            arg = args.at(0);
         }
-        else
+        else if (!keys_opt || args.count()) // support 0 or 1 positional arg when --keys is given
         {
-            cerr << "Need exactly one setting key.\n";
+            cerr << "Multiple settings not implemented yet. Please try again with either a single key or just the "
+                    "`--keys` flag for now.\n";
```
Saviq, line 102:
```suggestion
                    "`--keys` option for now.\n";
```

src/client/cli/cmd/get.cpp
ricab: Yep, thanks.

src/client/cli/cmd/get.cpp
ricab: Committed.

src/client/cli/cmd/get.cpp
```diff
@@ -67,34 +65,71 @@ QString cmd::Get::short_help() const
...
-    parser->addPositionalArgument("key", "Path to the setting whose configured value should be obtained.", "<key>");
+    parser->addPositionalArgument(
+        "arg", "Setting(s) key(s), i.e. path(s) to the setting(s) whose configured value(s) should be obtained.",
+        "[<arg> ...]");
 
     QCommandLineOption raw_option("raw", "Output in raw format. For now, this affects only the representation of empty "
                                          "values (i.e. \"\" instead of \"<empty>\").");
+    QCommandLineOption keys_option("keys", "List available settings keys. This outputs the whole list of currently "
+                                           "available settings keys, or a matching subset when any <arg> are "
+                                           "provided");
```
Saviq, line 83: And here, as you say it's a bit useless (well, it lets you validate the key), but what's not supported shouldn't be in help text.

Saviq (changes requested):
Could use a warning when remote keys couldn't be obtained?

src/client/cli/cmd/get.cpp
```diff
@@ -67,34 +65,71 @@ QString cmd::Get::short_help() const
...
+                               "no key is specified.");
+    return desc + "\n\n" + describe_common_settings_keys();
 }
 
 mp::ParseCode cmd::Get::parse_args(mp::ArgParser* parser)
 {
-    parser->addPositionalArgument("key", "Path to the setting whose configured value should be obtained.", "<key>");
+    parser->addPositionalArgument(
+        "arg", "Setting(s) key(s), i.e. path(s) to the setting(s) whose configured value(s) should be obtained.",
+        "[<arg> ...]");
```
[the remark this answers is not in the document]
ricab: OK, I gave it another shot. How do you like it now?

src/client/cli/cmd/get.cpp
ricab: Let me know if it's OK now or if there are further changes you'd like to see.

ricab:
@Saviq I pushed the changes we discussed, but with a warning only for the unreachable daemon. Because of the way the code is done, it would be a pain to change it to log and return an error, but still output any keys at all. Let me know what you think.

Saviq:
@ricab did we not say an error getting logged? Regardless of exit code?

```plain
$ multipass get --keys
client.gui.autostart
client.gui.hotkey
client.primary-name
$ multipass get --keys -v
[2022-04-11T12:25:20.521] [warning] [remote settings] Could not reach daemon.
client.gui.autostart
client.gui.hotkey
client.primary-name
```

ricab:
Thanks for the review. 

I was about to comment: logging an error but otherwise succeeding leaves a stone in my shoe. But I think the real issue is not showing warnings by default. Warnings are supposed to warn, they should be visible. Hopefully we can revise them at some point and make warning the default.

Issue #2497, opened 2022-03-29
Rename "Workflow" implementation code to "Blueprints"


townsend2010:
This will make it clear in the code what it is for and remove the legacy name.

PR #2498, opened 2022-03-29
Rename "workflows" user strings to "Blueprints"

### LINKED ISSUE #2496: Change user facing "workflows" strings to "blueprints"
This is for the rebranding of "Workflows" to "Blueprints".

Another bug report will be entered for renaming the underlying implementation in the code.

townsend2010:
Fixes #2496

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -466,7 +466,7 @@ void validate_image(const mp::LaunchRequest* request, const mp::VMImageVault& va
     catch (const mp::IncompatibleWorkflowException&)
     {
         throw std::runtime_error(
-            fmt::format("The \"{}\" workflow is not compatible with this host.", request->image()));
+            fmt::format("The \"{}\" Blueprint is not compatible with this host.", request->image()));
     }
     catch (const std::out_of_range&)
     {
@@ -1080,7 +1080,7 @@ try // clang-format on
             catch (const std::exception&)
             {
                 throw std::runtime_error(
-                    fmt::format("Unable to find an image or workflow matching \"{}\"", request->search_string()));
+                    fmt::format("Unable to find an image or Blueprint matching \"{}\"", request->search_string()));
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/daemon/daemon.cpp
```diff
@@ -466,7 +466,7 @@ void validate_image(const mp::LaunchRequest* request, const mp::VMImageVault& va
     catch (const mp::IncompatibleWorkflowException&)
```
luis4a0, line 466: Shall we change the exception name to `mp::IncompatibleBlueprintException`?

src/platform/platform_linux.cpp
```diff
@@ -245,7 +245,7 @@ std::map<std::string, mp::NetworkInterfaceInfo> mp::platform::Platform::get_netw
 
 QString mp::platform::Platform::get_workflows_url_override() const
```
luis4a0, line 246: Same thing here, does this function need to be called `get_blueprints_url_override()`?

luis4a0 (commented):
Hey @townsend2010, thanks for this! Looks good, some quiestion: does "Blueprints" need to start with a capital in the strings?

This, in addition to the function and exception names containing the word "workflow", maybe they need to be changed. What do you think?

townsend2010:
Hey @luis4a0!

Thanks for the review!

> does "Blueprints" need to start with a capital in the strings?

Good question.  We used lower case "workflows" before, but in my opinion, "Blueprints" is now the real product name, so I'm thinking it should be upper case as it's a proper noun.  @ricab, @nathan-j-hart, @Saviq, any opinions on this?

src/daemon/daemon.cpp
townsend2010: There is a bunch of renaming to do with the implementation and that is tracked in #2497.  Since this particular PR is *only* for the release and only affects user facing strings, we should wait on this.

src/platform/platform_linux.cpp
townsend2010: Same as above.

ricab:
Yeah, upper case looks good to me.

luis4a0:
It's OK for me to use capitals if it's a product name, didn't know it was.

luis4a0:
bors retry

:angry:

nathan-j-hart:
+1 to capital letters for Blueprints. It conveys that it is a Multipass-specific feature, and not some general concept the user should be familiar with from elsewhere

PR #2501, opened 2022-03-30
[snap] drop packaged libssh

Saviq:
We build our own.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -159,6 +159,8 @@ parts:
     - on arm64: [qemu-system-arm]
     - qemu-utils
     - libslang2
+    stage:
+    - -usr/lib/*/libssh*
     organize:
       usr/lib/*/pulseaudio/libpulsecommon-*.so: usr/lib/
       usr/share/seabios/*: qemu/
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -94,6 +94,13 @@ apps:
       - system-observe
 
 parts:
+  libvirt:
+    plugin: nil
+    stage-packages:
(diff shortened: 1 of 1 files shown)


Saviq:
bors try

townsend2010 (commented):
I still see `libssh.so.4.8.4` in the snap.  I don't think it breaks anything, but if we are doing this exercise, it should go away.  That said, it may not be trivial to get Snapcraft to remove it, so not going to block this.

Saviq:
bors try

Saviq:
How about now?

bors try

Saviq:
Just for a confirmation on libvirt:

```plain
$ sudo multipass get local.driver
libvirt
$ multipass list
primary                 Running           192.168.122.18   Ubuntu 20.04 LTS
```

snap/snapcraft.yaml
```diff
@@ -126,7 +133,6 @@ parts:
     - libqt5x11extras5
     - libssl1.1
     - libxml2
-    - libvirt0
```
townsend2010, line 129: I think we need a:
```yaml
    after:
      - libvirt
```

for the `multipass` part, right?

townsend2010 (commented):
Inline question below.

snap/snapcraft.yaml
Saviq: No, this is for staging. The `build-packages:` above make it so `libvirt0` is available during the build anyway.

townsend2010: Oh, duh, right :+1:

townsend2010 (approved):
Ok, great, thanks!

bors merge

PR #2526, opened 2022-04-14
Port auto-mount code in launch to settings handlers

ricab:
Let's see how this goes.


ricab:
Note that, if there is a problem fetching the mounts setting for some reason, Multipass will say:

```
$ multipass launch -n primary
Launched: primary
launch failed: [error decription]
```

This is as before, no changes there. However, if the same problem arises on `start` or `shell`, Multipass will now report that `start` or `shell` failed, respectively, while before it would have said that `launch` had failed. If the `mount` itself fails, it will say that `mount` fail (as before).

townsend2010 (approved):
Great, LGTM!

bors merge

townsend2010:
bors retry

PR #2527, opened 2022-04-14
[cli] Don't attempt automount with disabled petenv

### LINKED ISSUE #2447: Random `instance "Home" does not exist` error if launching with `--name` and petname empty
```plain
$ multipass set client.primary-name=
$ multipass launch  
Launched: needed-kea                                                            
instance "Home" does not exist
```

https://github.com/canonical/multipass/blob/a5a8109d442c32fa37be1f2141b2245a7164d79d/src/client/cli/cmd/launch.cpp#L130-L140

We should be looking at the returned name here, not the requested (since that is empty).

ricab:
Fixes #2447.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -126,7 +126,7 @@ mp::ReturnCode cmd::Launch::run(mp::ArgParser* parser)
     }
 
     auto ret = request_launch(parser);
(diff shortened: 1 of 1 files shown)


townsend2010 (approved):
Yep, thanks!

bors merge

Saviq:
bors retry

PR #2529, opened 2022-04-18
[daemon] Move hypervisor health check sooner in ctor

### LINKED ISSUE #2519: LXD VMs constructed with wrong storage pool on daemon startup
When the daemon starts and restores persisted LXD VMs, `LXDVirtualMachineFactory::create_virtual_machine` is called before `LXDVirtualMachineFactory::hypervisor_health_check`. But the later is [where the storage pool is defined](https://github.com/canonical/multipass/blob/3c0c730077e2964e19e15c99f3383a697d0f5c4b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp#L167-L193). Until then, the former passes an empty pool to the `LXDVirtualMachine` constructor. 

Until #2353, this would have no visible effect. The pool parameter would [only be used](https://github.com/canonical/multipass/blob/d8df488fda5e72673f813f3d15be5252b9f5ae9e/src/platform/backends/lxd/lxd_virtual_machine.cpp#L182) for VMs that did not already exist on LXD's side, but those would actually be [skipped](https://github.com/canonical/multipass/blob/2736f3d55134f8ee8d7684f0ee56cb61e41157c2/src/daemon/daemon.cpp#L878-L882) and [deleted](https://github.com/canonical/multipass/blob/2736f3d55134f8ee8d7684f0ee56cb61e41157c2/src/daemon/daemon.cpp#L950-L954) by the daemon instead. With instance mod, however, the storage pool parameter is saved and later used to change the disk property. That operation thus [fails](https://github.com/canonical/multipass/pull/2353#issuecomment-1093283716) for LXD instances that existed prior to the last daemon startup.

**To Reproduce**
With instance mod in:
1. `multipass set local.driver=lxd`
1. `multipass launch --name foo`
1. `snap restart multipass`
1. `multipass set local.foo.disk=20G`

townsend2010:
This allows any initialization of any hypervisor specific primitives to be complete
before any existing instances are instantiated.

Fixes #2519

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -856,6 +856,15 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
     connect_rpc(daemon_rpc, *this);
     std::vector<std::string> invalid_specs;
 
+    try
+    {
+        config->factory->hypervisor_health_check();
+    }
+    catch (const std::runtime_error& e)
+    {
+        mpl::log(mpl::Level::warning, category, fmt::format("Hypervisor health check failed: {}", e.what()));
+    }
+
     for (auto& entry : vm_instance_specs)
     {
         const auto& name = entry.first;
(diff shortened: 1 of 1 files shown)


luis4a0:
bors +r

luis4a0:
Yes, I meant bors r+

townsend2010:
bors retry

Saviq:
bors retry

PR #2540, opened 2022-04-28
[tests] stop installing gtest

Saviq:


ricab:
Anything blocking this?

townsend2010:
@ricab,

Nothing blocking per se, but it didn't look like it did what it was supposed to do and so @Saviq put in https://github.com/canonical/multipass-private/pull/461 instead.  Not sure why this didn't work, so I was reluctant to just close until we understood why.

ricab:
Got it, thanks for explaining.

Saviq:
Well, it's still correct to do in our case (as we don't package anything GTest).

townsend2010 (approved):
Ok, makes sense to use this.

bors merge

PR #2544, opened 2022-04-29
[daemon] Remove `snapcraft:core` remote

luis4a0:
It was on the non-bridging remote list. Since we don't support it anymore, it doesn't make sense for it to stay there.


townsend2010 (changes requested):
Thanks!

Needs a test fixed: `DaemonRefuseSnapcraft/RefuseBridging.old_image/0, where GetParam() = ("snapcraft", "core")`

townsend2010 (approved):
LGTM, thanks!

bors merge

Issue #2553, opened 2022-05-05
[regression] Wrong owner/group used when making new directory in a mount


townsend2010:
On 1.9.0, making a new directory in a mount gives it `root`/`root` ownership instead of `ubuntu`/`ubuntu`.  This also translates to `root`/`root` ownership on the host.

PR #2556, opened 2022-05-05
Fix ownership of a folder created under a mount

luis4a0:
The reverse id mapping function returned `-1` if the current user was not found on the reverse mappings. Since the folder is created as root, when `0` did not exist in the reverse mappings the new folder retained the `0` id, because `-1` makes `chown` to keep the owner.

We use now the user/groups id's of the parent folder if the current user is not found on the reverse mappings. This was the behavior before implementing reverse id mapping.

Fixes https://github.com/canonical/multipass/issues/2553.


townsend2010 (approved):
Ok, works well, thanks!

bors merge

PR #2576, opened 2022-05-17
[singleton] Allow forwarding of base class ctor args

townsend2010:
This helps make singleton classes that require initialization and has no default ctor.


townsend2010:
Not going to go this route.

PR #2584, opened 2022-05-23
[blueprints] Rename all references of workflows to blueprints

### LINKED ISSUE #2497: Rename "Workflow" implementation code to "Blueprints"

townsend2010:
Fixes #2497


luis4a0 (commented):
Hey @townsend2010, thanks! Looks good to me, but it doesn't compile on Mac ARM, I think because the Qt version. How can we fix that?

luis4a0 (approved):
Ok, so this is an issue of private code. Now it's fixed, so let's try to bors it.

PR #2593, opened 2022-05-26
[cli] cloud-init url option

### LINKED ISSUE #1461: cloud-init url

andrei-toterman:
This adds support for `http` and `https` URLs to be specified as an argument for the `--cloud-init` option of the `launch` command.

fix #1461

### DIFF
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -79,7 +79,7 @@ mp::Client::Client(ClientConfig& config)
     add_command<cmd::Alias>(aliases);
     add_command<cmd::Aliases>(aliases);
     add_command<cmd::Authenticate>();
-    add_command<cmd::Launch>();
+    add_command<cmd::Launch>(config.url_downloader);
     add_command<cmd::Purge>(aliases);
     add_command<cmd::Exec>(aliases);
     add_command<cmd::Find>();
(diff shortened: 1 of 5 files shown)


townsend2010 (commented):
Hey @andrei-toterman!

Thanks for this.  I think there may be a less intrusive way of accomplishing the mock you need without introducing a new member to `Launch` and adding it to the `config`.

The `URLDownloader` class makes use of a singleton called `MP_ NETMGRFACTORY` which allows injecting of a `MockQNetworkAccessManager`. See `test_url_downloader.cpp`.

So what you could do in the new unit test itself is to inject the mock early like:
```
auto [mock_network_access_manager, guard] = mpt::MockNetworkManagerFactory::inject();
```

Then much like the `URLDownloader` tests, you can return mcok replies and the such for expected functions.

I hope this makes sense.  If not or you are having troubles with this, then it can wait until I'm back.

townsend2010 (approved):
LGTM, thanks!

bors merge

townsend2010:
bors retry

PR #2596, opened 2022-05-26
[settings] Allow lowering cpu and memory instance settings

### LINKED ISSUE #2580: [settings] Need to allow lowering CPU's and memory
Currently, instance modification will *only* allow raising values for CPU's, memory, and disk.  However, if a user overallocates CPU's or memory, there is no way to lower the setting again.  Disk is different in that it is not trivial to shrink a partition in a virtual disk once it has grown.

townsend2010:
Fixes #2580

### DIFF
--- a/src/daemon/instance_settings_handler.cpp
+++ b/src/daemon/instance_settings_handler.cpp
@@ -118,7 +118,7 @@ mp::MemorySize get_memory_size(const QString& key, const QString& val)
 void update_cpus(const QString& key, const QString& val, mp::VirtualMachine& instance, mp::VMSpecs& spec)
 {
     bool converted_ok = false;
-    if (auto cpus = val.toInt(&converted_ok); !converted_ok || cpus < 1)
+    if (auto cpus = val.toInt(&converted_ok); !converted_ok || cpus < std::stoi(mp::min_cpu_cores))
         throw mp::InvalidSettingException{key, val, "Need a positive integer (in decimal format)"};
     else if (cpus != spec.num_cores) // NOOP if equal
     {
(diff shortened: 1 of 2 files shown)


src/daemon/instance_settings_handler.cpp
```diff
@@ -118,11 +118,9 @@ mp::MemorySize get_memory_size(const QString& key, const QString& val)
 void update_cpus(const QString& key, const QString& val, mp::VirtualMachine& instance, mp::VMSpecs& spec)
 {
     bool converted_ok = false;
-    if (auto cpus = val.toInt(&converted_ok); !converted_ok || cpus < 1)
+    if (auto cpus = val.toInt(&converted_ok); !converted_ok || cpus < std::stoi(mp::min_cpu_cores))
         throw mp::InvalidSettingException{key, val, "Need a positive integer (in decimal format)"};
```
luis4a0, line 122: Can we improve the error message? It's not anymore just a positive integer, the message can be something like "Less than %1 CPUs are not allowed."

luis4a0 (approved):
Thanks @townsend2010! Looks great, just one inline comment.

src/daemon/instance_settings_handler.cpp
townsend2010: Hey @luis4a0!

I'm not sure what you mean.  Nothing really changed here other than I use the already defined `min_cpu_cores` rather than the hardcoded `1`, but it still only accepts a positive integer (in decimal format).

luis4a0: Well, if `min_cpu_cores` is set for whatever reason to 2, the current error message will not be confusing.

townsend2010: Hmm, I see where you are coming from, but that would also apply if the original hard coded `1` was changed to a hard coded `2`.  But yeah, perhaps that error message could be changed a bit to reflect whatever `min_cpu_cores` value is.  You convinced me to change it :slightly_smiling_face:

townsend2010: Ok, changed the message slightly, but left the other wording intact since it's possible to pass in values such as `1.5`, `foo`, etc., so need to be explicit about it being a positive integer in decimal format.

Issue #2298, opened 2021-10-19
Unalias multiple aliases at once


ricab:
**What are you trying to do?**
```
$ multipass unalias alias1 alias2
Wrong number of arguments given
```

**What's your proposed solution?**
`multipass unalias [--all|alias [alias...]]`

luis4a0:
I should say "fixes" instead of "addresses" in the PR to close this issue automatically :facepalm:

PR #2615, opened 2022-06-14
Add `unalias --all`

luis4a0:
This PR adds the argument `--all` to the `unalias` command. It also makes `unalias` capable to remove various alias at once.

Addresses #2298.


src/client/cli/cmd/unalias.cpp
```diff
@@ -32,8 +32,34 @@ mp::ReturnCode cmd::Unalias::run(mp::ArgParser* parser)
...
+            auto i = bad_aliases.begin();
+            if (bad_aliases.size() == 1)
+                cerr << fmt::format("Unexisting alias: {}.\n", *i);
+            else
+            {
+                cerr << fmt::format("Unexisting aliases: {}", *i);
+                for (++i; i != bad_aliases.end(); ++i)
+                    cerr << fmt::format(", {}", *i);
+                cerr << ".\n";
+            }
```
townsend2010, line 54: This feels a little awkward to me.  How about the following:
```c++
            cerr << fmt::format("Nonexistent {}: ", bad_aliases.size() == 1 ? "alias" : "aliases");

            for (const auto& bad_alias : bad_aliases)
            {
                cerr << fmt::format("{}{}", bad_alias, &bad_alias != &bad_aliases.back() ? ", " : "");
            }

            cerr << ".\n";
```

townsend2010 (changes requested):
Works, well, thanks!  Just some comments inline to address.

src/client/cli/cmd/unalias.cpp
andrei-toterman: How about:
```cpp
cerr << fmt::format("Nonexistent {}: {}", bad_aliases.size() == 1 ? "alias" : "aliases", fmt::join(bad_aliases, ", "));
```

townsend2010: Oh @andrei-toterman, nice!

src/client/cli/cmd/unalias.cpp
```diff
@@ -32,8 +32,34 @@ mp::ReturnCode cmd::Unalias::run(mp::ArgParser* parser)
...
+                bad_aliases.push_back(alias_to_remove);
+
+        if (!bad_aliases.empty())
+        {
+            auto i = bad_aliases.begin();
+            if (bad_aliases.size() == 1)
+                cerr << fmt::format("Unexisting alias: {}.\n", *i);
+            else
+            {
+                cerr << fmt::format("Unexisting aliases: {}", *i);
```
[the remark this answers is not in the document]
luis4a0: Right.

src/client/cli/cmd/unalias.cpp
```diff
@@ -32,8 +32,34 @@ mp::ReturnCode cmd::Unalias::run(mp::ArgParser* parser)
...
+
+        for (const auto& alias_to_remove : aliases_to_remove)
+            if (!aliases.exists_alias(alias_to_remove))
+                bad_aliases.push_back(alias_to_remove);
+
+        if (!bad_aliases.empty())
+        {
+            auto i = bad_aliases.begin();
+            if (bad_aliases.size() == 1)
+                cerr << fmt::format("Unexisting alias: {}.\n", *i);
```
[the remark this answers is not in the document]
luis4a0: Right.

townsend2010 (changes requested):
Thanks, almost there!

One thing I found is that the same alias is allowed to be entered multiple times and then the following error is shown:
`Warning: 'No such file or directory' when removing alias script for mp-ls`

IMO, duplicate aliases should just be ignored and treated as a single alias.

Also, and I don't know if it's really a big deal if we handle duplicate aliases like I propose above, but the `unalias` bash completion will show the same alias as an alias to choose even if it's already selected in the command line.

luis4a0:
Done the change in bash completions. Now, each alias is suggested only once. This can be refactored in the future and be reused in commands on which we don't want to repeat (and to avoid repeating options like `--all` or `--help`.

townsend2010 (approved):
Ok, great, thanks!

bors merge

townsend2010:
bors retry

townsend2010:
Ok, let's try one more time...

bors retry

Issue #717, opened 2019-04-10
Support to configure the mirror url of cloud image  


adglkh:
The user experience of using multipass in China is not quite good.

- The process of cloud image download from the official site https://cloud-images.ubuntu.com is interrupted quite often.
- I also hit the issue from time to time in poor network conditions  https://github.com/CanonicalLtd/multipass/issues/386 
- If i was lucky enough, the whole process to download an image from the above URL would take 1~2h, which is really painful.

So I'm wondering if there's a way to configure mirrors in multipass so that I can use something like the URL[1] as a mirror to boost the image download speed.
That could be really helpful for multipass users in China.

Thanks
Gary 

[1] https://mirrors.tuna.tsinghua.edu.cn/ubuntu-cloud-images/

Saviq:
Hey @adglkh,

Certainly, this would be a welcome addition. #276 needs to be solved before, though, so we can authenticate the images downloaded.

wusphinx:
when you visit https://mirrors.tuna.tsinghua.edu.cn/ubuntu-cloud-images/ , you can't find the directory **release**, which will cause something wrong as below: 
```
launch failed: Remote "release" is unknown or unreachable. If image mirror is enabled, please confirm it is valid.
```
but, you can try the following commands
```
multipass set local.image.mirror=https://mirrors.cloud.tencent.com/ubuntu-cloud-images/
```
it works fine.

PR #2632, opened 2022-06-23
Download image from mirror site

rim99:
Hi team, I would like to present a PR for issue: https://github.com/canonical/multipass/issues/717

**Feature: mirror of cloud images**

By setting `local.image.mirror`, users can download images from mirror sites instead of the official one. The feature can be disabled by leaving the config empty.

**Implementation**

When updating product versions, it will choose the manifest from mirror site if possible. Since mirror sites is syncing with official site periodically, the latest version in official site may not exist in mirror sites.

Only if the content of a version is *identical* to the same version from official site, it will be updated into the `manifests` cache of `UbuntuVMImageHost` instance.

### DIFF
--- a/include/multipass/constants.h
+++ b/include/multipass/constants.h
@@ -51,14 +51,15 @@ constexpr auto bridged_network_name = "bridged";
 constexpr auto settings_extension = ".conf";
 constexpr auto daemon_settings_root = "local";
 
-constexpr auto petenv_key = "client.primary-name";     // This will eventually be moved to some dynamic settings schema
-constexpr auto driver_key = "local.driver";            // idem
-constexpr auto passphrase_key = "local.passphrase";    // idem
-constexpr auto bridged_interface_key = "local.bridged-network"; // idem
-constexpr auto mounts_key = "local.privileged-mounts"; // idem
-constexpr auto autostart_key = "client.gui.autostart"; // idem
+constexpr auto petenv_key = "client.primary-name";  // This will eventually be moved to some dynamic settings schema
+constexpr auto driver_key = "local.driver";         // idem
+constexpr auto passphrase_key = "local.passphrase"; // idem
+constexpr auto bridged_interface_key = "local.bridged-network";       // idem
+constexpr auto mounts_key = "local.privileged-mounts";                // idem
+constexpr auto autostart_key = "client.gui.autostart";                // idem
 constexpr auto winterm_key = "client.apps.windows-terminal.profiles"; // idem
 constexpr auto hotkey_key = "client.gui.hotkey";                      // idem
+constexpr auto mirror_key = "local.image.mirror";                     // idem; this defines the mirror of simple streams
 
 [[maybe_unused]] // hands off clang-format
 constexpr auto key_examples = {autostart_key, driver_key, mounts_key};
--- a/include/multipass/simple_streams_manifest.h
+++ b/include/multipass/simple_streams_manifest.h
@@ -28,6 +28,7 @@
 #include <QString>
 
 #include <memory>
+#include <optional>
 #include <vector>
 
 namespace multipass
@@ -35,11 +36,12 @@ namespace multipass
 
 struct SimpleStreamsManifest
 {
-    static std::unique_ptr<SimpleStreamsManifest> fromJson(const QByteArray& json, const QString& host_url);
+    static std::unique_ptr<SimpleStreamsManifest>
+    fromJson(const QByteArray& json, const std::optional<QByteArray>& json_from_mirror, const QString& host_url);
 
     const QString updated_at;
     const std::vector<VMImageInfo> products;
     const QMap<QString, const VMImageInfo*> image_records;
 };
-}
+} // namespace multipass
 #endif // MULTIPASS_SIMPLE_STREAMS_MANIFEST_H
--- a/src/daemon/daemon_config.cpp
+++ b/src/daemon/daemon_config.cpp
@@ -38,6 +38,7 @@
 
 #include <chrono>
 #include <memory>
+#include <optional>
 
 namespace mp = multipass;
 namespace mpl = multipass::logging;
@@ -134,10 +135,12 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
         image_hosts.push_back(std::make_unique<mp::CustomVMImageHost>(QSysInfo::currentCpuArchitecture(),
                                                                       url_downloader.get(), manifest_ttl));
         image_hosts.push_back(std::make_unique<mp::UbuntuVMImageHost>(
-            std::vector<std::pair<std::string, std::string>>{
-                {mp::release_remote, "https://cloud-images.ubuntu.com/releases/"},
-                {mp::daily_remote, "https://cloud-images.ubuntu.com/daily/"},
-                {mp::appliance_remote, "https://cdimage.ubuntu.com/ubuntu-core/appliances/"}},
+            std::vector<std::pair<std::string, UbuntuVMImageRemote>>{
+                {mp::release_remote, UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/", "releases/",
+                                                         std::make_optional<QString>(mp::mirror_key)}},
+                {mp::daily_remote, UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/", "daily/",
+                                                       std::make_optional<QString>(mp::mirror_key)}},
+                {mp::appliance_remote, UbuntuVMImageRemote{"https://cdimage.ubuntu.com/", "ubuntu-core/appliances/"}}},
             url_downloader.get(), manifest_ttl));
     }
     if (vault == nullptr)
--- a/src/daemon/daemon_init_settings.cpp
+++ b/src/daemon/daemon_init_settings.cpp
@@ -59,6 +59,26 @@ QString driver_interpreter(QString val)
     return val;
 }
 
+QString image_mirror_interpreter(QString val)
+{
+    if (val.size() == 0)
+    {
+        return val;
+    }
+
+    if (!val.startsWith("https://"))
+    {
+        throw mp::InvalidSettingException(mp::mirror_key, val,
+                                          "The hostname of mirror must contain protocol name: https");
+    }
+
+    if (!val.endsWith("/"))
+    {
+        val.append("/");
+    }
+    return val;
+}
+
 } // namespace
 
 void mp::daemon::monitor_and_quit_on_settings_change() // temporary
@@ -79,6 +99,7 @@ void mp::daemon::register_global_settings_handlers()
     settings.insert(std::make_unique<CustomSettingSpec>(mp::passphrase_key, "", [](QString val) {
         return val.isEmpty() ? val : MP_UTILS.generate_scrypt_hash_for(val);
     }));
+    settings.insert(std::make_unique<CustomSettingSpec>(mp::mirror_key, "", image_mirror_interpreter));
 
     MP_SETTINGS.register_handler(
         std::make_unique<PersistentSettingsHandler>(persistent_settings_filename(), std::move(settings)));
--- a/src/daemon/ubuntu_image_host.cpp
+++ b/src/daemon/ubuntu_image_host.cpp
@@ -17,8 +17,10 @@
 
 #include "ubuntu_image_host.h"
 
+#include <multipass/constants.h>
 #include <multipass/platform.h>
 #include <multipass/query.h>
+#include <multipass/settings/settings.h>
 #include <multipass/simple_streams_index.h>
 #include <multipass/url_downloader.h>
 
@@ -27,6 +29,8 @@
 #include <multipass/exceptions/unsupported_image_exception.h>
 #include <multipass/exceptions/unsupported_remote_exception.h>
 
+#include <QJsonDocument>
+#include <QJsonObject>
 #include <QUrl>
 
 #include <algorithm>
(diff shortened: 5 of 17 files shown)


townsend2010:
Hi @rim99,

Thank you so much for this!  We will begin reviewing it soon, but in the meantime, I'll run it through CI.

townsend2010:
A couple of things off the top:
1. You will need to sign the [Canonical contributor license agreement](https://ubuntu.com/legal/contributors).
2. You will need to run your changes through `clang-format-12`.

rim99:
Hi @townsend2010, I just signed a new CLA contract, could you help to rerun the CI `cla-check` again, pelase? Thanks

townsend2010:
Hi @rim99,

Thanks, you are good for signing the CLA!  I'll review the code soon.

townsend2010:
Hi @rim99,

I just wanted to let you know that this isn't being ignored.  We are in the late stages of releasing version 1.10 and once that is done, I'll get on starting to review this PR.

Thanks!

src/daemon/ubuntu_image_host.cpp
```diff
@@ -220,28 +224,39 @@ std::vector<std::string> mp::UbuntuVMImageHost::supported_remotes()
...
 {
-    for (const auto& remote : remotes)
+    for (const auto& [remote_name, remote_info] : remotes)
     {
         try
         {
-            check_remote_is_supported(remote.first);
+            check_remote_is_supported(remote_name);
+            auto official_site = remote_info.get_official_url();
+            auto manifest_bytes_from_official = download_manifest(official_site, url_downloader);
```
townsend2010, line 243: Wouldn't it be better to *only* download the official manifest if the mirror url config setting is not set?  This would save a little bandwidth and time fetching a manifest that won't be used.

townsend2010 (changes requested):
Hey @rim99,

Thanks for this.  I've made some inline comments, but I have a larger question about this approach.

In my mind the mirror configuration setting should override the default official mirror, but it looks to me that your code fetches both the official and the configured mirror and then if the configured mirror fails to retrieve the manifest, it will use the official.  Am I reading that correct?  We can discuss further based on your answer :slightly_smiling_face: 

Thanks again!

include/multipass/constants.h
```diff
@@ -51,14 +51,15 @@ constexpr auto bridged_network_name = "bridged";
...
-constexpr auto autostart_key = "client.gui.autostart"; // idem
+constexpr auto petenv_key = "client.primary-name";  // This will eventually be moved to some dynamic settings schema
+constexpr auto driver_key = "local.driver";         // idem
+constexpr auto passphrase_key = "local.passphrase"; // idem
+constexpr auto bridged_interface_key = "local.bridged-network";       // idem
+constexpr auto mounts_key = "local.privileged-mounts";                // idem
+constexpr auto autostart_key = "client.gui.autostart";                // idem
 constexpr auto winterm_key = "client.apps.windows-terminal.profiles"; // idem
 constexpr auto hotkey_key = "client.gui.hotkey";                      // idem
+constexpr auto mirror_key = "local.image.mirror";
```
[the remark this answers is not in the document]
luis4a0: Hey @townsend2010! I'm also on the fence about this. On one hand, `image` would make it easier to understand what is that mirror for. But, on the other hand, `simplestreams` would be more descriptive of what Multipass expects as an image. I vote thus for the latter choice.

Making whatever choice, a comment in the code wouldn't hurt at all :)

src/daemon/daemon_config.cpp
```diff
@@ -134,10 +135,12 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
         image_hosts.push_back(std::make_unique<mp::UbuntuVMImageHost>(
-            std::vector<std::pair<std::string, std::string>>{
-                {mp::release_remote, "https://cloud-images.ubuntu.com/releases/"},
-                {mp::daily_remote, "https://cloud-images.ubuntu.com/daily/"},
-                {mp::appliance_remote, "https://cdimage.ubuntu.com/ubuntu-core/appliances/"}},
+            std::vector<std::pair<std::string, UbuntuVMImageRemote>>{
+                {mp::release_remote, UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/", "releases/",
+                                                         make_optional<QString>(mp::mirror_key)}},
+                {mp::daily_remote, UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/", "daily/",
+                                                       make_optional<QString>(mp::mirror_key)}},
```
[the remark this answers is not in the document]
rim99: done

src/daemon/daemon_config.cpp
```diff
@@ -134,10 +135,12 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
         image_hosts.push_back(std::make_unique<mp::CustomVMImageHost>(QSysInfo::currentCpuArchitecture(),
                                                                       url_downloader.get(), manifest_ttl));
         image_hosts.push_back(std::make_unique<mp::UbuntuVMImageHost>(
-            std::vector<std::pair<std::string, std::string>>{
-                {mp::release_remote, "https://cloud-images.ubuntu.com/releases/"},
-                {mp::daily_remote, "https://cloud-images.ubuntu.com/daily/"},
-                {mp::appliance_remote, "https://cdimage.ubuntu.com/ubuntu-core/appliances/"}},
+            std::vector<std::pair<std::string, UbuntuVMImageRemote>>{
+                {mp::release_remote, UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/", "releases/",
+                                                         make_optional<QString>(mp::mirror_key)}},
```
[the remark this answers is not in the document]
rim99: done

include/multipass/simple_streams_manifest.h
```diff
@@ -28,18 +28,20 @@
...
 #include <vector>
 
 namespace multipass
 {
 
 struct SimpleStreamsManifest
 {
-    static std::unique_ptr<SimpleStreamsManifest> fromJson(const QByteArray& json, const QString& host_url);
+    static std::unique_ptr<SimpleStreamsManifest>
+    fromJson(const QByteArray& json, const std::optional<QByteArray> json_from_mirror, const QString& host_url);
```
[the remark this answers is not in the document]
rim99: Done

src/daemon/daemon_init_settings.cpp
```diff
@@ -59,6 +59,26 @@ QString driver_interpreter(QString val)
...
 }
 
+QString image_mirror_interpreter(QString val)
+{
+    if (val.size() == 0)
+    {
+        return val;
+    }
+
+    if (!val.startsWith("http://") && !val.startsWith("https://"))
```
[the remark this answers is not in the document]
rim99: Makes sense

include/multipass/constants.h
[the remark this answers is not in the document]
rim99: If we don't have strong preference here, I would recommend to stick to current `local.image.mirror`. Since it's not just a key of a configuration, but also an API exposed to the end users. Easy for user to understanding should be a good quality of a product. So I just added a comment here.

However, if we still insist to use the name `simplestreams`, I'm happy to refactor it.

rim99:
Hi @townsend2010, thanks for the feedback.

Regarding to the usage of official and mirror hosts, here is what I thought. Please correct me if I'm wrong, since you guys have more context.

Currently, before Multipass downloading the image, it will retrieve a manifest from the host, which contains the checksum of related images. As you can see, we use the image and checksum from the same host for confirmation. This won't be a problem if we are connecting to the official host, Since the information is ensured by TLS connections.

However, if the host is designated by end users, it shouldn't be trusted anymore. Unless we have some kind of whitelist to confirm the *mirror* host is a valid one. By retrieving the manifest twice, one from the official host, the other from mirror. We have a way to verify that the mirror host does tell the truth. So we can continue to download the images.

What if the mirror is not a valid one? 2 cases I can come up with:
1. the host itself is totally a fake one. It doesn't even have a manifest. The downloading process would fail with a `mp::DownloadException`, which is handled here: https://github.com/canonical/multipass/blob/07d671cd61ac76428f32b48d1fc8e244e3963efa/src/daemon/ubuntu_image_host.cpp#L250, This won't call method: `mp::SimpleStreamsManifest::fromJson` So leaving the `manifests` of `UbuntuVMImageHost` instance empty. 
2. the host have a manifest, however, with a fake checksum. In this case, it will be ruled out of the `manifests` of `UbuntuVMImageHost` instance. Since the method `mp::SimpleStreamsManifest::fromJson` would return `std::nullopt`.

As you can see, for both cases, the the `manifests` of `UbuntuVMImageHost` instance would be empty. And the `launch` command would stopped at the following place. No image would be retrieved from the official host at all.

src/daemon/ubuntu_image_host.cpp
```diff
@@ -275,7 +290,9 @@ mp::SimpleStreamsManifest* mp::UbuntuVMImageHost::manifest_from(const std::strin
                            });
 
     if (it == manifests.cend())
-        throw std::runtime_error(fmt::format("Remote \"{}\" is unknown or unreachable.", remote));
+        throw std::runtime_error(fmt::format(
+            "Remote \"{}\" is unknown or unreachable. If image mirror is enabled, please confirm it is valid.",
+            remote));
```
rim99, line 295: The `launch` command would fail from here, it the mirror is invalid.

src/daemon/ubuntu_image_host.cpp
rim99: If the mirror url config setting is not set, the manifest would be `std::nullopt`. It does download manifest from official  host only.
https://github.com/rim99/multipass/blob/57711ba301ba175265bfe35bf499501bcd623658/src/daemon/ubuntu_image_host.cpp#L246

townsend2010: Ok, I've read your reasoning and I understand your points.  So really, this is *preferring* a mirror if it exists and is a copy of the official site.  One thing I'm a bit concerned about with this approach is if a user is behind a firewall that prevents access to the official site, but allows access to the mirror site.  This wouldn't work for this approach.  But I guess this is no worse than what is already there in that the official site can't be reached anyways and the user can't use Multipass.

Thank you for the explanation and I'll give it some more thought.

include/multipass/constants.h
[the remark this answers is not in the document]
townsend2010: Ok, since this is user facing and "image" makes much more sense to an end user than "simplestreams", let's leave it as is.

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -115,6 +127,14 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(c
         {
             const auto version_string = it.key();
             const auto version = versions[version_string].toObject();
+            const auto version_from_official = manifest_products_from_official[product_key]
+                                                   .toObject()["versions"]
+                                                   .toObject()[version_string]
+                                                   .toObject();
+
+            if (version != version_from_official)
+                continue;
```
rim99, line 136: Hi @townsend2010 , 

This part of code will make sure the invalid remote info will be ruled out.

townsend2010: Hi @rim99,

Ok, I see, thanks for pointing that out.  Are we intending the manifest from the mirror to be **exactly** the same as the manifest from the official host?  If so, then I think this should be more robust in checking, like computing a checksum for the official manifest and then computing the checksum of the mirror manifest and comparing.  Otherwise, if there are differences, I think this should check *everything* possible that would be the same, not just `version`.

rim99: These are the json file for comparing:
* [manifest from Official](https://cloud-images.ubuntu.com/releases/streams/v1/com.ubuntu.cloud:released:download.json)
* [manifest from a mirror](https://repo.huaweicloud.com/ubuntu-cloud-images/releases/streams/v1/com.ubuntu.cloud:released:download.json)

Initially, I was trying to validate it by comparing whole file. And it doesn't works since , like I mentioned before, the mirror only syncs with official site periodically. There are chances that, a version released at the official site, but doesn't exist at the mirror site. 

So current implementation picks versions from manifest of mirror site, since they are available on the mirror site at the moment. It doesn't compare the info other than version, because these are useless for an image, which is what it want ultimately. The crucial info has been verified.

townsend2010: Hi @rim99,

Ok, I follow what you are saying now and I agree with your assessment.  Thanks!

townsend2010:
Hi @rim99,

This is looking good.  I still need to test though.

One thing, in order to have a cleaner history, instead of merging `main` into this branch, it's better to rebase this branch on `main` and force push to avoid the merge commit.  Could you please do that?

rim99:
Hi @townsend2010 ,

I just rebased this branch, and squashed all commits together. Pleas take a look, thanks.

townsend2010:
Hi @rim99,

It wasn't necessary to squash all of the commits, but that's ok.  I'll get this reviewed as soon as possible.  Thanks!

townsend2010 (changes requested):
Seems a commit was lost as the change I suggested and you fixed to prepend `std::` to `optional` is no longer there and breaking the build.

See https://github.com/canonical/multipass/runs/7546991575?check_suite_focus=true#step:11:4307.

rim99:
Cool, just fixed this.

townsend2010:
bors try

townsend2010 (approved):
Ok, I think this is good.  Thanks!

bors merge

PR #2639, opened 2022-06-29
Recursive transfer

### LINKED ISSUE #1356: `transfer` missing recursive operation
I wanted to transfer in a directory hierarchy. But the transfer command does not take a -r flag.

_Originally posted by @robogeek in https://github.com/canonical/multipass/issues/1355_

andrei-toterman:
This PR adds the `-r` flag to the `transfer` command, allowing the transfer of directories

fix #1356

### DIFF
--- a/include/multipass/format.h
+++ b/include/multipass/format.h
@@ -18,10 +18,12 @@
 #ifndef MULTIPASS_FORMAT_H
 #define MULTIPASS_FORMAT_H
 
-#include <QString>
-#include <filesystem>
 #include <fmt/format.h>
 #include <fmt/ostream.h>
+
+#include <filesystem>
+
+#include <QString>
 
 namespace fmt
 {
--- a/include/multipass/format.h
+++ b/include/multipass/format.h
@@ -44,6 +44,7 @@ struct formatter<QString>
     }
 };
 
+// TODO this will be included in fmtlib v9
 template <>
 struct formatter<std::filesystem::path>
 {
--- a/include/multipass/format.h
+++ b/include/multipass/format.h
@@ -18,10 +18,10 @@
 #ifndef MULTIPASS_FORMAT_H
 #define MULTIPASS_FORMAT_H
 
+#include <QString>
+#include <filesystem>
 #include <fmt/format.h>
 #include <fmt/ostream.h>
-#include <filesystem>
-#include <QString>
 
 namespace fmt
 {
@@ -59,6 +59,5 @@ struct formatter<std::filesystem::path>
 };
 
 } // namespace fmt
-
 
 #endif // MULTIPASS_FORMAT_H
--- a/include/multipass/format.h
+++ b/include/multipass/format.h
@@ -20,6 +20,7 @@
 
 #include <fmt/format.h>
 #include <fmt/ostream.h>
+#include <filesystem>
 #include <QString>
 
 namespace fmt
@@ -38,6 +39,22 @@ struct formatter<QString>
     auto format(const QString& a, FormatContext& ctx)
     {
         return format_to(ctx.out(), "{}", a.toStdString()); // TODO: remove the copy?
+    }
+};
+
+template <>
+struct formatter<std::filesystem::path>
+{
+    template <typename ParseContext>
+    constexpr auto parse(ParseContext& ctx)
+    {
+        return ctx.begin();
+    }
+
+    template <typename FormatContext>
+    auto format(const std::filesystem::path& path, FormatContext& ctx)
+    {
+        return format_to(ctx.out(), "'{}'", path.string());
     }
 };
 
--- a/include/multipass/recursive_dir_iterator.h
+++ b/include/multipass/recursive_dir_iterator.h
@@ -1,0 +1,250 @@
+/*
+ * Copyright (C) 2021-2022 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef RECURSIVE_DIR_ITERATOR_H
+#define RECURSIVE_DIR_ITERATOR_H
+
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+namespace multipass
+{
+class DirectoryEntry
+{
+    fs::directory_entry entry;
+
+public:
+    DirectoryEntry() = default;
+    DirectoryEntry(const DirectoryEntry&) = default;
+    DirectoryEntry(DirectoryEntry&&) = default;
+    DirectoryEntry(const fs::directory_entry& _entry) : entry{_entry}
+    {
+    }
+
+    virtual DirectoryEntry& operator=(const DirectoryEntry&) = default;
+    virtual DirectoryEntry& operator=(DirectoryEntry&&) noexcept = default;
+
+    virtual ~DirectoryEntry() = default;
+
+    virtual void assign(const fs::path& path)
+    {
+        entry.assign(path);
+    }
+
+    virtual void assign(const fs::path& path, std::error_code& err)
+    {
+        entry.assign(path, err);
+    }
+
+    virtual void replace_filename(const fs::path& path)
+    {
+        entry.replace_filename(path);
+    }
+
+    virtual void replace_filename(const fs::path& path, std::error_code& err)
+    {
+        entry.replace_filename(path, err);
+    }
+
+    virtual void refresh()
+    {
+        entry.refresh();
+    }
+
+    virtual void refresh(std::error_code& err) noexcept
+    {
+        entry.refresh(err);
+    }
+
+    virtual const fs::path& path() const noexcept
+    {
+        return entry.path();
+    }
+
+    virtual bool exists() const
+    {
+        return entry.exists();
+    }
+
+    virtual bool exists(std::error_code& err) const noexcept
+    {
+        return entry.exists(err);
+    }
+
+    virtual bool is_block_file() const
+    {
+        return entry.is_block_file();
+    }
+
+    virtual bool is_block_file(std::error_code& err) const noexcept
+    {
+        return entry.is_block_file(err);
+    }
+
+    virtual bool is_character_file() const
+    {
+        return entry.is_character_file();
+    }
+
+    virtual bool is_character_file(std::error_code& err) const noexcept
+    {
+        return entry.is_character_file(err);
+    }
+
+    virtual bool is_directory() const
+    {
+        return entry.is_directory();
+    }
+
+    virtual bool is_directory(std::error_code& err) const noexcept
+    {
+        return entry.is_directory(err);
+    }
+
+    virtual bool is_fifo() const
+    {
+        return entry.is_fifo();
+    }
+
+    virtual bool is_fifo(std::error_code& err) const noexcept
+    {
+        return entry.is_fifo(err);
+    }
+
+    virtual bool is_other() const
+    {
+        return entry.is_other();
+    }
+
+    virtual bool is_other(std::error_code& err) const noexcept
+    {
+        return entry.is_other(err);
+    }
+
+    virtual bool is_regular_file() const
+    {
+        return entry.is_regular_file();
+    }
+
+    virtual bool is_regular_file(std::error_code& err) const noexcept
+    {
+        return entry.is_regular_file(err);
+    }
+
+    virtual bool is_socket() const
+    {
+        return entry.is_socket();
+    }
+
+    virtual bool is_socket(std::error_code& err) const noexcept
+    {
+        return entry.is_socket(err);
+    }
+
+    virtual bool is_symlink() const
+    {
+        return entry.is_symlink();
+    }
+
+    virtual bool is_symlink(std::error_code& err) const noexcept
+    {
+        return entry.is_symlink(err);
+    }
+
+    virtual uintmax_t file_size() const
+    {
+        return entry.file_size();
+    }
+
+    virtual uintmax_t file_size(std::error_code& err) const noexcept
+    {
+        return entry.file_size(err);
+    }
+
+    virtual uintmax_t hard_link_count() const
+    {
+        return entry.hard_link_count();
+    }
+
+    virtual uintmax_t hard_link_count(std::error_code& err) const noexcept
+    {
+        return entry.hard_link_count(err);
+    }
+
+    virtual fs::file_time_type last_write_time() const
+    {
+        return entry.last_write_time();
+    }
+
+    virtual fs::file_time_type last_write_time(std::error_code& err) const noexcept
+    {
+        return entry.last_write_time(err);
+    }
+
+    virtual fs::file_status status() const
+    {
+        return entry.status();
+    }
+
+    virtual fs::file_status status(std::error_code& err) const noexcept
+    {
+        return entry.status(err);
+    }
+
+    virtual fs::file_status symlink_status() const
+    {
+        return entry.symlink_status();
+    }
+
+    virtual fs::file_status symlink_status(std::error_code& err) const noexcept
+    {
+        return entry.symlink_status(err);
+    }
+
+    virtual bool operator==(const DirectoryEntry& rhs) const noexcept
+    {
+        return entry == rhs.entry;
+    }
+};
+
+class RecursiveDirIterator
+{
+    fs::recursive_directory_iterator iter;
+    DirectoryEntry current;
+
+public:
+    RecursiveDirIterator() = default;
+    RecursiveDirIterator(const fs::path& path, std::error_code& err) : iter{path, err}
+    {
+    }
+
+    virtual bool hasNext()
+    {
+        return iter != fs::end(iter);
+    }
+
+    virtual const DirectoryEntry& next()
+    {
+        return current = *iter++;
+    }
+
+    virtual ~RecursiveDirIterator() = default;
+};
+} // namespace multipass
+
+#endif // RECURSIVE_DIR_ITERATOR_H
--- a/include/multipass/recursive_dir_iterator.h
+++ b/include/multipass/recursive_dir_iterator.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2022 Canonical, Ltd.
+ * Copyright (C) 2022 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/include/multipass/recursive_dir_iterator.h
+++ b/include/multipass/recursive_dir_iterator.h
@@ -20,14 +20,13 @@
 
 #include <filesystem>
 
-namespace fs = std::filesystem;
-
 namespace multipass
 {
+namespace fs = std::filesystem;
+
+// this is just a thin wrapper around std::filesystem::directory_entry, used for mocking purposes
 class DirectoryEntry
 {
-    fs::directory_entry entry;
-
 public:
     DirectoryEntry() = default;
     DirectoryEntry(const DirectoryEntry&) = default;
@@ -36,8 +35,8 @@ public:
     {
     }
 
-    virtual DirectoryEntry& operator=(const DirectoryEntry&) = default;
-    virtual DirectoryEntry& operator=(DirectoryEntry&&) noexcept = default;
+    DirectoryEntry& operator=(const DirectoryEntry&) = default;
+    DirectoryEntry& operator=(DirectoryEntry&&) noexcept = default;
 
     virtual ~DirectoryEntry() = default;
 
@@ -220,13 +219,14 @@ public:
     {
         return entry == rhs.entry;
     }
+
+private:
+    fs::directory_entry entry;
 };
 
+// wrapper class around std::filesystem::recursive_directory_iterator used for mocking purposes
 class RecursiveDirIterator
 {
-    fs::recursive_directory_iterator iter;
-    DirectoryEntry current;
-
 public:
     RecursiveDirIterator() = default;
     RecursiveDirIterator(const fs::path& path, std::error_code& err) : iter{path, err}
@@ -244,6 +244,10 @@ public:
     }
 
     virtual ~RecursiveDirIterator() = default;
+
+private:
+    fs::recursive_directory_iterator iter;
+    DirectoryEntry current;
 };
 } // namespace multipass
 
--- a/include/multipass/ssh/sftp_client.h
+++ b/include/multipass/ssh/sftp_client.h
@@ -22,6 +22,9 @@
 
 #include <libssh/sftp.h>
 
+#include <QFlags>
+#include <filesystem>
+#include <functional>
 #include <iostream>
 #include <memory>
 #include <string>
@@ -29,7 +32,22 @@
 namespace multipass
 {
 using SSHSessionUPtr = std::unique_ptr<SSHSession>;
-using SFTPSessionUPtr = std::unique_ptr<sftp_session_struct, void (*)(sftp_session)>;
+using SFTPSessionUPtr = std::unique_ptr<sftp_session_struct, std::function<void(sftp_session)>>;
+
+namespace fs = std::filesystem;
+
+enum TransferFlags
+{
+    Recursive = 1,
+};
+
+SFTPSessionUPtr make_sftp_session(ssh_session session);
+std::unique_ptr<std::ostream> open_write_file(const fs::path& path);
+std::unique_ptr<std::istream> open_read_file(const fs::path& path);
+std::string get_full_file_target(const fs::path& source_path, const fs::path& target_path);
+std::string get_full_file_target(sftp_session sftp, const fs::path& source_path, const fs::path& target_path);
+fs::path get_full_dir_target(const fs::path& source_path, const fs::path& target_path);
+fs::path get_full_dir_target(sftp_session sftp, const fs::path& source_path, const fs::path& target_path);
 
 class SFTPClient
 {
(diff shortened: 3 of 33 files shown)


townsend2010:
Kicking this to try to get a clean build on the Windows/macOS platforms.

townsend2010 (changes requested):
This is not building successfully on macOS.  Here is an example: https://github.com/canonical/multipass-private/runs/7413114751?check_suite_focus=true#step:23:3918

townsend2010:
Hey @andrei-toterman!

Thanks for this.  I've done an initial pass and will provide some feedback, but before I do that, I'm just wondering if it would be possible to add some simple tests for the implementations in `file_ops.cpp` and `recursive_dir_iterator.h`?  Let me know if you need any help with that.

andrei-toterman:
Hey @townsend2010!

The methods and classes in those files are basically just virtual wrappers around the `std::filesystem` functions and classes, so that I can mock them in tests. I'm not really sure how I could test those wrappers, so I would appreciate some help! :)

townsend2010:
For posterity, @andrei-toterman and I discussed the missing tests and we have a plan on covering those files.

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,54 @@
...
 namespace multipass
 {
 using SSHSessionUPtr = std::unique_ptr<SSHSession>;
-using SFTPSessionUPtr = std::unique_ptr<sftp_session_struct, void (*)(sftp_session)>;
+using SFTPSessionUPtr = std::unique_ptr<sftp_session_struct, std::function<void(sftp_session)>>;
+
+enum TransferFlags
+{
+    Recursive = 1,
+};
```
townsend2010, line 40: A single member `enum` seems a bit strange.  Any plans for other transfer types or perhaps this is just forward thinking for the future?

include/multipass/ssh/sftp_dir_iterator.h
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_SFTP_DIR_ITERATOR_H
+#define MULTIPASS_SFTP_DIR_ITERATOR_H
+
+#include <filesystem>
+#include <functional>
+#include <libssh/sftp.h>
```
townsend2010, line 23: It's preferred have all of the standard lib headers defined together with `#include <libssh/sftp.h>` defined first with a blank line below it before the standard lib block of headers.

include/multipass/ssh/sftp_dir_iterator.h
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: Just 2022 please.

include/multipass/ssh/sftp_utils.h
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: Idem - 2022 please.

include/multipass/ssh/sftp_dir_iterator.h
```diff
@@ -0,0 +1,53 @@
...
+    sftp_session sftp;
+    std::stack<SFTPDirUPtr, std::vector<SFTPDirUPtr>> dirs;
+    SFTPAttributesUPtr previous_attr;
+
+    void push_dir(const fs::path& path);
+
+public:
+    SFTPDirIterator() = default;
+    SFTPDirIterator(sftp_session sftp, const fs::path& path);
+    virtual bool hasNext();
```
townsend2010, line 46: Probably could make this a `const` method.

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,54 @@
...
+    SFTPClient() = default;
     SFTPClient(const std::string& host, int port, const std::string& username, const std::string& priv_key_blob);
     SFTPClient(SSHSessionUPtr ssh_session);
 
-    void push_file(const std::string& source_path, const std::string& destination_path);
-    void pull_file(const std::string& source_path, const std::string& destination_path);
-    void stream_file(const std::string& destination_path, std::istream& cin);
-    void stream_file(const std::string& source_path, std::ostream& cout);
+    virtual bool is_dir(const fs::path& path);
+    virtual bool push(const fs::path& source_path, const fs::path& target_path, QFlags<TransferFlags> flags,
```
townsend2010, line 52: Could `flags` be `const` here?

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,54 @@
...
+                      std::ostream& err_sink);
+    virtual bool pull(const fs::path& source_path, const fs::path& target_path, QFlags<TransferFlags> flags,
```
townsend2010, line 54: Idem - `const` `flags`

src/ssh/sftp_client.cpp
```diff
@@ -16,189 +16,309 @@
...
-
-namespace mp = multipass;
-
-namespace
-{
-// TODO: For push/pull, use actual file permissions
 constexpr int file_mode = 0664;
 constexpr auto max_transfer = 65536u;
-const std::string stream_file_name{"stream_output.dat"};
+const char* stream_file_name = "stream_output.dat";
```
townsend2010, line 32: Just curious why the type had to be changed here.

src/ssh/sftp_dir_iterator.cpp
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: Idem- 2022 please.

src/ssh/sftp_dir_iterator.cpp
```diff
@@ -0,0 +1,77 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <multipass/ssh/sftp_dir_iterator.h>
+
+#include <multipass/ssh/sftp_utils.h>
+
+#include <multipass/format.h>
```
townsend2010, line 22: Unless there is some issue on the order of `#include`'s, could you please group the `multipass` includes together, then a blank line and then the `stdexcept`?

src/ssh/sftp_dir_iterator.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <stdexcept>
+
+namespace multipass
+{
+void SFTPDirIterator::push_dir(const fs::path& path)
+{
+    auto dir = mp_sftp_opendir(sftp, path.u8string().c_str());
+    if (!dir)
+        throw std::runtime_error{
+            fmt::format("[sftp] cannot open remote directory {}: {}", path, ssh_get_error(sftp->session))};
```
townsend2010, line 32: I think a blank line here would help a little with readability.

src/ssh/sftp_utils.cpp
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: Idem- copyright year :slightly_smiling_face:

src/ssh/sftp_utils.cpp
```diff
@@ -0,0 +1,155 @@
...
+        throw std::runtime_error{
+            fmt::format("[sftp] cannot overwrite local non-directory {} with directory", target_path)};
+    else if (err)
+        throw std::runtime_error{fmt::format("[sftp] cannot access {}: {}", target_path, err.message())};
+
+    if (!MP_FILEOPS.exists(target_path, err))
+    {
+        if (!MP_FILEOPS.create_directory(target_path, err))
+            throw std::runtime_error{
+                fmt::format("[sftp] cannot create local directory {}: {}", target_path, err.message())};
```
townsend2010, line 97: A blank line here for readability please.

src/ssh/sftp_utils.cpp
```diff
@@ -0,0 +1,155 @@
...
+                fmt::format("[sftp] cannot create remote directory {}: {}", target_path, ssh_get_error(sftp->session))};
+
+        return target_path;
+    }
+
+    auto child_path = target_path / source_path.filename();
+    auto child_info = mp_sftp_stat(sftp, child_path.u8string().c_str());
+    if (child_info && child_info->type != SSH_FILEXFER_TYPE_DIRECTORY)
+        throw std::runtime_error{
+            fmt::format("[sftp] cannot overwrite remote non-directory {} with directory", child_path)};
```
townsend2010, line 136: Blank line please.

tests/mock_file_ops.h
```diff
@@ -52,6 +52,20 @@ class MockFileOps : public FileOps
     MOCK_CONST_METHOD3(open, void(std::fstream&, const char*, std::ios_base::openmode));
     MOCK_CONST_METHOD1(commit, bool(QSaveFile&));
 
+    MOCK_METHOD(std::unique_ptr<std::ostream>, open_write, (const fs::path& path), (override, const));
```
townsend2010, line 55: Nice, it's about time we start moving to `MOCK_METHOD`.  We should think about a PR that does that everywhere...

tests/mock_recursive_dir_iterator.h
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: Idem- copyright year please.

tests/mock_sftp_client.h
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: I won't even say it this time :wink:

tests/mock_sftp_dir_iterator.h
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (C) 2019-2022 Canonical, Ltd.
```
townsend2010, line 2: :copyright: year

tests/test_sftp_utils.cpp
```diff
@@ -0,0 +1,342 @@
+#include "common.h"
```
townsend2010, line 1: Idem - Missing copyright block...

tests/test_sftp_dir_iterator.cpp
```diff
@@ -0,0 +1,98 @@
+#include "common.h"
```
townsend2010, line 1: Missing the copyright block.

townsend2010 (changes requested):
Hey @andrei-toterman!

Whoa, this is a much, MUCH bigger task/PR than I ever imagined!  I've given my first pass of code review and will begin the functional testing.  Thanks for this!

include/multipass/ssh/sftp_client.h
andrei-toterman: yes, there will also be a `-p` flag PR after this one :smile:

include/multipass/ssh/sftp_dir_iterator.h
andrei-toterman: Oops, sorry, I didn't notice the year differences, so I just copy-pasted these from whichever file I found first :sweat_smile:

src/ssh/sftp_client.cpp
andrei-toterman: honestly, it didn't have to be changed, it just bothered me that the string was just long enough to not qualify for a small string optimization, and there was no point in having it allocated on the heap, but in the end i guess the compiler probably optimized the whole thing anyway. i'll change it back :)

src/ssh/sftp_dir_iterator.cpp
andrei-toterman: where exactly? after line 32 or 29?

include/multipass/ssh/sftp_client.h
townsend2010: Ok, cool, makes sense now. :slightly_smiling_face:

src/ssh/sftp_dir_iterator.cpp
townsend2010: Ugh, stupid Github comments don't make it clear where exactly the reviewer was referring to.  I'm talking about adding a blank line between line 32 and line 33.

ricab (commented):
Hi @andrei-toterman, before looking deeper into the code, I put this through some intensive testing. The good news is that it worked great on most tests! 

I found only a couple of minor wrinkles, but of minor significance. If you prefer we can open issues for those and follow up separately. I think that'd be preferable, to avoid making this PR any bigger.

Tests passed with flying colors:

- :heavy_check_mark: source cwd (.)
  &rarr; works as expected
- :heavy_check_mark: target cwd (.)
  &rarr; works as expected
- :heavy_check_mark: dir with .dot-files 
  &rarr; works as expected
- :heavy_check_mark: dir with FIFO
  &rarr; refuses to copy non-regular file, as expected
- :heavy_check_mark: verify user and group ownership
  &rarr; ubuntu/ubuntu, as expected
- :heavy_check_mark: overwrite with same dir
  &rarr; timestamp not updated -> same as cp, scp
- :heavy_check_mark: overwrite dir with more contents
  &rarr; works as expected
- :heavy_check_mark: overwrite dir with fewer contents
  &rarr; contents merged, not overwritten, same as cp, scp
- :heavy_check_mark: overwrite with modified contents
  &rarr; good, timestamp not updated -> same as cp, scp
- :heavy_check_mark: parent dir not there in target
  &rarr; fails as expected
- :heavy_check_mark: transfer dir without `-r`
  &rarr; fails as expected
- :heavy_check_mark: direct link to directory (i.e. `mp transfer -r link-to-dir instance:`)
  &rarr; link is resolved and target is copied

Wrinkles:

- :x: links within directory
  &rarr; copied as links, rather than target (this is unlike what `scp` does, and unlike `mp transfer` without `-r`):

```bash  
$ ls -la
total 8
drwxrwxr-x 2 ricab ricab 4096 ago 24 11:39 .
drwxr-xr-x 9 ricab ricab 4096 ago 24 11:38 ..
lrwxrwxrwx 1 ricab ricab    7 ago 24 11:39 asdf -> ../asdf
-rw-rw-r-- 1 ricab ricab    0 ago 23 17:23 fdsa
$ mp transfer asdf special-gryphon:
$ mp exec special-gryphon -- ls -la asdf
-rw-rw-r-- 1 ubuntu ubuntu 5 Aug 24 11:40 asdf
$ mp transfer -r . special-gryphon:test
$ mp exec special-gryphon -- ls -la test
total 8
drwxrwxr-x 2 ubuntu ubuntu 4096 Aug 24 11:41 .
drwxr-xr-x 5 ubuntu ubuntu 4096 Aug 24 11:41 ..
lrwxrwxrwx 1 ubuntu ubuntu    7 Aug 24 11:41 asdf -> ../asdf
-rw-rw-r-- 1 ubuntu ubuntu    0 Aug 24 11:41 fdsa
```

Note: we may need a way to handle recursion... `scp` blindly follows recursions and fails after a certain depth (exit code 1), but that may be because of an OS limit. If that's the case, it would be enough for us too.

- :x: verify permissions
  &rarr; dir permissions not being preserved, unlike `scp` and `tranfer` for regular files:

```bash
$ ll test
total 8
dr-x--x--x 2 ricab ricab 4096 ago 23 17:02 .
drwxr-xr-x 7 ricab ricab 4096 ago 23 17:02 ..
$ mp transfer -r test/ primary:.
$ mp exec -n primary -- ls -la test
total 8
drwxrwxr-x 2 ubuntu ubuntu 4096 Aug 23 17:03 .
drwxr-xr-x 8 ubuntu ubuntu 4096 Aug 23 17:03 ..
```

andrei-toterman:
Hi, @ricab!

Thanks for the thorough testing! Regarding the first issue with links being copied as links, I think we discussed this earlier in a daily standup and concluded that this might be the desired behavior. Indeed, `scp` blindly follows links up to a certain limit, then crashes, but `sftp` doesn't copy links at all, and `cp` copies the link as-is with the original target. And I think we decided to handle it like `cp` does. But we can reevaluate the options and decide again :smiley:

ricab (changes requested):
Hey @andrei-toterman, first of all, great code! Well thought, careful, precise. Good structure, encapsulated concepts and stylish tricks too :wink: And it worked really well on my testing. 

Although this was meant as a "lighter" review &ndash; I only glanced over portions of lower-level code and skipped the tests entirely &ndash; it is a big PR and my first review of your code, so I still have many comments with praise, questions and suggestions for improvement. Most of the requests inline are for polishing and detail-level sort of thing. However, there are a couple of areas that I think have some margin of improvement: exception handling and separation of interface  and implementation. 

Of course, I may also be missing and overlooking things, so please let me know what you think too!

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,55 @@
...
 public:
+    SFTPClient() = default;
     SFTPClient(const std::string& host, int port, const std::string& username, const std::string& priv_key_blob);
     SFTPClient(SSHSessionUPtr ssh_session);
 
-    void push_file(const std::string& source_path, const std::string& destination_path);
-    void pull_file(const std::string& source_path, const std::string& destination_path);
-    void stream_file(const std::string& destination_path, std::istream& cin);
-    void stream_file(const std::string& source_path, std::ostream& cout);
+    virtual bool is_dir(const fs::path& path);
```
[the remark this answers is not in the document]
andrei-toterman: It's for checking if a path is a directory inside an instance, not locally. Those other methods check if the given path is a directory on the user's machine, not inside the VM.

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,55 @@
...
+    virtual bool push(const fs::path& source_path, const fs::path& target_path, const QFlags<TransferFlags> flags,
+                      std::ostream& err_sink);
+    virtual bool pull(const fs::path& source_path, const fs::path& target_path, const QFlags<TransferFlags> flags,
```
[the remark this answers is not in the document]
andrei-toterman: Okay, so here it was a complicated situation.

I think that when an error occurs, the user should be notified, but the transfer should still keep going and copy as many files as it can. That's why I think we can't let these functions throw exceptions when an error occurs, since that would interrupt the whole transfer.

Another alternative would be to accumulate all errors inside the function and throw a big exception at the end with all the errors. But then the errors wouldn't be reported in real time, only when the whole transfer is finished.

My approach was to use the `err_sink` to immediately be able to report errors as they occur, then return a success/failure bool at the end. This was the best way I could think of to accomplish a system that doesn't quit at the first error and tries to copy as much as it can, but that also reports all errors live, not just at the end.

src/ssh/sftp_dir_iterator.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if (dirs.empty())
+        return std::move(previous_attr);
+
+    auto dir = dirs.top().get();
+    while (auto attr = mp_sftp_readdir(sftp, dir))
+    {
+        if (strcmp(attr->name, ".") == 0 || strcmp(attr->name, "..") == 0)
+            continue;
+
+        auto path = fmt::format("{}/{}", dir->name, attr->name);
```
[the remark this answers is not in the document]
andrei-toterman: It shouldn't be tied to the host OS. `SFTPDirIterator` iterates over the directories inside the VM, not locally, and the VM always runs Linux :smile: 
This was actually a big hassle to handle, as I initially used the `std::filesystem::path` default behavior everywhere, which always uses OS specific separators. On Windows, of course, this will create paths with '\', but then working with those paths inside the VM through SFTP wouldn't make much sense, since '\' will not be used as a path separator. That's why I had to go through the entire code again and find all paths that are going to be used inside the VM and make sure they always use '/'. Honestly sometimes I regret using `std::filesystem::path` instead of plain old `std::string` with '/', just because of this pesky little '\'...

include/multipass/recursive_dir_iterator.h
```diff
@@ -0,0 +1,250 @@
...
+{
+    fs::directory_entry entry;
+
+public:
+    DirectoryEntry() = default;
+    DirectoryEntry(const DirectoryEntry&) = default;
+    DirectoryEntry(DirectoryEntry&&) = default;
+    DirectoryEntry(const fs::directory_entry& _entry) : entry{_entry}
+    {
+    }
```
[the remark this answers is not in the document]
andrei-toterman: Sorry, can you please clarify? :sweat_smile:

include/multipass/format.h
```diff
@@ -41,7 +44,22 @@ struct formatter<QString>
     }
 };
 
-} // namespace fmt
+template <>
+struct formatter<std::filesystem::path>
```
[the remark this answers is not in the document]
andrei-toterman: I tried to upgrade to the version that adds `std::filesystem::path` support, but it created some problems elsewhere, particularly with `QByteArray`, so it wouldn't be a seamless transition. We shall then leave this for a future PR.

include/multipass/ssh/sftp_dir_iterator.h
```diff
@@ -0,0 +1,54 @@
...
+namespace multipass
+{
+using SFTPAttributesUPtr = std::unique_ptr<sftp_attributes_struct, std::function<void(sftp_attributes)>>;
+using SFTPDirUPtr = std::unique_ptr<sftp_dir_struct, std::function<int(sftp_dir)>>;
+
+class SFTPDirIterator
+{
+    sftp_session sftp;
+    std::stack<SFTPDirUPtr, std::vector<SFTPDirUPtr>> dirs;
+    SFTPAttributesUPtr previous_attr;
```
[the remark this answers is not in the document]
andrei-toterman: yeah, i get the confusion. my reasoning was that it holds the previously read attribute, since when returning from `next()`, we either read a new attribute and return the previously read one by swapping it with the newly read one, or we don't read any new attribute and just return the previously read one when we're at the end of a directory. but yeah `next_attribute` makes more sense, since it is always the next attribute that is going to be returned

include/multipass/ssh/sftp_utils.h
```diff
@@ -0,0 +1,62 @@
...
+
+struct SFTPUtils : public Singleton<SFTPUtils>
+{
+    SFTPUtils(const Singleton<SFTPUtils>::PrivatePass&) noexcept;
+
+    virtual fs::path get_full_local_file_target(const fs::path& source_path, const fs::path& target_path);
+    virtual fs::path get_full_remote_file_target(sftp_session sftp, const fs::path& source_path,
+                                                 const fs::path& target_path);
+    virtual fs::path get_full_local_dir_target(const fs::path& source_path, const fs::path& target_path);
+    virtual fs::path get_full_remote_dir_target(sftp_session sftp, const fs::path& source_path,
```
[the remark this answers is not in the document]
andrei-toterman: now that i think about it better, maybe just `get_{local,remote}_{file,dir}_path` is good enough, since there's noting really 'full' or 'complete' about the paths, they're just translated from one base path to the other

include/multipass/ssh/sftp_client.h
```diff
@@ -18,31 +18,55 @@
...
+    SFTPClient() = default;
     SFTPClient(const std::string& host, int port, const std::string& username, const std::string& priv_key_blob);
     SFTPClient(SSHSessionUPtr ssh_session);
 
-    void push_file(const std::string& source_path, const std::string& destination_path);
-    void pull_file(const std::string& source_path, const std::string& destination_path);
-    void stream_file(const std::string& destination_path, std::istream& cin);
-    void stream_file(const std::string& source_path, std::ostream& cout);
+    virtual bool is_dir(const fs::path& path);
+    virtual bool push(const fs::path& source_path, const fs::path& target_path, const QFlags<TransferFlags> flags,
```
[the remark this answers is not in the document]
andrei-toterman: [Let the fight begin](https://github.com/canonical/multipass/pull/2639#discussion_r944659578) :triumph:

include/multipass/format.h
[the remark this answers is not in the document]
ricab: Sounds good. Can you just add the TODO then, please?

include/multipass/ssh/sftp_client.h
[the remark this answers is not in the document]
ricab: Duh :man_facepalming: OK, how about making that clearer in the name? `is_remote_dir` perhaps? Just to make life easier for distracted people like myself

src/ssh/sftp_dir_iterator.cpp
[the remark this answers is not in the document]
ricab: OK, don't mind me then. Ideally the SFTP code shouldn't assume the remote OS, but I guess that's fine in here. We can generalize if we ever have the need.

include/multipass/ssh/sftp_utils.h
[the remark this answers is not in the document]
ricab: OK, I will let you choose what you think is best then.

include/multipass/ssh/sftp_client.h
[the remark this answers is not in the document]
ricab: LOL. @townsend2010 calling your attention to this one ;) WDYT?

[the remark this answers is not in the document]
townsend2010: :wink: Ok, well, I did put my comment in question form :laughing: 

At any rate, I'm fine with whatever we use for the parameter **as long as** we keep `flags` from being modifiable in `push()` which is my main concern.

include/multipass/ssh/sftp_client.h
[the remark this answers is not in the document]
ricab: The point is that we only want to continue copying on expected errors, I believe. If there is an unexpected error, we should instead fail early and interrupt. Continuing normally on unexpected errors would make catching bugs that cascaded from them harder to understand and link to the initial issue. If, however, we find new error cases that we want to continue copying on, we can always start treating them as expected too, selectively. So, if this were critical software, I would say the "best effort" approach that you went for would be the best option. In our case, I think "fail early" is better.

As to how to achieve that, and aiming for simplicity (fewer parts to break, less code to understand, maintain...), I would avoid using Qt signals. It would be another mechanism that could fail when we were already handling a failure. And I think it would add unnecessary complexity. 

We have everything we need built in C++ itself. I tried to describe how in [other](https://github.com/canonical/multipass/pull/2639/#discussion_r959507521) [comments](https://github.com/canonical/multipass/pull/2639/#discussion_r959529597), probably not very clearly... The idea would be to throw dedicated exceptions (e.g. `SFTPException`) on problems that we can recover from and continue on. Then, the functions that now return `bool` would catch those dedicated exceptions, accumulate the relevant messages but otherwise continue with the copy, and then throw again when done (using `fmt::join`, to compose the final message, for example). Other exceptions would just be let through.

include/multipass/recursive_dir_iterator.h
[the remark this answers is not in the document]
ricab: Oops, just noticed this one. Forget about it, I must have read it as a copy constructor...

src/ssh/sftp_dir_iterator.cpp
```diff
@@ -70,8 +69,8 @@ SFTPAttributesUPtr SFTPDirIterator::next()
         return next();
     }
 
-    auto err_msg = fmt::format("[sftp] cannot read remote directory '{}': {}", dir->name, ssh_get_error(sftp->session));
+    SFTPError err{"cannot read remote directory '{}': {}", dir->name, ssh_get_error(sftp->session)};
     dirs.pop();
-    throw std::runtime_error{err_msg};
+    throw std::move(err);
```
ricab, line 74: Better remove the move here. The reasoning is similar to return statements: the standard already requires that err be treated as an rvalue in this case and, by using the cast, you prevent the compiler from entirely eliding the move because the operand of the throw exception is no longer an automatic local object.

ricab (approved):
Alright, LGTM! I have only a small detail inline, but otherwise good to go.

bors delegate+

andrei-toterman:
bors retry

townsend2010:
Ok, let's try again...

bors retry

PR #2647, opened 2022-07-01
[daemon] Move starting instances back to main thread

### LINKED ISSUE #2646: [regression] Errors in starting instances are not detected and start detection will continue
For example, allocating too much memory to an instance using the `qemu` driver causes `qemu` to error out, but the daemon does not detect this and will continue to keep trying to detect when the instance is ready.

townsend2010:
Fixes #2646

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1706,9 +1706,33 @@ try // clang-format on
         }
     }
 
+    fmt::memory_buffer start_errors;
+    for (const auto& name : vms)
+    {
+        auto& vm = vm_instances.find(name)->second;
+        auto state = vm->current_state();
+        if (state == VirtualMachine::State::unknown)
+        {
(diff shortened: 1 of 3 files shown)


luis4a0:
Hi @andrei-toterman, thanks for the thorough testing! You're right on libvirt this seems buggy, but this behavior was already present in 1.9.2. So this is not a regression, but a preexisting issue with libvirt. Thanks for going deep into this!

luis4a0:
@andrei-toterman @townsend2010, do you agree we merge this now and later investigate the issue with libvirt?

andrei-toterman:
@luis4a0 I agree. This still fixes the majority of cases

luis4a0:
Great @andrei-toterman, so we have two votes for merging. Let's go!

bors r+

luis4a0:
bors retry

luis4a0:
Something weird is happening here... bors is consistently timing out.

townsend2010:
One more time...

bors retry

PR #2652, opened 2022-07-12
Integrate aliases to Blueprints

luis4a0:
This PR adds the ability to add aliases to the Blueprints definition files, in this way:
```
aliases:
  <alias>: <name>:<command>
```


townsend2010 (changes requested):
Hey @luis4a0!

Good stuff here!  I've taken a cursory look over the code and don't see anything glaring, but I'll take a closer look very soon.

That said, there is one problem.  If you launch a Blueprint and pass `-n foo`, the instance name will become `foo`, but the alias creation does not take this into account.  `--name` will override the default instance name.

luis4a0:
Hey @townsend2010, thanks!

I'll take care of that.

src/client/cli/cmd/launch.cpp
```diff
@@ -435,6 +436,23 @@ mp::ReturnCode cmd::Launch::request_launch(const ArgParser* parser)
         if (timer)
             timer->pause();
 
+        std::vector<std::string> warning_aliases;
+        auto n = reply.aliases_to_be_created_size();
+        for (auto i = 0; i < n; ++i)
```
townsend2010, line 441: Is it not possible to just iterate over `reply.aliases_to_be_created()` directly like:

`for (const auto& alias_to_be_created : reply.aliases_to_be_created()`

or similar?

townsend2010 (commented):
Looking good!  Inline question below.

src/client/cli/cmd/launch.cpp
luis4a0: Oh, I didn't know RPC messages could be iterated like that. I changed it, thanks!

src/client/cli/cmd/launch.cpp
```diff
@@ -435,6 +436,21 @@ mp::ReturnCode cmd::Launch::request_launch(const ArgParser* parser)
...
+                                             alias_to_be_created.working_directory()};
+            if (create_alias(aliases, alias_to_be_created.name(), alias_definition, cout, cerr) != ReturnCode::Ok)
+                warning_aliases.push_back(alias_to_be_created.name());
+        }
+
+        if (warning_aliases.size())
+            cout << fmt::format("Warning: unable to create {} {}.\n", warning_aliases.size() == 1 ? "alias" : "aliases",
+                                fmt::join(warning_aliases, ", "));
+        else
+            cout << "No aliases warning.\n";
```
townsend2010, line 452: Is this cruft left from testing?  It seems awkward otherwise :wink:

townsend2010 (changes requested):
Ok, one last comment inline and I believe this is good to go.  Thanks!

src/client/cli/cmd/launch.cpp
luis4a0: Yes, definitely. I removed it.

townsend2010 (approved):
Ok, let's do it!  Thanks!

bors merge

luis4a0:
bors retry

PR #2654, opened 2022-07-13
[mounts] Add mount handlers for different types mounts

townsend2010:
This is the refactoring and plumbing necessary to support different mount types.


andrei-toterman (commented):
Hey @townsend2010! Everything looks very good, just a few comments inline :)

src/daemon/daemon_config.h
```diff
@@ -56,6 +57,7 @@ struct DaemonConfig
     const std::shared_ptr<logging::MultiplexingLogger> logger;
     const std::unique_ptr<QNetworkProxy> network_proxy;
     const std::unique_ptr<VMBlueprintProvider> blueprint_provider;
+    const std::vector<std::unique_ptr<MountHandler>> mount_handlers;
```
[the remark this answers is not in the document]
townsend2010: Good idea!  I guess I was so focused on following the `std::vector<...>` that is defined elsewhere... :wink:

src/client/cli/cmd/mount.cpp
```diff
@@ -46,6 +49,18 @@ auto convert_id_for(const QString& id_string)
...
     return id;
 }
+
+auto checked_mount_type(const QString& type)
+{
+    if (type == "integrated")
+        return mp::MountRequest_MountType_INTEGRATED;
+
+    if (type == "native")
+        return mp::MountRequest_MountType_NATIVE;
```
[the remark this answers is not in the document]
townsend2010: Good idea, will do!

tests/test_sshfsmount.cpp
```diff
@@ -173,6 +174,10 @@ struct SshfsMount : public mp::test::SftpServerTest
         EXPECT_TRUE(next_expected_cmd == commands.end()) << "\"" << next_expected_cmd->first << "\" not executed";
     }
 
+    static void on_install()
+    {
+    }
+
```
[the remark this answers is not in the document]
townsend2010: I must have had some grandiose idea that didn't come to fruition :slightly_smiling_face: I'll remove it if I can't figure out my intent...

src/client/cli/cmd/mount.cpp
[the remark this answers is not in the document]
townsend2010: Done

tests/test_sshfsmount.cpp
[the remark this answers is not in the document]
townsend2010: Removed

src/daemon/daemon_config.h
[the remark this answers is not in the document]
townsend2010: Done

src/client/cli/cmd/mount.cpp
```diff
@@ -35,6 +36,8 @@ namespace cmd = multipass::cmd;
 namespace
 {
 constexpr auto category = "mount cmd";
+const QString default_mount_type{"integrated"};
+const QString native_mount_type{"native"};
```
townsend2010, line 40: Hey @luis4a0, @ricab, @andrei-toterman, and @sharder996,

What do you guys think of these names for the type passed into `mount`?  Since this is user facing, I'd like a consensus on this before it's committed.

luis4a0: Hey @townsend2010!

I'd go for something a bit different. `integrated` does not give for me the idea that it's the default in Multipass, I think `default` would be OK to confer to users the idea of method which was always there in Multipass.

OTOH, `native` seems good to me. Maybe `optimized` or `platform` would be a bit more clear.

townsend2010: Hey @luis4a0,

The notion of what is "default" may change as the native mounts mature and we want to make those default, so I don't think `default` should be a user facing type.  I think it should be a simple, descriptive name for each type although I'm not quite sure `integrated` conveys that, but I can't think of a term that does.  I'd like to avoid calling out what the underlying technology is too, so I'd like to avoid things such as `sshfs`, `smb`, etc.

ricab: I agree with both of you. "Default" may change, but `integrated` sounds a little opaque to me too. How about `portable` and `optimized`? Those would convey the main advantage of each option.

luis4a0: ++`portable`!

luis4a0: (well, the exclamation mark above is an exclamation mark, not a factorial sign)

sharder996: Generally, I think the original names are fine as they describe the technology and not a relative attribute of it. For that reason I would shy away from using `optimized` because `optimized` could have a different meaning depending on use cause or additional future options.

andrei-toterman: I tend to agree that `optimized` might be a little misleading, since, at least to me, it gives the impression that the underlying technology is the same, but it just has a better implementation. 
Since the end goal would be to hopefully make the performance mounts the default choice, how about:
for `sshfs`: `legacy`, since it's an older system, but it can still be used to get the job done. other alternatives: `primitive`, `basic`, `inherent`, `builtin`, `inbuilt`
for performance mounts: `native` and `platform` sound good, but `platform` is not an adjective, so it sounds kinda weird. alternatives: `express`, `specialized`, `specific`, `modern`, `targeted`

townsend2010: After some discussion on the daily sync, we settled on `classic` and `native` for the mount types.

townsend2010:
Hey @andrei-toterman,

I would appreciate it if you gave this another review pass and then if you are good with it, let's go ahead and get it in.  Thanks!

andrei-toterman (changes requested):

src/rpc/multipass.proto
```diff
@@ -259,15 +259,20 @@ message TargetPathInfo {
 }
 
 message MountRequest {
+    enum MountType {
+        INTEGRATED = 0;
```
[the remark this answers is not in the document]
townsend2010: Yep, missed that, thanks!  Now fixed:)

andrei-toterman (approved):
Very nice! Let's go!

bors merge

townsend2010:
Hey @andrei-toterman!

I made the `-t` flag hidden.  Could you please look over this one last time before bors'ing it (or provide feedback)?

andrei-toterman (approved):
Hey @townsend2010! Took another look at the whole thing just to be sure. It's all looking great, in my opinion. So let's give it a go once more!

bors merge

andrei-toterman:
bors retry

andrei-toterman:
@townsend2010 i think there is a conflict between public/private versions of `src/sshfs_mount/CMakeLists.txt`. Can you please take a look?

townsend2010:
Hey @andrei-toterman!

> Can you please take a look?

Bah, I tend to forget this fact, but when `bors` runs builds on the Windows/macOS side, it doesn't use the private counterpart branch, so when there is a conflict, we have to merge manually.  I will do that now.  Thanks!

townsend2010:
Actually, I will create the private PR and let it build just to be sure that side is good and then I'll merge this PR and that PR.

PR #2664, opened 2022-07-15
Fix execution of `sudo` through `exec` command

### LINKED ISSUE #2663: [regression] New `exec` working directory mapping does not take sudo into account
There are reports of v1.10.0 of Multipass [breaking Snapcraft builds](https://forum.snapcraft.io/t/snapcraft-failing/30877).

Looking at the error pasted in the output, it looks like `sudo`/`sudo -u foo` in the `exec` command is not being taken into account when executing the `cd /working/dir`.  I see two options here for this:
1. Just always prepend `sudo` to the `cd` command and be done with it. 
2. Look for `sudo`/`sudo -u foo` in the `exec` command and prepend that to the `cd` command when found.

Also, when this is fixed, we still need to ensure that Snapcraft builds Snaps fine with the working directory mapping enabled.

luis4a0:
Fixes #2663.

### DIFF
--- a/src/client/cli/cmd/exec.cpp
+++ b/src/client/cli/cmd/exec.cpp
@@ -169,7 +169,16 @@ mp::ReturnCode cmd::Exec::exec_success(const mp::SSHInfoReply& reply, const mp::
         std::vector<std::vector<std::string>> all_args;
         if (dir)
         {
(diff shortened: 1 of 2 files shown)


townsend2010 (approved):
Great, WFM, thanks!

bors merge

PR #2667, opened 2022-07-18
Linux qemu uefi

### LINKED ISSUE #1477: Boot instances using UEFI

I would like to test some of Ubuntu's UEFI configuration, but multipass still launches instances using GPT instead of UEFI.

Not exactly _mine_, but:

```
< ChrisTownsend> Odd_Bloke: After a bit of reading, we need to point qemu to the TianoCore bios in order to support that.
```

(An ideal implementation for my testing would let me select/modify the method used for boot.)

andrei-toterman:
Make QEMU boot using UEFI on Linux

fix #1477

### DIFF
--- a/tests/qemu/linux/test_qemu_platform_detail.cpp
+++ b/tests/qemu/linux/test_qemu_platform_detail.cpp
@@ -153,10 +153,20 @@ TEST_F(QemuPlatformDetail, platform_args_generate_net_resources_removes_works_as
     const auto platform_args = qemu_platform_detail.vm_platform_args(vm_desc);
 
     // Tests the order and correctness of the arguments returned
-    EXPECT_THAT(platform_args, ElementsAre("--enable-kvm", "-bios", _, "-cpu", "host", "-nic",
-                                           QString::fromStdString(fmt::format(
-                                               "tap,ifname={},script=no,downscript=no,model=virtio-net-pci,mac={}",
-                                               tap_name, vm_desc.default_mac_address))));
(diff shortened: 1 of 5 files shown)


luis4a0:
Hey @andrei-toterman, thanks!

Looks good to me. It works here. Minor comment inline.

tests/qemu/linux/test_qemu_platform_detail.cpp
```diff
@@ -153,12 +153,10 @@ TEST_F(QemuPlatformDetail, platform_args_generate_net_resources_removes_works_as
     const auto platform_args = qemu_platform_detail.vm_platform_args(vm_desc);
 
     // Tests the order and correctness of the arguments returned
-    std::vector<QString> expected_platform_args{
-        "--enable-kvm", "-cpu", "host", "-nic",
-        QString::fromStdString(fmt::format("tap,ifname={},script=no,downscript=no,model=virtio-net-pci,mac={}",
-                                           tap_name, vm_desc.default_mac_address))};
-
-    EXPECT_THAT(platform_args, ElementsAreArray(expected_platform_args));
+    EXPECT_THAT(platform_args, ElementsAre("--enable-kvm", "-bios", _, "-cpu", "host", "-nic",
```
luis4a0, line 156: Nice, but can you test that the wildcard parameter is the expected one on each platform?

luis4a0:
It seems Windows test timed out. I'll retry once.

bors retry

PR #2668, opened 2022-07-18
Change long option

### LINKED ISSUE #2490: `multipass launch` should allow `--memory` as a synonym of `--mem`
That would make it more consistent with the corresponding instance-mod key.

sharder996:
This PR changes the long option for memory allocation of the `launch` command from `--mem` to `--memory`. `--mem` is still supported for backwards compatibility, but will not be displayed in the help text. If both options are used, `--memory` will take precedence.

Fixes #2490 for original change request.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -236,7 +236,7 @@ _multipass_complete()
             opts="${opts} --all --purge"
         ;;
         "launch")
-            opts="${opts} --cpus --disk --mem --name --cloud-init --network --bridged --mount"
+            opts="${opts} --cpus --disk --memory --name --cloud-init --network --bridged --mount"
         ;;
         "mount")
             opts="${opts} --gid-map --uid-map"
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -204,12 +204,17 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                            "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
                                            min_disk_size, default_disk_size)),
         "disk", QString::fromUtf8(default_disk_size));
+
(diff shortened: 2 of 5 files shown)


luis4a0 (commented):
Hey @sharder996, thanks!

Looks good, hiding the going-to-be-deprecated option seems the best choice to me.

One detail, though: I think that specifying `--mem` and `--memory` toghether must make Multipass to fail.

luis4a0:
@sharder996 it seems you answered before I update my question. I didn't read that `--memory` takes precedence over `--mem` in the description, sorry for that.

My point is that, in case the user knows that there is a `--memory` parameter, we should fail saying that `--mem` must not be used. The point in leaving the old parameter is to allow for old scripts to be used, but I don't think we should permit using it interchangeably with the new parameter. I'm mentioning this in order to open a discussion, not yet to force a change in your code. What do you think?

townsend2010:
I think if both `--mem` and `--memory` are specified at the same time, that should be a failure since the user would be aware of `--memory`.

Also, to take this one step further, I think if `--mem` is only specified, a warning should be printed out saying `--mem` is being deprecated in favor of `--memory` and please update any scripts, etc.

luis4a0:
Hey @sharder996. Thanks, it's much better now!

A tiny bit of coverage remaining and we're done. Can you please add coverage to the error messages?

luis4a0:
Excellent, thanks @sharder996! A final detail: can you please change "in favor" to "in favour", so we use UK English in the output?

luis4a0:
bors retry

luis4a0:
bors retry

luis4a0:
bors retry

luis4a0:
Weird, as usual

bors retry

townsend2010:
We'll try this one more time since the other branch was out of sync...

bors retry

PR #2674, opened 2022-07-21
Remove `multipass/optional.h`

### LINKED ISSUE #2630: Remove multipass::optional
`std::optional` should be fully supported on all platforms now, so we should be able to remove `multipass::optional`.

sharder996:
`C++17` is now supported on all platforms so we can remove logic that uses `<experimental/optional>` for older versions. `std::optional` documentation [here](https://en.cppreference.com/w/cpp/utility/optional).

Fixes #2630

Private repo PR: canonical/multipass-private#470


sharder996:
Some linting changes in here as well as VSCode automatically lints the file when it is saved.

luis4a0 (approved):
Looks good, thanks!

bors r+

luis4a0:
bors retry

PR #2676, opened 2022-07-21
Fix repeated [UG]ID maps

### LINKED ISSUE #2665: [mounts] Repeated id/gid mappings
I'm seeing where repeated uid/gid mappings are occurring and seem to increase with instance stop/start cycles with mounts defined.

```
$ multipass info test-mount
Name:           test-mount
State:          Running
IPv4:           10.88.239.162
Release:        Ubuntu 22.04 LTS
Image hash:     d90ea6784789 (Ubuntu 22.04 LTS)
Load:           0.01 0.04 0.00
Disk usage:     1.9G out of 4.7G
Memory usage:   253.6M out of 1.9G
Mounts:         /home/townsend => Home
                    UID map: 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default
                    GID map: 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default, 1000:default
                               => test-mount
                    UID map: 1000:default, 1000:default, 1000:default, 1000:default
                    GID map: 1000:default, 1000:default, 1000:default, 1000:default
```

Yikes!

luis4a0:
This PR adds checks in three points of the execution to assure that no [UG]ID mappings are repeated. Namely, when the instance DB is read from file, when it is written to file and when the SSHFS server parses its arguments. Moreover, a bug in the latter scenario, giving an erroneous mappings check, was fixed.

Fixes #2665.

### DIFF
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -18,12 +18,33 @@
 #ifndef MULTIPASS_ID_MAPPINGS_H
 #define MULTIPASS_ID_MAPPINGS_H
 
+#include <multipass/format.h>
+#include <multipass/logging/log.h>
+
+#include <set>
 #include <utility>
 #include <vector>
+
+namespace mpl = multipass::logging;
 
 namespace multipass
 {
 typedef typename std::vector<std::pair<int, int>> id_mappings;
+
+inline id_mappings unique_id_mappings(const id_mappings& xid_mappings)
+{
+    id_mappings ret;
+    std::set<std::pair<int, int>> id_set;
+
+    for (const auto& id_map : xid_mappings)
+        if (id_set.insert(id_map).second)
+            ret.push_back(id_map);
+        else
+            mpl::log(mpl::Level::debug, "id_mappings",
+                     fmt::format("Not inserting repeated map {}:{}", id_map.first, id_map.second));
+
+    return ret;
 }
+} // namespace multipass
 
 #endif // MULTIPASS_ID_MAPPINGS_H
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -21,6 +21,8 @@
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 
+#include <algorithm>
+#include <iterator>
 #include <set>
 #include <utility>
 #include <vector>
@@ -36,12 +38,17 @@ inline id_mappings unique_id_mappings(const id_mappings& xid_mappings)
     id_mappings ret;
     std::set<std::pair<int, int>> id_set;
 
-    for (const auto& id_map : xid_mappings)
-        if (id_set.insert(id_map).second)
-            ret.push_back(id_map);
-        else
-            mpl::log(mpl::Level::debug, "id_mappings",
-                     fmt::format("Not inserting repeated map {}:{}", id_map.first, id_map.second));
+    auto is_mapping_repeated = [&id_set](const auto& m) {
+        if (id_set.insert(m).second)
+            return false;
+
+        mpl::log(mpl::Level::debug, "id_mappings", fmt::format("Dropping repeated mapping {}:{}", m.first, m.second));
+
+        return true;
+    };
+
+    std::remove_copy_if(xid_mappings.cbegin(), xid_mappings.cend(), std::back_insert_iterator(ret),
+                        is_mapping_repeated);
 
     return ret;
 }
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -18,6 +18,7 @@
 #ifndef MULTIPASS_UTILS_H
 #define MULTIPASS_UTILS_H
 
+#include <multipass/id_mappings.h>
 #include <multipass/logging/level.h>
 #include <multipass/path.h>
 #include <multipass/singleton.h>
@@ -114,6 +115,9 @@ std::string run_in_ssh_session(SSHSession& session, const std::string& cmd);
 // yaml helpers
 std::string emit_yaml(const YAML::Node& node);
 std::string emit_cloud_config(const YAML::Node& node);
+
+// [ug]id mappings helpers
+id_mappings unique_id_mappings(const id_mappings& xid_mappings);
 
 // enum helpers
 template <typename RegisteredQtEnum>
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -18,7 +18,6 @@
 #ifndef MULTIPASS_UTILS_H
 #define MULTIPASS_UTILS_H
 
-#include <multipass/id_mappings.h>
 #include <multipass/logging/level.h>
 #include <multipass/path.h>
 #include <multipass/singleton.h>
@@ -115,9 +114,6 @@ std::string run_in_ssh_session(SSHSession& session, const std::string& cmd);
 // yaml helpers
 std::string emit_yaml(const YAML::Node& node);
 std::string emit_cloud_config(const YAML::Node& node);
-
-// [ug]id mappings helpers
-id_mappings unique_id_mappings(const id_mappings& xid_mappings);
 
 // enum helpers
 template <typename RegisteredQtEnum>
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -42,7 +42,6 @@
 #include <fstream>
 #include <random>
 #include <regex>
-#include <set>
 #include <sstream>
 
 #include <openssl/evp.h>
@@ -630,22 +629,4 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
 std::string mp::utils::emit_cloud_config(const YAML::Node& node)
 {
     return fmt::format("#cloud-config\n{}\n", emit_yaml(node));
-}
-
-mp::id_mappings mp::utils::unique_id_mappings(const mp::id_mappings& xid_mappings)
-{
-    mp::id_mappings ret;
-    std::set<std::pair<int, int>> id_set;
-
-    for (const auto& id_map : xid_mappings)
-        if (id_set.insert(id_map).second)
-        {
-            ret.push_back(id_map);
-            mpl::log(mpl::Level::trace, category, fmt::format("Inserting map {}:{}", id_map.first, id_map.second));
-        }
-        else
-            mpl::log(mpl::Level::trace, category,
-                     fmt::format("Not inserting repeated map {}:{}", id_map.first, id_map.second));
-
-    return ret;
 }
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -43,6 +43,7 @@
 #include <random>
 #include <regex>
 #include <sstream>
+#include <unordered_set>
 
 #include <openssl/evp.h>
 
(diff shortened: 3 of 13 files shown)


ricab (changes requested):
Hi @luis4a0, Thank you for your efforts on this. 

I like the choice of using a data structure to ensure uniqueness, I agree that is the way to go :+1:  However, I don't understand why you use is as an accessory in a utility function rather than just replace the original vector. For example, I tested [this small diff](https://paste.ubuntu.com/p/y9XJXVH87V/) and it seems to achieve the same thing. Am I missing something?

With respect to choosing `unordered_set` (rather than `set`): you chose it because of complexity and perhaps cache misses, right? I agree it is very important to be aware of those :slightly_smiling_face: But the thing is... most of the time our mappings have only one element. They will never be big enough for any of that to matter. I wouldn't be surprised if `set` was actually faster for such small cases, but in any case, the difference is absolutely negligible and I am not sure it warrants more code for a dedicated hash function (more code to read, maintain, make sure it has no bugs...).

Finally, even though a set fixes the issue, it would be good to pinpoint where the problem was coming from. We could be overlooking something else too. Did you happen to find out yet?

luis4a0:
Hey @ricab, thanks for reviewing this!

I don't replace the original data structure because it is central for an ID map to keep the order in which it was entered by the user. Normally, when using one-way ID maps, we used to have a map, as you say. But when we introduced the bidirectional mapping, it became necessary to keep a vector, because conflicts may arise on which we have to answer with the first element we find. So unfortunately that is not an option.

Imagine there is are only two mappings: (A, B) and (A, C). The user specified them in this order. This is possible in our bidirectional mapping scheme: a file owned by A on the host will be mapped to B and not to C (because this pair was specified first). A file owned by B on the instance will be mapped to being owned by A on the host. And a file owned by C on the instance will be owned by A on the host. In this simple example, the order is important, if we store the mappings in a set we will have an abiguity to answer who owns on the instance a file owned by A on the host.

Your second point, about choosing `unordered_set` I must say you're right. I'll change that.

About your last point, about the actual bug we are fixing, I am not sure what happens. The point is that the JSON file containing the mappings gets sometimes corrupted. I think (not 100% sure, and I don't use to blame others, but here we go) something happens with `QJsonObject` iterators.

ricab:
Ohh I see, very good point @luis4a0. So I was indeed missing something! :wink:  

OK, let's keep the vector then. Ideally we'd have a data-structure that did both (ensure uniqueness and maintain insertion order), but failing that, we could place the utility function together with the typedef, in `id_mappings.h` (we would rarely need one without the other). WDYT?

luis4a0:
Sure, a vector isn't the best data structure for that in terms of complexity, but we discussed with Saviq at the moment and we decided to keep it simple, given that the data structure will never hold an enormous amount of data such as to pay off the effort of implementing something complicated.

You're right the check function could be in `id_mappings.h`. You mean creating the corresponding `id_mappings.cpp`, right?

ricab:
> Sure, a vector isn't the best data structure for that in terms of complexity, but we discussed with Saviq at the moment and we decided to keep it simple, given that the data structure will never hold an enormous amount of data such as to pay off the effort of implementing something complicated.

The main gain would be not having to worry about uniqueness in client code. But I didn't find any such structure in the libs we already use and it's not worth rolling one ourselves.

> You're right the check function could be in `id_mappings.h`. You mean creating the corresponding `id_mappings.cpp`, right?

As you prefer. If you think it is small enough to inline, that's fine too.

src/utils/utils.cpp
```diff
@@ -630,3 +641,21 @@ std::string mp::utils::emit_cloud_config(const YAML::Node& node)
...
+
+    for (const auto& id_map : xid_mappings)
+        if (id_set.insert(id_map).second)
+        {
+            ret.push_back(id_map);
+            mpl::log(mpl::Level::trace, category, fmt::format("Inserting map {}:{}", id_map.first, id_map.second));
+        }
+        else
+            mpl::log(mpl::Level::trace, category,
+                     fmt::format("Not inserting repeated map {}:{}", id_map.first, id_map.second));
```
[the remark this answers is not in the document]
luis4a0: > As you prefer. If you think it is small enough to inline, that's fine too.

I think it's ok to leave the implementation in the header, because moving it to `src` would mean to make a refactor (this is used by the daemon, the SSHFS server and the client). That's why I thought at first it would fit in utils. I implemented it inside the header file.

> How about [std::remove_copy_if](https://en.cppreference.com/w/cpp/algorithm/remove_copy)? You could have a predicate that captured your id_set to do the same insert check.

Right, that will simplify the code.

> As for logging, I would drop the one in line 654 and promote the one in 657 to debug. WDYT?
Right, no need to declare that an ID is correct. I did the change.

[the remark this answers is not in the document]
ricab: Sounds good. Let me know when this is ready for a second pass.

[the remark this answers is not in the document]
luis4a0: @ricab thanks, I think it is ready. I used `std::remove_copy_if` as you suggested. It looks better, but the log message avoids us to make the function a one-liner :)

I opted for leaving the function in the header but, instead of inlining it, I used a template. WDYT?

include/multipass/id_mappings.h
```diff
@@ -18,12 +18,40 @@
...
 {
 typedef typename std::vector<std::pair<int, int>> id_mappings;
+
+template <class IDMap>
+id_mappings unique_id_mappings(const IDMap& xid_mappings)
+{
+    id_mappings ret;
+    std::set<std::pair<int, int>> id_set;
+
+    auto pred = [&id_set](const auto& m) {
```
ricab, line 42: Would you mind renaming to `is_mapping_repeated`?

include/multipass/id_mappings.h
```diff
@@ -18,12 +18,40 @@
...
+        if (id_set.insert(m).second)
+            return false;
+
+        mpl::log(mpl::Level::debug, "id_mappings", fmt::format("Not inserting repeated map {}:{}", m.first, m.second));
```
ricab, line 46:
```suggestion
        mpl::log(mpl::Level::debug, "id_mappings", fmt::format("Dropping repeated mapping {}:{}", m.first, m.second));
```

include/multipass/id_mappings.h
```diff
@@ -18,12 +18,40 @@
...
+
+        return true;
+    };
+
+    std::remove_copy_if(xid_mappings.cbegin(), xid_mappings.cend(), std::back_insert_iterator(ret), pred);
```
ricab, line 51: :+1:

include/multipass/utils.h
```diff
@@ -18,6 +18,7 @@
 #ifndef MULTIPASS_UTILS_H
 #define MULTIPASS_UTILS_H
 
+#include <multipass/id_mappings.h>
```
ricab, line 21: Forgotten include.

src/utils/utils.cpp
```diff
@@ -42,6 +42,7 @@
 #include <fstream>
 #include <random>
 #include <regex>
+#include <set>
```
ricab, line 45: And here :)

include/multipass/id_mappings.h
```diff
@@ -18,12 +18,40 @@
...
 #include <vector>
 
+namespace mpl = multipass::logging;
+
 namespace multipass
 {
 typedef typename std::vector<std::pair<int, int>> id_mappings;
+
+template <class IDMap>
+id_mappings unique_id_mappings(const IDMap& xid_mappings)
```
ricab, line 37: Why did you choose a template rather than inline? This is very clearly bound to the `typedef` above, right?

ricab (changes requested):
Hey @luis4a0, I was about to say the PR looked good and I only had a few details inline... when I actually went to test it.

The thing is, whatever mechanism is duplicating mappings is also copying mappings from instance to instance. For instance, running with vanilla 1.10.1, I launched `primary` (which automatically mounts home) and then mounted something manually: `mp mount -g 123:321 -g 321:123 -g 123:321 -g 321:321 -g 321:123 -u 234:432 -u 432:432 -u 432:432 -u 234:432 tmp primary`. I got:

```
$ mp info primary
Name:           primary
State:          Running
IPv4:           10.198.95.120
Release:        Ubuntu 20.04.4 LTS
Image hash:     692406940d6a (Ubuntu 20.04 LTS)
Load:           0.07 0.12 0.08
Disk usage:     1.4G out of 4.7G
Memory usage:   161.3M out of 976.8M
Mounts:         /home/ricab/tmp => /home/ricab/tmp
                    UID map: 234:432, 432:432, 432:432, 234:432
                    GID map: 123:321, 321:123, 123:321, 321:321, 321:123
                /home/ricab     => Home
                    UID map: 1000:default
                    GID map: 1000:default
```

As expected. However, after a few daemon restarts:

```
$ mp info primary
Name:           primary
State:          Running
IPv4:           10.198.95.120
Release:        Ubuntu 20.04.4 LTS
Image hash:     692406940d6a (Ubuntu 20.04 LTS)
Load:           0.84 0.31 0.15
Disk usage:     1.4G out of 4.7G
Memory usage:   158.8M out of 976.8M
Mounts:         /home/ricab     => Home
                    UID map: 234:432, 432:432, 432:432, 234:432, 1000:default, 234:432, 432:432, 432:432, 234:432, 234:432, 432:432, 432:432, 234:432, 1000:default
                    GID map: 123:321, 321:123, 123:321, 321:321, 321:123, 1000:default, 123:321, 321:123, 123:321, 321:321, 321:123, 123:321, 321:123, 123:321, 321:321, 321:123, 1000:default
                /home/ricab/tmp => /home/ricab/tmp
                    UID map: 234:432, 432:432, 432:432, 234:432, 1000:default, 234:432, 432:432, 432:432, 234:432
                    GID map: 123:321, 321:123, 123:321, 321:321, 321:123, 1000:default, 123:321, 321:123, 123:321, 321:321, 321:123
```

Switching to your PR, I get the repetitions filtered, but the mappings for the second mount were still transported to the first one:

```
ricab@ricab-xps13:~ $ snap refresh --channel edge/pr2676 multipass
multipass (edge/pr2676) 1.11.0-dev.56.pr2676+ged8f45c5 from Canonical✓ refreshed
ricab@ricab-xps13:~ $ mp info primary
Name:           primary
State:          Running
IPv4:           10.198.95.120
Release:        Ubuntu 20.04.4 LTS
Image hash:     692406940d6a (Ubuntu 20.04 LTS)
CPU(s):         1
Load:           0.20 0.23 0.13
Disk usage:     1.4G out of 4.7G
Memory usage:   159.0M out of 976.8M
Mounts:         /home/ricab/tmp => /home/ricab/tmp
                    UID map: 234:432, 432:432, 1000:default
                    GID map: 123:321, 321:123, 321:321, 1000:default
                /home/ricab     => Home
                    UID map: 234:432, 432:432, 1000:default
                    GID map: 123:321, 321:123, 321:321, 1000:default
```

I really think we need to pinpoint where the issue is coming from and fix it there. That will probably render `unique_id_mappings` obsolete. If so, feel free to ignore the corresponding inline comments.

luis4a0:
Hey @ricab, thanks! That's astonishing! It worked on my tests, I'll try to reproduce it here.

ricab:
OK. Let me know what you find!

include/multipass/id_mappings.h
luis4a0: Done.

include/multipass/id_mappings.h
luis4a0: Done.

include/multipass/utils.h
luis4a0: Done.

src/utils/utils.cpp
luis4a0: Done.

include/multipass/id_mappings.h
luis4a0: Changed to inline.

ricab (approved):
Alright, looks good!

bors r+

ricab:
Bahh, looks like outdated certificates on cloudbase.it nodes again... I can get to it here, hopefully bors will too in a couple retries

bors retry

PR #2684, opened 2022-07-28
Fix LXD not updating state

### LINKED ISSUE #2672: [lxd] Multipass not reacting correctly with internally shut down instances
> if you don't use multipass stop, then it's really hard for multipassd to know the state of a VM. That said, we could probably be better at detecting instance states that are modified outside of Multipass when using the LXD driver.

@townsend2010 are you aware that `multipass stop --all` is not enough to prevent instances from auto-booting the next time the host boots?

Example:
- An instance is powered off internally, using a command like `poweroff`
- `multipass stop --all` is run on the host machine.
- Host machine is rebooted, and the instance (unfortunately!) auto-boots, despite both above attempts to prevent that :/

(This should be easily fixable, no?)

PS This is separate from the question as to whether or not Multipass will one day act like a modern BIOS:

> there really should be a Multipass/LXD option (rather like in a PC's BIOS/firmware) allowing you to toggle between either of these 2 choices:
>
> 1. Auto-boot VM's/instances after power failure / glitch / etc.
> 2. Do NOT auto-boot VM's/instances after power failure / glitch / etc.

_Originally posted by @holta in https://github.com/canonical/multipass/issues/2610#issuecomment-1189353167_

sharder996:
LXD instances do not update instance state within multipass when they are externally shutdown (ie. `sudo poweroff` within the instance) and so will reboot them when the daemon restarts. To fix this, we force multipass to update the state of LXD instances before shutting down to ensure that they are put into the correct state when multipass reboots.

Fixes #2672

### DIFF
--- a/src/platform/backends/lxd/lxd_virtual_machine.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine.cpp
@@ -196,6 +196,7 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
     update_shutdown_status = false;
 
+    current_state();
     if (state == State::running)
     {
         try
@@ -207,6 +208,10 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
         {
             stop();
         }
+    }
+    else
+    {
+        update_state();
     }
 }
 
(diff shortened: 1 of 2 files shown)


andrei-toterman (approved):

andrei-toterman:
thanks, @sharder996! looks good to me.

bors merge

andrei-toterman:
dammit

bors retry

andrei-toterman:
:roll_eyes: 

bors retry

townsend2010:
And `bors` was doing so well before this...

bors retry

PR #2685, opened 2022-07-29
[cli] Report memory sizes consistently

ricab:
I noticed that there was another calculation for human readable memory sizes in the table formatter. Dropping it for consistent output across Multipass (and less code to maintain).


sharder996:
LGTM

bors merge

PR #2693, opened 2022-08-03
Add `force` option to lxd `stop` request on error

### LINKED ISSUE #2492: Multipass can't delete LXD instances that go into Error state

sharder996:
If `lxd` fails to stop an instance gracefully, retry the request with the `force` option applied.

Fixes #2492


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -378,9 +386,9 @@ const QUrl mp::LXDVirtualMachine::network_leases_url()
     return base_url.toString() + "/networks/" + bridge_name + "/leases";
 }
 
-void mp::LXDVirtualMachine::request_state(const QString& new_state)
+void mp::LXDVirtualMachine::request_state(const QString& new_state, const bool force)
```
andrei-toterman, line 389: i have mixed feelings about this. on one hand, it's simple and it gets the job done and it doesn't interfere with the other uses of this function. but on the other hand, the other possible `new_state`s do not take a `force` option, which would make possible future uses of this function a little confusing, and it would also make it more awkward if we ever need to pass different flags to different actions.

one possible alternative would be to pass a `const QJsonObject& extra_fields = {}` instead of a `const bool force` and merge the two json objects, then call the method as `request_state("stop", {{"force", true}})`. this would allow for clearer and more extensive customization, although it may be too much for such simple use cases.

so, this can stay as is if consider the drawbacks to be negligible. what do you think?

sharder996: There aren't a large number of different arguments that could be passed to `lxd` but using an `args` parameter would make it look cleaner and enable future changes on our end without having to change this function.

https://github.com/lxc/lxd/blob/407205d4bd5d59e4a85d846988011a6fc09a9e72/doc/rest-api.yaml#L2072

andrei-toterman (commented):
Now that I look at it a little better, I don't think this approach will work nicely with #1909. The `LXDVirtualMachine` class inherits its `stop` function from the abstract `VirtualMachine` class, from which the other backends inherit as well. The virtual `stop` function should be able to take a `bool force` flag so that any backend could do a forced stop. But in this current implementation, the stop request is automatically forced if the normal one fails. I think this should be handled at a higher level, since maybe we don't want to shut down an instance by force no matter what, only in some cases, like when deleting and purging it.

PR #2705, opened 2022-08-15
[tests] Fix URLDownloader tests

townsend2010:
There is a race that has been showing up on the Windows build.


luis4a0 (approved):
Thanks @townsend2010! Increasing the test timeout seems to fix the race condition. If the problem arises again, we can try with a bigger value (10ms is still a very small value which won't compromise speed, in the future we can safely use 50 or 100ms).

townsend2010:
Ugh....

bors retry

luis4a0:
Well, at least it's not windows this time.

bors retry

townsend2010:
If bors fails this time, then let's merge manually.

PR #2711, opened 2022-08-18
Parallel instance starts on daemon startup

### LINKED ISSUE #2520: [qemu] daemon starting instance twice if triggered on startup
When an instance gets started on Multipass startup, but a client requests a start as well, daemon may try and start it twice:

```shell
$ multipass start
$ snap stop multipass
$ snap start multipass; while true; do multipass start; done
start failed: cannot connect to the multipass socket
# ...
start failed: cannot connect to the multipass socket
Starting primary \^C
```

Log excerpt:
```plain
Apr 12 13:19:47 nuc multipassd[320423]: primary needs starting. Starting now...
Apr 12 13:19:48 nuc multipassd[320423]: [322406] started: qemu-system-x86_64 --enable-kvm -cpu host -nic tap,ifname=tap-1e376645a40,script=no,downscript=no,model=virtio-net-pci,mac=52:54:00:be:37:3d -device virtio-scsi-pci,id
=scsi0 -drive file=/var/snap/multipass/common/data/multipassd/vault/instances/primary/ubuntu-20.04-server-cloudimg-amd64.img,if=none,format=qcow2,discard=unmap,id=hda -device scsi-hd,drive=hda,bus=scsi0.0 -smp 1 -m 1024M -qmp
 stdio -chardev null,id=char0 -serial chardev:char0 -nographic -cdrom /var/snap/multipass/common/data/multipassd/vault/instances/primary/cloud-init-config.iso -loadvm suspend -machine pc-i440fx-focal
Apr 12 13:19:48 nuc multipassd[320423]: Waiting for SSH to be up
Apr 12 13:19:48 nuc multipassd[320423]: QProcess: Destroyed while process ("qemu-system-x86_64") is still running.
Apr 12 13:19:48 nuc multipassd[320423]: QIODevice::read (QProcess): device not open
Apr 12 13:19:48 nuc multipassd[320423]: QMP:
Apr 12 13:19:48 nuc multipassd[320423]: QIODevice::read (QProcess): device not open
Apr 12 13:19:48 nuc multipassd[320423]: QMP:
Apr 12 13:20:18 nuc multipassd[320423]: process state changed to NotRunning
Apr 12 13:20:18 nuc multipassd[320423]: [322981] started: qemu-system-x86_64 -nographic -dump-vmstate /tmp/multipassd.wmJFCU
Apr 12 13:20:18 nuc multipassd[320423]: Applied AppArmor policy: multipass.primary.qemu-system-x86_64
Apr 12 13:20:18 nuc multipassd[320423]: process state changed to Starting
Apr 12 13:20:18 nuc multipassd[322984]: Caught an unhandled exception: Failed to apply AppArmor policy multipass.primary.qemu-system-x86_64: errno=2 (No such file or directory)
Apr 12 13:20:18 nuc multipassd[322984]: QSocketNotifier: Invalid socket 30 and type 'Read', disabling...
Apr 12 13:20:48 nuc multipassd[320423]: Qemu failed to start: program: qemu-system-x86_64; error: Process operation timed out
Apr 12 13:20:48 nuc multipassd[320423]: Caught an unhandled exception: failed to start qemu instance: program: qemu-system-x86_64; error: Process operation timed out
```

AFAICS it tried to dump vmstate on a running vm.

[multipassd.log](https://github.com/canonical/multipass/files/8473264/multipassd.log)

sharder996:
When trying to start an instance through the cli, a parallel instance start can occurs if done while the daemon is bootstrapping. This PR ensures that the instance state is changed at the beginning of process initialization rather than waiting for a signal handler. This ensures that process state remains accurate immediately after `vm_instance->start()` is called. An additional lock is also added to make sure that the checking of instance state and an instance start are treated as an atomic operation.

Fixes #2520


ricab:
Hey @sharder996 could you please rebase this? That would publish the snap again.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -261,6 +261,8 @@ void mp::QemuVirtualMachine::start()
     }
 
     vm_process->start();
+    state = VirtualMachine::State::starting;
```
ricab, line 264: Why did you need to move this here? We should be holding the new lock at this point and `on_started` would be called before returning from the next line, because it's [registered as a callback for when the process starts](https://github.com/canonical/multipass/blob/4b03174758438540a0b6d3ec2851c446e797d185/src/platform/backends/qemu/qemu_virtual_machine.cpp#L474). Unless it fails to start, in which case it should actually not be in the `starting` state, right?

ricab (changes requested):
Hey @sharder996, this works nicely and it does not delay being able to connect to daemon. Good job! I have a couple of minutia inline. And then only a missing piece: tests. Did you have trouble writing tests for this?

src/platform/backends/qemu/qemu_virtual_machine.cpp
sharder996: I lock itself was not able to solve the issue because the second `start()` call was happening before the callback `on_started()` was able to be called by the process created by the first `start()` call.

ricab: Hmm, I still don't get it. Can you offer an example of that? A stack trace would do (real or imaginary). I am probably missing something, but here's what I am seeing.

Until `on_started` is called, the main thread is still [waiting for it to start](https://github.com/canonical/multipass/pull/2711/files#diff-b70c5ab44722c02494a880259a18034fce5b14fb187dcc9830192a02d50078d4R266), and so still holding the lock. Notice that [`waitForStarted`](https://doc.qt.io/qt-5/qprocess.html#waitForStarted) *Blocks until the process has started and the `started()` signal has been emitted*. 

The `started()` signal is connected in the code I linked above with a [`DirectConnection`](https://doc.qt.io/qt-5/qt.html#ConnectionType-enum) (that's what `AutoConnection` means in this case since the receiving lambda is created when connecting). So, if the signal hasn't been handled, it hasn't been emitted. And we just saw that, if `started()` hasn't been emitted, `waitForStarted` hasn't returned yet (or there was an error).

sharder996: So, after re-testing this again it seems like the lock in of itself is enough to prevent concurrent `start()` commands. And, I would tend to agree with you in that `waitForStarted()` should block until the process has started, but thought that changing where the vm state is updated was necessary.

ricab (approved):
Alright, LGTM, thanks!

bors r+

PR #2713, opened 2022-08-19
Add workspaces to Blueprints

### LINKED ISSUE #2670: [blueprints] Support automatically creating a workspace and mounting it in instance
Blueprints need the ability automatically create a workspace and automatically mount that in the created instance.

The created workspace on the host should be in the form `$HOME\multipass\$instance-name` and should be mounted in the instance in the same manner.

The key in the yaml should look something like:
```yaml
instances:
  foo:
    workspace: true
```

luis4a0:
This PR adds the ability of creating a workspace to Blueprints. A workspace is a host folder in `${HOME}/multipass`, which is automatically mounted on the Blueprint instance.

A workspace is created by adding in the Blueprints YAML file the following:
```plain
instances:
  foo:
    workspace: true
```

Fixes #2670.


townsend2010 (approved):
LGTM, thanks!

bors merge

townsend2010:
Arghhh, `clang` is segfaulting more and more on the Intel Mac :frowning_face: 

bors retry

PR #2715, opened 2022-08-23
Ignore LXDNotFoundException on purge

### LINKED ISSUE #2714: Multipass crashes on `delete -p` when using the LXD driver
**Describe the bug**
Multipass crashes with an exception reaching a forbidden boundary when trying to purge-delete an instance. This happens after the instance has been successfully deleted in LXD and Multipass requests its state again. Presumably, that's coming out of [here](https://github.com/canonical/multipass/blob/main/src/platform/backends/lxd/lxd_virtual_machine.cpp#L199).

```
$ mp delete -p primary
delete failed: cannot connect to the multipass socket
```

And this in the log, after the instance is deleted:

```
ago 23 17:28:12 ricab-xps13 multipassd[77722]: Requesting LXD: GET unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0/virtual-machines/primary/state?project=multipass
ago 23 17:28:12 ricab-xps13 multipass.multipassd[77722]: terminate called after throwing an instance of 'multipass::LXDNotFoundException'
ago 23 17:28:12 ricab-xps13 multipass.multipassd[77722]:   what():  LXD object not found
```

**To Reproduce**
How, and what happened?
1. `multipass set local.driver=lxd`
1. `multipass launch`
1. `multipass delete -p <instance>`

sharder996:
Ignore `LXDNotFoundException` which is thrown when an instance has been purged. This isn't a particularly nice fix, but seems to be the simplest without a complete refactor of the LXD backend. Another potential way of accounting for this situation could be through the use of the `update_shutdown_status` boolean, but I felt like that would just add unnecessary complexity to an already complex state machine.

Fixes #2714

### DIFF
--- a/src/platform/backends/lxd/lxd_virtual_machine.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine.cpp
@@ -29,6 +29,7 @@
 #include <multipass/memory_size.h>
 #include <multipass/network_access_manager.h>
 #include <multipass/snap_utils.h>
+#include <multipass/top_catch_all.h>
 #include <multipass/utils.h>
 #include <multipass/virtual_machine_description.h>
 #include <multipass/vm_status_monitor.h>
@@ -196,30 +197,33 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
     update_shutdown_status = false;
 
-    try
-    {
-        current_state();
+    mp::top_catch_all(vm_name, [this]() {
+        try
+        {
+            current_state();
 
-        if (state == State::running)
-        {
-            try
+            if (state == State::running)
             {
-                if (!QFileInfo::exists(mp::utils::snap_common_dir() + "/snap_refresh"))
+                try
+                {
+                    if (!QFileInfo::exists(mp::utils::snap_common_dir() + "/snap_refresh"))
+                        stop();
+                }
+                catch (const mp::SnapEnvironmentException&)
+                {
                     stop();
+                }
             }
-            catch (const mp::SnapEnvironmentException&)
+            else
             {
-                stop();
+                update_state();
             }
         }
-        else
+        catch (const LXDNotFoundException& e)
         {
-            update_state();
+            mpl::log(mpl::Level::debug, vm_name, fmt::format("LXD object not found"));
         }
-    }
-    catch (const LXDNotFoundException& e)
-    {
-    }
+    });
 }
 
 void mp::LXDVirtualMachine::start()
--- a/src/platform/backends/lxd/lxd_virtual_machine.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine.cpp
@@ -196,22 +196,29 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
     update_shutdown_status = false;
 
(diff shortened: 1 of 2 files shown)


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -196,22 +196,29 @@ mp::LXDVirtualMachine::~LXDVirtualMachine()
 {
     update_shutdown_status = false;
 
-    current_state();
-    if (state == State::running)
+    try
```
ricab, line 199: An overall wrapping with `top_catch_all` would still be nice (in addition to the existing try/catches), to account for any other exceptions that could be thrown here and avoid terminating in the same way.

ricab (changes requested):
Hey @sharder996, thanks for the fix! I think you're right, this is what we need for the time being. Only a few requests inline.

ricab:
Hey @sharder996, I haven't looked back at this yet, but it seems it needs a rebase.

tests/lxd/test_lxd_backend.cpp
```diff
@@ -653,6 +654,65 @@ TEST_F(LXDBackend, machine_does_not_update_state_in_dtor)
     EXPECT_TRUE(stop_requested);
 }
 
+TEST_F(LXDBackend, machine_logs_not_found_exception_in_dtor)
```
ricab, line 657: Could you use camelCase for this? We have a bunch of snake_cased tests still, but we're trying to move from it because gtest advises against underscores in test names (it can mess with how it uses underscores to generate internal test names).

tests/lxd/test_lxd_backend.cpp
```diff
@@ -26,6 +26,7 @@
 #include "tests/mock_platform.h"
 #include "tests/mock_status_monitor.h"
 #include "tests/stub_status_monitor.h"
+#include "tests/stub_terminal.h"
```
ricab, line 29: Is this a leftover from development?

tests/lxd/test_lxd_backend.cpp
```diff
@@ -653,6 +654,65 @@ TEST_F(LXDBackend, machine_does_not_update_state_in_dtor)
...
+    {
+        mp::LXDVirtualMachine machine{
+            default_description, mock_monitor,        mock_network_access_manager.get(), base_url,
+            bridge_name,         default_storage_pool};
+        machine.shutdown();
+    }
+
+    EXPECT_TRUE(vm_shutdown);
+    EXPECT_TRUE(stop_requested);
+}
```
ricab, line 714: I generally request that functions be kept small and copy-pasting code be avoided (the two go hand in hand). 

However, lets leave it alone in here: you are just reproducing the pattern that is used throughout the file and extracting reusable lambda bits for mocking is not obvious and not for this PR.

ricab (commented):
Hey @sharder996, this is looking good now, I have only a couple of tiny requests before we ship it :wink: And I did not see any leaking mocks, BTW.

ricab (approved):
Alright, LGTM!

bors r+

ricab:
bors retry

PR #2717, opened 2022-08-24
Make core 20/22 not change IP address over reboot

### LINKED ISSUE #2690: Multipass fails to notice IP change over reboot
```plain
$ multipass launch http://cdimage.ubuntu.com/ubuntu-core/20/stable/current/ubuntu-core-20-amd64.img.xz

$ multipass ls
fluttering-jackal       Restarting        --               Not Available

$ multipass stop fluttering-jackal

$ multipass start fluttering-jackal

$ multipass ls
fluttering-jackal       Running           10.127.143.204   Not Available

$ multipass exec fluttering-jackal grep PRETTY /etc/os-release
PRETTY_NAME="Ubuntu Core 20"
```

The problem is that the machine changes client-id over reboots, so DNSMasq gives up a new IP and Multipass fails to notice:
```plain
$ journalctl -u snap.multipass.multipassd.service --since -1h |grep DHCPACK
sie 01 15:27:08 michal-laptop dnsmasq-dhcp[10199]: DHCPACK(mpqemubr0) 10.127.143.203 52:54:00:59:fc:d9 ubuntu
sie 01 15:28:20 michal-laptop dnsmasq-dhcp[10199]: DHCPACK(mpqemubr0) 10.127.143.204 52:54:00:59:fc:d9 fluttering-jackal
sie 01 15:32:55 michal-laptop dnsmasq-dhcp[10199]: DHCPACK(mpqemubr0) 10.127.143.204 52:54:00:59:fc:d9 fluttering-jackal
sie 01 15:34:57 michal-laptop dnsmasq-dhcp[10199]: DHCPACK(mpqemubr0) 10.127.143.204 52:54:00:59:fc:d9 fluttering-jackal
```

andrei-toterman:
This PR makes it so that when a Linux QEMU instance reboots, it checks the `dnsmasq.leases` file for leases that have the same MAC address as the instance but under a different host name and releases them. This is needed for core20 and core22, which change host names during launching and get the daemon confused about which IP is the correct one.

It also makes sure that when an instance is done rebooting, it has a `Running` state, since those instances were getting stuck in a `Restarting` state.

fix #2690

### DIFF
--- a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
+++ b/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
@@ -136,7 +136,8 @@ mp::QemuPlatformDetail::~QemuPlatformDetail()
 
 std::optional<mp::IPAddress> mp::QemuPlatformDetail::get_ip_for(const std::string& hw_addr)
 {
-    return dnsmasq_server->get_ip_for(hw_addr);
+    auto ip_and_host = dnsmasq_server->get_ip_and_host_for(hw_addr);
+    return ip_and_host ? std::make_optional(ip_and_host.value().first) : std::nullopt;
 }
 
 void mp::QemuPlatformDetail::remove_resources_for(const std::string& name)
@@ -185,6 +186,14 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
                          << QString::fromStdString(
                                 fmt::format("tap,ifname={},script=no,downscript=no,model=virtio-net-pci,mac={}",
                                             tap_device_name, vm_desc.default_mac_address));
+}
+
+// FIXME: after moving to core22, this will be handled by dnsmasq --dhcp-ignore-clid
+void multipass::QemuPlatformDetail::release_mac_with_different_hostname(const std::string& hw_addr,
+                                                                        const std::string& name)
+{
+    if (auto ip_and_host = dnsmasq_server->get_ip_and_host_for(hw_addr); ip_and_host && ip_and_host->second != name)
+        dnsmasq_server->release_mac(hw_addr);
 }
 
 mp::QemuPlatform::UPtr mp::QemuPlatformFactory::make_qemu_platform(const Path& data_dir) const
--- a/tests/qemu/linux/test_qemu_platform_detail.cpp
+++ b/tests/qemu/linux/test_qemu_platform_detail.cpp
@@ -211,3 +211,19 @@ TEST_F(QemuPlatformDetail, writing_ipforward_file_failure_logs_expected_message)
 
     mp::QemuPlatformDetail qemu_platform_detail{data_dir.path()};
 }
+
+TEST_F(QemuPlatformDetail, release_mac_with_different_hostname)
+{
+    const mp::IPAddress ip_address{fmt::format("{}.5", subnet)};
+
+    EXPECT_CALL(*mock_dnsmasq_server, get_ip_and_host_for)
+        .WillOnce(Return(std::make_optional(std::make_pair(ip_address, name))))
+        .WillOnce(Return(std::make_optional(std::make_pair(ip_address, name + "not"))))
+        .WillOnce(Return(std::nullopt));
+    EXPECT_CALL(*mock_dnsmasq_server, release_mac(hw_addr));
+
+    mp::QemuPlatformDetail qemu_platform_detail{data_dir.path()};
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr, name);
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr, name);
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr + "not", name);
+}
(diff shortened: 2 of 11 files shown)


luis4a0 (approved):
Hey @andrei-toterman, thanks! Code looks good and it indeed works. As for the coverage, I can say although this reduces a bit the total coverage, the uncovered code changed was uncovered before. So I'm OK with that.

luis4a0:
bors r+

:crossed_fingers:

townsend2010 (commented):
This looks like the new `dnsmasq` option **doesn't** work with version we use in the snap: https://github.com/canonical/multipass/runs/8043246718?check_suite_focus=true#step:3:1346

CI FTW!!!

townsend2010 (changes requested):
Yep, confirmed the `--dhcp-ignore-clid` option is not supported by `dnsmasq` we include in the snap.

luis4a0:
> Yep, confirmed the `--dhcp-ignore-clid` option is not supported by `dnsmasq` we include in the snap.

Hey @townsend2010! Sorry, this escaped my sight. What can we do, then?

The options I see are:
1. build with `core22`, we'll have there a new version of the deb package
2. make a `dnsmasq` snap and make our snap depend on it
3. build `dnsmasq` from source

What do you think?

townsend2010:
Hey @luis4a0 and @andrei-toterman,

> build with `core22`, we'll have there a new version of the deb package

I would like to move to `core22`, but that is really risky at this point in the cycle.  That in and off itself is a Roadmap item and should be done early in the cycle, so I say no at this point.

> make a `dnsmasq` snap and make our snap depend on it

Calling executables in one snap from another snap in this manner is not really supported by design.  Caveat: It is possible in some scenarios to do this either via portals or via dbus/desktop files mechanism, but that is a bunch of heavy lifting for this and something I will avoid.

>     build `dnsmasq` from source

Of these 3 options, this is the most likely to happen, but it's still quite a bit of heavy lifting for something we could drop when we move to the `core22` base, but it certainly something could be done within this cycle and still get this in.

@andrei-toterman, I believe you had another solution working before using this cleaner solution before, correct?  If so, I think it may be better to just use your original solution with a `FIXME:` to change to this solution once we move to the `core22` base.

andrei-toterman:
@townsend2010 Sure, I'll update the PR with the old fix :smile:

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -187,6 +188,15 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
                                             tap_device_name, vm_desc.default_mac_address));
 }
 
+// FIXME: after moving to core22, this will be handled by dnsmasq --dhcp-ignore-clid
+void multipass::QemuPlatformDetail::release_mac_with_different_hostname(const std::string& hw_addr,
+                                                                        const std::string& name)
+{
+    if (auto ip_and_host = dnsmasq_server->get_ip_and_host_for(hw_addr);
+        ip_and_host.has_value() && ip_and_host.value().second != name)
+        dnsmasq_server->release_mac(hw_addr);
```
luis4a0, line 197: I know this is very picky, but I couldn't understand this at first sight because of the unfortunate indentation. What about:
```
    if (auto ip_and_host = dnsmasq_server->get_ip_and_host_for(hw_addr); ip_and_host && ip_and_host->second != name)
        dnsmasq_server->release_mac(hw_addr);
```
?

The whole `if` block would fit in one line. It is completely valid to test the `std::optional` for nullity and the `operator*` is equivalent to `value()` when the optional contains a value.

luis4a0 (approved):
Hey @andrei-toterman, now it works :P

luis4a0:
bors cancel

luis4a0:
Sorry @andrei-toterman, I approved instead of waiting for you to answer. Just those two details and we'll be done.

tests/qemu/linux/test_qemu_platform_detail.cpp
```diff
@@ -209,3 +211,19 @@ TEST_F(QemuPlatformDetail, writing_ipforward_file_failure_logs_expected_message)
...
+        .WillOnce(Return(std::make_optional(std::make_pair(ip_address, name))))
+        .WillOnce(Return(std::make_optional(std::make_pair(ip_address, name + "not"))))
+        .WillOnce(Return(std::nullopt));
+    EXPECT_CALL(*mock_dnsmasq_server, release_mac(hw_addr));
+
+    mp::QemuPlatformDetail qemu_platform_detail{data_dir.path()};
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr, name);
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr, name);
+    qemu_platform_detail.release_mac_with_different_hostname(hw_addr + "not", name);
+}
```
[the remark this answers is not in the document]
andrei-toterman: Oops, yup, I'll add it!

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
andrei-toterman: Indeed it is not very readable. Thanks for the tips!

luis4a0:
Thanks @andrei-toterman! Let's give it a try!

bors r+

PR #2718, opened 2022-08-24
sshfs infinitely looping on missing mount directory

### LINKED ISSUE #1677: sshfs_server gets in tight loop if host mount directory is missing on instance start
If a mount is defined and running and then the mounted host directory is removed, but the mount is still defined, subsequent starts of the instance will lead to `sshfs_server` taking 100% CPU time.

sharder996:
This PR add a check to make sure that the given directory exists before attempting to mount it. If the directory is not found, the host and target directories will be removed from the database.

Fixes #1677

### DIFF
--- a/src/client/cli/cmd/restart.cpp
+++ b/src/client/cli/cmd/restart.cpp
@@ -46,7 +46,16 @@ mp::ReturnCode cmd::Restart::run(mp::ArgParser* parser)
         return standard_failure_handler_for(name(), cerr, status);
     };
 
-    spinner.start(instance_action_message_for(request.instance_names(), "Restarting "));
+    auto streaming_callback = [this, &spinner](mp::RestartReply& reply) {
+        if (!reply.log_line().empty())
+        {
+            spinner.print(cerr, reply.log_line());
+        }
+
+        spinner.stop();
+        spinner.start(reply.reply_message());
+    };
+
     request.set_verbosity_level(parser->verbosityLevel());
 
     std::unique_ptr<multipass::utils::Timer> timer;
@@ -58,7 +67,14 @@ mp::ReturnCode cmd::Restart::run(mp::ArgParser* parser)
         timer->start();
     }
 
-    return dispatch(&RpcMethod::restart, request, on_success, on_failure);
+    ReturnCode return_code;
+    do
+    {
+        spinner.start(instance_action_message_for(request.instance_names(), "Restarting "));
+    } while ((return_code = dispatch(&RpcMethod::restart, request, on_success, on_failure, streaming_callback)) ==
+             ReturnCode::Retry);
+
+    return return_code;
 }
 
 std::string cmd::Restart::name() const
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2667,11 +2667,15 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
                 }
                 catch (const std::exception& e)
                 {
-                    fmt::format_to(errors, "Removing \"{}\": {}\n", target_path, e.what());
+                    mpl::log(mpl::Level::error, category, fmt::format("Removing \"{}\": {}\n", target_path, e.what()));
                     invalid_mounts.push_back(target_path);
                 }
-                persist_instances();
             }
+
+            for (const auto& mount : invalid_mounts)
+                vm_instance_specs[name].mounts.erase(mount);
+
+            persist_instances();
         }
     }
     catch (const std::exception& e)
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2667,7 +2667,8 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
                 }
                 catch (const std::exception& e)
                 {
-                    mpl::log(mpl::Level::error, category, fmt::format("Removing \"{}\": {}\n", target_path, e.what()));
+                    mpl::log(mpl::Level::warning, category,
+                             fmt::format("Removing \"{}\": {}\n", target_path, e.what()));
                     invalid_mounts.push_back(target_path);
                 }
             }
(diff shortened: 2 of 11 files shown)


sharder996:
While a `warning` level log message is recorded on a mount being removed, the cli does not display anything. Since the current state of `main` only allows for passing back errors (resulting in a message of `start failed` which is inappropriate since the command still works) I am putting up what I have so far so that we can discuss how we might want to go about displaying warnings as well.

andrei-toterman:
Hi @sharder996! Nice work, and it indeed seems to function properly. But I got one proposal, open for debate: What if we don't remove the mount entry if it failed? We can check if the directory exists, and if it doesn't, just report it to the user and skip its mounting, but leave the mount entry there, just in case the directory reappears back, so the user doesn't have to mount it again. Opinions?

sharder996:
Hello @andrei-toterman I think we should remove mount entries if an exception occurs while mounting them. Not removing them, invalid mounts would be displayed with valid mounts in `multipass info <instance>` with nothing differentiating them. Also, if the user broke a mount by deleting/moving a source directory, I think it should be expected that they have to re-mount the directory.

But, I think the decision of what we should do here basically boils down to the question of "Do we want to try to fix the user's problem for them or have them fix it themselves?".

townsend2010:
Hey @sharder996 and @andrei-toterman,

> Do we want to try to fix the user's problem for them or have them fix it themselves?

If a directory is missing on the host, I believe the user should have to fix it since it's most likely removed intentionally.  I also think we should treat it as a failure and not leave the mount defined.  The user must either fix the issue or just not have the particular mount defined.  My 2 cents :slightly_smiling_face:

src/daemon/daemon.cpp
```diff
@@ -2662,11 +2663,26 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
...
-                persist_instances();
             }
+
+            for (const auto& mount : invalid_mounts)
+                vm_instance_specs[name].mounts.erase(mount);
+
+            if (server && !fmt::to_string(warnings).empty())
+            {
+                Reply reply;
+                reply.set_log_line(fmt::to_string(warnings));
```
andrei-toterman, line 2681: this is calling `fmt::to_string` twice, which involves two string allocations. you could use `warnings.size() > 0` instead

src/sshfs_mount/sshfs_mounts.cpp
```diff
@@ -65,6 +67,12 @@ mp::SSHFSMounts::SSHFSMounts(const SSHKeyProvider& key_provider) : key(key_provi
 void mp::SSHFSMounts::start_mount(VirtualMachine* vm, const std::string& source_path, const std::string& target_path,
                                   const mp::id_mappings& gid_mappings, const mp::id_mappings& uid_mappings)
 {
+    if (!MP_FILEOPS.exists(QDir{QString::fromStdString(source_path)}))
+    {
+        mount_processes[vm->vm_name].erase(target_path);
+        throw std::runtime_error(fmt::format("Mount path `{}` does not exist.", source_path));
```
andrei-toterman, line 73: i see that you're using backticks here, but in most other places single quotes are used for printing paths

src/daemon/daemon.cpp
```diff
@@ -2662,11 +2663,26 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::stri
                 }
                 catch (const std::exception& e)
                 {
-                    fmt::format_to(errors, "Removing \"{}\": {}\n", target_path, e.what());
+                    // TODO: Combine these into one warning level log once they are displayed in the cli by
+                    // default
+                    mpl::log(mpl::Level::info, category, fmt::format("Removing \'{}\': {}\n", target_path, e.what()));
+                    fmt::format_to(warnings,
+                                   fmt::format("Removing mount \'{}\' from {}: {}\n", target_path, name, e.what()));
```
andrei-toterman, line 2670: i don't think single quotes need to be escaped like that, they should be printed just fine without the backslash

src/client/cli/cmd/restart.cpp
```diff
@@ -46,7 +46,16 @@ mp::ReturnCode cmd::Restart::run(mp::ArgParser* parser)
...
-    spinner.start(instance_action_message_for(request.instance_names(), "Restarting "));
+    auto streaming_callback = [this, &spinner](mp::RestartReply& reply) {
+        if (!reply.log_line().empty())
+        {
+            spinner.print(cerr, reply.log_line());
+        }
+
+        spinner.stop();
+        spinner.start(reply.reply_message());
+    };
```
andrei-toterman, line 57: i see there are some similar lines in `start.cpp` and they are covered. could these be covered in the same way?

andrei-toterman (changes requested):

andrei-toterman:
Alright, looks good to me! The code that still remains uncovered is pretty much in sections that were already uncovered, so I guess it is okay!

bors merge

andrei-toterman (approved):

andrei-toterman (approved):
Alright, looks good to me! The code that still remains uncovered is pretty much in sections that were already uncovered, so I guess it is okay!

bors merge

townsend2010:
bors retry

townsend2010:
One more time...

bors retry

townsend2010:
Hmm, bors keeps failing on the same test on Windows with this PR.  Something fishy is going on, but it needs to be investigated before we merge.

townsend2010:
The test is `Client/TimeoutSuite.command_exits_on_timeout/2, where GetParam() = "restart"` and this code touches the `restart` handling.  It may be introducing a race on Windows since it's a slow VM.

sharder996:
@townsend2010 The first time bors failed, it failed on `where GetParam() = "start"` which was unmodified. Indication that this test failure is unrelated? I took a look at the `restart` test in question (which comes from the same test suite), but I'm not seeing anything to suggest why it's failing.

PR #2723, opened 2022-08-28
bors: only wait for builds (and unit tests)

Saviq:
This way it won't block due to frequent infrastructure issues.
Also change the committer name / email so that
multipass-ci-bot@canonical.com gets emailed on failures.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,20 +1,13 @@
 status = [
            "Linux",
-           "CI (Linux, linux, lxd)",
-           "CI (Linux, linux, qemu)",
            "macOS-x86_64",
-           "CI (macOS, macos-10.15, hyperkit, x86_64)",
-           "CI (macOS, macos-10.15, qemu, x86_64)",
-           "CI (macOS, macos-11, hyperkit, x86_64)",
-           "CI (macOS, macos-11, qemu, x86_64)",
            "macOS-arm64",
-           "CI (macOS, qemu, arm64)",
            "Windows",
-           "CI (Windows, windows-10, hyperv)",
-           "CI (Windows, windows-11, hyperv)"
          ]
 block-labels = [ "no-merge" ]
-timeout-sec = 12000
+committer.name = "Multipass CI Bot"
+committer.email = "multipass-ci-bot@canonical.com"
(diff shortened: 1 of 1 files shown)


Saviq:
bors try

ricab:
Hmm nice. What does the bot when emailed? Anywhere we can subscribe to that?

Saviq:
> Hmm nice. What does the bot do when emailed? Anywhere we can subscribe to that?

It's a simple forward, IS manages to whom.

@townsend2010 can you confirm you got a failure email yesterday?

townsend2010:
> @townsend2010 can you confirm you got a failure email yesterday?

Yes, I did receive it!

Saviq:
> Yes, I did receive it!

Sounds like a workable solution, then?

townsend2010:
> Sounds like a workable solution, then?

Hmm, now that I think about it, I did get an email about testing the forward, but you said in the email that you think GH is not sending emails on failures that it didn't trigger itself and that it may need a custom action. Is the "failure" email you are talking about or was there a different email?

Saviq:
@townsend2010 did you get a notification about this failing?

https://github.com/canonical/multipass/actions/runs/2942985393/attempts/1

townsend2010:
Hrmph, no, I didn't :slightly_frowning_face:

Saviq:
> Hrmph, no, I didn't slightly_frowning_face

Right, then that's what I meant, even though originally I believed it didn't send them because _I_ didn't get them. But I'm not on the forward any more.

townsend2010:
Hey @Saviq!

The lack of getting the failure email is not dependent on this PR, right?

Saviq:
@townsend2010 well, merging this would leave you without notifications about integration failures

townsend2010:
Ok, what I meant was is fixing the lack of notifications something we do in this PR or is it somewhere else that it needs fixed?

Saviq:
Ah. Yes, it can be here. It will need to be an extra step in the job. Just need to find how best to approach it.

townsend2010:
bors is dead, so let's just close this now.

PR #2724, opened 2022-08-28
ci: allow cancellation of BuildAndTest

Saviq:
Jobs/steps need explicit `!cancelled()` if they use any of the status
check functions:

https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -59,10 +59,13 @@ jobs:
     # Let Lint fail on pull requests
     if: |
       ${{
-        success()
-        || needs.Lint.result == 'skipped'
-        || ( needs.Lint.result == 'failure'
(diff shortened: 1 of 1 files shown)


ricab:
Doesn't `success()` imply `!cancelled()`? The link you posted says that success "Returns true when none of the previous steps have failed or been canceled."

ricab:
Oh nevermind, it's being or-ed with other things.

Saviq:
> Doesn't `success()` imply `!cancelled()`? The link you posted says that success "Returns true when none of the previous steps have failed or been canceled."

You'd think so, wouldn't you... this didn't work (cancelling didn't interrupt builds), so that may be true in fact. I'll have to dig more.

PR #2727, opened 2022-08-29
sshfs not cleanly shutdown when stopping an instance

### LINKED ISSUE #1132: sshfs_server not cleanly shut down when stopping an instance

sharder996:
This PR adds logic to force a process to wait until the process has exited before continuing. If the process does not terminate within the specified amount of time, kill it.

Fixes #1132

### DIFF
--- a/src/sshfs_mount/sshfs_mounts.cpp
+++ b/src/sshfs_mount/sshfs_mounts.cpp
@@ -154,7 +154,8 @@ bool mp::SSHFSMounts::stop_mount(const std::string& instance, const std::string&
 
         if (!sshfs_mount->wait_for_finished(1000))
         {
-            mpl::log(mpl::Level::info, category, "failed to terminate nicely, killing");
+            mpl::log(mpl::Level::info, category,
+                     fmt::format("Failed to terminate mount '{}' in instance \"{}\", killing", path, instance));
             sshfs_mount->kill();
         }
         return true;
--- a/src/sshfs_mount/sshfs_mounts.cpp
+++ b/src/sshfs_mount/sshfs_mounts.cpp
@@ -172,10 +172,11 @@ void mp::SSHFSMounts::stop_all_mounts_for_instance(const std::string& instance)
     }
     else
     {
-        for (auto it = mounts_it->second.cbegin(), next = it; it != mounts_it->second.cend(); it = next)
+        for (auto it = mounts_it->second.cbegin(); it != mounts_it->second.cend();)
         {
-            next++;
-            stop_mount(instance, it->first);
+            // Clever postfix increment with member access needed to prevent iterator invalidation since iterable is
+            // modified in sshfs connection signal handler
+            stop_mount(instance, it++->first);
         }
     }
     mount_processes[instance].clear();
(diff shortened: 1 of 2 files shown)


src/sshfs_mount/sshfs_mounts.cpp
```diff
@@ -162,15 +167,14 @@ void mp::SSHFSMounts::stop_all_mounts_for_instance(const std::string& instance)
...
     {
-        for (auto& sshfs_mount : mounts_it->second)
+        for (auto it = mounts_it->second.cbegin(), next = it; it != mounts_it->second.cend(); it = next)
         {
-            mpl::log(mpl::Level::debug, category,
-                     fmt::format("Stopping mount '{}' in instance \"{}\"", sshfs_mount.first, instance));
-            sshfs_mount.second->terminate();
+            next++;
+            stop_mount(instance, it->first);
         }
```
andrei-toterman, line 178: i don't really get this for-loop. what is the point of `next`? can't this be simplified like this?
```cpp
for (auto it = mounts_it->second.cbegin(), next = it; it != mounts_it->second.cend(); it = next)
{
    next++;
    stop_mount(instance, it->first);
}
```
:arrow_down: 
```cpp
for (auto it = mounts_it->second.cbegin(), next = it; it != mounts_it->second.cend(); it = next)
{
    stop_mount(instance, it->first);
    next++;
}
```
:arrow_down: 
```cpp
for (auto it = mounts_it->second.cbegin(), next = it; it != mounts_it->second.cend(); it = ++next)
    stop_mount(instance, it->first);
```
:arrow_down: 
```cpp
for (auto it = mounts_it->second.cbegin(); it != mounts_it->second.cend(); it++)
    stop_mount(instance, it->first);
```
:arrow_down: 
```cpp
for (const auto& it : mounts_it->second)
    stop_mount(instance, it->first);
```
:arrow_down: 
```cpp
for (const auto& [path, _] : mounts_it->second)
    stop_mount(instance, path);
```

sharder996: The purpose behind `next` is to keep track of the position of the next item in the `mount_processes` map. When `stop_mount` is called, the connection listener will delete the item from the map [here](https://github.com/canonical/multipass/blob/645f9df7462cf10f63d4568cd66804817c0ba31b/src/sshfs_mount/sshfs_mounts.cpp#L101) thus losing a pointer to the next item. `next` will keep track of the next item eliminating a potential race.

andrei-toterman: Ah, it's about iterator invalidation in maps! Now I understand, thanks for explaining!
Can you please leave a comment above the loop explaining the reason behind it? I'm sure that will be helpful in the future.
Also since this is about keeping the iterator one step ahead of the deletes, how about
```cpp
for (auto it = mounts_it->second.cbegin(); it != mounts_it->second.cend();)
    stop_mount(instance, it++->path);
```

sharder996: Made your suggested changes to the loop and also added a comment.

andrei-toterman:
bors retry

townsend2010:
Bah, bors shows everything passed in https://app.bors.tech/batches/247858 yet it timed out.  I'm going to merge this manually.

PR #2732, opened 2022-08-31
[gui] not reacting to primary instance being purged

### LINKED ISSUE #1241: [gui] not reacting to primary instance being purged
1. `multipass start`
1. `multipass delete -p --all`
1. GUI doesn't realize the primary instance went away

or

1. `multipass start`
1. `multipass delete --all`
1. GUI correctly reacts to instance being deleted
1. `multipass purge`
1. GUI doesn't realize the primary instance was purged

sharder996:
This PR fixes the text shown in the gui when no primary instance is found.

Fixes #1241

### DIFF
--- a/src/client/gui/gui_cmd.cpp
+++ b/src/client/gui/gui_cmd.cpp
@@ -399,28 +399,25 @@ void cmd::GuiCmd::handle_petenv_instance(const google::protobuf::RepeatedPtrFiel
     // petenv doesn't exist yet
     if (petenv_instance == instances.cend())
     {
-        if (current_petenv_name != petenv_name)
-        {
-            petenv_start_action.setText("Start");
-            petenv_start_action.setEnabled(false);
-            petenv_shell_action.setEnabled(true);
-            petenv_stop_action.setEnabled(false);
+        petenv_start_action.setText("Start");
+        petenv_start_action.setEnabled(false);
+        petenv_shell_action.setEnabled(false);
+        petenv_stop_action.setEnabled(false);
 
+        current_petenv_name = petenv_name;
+    }
+    else
+    {
+        auto state = petenv_instance->instance_status();
+
+        if (petenv_state.status() != state.status() || petenv_name != current_petenv_name)
+        {
+            petenv_start_action.setText(set_title_string_for(fmt::format("Start \"{}\"", petenv_name), state));
+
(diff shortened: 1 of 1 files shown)


src/client/gui/gui_cmd.cpp
```diff
@@ -399,28 +399,25 @@ void cmd::GuiCmd::handle_petenv_instance(const google::protobuf::RepeatedPtrFiel
...
-            petenv_start_action.setText("Start");
-            petenv_start_action.setEnabled(false);
-            petenv_shell_action.setEnabled(true);
-            petenv_stop_action.setEnabled(false);
-
-            current_petenv_name = petenv_name;
-        }
+        petenv_start_action.setText("Start");
+        petenv_start_action.setEnabled(false);
+        petenv_shell_action.setEnabled(false);
```
andrei-toterman, line 404: i think this should remain `true`. so far, if there was no primary instance, clicking that button would launch a terminal and create the primary instance, and that's a nice one-click action that would put you into the shell of the primary instance

src/client/gui/gui_cmd.cpp
```diff
@@ -399,28 +399,25 @@ void cmd::GuiCmd::handle_petenv_instance(const google::protobuf::RepeatedPtrFiel
...
-        petenv_state = state;
-        current_petenv_name = petenv_name;
+        if (petenv_state.status() != state.status() || petenv_name != current_petenv_name)
+        {
+            petenv_start_action.setText(set_title_string_for(fmt::format("Start \"{}\"", petenv_name), state));
+
+            set_input_state_for({&petenv_start_action, &petenv_shell_action, &petenv_stop_action}, state);
+            petenv_state = state;
+            current_petenv_name = petenv_name;
+        }
```
andrei-toterman, line 420: i know this is purely a matter of preference in this case, but i think an early `return` is better than an `else` + it doesn't really change any functionality, it's just a style thing. but i also get the appeal in this simple case, so you can leave it like this if you like it :smiley:

andrei-toterman (changes requested):

andrei-toterman (approved):
Perfect!

bors merge

townsend2010:
Seems a race in the Windows tests is starting to show up.  We'll try it once more...

bors retry

PR #2742, opened 2022-09-13
Handle bad settings read from disk

ricab:
Be robust to bad setting values read from disk, by logging the occurrence, resetting, and returning the default.

The lazy approach taken here reuses `SettingSpec::interpret`, but it imposes a new constraint on it: it needs to be able to validate already "interpreted" (marshalled) settings. I want to see how this goes on different platforms and integration tests. 

If this works, we can rely on it to reset `hyperkit` to `qemu` once the former is gone (because the latter will be the default by then). Otherwise, we could try to finally move to a marshall/unmarshall scheme, or just tackle the hyperkit case with dedicated code.

### DIFF
--- a/src/settings/persistent_settings_handler.cpp
+++ b/src/settings/persistent_settings_handler.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2022 Canonical, Ltd.
+ * Copyright (C) 2021-2023 Canonical, Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


ricab:
Un-drafting to get a Windows run too.

ricab:
This still needed some work after all. One unit test missing at this point.

ricab:
Actually I think this is good enough. The only missing coverage is on the QSettings wrapper (which we decided before didn't need testing).

ricab:
Hey @townsend2010, just to let you know that this is now ready for review.

townsend2010:
bors try

townsend2010 (approved):
LGTM, thanks!

bors merge

townsend2010:
The danged `clang` crash in CI...

bors retry

PR #2745, opened 2022-09-14
Blueprints v2 with loading from URL

### LINKED ISSUE #2671: [blueprints] Allow launching images via URL
In order for Blueprints to launch custom images, Blueprints need to allow launching via URL.  Architecture needs to be taken into account when specifying images.

luis4a0:
This PR introduces fetching images from URL on Blueprints. Since the images are only intended to be created in AMD64 and ARM64, adding this feature to Blueprints makes the YAML format used in the old Blueprints incompatible. For this, we introduce a new version of Blueprints, v2.

Fixes #2671.


Saviq:
If I may chime in, does this really require a bump to `v2`? Can't you instead maintain backwards compatibility?

townsend2010:
> Can't you instead maintain backwards compatibility?

Sure, it's *possible*, but it would add lots of complexity for Blueprints maintainers and for multipass.  The notion of Blueprint-specific images will *only* apply to amd64 and arm64 and after team discussion, we felt that separating the Blueprints that either support the v1 way such as the base cloud image, lots of bootstrapping of cloud-init, etc. should be separate from specifying an image and only allowing the Blueprints to run on those two architectures.  It's a clean break between the two.  Also, doing this will allow a corresponding v1 Blueprint that runs on a different architecture to still exist with muddying up the Blueprint with all sorts of logic to do things depending on if there a Blueprint-specific image or not.

Saviq:
> Blueprint-specific images will only apply to amd64 and arm64

Shouldn't that be a per-blueprint thing? I mean, if someone comes in and provides an armhf image, that should be supported?

> doing this will allow a corresponding v1 Blueprint that runs on a different architecture to still exist

That I grok, but only if that doesn't mean that amd64/arm64 will _only_ support v2, and others _only_ support v1?

luis4a0:
Hey @Saviq! Imagine the situation with our Docker Blueprint. We want to provide only AMD64 and ARM64 builds, while generating the VM from a plain jammy image for the other architectures. So we'll have a cloud-init for AMD64, ARM64, and a different cloud-init for the other archs. If we do all the definitions on the same YAML, even modifying the syntax for that, would be a mess. Introducing v2 will tidy things up in this case.

Saviq:
@luis4a0 sure, I get why it's cleaner this way, but IMO limiting, if you can't "escape" it. If Multipass parses _both_ the `v1` and `v2`, and selects the highest one supporting the current arch, that allows for both a `v1/foo.yaml` and a `v2/foo.yaml` to exist. This way you can gradually introduce bespoke images. But you have to maintain two blueprints, even if a lot of them will be the same (that obviously depends on what the blueprint actually does - if most of it _can_ be moved to the bespoke image, then the `v2` will be much simpler).

For reference, Snapcraft has the same problem, and they solved it with advanced grammar, so there is prior art: https://snapcraft.io/docs/snapcraft-advanced-grammar. You could think of adopting a similar approach long-term.

townsend2010:
Maybe restricting the v2 Blueprints to just amd64 and arm64 is not necessary although we did some analysis and this affects only Linux and only 0.5% of those users are potentially affected (and only a subset of those users will use a Blueprint), so drawing a line in the sand is not really that far off base.

I really think introducing v2 here is better for the reasons @luis4a0 pointed out, it keeps it neat and tidy.  We *could* do something like this though:
1. If v2 Blueprint exists and has image for host architecture, use it.
2. If not, then check if a corresponding v1 Blueprint exists. If so, check `runs-on` with host architecture and use it if there is a match.
3. Otherwise, Blueprint doesn't work and it won't be available.

As you said, if most of a Blueprints bootstrapping can moved into an image, then v2 will be simpler.  But if it can't, there is really no reason for a v2 Blueprint and image and the Blueprint should just stay v1.

Regarding the Snapcraft advanced grammar, we didn't consider this and IMO, it muddies the Blueprint again if we introduce something like that.  There is a trade-off no matter which path we go down, but I think we are trying to cater to a very, very small section of users and I'd like to keep things as clean as we can.

Saviq:
> We _could_ do something like this though:
>  1. If v2 Blueprint exists and has image for host architecture, use it.
>  2. If not, then check if a corresponding v1 Blueprint exists. If so, check `runs-on` with host architecture and use it if there is a match.
>  3. Otherwise, Blueprint doesn't work and it won't be available.

Right, that's the backwards compatibility I had in mind. Just clarify one thing for me - can you still use image aliases in `v2`?

> Regarding the Snapcraft advanced grammar, we didn't consider this and IMO, it muddies the Blueprint again

That is true, it doesn't make for an easy read.

townsend2010:
> Right, that's the backwards compatibility I had in mind. Just clarify one thing for me - can you still use image aliases in `v2`?

There really shouldn't be a need for that.  If you want to use an alias, that is what `v1` would be for.  I see `v2` as "Here's an image that does all of the otherwise heavylifting of installing packages that you see in v1 Blueprints."  Without a bespoke image in `v2`, there really isn't a point.  But if you have an idea I'm not thinking off, I'm all, uh, eyes.

Saviq:
> There really shouldn't be a need for that. If you want to use an alias, that is what `v1` would be for. I see `v2` as "Here's an image that does all of the otherwise heavylifting of installing packages that you see in v1 Blueprints." Without a bespoke image in `v2`, there really isn't a point. But if you have an idea I'm not thinking off, I'm all, uh, eyes.

That's not what's versions are in my mind… You shouldn't be stuck with two "active" versions for too long. And if someone can only maintain an image for either amd64 or arm64, they'll need to maintain two blueprints, even if they could encode the logic in just one (conditional on e.g. what's preinstalled or not - they can do it just fine within the commands ran, most of them will even be idempotent). I just worry you'll end up stuck with the two modes forever and have to explain / document / otherwise deal with it.

Which is why I ultimately think it should still be "v1", just with a new grammar for `images` supported, and aliases supported in there as well. Let the blueprint author decide how they approach it.

It's your call of course, just my ¢2.

townsend2010:
I certainly understand your point and thought about this too.  It's not just about the `images` grammar- that's actually the uncomplicated the part.  The complicated part is the handling of the `cloud-init`:`vendor-data` part.  We just pass this in directly to cloud-init, but there would be many conditional parts in this section.  We would basically have to have two separate `vendor-data`'s depending on whether a bespoke image is available or not.  That would get very messy in the Blueprint IMO.

Also, `v2` is just an arbitrary name.  We could call it something else like `v1`/image-based/foo.yaml` and it would accomplish the same thing and not look like a version jump.

In the end, there's going to be a compromise with this.  The question that needs to be answered is do we want to still use a unified Blueprint that supports all of this (which will be complicated for the author) or have the option of two separate Blueprints in which the Blueprints themselves will be much cleaner?

Saviq:
ACK on all points. I suppose I just dislike the inevitable duplication, and the alternative of _only_ supporting a{md,rm}64 in a given blueprint (which is what I imagine this split will lead to). You clearly thought this through and I'm outvoted, so will not push this more :)

townsend2010:
> I suppose I just dislike the inevitable duplication

I don't really like it either, but I think the alternative of having a complicated Blueprint is worse, but yes, I certainly understand you on this.

> the alternative of *only* supporting a{md,rm}64 in a given blueprint

Well, that won't be a restriction after thinking about what you said as we will allow authors to use bespoke images for other architectures.  Yes, I don't think there will be too many takers on maintaining a plethora of images, but images will be a very important factor for the best user experience when using Blueprints and we want to encourage this.

Anyways, I always appreciate your comments and it's good for discussion :wink:

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -41,39 +41,108 @@ namespace mpl = multipass::logging;
...
+}
+
+auto blueprints_map_for(const std::string& archive_file_path, bool& needs_update, const std::string& arch)
 {
     std::map<std::string, YAML::Node> blueprints_map;
     std::ifstream zip_stream{archive_file_path, std::ios::binary};
     auto zip_archive = MP_POCOZIPUTILS.zip_archive_for(zip_stream);
 
-    for (auto it = zip_archive.headerBegin(); it != zip_archive.headerEnd(); ++it)
+    for (auto blueprint_dir_version : blueprint_dir_versions)
```
townsend2010, line 93: Perhaps `const auto blueprint_dir_version`?

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -135,14 +208,38 @@ mp::Query mp::DefaultVMBlueprintProvider::fetch_blueprint_for(const std::string&
         }
     }
 
-    auto blueprint_instance = blueprint_config["instances"][blueprint_name];
+    auto blueprint_instance = blueprint_config[instances_key][blueprint_name];
 
     // TODO: Abstract all of the following YAML schema boilerplate
-    if (blueprint_instance["image"])
+
+    if (blueprint_config["blueprint-version"].as<std::string>() == "v2")
```
townsend2010, line 215: I think instead of hard coding `v2` here, it would be better to leverage the constant in `blueprint_dir_versions`.

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -135,14 +208,38 @@ mp::Query mp::DefaultVMBlueprintProvider::fetch_blueprint_for(const std::string&
...
+                mpl::log(mpl::Level::debug, category, fmt::format("Downloading SHA256 from {}", sha256_string));
+                auto downloaded_sha256 = url_downloader->download(QUrl(QString::fromStdString(sha256_string)));
+                if (downloaded_sha256.size() > 64)
+                    downloaded_sha256.truncate(64); // To account for newlines or other content.
+                sha256_string = QString(downloaded_sha256).toStdString();
+            }
+            mpl::log(mpl::Level::debug, category, fmt::format("Add SHA256 \"{}\" to image record", sha256_string));
+            vm_desc.image.id = sha256_string;
+        }
+        else
```
townsend2010, line 236: Idem, braces for the `else`.

src/daemon/default_vm_image_vault.h
```diff
@@ -49,7 +50,8 @@ class DefaultVMImageVault final : public BaseVMImageVault
     ~DefaultVMImageVault();
 
     VMImage fetch_image(const FetchType& fetch_type, const Query& query, const PrepareAction& prepare,
-                        const ProgressMonitor& monitor) override;
+                        const ProgressMonitor& monitor, const bool unlock,
+                        const std::optional<std::string> checksum) override;
```
townsend2010, line 54: I think this should be a `const` reference.

townsend2010 (commented):
Hey @luis4a0!

After a very long time of this PR sitting here, I did my first pass.  Please see comments inline.

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -41,39 +41,108 @@ namespace mpl = multipass::logging;
...
+                std::vector<std::string> runs_on = blueprint_node[runs_on_key].as<std::vector<std::string>>();
+
+                return std::find(runs_on.cbegin(), runs_on.cend(), arch) != runs_on.cend();
+            }
+            catch (const YAML::BadConversion&)
+            {
+                throw mp::InvalidBlueprintException(fmt::format(bad_conversion_template, runs_on_key, blueprint_name));
+            }
+        }
+        else
```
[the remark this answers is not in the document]
luis4a0: Hey @townsend2010, thanks for the review! Right, at some point I stopped using them and I know it's a bad habit. Corrected.

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -41,39 +41,108 @@ namespace mpl = multipass::logging;
...
+
+    if (blueprint_node[instances_key] && blueprint_node[instances_key][blueprint_name] &&
+        blueprint_node[instances_key][blueprint_name]["images"])
+    {
+        if (blueprint_node[instances_key][blueprint_name]["images"][arch])
+        {
+            if (blueprint_node[instances_key][blueprint_name]["images"][arch]["url"])
+                return true;
+        }
+        else
```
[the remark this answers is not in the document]
luis4a0: Added braces.

src/blueprint_provider/default_vm_blueprint_provider.cpp
luis4a0: Right, fixed.

src/blueprint_provider/default_vm_blueprint_provider.cpp
luis4a0: That's a good idea. I made the strings `"v1"` and `"v2"` `constexpr`. Is that what you mean?

src/blueprint_provider/default_vm_blueprint_provider.cpp
luis4a0: Done, added the braces.

src/daemon/default_vm_image_vault.h
luis4a0: Do you mean the `checksum` argument? I thought `optional` was implemented as a pointer so the refetence was not needed. But anyway, it makes the function signature look neat.

src/blueprint_provider/default_vm_blueprint_provider.cpp
townsend2010: Yeah, pretty much:)

src/daemon/default_vm_image_vault.h
townsend2010: Yes, that argument.  Thanks for changing this.  From https://en.cppreference.com/w/cpp/utility/optional, it says:

> If an optional<T> contains a value, the value is guaranteed to be allocated as part of the optional object footprint, i.e. no dynamic memory allocation ever takes place. Thus, an optional object models an object, not a pointer, even though [operator*()](https://en.cppreference.com/w/cpp/utility/optional/operator*) and [operator->()](https://en.cppreference.com/w/cpp/utility/optional/operator*) are defined.

luis4a0: Amazing, thanks!

townsend2010 (commented):
Ok, just one more question inline and I think this is almost ready to merge :slightly_smiling_face:

src/blueprint_provider/default_vm_blueprint_provider.cpp
```diff
@@ -41,39 +41,117 @@ namespace mpl = multipass::logging;
...
+
+bool runs_on(const std::string& blueprint_name, const YAML::Node& blueprint_node, const std::string& arch)
+{
+    if (blueprint_node["blueprint-version"].as<std::string>() == version_v1)
+    {
+        if (blueprint_node[runs_on_key])
+        {
+            try
+            {
+                std::vector<std::string> runs_on = blueprint_node[runs_on_key].as<std::vector<std::string>>();
```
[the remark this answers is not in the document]
luis4a0: Hey @townsend2010, sorry, I can't get what you mean here. Are you referring to `runs_on_key`? It's compiling and running fine on my computer.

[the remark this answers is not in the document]
townsend2010: I mean does `const auto runs_on =` not work instead of explicitly using the type?  I prefer the `auto` convention, but if it doesn't work, then I understand.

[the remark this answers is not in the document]
luis4a0: Oh, right, sorry. Yes, `const auto` works fine :). I corrected it.

townsend2010 (approved):
Ok, let's do it! :rocket: 

bors merge

townsend2010:
Integration tests actually passed: https://github.com/canonical/multipass/actions/runs/4088953335

Merging manually...\o/

PR #2753, opened 2022-09-22
Recursive transfer dir perms

andrei-toterman:
This PR fixes a bug in the `transfer -r` command that did not set the copied directories permissions accordingly

### DIFF
--- a/src/ssh/sftp_client.cpp
+++ b/src/ssh/sftp_client.cpp
@@ -19,7 +19,6 @@
 
 #include "ssh_client_key_provider.h"
 #include <multipass/file_ops.h>
-#include <multipass/format.h>
 #include <multipass/logging/log.h>
 #include <multipass/ssh/sftp_utils.h>
 #include <multipass/ssh/throw_on_error.h>
@@ -124,12 +123,9 @@ void SFTPClient::push_file(const fs::path& source_path, const fs::path& target_p
 
     do_push_file(*local_file, target_path);
 
-    auto target_attr = mp_sftp_stat(sftp.get(), target_path.u8string().c_str());
     std::error_code _;
     auto status = MP_FILEOPS.status(source_path, _);
-
-    target_attr->permissions = static_cast<mode_t>(status.permissions());
-    if (sftp_setstat(sftp.get(), target_path.u8string().c_str(), target_attr.get()) != SSH_FX_OK)
+    if (sftp_chmod(sftp.get(), target_path.u8string().c_str(), static_cast<mode_t>(status.permissions())) != SSH_FX_OK)
         throw SFTPError{"cannot set permissions for remote file {}: {}", target_path, ssh_get_error(sftp->session)};
 
     if (local_file->fail() && !local_file->eof())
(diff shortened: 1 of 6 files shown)


src/ssh/sftp_client.cpp
```diff
@@ -161,6 +157,10 @@ bool SFTPClient::push_dir(const fs::path& source_path, const fs::path& target_pa
     auto local_iter = MP_FILEOPS.recursive_dir_iterator(source_path, err);
     if (err)
         throw SFTPError{"cannot open local directory {}: {}", source_path, err.message()};
+
+    std::vector<std::pair<fs::path, fs::perms>> subdirectory_perms{
+        {target_path, MP_FILEOPS.status(source_path, err).permissions()}};
```
luis4a0, line 162: This type is repeated, could you add a `typedef`?

src/ssh/sftp_client.cpp
```diff
@@ -215,6 +218,18 @@ bool SFTPClient::push_dir(const fs::path& source_path, const fs::path& target_pa
         }
     }
 
+    for (auto it = subdirectory_perms.crbegin(); it != subdirectory_perms.crend(); ++it)
```
luis4a0, line 221: Is it important to have all the subdirs sorted? I see you start by the end, there must be a reason I am missing. But, if not, I'd use and `std::map` or `std::unordered_map` instead of a vector of pairs. WDYT?

luis4a0 (changes requested):

luis4a0:
Hey @andrei-toterman! Looks good, thanks! Just minor remarks inline.

src/ssh/sftp_client.cpp
andrei-toterman: It's not really repeated. One vector has `std::pair<fs::path, fs::perms>` and the other has `std::pair<fs::path, mode_t>`, since one will be used by `std::filesystem::permissions` and the other will be used by `sftp_chmod` :smile:

src/ssh/sftp_client.cpp
andrei-toterman: My reasoning was that the permissions should be set in the reverse order in which the directories were created, to ensure that the permissions of a parent directory don't disallow setting the permissions of a child directory. For example, when copying `parent/child`, the `parent` dir is created, then the `child` one. But maybe the permissions of `parent` are in such a way that they do not allow setting the permissions of `child`, so we must first set the permissions of `child` and only then the ones of `parent`.

luis4a0: Good, thanks. Makes sense. I'm OK with the vector of pairs, then. I'm not worried about complexity because (i) we never search there and (ii) the vector will never be enormous. It was just a matter of readability.

src/ssh/sftp_client.cpp
luis4a0: Oh, right, didn't catch that. Thanks!

luis4a0 (changes requested):
Great, a last thing: can you please add `--recursive` to bash completions in `completions/bash/multipass`?

andrei-toterman:
Hey @luis4a0! Sure, I'll do it, but since I will also add a completion for `parents` in the other PR, I will do both of them in one commit in the other PR's branch, to avoid dealing with the conflict of doing them separately

andrei-toterman:
Oops, I said I will add them in the other branch, but added them here. Well, this works as well ¯\\\_(ツ)\_/¯

luis4a0:
No worries, thanks!

luis4a0 (approved):
Let's go!

bors r+

andrei-toterman:
bors retry

luis4a0:
bors retry

luis4a0:
bors retry

PR #2755, opened 2022-09-23
[rpc] Convert RPC to use bi-directional streaming

townsend2010:


townsend2010:
Hey @andrei-toterman,

I've now updated this.  Could you please give it another review pass?  Thanks!

andrei-toterman:
Hey, @townsend2010! Code looks good, I tested it locally and everything seems to work just like before, as expected. I see that there are some conflicts with the private side though. Will this need a separate private PR as well?

townsend2010:
Oh, and of course `bors` will fail because of the conflict, but it will be good to get it to run up to that point and then we can merge manually.

andrei-toterman (approved):
Alright, LGTM!

bors merge

PR #2763, opened 2022-09-30
[snap] update base to core22

### LINKED ISSUE #2762: [snap] Update Multipass snap package to use core22 base
We need to update the Multipass Snap package to use the core22 base.  This will be needed to move to Qt6.

andrei-toterman:
fix #2762

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -150,8 +150,9 @@ jobs:
       if: ${{ matrix.build-type == 'Debug' }}
       timeout-minutes: 2
       run: |
-        /snap/bin/lxc start snapcraft-multipass
-        /snap/bin/lxc exec snapcraft-multipass -- \
+        instance_name=`/snap/bin/lxc --project snapcraft --format=csv --columns=n list`
+        /snap/bin/lxc --project snapcraft start $instance_name
+        /snap/bin/lxc --project snapcraft exec $instance_name -- \
            env CTEST_OUTPUT_ON_FAILURE=1 \
                LD_LIBRARY_PATH=/root/stage/usr/lib/x86_64-linux-gnu/:/root/stage/lib/:/root/parts/multipass/build/lib/ \
                /root/parts/multipass/build/bin/multipass_tests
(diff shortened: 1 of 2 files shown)


luis4a0:
Hey @andrei-toterman, I have this issue when trying to compile throuth snapcraft:
```
2023-01-20 12:27:14.161 :: 2023-01-20 12:21:53.726 :: fatal: detected dubious ownership in repository at '/root/parts/multipass/src'
2023-01-20 12:27:14.161 :: 2023-01-20 12:21:53.726 :: To add an exception for this directory, call:
2023-01-20 12:27:14.161 :: 2023-01-20 12:21:53.726 ::
2023-01-20 12:27:14.161 :: 2023-01-20 12:21:53.726 :: 	git config --global --add safe.directory /root/parts/multipass/src
```

I remember we discussed about this, but can't remember the outcome of the discussion. What can we do here?

andrei-toterman:
Hey, @luis4a0! That is happening when snapping with Multipass instead of LXD, right? Unfortunately there's not much we can do. It's a bug within snapcraft itself (see [here](https://bugs.launchpad.net/snapcraft/+bug/1995293)). The issue stems from the fact that when snapping with Multipass, there are no UID/GID mappings used, which results in the "dubious ownership" issue. Snapcraft used to use the correct mappings, as it does with LXD, but for some reason, it doesn't do it anymore :man_shrugging:

luis4a0:
Ok, so we'll need to wait for Snapcraft people to fix that. I'll use LXD in the meantime, thanks!

townsend2010:
It's unclear to me at this time if this affects *all* snap builds using Multipass with `core22` or if it's isolated to a particular build of a snap?  If it's all snaps, then we should up the urgency for the *craft folks to fix it on their end since we need to get this in and don't want to break a bunch of snap builds.

townsend2010:
Ok, after some thought and discussion, the Snapcraft bug is not a blocker for us moving Multipass to `core22` since it still happens when Multipass isn't based on `core22`.

@luis4a0, let's please move ahead with getting this approved and merged.

luis4a0:
Hey @andrei-toterman, this builds correctly for me, and the resulting snap works. Just wanted to point out that there are some warnings you may want to address before merging: https://pastebin.canonical.com/p/GQKHhSKd6J/

Thanks!

andrei-toterman:
Ok, I've looked into it, and apparently [snapcraft does not perform automatic ELF patching in snaps with core22 base, but it will get fixed in the next version of snapcraft](https://snapcraft.io/docs/linters-classic#heading--issues-auto), and since we only get warnings, not errors, and the snap seems to work, I think this does not block the PR and we can hopefully just let snapcraft solve the issue in the future instead of trying to manually patch them.

luis4a0 (approved):
Hey @andrei-toterman! Excellent, so I'm OK for this to be merged!

luis4a0:
bors r+

:crossed_fingers:

luis4a0:
bors retry

townsend2010:
A big backlog of builds on the private side caused this to timeout.  Re-running one more time...

bors retry

luis4a0:
bors retry

townsend2010:
This has passed Linux integration tests and is failing for unrelated reasons elsewhere.  Going to merge manually,

PR #2770, opened 2022-10-04
[mounts] More common mount plumbing refactoring

townsend2010:


src/client/cli/cmd/mount.cpp
```diff
@@ -90,6 +90,22 @@ mp::ReturnCode cmd::Mount::run(mp::ArgParser* parser)
...
+            const auto [username, password] = mcp::get_user_password(term);
+
+            request.mutable_user_credentials()->set_username(username);
+            request.mutable_user_credentials()->set_password(password);
+
+            client->Write(request);
+
+            return;
+        }
+
```
luis4a0, line 108: If I am correct, this code is repeated verbatim in `restart.cpp` and `start.cpp`. Could this fragment be refactored into a new function?

luis4a0 (changes requested):
Hey @townsend2010, thanks for this! Only two issues I see. The first one is that this didn't compile on Windows, does this need a private counterpart? The second one is a comment inline. Thanks!

townsend2010:
Hey @luis4a0!

Thanks for the review!

> The first one is that this didn't compile on Windows, does this need a private counterpart?

Ah, right, the branch I'm working on for SMB mounts does implement this for Windows, but this should have a private counterpart that adds a stub for this.  I'll fix that.

src/client/cli/cmd/mount.cpp
townsend2010: Well, it's not *quite* verbatim, the request type is different in each case.  I did give it some thought on factoring this out, but then we'd have to template the function and that gets messing with a free function, so then I'd need to include it in a class and implement it in the header.  I agree it seems quite redundant.  I'll give it a bit more thought.

luis4a0: Oh, right, I didn't note that. I'm OK with repeating few lines of code in that case.

townsend2010: Ok, refactored this some more.

luis4a0 (approved):
Yeah, looks great! Thanks!

PR #2771, opened 2022-10-05
Update README.md

kshillingstad:
I am proposing a small change to run commands inside an instance from outside section. Using this option for exec is more robust for scripting VM deployments and solved a number of parsing issues for my team.
https://multipass.run/docs/exec-shells

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ way down the Inception levels... ;)
 ## Run commands inside an instance from outside
 
 ```
-$ multipass exec dancing-chipmunk -- lsb_release -a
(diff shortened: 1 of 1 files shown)


luis4a0:
Hi @kshillingstad, thank you very much for taking the time for this. We want to keep thigs as simple as possible in the readme, which is what users read first. That is why we only add details on advanced use cases deeper in the documentation. On the other hand, we need to update the readme in order to add links to our full documentation. I'll close this pull request and will open a new PR with the updates. Thanks again!

PR #2773, opened 2022-10-05
[performance mounts] Add Qemu 9p mounts

sharder996:


luis4a0:
Hey @sharder996, great job there!

Before going into the code, I tested it locally on Linux. Everything works perfectly, except one thing (we already had this issue on Mac): when mounting with default id mapping, `multipass info` correctly shows
```
Mounts:         /home/luis/multipass-build/a => aaa
                    UID map: 1000:default
                    GID map: 1000:default
```

However, inside the instance, `ls -l aaa/` shows
```
drwxrwxr-x 3 nobody nogroup 3 Sep 23 11:38 b
```
that is, id `default` is not mapped to `ubuntu` in this case. Is it possible to fix that?

luis4a0:
@sharder996 another nit in the interface:
```
$ multipass mount -t native a snapcraft-multipass:aaa --gid-map 1000:1000
mount failed: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
```

There must be an error parsing id maps, since the same command with `-t classic` works.

luis4a0:
Hey @sharder996! Thanks!

Id mapping now works better. If launching with no UID nor GID maps, it works and the user can write to the mounted folder on the instance. However, if mounting with `--uid-map 1000:1000` (knowing that the user id is `1000` on both the host and the instance) it appears to work but on the instance it is not possible to create a file on the mounted folder:
```
$ multipass mount -t native a relieved-swine:aaa --uid-map 1000:1000
$ multipass shell relieved-swine
$ cat > aaa/ffff
-bash: aaa/ffff: Operation not permitted
```

Can you please take a look to that? Thanks!

sharder996:
Hey @luis4a0! 

I found and fixed the issue with file permissions not being correctly set in the mounted directory. However, I feel like the solution is a bit messy as we need to persist another piece of information between `init_mount()` and `start_mount()` function calls. What's your opinion on maybe refactoring [SSHFSServerConfig](https://github.com/canonical/multipass/blob/6e550f10aa37164b9a202f9c6994bf2b0b12f33b/include/multipass/sshfs_server_config.h#L29) into something more generalized or creating another struct?

luis4a0:
Hey @sharder996! Thanks for that! I'll take a look to the code and try to think about a refactor. Will discuss later on!

include/multipass/mount_config.h
```diff
@@ -0,0 +1,36 @@
...
+#include <unordered_map>
+
+namespace multipass
+{
+
+struct MountConfig
+{
+    VirtualMachine* vm;
+    id_mappings uid_mappings;
+};
```
ricab, line 33: Thanks, this makes it much easier to understand (where this is used). 

You have it in a header with a view to use it for other mounts, correct? Do you know if this pattern fits the code for other mounts (existing or upcoming)?

sharder996: Yes, the idea was that if we have another kind of mount where we need some configuration data, we could just add additional fields to the struct.

I say that, but I don't have any insight fits the requirements for other mounts. I had decided not to merge this with [SSHFSServerConfig](https://github.com/canonical/multipass/blob/765d09de8043f69a4beaf4a9404a68cddd60cf80/include/multipass/sshfs_server_config.h#L29) as I feel SSHFS is too different with it's need to spawn a separate process for each mount.

luis4a0:
Hey @sharder996! We are almost there. I have two comments, though.

First, I am seeing extra newline characters in the CLI output. For instance:
```
$ multipass launch jammy









Launched: purified-raccoon
```
(nine extra newlines) and
```
$ multipass mount -t native snap purified-raccoon:snap1

                                                                                                                                                              
$
```
(two extra newlines).

On the other hand, I see the following behavior when mounting and unmounting without starting the VM between these operations:
```
$ multipass mount -t native snap purified-raccoon:snap1
$ multipass umount purified-racoon
umount failed:
```
(fails with no reason). Is this expected?

Thanks!

luis4a0 (approved):
Hey @sharder996, thanks a lot for this wonderful work! After many tests and review rounds, I think this is ready for merging. I believe the code uncovered by tests is really hard to cover now, so we can leave it for the future.

luis4a0:
bors cancel

luis4a0:
Ok, everything seems to be fine. Let's rock.

bors r+

luis4a0:
bors cancel

luis4a0:
Shall retry. Nevermind.

luis4a0:
Last retry before examining what's happening in the CI.

bors retry

townsend2010:
bors retry

townsend2010:
Ok, a Windows integration test failed which is unrelated to this, so merging manually.

PR #2779, opened 2022-10-14
[cli] Remove padded chars in cli output

### LINKED ISSUE #958: `multipass launch` pads its output to 80 characters with whitespace
You can't _see_ this per se, but copying/pasting whole lines from my console demonstrates the problem:

```shell
% multipass launch bionic
Launched: steady-koel                                                           
```

```python3
>>> len("Launched: steady-koel                                                           ")
80
```

I would expect no trailing whitespace.

sharder996:
Use [ANSI escape characters](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797) to manipulate cursor and overwrite test in the cli. 

---

public side of [pr512](https://github.com/canonical/multipass-private/pull/512)

Fixes #958

### DIFF
--- a/include/multipass/cli/client_platform.h
+++ b/include/multipass/cli/client_platform.h
@@ -44,6 +44,7 @@ public:
     Platform(const Singleton::PrivatePass&) noexcept;
 
     virtual std::pair<std::string, std::string> get_user_password(Terminal* term) const;
+    virtual void enable_ansi_escape_chars() const;
 };
 
 void parse_transfer_entry(const QString& entry, QString& path, QString& instance_name);
--- a/src/client/cli/CMakeLists.txt
+++ b/src/client/cli/CMakeLists.txt
@@ -17,6 +17,7 @@ add_library(client STATIC
   client.cpp)
 
 target_link_libraries(client
+  client_platform
   commands
   fmt
   formatter
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -45,6 +45,7 @@
 
 #include <multipass/cli/argparser.h>
 #include <multipass/cli/client_common.h>
+#include <multipass/cli/client_platform.h>
 #include <multipass/constants.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
@@ -103,6 +104,8 @@ mp::Client::Client(ClientConfig& config)
     add_command<cmd::Version>();
 
     sort_commands();
+
+    MP_CLIENT_PLATFORM.enable_ansi_escape_chars();
 }
 
 void mp::Client::sort_commands()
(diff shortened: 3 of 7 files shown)


sharder996:
Works on Linux. Need to verify on Windows/MacOS.

andrei-toterman (changes requested):
I don't think this has much to do with the operating system itself, but with the terminal application that is used. The new Windows Terminal supports ANSI codes out of the box. And, if set up correctly, both Command Prompt and PowerShell support ANSI codes (at least in Windows 10 versions older than 2016). For example, the terminal integrated in VSCode enables ANSI support for them
image

You can look into [`GetStdHandle`](), [`GetConsoleMode`](https://learn.microsoft.com/en-us/windows/console/getconsolemode), [`SetConsoleMode`](), and `ENABLE_VIRTUAL_TERMINAL_PROCESSING` to try to enable ANSI codes on Windows, and fall back to the `std::setw` option if that fails. It could be something similar to this
```cpp
// in AnimatedSpinner ctor
auto terminal_handle = GetStdHandle(STD_OUTPUT_HANDLE);
unsigned long mode;
auto supports_ansi = GetConsoleMode(terminal_handle, &mode) &&
                     ((mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING) ||
                     SetConsoleMode(terminal_handle, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING));

// in clear_line
if (supports_ansi) {
    // ansi stuff
} else {
    // fallback to std::setw
}
```

On Unix-like systems, checking the `TERM` environment variable is a common way of checking if the terminal supports ANSI codes.

I'm really not sure how to handle the situation in which `stdout` is not a terminal, i.e. when piping the output into another program. But I don't think that is in the scope of this PR, since right now the output is a mess and I don't see any people complaining about it.

andrei-toterman:
I think abstracting this into a `bool terminal_supports_ansi()` method of `MP_PLATFORM` would be a good idea :)

sharder996:
@andrei-toterman This is ready for review again!

Added a platform function for enabling ANSI codes on Windows which is going to fail when redirecting output to a file. But, that output was a already a mess in the first place and choked full of whitespace and newlines. 
The minimum Windows OS of Multipass is Windows 10 and I think expecting users will have installed an update from 2016 is pretty reasonable which is why I've chosen to ignore the return value from `SetConsoleMode()`. The alternative being to save the output and choose how to clear a line based on it. However, getting the value from the constructor of the client to the animated spinner would not be straightforward as the console mode only needs to set once per process, so setting it in the animated spinner would be a bit overkill.

andrei-toterman (approved):
Hey, @sharder996! Thanks for this!

bors merge

andrei-toterman:
bors retry

PR #2782, opened 2022-10-17
Update to qt6

### LINKED ISSUE #2627: Update to Qt6
Update to use QT ~5.15.5~ 5.15.3 or later
https://www.qt.io/blog/commercial-lts-qt-5.15.5-released

Contains a fix for this proxy issue: 
https://github.com/canonical/multipass/issues/2325#issuecomment-1097681715

andrei-toterman:
Fixes #2627


include/multipass/standard_paths.h
```diff
@@ -45,7 +45,7 @@ class StandardPaths : public Singleton<StandardPaths>
     static constexpr auto PicturesLocation = StandardLocation::PicturesLocation;
     static constexpr auto TempLocation = StandardLocation::TempLocation;
     static constexpr auto HomeLocation = StandardLocation::HomeLocation;
-    static constexpr auto DataLocation = StandardLocation::DataLocation;
+    static constexpr auto DataLocation = StandardLocation::AppDataLocation;
```
ricab, line 48: I understand this saves us from changing client code everywhere, but I think we should stick to what we're wrapping. These wrappers exist only for the purpose of testing and I believe we should keep them as thin and faithful to the original as possible. The moment we start mapping things differently or adding functionality, they themselves would need to be tested.

src/settings/persistent_settings_handler.cpp
```diff
@@ -29,10 +29,7 @@ namespace
 {
 std::unique_ptr<mp::WrappedQSettings> persistent_settings(const QString& filename)
 {
-    auto ret = mp::WrappedQSettingsFactory::instance().make_wrapped_qsettings(filename, QSettings::IniFormat);
-    ret->setIniCodec("UTF-8");
```
ricab, line 33: I suppose it's UTF-8 by default now, is it?

src/daemon/daemon.cpp
```diff
@@ -2684,16 +2682,17 @@ void mp::Daemon::init_mounts(const std::string& name)
...
-    auto future_watcher = async_future_watchers.back().get();
-    QObject::connect(future_watcher, &QFutureWatcher<AsyncOperationStatus>::finished,
-                     [this, future_watcher, finished_op] {
-                         finished_op();
-                         finish_async_operation(future_watcher->future());
-                     });
+    QObject::connect(future_watcher_p, &QFutureWatcher<AsyncOperationStatus>::finished, [this, uuid, finished_op] {
+        finished_op();
+        finish_async_operation(uuid);
+    });
```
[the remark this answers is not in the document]
andrei-toterman: Previously, we just kept some `QFutureWatcher`s in a `std::vector` and removed them in `finish_async_operation` based on the equality of the `QFuture` that they store. But in Qt6, the equality operator for `QFuture` [was removed](https://doc.qt.io/qt-6/qtcore-changes-qt6.html#equality-operators), and there was no way of comparing the futures directly anymore. So I tried coming up with a way to ensure that when a `QFutureWacther` is created, there is a unique way of identifying it when it needs to be removed, and UUIDs seemed good enough

src/settings/persistent_settings_handler.cpp
andrei-toterman: [yup](https://doc.qt.io/qt-6.2/qsettings.html#compatibility-with-older-qt-versions)

include/multipass/standard_paths.h
andrei-toterman: I see your point. Now that I look closer, `DataLocation` is not used anywhere in our code and the `QStandardPaths::StandardLocation::DataLocation` has been removed in favor of `QStandardPaths::StandardLocation::AppDataLocation` and we already have a `AppDataLocation` constant that wraps that value. So I guess the `DataLocation` can be removed entirely, right?

src/daemon/daemon.cpp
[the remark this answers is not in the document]
ricab: Got it, thanks for the explanation.

src/settings/persistent_settings_handler.cpp
ricab: Cool, thx!

include/multipass/standard_paths.h
ricab: Oh ok, yep I think it should be removed if it is not available in the *wrappee* anymore.

townsend2010 (changes requested):
Hey @andrei-toterman!

Thanks for this! Unfortunately, the tray icon GUI does not work for me:
```
$ multipass.gui 
Warning: Ignoring WAYLAND_DISPLAY on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
qt.qpa.plugin: Could not find the Qt platform plugin "wayland" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Aborted (core dumped)
```

andrei-toterman:
Hey, @townsend2010! Yeah, apparently I didn't stage some things properly. Should be fixed now.

townsend2010:
Hrm, it's failing the Windows build now: https://github.com/canonical/multipass-private/actions/runs/4045828641/jobs/6957841208

townsend2010 (approved):
This fails building on the private side due to the private side change in the workflow.

I'm gonna `bors` this to get the commit message, but it will fail, so I'll just merge manually.

townsend2010:
bors merge

PR #2783, opened 2022-10-19
Use `--dhcp-ignore-clid` option for `dnsmasq`

andrei-toterman:
This is a follow-up to #2717 which required a more convoluted fix to get the correct IP address based only on the MAC address of an instance. In a newer version of `dnsmasq` available in core22, there is the `--dhcp-ignore-clid` option which does just that.


townsend2010 (approved):
Hey @andrei-toterman!

Works well and much cleaner than the old solution.  Let's get it in!

bors merge

townsend2010:
Major outage that should be fixed now.  Try again...

bors retry

townsend2010:
bors retry

townsend2010:
bors retry

PR #2786, opened 2022-10-20
Print client warnings by default

luis4a0:
In order to show the user messages related to Hyperkit deprecation, the client must print warnings. We also need to decide which warnings are worth printing and which are not.

Implements https://github.com/canonical/multipass/projects/15#card-85698674

### DIFF
--- a/src/client/cli/argparser.cpp
+++ b/src/client/cli/argparser.cpp
@@ -85,10 +85,8 @@ auto verbosity_level_in(const QStringList& arguments)
             return 1;
         if (arg == QStringLiteral("-vv"))
             return 2;
-        if (arg == QStringLiteral("-vvv"))
+        if (QRegExp{"-vvv+"}.exactMatch(arg))
             return 3;
-        if (QRegExp{"-vvvv+"}.exactMatch(arg))
-            return 4;
     }
     return 0;
 }
@@ -117,7 +115,7 @@ mp::ParseCode mp::ArgParser::parse(const std::optional<mp::AliasDict>& aliases)
     QCommandLineOption help_option(help_option_names, "Displays help on commandline options");
     QCommandLineOption verbose_option({"v", "verbose"},
                                       "Increase logging verbosity. Repeat the 'v' in the short option for more detail. "
-                                      "Maximum verbosity is obtained with 4 (or more) v's, i.e. -vvvv.");
+                                      "Maximum verbosity is obtained with 3 (or more) v's, i.e. -vvv.");
     QCommandLineOption version_option({"V", "version"}, "Show version details");
     version_option.setFlags(QCommandLineOption::HiddenFromHelp);
     parser.addOption(help_option);
@@ -374,9 +372,9 @@ QStringList mp::ArgParser::unknownOptionNames() const
 
 void mp::ArgParser::setVerbosityLevel(int verbosity)
 {
-    if (verbosity > 4 || verbosity < 0)
+    if (verbosity > 3 || verbosity < 0)
     {
-        cerr << "Verbosity level is incorrect. Must be between 0 and 4.\n";
+        cerr << "Verbosity level is incorrect. Must be between 0 and 3.\n";
     }
     else if (verbosity_level != verbosity)
     {
--- a/tests/test_argparser.cpp
+++ b/tests/test_argparser.cpp
@@ -57,7 +57,7 @@ TEST_P(TestVerbosity, test_various_vs)
     auto parser = mp::ArgParser{args, cmds, oss, oss};
     parser.parse();
 
-    const auto expect = v > 4 ? 4 : v;
+    const auto expect = v > 3 ? 3 : v;
     EXPECT_EQ(parser.verbosityLevel(), expect);
 }
 
--- a/tests/test_argparser.cpp
+++ b/tests/test_argparser.cpp
@@ -40,6 +40,18 @@ struct TestVerbosity : public TestWithParam<int>
 {
 };
 
+TEST_F(TestVerbosity, verbosity_is_zero_with_no_arguments)
+{
+    std::ostringstream oss;
+    const auto cmds = std::vector<mp::cmd::Command::UPtr>{};
+    auto args = QStringList{"multipass_tests"};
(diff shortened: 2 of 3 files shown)


src/client/cli/argparser.cpp
```diff
@@ -85,10 +85,8 @@ auto verbosity_level_in(const QStringList& arguments)
             return 1;
         if (arg == QStringLiteral("-vv"))
             return 2;
-        if (arg == QStringLiteral("-vvv"))
+        if (QRegExp{"-vvv+"}.exactMatch(arg))
             return 3;
-        if (QRegExp{"-vvvv+"}.exactMatch(arg))
-            return 4;
     }
     return 0;
```
ricab, line 91: I think this needs to change to `return 1`.

ricab (changes requested):
Hey @luis-penaranda, thank you for this. It is mostly there, but I couldn't get it to actually print any warnings by default quite yet. For instance, if I disconnect from the internet and try to launch, I see:

```
$ mp launch
launch failed: Remote "" is unknown or unreachable.
```

But if I provide `-v`:

```
$ mp launch -v
[2022-10-26T14:29:49.574] [warning] [url downloader] Cannot retrieve headers for https://cdimage.ubuntu.com/ubuntu-core/16/stable/current/ubuntu-core-16-amd64.img.xz: Host cdimage.ubuntu.com not found
[2022-10-26T14:29:49.578] [warning] [VMImageHost] Could not update manifest: failed to download from 'https://cdimage.ubuntu.com/ubuntu-core/16/stable/current/ubuntu-core-16-amd64.img.xz': Host cdimage.ubuntu.com not found
[2022-10-26T14:29:49.579] [warning] [url downloader] Cannot retrieve headers for https://cloud-images.ubuntu.com/buildd/releases/bionic/release/bionic-server-cloudimg-amd64-disk.img: Host cloud-images.ubuntu.com not found
[2022-10-26T14:29:49.579] [warning] [VMImageHost] Could not update manifest: failed to download from 'https://cloud-images.ubuntu.com/buildd/releases/bionic/release/bionic-server-cloudimg-amd64-disk.img': Host cloud-images.ubuntu.com not found
launch failed: Remote "" is unknown or unreachable.
```

src/client/cli/argparser.cpp
luis4a0: In fact, this return line is never reached, because the function is called only when verbosity is set. I prefer to leave it as zero, for considering the case on which the function is called when verbosity is not set. WDYT?

ricab: Sorry, I had not realized that the other levels would need to change too. I think this function should provide the verbosity corresponding to a command line, in any case (even if it is not used in all cases). So I suppose something like the following?

```c++
auto verbosity_level_in(const QStringList& arguments)
{
    for (const QString& arg : arguments)
    {
        if (arg == QStringLiteral("-v") || arg == QStringLiteral("--verbose"))
            return 2;
        if (arg == QStringLiteral("-vv"))
            return 3;
        if (arg == QStringLiteral("-vvv+"))
            return 4;
    }
    return 1;
}
```

It would be inconsistent to return 0 in the default case then.

luis4a0: But that `return 1` is never reached. To be clear, let's abstract from this function and state which the levels should be:
- no verbosity arguments should be level 1
- `-v` or `--verbose` should be level 2
- `-vv` should be level 3
- `-vvv` or more should be level 4
Is that what you had in mind?

Additionally, we can count in the future for an option `--quiet` to force verbosity level 0.

luis4a0: Oh, right, sorry I didn't see that you are proposing that this function must be always called. +1 to that. So we are speaking the same language.

ricab: Yes, that's what I think we need. 

I understand that the last return is never reached *today*. However, a function should provide a consistent unit of functionality, without assuming how it is called. According to the function name, that functionality is to provide the `verbosity_level_in` a cmd line (as represented by a list of arguments). 

If there are no `-v`s, the verbosity in the command line should now become 1. If we add `--quiet` in the future, this function would then need to take that into account.

ricab: Oh, crossed msgs. Well, I think ideally it would make sense for this function to be the only unit determining the verbosity. I don't need that to be the case with this PR, but if it is a small change, I suppose that doesn't harm.

luis4a0: Of course, it doesn't harm.

ricab:
Hey @luis4a0, brace yourself... I am sorry, I am afraid there's a little u-turn on this... So looking at the output yesterday in my room I realized I wasn't happy with where we'd be going with this whole increased verbosity thing. We've discussed here too and I believe we need to abort this, sorry.

My reasoning for this was that, if we had a warning, we'd better show it to the users, or we wouldn't be warning them of anything. But then looking at those messages, I realized that we actually have different targets for the warnings we want to have the daemon log and the messages we display in the client. Like the example I posted above, with the network being disconnected. Currently we have a simple, clean, high-level message informing the user of the problem. With the increased verbosity though, it gets a little messy, without being very informative to the user (the user shouldn't have to care about VMImageHost details). What I thought before was that we would decrease those messages to info, but that's not really very good, because we still want them to stand out in the logs. The user is one target, caring only about high level stuff. We, or anyone investigating the logs, are a different target, with a need for low-level and implementation details.

When I wrote the spec, I thought that this was a technical requirement for us to be able to tell the user to switch to QEMU, but I have since understood that log forwarding to the client better and I realized that we can just bypass the logging facility for this entirely, without having to change anything in the client.

I am really sorry for having wasted your time on this. Let me know if you want to discuss it, or if there is anything I can explain better!

luis4a0:
Hey @ricab! Don't worry at all! I completely understand, and I also see that this is a part of a huge feature. On the contrary, I'm happy I could read a part of code I didn't before :)

On the other hand, I'd be happy to help if you need something more for this feature!

townsend2010:
As indicated by @ricab, this is no longer needed.

PR #2793, opened 2022-10-27
[daemon] Add alias to pollinate

townsend2010:


townsend2010:
@nathan-j-hart

src/daemon/daemon.cpp
```diff
@@ -114,21 +114,38 @@ mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
...
+    config["timezone"] = request->time_zone();
     config["system_info"]["default_user"]["name"] = username;
 
     auto pollinate_user_agent_string =
         fmt::format("multipass/version/{} # written by Multipass\n", multipass::version_string);
     pollinate_user_agent_string += fmt::format("multipass/driver/{} # written by Multipass\n", backend_version_string);
     pollinate_user_agent_string +=
         fmt::format("multipass/host/{} # written by Multipass\n", multipass::platform::host_version());
+    pollinate_user_agent_string += fmt::format("multipass/alias/{}{} # written by Multipass\n",
+                                               !remote_name.empty() ? remote_name + ":" : "", pollinate_alias);
```
townsend2010, line 148: Hey @nathan-j-hart, this is capturing the `remote` (if provided by the user) and the "alias" used to `launch` the instance.  Is that good for what you need?

nathan-j-hart: where `remote` is whether it's a URL or file? alias is the biggest need here so looks good from that standpoint

townsend2010: `remote` (if provided) will tell us if it's say, for example, a `snapcraft:core18` launch (which is a buildd image for building snaps) vs. a `core18` launch (which is the actual Ubuntu Core 18 distro).  This helps discern this ambiguity.

townsend2010: Also, `alias` will either be `http` or `file` only if it's those types of launches.  No other part of the url will be part of the string.

nathan-j-hart: Right, ok in that case this looks great to me

src/daemon/daemon.cpp
```diff
@@ -114,21 +114,38 @@ mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
...
+                                   const std::string& backend_version_string, const mp::CreateRequest* request)
 {
     auto ssh_key_line = fmt::format("ssh-rsa {} {}@localhost", key_provider.public_key_as_base64(), username);
+    QString pollinate_alias = QString::fromStdString(request->image());
 
+    if (pollinate_alias.isEmpty())
+    {
+        pollinate_alias = "default";
+    }
+    else if (pollinate_alias.startsWith("http"))
```
andrei-toterman, line 127: not to be too pedantic, but shouldn't this check for `http://` and `https://`?

src/daemon/daemon.cpp
```diff
@@ -114,21 +114,38 @@ mp::Query query_from(const mp::LaunchRequest* request, const std::string& name)
...
+    {
+        pollinate_alias = "http";
+    }
+    else if (pollinate_alias.startsWith("file"))
```
andrei-toterman, line 131: same here, `file://` instead of `file`

andrei-toterman (commented):
Alright, looks good! Only a minuscule question inline.

tests/test_daemon.cpp
```diff
@@ -592,18 +601,22 @@ TEST_P(DaemonCreateLaunchTestSuite, adds_ssh_keys_to_cloud_init_config)
...
         }));
 
-    send_command({GetParam()});
+    const auto [command, _] = GetParam();
+    send_command({command});
 }
 
 TEST_P(DaemonCreateLaunchTestSuite, adds_pollinate_user_agent_to_cloud_init_config)
 {
+    const auto [command, alias] = GetParam();
```
andrei-toterman, line 610: also since this is the only place where the alias from `GetParam()` is used, wouldn't it be cleaner to just create a new `DaemonCreateLaunchAliasTestSuite` that implements `WithParamInterface<std::tuple<std::string, string>>` to be used here and to not modify all the other tests to ignore the alias?

src/daemon/daemon.cpp
townsend2010: Actually, upon further consideration, I think this should stay due to this already occurring in the following lines:
https://github.com/canonical/multipass/blob/release/1.11/src/daemon/daemon.cpp#L109-L112

andrei-toterman: okay, makes sense!

townsend2010:
Yay, it passed ci!!!!  @andrei-toterman, please `bors` it at your convenience:)

andrei-toterman (approved):
Nice!

bors merge

Saviq:
bors retry

Saviq:
bors retry

Saviq:
bors retry

townsend2010:
bors retry

PR #2802, opened 2022-11-01
Fix default UID/GID mappings.

### LINKED ISSUE #2801: Wrong UID or GID mappings in mounts
**Describe the bug**
When a mount command specifies only one of UID or GID maps, it does not set a default value for the unspecified one.

**To Reproduce**
`multipass mount folder instance --gid-map=501:8`
Inside the instance, the files and folders inside the directory are owned by an user with UID equal to the UID of the user executing the client on the host.

**Expected behavior**
The default instance user (usually `ubuntu` with UID  `1000`) should own the files in the mounted directory in the instance.

luis4a0:
The client specified default maps only when both UID, GID maps were not specified. It failed in case one and only one of the two was specified.

Fixes #2801.

### DIFF
--- a/src/client/cli/cmd/mount.cpp
+++ b/src/client/cli/cmd/mount.cpp
@@ -241,6 +241,15 @@ mp::ParseCode cmd::Mount::parse_args(mp::ArgParser* parser)
             }
         }
     }
+    else
+    {
+        mpl::log(mpl::Level::debug, category,
+                 fmt::format("{}:{} {}(): adding default uid mapping", __FILE__, __LINE__, __FUNCTION__));
+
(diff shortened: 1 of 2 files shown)


sharder996 (approved):
LGTM

bors merge

luis4a0:
Hey @sharder996, thanks for the review! I added a couple tests, should be ready for reviewing again now.

sharder996:
Still lgtm

bors r+

Saviq:
bors retry

Saviq:
bors retry

Saviq:
bors retry

PR #2810, opened 2022-11-03
Accept decimal memory values

### LINKED ISSUE #2677: Multipass should accept decimals back for memory values
**Describe the bug**

When retrieving the memory size of an instance using "get", the number returned is a float with one decimal place, of which the decimal is always zero.

**To Reproduce**
How, and what happened?
1. `multipass get local.example.memory`
1. It returns "`256.0MiB`".

**Expected behavior**

That it returns "`256M`" or "`256MiB`".

**Logs**

**Additional info**
 - OS: macOS 12.4
- `multipass version` 1.10.1
- `multipass info --all`
```
Name:           example
State:          Stopped
IPv4:           --
Release:        --
Image hash:     e1ce033239f0 (Ubuntu 22.04 LTS)
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --
```

**Additional context**
Trying to set back the same number, results in an error:
```

multipass set local.example.memory=256.0MiB
set failed: Invalid setting 'local.s1.memory=256.0MiB': 256.0MiB is not a valid memory size - need a non-negative integer (in base 10) followed by the desired unit (e.g. 1234B, 42MiB)

```
Dropping the `.0` decimal works:
`multipass set local.example.memory=256MiB`

I don't know C++, but to me this looks like where it is done: https://github.com/canonical/multipass/blob/v1.10.1/src/utils/memory_size.cpp#L130-L131

sharder996:
Adds the ability to input decimal amounts for memory specification.

Fixes #2677

### DIFF
--- a/include/multipass/exceptions/invalid_memory_size_exception.h
+++ b/include/multipass/exceptions/invalid_memory_size_exception.h
@@ -29,8 +29,8 @@ class InvalidMemorySizeException : public std::runtime_error
 {
 public:
     InvalidMemorySizeException(const std::string& val)
-        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative decimal (in base 10) "
-                                    "followed by the desired unit (e.g. 1234B, 42MiB, 0.5G)",
+        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative integer (in base 10) "
+                                    "or a decimal followed by K, M, or G (e.g. 1234B, 42MiB, 0.5G)",
                                     val))
     {
     }
--- a/include/multipass/exceptions/invalid_memory_size_exception.h
+++ b/include/multipass/exceptions/invalid_memory_size_exception.h
@@ -29,8 +29,8 @@ class InvalidMemorySizeException : public std::runtime_error
 {
 public:
     InvalidMemorySizeException(const std::string& val)
-        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative integer (in base 10) "
-                                    "followed by the desired unit (e.g. 1234B, 42MiB)",
+        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative decimal (in base 10) "
+                                    "followed by the desired unit (e.g. 1234B, 42MiB, 0.5G)",
                                     val))
     {
     }
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -204,8 +204,8 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                   "cpus", default_cpu_cores);
     QCommandLineOption diskOption(
         {"d", "disk"},
-        QString::fromStdString(fmt::format("Disk space to allocate. Positive decimals, in "
-                                           "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
+        QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
+                                           "bytes, or decimals, with K, M, G suffix.\nMinimum: {}, default: {}.",
                                            min_disk_size, default_disk_size)),
         "disk", QString::fromUtf8(default_disk_size));
 
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -211,8 +211,8 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
 
     QCommandLineOption memOption(
         {"m", "memory"},
-        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive decimals, "
-                                           "in bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
+        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive integers, "
+                                           "in bytes, or decimals, with K, M, G suffix.\nMinimum: {}, default: {}.",
                                            min_memory_size, default_memory_size)),
         "memory", QString::fromUtf8(default_memory_size)); // In MB's
     QCommandLineOption memOptionDeprecated(
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -204,14 +204,14 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                   "cpus", default_cpu_cores);
     QCommandLineOption diskOption(
         {"d", "disk"},
-        QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
+        QString::fromStdString(fmt::format("Disk space to allocate. Positive decimals, in "
                                            "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
                                            min_disk_size, default_disk_size)),
         "disk", QString::fromUtf8(default_disk_size));
 
     QCommandLineOption memOption(
         {"m", "memory"},
-        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive integers, "
+        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive decimals, "
                                            "in bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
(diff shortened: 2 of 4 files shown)


ricab:
Hey @sharder996, could you please rebase this, since the whitespace part was done separately? Thanks.

tests/test_memory_size.cpp
```diff
@@ -189,8 +230,13 @@ TEST(MemorySize, canCompareEqual)
...
     EXPECT_EQ(mp::MemorySize{"2048"}, mp::MemorySize{"2k"});
     EXPECT_EQ(mp::MemorySize{"2g"}, mp::MemorySize{"2048M"});
-    EXPECT_EQ(mp::MemorySize{"2g"}, mp::MemorySize{"2048M"});
     EXPECT_EQ(mp::MemorySize{"0m"}, mp::MemorySize{"0k"});
+    EXPECT_EQ(mp::MemorySize{"1.5G"}, mp::MemorySize{"1536M"});
+    EXPECT_EQ(mp::MemorySize{"1.0K"}, mp::MemorySize{"1024B"});
+    EXPECT_EQ(mp::MemorySize{"1.0K"}, mp::MemorySize{"1k"});
+    EXPECT_EQ(mp::MemorySize{"3.14K"}, mp::MemorySize{"3215"});
+    EXPECT_EQ(mp::MemorySize{"0.0001G"}, mp::MemorySize{"107374"});
+    EXPECT_EQ(mp::MemorySize{"0.095367432K"}, mp::MemorySize{"97B"});
```
ricab, line 239: Nice, you followed the same flooring approach in this case. However, if we accept something like this and discard fractional bytes, why not accept fractional bytes and react the same way? You could then add another test in the sequence of convertsLowerUnitByFlooring (e.g. check that 7.89B, in bytes, is 7).

Alternatively, we could refuse any input that implied fractional bytes, but that would be problematic because we'd then have to refuse something like 3.3K.

tests/test_memory_size.cpp
```diff
@@ -109,8 +109,9 @@ TEST_P(TestBadMemorySizeFormats, rejectsBadFormats)
 
 INSTANTIATE_TEST_SUITE_P(MemorySize, TestGoodMemorySizeFormats, ValuesIn(TestGoodMemorySizeFormats::generate_args()));
```
ricab, line 110: Could you also update `generate_args` to include some fractional sizes?

tests/test_memory_size.cpp
```diff
@@ -109,8 +109,9 @@ TEST_P(TestBadMemorySizeFormats, rejectsBadFormats)
...
 INSTANTIATE_TEST_SUITE_P(MemorySize, TestBadMemorySizeFormats,
-                         Values("321BB", "321BK", "1024MM", "1024KM", "1024GK", "K", "", "123.321K", "123.321", "6868i",
-                                "555iB", "486ki", "54Mi", "8i33", "4M2", "-2345", "-5MiB", "K", "4GM"));
+                         Values("321BB", "321BK", "1024MM", "1024KM", "1024GK", "K", "", "123.321", "6868i", "555iB",
+                                "486ki", "54Mi", "8i33", "4M2", "-2345", "-5MiB", "K", "4GM", "256.M", "186000.B",
+                                "3.14", ".5g", "4.2B", "42.", "2048.K", " 268. "));
```
ricab, line 114: While you're at it, why not accept `1.` and `.1` too? They're usually acceptable ways to input floats. You would have to protect the regexp against just `.` though. This is less relevant though, so leave it alone if you find it too troublesome.

ricab (changes requested):
Hey @sharder996, impressive stuff! However, at this point I think you'd be better off just separating the unit and then using `std::stof` on the rest. That would have been preferable for integers too, but floats are much trickier to parse.

I focused the review on the tests, to understand what contract you implemented. I think there are a couple of respects that can still be improved. The main one concerns handling of sizes with fractional bytes (details inline).

In any case, this is largely going in the right direction, thanks for your work on this!

tests/test_memory_size.cpp
sharder996: Done

sharder996:
Hey @ricab, thanks for the review! Couple notes in reply to your comments.

The motivation behind not capturing the input as a floating point number was to make a best effort in preserving as much precision as possible. 

I had wanted to to reject blatant fractional bytes, such as `1.1B`, since it is pretty obvious that bytes are atomic in terms of computer memory. When inputting higher order numbers, such as `3.3K`, I think it's fine to have implied fractional bytes (and actually be flooring the number) since the user is already giving an approximation of the memory they want to allocate.

The other two forms, `.5G` and `512.M`, I think can have an argument made for them as they can be considered numerically valid. However, the appearance of the latter still repels me.

ricab (changes requested):
Alright, looking great. Only a couple of things missing:
  - updating the message on `InvalidMemorySizeException`;
  - updating the help for `launch`.
  
I will go ahead and update the settings keys for memory and disk too. It won't be in effect for a while, but that's just how our docs work.

src/client/cli/cmd/launch.cpp
```diff
@@ -204,14 +204,14 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                   "cpus", default_cpu_cores);
     QCommandLineOption diskOption(
         {"d", "disk"},
-        QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
+        QString::fromStdString(fmt::format("Disk space to allocate. Positive decimals, in "
                                            "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
```
ricab, line 208: How about

```suggestion
        QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
                                           "bytes, or decimals, with K, M, G suffix.\nMinimum: {}, default: {}.",
```

sharder996: That would be more specific and possibly avoid some future confusion :)

src/client/cli/cmd/launch.cpp
```diff
@@ -205,13 +205,13 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
         QString::fromStdString(fmt::format("Disk space to allocate. Positive integers, in "
-                                           "bytes, or with K, M, G suffix.\nMinimum: {}, default: {}.",
+                                           "bytes, or decimals, with K, M, G suffix.\nMinimum: {}, default: {}.",
                                            min_disk_size, default_disk_size)),
         "disk", QString::fromUtf8(default_disk_size));
 
     QCommandLineOption memOption(
         {"m", "memory"},
-        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive integers, "
+        QString::fromStdString(fmt::format("Amount of memory to allocate. Positive decimals, "
```
ricab, line 214: This one too please :)

include/multipass/exceptions/invalid_memory_size_exception.h
```diff
@@ -29,8 +29,8 @@ class InvalidMemorySizeException : public std::runtime_error
 {
 public:
     InvalidMemorySizeException(const std::string& val)
-        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative integer (in base 10) "
-                                    "followed by the desired unit (e.g. 1234B, 42MiB)",
+        : runtime_error(fmt::format("{} is not a valid memory size - need a non-negative decimal (in base 10) "
```
ricab, line 32: And I think this needs rephrasing, otherwise it can get confusing:

```plain
$ multipass set local.teaching-sculpin.memory=1.1B
set failed: Invalid setting 'local.teaching-sculpin.memory=1.1B': 1.1B is not a valid memory size - need a non-negative decimal (in base 10) followed by the desired unit (e.g. 1234B, 42MiB, 0.5G)
```

ricab: I am thinking perhaps: "need a non-negative integer, possibly followed by 'B', or a decimal followed by a larger unit (e.g. 1234B, 42MiB, 0.5G)"

ricab (commented):
Alright, great job! Let's get this in.

bors r+

ricab (approved):

PR #2824, opened 2022-11-16
Verify launch mount source path before launching

### LINKED ISSUE #2817: [cli/launch] `--mount` option should check if source directories exist and abort launch if they don't
Related to #2815 

Right now, when doing `launch --mount <source>:<target>`, the instance is launched and then `<source>` and `<target>` are passed to the `mount` command. If `<source>` does not exist, the mount will fail, but the instance will already be launched. We should check ahead of time if the requested source directories for mounting exist, and abort the entire launch if they don't.

sharder996:
Verify client source directories before launching an instance and fail the launch if they are invalid.

Fixes #2817

### DIFF
--- a/tests/test_cli_client.cpp
+++ b/tests/test_cli_client.cpp
@@ -1239,18 +1239,20 @@ TEST_F(Client, launchCmdMountOptionFailsOnInvalidDir)
                      trash_stream, err),
         mp::ReturnCode::CommandLineError);
     EXPECT_THAT(err.str(), HasSubstr(fmt::format("Mount source path \"{}\" does not exist", fake_source)));
+    err.clear();
 
     EXPECT_EQ(
         send_command({"launch", "--name", instance_name, "--mount", fmt::format("{}:{}", fake_source, fake_target)},
                      trash_stream, err),
         mp::ReturnCode::CommandLineError);
-    EXPECT_THAT(err.str(), HasSubstr(fmt::format("Mount source path \"{}\" does not exist", fake_source)));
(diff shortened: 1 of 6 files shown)


sharder996:
Unit test has been written! @andrei-toterman

tests/test_cli_client.cpp
```diff
@@ -1220,6 +1220,39 @@ TEST_F(Client, launch_cmd_mount_option)
...
+        send_command({"launch", "--name", instance_name, "--mount", fmt::format("{}:{}", fake_source, fake_target)},
+                     trash_stream, err),
+        mp::ReturnCode::CommandLineError);
+    EXPECT_THAT(err.str(), HasSubstr(fmt::format("Mount source path \"{}\" does not exist", fake_source)));
+
+    EXPECT_EQ(
+        send_command({"launch", "--name", instance_name, "--mount", fmt::format("{}:{}", fake_source, fake_target)},
+                     trash_stream, err),
+        mp::ReturnCode::CommandLineError);
+    EXPECT_THAT(err.str(), HasSubstr(fmt::format("Mount source path \"{}\" does not exist", fake_source)));
```
andrei-toterman, line 1253: hmm shouldn't these return different error messages? i think you should clear `err` before each `send_command` invocation, so you don't compare against errors from previous calls.

andrei-toterman (changes requested):

tests/test_cli_client.cpp
sharder996: Sorry, copy paste error! I also added a line to clear the stream to make sure that the previous output is gone.

andrei-toterman (approved):
thanks, looks good!

bors merge

andrei-toterman:
a moment of appreciation for bors working on the first try

PR #2825, opened 2022-11-16
Catch `INSTANCE_EXISTS` error from daemon

sharder996:
When launching an instance with a duplicate name, the multipass cli will by default print out the serialized binary string of the protobuf message resulting in extra newlines appearing in the console. 

Ex:

```
scott:~$ multipass launch -n foo                                            
Launched: foo                                                                   
scott:~$ multipass launch -n foo
launch failed: instance "foo" already exists                                    


scott:~$  
```

After changes:

```
scott:~$ multipass launch -n foo
launch failed: instance "foo" already exists                                    
Invalid instance name supplied: foo
scott::~$
```

### DIFF
--- a/src/client/common/client_common.cpp
+++ b/src/client/common/client_common.cpp
@@ -145,9 +145,7 @@ mp::ReturnCode mp::cmd::standard_failure_handler_for(const std::string& command,
                                                      const grpc::Status& status, const std::string& error_details)
 {
     fmt::print(cerr, "{} failed: {}\n{}", command, status.error_message(),
-               !error_details.empty()
-                   ? fmt::format("{}\n", error_details)
-                   : !status.error_details().empty() ? fmt::format("{}\n", status.error_details()) : "");
+               !error_details.empty() ? fmt::format("{}\n", error_details) : "");
 
     return return_code_for(status.error_code());
 }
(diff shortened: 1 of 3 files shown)


luis4a0:
Hi @sharder996, thanks for this! Now that you're focusing here, maybe it's better to return only the message `launch failed: instance "foo" already exists` if the instance exists and the message `Invalid instance name supplied: foo` in other failure cases. I think it's a matter of rewriting the conditions inside the `if`s, am I right?

sharder996:
Hi @luis4a0! I found the underlying issue which is that we are printing out a binary serialization of a protobuf message which can contain embedded null characters manifesting as newlines in the terminal.

src/client/common/client_common.cpp
```diff
@@ -145,9 +145,7 @@ mp::ReturnCode mp::cmd::standard_failure_handler_for(const std::string& command,
                                                      const grpc::Status& status, const std::string& error_details)
 {
     fmt::print(cerr, "{} failed: {}\n{}", command, status.error_message(),
-               !error_details.empty()
-                   ? fmt::format("{}\n", error_details)
-                   : !status.error_details().empty() ? fmt::format("{}\n", status.error_details()) : "");
```
ricab, line 150: Good find, although this returns a string, it has [only *binary* details](https://grpc.github.io/grpc/cpp/classgrpc_1_1_status.html#a3eb2cecab0655d1f74c191847332824d).

luis4a0: Hey @ricab, thanks for taking a look to this. What do you mean with this? The output is indeed correct now.

luis4a0 (approved):
Hey @sharder996! This works OK, so let's go. Thanks!

luis4a0:
bors retry

PR #2826, opened 2022-11-17
remove backticks in multipass info

### LINKED ISSUE #2819: `info` Disk Usage breaks when shell = fish
**Describe the bug**
When using fish as the shell for the default ubuntu user, the command `multipass info` doesn't show the disk usage:

```
multipass info fish-test
Name:           fish-test
State:          Running
IPv4:           192.168.64.28
Release:        Ubuntu 22.04.1 LTS
Image hash:     c363ebe92173 (Ubuntu 22.04 LTS)
Load:           0.31 0.08 0.03
Disk usage:     --
Memory usage:   151.8M out of 961.9M
Mounts:         --
```

**To Reproduce**
How, and what happened?
1. Launch a new multipass instance with this cloud init file:
    ```
    #cloud-config

    packages:
      - fish

    system_info:
      default_user:
        shell: /usr/bin/fish
    ```
1. Run `multipass info` on the newly created VM.

Running `multipass info -v` gives me the following error:

```
[warning] [utils] no output after running 'df --output=used `awk '$2 == "/" { print $1 }' /proc/mounts` -B1 | sed 1d'
```
I'm guessing this is due to the backticks?

**Expected behavior**
Expected the `info` command to include the disk usage of the VM.

**Logs**

**Additional info**
 - OS: macOS 12.6.1
- `multipass version`: 1.10.1+mac
- `multipass info --all`

**Additional context**

sharder996:
Replace backticks in shell commmands with posix compliant `$()`.

Fixes #2819

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1330,9 +1330,9 @@ try // clang-format on
             info->set_memory_usage(mpu::run_in_ssh_session(session, "free -b | grep 'Mem:' | awk '{printf $3}'"));
             info->set_memory_total(mpu::run_in_ssh_session(session, "free -b | grep 'Mem:' | awk '{printf $2}'"));
             info->set_disk_usage(mpu::run_in_ssh_session(
-                session, "df --output=used `awk '$2 == \"/\" { print $1 }' /proc/mounts` -B1 | sed 1d"));
+                session, "df --output=used $(awk '$2 == \"/\" { print $1 }' /proc/mounts) -B1 | sed 1d"));
             info->set_disk_total(mpu::run_in_ssh_session(
-                session, "df --output=size `awk '$2 == \"/\" { print $1 }' /proc/mounts` -B1 | sed 1d"));
+                session, "df --output=size $(awk '$2 == \"/\" { print $1 }' /proc/mounts) -B1 | sed 1d"));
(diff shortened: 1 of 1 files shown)


ricab (changes requested):
While I think this is correct, it doesn't quite solve the original issue yet:

```sh
$ multipass info handsome-ling
info failed: fish: Expected a variable name after this $.
df --output=used $(awk '$2 == "/" { print $1 }' /proc/mounts) -B1 | sed 1d
```

This is because:

```fish
ubuntu@handsome-ling ~> echo $(echo bla)
fish: $(...) is not supported. In fish, please use '(echo)'.
echo $(echo bla)
     ^
```

Reading around, fish does not try to be POSIX compliant at all... 

I guess multipass should not depend on what shell is configured, so maybe @luis4a0's idea of wrapping the whole thing with a `bash -c` would be the best option.

luis4a0:
> Reading around, fish does not try to be POSIX compliant at all...

Let me be again the devil's advocate here: in that case, does it make sense to support a non-POSIX shell?

On the other hand, as you mention, `/bin/bash -c` would solve the issue for any shell.

ricab (approved):
You're right @sharder996, let's get this in. Thanks!

bors r+

PR #2828, opened 2022-11-17
[tests] Use explicit template instantiation for templated fixture functions

townsend2010:
This should help some with the compile time and memory footprint needed to compile the unit tests.


ricab (approved):
Ohhh, the Egg of Columbus :clap: Great idea!

There was a build failure on macOS, but I think it's unrelated.

bors r+

Issue #2807, opened 2022-11-02
[regression] Repeated mount not refused when instance is stopped


ricab:
**To Reproduce**
How, and what happened?
1. `multipass launch -n test`
1. `multipass mount foo test:foo`
2. `multipass mount bar test:foo`
    ```
    mount failed: The following errors occurred:
    "test:foo" is already mounted
    ```
1. `multipass stop test`
2. `multipass mount --type native bar test:foo`
(no error)

**Expected behavior**
Multipass would refuse the repeated mount.

**Additional info**
 - OS: macOS 10.16
- `multipass version`: 1.11 RC
- `multipass get local.driver`: `qemu`

ricab:
Actually, simpler:

```
$ multipass launch -n test
$ multipass stop test
$ multipass mount a test:foo
$ multipass mount b test:foo
$ echo $?
0
```

townsend2010:
Does this also occur in 1.10.1?

ricab:
Nope, I just tried it and Multipass correctly rejected the mount in 1.10. So regression.

Issue #2797, opened 2022-11-01
[regression][snap] `multipass restart` fails to restart mounts


sharder996:
**Describe the bug**
When restarting an instance, sshfs exits unexpectedly and then fails to reestablish the mount. Previously defined sshfs mount is then missing from restarted instance.

**Edit**
Native mounts do not restart as well. `sudo mount` is not being executed after the instance restarts. Doing it manually fixed the issue.

**To Reproduce**
1. `multipass launch -n vm1`
2. `multipass mount . vm1:test-dir`
3. `multipass restart vm1`

**Expected behavior**
Multipass successfully reestablishes the mount and/or restarts the instance with previous mounts intact.

**Logs**
```
Nov 01 10:02:21 scott-ubuntu multipassd[55906]: Rebooting vm4
Nov 01 10:02:21 scott-ubuntu multipassd[55906]: Executing 'sudo systemctl stop ssh'
Nov 01 10:02:21 scott-ubuntu multipassd[55906]: Executing 'sudo reboot'
Nov 01 10:02:21 scott-ubuntu multipassd[55906]: Waiting for SSH to be up
Nov 01 10:02:21 scott-ubuntu sshfs_server[57788]: sshfs in the instance appears to have exited unexpectedly.  Trying to recover.
Nov 01 10:02:21 scott-ubuntu sshfs_server[57788]: Executing 'findmnt --source :/home/scott  -o TARGET -n'
Nov 01 10:02:21 scott-ubuntu sshfs_server[57788]: Caught an unhandled exception: unable to create a channel for remote process: 'findmnt --source :/home/scott  -o TARGET -n', the SSH session is not connected
Nov 01 10:02:21 scott-ubuntu multipassd[55906]: Removing "test-dir": ssh connection failed: 'Socket error: Connection reset by peer'
Nov 01 10:02:23 scott-ubuntu multipassd[55906]: QMP: {"timestamp": {"seconds": 1667293343, "microseconds": 93395}, "event": "RESET", "data": {"guest": true, "reason": "guest-reset"}}
Nov 01 10:02:23 scott-ubuntu multipassd[55906]: VM restarting
Nov 01 10:02:23 scott-ubuntu multipassd[55906]: Waiting for SSH to be up
[...]

PR #2791, opened 2022-10-25
[mounts] Fix mount handling when suspending

townsend2010:
Fixes #2790

PR #2835, opened 2022-11-23
Mount handlers refactor

andrei-toterman:
This PR is a refactor of the mount handlers. Instead of having each mount handler keep track of all the mounts of a specific type, we will have each instance of a mount handler class represent a single mounted path. The daemon will keep track of these mounts and, since each mount can be thought of as a resource, we can use RAII to manage their creation and destruction to some extent. Previous tests have been removed in order to not block the next release, but they will be rewritten as soon as possible.

---

- [x] linux:
  * [x] qemu - @sharder996  
  * [x] lxd - @townsend2010 
- [x] macOS
  * [x] M1
    + [x] qemu - @luis4a0 
  * [x] Intel
    + [x] qemu - @ricab 
- [x] Windows
  * [x] hyperv - @townsend2010 
  
Packages (updated on Jan. 11):
  - https://multipass-ci.s3.amazonaws.com/pr2835/multipass-1.12.0-dev.534.pr2835%2Bg1ad1e985.win-win64.exe
  - https://multipass-ci.s3.amazonaws.com/pr2835/multipass-1.12.0-dev.534.pr2835%2Bg1ad1e985.mac-Darwin.pkg

Regressions to address in this PR:
 - [x] #2807
 - [x] #2797 
 - [x] #2849 
 - [x] #2851 
 - [x] #2857


ricab (changes requested):
Hey @andrei-toterman, preliminary review. Thanks for the code, this is a lot of work and it looks nice and clean. Some comments and requests inline.

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,161 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
+        throw std::runtime_error(
+            fmt::format("Failed to terminate SSHFS mount process: {}", process->read_all_standard_error()));
 }
-
-void mp::SSHFSMountHandler::stop_all_mounts_for_instance(const std::string& instance)
+SSHFSMountHandler::~SSHFSMountHandler()
 {
-    auto mounts_it = mount_processes.find(instance);
-    if (mounts_it == mount_processes.end() || mounts_it->second.empty())
+    try
```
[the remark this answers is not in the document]
andrei-toterman: It kind of is common, one difference being the `category` used in logging. But still, pulling it to the base class wouldn't work, since calling a virtual method inside a destructor bypasses virtual dispatch, so the base destructor wouldn't call the `stop` function of the derived class.

include/multipass/mount_handler.h
```diff
@@ -18,58 +18,66 @@
...
 
-    // Used to set up anything host side related
-    virtual void init_mount(VirtualMachine* vm, const std::string& target_path, const VMMount& vm_mount) = 0;
-    virtual void start_mount(VirtualMachine* vm, ServerVariant server, const std::string& target_path,
-                             const std::chrono::milliseconds& timeout = std::chrono::minutes(5)) = 0;
-    virtual void stop_mount(const std::string& instance, const std::string& path) = 0;
-    virtual void stop_all_mounts_for_instance(const std::string& instance) = 0;
-    virtual bool has_instance_already_mounted(const std::string& instance, const std::string& path) const = 0;
+    virtual void start(ServerVariant server, std::chrono::milliseconds timeout = std::chrono::minutes(5)) = 0;
+    virtual void stop() = 0;
```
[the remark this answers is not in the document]
andrei-toterman: The reason behind `start` is mainly for the QEMU mounts. They are initialized in the constructor, i.e. adding the mount arguments to the QEMU command line, when the instance is stopped. And then, only after the instance has started, the `start` method is called, which does the actual `mount` command. The SSHFS and SMB mounts can happen even when the instance is already running, so for them `start` is not really needed and they could do everything in the constructor. I have an idea of how to allow QEMU mounts to happen as the instance is running, which would allow every mount to only need the constructor, but I'm not entirely sure if it's a good idea, so I'll want to discuss it with all of you, and maybe introduce it in a future re-re-refactoring.

As for `stop`, I don't think we can get rid of it. The reason behind it is that unmounting can fail, and when it does, we want to inform the user of the failure and to not remove the mount. That is why `stop` can throw if anything fails. If we tried to handle unmounting only in the destructor, we wouldn't be able to keep the mount in case of unmount failure. So the closest way to get to RAII that I can think of is to make the `stop` method idempotent, and have the unmount happen either when manually calling `stop` and allowing it to throw, or when the object is dropped and any exception is caught and logged, in cases in which we don't care about preserving the mount on failure.

luis4a0:
Hey @andrei-toterman, thanks for this! The code looks great and Ricardo did a wonderful job inspecting it. I'd like to ask you about testing. I remember we discussed about it, and we agreed to defer adding unit tests. But we need to be sure that this works, so I'd like to discuss with you (i) what is the issue this is fixing? (ii) how to reproduce it? (iii) how to test this works? Thanks!

include/multipass/mount_handler.h
```diff
@@ -18,58 +18,68 @@
...
 
 protected:
-    MountHandler(const SSHKeyProvider& ssh_key_provider) : ssh_key_provider(&ssh_key_provider){};
+    MountHandler() = default;
```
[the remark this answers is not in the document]
andrei-toterman: Not entirely sure, so correct me if I'm wrong, but I think I need it for the mock/stub classes. If I don't provide a default constructor, those derived classes will need to call the base constructor which does some initialization and validation, which I don't think are needed for those derived classes.

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,161 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
     config.instance = vm->vm_name;
-    config.target_path = target_path;
-    config.source_path = vm_mount.source_path;
-    config.uid_mappings = vm_mount.uid_mappings;
-    config.gid_mappings = vm_mount.gid_mappings;
+    config.target_path = target;
+    config.source_path = mount.source_path;
+    config.uid_mappings = mount.uid_mappings;
+    config.gid_mappings = mount.gid_mappings;
     config.private_key = ssh_key_provider->private_key_as_base64();
```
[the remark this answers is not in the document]
andrei-toterman: I think that would be a little ugly without the designated initializers of C++20. The first field of `SSHFSServerConfig` is `host`, but that one gets assigned only in `start`, so in the initializer list I would probably have to have something like
```cpp
: MountHandler{vm, ssh_key_provider, target, mount},
  config{"",
         vm->ssh_port(),
         vm->ssh_username(),
         vm->vm_name,
         ssh_key_provider->private_key_as_base64(),
         mount.source_path,
         target,
         mount.uid_mappings,
         mount.gid_mappings}
```
Which I think is kind of error prone, in case something changes in `SSHFSServerConfig`, like the order of the fields.

andrei-toterman:
Hey @luis4a0! I think the issue with suspending was a different one, addressed in #2791. The issue that lead to this refactor was that each mount handler kept it's own internal state about what paths are mounted, but they were not aware of what other mount handlers have mounted, so doing something like this on the QEMU driver would result in a lot of weird behavior:
```plain
mp stop instance
mp mount -t native /source/path instance:target
mp start instance
mp mount -t classic /other/source/path instance:target
```
The classic SSHFS mount handler would not be aware that the native QEMU mount handler has already mounted on that path, so the mount would be overwritten. But the `mp info instance` would still show only the old mount as active. And `mp umount` would do a lot of weird stuff as well.

This PR tries to fix that by taking away the responsibility of the mount handlers to keep track of the mounts. Each mount handler will keep track of exactly one mount, and the daemon will keep track of all mounts through a `map[instance name, map[target path, mount handler base class]]`.

The best way to test this is, first of all, to make sure that mounts do indeed work correctly, and then to see if you can reproduce any weird behavior by trying to override mounts.

include/multipass/mount_handler.h
```diff
@@ -18,58 +18,68 @@
...
+            throw std::runtime_error(fmt::format("Mount source path \"{}\" does not exist.", mount.source_path));
+        if (err)
+            throw std::runtime_error(
+                fmt::format("Mount source path \"{}\" is not accessible: {}.", mount.source_path, err.message()));
+        if (source_status.type() != fs::file_type::directory)
+            throw std::runtime_error(fmt::format("Mount source path \"{}\" is not a directory.", mount.source_path));
+        if (source_status.permissions() != fs::perms::unknown &&
+            (source_status.permissions() & fs::perms::owner_read) == fs::perms::none)
+            throw std::runtime_error(fmt::format("Mount source path \"{}\" is not readable.", mount.source_path));
+    };
```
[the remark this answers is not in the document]
andrei-toterman: The only reason is that I am not really sure what would be an appropriate place to create a `cpp` file for it. Any suggestions? :sweat_smile:

luis4a0:
Hey @andrei-toterman, thanks! Now I have enough information for testing on my side.

src/sshfs_mount/sshfs_mount_handler.cpp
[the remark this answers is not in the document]
ricab: Very good point. What you could do is have a regular protected function in the base that you'd call explicitly in each derived class's destructor. But perhaps not worth the trouble on this occasion.

include/multipass/mount_handler.h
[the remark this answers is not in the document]
ricab: Right, and you can't easily extract that validation code into something mockable because, again, no virtual dispatching in the constructor. I suppose that validation code would have to be extracted into a mockable "validator", that the base class c'tor would receive (the real validator could be statically available as a member of the base class, for implementations to use). 

For now, the default constructor is not used, so I think there is no point providing it. How to approach mocking can be part of the work for tests, so maybe we can defer the decision and remove the constructor until it's needed. If it comes down to tests being a little less "unitary" (because they execute the validation code) as the cost to ensuring no empty handlers at compile time, perhaps that's worth paying(?) Maybe not, being protected, the risk of misuse is not big... We should probably look for prior art in other cases in our code. So yeah, can we defer the decision on mocking?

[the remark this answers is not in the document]
andrei-toterman: Actually, now that I think better, we won't really need in the future either. The only validation in the constructor comes from a single `MP_FILEOPS.status` call, which is already mockable, so it will be very easy to test. I'll remove the default constructor and I don't think it will need to come back again later.

src/sshfs_mount/sshfs_mount_handler.cpp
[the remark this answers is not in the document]
ricab: Yeah, I understand your point, and yes, naming parameters will be really nice. But the risk of wrong parameter ordering affects any function with multiple parameters of the same type. Ideally, one would create different types for the different strings, to make it safer. But I think it is not worth the trouble. Granted, with regular functions you don't have an alternative like the one you use here, but I am not quite convinced it is better. It is also error prone, in that it is easy to leave stuff uninitialized if new fields are added. And being forced to pass that empty string makes it clearer that it is not there yet. You might even pass "not yet defined" or something.

But OK, all in all I recognize this is splitting hairs and there is a margin of subjectivity. It's your code and you have a good justification, so leave it as you prefer. And feel free to argue and convince me anyway if you want :)

include/multipass/mount_handler.h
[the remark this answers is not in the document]
ricab: OK, I am happy with that, thanks. Very much interested in you idea for live mounting in QEMU too.

include/multipass/mount_handler.h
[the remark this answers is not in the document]
ricab: Hmm, good question. Perhaps Daemon? It looks like there is already a circular dependency between daemon and platform, so the VMs should still be able to use it.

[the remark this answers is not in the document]
ricab: If you find it is too much trouble, leave it.

src/daemon/daemon.cpp
```diff
@@ -1928,11 +1887,7 @@ try // clang-format on
             if (instance->current_state() == VirtualMachine::State::delayed_shutdown)
                 delayed_shutdown_instances.erase(name);
 
-            for (const auto& mount_handler : config->mount_handlers)
-            {
-                mount_handler.second->stop_all_mounts_for_instance(name);
-            }
-
+            mounts[name].clear();
```
ricab, line 1890: No stopping needed, to avoid corrupting data on the host? In any implementation?

ricab (commented):
Still only partial review...

src/platform/backends/qemu/qemu_mount_handler.cpp
```diff
@@ -38,54 +29,61 @@ namespace
...
+    : MountHandler{vm, ssh_key_provider, target, mount}
 {
-    using St = VirtualMachine::State;
-    const auto skip_states = {St::off, St::stopped, St::suspended};
-
-    auto state = vm->current_state();
-    if (std::none_of(cbegin(skip_states), cend(skip_states), [&state](const auto& st) { return state == st; }))
-        throw std::runtime_error("Please shutdown virtual machine before defining native mount.");
+    if (!(qemu_vm = dynamic_cast<QemuVirtualMachine*>(vm)))
+        throw std::runtime_error{"QEMU mount must be used on a QEMU VM"};
```
ricab, line 39: Why not receive a `QemuVirtualMachine` to enforce that at compilation time? The VMs make these, so you shouldn't need the RTTI. If some other caller wants to construct directly, they'd get the onus of ensuring the right type.

src/daemon/daemon.cpp
andrei-toterman: Hmm, even when deleting the instance? I know we discussed this during a daily sync, and from what I understood, we should prevent only stopping and suspending from happening in case of unmount failure. Should we also prevent non-purge deleting?

src/platform/backends/qemu/qemu_mount_handler.cpp
andrei-toterman: Ah, yes, very good point. Thanks!

src/daemon/daemon.cpp
ricab: Right, we did discuss that. OK, let's stick to what we said then. I guess corruptions could only arise if the mount was still being written to, which the user would have to ensure was not the case before deleting. I just don't know when writes are flushed...

include/multipass/virtual_machine.h
```diff
@@ -76,8 +77,8 @@ class VirtualMachine : private DisabledCopyMove
     virtual void update_cpus(int num_cores) = 0;
     virtual void resize_memory(const MemorySize& new_size) = 0;
     virtual void resize_disk(const MemorySize& new_size) = 0;
-    virtual void add_vm_mount(const std::string& target_path, const VMMount& vm_mount) = 0;
-    virtual void delete_vm_mount(const std::string& target_path) = 0;
+    virtual std::unique_ptr<MountHandler> make_native_mount_handler(const SSHKeyProvider* ssh_key_provider,
+                                                                    std::string target, const VMMount& mount) = 0;
```
luis4a0, line 81: Is there a special motivation to pass a string by copy? I remember you answered something about `QString` at some point, just checking you know what you're doing.

src/daemon/daemon.cpp
```diff
@@ -1982,50 +1937,52 @@ try // clang-format on
...
         {
-            for (const auto& mount_handler : config->mount_handlers)
+            for (auto it = vm_mounts.begin(); it != vm_mounts.end();)
             {
-                mount_handler.second->stop_all_mounts_for_instance(name);
+                // iterator must be advanced before used in order to prevent iterator invalidation caused by deleting
+                // from the iterated map
+                const auto& [target, mount] = *it++;
+                do_unmount(*mount, target);
             }
```
luis4a0, line 1975: I know this works, but it scares me a bit the fact that a function call can invalidate an iterator. What about unmounting everything and then emptying `vm_mounts`?

andrei-toterman: I don't think that will work. Stopping and destructing a mount should happen in tandem. If we stop the mounts, but do not delete them from the map, if one of the mounts fails and throws from `stop`, the ones that already stopped will still be in the map when they souldn't.

include/multipass/virtual_machine.h
andrei-toterman: Nope, no special reason and no relation to QString (which could be passed by value due to its copy-on-write implementation). My mistake. I will make sure to pass stuff by const reference accordingly :smile:

src/daemon/daemon.cpp
andrei-toterman: Why would it return an iterator to the first element? One alternative I see is for it to return an iterator to the next element after the deleted one, like `erase` does on a map. The best implementation for that approach I could come up with looks like this
```cpp
auto do_unmount = [&](auto mount_iter) {
    auto& [target, mount] = *mount_iter;
    try
    {
        mount->stop();
        vm_spec_mounts.erase(target);
        return vm_mounts.erase(mount_iter);
    }
    catch (const std::runtime_error& e)
    {
        fmt::format_to(std::back_inserter(errors), "failed to unmount \"{}\" from '{}': {}\n", target, name,
                       e.what());
        return ++mount_iter;
    }
}
// Empty target path indicates removing all mounts for the VM instance
if (target_path.empty())
{
    for (auto it = vm_mounts.begin(); it != vm_mounts.end(); it = do_unmount(it))
        ;
}
else
{
    if (auto it = vm_mounts.find(target_path); it != vm_mounts.end())
    {
        do_unmount(it);
    }
    else
    {
        fmt::format_to(std::back_inserter(errors), "path \"{}\" is not mounted in '{}'\n", target_path, name);
    }
}
```

Opinions?

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
-                {
-                    auto reply = make_reply_from_server(*server);
-                    reply.set_reply_message("Enabling support for mounting");
-                    server->Write(reply);
-                }
-            };
+        [this, timeout](auto&& server) {
+            SSHSession session{vm->ssh_hostname(), vm->ssh_port(), vm->ssh_username(), *ssh_key_provider};
+            if (has_sshfs(vm->vm_name, session))
+                return;
```
ricab, line 135: Why pull this into the lambda? This is independent of the server, so why mix it with the variant visit? You get one more capture, but you can trade it in below :)

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
+
+            if (server)
+            {
+                auto reply = make_reply_from_server(server);
+                reply.set_reply_message("Enabling support for mounting");
+                server->Write(reply);
+            }
 
-            install_sshfs_for(vm->vm_name, session, on_install, timeout);
+            install_sshfs_for(vm->vm_name, session, timeout);
```
ricab, line 144: And this might as well be pulled out of the visit, to isolate the code that depends on the variant. No need to capture the timeout then.

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
-        });
+    QObject::connect(process.get(), &Process::finished, [this](const ProcessState& exit_state) {
+        if (exit_state.completed_successfully())
+        {
+            mpl::log(mpl::Level::info, category,
+                     fmt::format("Mount \"{}\" in instance '{}' has stopped", target, vm->vm_name));
+        }
+        else
+        {
+            // not error as it failing can indicate we need to install sshfs in the VM
```
ricab, line 160: Is this still true? We install above if we don't find it, right?

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
+    start_and_block_until_connected(process.get());
 
     // Check in case sshfs_server stopped, usually due to an error
-    auto process_state = sshfs_server_process->process_state();
+    auto process_state = process->process_state();
     if (process_state.exit_code == 9) // Magic number returned by sshfs_server
-    {
-        throw mp::SSHFSMissingError();
-    }
+        throw SSHFSMissingError();
```
ricab, line 180: OK, I am confused... Didn't we just install it?

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
     {
-        for (auto it = mounts_it->second.cbegin(); it != mounts_it->second.cend();)
-        {
-            // Clever postfix increment with member access needed to prevent iterator invalidation since iterable is
-            // modified in sshfs connection signal handler
-            stop_mount(instance, it++->first);
-        }
+        mpl::log(
+            mpl::Level::warning, category,
+            fmt::format("Failed to gracefully stop mount \"{}\" in instance '{}': {}", target, vm->vm_name, e.what()));
```
ricab, line 206: You need to kill the process at this point, right? Otherwise, slots might be called after it was deleted.

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
+        if (exit_state.completed_successfully())
+        {
+            mpl::log(mpl::Level::info, category,
+                     fmt::format("Mount \"{}\" in instance '{}' has stopped", target, vm->vm_name));
+        }
+        else
+        {
+            // not error as it failing can indicate we need to install sshfs in the VM
+            mpl::log(mpl::Level::warning, category,
+                     fmt::format("Mount \"{}\" in instance '{}' has stopped unexpectedly: {}", target, vm->vm_name,
```
ricab, line 162: I find this message misleading. The process completing "unsuccessfully" and "unexpectedly" are two different things. It may complete unsuccessfully after it is killed, or even terminated (e.g. if it encountered some error while quitting); and, as far as we know, it could stop unexpectedly without signalling any error.

tests/stub_virtual_machine.h
```diff
@@ -106,12 +107,11 @@ struct StubVirtualMachine final : public multipass::VirtualMachine
...
-    void add_vm_mount(const std::string&, const VMMount&) override
-    {
-    }
-
-    void delete_vm_mount(const std::string&) override
+    std::unique_ptr<MountHandler> make_native_mount_handler(const SSHKeyProvider* ssh_key_provider,
+                                                            const std::string& target, const VMMount& mount) override
     {
+        // TODO replace with StubMountHandler when writing tests
+        return nullptr;
```
ricab, line 114: Until then, throwing might be a better option (easier to catch if this is called inadvertently).

ricab (commented):
Alright, here is another round of comments. Nothing big, it's mostly good to go. Good job!

src/sshfs_mount/sshfs_mount_handler.cpp
andrei-toterman: The destructor of `QProcess` kills the process, so it should be killed immediately after this point.

src/platform/backends/qemu/qemu_mount_handler.cpp
```diff
@@ -38,54 +29,58 @@ namespace
...
-}
-
-void mp::QemuMountHandler::init_mount(VirtualMachine* vm, const std::string& target_path, const VMMount& vm_mount)
+QemuMountHandler::QemuMountHandler(QemuVirtualMachine* vm, const SSHKeyProvider* ssh_key_provider, std::string target,
+                                   const VMMount& mount)
+    : MountHandler{vm, ssh_key_provider, target, mount},
+      qemu_vm{vm},
+      // Create a reproducible unique mount tag for each mount. The cmd arg can only be 31 bytes long so part of the
+      // uuid must be truncated. First character of tag must also be alpabetical.
+      tag{QUuid::createUuidV3(QUuid(), QString::fromStdString(target))
```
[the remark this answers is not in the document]
andrei-toterman: Not entirely sure, as I didn't make this decision.
@sharder996, I think you wrote this initially, right? Any reason to use a UUID v3 generated from the target? I don't really see a reason to tie the UUID to the target. Can't they be totally unrelated?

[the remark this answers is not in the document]
sharder996: It's generated from the target because each mount needs to have a unique tag. Since the mount target also needs to be unique, I used it as a seed to create the mount tag.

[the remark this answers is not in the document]
andrei-toterman: I understand that the tag must be unique, but a non-seeded UUID should pretty much guarantee that. Nevertheless, I don't think it would be a bad idea to allow `mp::utils::make_uuid` to be seeded. So I'll add that :)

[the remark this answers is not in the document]
sharder996: Oh, the other reason I needed to seed the tag is because we need it in two places; once to pass into qemu when creating the process and the second when actually mounting.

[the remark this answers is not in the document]
andrei-toterman: Ah, yes, indeed. Well, now the tag gets generated only once in the constructor, so I guess we can just skip the whole seed thing entirely and use `mp::utils::make_uuid` as is.

src/daemon/daemon.cpp
luis4a0: @ricab my concern was that the function `do_unmount()` invalidated the iterator but the caller had to take care of the invalid iterator. I proposed to address the invalidation in the same function, or something like that. But if you don't consider it's important, nevermind.

src/platform/backends/qemu/qemu_mount_handler.cpp
```diff
@@ -38,54 +29,58 @@ namespace
...
-
-    vm->add_vm_mount(target_path, vm_mount);
-    mounts[vm->vm_name][target_path] = vm;
+             fmt::format("initializing native mount {} => {} in '{}'", mount.source_path, target, vm->vm_name));
+
+    const auto uid_map = mount.uid_mappings.empty() ? std::make_pair(1000, 1000) : mount.uid_mappings[0];
+    const auto gid_map = mount.gid_mappings.empty() ? std::make_pair(1000, 1000) : mount.gid_mappings[0];
+    const auto uid_arg = QString("uid_map=%1:%2,").arg(uid_map.first).arg(uid_map.second == -1 ? 1000 : uid_map.second);
+    const auto gid_arg = QString{"gid_map=%1:%2,"}.arg(gid_map.first).arg(gid_map.second == -1 ? 1000 : gid_map.second);
+    qemu_vm->mount_args[target] = {
```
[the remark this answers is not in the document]
andrei-toterman: Yeah, I agree that's it's not a nice approach. But we need to set, erase and read from the map, so we need full access to it basically. Would it be nicer to have three separate methods for that, or only one that returns a reference to the map and use that to modify it?

And while we're on this topic, I want to address a related issue and ask for an opinion. The reason we need to read from the map is to help make `stop` idempotent. If we call `stop` once and it succeeds, it will erase the respective entry from the map, and any further call to `stop` will read the map again and not do anything, since there will be no entry in the map. And we do something similar for the other implementations as well. We check some state to determine whether a successful `stop` has happened, and do nothing if it did. Would it be an inelegant approach to just have boolean flags that we set on success and use them to determine whether `start`/`stop` have completed successfully?

src/sshfs_mount/sshfs_mount_handler.cpp
ricab: Ahh right! Good. I wonder why there was that comment about the delete_later in the code...

andrei-toterman: From my understanding, it was based on a false assumption. It stated that the slots connected to the process signals could be called on the event loop, so in order to ensure that by the time they are called, the process is not destroyed, it used a `qt_delete_later_unique_ptr`, so that the process is also destroyed on the event loop, after all the slots have been called. But this should never be the case. The concern about the slots being called on the event loop is valid only for `Qt::QueuedConnection`. But when connecting a lambda as a slot to a signal, the connection is always `Qt::DirectConnection`, which means the slots will be called immediately, and the `emit` call will only return after all the slots have been called. That's why now the mount handler uses a simple `unique_ptr` to hold the process. By the time it is destroyed, it can't have any slots waiting to be called.

src/daemon/daemon.cpp
ricab: Ahh ok, I see what you mean @luis4a0. `do_unmount` will need to change the iterator one way or another, so I guess the most important is to make that clear in the code. The lambda is already pretty localized and there is a clear comment, but I think receiving the iterator in `do_unmount` (as in Andrei's alternative code) makes more sense. 

I am not sure that incrementing it in the unmount lambda is the best. We don't always need the iterator any longer and unmount should be concerned with unmounting, not iterating. Perhaps renaming the param to `expiring_it` would make it clear enough that the client could not use the iterator any longer. Maybe even receiving by rvalue reference, even though copies would still be invalidated...

luis4a0: Yeah, renaming to `expiring_it` is a good approach which self-documents the situation.

andrei-toterman: One thing that I think should be taken into consideration is that this happens in other parts of the code (e.g. stopping/suspending an instance), but without the use of a lambda. The reason I used a lambda here is simple: the same code applies to both unmounting all mounts or one mount. But in other places, only all mounts are removed, so it doesn't make sense to have a lambda in those places, so the iterator is advanced inside the loop, with the same `auto& [target, mount] = *it++` trick. I think keeping it the same in all places makes it easier to understand.

luis4a0: `*it++` is fine, my original worry is that it is used to take into account the invalidation done in another function. I agree it's better to handle the situation in the same way in all places, so maybe the renaming proposed by Ricardo is the best option.

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
+        mpl::log(mpl::Level::warning, category, fmt::format("Snap support is not installed in '{}'", name));
         throw std::runtime_error(
-            fmt::format("Snap support needs to be installed in \'{}\' in order to support mounts.\n"
+            fmt::format("Snap support needs to be installed in '{}' in order to support mounts.\n"
                         "Please see https://docs.snapcraft.io/installing-snapd for information on\n"
                         "how to install snap support for your instance's distribution.\n\n"
                         "If your distribution's instructions specify enabling classic snap support,\n"
                         "please do that as well.\n\n"
-                        "Alternatively, install `sshfs` manually inside the instance.",
+                        "Alternatively, install `multipass-sshfs` manually inside the instance.",
```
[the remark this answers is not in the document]
andrei-toterman: True, I don't think it makes sense here anymore. I'm not entirely familiar with how the SSHFS mounts implementation changed over time, but it seems to require `multipass-sshfs` to be installed in order for the mount to start for a long time now, so I guess that sentence can go away.

src/sshfs_mount/sshfs_mount_handler.cpp
andrei-toterman: Indeed, the process wouldn't start if it was not installed, so this is not true anymore. I think the log can be promoted to `error` then.

src/platform/backends/qemu/qemu_mount_handler.cpp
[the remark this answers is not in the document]
ricab: > [...] we need to set, erase and read from the map, so we need full access to it basically [...]

In such cases, I think it is preferable to have an (appropriately named) accessor method that returns a non-const reference (e.g. `modifiable_mount_args`. That pulls it to the interface explicitly and makes it clear that it can be changed elsewhere. An example of a practical situation this would be helpful for is when grepping the code of QEMU VM class to find where `mount_args` is changed. With the current implementation, it would be easy to miss the friend declaration that enables the change in the mount handler. On the contrary, an accessor would be hit by the grep, so people would know they'd need to search for it too.

> [...] boolean flags [...]

Hmm, good question. On the one hand, that would be a sort of duplication of truth, so there is always the risk of going out of sync. On the other, it would avoid a search. So it could be considered a justified form of caching. Also, I may be missing something, but I think you may need something of that sort in samba mounts because, IIUC, the `share_name` is only set in the constructor and clearing it would prevent the mount from starting again after it was stopped. 

Finally, I guess it makes sense conceptually for the base class that offers stop/start to keep track of whether or not it is active, regardless of actual implementations. So one approach could be to implement start/stop with the *template method* pattern in the base class. That is, the base `MountHandler` would have something like:

```c++
public:
(not virtual) void start(...)
{
    if (!m_active)
    {
        start_impl(...);
        m_active = true;
    }
}

(not virtual) void stop()
{
    if (active)
      stop_impl();
      m_active = true;
}

bool is_active() { return m_active; }

protected:
virtual void start_impl(...) = 0;
virtual void stop_impl() = 0;

private:
bool m_active;
```

Descendants would only implement *_impl versions and not need to care about idempotence any longer &ndash; they could still assert that they were being called from the right state, or even throw a `logic_error` if we fear that situation would arise in real usage. Of course, all of this would also need locks to ensure proper thread-safety...

src/sshfs_mount/sshfs_mount_handler.cpp
```diff
@@ -60,229 +48,162 @@ void start_and_block_until(mp::Process* process, Signal signal, std::function<bo
...
-        [this, vm, &session, &timeout](auto&& server) {
-            auto on_install = [this, server] {
-                if (server)
-                {
-                    auto reply = make_reply_from_server(*server);
-                    reply.set_reply_message("Enabling support for mounting");
-                    server->Write(reply);
-                }
-            };
+        [this, timeout](auto&& server) {
```
ricab, line 132: Just noticed: why use a forwarding reference if it is not forwarded?

src/sshfs_mount/sshfs_mount_handler.cpp
ricab: OK, if I am seeing it right, we should not get into this situation anymore and we could replace this with an `assert(process_state.exit_code != 9)`. However, that should have been the case before this PR and I guess it's out of scope, so feel free to ignore.

src/sshfs_mount/sshfs_mount_handler.cpp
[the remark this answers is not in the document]
ricab: OK, after discussing with @townsend2010, it is actually the other way around: the message is correct and the code is wrong. That is, we would like to honor what the message suggests, so we should keep it. We can consider the fact that we do not honor it right now as a latent bug. So maybe just drop any change to this string and we can make everything consistent at a later time.

src/sshfs_mount/sshfs_mount_handler.cpp
ricab: Same here, ignore the mismatch between code and comment/log-level and revert to what was there before.

src/sshfs_mount/sshfs_mount_handler.cpp
ricab: Again, yeah just ignore.

src/daemon/daemon.cpp
andrei-toterman: Okay, after all of this, I'm not entirely sure I understand what the final proposed solution is, so can you please spell it out for me? :sweat_smile:

src/platform/backends/qemu/qemu_mount_handler.cpp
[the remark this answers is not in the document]
andrei-toterman: Nice, I like this. Will implement!

src/daemon/daemon.cpp
andrei-toterman: Alright, in today's sync, @luis4a0 clarified to me that the final solution is to leave the lambda as is and to better document the iterator invalidation through a better variable name and a comment.

ricab: That and receiving the iterator directly in the lambda. That's the param I would call "expiring_it" (although naming the argument like that is fine too). As a bonus, that would avoid a subscript operation (computing hash and corresponding map node).

src/platform/backends/qemu/qemu_mount_handler.h
```diff
@@ -18,26 +18,25 @@
...
-                     const std::chrono::milliseconds& timeout = std::chrono::minutes(5)) override;
-    void stop_mount(const std::string& instance, const std::string& path) override;
-    void stop_all_mounts_for_instance(const std::string& instance) override;
-    bool has_instance_already_mounted(const std::string& instance, const std::string& path) const override;
+    void start_impl(ServerVariant server, std::chrono::milliseconds timeout = std::chrono::minutes(5)) override;
+    void stop_impl() override;
 
 private:
-    std::unordered_map<std::string, std::unordered_map<std::string, VirtualMachine*>> mounts;
+    QemuVirtualMachine::MountArgs& vm_mount_args;
```
ricab, line 38: This is OK in here because the class is noncopyable, but [careful with reference attributes](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references) in the generic case.

include/multipass/mount_handler.h
```diff
@@ -45,31 +43,61 @@ class MountHandler : private DisabledCopyMove
...
-                             const std::chrono::milliseconds& timeout = std::chrono::minutes(5)) = 0;
-    virtual void stop_mount(const std::string& instance, const std::string& path) = 0;
-    virtual void stop_all_mounts_for_instance(const std::string& instance) = 0;
-    virtual bool has_instance_already_mounted(const std::string& instance, const std::string& path) const = 0;
+    void start(ServerVariant server, std::chrono::milliseconds timeout = std::chrono::minutes(5))
+    {
+        if (active)
+            return;
+        start_impl(server, timeout);
+        active = true;
```
ricab, line 51: Hmm, looking at the code, I don't think the `atomic_bool` is enough in here after all. I thought of `if (active.exchange(true))` instead, which would prevent concurrent starts from calling `start_impl` at the same time. However, without further care, that would also allow the state to be set to active when `start_impl` failed. And yet, it would not help us with mixed starts/stops, when executed concurrently. A lock is needed for that, AFAICS, at which point the atomic becomes redundant. 

As it is, the reader may get the impression that concurrency has been dealt with, upon seeing the atomic. So, if you don't feel like doing yet another significant iteration on the code to add a lock, I would ask for just a "TODO" comment making it clear that this still needs to be made thread safe.

ricab (approved):
Alright, nicely done @andrei-toterman, thanks! 

I wanna get a move on with this and my last request for a comment will not affect compiled packages, so approving now in order to move on with testing ASAP.

@luis4a0, we'll wait for your approval too before actual merging.

bors try

ricab:
Ugh?

bors retry

ricab:
It looks like there are some [degradations on GH](https://www.githubstatus.com/), so we'll have to wait for tomorrow anyway. I just found an issue on the Windows side anyway, so the timing could be worse.

src/utils/utils.cpp
```diff
@@ -423,10 +423,7 @@ QString mp::utils::get_multipass_storage()
 
 QString mp::utils::make_uuid()
 {
-    auto uuid = QUuid::createUuid().toString();
-
-    // Remove curly brackets enclosing uuid
-    return uuid.mid(1, uuid.size() - 2);
+    return QUuid::createUuid().toString(QUuid::WithoutBraces);
```
ricab, line 426: :+1:

include/multipass/mount_handler.h
andrei-toterman: Ah, yes, I see what you mean. Would something like this solve it?
```cpp
void start(...) {
    std::lock_guard active_lock{active_mutex};
    if (active)
        return;
    start_impl(...);
    active = true;
}
...
private:
    bool active;
    std::mutex active_mutex;
```

ricab: It would indeed, that would be perfect!

andrei-toterman: Alright, done!

ricab:
OK, I think we can start testing this on the different platforms (picking up the artifacts from the latest CI run). Sign up in the (upcoming) checklist in this PR's description.

townsend2010:
Merge conflicts need fixing...

townsend2010:
This is looking really good functionally wise.  I've tested on all 3 platforms and it's holding up very nicely even when using a mix of mount types in the same instance!  Great work!  I don't think this will get merged before the holiday shutdown, but it should be in good shape after the new year.  A sanity check of the commits after the last code review is still needed and can be done later.

include/multipass/mount_handler.h
```diff
@@ -45,31 +42,62 @@ class MountHandler : private DisabledCopyMove
...
-                             const std::chrono::milliseconds& timeout = std::chrono::minutes(5)) = 0;
-    virtual void stop_mount(const std::string& instance, const std::string& path) = 0;
-    virtual void stop_all_mounts_for_instance(const std::string& instance) = 0;
-    virtual bool has_instance_already_mounted(const std::string& instance, const std::string& path) const = 0;
+    void start(ServerVariant server, std::chrono::milliseconds timeout = std::chrono::minutes(5))
+    {
+        std::lock_guard active_lock{active_mutex};
+        start_impl(server, timeout);
+        active = true;
+    }
```
ricab, line 50: This is the only thing that I am still a little unsure of. The motivation for this scheme was to free implementations from being idempotent (with thread-safety as a bonus). You later said that concrete implementations really ought to be idempotent. OK, but are we sure they are now? I thought you had meanwhile fixed the case that needed that - restart - no? Or are you protecting against other cases where `active` could potentially go out of sync? But if that is a real concern, does it make sense to try to track the state at all? And why the asymmetry with respect to stop? Why do we need `start_impl`s to be idempotent, but not `stop_impl`s?

ricab (commented):
ughh...

include/multipass/mount_handler.h
andrei-toterman: Ahh, yes, I forgot to remove this. Indeed, this was needed by restart, since it didn't stop the mounts before starting them again. But now it does, so `start` can go back to being idempotent.

src/platform/backends/qemu/qemu_mount_handler.cpp
andrei-toterman: Hi, @ricab! Unfortunately I have to address this point again, as I've run into some difficulties while writing tests. When testing the `QemuMountHandler`, I need to pass a mocked VM to the handler, for stuff like `vm->current_state()`. But if the constructor parameter is of type `QemuVirtualMachine*` instead of the base `VirtualMachine*`, I cannot pass a mock in a clean way. I tried to get around it with `reinterpret_cast`, but that brings other problems when the handler calls `QemuVirtualMachine` specific methods, like `modifiable_mount_args`, which I cannot mock. Of course, the previous `dynamic_cast<QemuVirtualMachine*>` was not a complete solution, since it won't be able to cast from `MockVirtualMachine*`, so it only moves the type problem to runtime. But then at least we could do the dynamic cast a little later, after some other checks that could be mocked. Any ideas or suggestions on how to approach this?

andrei-toterman: I see two options: create a `MockQemuVirtualMachine` that inherits from `QemuVirtualMachine`, so that I can cleanly mock anything I need. Or pull the `QemuVirtualMachine` specific methods (just `modifiable_mount_args` as of now) into the base `VirtualMachine` and use the existing `MockVirtualMachine`. In my opinion, the second approach is kind of ugly. Every type of virtual machine will then inherit an unreasonable method for their specific implementation, and will end up throwing an exception when that method is called. I think the first approach would be cleaner, except that I would need to re-mock all the `VirtualMachine` methods. Do you know any nice multiple inheritance trick so that I can do something like
```cpp
struct MockQemuVirtualMachine : MockVirtualMachine, QemuVirtualMachine {
    // mock only QemuVirtualMachine specific stuff
}
```
so that it can be passed as a `QemuVirtualMachine*` but still use the mocked stuff from `MockVirtualMachine`?

ricab: OK, discussed elsewhere, but a couple of lines of code to make the idea more concrete:

```cpp
template <typename VMClass = VM>
class MockVM : public VMClass
{
// Mock VM methods
};

class MockQemuVM : public MockVM<QemuVM>
{
// Mock QemuVM methods
};
```

andrei-toterman: It works pretty nicely! The only minor inconvenience is that there are still several places that need to be modified with a `MockVirtualMachine<>` even though there is a default template parameter, like in the return type of functions or when used as a type in another template, like, `NiceMock<MockVirtualMachine<>>`.

So in order to avoid all that, I'm thinking to change the class name to something different and alias the concrete type with the default value to `MockVirtualMachine`
```cpp
template<typename T = VirtualMachine, typename = std::enable_if_t<std::is_base_of_v<VirtualMachine, T>>>
struct MockVirtualMachineT : public T {
...
}

using MockVirtualMachine = MockVirtualMachineT<>;
```

andrei-toterman: Ah, and the other issue is that `QemuVirtualMachine` is declared as `final`, but I assume it's not a really big deal if it is removed.

townsend2010 (approved):
@andrei-toterman, really great work here!  I'm giving my final approval and wait on the other approvals before merging.

luis4a0 (approved):
Succesfully tested on qemu on Mac M1.

sharder996 (approved):
LGTM on Linux with qemu! No issues found, good job!

townsend2010:
Ok, running `bors` to get proper merge message.

bors r=ricab,luis4a0,sharder996,townsend2010

PR #2836, opened 2022-11-23
Ignore trailing slash on mount target path

### LINKED ISSUE #1405: mount/unmount foreign folder bug
This happens when mounting a folder in the client using a trailing slash in the destination path, issuing  `multipass mount path/to/mount diplomatic-lungfish:mountpoint/`.

When trying to unmount it, the path must have the final slash. This is, the command
`multipass umount diplomatic-lungfish:mountpoint/` will work, but
`multipass umount diplomatic-lungfish:mountpoint` won't, the command will print
`"mountpoint" is not mounted"`
`mountpoint" not found in database`.

This also happens if mounting without using a final backslash and trying to unmount using one.

This bug is related to https://github.com/canonical/multipass/issues/1378 .

sharder996:
Clean target path before using to normalize searching/mounting/etc

Fixes #1405

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1537,8 +1537,8 @@ try // clang-format on
             continue;
         }
 
-        auto target_path = path_entry.target_path();
-        if (mp::utils::invalid_target_path(QString::fromStdString(target_path)))
+        auto target_path = QDir::cleanPath(QString::fromStdString(path_entry.target_path()));
+        if (mp::utils::invalid_target_path(target_path))
         {
             fmt::format_to(std::back_inserter(errors), "Unable to mount to \"{}\"\n", target_path);
             continue;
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -1990,11 +1990,11 @@ try // clang-format on
             continue;
         }
 
-        auto target_path = path_entry.target_path();
+        auto target_path = QDir::cleanPath(QString::fromStdString(path_entry.target_path()));
```
luis4a0, line 1993: Hey @sharder996, why don't keep `target_path` as a `std::string`? This way, the rest of the code stays unchanged.

sharder996: Overall that would result in less conversion. Not sure which is more icky, `fromStdString().toStdString()` or multiple `toStdString()` lol

luis4a0: Yeah, it's not nice to have two string types and convert back and forth between them. But I like how it is now, because the diff is only two lines.

luis4a0 (approved):
Ok, this works for me. Let´s rock!

bors r+

townsend2010:
Ugh, random crashes in `multipass_tests`...try again....

bors retry

townsend2010:
bors retry

luis4a0:
bors retry

townsend2010:
Ok, this is your last chance bors...

bors retry

PR #2846, opened 2022-12-05
[actions] Specify Ubuntu 20.04 since 22.04 is currently broken for LXD action

townsend2010:


townsend2010:
I have a better solution at #2847.

PR #2854, opened 2022-12-13
Hyperkit migration user feedback

ricab:
Implements proper communication with the user for the Hyperkit migration.


ricab:
Replaced by #2860.

PR #2856, opened 2022-12-14
Fix spinner callbacks

### LINKED ISSUE #2888: Empty spinner message after log in `(re)start`
**Describe the bug**
`Start` and `restart` commands show a spinner with no message (beyond a space) after printing log messages:

https://asciinema.org/a/ZLXiZVVqHaWhPMLTNeaCy8TBu

**To Reproduce**
How, and what happened?
1. `multipass (re)start -vv`

ricab:
Fixes #2888 and #2889, extracts the repeated `streaming_callback`s in question, and adds tests.

### DIFF
--- a/src/platform/client/client_platform.cpp
+++ b/src/platform/client/client_platform.cpp
@@ -57,7 +57,7 @@ int mcp::getgid()
     return ::getgid();
 }
 
-std::pair<std::string, std::string> mcp::get_user_password(mp::Terminal*)
+std::pair<std::string, std::string> mcp::Platform::get_user_password(mp::Terminal*) const
 {
     return {};
 }
--- a/include/multipass/cli/client_common.h
+++ b/include/multipass/cli/client_common.h
@@ -58,7 +58,7 @@ template <typename Request, typename Reply>
 void handle_user_password(grpc::ClientReaderWriterInterface<Request, Reply>* client, Terminal* term)
 {
     Request request;
-    const auto [username, password] = multipass::cli::platform::get_user_password(term);
+    const auto [username, password] = MP_CLIENT_PLATFORM.get_user_password(term);
 
     request.mutable_user_credentials()->set_username(username);
     request.mutable_user_credentials()->set_password(password);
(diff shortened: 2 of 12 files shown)


sharder996:
Looks good, but the behavior that you're fixing is still present in other Multipass commands such as `stop`, `suspend`, etc. Would it be much effort to fix those right away as well?

ricab:
Oh definitely! I found this issue when testing Hyperkit deprecation messaging, which isn't applicable on those commands, so I missed them. Well spotted. Adding to my TODO list :slightly_smiling_face:

ricab:
Hey @sharder996, looking back into this, `stop` and `suspend` are actually suffering from a slightly different problem: the output gets messed up because the default `streaming_callback` steps on the spinner's toes by printing directly to `cerr`. 

I am working on a fix for that too but, if you have no objections, I'd submit as a separate PR, just to keep it logically separate and unblock this one. And I think I will create a proper bug report for both cases, just to have it all tidy and proper.

sharder996:
Sounds good to me. I don't have any other objections and this looks good, so lets get it merged!

bors r+

ricab:
Thanks for the review @sharder996! I just noticed one command that is missing in this one though: `mount`. There we have the same problem. And I need to rebase too.

bors cancel

ricab:
OK, this should be good now.

ricab:
Hey @sharder996, I didn't like to see all those copy-paste callbacks untested, so I extracted them and started testing them. But then it was easier to just do everything in a single PR, so I am expanding this one to encompass all of it. Hope that sounds reasonable.

src/platform/client/client_platform.cpp
```diff
@@ -57,7 +57,7 @@ int mcp::getgid()
     return ::getgid();
 }
 
-std::pair<std::string, std::string> mcp::get_user_password(mp::Terminal*)
```
sharder996, line 60: Changing this breaks a reference to it in `client_platform_windows.cpp:76`

sharder996 (changes requested):

src/platform/client/client_platform.cpp
ricab: Yes, the private PR linked above fixes that.

sharder996 (approved):
LGTM! Missed the link to the private side PR in the list of commits. Nice job with the tests and getting everything covered!

sharder996:
bors cancel

sharder996:
Ok, this looks good! Lets merge now

bors r+

PR #2874, opened 2023-01-10
Add platform permission func

sharder996:
Windows handles permissions differently than other OS's so we need to add the ability to have a platform specific veersion.

Private side PR [here](https://github.com/canonical/multipass-private/pull/494)

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -1336,9 +1336,9 @@ TEST_F(Daemon, writesAndReadsMountsInJson)
 
     // Create a temp folder containing three subfolders, and create mount points for all of them.
     auto temp_mount_dir = QDir(mpt::TempDir().path());
-    auto temp_mount_1 = make_dir(temp_mount_dir, QString("a"), QFileDevice::Permissions{}).toStdString();
-    auto temp_mount_2 = make_dir(temp_mount_dir, QString("b"), QFileDevice::Permissions{}).toStdString();
-    auto temp_mount_3 = make_dir(temp_mount_dir, QString("c"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_1 = MP_UTILS.make_dir(temp_mount_dir, QString("a"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_2 = MP_UTILS.make_dir(temp_mount_dir, QString("b"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_3 = MP_UTILS.make_dir(temp_mount_dir, QString("c"), QFileDevice::Permissions{}).toStdString();
 
     mp::id_mappings uid_mappings_1{{123, 321}};
     mp::id_mappings gid_mappings_1{{456, 654}};
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -109,11 +109,6 @@ struct Daemon : public mpt::DaemonTestFixture
             return mock_utils.Utils::filesystem_bytes_available(data_directory);
         });
 
-        ON_CALL(mock_utils, make_dir(_, _, _))
-            .WillByDefault([this](const QDir& a_dir, const QString& name, QFileDevice::Permissions permissions) {
-                return mock_utils.Utils::make_dir(a_dir, name, permissions);
-            });
-
         EXPECT_CALL(mock_platform, get_blueprints_url_override()).WillRepeatedly([] { return QString{}; });
         EXPECT_CALL(mock_platform, multipass_storage_location()).Times(AnyNumber()).WillRepeatedly(Return(QString()));
         EXPECT_CALL(mock_platform, create_alias_script(_, _)).WillRepeatedly(Return());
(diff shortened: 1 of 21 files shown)


ricab (commented):
Hey @sharder996, one last little piece missing.

tests/test_daemon.cpp
```diff
@@ -109,6 +109,11 @@ struct Daemon : public mpt::DaemonTestFixture
             return mock_utils.Utils::filesystem_bytes_available(data_directory);
         });
 
+        ON_CALL(mock_utils, make_dir(_, _, _))
+            .WillByDefault([this](const QDir& a_dir, const QString& name, QFileDevice::Permissions permissions) {
+                return mock_utils.Utils::make_dir(a_dir, name, permissions);
+            });
```
ricab, line 115: This is a big hammer, it affects all daemon tests. I understand you're just recovering what happened before, but now that we have the mock we might as well use it and prevent inadvertent filesystem changes affecting other tests. 

I had a look at the two tests that failed without this. Here's an alternative that fixes them individually: https://paste.ubuntu.com/p/XczdJ7MX9V/

tests/test_daemon.cpp
```diff
@@ -1336,9 +1341,9 @@ TEST_F(Daemon, writesAndReadsMountsInJson)
 
     // Create a temp folder containing three subfolders, and create mount points for all of them.
     auto temp_mount_dir = QDir(mpt::TempDir().path());
-    auto temp_mount_1 = make_dir(temp_mount_dir, QString("a"), QFileDevice::Permissions{}).toStdString();
-    auto temp_mount_2 = make_dir(temp_mount_dir, QString("b"), QFileDevice::Permissions{}).toStdString();
-    auto temp_mount_3 = make_dir(temp_mount_dir, QString("c"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_1 = MP_UTILS.make_dir(temp_mount_dir, QString("a"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_2 = MP_UTILS.make_dir(temp_mount_dir, QString("b"), QFileDevice::Permissions{}).toStdString();
+    auto temp_mount_3 = MP_UTILS.make_dir(temp_mount_dir, QString("c"), QFileDevice::Permissions{}).toStdString();
```
ricab, line 1346: The permissions here are empty, but there is no special treatment for empty permissions anymore. I am not sure how this worked. Perhaps the platform mock was not doing anything to set the permissions. In the alternative above, I just bypass all that with a dedicated lambda, but feel free to use your trick to call the base here, as long as we're sure the permissions are handled correctly.

ricab (commented):
Almost there!

tests/test_daemon.cpp
ricab: Maybe each temp dir could be just `QDir(mpt::TempDir().path());`?

sharder996: `QDir(mpt::TempDir().path())` doesn't work since it doesn't actually create a directory.

tests/test_daemon.cpp
sharder996: I see your point, but I'd rather just move the bypass to the specific test then rewrite a simplified version of the `make_dir()` function in a lambda.

tests/test_daemon.cpp
ricab: Doesn't `mpt::TempDir()` create a temporary directory?

tests/test_daemon.cpp
ricab: Sure, makes sense.

tests/test_daemon.cpp
sharder996: I found that it wasn't creating anything. And then the test would fail.

tests/test_daemon.cpp
```diff
@@ -283,6 +278,11 @@ TEST_F(Daemon, data_path_with_storage_valid)
 
     EXPECT_CALL(mock_platform, multipass_storage_location()).WillOnce(Return(mp::utils::get_multipass_storage()));
 
+    ON_CALL(mock_utils, make_dir(_, _, _))
+        .WillByDefault([this](const QDir& a_dir, const QString& name, QFileDevice::Permissions permissions) {
+            return mock_utils.Utils::make_dir(a_dir, name, permissions);
+        });
+
```
ricab, line 285: This one doesn't actually need to make a directory, so I think it would be preferable to just mock the required return, as usual. But I won't block on that.

tests/test_daemon.cpp
ricab: Oh right, because that `TempDir` is a temporary in that expression, so it would delete the directory after creating the `QDir`. You'd have to bind it to a variable. Anyway, it's fine this way too.

ricab (approved):
Alrighty then

bors r+

sharder996:
Build failed because of IS outage.

bors r+

townsend2010:
bors retry

townsend2010:
bors retry

townsend2010:
Ah, right, this has a private counterpart, so `bors` gonna fail...merging manually.

townsend2010:
Ok, regular CI is all green, so merging this now.

PR #2877, opened 2023-01-12
Handle Qemu native mount when resuming from suspended state

### LINKED ISSUE #2876: Multipass won't restart with native mounts on an instance
When restarting Multipass (after a reboot or with `launchctl load`/`unload`) having an instance with a native mount running, the daemon will suspend the instance when shutting down, but won't start it again because there are native mounts on it. The daemon must catch the exception generated by this and handle it accordingly.

**Logs**
The log file keeps adding the following lines at the end:
```
[2023-01-12T11:07:40.285] [info] [VMImageHost] Did not find any supported products in "appliance"
[2023-01-12T11:07:40.693] [info] [rpc] gRPC listening on unix:/var/run/multipass_socket
[2023-01-12T11:07:40.698] [debug] [qemu-img] [3306] started: qemu-img snapshot -l /var/root/Library/Application Support/multipassd/qemu/vault/instances/topical-moccasin/ubuntu-22.04-server-cloudimg-arm64.img
[error] [daemon] Caught an unhandled exception: Please shutdown the instance before attempting native mounts.
```

**Additional info**
- OS: macOS 13.1
- `multipass version`
multipass   1.11.0-rc.453+g51435a61.mac
`qemu`

andrei-toterman:
fix #2876

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -129,7 +129,7 @@ std::string qenum_to_string(RegisteredQtEnum val);
 
 // other helpers
 QString get_multipass_storage();
-QString make_uuid();
+QString make_uuid(const std::optional<std::string>& seed = std::nullopt);
 template <typename OnTimeoutCallable, typename TryAction, typename... Args>
 void try_action_for(OnTimeoutCallable&& on_timeout, std::chrono::milliseconds timeout, TryAction&& try_action,
                     Args&&... args);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2638,9 +2638,29 @@ grpc::Status mp::Daemon::cmd_vms(const std::vector<std::string>& tgts, std::func
 void mp::Daemon::init_mounts(const std::string& name)
 {
     auto& vm_mounts = mounts[name];
-    for (const auto& [target, vm_mount] : vm_instance_specs[name].mounts)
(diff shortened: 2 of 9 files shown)


townsend2010 (changes requested):
Hey @andrei-toterman,

Although this now keeps the daemon from crashing, the instance cannot be started since `qemu` exits on startup.

Steps to reproduce:
1. Create instance.
2. Stop instance.
3. Add `native` mount.
4. Stop the `multipass` snap.
5. Start the `multipass` snap.

Watch the logs and you'll see that `qemu` exits with "Permission denied" since the mount point is not being added to the Apparmor profile.  I think the `-virtfs` command line option needs to be sanitized out since we are really removing the native mount for the instance.

townsend2010:
I should also note that once things get in the aforementioned state, trying to do *anything* with the instance causes `multipassd` to segfault.

andrei-toterman:
Ok, after some more research and experimentation, I found out that resuming mounts from a suspended state is actually possible. It's only the adding of new devices that is not compatible with `loadvm`. The vm snapshot includes all the devices, and if the exact same devices are given when doing a `loadvm`, it will work. So after all, the solution is indeed to make the `QemuMountHandler` look for the mount arguments that were saved in the json file and use them if the mount tag matches.

townsend2010 (approved):
Hey @andrei-toterman!

Thanks for this!  Looks good!

bors merge

townsend2010:
bors retry

PR #2879, opened 2023-01-13
[timer] Handle spurious wake ups

### LINKED ISSUE #2878: Timer class does not deal with spurious wake ups
There are intermittent failing `Timer` related tests that fail on Windows.  These tests are usually related to timeouts not being seen and thus the `exit()` function not being called as expected.

After some code analysis of the `Timer` class, it appears that `std::condition_variable::wait_for()` is being spuriously woken up since `exit()` is not called.  The other very unlikely scenario in which this behavior may occur is if `wait_for()` does not return for some time after the timeout is reached.

It appears that spurious wake ups are more common on Windows, so we really should have a predicate for these to account for these spurious wake ups before the timeout occurs.

townsend2010:
Fixes #2878

### DIFF
--- a/include/multipass/timer.h
+++ b/include/multipass/timer.h
@@ -67,9 +67,9 @@ public:
     TimerSyncFuncs(const Singleton<TimerSyncFuncs>::PrivatePass&) noexcept;
 
     virtual void notify_all(std::condition_variable& cv) const;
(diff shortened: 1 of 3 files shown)


ricab (approved):
Yep, great catch and right solution.

bors r+

townsend2010:
Unrelated failures in the integration tests.  Merging manually.

PR #2893, opened 2023-01-23
[cli/transfer] don't treat windows absolute paths as instance:path pair

### LINKED ISSUE #2891: [regression] multipass transfer on windows - colon in the local/destination path now throws error
On Windows, local paths by definition have to include a colon after the drive letter. And when you use that, even via $HOME, or ~ in git bash, it now throws the error "Cannot specify an instance name for both source and destination". It only started doing this with the latest version 1.11.0.

I have a number of scripts that worked before and now throw the above error and so I have reverted to 1.10.1 to keep things working for now.

andrei-toterman:
fix #2891

### DIFF
--- a/src/client/cli/cmd/transfer.cpp
+++ b/src/client/cli/cmd/transfer.cpp
@@ -25,6 +25,8 @@
 
 #include <fmt/std.h>
 
+#include <QRegularExpression>
+
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
 namespace mcp = multipass::cli::platform;
@@ -163,9 +165,16 @@ std::vector<std::pair<std::string, fs::path>> cmd::Transfer::args_to_instance_an
     instance_entry_args.reserve(args.size());
     for (const auto& entry : args)
     {
+        // this is needed so that Windows absolute paths are not split at the colon following the drive letter
+        if (QRegularExpression{R"(^[A-Za-z]:\\.*)"}.match(entry).hasMatch())
+        {
+            instance_entry_args.emplace_back("", entry.toStdString());
+            continue;
+        }
+
         const auto source_components = entry.split(':');
         const auto instance_name = source_components.size() == 1 ? "" : source_components.first().toStdString();
-        const auto file_path = source_components.size() == 1 ? source_components.first() : source_components.at(1);
+        const auto file_path = source_components.size() == 1 ? source_components.first() : entry.section(':', 1);
         if (!instance_name.empty())
             request.add_instance_name()->append(instance_name);
         instance_entry_args.emplace_back(instance_name, file_path.isEmpty() ? "." : file_path.toStdString());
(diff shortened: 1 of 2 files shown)


src/client/cli/cmd/transfer.cpp
```diff
@@ -163,9 +165,16 @@ std::vector<std::pair<std::string, fs::path>> cmd::Transfer::args_to_instance_an
...
+        if (QRegularExpression{R"(^[A-Za-z]:[\\/].*)"}.match(entry).hasMatch())
+        {
+            instance_entry_args.emplace_back("", entry.toStdString());
+            continue;
+        }
+
         const auto source_components = entry.split(':');
         const auto instance_name = source_components.size() == 1 ? "" : source_components.first().toStdString();
-        const auto file_path = source_components.size() == 1 ? source_components.first() : source_components.at(1);
+        const auto file_path = source_components.size() == 1 ? source_components.first() : entry.section(':', 1);
```
ricab, line 177: Is this meant to handle linux paths with colons in them? But then how do we know which colon is the splitting one. If we don't know we can handle such cases correctly, I wonder if we're not better off just rejecting them from the get go? Or are you reverting to what we did earlier?

ricab (commented):
Thanks for the quick fix @andrei-toterman! Only one aspect I would like to clarify...

src/client/cli/cmd/transfer.cpp
ricab: Yep, makes sense :+1:

src/client/cli/cmd/transfer.cpp
```diff
@@ -163,9 +165,16 @@ std::vector<std::pair<std::string, fs::path>> cmd::Transfer::args_to_instance_an
     instance_entry_args.reserve(args.size());
     for (const auto& entry : args)
     {
+        // this is needed so that Windows absolute paths are not split at the colon following the drive letter
+        if (QRegularExpression{R"(^[A-Za-z]:[\\/].*)"}.match(entry).hasMatch())
```
ricab, line 169: Hey @andrei-toterman, after testing, this is a little problematic with single letter instance names:

```
$ multipass launch -n C
Launched: C
$ multipass transfer -rp C:\Users\ricab\bin\yasm.exe C:/home/ubuntu/yasm
An instance name is needed for either source or destination
```

A proper fix would probably require expanding the command line to deal with ambiguous cases somehow, but I think this could be *mostly* worked around by dropping the forward slash from the regex. People will need the backward slash to specify windows paths, but I think that's OK. It will still not work if the file name in the instance starts with a backward slash, but that's a corner case we can probably live with. WDYT?

andrei-toterman: Yeah, unfortunately instances with single letter names are not differentiated properly by the regex, and I assumed that would be a corner case as well. But your solution makes more sense. Thanks, I'll do it!

ricab (approved):
Alright, let's get it in.

bors r+

PR #2897, opened 2023-01-23
[tests] Fix MemorySize tests depending on locale

ricab:
Avoids unsupported number representations, e.g. with a comma separating integers and decimal parts.

### DIFF
--- a/tests/test_memory_size.cpp
+++ b/tests/test_memory_size.cpp
@@ -19,6 +19,7 @@
 #include "common.h"
 
 #include <multipass/exceptions/invalid_memory_size_exception.h>
+#include <multipass/format.h>
 #include <multipass/memory_size.h>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


sharder996 (commented):
Hey @ricab, this looks fine, but just curious; couldn't one have just replaced `std::to_string()` with `fmt::format()` everywhere and eliminated the need for the template? It's not providing any value outside the file and it's a just simple wrapper.

ricab:
Hi @sharder996. We could, but we'd have to specify the format each time, which would make it a little harder for the reader inline. Contrast the two:

```cpp
EXPECT_EQ(mp::MemorySize{fmt::format("{}", val) + "M"}.in_kilobytes(), val * kilo);
EXPECT_EQ(mp::MemorySize{to_str(val) + "M"}.in_kilobytes(), val * kilo);
```

The second form follows DRY in that it avoids repeating the (empty) format everywhere, and it's just a little bit easier to read: `to_str` has a lighter syntax and immediately makes the intent clear.

ricab:
I don't know, I tend to think that adding support for locales would have to be a holistic task, where we visited every input/output...

sharder996:
Adding support for different locales isn't an issue right now and it doesn't block this.

bors r+

PR #2899, opened 2023-01-23
[icons] Use new Multipass logo assets

### LINKED ISSUE #2895: [linux] Update Multipass icon for snap package and Snap Store
We are switching to a new logo, so the snap package needs to be updated to use this new asset.

Along with this, the Snap Store will need the Multipass logo updated as well.

townsend2010:
Fixes #2895


andrei-toterman:
This was resolved in #3282 as part of moving to the new Flutter GUI.

PR #2903, opened 2023-01-25
Mount handler tests

andrei-toterman:
Follow-up to #2835 to add unit tests for the new mount handlers implementations.

PR #2910, opened 2023-02-02
[mount_handler/sshfs] Get SSH port when starting the mount

### LINKED ISSUE #2904: [regression, virtualbox] Daemon removes mounts on startup & fails to start instance
**Describe the bug**
Hangs on the following screen after the update
image

Will try get some more logs tomorrow. But i'm assuming its the version change thats broke it. It was working before.

New containers start fine. I'm using --network en0 on macos latest os intel

Please just close if not an issue.

**To Reproduce**
How, and what happened?
1. update to *.11 from .10

**Expected behavior**
Expected container to boot

**Logs**

**Additional info**
 - OS: [e.g. macOS 10.15]
- `multipass version`
- `multipass info --all`

**Additional context**

townsend2010:
This avoids the case when the port may not be available yet or chages.

Fixes #2904

### DIFF
--- a/src/sshfs_mount/sshfs_mount_handler.cpp
+++ b/src/sshfs_mount/sshfs_mount_handler.cpp
@@ -114,7 +114,7 @@ SSHFSMountHandler::SSHFSMountHandler(VirtualMachine* vm, const SSHKeyProvider* s
     : MountHandler{vm, ssh_key_provider, target, mount},
       process{nullptr},
       config{"",
-             vm->ssh_port(),
+             0,
              vm->ssh_username(),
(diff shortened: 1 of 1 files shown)


ricab (approved):
This looks good and works well. I confirm it fixed the issue on macOS. Thank you!

Before merging, I imagine this will have an impact on #2903 though. @townsend2010, @andrei-toterman, how do you want to coordinate with for that?

townsend2010:
Hey @ricab, thanks for the review!

I think this needs to get in now and the unit tests PR will need to be adapted for this.

PR #2918, opened 2023-02-09
[readme] Replace mention of Hyperkit with QEMU

ricab:
Not sure if we should say "QEMU on Linux and macOS, and Hyper-V on Windows". Or maybe even list all the drivers? Or not list them at all and just say "It works with multiple hypervisors and runs VMs with minimal overhead"?


naturedamends:
But its not that simple. I only use hyperv and virtual box on macOS

PR #2932, opened 2023-02-20
Detect mount active state

### LINKED ISSUE #2917: [macOS] Inconsistent mounting when using `mount --type native` on restart
**Describe the bug**

Hello team. Really love the new `mount --type native` option that landed in `1.11.0` - nice to not have SSHFS along for the ride. But I've found a little bug/quirk - that I can now reproduce.

I noted that after modifying an existing instance to re-mount using `mount --type native` - I noted that on restart of the instance - the native mount would fail sometimes.

Turns out I've been able to isolate it to when the instance is either shutdown/restart within the instance itself - rather than using `multipass stop` / `multipass restart`.

**To Reproduce**

1. `multipass start INSTANCE_NAME`
2. `multipass shell INSTANCE_NAME`
3. Confirm mount point exists
4. `sudo shutdown now -P`
5. `multipass start INSTANCE_NAME`
6. `multipass shell INSTANCE_NAME`
7. Mount will no longer exist....
8. Now `multipass stop INSTANCE_NAME`
9. `multipass start INSTANCE_NAME`
10. `multipass shell INSTANCE_NAME`
11. Mount is now back 🎉 

Can also use `multipass restart` as well to "fix" an image broken via `shutdown / reboot` operations from within the instance itself.

Sorry, hope that's clear!

**Expected behavior**

The native mount continues to work on restart of instance.

**Additional info**

- OS: macOS `12.6.3`
- `multipass version`: `multipass   1.11.1+mac`
- `multipass info --all`

```
Name:           magnetik-docker
State:          Running
IPv4:           192.168.65.2
                192.168.0.61
                172.19.0.1
                172.18.0.1
                172.21.0.1
                172.17.0.1
                172.23.0.1
                172.20.0.1
                172.22.0.1
                172.24.0.1
Release:        Ubuntu 20.04.5 LTS
Image hash:     0f1a20bcda28 (Ubuntu 20.04 LTS)
CPU(s):         1
Load:           0.27 0.08 0.03
Disk usage:     8.1GiB out of 9.5GiB
Memory usage:   182.6MiB out of 970.2MiB
Mounts:         /Users/magnetik => /Users/magnetik
                    UID map: 501:default
                    GID map: 20:default
```

andrei-toterman:
This PR modifies the mount handlers by adding specialized checks for whether a mount is still active or not. It also makes use of the way in which some backends can notify the daemon when a VM reboots, so that the mounts can also be stopped. This allows the mount handlers to have a consistent state when a VM is shut down internally and, in those specific backend cases, when rebooted internally.

fix #2917

### DIFF
--- a/include/multipass/mount_handler.h
+++ b/include/multipass/mount_handler.h
@@ -45,39 +45,42 @@ public:
     void start(ServerVariant server, std::chrono::milliseconds timeout = std::chrono::minutes(5))
     {
         std::lock_guard active_lock{active_mutex};
-        if (is_active())
-            return;
-        start_impl(server, timeout);
+        if (!is_active())
+            start_impl(server, timeout);
         active = true;
     }
 
     void stop(bool force = false)
     {
         std::lock_guard active_lock{active_mutex};
-        if (!is_active())
-            return;
-        stop_impl(force);
+        if (is_active())
+            stop_impl(force);
         active = false;
+    }
+
+    virtual bool is_active()
+    {
+        return active;
     }
 
 protected:
     MountHandler() = default;
     MountHandler(VirtualMachine* vm, const SSHKeyProvider* ssh_key_provider, const std::string& target,
-                 const VMMount& mount)
-        : vm{vm}, ssh_key_provider{ssh_key_provider}, target{target}, active{false}
+                 const std::string& source)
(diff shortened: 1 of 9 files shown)


sharder996 (approved):
LGTM, just one change needed on the private side. Looks like there aren't really any meaningful tests that can be added to cover the new lines so we'll just have to take the hit.

sharder996:
Private side looks good now as well.

bors merge

PR #2935, opened 2023-02-22
[backend/qemu] Increase shutdown/suspend timeout to 5 minutes

### LINKED ISSUE #2933: [qemu] Allow more than 30 seconds to shutdown/suspend an instance
Currently, only 30 seconds are allowed to shutdown or suspend an instance before it kills `qemu`.  We should allow more time as the instance or host machine is busy.

townsend2010:
Fixes #2933

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -57,6 +57,8 @@ constexpr auto mount_data_key = "mount_data";
 constexpr auto mount_source_key = "source";
 constexpr auto mount_arguments_key = "arguments";
 
+constexpr int timeout = 300000; // 5 minute timeout for shutdown/suspend
+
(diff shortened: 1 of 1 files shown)


sharder996:
We're doing some similar with LXD [here](https://github.com/canonical/multipass/blob/0c58deb74c38bfc364311cae38b4b14b3646fec9/src/platform/backends/lxd/lxd_virtual_machine.cpp#L189) that has been left with a comment that we should probably be passing in the timeout from the daemon (or at least consulting it). The engineering required to do that might not be feasible at this time, but maybe we should create a ticket so it doesn't get buried.

sharder996 (approved):
Ok, LGTM! Just didn't want to lose track of the other issue.

bors merge

sharder996:
bors retry

townsend2010:
Last time...

bors retry

Issue #2936, opened 2023-02-28
[cli/info] `multipass info --all` does not list deleted instances


andrei-toterman:
**Describe the bug**
Listing information about all instances does not include deleted instances. `multipass info` works on a deleted instance, `multipass ls` includes deleted instances, so `multipass info --all` should include the deleted instances info as well.

**To Reproduce**
1. `multipass launch -n test`
1. `multipass delete test`
1. `multipass info test` shows info about deleted instance
1. `multipass info --all` does not show info about the deleted instance

**Expected behavior**
Show info about all instances on `multipass info --all`

**Additional info**
- OS: any
- `multipass version` 1.11.1
- `multipass get local.driver` any

Aryansingh0103:
I would like to work on this can you please assign me this work @andrei-toterman

Aryansingh0103:
Sir can you help me by suggesting where this file is as I am searching for it, I just want to know where the file is

andrei-toterman:
Sure! You should look inside `src/daemon/daemon.cpp`.

georgeliao:
Since **multipass info --all** does show the state of the instance, it makes sense to show the deleted instances as well. Apart from that, the fact that you manually run it and find it out this behavior does not seem to be right. My wondering is that whether this can be unit tested. If so, maybe unit tests should be added.

ricab:
Indeed, this should be covered by unit tests. We have been steadily improving our test coverage over the years, but there are still blind spots. I agree that implementing tests for this should be part of fixing this issue.

Aryansingh0103:
thankyou sir, I think my code will work but it has some flaws also in it but I'm still trying to improve it .

PR #2944, opened 2023-03-02
#2936  Issue on [cli/info] multipass info --all does not list deleted instances : modify the list_instances function in the multipass-info.cpp and also…

Aryansingh0103:
… added main  function to pass the show_deleted flag to it.

I've added a show_deleted parameter to control whether information about deleted instances should be included in the output. I've also added a check for deleted instances and a loop to print information about each deleted instance if the show_deleted flag is set. And In order to use the modified list_instances function, we need to modify the info_main function to pass the show_deleted flag to it.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,20 @@
+{
+    "configurations": [
+        {
+            "name": "Win32",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [
+                "_DEBUG",
+                "UNICODE",
+                "_UNICODE"
+            ],
+            "compilerPath": "C:\\Program Files (x86)\\mingw-w64\\i686-8.1.0-posix-dwarf-rt_v6-rev0\\mingw32\\bin\\gcc.exe",
+            "cStandard": "c17",
+            "cppStandard": "gnu++14",
+            "intelliSenseMode": "windows-gcc-x86"
+        }
+    ],
+    "version": 4
+}
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -92,3 +92,4 @@ mp::ParseCode cmd::Info::parse_args(mp::ArgParser* parser)
 
     return status;
 }
+
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -369,6 +369,81 @@ std::unordered_map<std::string, mp::VMSpecs> load_db(const mp::Path& data_path, 
     }
     return reconstructed_records;
 }
+int list_instances(bool show_deleted) {
+    std::vector<instance::Instance> instances = instance::list_all();
+    std::vector<instance::Instance> deleted_instances = instance::list_deleted();
+
+    if (instances.empty() && deleted_instances.empty()) {
+        std::cout << "No instances found." << std::endl;
+        return EXIT_SUCCESS;
+    }
+
+    // Sort instances by name.
+    std::sort(instances.begin(), instances.end(), [](const auto& a, const auto& b) {
+        return a.get_name() < b.get_name();
+    });
+
+    // Print info for each instance.
+    for (const auto& inst : instances) {
+        std::cout << "Name: " << inst.get_name() << std::endl;
+        std::cout << "State: " << inst.get_state() << std::endl;
+        std::cout << "IPv4: " << inst.get_ipv4() << std::endl;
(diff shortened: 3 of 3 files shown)


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,20 @@
+{
```
georgeliao, line 1: Why is this vscode file here? Is it something local to you and it was added accidentally?

andrei-toterman (changes requested):
Hi, @Aryansingh0103! I'm glad that you've decided to try to contribute to Multipass!

First things first: I see that you've committed your `.vscode` directory. That should definitely be avoided, as it's specific only to your setup. To be honest, we could also improve our `.gitignore` file with some more common entries like that one.

Now, for the actual fix itself, let me give you a brief explanation about how Multipass functions. There are two main components: the daemon `multipassd` and the CLI `multipass`. The daemon is a process that always runs in the background and manages the state of the virtual machines. The CLI is a tool that you use in your terminal in order to interact with the daemon, by sending it requests and displaying its replies back to the user.

When you do `multipass info --all` in the terminal, you use the `multipass` CLI to send a request to the `multipassd` daemon to give you all the information about the instances. The request that is being sent is correct, but the daemon does not return a complete reply. Therefore, we have nothing to fix within the CLI, only within the daemon code.

Each request that the daemon may receive is handled by a related method inside the `Daemon` class. In the case of `multipass info`, the request is handled inside the `Daemon::info` method. Inside that method, we look at the received request and we construct a reply with the appropriate information. The important part of the code for this issue is this one, found [here](https://github.com/canonical/multipass/blob/main/src/daemon/daemon.cpp#LL1346-L1357)
```cpp
std::vector<decltype(vm_instances)::key_type> instances_for_info;

if (request->instance_names().instance_name().empty())
{
    for (auto& pair : vm_instances)
        instances_for_info.push_back(pair.first);
}
else
{
    for (const auto& name : request->instance_names().instance_name())
        instances_for_info.push_back(name);
}
```

The first vector, `instances_for_info`, is a vector of strings representing the names of the VMs that we want to query for information. In the following `if` statement, we check if the request we received has any VM names specified. If it doesn't, as in the case of `multipass info --all`, we go through our non-deleted instances, which we keep inside the `vm_instances` hashmap, and we add the name of that VM, which is used as a key in the hashmap and retrieved from `pair.first`, to the list of VM names that we're interested into. Conversely, if we do specify VM names in `multipass info`, we go on the `else` branch and we only add the names that we specified to `instances_for_info`.

There is no issue with what happens with `instances_for_info` after this point. The issue is that we keep deleted and non-deleted VMs in two different hashmaps. The non-deleted ones are, as seen, inside `vm_instances`. The deleted ones are inside a hashmap called `deleted_instances`. And when we want to add all the VMs to `instances_for_info`, we only iterate through `vm_instances`, but not through `deleted_instances` as well. So I suggest that you do something similar with
```cpp
for (auto& pair : vm_instances)
    instances_for_info.push_back(pair.first);
```
but for the `deleted_instances` as well.

I hope this explanation clears things up a bit. Don't hesitate to update the PR as many times as needed :smile:

Aryansingh0103:
can I add this code 
std::vector<decltype(vm_instances)::key_type> instances_for_info;

for (auto& pair : vm_instances)
{
    instances_for_info.push_back(pair.first);
}

for (auto& pair : deleted_instances)
{
    auto it = std::find(instances_for_info.begin(), instances_for_info.end(), pair.first);
    if (it == instances_for_info.end())
    {
        instances_for_info.push_back(pair.first);
    }
}

With these changes, the instances_for_info vector will contain the names of all non-deleted and deleted VMs, ensuring that all VMs are included in the output of the multipass info --all command. The code also checks for duplicates before adding VM names to the instances_for_info vector to avoid including the same VM multiple times.

.vscode/c_cpp_properties.json
Aryansingh0103: it was added accidentally

andrei-toterman:
You're on the right track! It's true that we don't want duplicates in the `instances_for_info` vector, but since `vm_instances` holds all non-deleted instances and `deleted_instances` holds all the deleted ones, there is no overlap between them, so you don't need to manually check for duplicates, as they simply won't happen.

andrei-toterman:
Hi, @Aryansingh0103! Unfortunately I must close this PR, as we need a fix for this very soon. I hope you'll find other opportunities to contribute to Multipass or other open source projects :smile:

PR #2951, opened 2023-03-06
[image vault] Refactor error handling

sharder996:
Refactor where exceptions are thrown when no images are found in the vault. It should not be up to the vault to decide that no images being found should be interpreted as an error. So, this PR removes that behavior and leaves it up to the calling class.


ricab:
Hey @sharder996, to confirm, you don't mean to merge this on its own anymore, do you?

sharder996:
@ricab If it all looks good and can be merged with [#2701](https://github.com/canonical/multipass/pull/2701), then yes, this can be closed.

ricab:
@sharder996 how about now, can we close?

PR #2952, opened 2023-03-06
[cli] Fix param name in lambda

ricab:


luis4a0 (approved):
Excellent, thanks for this @ricab!

ricab:
Thanks for the quick review @luis4a0 !

PR #2963, opened 2023-03-11
[cli] Stop erroneous newlines from appearing in empty terminal

### LINKED ISSUE #2953: [regression/edge] Output on launch is showing many new lines now
When launching a new instance using the snap from `edge`, there are many newlines now displayed.  I think this is a regression caused by #2779.

Example:
```
$ multipass launch -n test









Launched: test

$
```

sharder996:
Add an extra ASCII escape character that moves the cursor up 0 lines. This stops newlines from appearing in a new/empty terminal when running a command that uses the animated spinner. Don't ask exactly why this works, but it does :shrug: 

---

Fixes #2953

### DIFF
--- a/src/client/cli/cmd/animated_spinner.cpp
+++ b/src/client/cli/cmd/animated_spinner.cpp
@@ -25,7 +25,7 @@ namespace
 {
 void clear_line(std::ostream& out)
 {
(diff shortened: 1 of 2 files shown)


andrei-toterman (approved):
Thanks, @sharder996! Seems to work!

bors merge

PR #2968, opened 2023-03-14
[cli] Show number of snapshots in `multipass info`

sharder996:
Just did the other formatters right away as well

---

https://github.com/canonical/multipass/projects/16#card-88442204


ricab (approved):
Great!

PR #2979, opened 2023-03-20
[cli] Only update progress when percentage goes up

### LINKED ISSUE #1532: [windows] download progress flickering
**Describe the bug**
Progress updates in all of cmd.exe, powershell, windows terminal cause flicker of the whole line. It looks that the line is updated more often than necessary.

**To Reproduce**
1. Install Multipass fresh
1. `multipass launch`

**Expected behavior**
Progress update only affects the terminal when the download percentage changes.

**Additional info**
 - OS: Windows 10
- multipass-1.2.1

sharder996:
Fixes #1532

### DIFF
--- a/src/platform/backends/lxd/lxd_vm_image_vault.cpp
+++ b/src/platform/backends/lxd/lxd_vm_image_vault.cpp
@@ -522,6 +522,7 @@ void mp::LXDVMImageVault::poll_download_operation(const QJsonObject& json_reply,
                           .arg(json_reply["metadata"].toObject()["id"].toString()));
 
         // Instead of polling, need to use websockets to get events
+        auto last_download_progress = -2;
         while (true)
         {
             try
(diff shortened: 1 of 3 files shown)


src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -522,6 +522,7 @@ void mp::LXDVMImageVault::poll_download_operation(const QJsonObject& json_reply,
                           .arg(json_reply["metadata"].toObject()["id"].toString()));
 
         // Instead of polling, need to use websockets to get events
+        auto last_download_progress = -2;
```
sharder996, line 525: I use `-2` because there are two different states that are being checked here. The above reply can either return a percentage for download progress (located under `download_progress -> rootfs` in the json reply, or a percentage for metadata download progress (located under `download_progress -> metadata`. The above helper function `parse_percent_as_int()` will return `-1` when downloading metadata, so a starting value of `-2` is used so that the downloading message will still be displayed in the cli.

sharder996:
Hey @townsend2010, thanks for the review!

It shouldn't be flickering, except when the percentage changes. The whole line gets updated then so it may flash briefly. Also, this PR only affects the download percentage message so other messages, such as those that have a spinner, may still flicker because of how often they get updated.

PR #2984, opened 2023-03-21
[cli] Add valid name desc to help

### LINKED ISSUE #1277: [documentation] 'launch' should explain what a valid instance name is
During a conversation with a user on IRC, there was a question about why underscores are not allowed for instance names.  I explained that an instance name needs to be a valid hostname.  The user then pointed out that our documentation fails to explain what a valid instance name is.  We should document that :grin:

sharder996:
Add an explanation to the `launch --help` for what valid instance names are.

TODO: If the wording looks good, [multipass.run](https://multipass.run/docs/launch-command) docs needs to be updated as well

Fixes #1277

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -224,9 +224,7 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                          "letter, and must end with an alphanumeric character."};
     const auto name_option_desc =
         petenv_name.isEmpty()
-            ? QString{"Name for the instance.\nValid names must consist of letters, numbers, or hyphens, must start "
-                      "with a letter, and must end with an alphanumeric character.\n%3"}
-                  .arg(valid_name_desc)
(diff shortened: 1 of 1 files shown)


src/client/cli/cmd/launch.cpp
```diff
@@ -220,12 +220,16 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
         QString::fromUtf8(default_memory_size));
     memOptionDeprecated.setFlags(QCommandLineOption::HiddenFromHelp);
 
+    const auto valid_name_desc = QString{"Valid names must consist of letters, numbers, or hyphens, must start with a "
+                                         "letter, and must end with an alphanumeric character."};
     const auto name_option_desc =
         petenv_name.isEmpty()
-            ? QString{"Name for the instance."}
+            ? QString{"Name for the instance.\nValid names must consist of letters, numbers, or hyphens, must start "
+                      "with a letter, and must end with an alphanumeric character.\n%3"}
```
ricab, line 228: Did you mean to replace this?

ricab (approved):
Yup, thanks!

bors r+

PR #2998, opened 2023-03-29
tweak snapshot cmd

ricab:
Turn name into an option in `snapshot` and allow `-c` for comments.


sharder996 (approved):
I think this is good. @ricab, I'll leave merging up to you if you want someone else to take a look at this as well.

ricab:
Hmm, let's merge. I mentioned it in the meeting and it's only going into the feature branch. We can always revert.

Thanks for reviewing @sharder996!

PR #3000, opened 2023-03-29
[building and passing tests] Snapshots info RPC

sharder996:
Something to build the different `multipass info` queries off of.

### DIFF
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -26,15 +26,15 @@ namespace cmd = multipass::cmd;
 
 namespace
 {
-std::vector<mp::InstanceOrSnapshot> add_instance_and_snapshot_names(const mp::ArgParser* parser)
+std::vector<mp::InstanceSnapshotPair> add_instance_and_snapshot_names(const mp::ArgParser* parser)
 {
-    std::vector<mp::InstanceOrSnapshot> instance_snapshot_names;
+    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names;
 
     for (const auto& arg : parser->positionalArguments())
     {
         const auto tokens = arg.split('.');
 
-        mp::InstanceOrSnapshot inst_snap_name;
+        mp::InstanceSnapshotPair inst_snap_name;
         inst_snap_name.set_instance_name(tokens[0].toStdString());
         if (tokens.size() > 1)
             inst_snap_name.set_snapshot_name(tokens[1].toStdString());
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -29,14 +29,16 @@ namespace
 // TODO@snapshots move this to common_cli once required by other commands
 std::vector<mp::InstanceSnapshotPair> add_instance_and_snapshot_names(const mp::ArgParser* parser)
 {
-    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names(parser->positionalArguments().count());
+    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names;
+    instance_snapshot_names.reserve(parser->positionalArguments().count());
 
     for (const auto& arg : parser->positionalArguments())
     {
         mp::InstanceSnapshotPair inst_snap_name;
-        inst_snap_name.set_instance_name(arg.left(arg.indexOf('.')).toStdString());
-        if (arg.indexOf('.') >= 0)
-            inst_snap_name.set_snapshot_name(arg.right(arg.length() - arg.indexOf('.') - 1).toStdString());
+        auto index = arg.indexOf('.');
+        inst_snap_name.set_instance_name(arg.left(index).toStdString());
+        if (index >= 0)
+            inst_snap_name.set_snapshot_name(arg.right(arg.length() - index - 1).toStdString());
 
         instance_snapshot_names.push_back(inst_snap_name);
     }
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -24,6 +24,28 @@
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
 
+namespace
+{
+std::vector<mp::InstanceOrSnapshot> add_instance_and_snapshot_names(const mp::ArgParser* parser)
+{
+    std::vector<mp::InstanceOrSnapshot> instance_snapshot_names;
+
+    for (const auto& arg : parser->positionalArguments())
+    {
+        const auto tokens = arg.split('.');
+
+        mp::InstanceOrSnapshot inst_snap_name;
+        inst_snap_name.set_instance_name(tokens[0].toStdString());
+        if (tokens.size() > 1)
+            inst_snap_name.set_snapshot_name(tokens[1].toStdString());
+
+        instance_snapshot_names.push_back(inst_snap_name);
+    }
+
+    return instance_snapshot_names;
+}
+} // namespace
+
 mp::ReturnCode cmd::Info::run(mp::ArgParser* parser)
 {
     auto ret = parse_args(parser);
@@ -85,7 +107,8 @@ mp::ParseCode cmd::Info::parse_args(mp::ArgParser* parser)
     if (parse_code != ParseCode::Ok)
         return parse_code;
 
-    request.mutable_instance_names()->CopyFrom(add_instance_names(parser));
+    for (const auto& item : add_instance_and_snapshot_names(parser))
+        request.add_instances_snapshots()->CopyFrom(item);
     request.set_no_runtime_information(parser->isSet(noRuntimeInfoOption));
 
     status = handle_format_option(parser, &chosen_formatter, cerr);
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -26,18 +26,17 @@ namespace cmd = multipass::cmd;
 
 namespace
 {
+// TODO@snapshots move this to common_cli once required by other commands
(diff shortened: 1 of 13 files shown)


src/daemon/daemon.cpp
```diff
@@ -1672,23 +1674,27 @@ try // clang-format on
...
                 mpu::run_in_ssh_session(session, "cat /etc/os-release | grep 'PRETTY_NAME' | cut -d \\\" -f2");
-            info->set_current_release(!current_release.empty() ? current_release : original_release);
+            instance_info->set_current_release(!current_release.empty() ? current_release : original_release);
         }
         return grpc::Status::OK;
     };
 
+    mp::InstanceNames instance_names;
+    for (const auto& n : request->instances_snapshots())
+        instance_names.add_instance_name(n.instance_name());
```
ricab, line 1694: This is only valid until snapshots are part of the deal. Could you please add a `TODO@snapshots`?

ricab (commented):
OK, this looks good. Only a few nitpicks and remarks.

src/client/cli/formatter/json_formatter.cpp
```diff
@@ -63,29 +63,31 @@ std::string mp::JsonFormatter::format(const InfoReply& reply) const
 
     info_json.insert("errors", QJsonArray());
 
-    for (const auto& info : reply.info())
+    for (const auto& info : reply.detailed_report().details())
```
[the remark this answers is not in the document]
sharder996: Key-value pairs in JSON are automatically sorted in alphabetical order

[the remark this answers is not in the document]
ricab: Hmm, you mean Qt does that, right? I think JSON itself is unordered (apart from arrays). Either way, we can't guarantee a specific ordering, so that's fine.

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -58,42 +58,47 @@ std::string mp::TableFormatter::format(const InfoReply& reply) const
 {
     fmt::memory_buffer buf;
 
-    for (const auto& info : format::sorted(reply.info()))
+    for (const auto& info : format::sorted(reply.detailed_report().details()))
```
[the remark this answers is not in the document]
sharder996: Currently no update needed since the required message structure stayed the same (eg. `details()` is a repeated field where each item containing a subfield named `name`. This is still true after the rpc changes)

src/client/cli/cmd/info.cpp
```diff
@@ -24,6 +24,28 @@
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
 
+namespace
+{
+std::vector<mp::InstanceSnapshotPair> add_instance_and_snapshot_names(const mp::ArgParser* parser)
```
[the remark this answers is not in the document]
sharder996: Yup, isn't required anywhere else at the moment, so I just put it here.

src/client/cli/formatter/table_formatter.cpp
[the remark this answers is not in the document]
ricab: Right, but once we start including snapshots they need to appear last. The relevant paragraph from the spec:

> Snapshot and instance arguments can be mixed. In that case, all requested instances will be displayed first, sorted alphabetically (as today), followed by all requested snapshots, sorted with the same criteria as in the snapshot overview. An MVP may refuse such mixing. The --all flag will now designate all instances and all snapshots.

src/client/cli/cmd/info.cpp
```diff
@@ -24,6 +24,28 @@
...
+    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names;
+
+    for (const auto& arg : parser->positionalArguments())
+    {
+        const auto tokens = arg.split('.');
+
+        mp::InstanceSnapshotPair inst_snap_name;
+        inst_snap_name.set_instance_name(tokens[0].toStdString());
+        if (tokens.size() > 1)
+            inst_snap_name.set_snapshot_name(tokens[1].toStdString());
```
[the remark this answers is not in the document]
sharder996: Instead of using split and possibly losing some part of the string, I changed it to set the instance name to everything before the first occurrence of a `.` if it exists and the snapshot to everything after. Since `.` is an invalid character for instances and snapshots regardless the daemon will return no results found if more periods/leading/ending periods are entered.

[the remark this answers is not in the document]
ricab: OK. I am usually an adept of fail-fast, but I guess we can leave it up to the daemon to decide what names are and are not acceptable.

src/client/cli/cmd/info.cpp
```diff
@@ -26,18 +26,17 @@ namespace cmd = multipass::cmd;
 
 namespace
 {
+// TODO@snapshots move this to common_cli once required by other commands
 std::vector<mp::InstanceSnapshotPair> add_instance_and_snapshot_names(const mp::ArgParser* parser)
 {
-    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names;
+    std::vector<mp::InstanceSnapshotPair> instance_snapshot_names(parser->positionalArguments().count());
```
ricab, line 32: In this case, the difference is irrelevant, so I am not going to ask you to change it any further, but this approach has the downside that you perform a bunch of unnecessary initializations: N pairs and 2N strings. It doesn't work with types that are not default constructible either. Finally, if for any reason you overestimate N, you end up with the wrong size. The canonical approach is:

```
std::vector<T> vec{};
vec.reserve(N);
```

Note that `reserve` this does not change size, only capacity. That's what we need to avoid growing the vector and copying things around. It's a pity that you need 2 lines, but that's the interface of `std::vector`.

src/client/cli/cmd/info.cpp
```diff
@@ -26,18 +26,17 @@ namespace cmd = multipass::cmd;
...
-
         mp::InstanceSnapshotPair inst_snap_name;
-        inst_snap_name.set_instance_name(tokens[0].toStdString());
-        if (tokens.size() > 1)
-            inst_snap_name.set_snapshot_name(tokens[1].toStdString());
+        inst_snap_name.set_instance_name(arg.left(arg.indexOf('.')).toStdString());
+        if (arg.indexOf('.') >= 0)
+            inst_snap_name.set_snapshot_name(arg.right(arg.length() - arg.indexOf('.') - 1).toStdString());
 
         instance_snapshot_names.push_back(inst_snap_name);
```
ricab, line 41: Oh actually you need to change, either above or here, otherwise you are adding elements on top of what you constructed.

src/client/cli/cmd/info.cpp
```diff
@@ -26,18 +26,17 @@ namespace cmd = multipass::cmd;
...
     {
-        const auto tokens = arg.split('.');
-
         mp::InstanceSnapshotPair inst_snap_name;
-        inst_snap_name.set_instance_name(tokens[0].toStdString());
-        if (tokens.size() > 1)
-            inst_snap_name.set_snapshot_name(tokens[1].toStdString());
+        inst_snap_name.set_instance_name(arg.left(arg.indexOf('.')).toStdString());
+        if (arg.indexOf('.') >= 0)
+            inst_snap_name.set_snapshot_name(arg.right(arg.length() - arg.indexOf('.') - 1).toStdString());
```
ricab, line 39: You are repeatedly searching for `'.'` here. Better do it once, save it in a variable, and use that.

ricab (changes requested):
Getting there, but I believe there's a bug now.

src/client/cli/cmd/info.cpp
sharder996: I thought there were cases similar to this where the compiler automatically optimizes this out. Regardless, fixed it.

src/client/cli/cmd/info.cpp
sharder996: Right..., that function slipped my mind.

src/client/cli/cmd/info.cpp
ricab: It might, but it also might not. Little details can affect that and I don't know if the compiler has enough information to decide that that particular function will return the same thing in all calls. Best rely on what the compiler is required to do.

src/client/cli/cmd/info.cpp
```diff
@@ -29,14 +29,16 @@ namespace
...
     for (const auto& arg : parser->positionalArguments())
     {
         mp::InstanceSnapshotPair inst_snap_name;
-        inst_snap_name.set_instance_name(arg.left(arg.indexOf('.')).toStdString());
-        if (arg.indexOf('.') >= 0)
-            inst_snap_name.set_snapshot_name(arg.right(arg.length() - arg.indexOf('.') - 1).toStdString());
+        auto index = arg.indexOf('.');
+        inst_snap_name.set_instance_name(arg.left(index).toStdString());
+        if (index >= 0)
+            inst_snap_name.set_snapshot_name(arg.right(arg.length() - index - 1).toStdString());
```
ricab, line 41: Sorry, I still think this should verify the format before sending it over to the daemon. The client is parsing it anyway, so it already needs to understand the structure. And otherwise errors can be confusing:

```
$ mp info ..
info failed: The following errors occurred:
instance "" does not exist
```

ricab (changes requested):
Sorry, trying it out a little more I still think those arguments need a little more checking.

src/client/cli/cmd/info.cpp
ricab: Actually, let's leave this for a follow up PR. The argument should only be parsed when introducing snapshot details anyway. And then there are other things that need to be changed, like the help and other error messages. For instance:

```
$ mp info
Name argument or --all is required
```

sharder996:
CI looks good so merging on behalf of @ricab

Issue #3010, opened 2023-04-04
Daemon logs Qt warning first thing


ricab:
This threw me off when searching for a different problem:

```
$ sudo ./bin/multipassd --logger=stderr -V debug
[2023-04-04T10:05:24.949] [warning] [Qt] Empty filename passed to function
[...]
```

PR #3011, opened 2023-04-04
Fix Qt warning

ricab:
Avoid setting permissions on an empty dir. Also, let new storage dirs inherit permissions.

### DIFF
--- a/src/daemon/daemon_config.cpp
+++ b/src/daemon/daemon_config.cpp
@@ -124,7 +124,10 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
         else
             data_directory = MP_STDPATHS.writableLocation(StandardPaths::AppDataLocation);
     }
-    MP_PLATFORM.set_permissions(storage_path, QFileDevice::ReadOwner | QFileDevice::WriteOwner | QFileDevice::ExeOwner);
+
+    if (!storage_path.isEmpty())
+        MP_PLATFORM.set_permissions(storage_path,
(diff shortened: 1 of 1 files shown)


ricab:
Hey @sharder996, does this look alright to you? I know you worked on this ~recently. I moved permissions-setting on `storage_path` so that new sub-directories can inherit them.

sharder996 (commented):
Hey @ricab, what was the warning that Qt was giving you? Regardless, two things that I noticed:
- The order to setting permissions and creating the cache/data directory shouldn't matter because of permissions inheritance. I haven't checked this, but maybe the ordering fixes your warning.
- I avoided using `if (!storage_path.isEmpty())` when setting permissions the first time around because the permissions should be set regardless (eg. we want to overwrite/update permissions for an existing installation of Multipass when a user chooses to keep their data).

ricab:
Hey @sharder996, so the warning is the one in issue #3010. It was in a commit message and I am not sure why GH is not linking it properly...

The order change was a drive-by thing, just because I noticed that data and cache directories would be created before setting the permissions on the parent, and so I thought they wouldn't inherit them (unless that function operates recursively?). 

The `if (!storage_path.isEmpty())` part was where the actual fix was. Without it, we were trying to set permissions on the path `""`, which would trigger the warning by Qt.

ricab:
OK, I made this address only the warning for the empty filename. #3013 addresses the ordering.

sharder996 (approved):
LGTM
bors merge

sharder996:
bors retry

sharder996:
one more time

bors retry

ricab:
Hey @sharder996, is that last merge intentional?

sharder996:
bors merge

PR #3012, opened 2023-04-04
Implement generic snapshot loading from disk

ricab:
Implement loading of snapshots that were previously persisted by Multipass.


ricab:
Declaring this good for what it does. Assumes alphabetic ordering of snapshot files for now. 

Further changes to come in the realm of persistence, but in a separate PR.

sharder996 (approved):
LGTM

PR #3023, opened 2023-04-06
[snapshots] Add snapshot overview to other formatters

sharder996:


ricab:
This is outdated now, could you please rebase?

src/client/cli/formatter/csv_formatter.cpp
```diff
@@ -50,32 +50,48 @@ std::string format_images(const google::protobuf::RepeatedPtrField<mp::FindReply
...
+    }
+    else if (reply.info_contents_case() == mp::InfoReply::kSnapshotOverview)
+    {
+        fmt::format_to(std::back_inserter(buf), "Instance,Snapshot,Parent,Comment\n");
 
-        fmt::format_to(std::back_inserter(buf), ",\"{}\";,{},{}\n", fmt::join(instance_details.ipv4(), ","),
-                       info.cpu_count(), instance_details.num_snapshots());
+        for (const auto& item : format::sort_snapshots(reply.snapshot_overview().overview()))
+        {
+            const auto snapshot = item.fundamentals();
```
ricab, line 89: Const references are preferable in such cases. If it returns by value, the lifetime of the temporary is extended by the const ref. If the function returns by reference (as here), you avoid a copy.

src/client/cli/formatter/json_formatter.cpp
```diff
@@ -63,90 +63,117 @@ std::string mp::JsonFormatter::format(const InfoReply& reply) const
...
+        }
+    }
+    else if (reply.info_contents_case() == mp::InfoReply::kSnapshotOverview)
+    {
+        std::unordered_map<std::string, QJsonObject> instance_map;
 
-            mounts.insert(QString::fromStdString(mount.target_path()), entry);
+        for (const auto& item : reply.snapshot_overview().overview())
+        {
+            const auto snapshot = item.fundamentals();
```
ricab, line 161: idem: const ref

src/client/cli/formatter/yaml_formatter.cpp
```diff
@@ -65,93 +65,116 @@ std::string mp::YamlFormatter::format(const InfoReply& reply) const
...
         }
-        instance_node["mounts"] = mounts;
-
-        info_node[info.name()].push_back(instance_node);
     }
+    else if (reply.info_contents_case() == mp::InfoReply::kSnapshotOverview)
+    {
+        for (const auto& item : format::sort_snapshots(reply.snapshot_overview().overview()))
+        {
+            const auto snapshot = item.fundamentals();
```
ricab, line 162: idem

src/client/cli/formatter/yaml_formatter.cpp
```diff
@@ -65,93 +65,116 @@ std::string mp::YamlFormatter::format(const InfoReply& reply) const
...
+            YAML::Node instance_node;
+            YAML::Node snapshot_node;
+
+            snapshot_node["parent"] = YAML::Null;
+            snapshot_node["comment"] = YAML::Null;
+            if (!snapshot.parent().empty())
+                snapshot_node["parent"] = snapshot.parent();
+            if (!snapshot.comment().empty())
+                snapshot_node["comment"] = snapshot.comment();
```
ricab, line 171: Why not use the ternary operator here? It would reduce 6 LoC to 2 here AFAICS (modulo formatting).

tests/test_output_formatter.cpp
```diff
@@ -253,13 +260,93 @@ auto construct_multiple_instances_info_reply()
...
+        entry->mutable_instance_status()->set_status(mp::InstanceStatus::SUSPENDED);
+        entry->mutable_instance_info()->set_image_release("18.10");
+        entry->mutable_instance_info()->set_id("1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd");
+    }
+    else if (reply.info_contents_case() == mp::InfoReply::kSnapshotOverview)
+    {
+        auto entry = reply.mutable_snapshot_overview()->add_overview();
+        entry->set_instance_name(petenv_name());
+        entry->mutable_fundamentals()->set_snapshot_name("snapshot1");
+        entry->mutable_fundamentals()->set_comment("An examplary comment");
```
ricab, line 348: "exemplary" :nerd_face:

ricab (changes requested):
Thank you Scott. Looks good in the main. I have a few smallish requests.

The first commit appears to be obsolete (empty). Can we drop it?

src/client/cli/formatter/yaml_formatter.cpp
sharder996: Doesn't work here because `YAML::Null` and `std::string` are different types.

src/client/cli/formatter/json_formatter.cpp
```diff
@@ -63,90 +63,117 @@ std::string mp::JsonFormatter::format(const InfoReply& reply) const
...
-            entry.insert("gid_mappings", mount_gids);
-            entry.insert("source_path", QString::fromStdString(mount.source_path()));
+            instance_info.insert("mounts", mounts);
+
+            info_obj.insert(QString::fromStdString(info.name()), instance_info);
+        }
+    }
+    else if (reply.info_contents_case() == mp::InfoReply::kSnapshotOverview)
+    {
+        std::unordered_map<std::string, QJsonObject> instance_map;
```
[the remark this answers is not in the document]
sharder996: The reason I had a separate map is because Qt JSON objects don't allow you insert into nested objects using a reference (like an iterator). And because I don't want to rely on the snapshot info being grouped together by instance, I had first grouped the info together in a map before inserting it all into the main `info_json` object.

I pushed an alternate approach, but it still has the downside of creating a new QJsonObject every time snapshot info has to be inserted for an instance that has already been seen in the loop.

src/client/cli/formatter/yaml_formatter.cpp
sharder996: Close, but this it turns into:
```
snapshot_node["parent"] = snapshot.parent().empty() ? YAML::Node(YAML::NodeType::Null) : YAML::Node(snapshot.parent());
```
It's getting longer, but still ends up saving a few LoC :smile:

ricab: Heh, true. But this seems to work here and avoid the line splitting:

```
snapshot_node["parent"] = snapshot.parent().empty() ? YAML::Node(YAML::Null) : YAML::Node(snapshot.parent());
```

I think just `YAML::Node{}` also creates a null node, but I am not sure.

src/client/cli/formatter/yaml_formatter.cpp
```diff
@@ -74,14 +73,11 @@ std::string mp::YamlFormatter::format(const InfoReply& reply) const
...
-        if (instance_details.current_release().empty())
-            instance_node["release"] = YAML::Null;
-        else
-            instance_node["release"] = instance_details.current_release();
-
-        instance_node["cpu_count"] = YAML::Null;
-        if (!info.cpu_count().empty())
-            instance_node["cpu_count"] = info.cpu_count();
+        instance_node["release"] = instance_details.current_release().empty()
+                                       ? YAML::Node(YAML::NodeType::Null)
```
ricab, line 77: You should be able to use just `YAML::Null` here too?

src/client/cli/formatter/yaml_formatter.cpp
```diff
@@ -74,14 +73,11 @@ std::string mp::YamlFormatter::format(const InfoReply& reply) const
...
+                                       : YAML::Node(instance_details.current_release());
+        instance_node["cpu_count"] =
+            info.cpu_count().empty() ? YAML::Node(YAML::NodeType::Null) : YAML::Node(info.cpu_count());
```
ricab, line 80: And here, and in the other cases below.

src/client/cli/formatter/json_formatter.cpp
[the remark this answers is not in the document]
ricab: OK, got it. I think we can keep the latter approach.

ricab (approved):
LGTM, yeyy!

bors r+

PR #3039, opened 2023-04-13
[image_host/custom] Fix wrong buildd daily image

townsend2010:


ricab (approved):
Yeah, that fixes it, thanks!

bors r+

PR #3048, opened 2023-04-18
[snapshots] Reposition num snapshots

sharder996:


ricab (approved):
Looking good, thank you!

PR #3050, opened 2023-04-19
Improve argument quoting

### LINKED ISSUE #2707: `docker-compose` alias does not behave the same as local `docker-compose`
**Describe the bug**
I was hoping to utilize Multipass' Docker preset [with `wp-env`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/) but have run into problems because a `docker-compose` alias doesn't behave the same way that a local `docker-compose` with the `DOCKER_HOST` environment variable set. 

When I run the `wp-env start` command, internally, it runs `child_process.spawn()` on `docker-compose`. This bypasses the local shell and just passes the arguments directly to the target executable. When I run this using a `docker:docker-compose` alias, I get the following error:

```
bash: -c: line 1: syntax error near unexpected token `('
bash: -c: line 1: `'cd' '/Users/obliviousharmony/Code/WordPress/gutenberg/packages/env'&&'docker-compose' '-f' '/Users/obliviousharmony/.wp-env/fcc19ccf5624bfb3d7f4b1258f0d4ea7/docker-compose.yml' 'run' '-T' 'tests-cli' 'bash' '-c' "set -eo pipefail && wp core install --url="localhost:8889" --title="env" --admin_user=admin --admin_password=password --admin_email=wordpress@example.com --skip-email && wp config set WP_DEBUG false --anchor="define( 'WP_DEBUG'," --raw && wp config set SCRIPT_DEBUG false --anchor="define( 'WP_DEBUG'," --raw && wp config set WP_ENVIRONMENT_TYPE "local" --anchor="define( 'WP_DEBUG'," && wp config set WP_PHP_BINARY "php" --anchor="define( 'WP_DEBUG'," && wp config set WP_TESTS_EMAIL "admin@example.org" --anchor="define( 'WP_DEBUG'," && wp config set WP_TESTS_TITLE "Test Blog" --anchor="define( 'WP_DEBUG'," && wp config set WP_TESTS_DOMAIN "localhost" --anchor="define( 'WP_DEBUG'," && wp config set WP_SITEURL "http://localhost:8889/" --anchor="define( 'WP_DEBUG'," && wp config set WP_HOME "http://localhost:8889/" --anchor="define( 'WP_DEBUG',""'
```

**To Reproduce**
How, and what happened?

1. Use the `docker` workflow to set up a Docker environment
2. Shell into the `docker` container and install NodeJS, I used v14 for this reproduction
3. Use `multipass alias docker:docker` to set up a local alias
4. Create a file named `2707-repro.js` in your home directory with this content:
```javascript
const { spawn } = require('child_process');

const test = spawn('docker', [
    'run',
    'ubuntu',
    'bash',
    '-c',
    `echo "test: ('complete')"`
]);

let output = '';
test.stdout.on('data', (data) => {
    output += data;
});

let errOutput = '';
test.stderr.on('data', (data) => {
    errOutput += data;
});

test.on('close', () => {
    console.log(output);
    if (errOutput != '') {
        console.error(errOutput);
        process.exit(-1);
    }
});

```
5. `cd` into your home directory on the host and run `node 2707-repro.js`. You should see the same error experienced with `wp-env`.
6. `cd` into the mapped home directory from a shell in the `docker` VM and run the above command again. You should see the `ls -la` and `echo` output from within the `ubuntu` Docker container without any error.

**Expected behavior**
I _was_ expecting it to behave exactly like running `docker-compose` locally. The [potentially relevant documentation](https://multipass.run/docs/exec-shells) seems to indicate that it should be processed in the same way that I expect, but, in practice, something seems to be going wrong.

**Logs**
This doesn't really seem related to be honest, everything is working perfectly fine!

**Additional info**
`multipass version`
```
multipass   1.10.1+Mac
multipassd  1.10.1+Mac
```
- `multipass info --all`
```
Name:           docker
State:          Running
IPv4:           192.168.64.2
                172.17.0.1
                172.18.0.1
Release:        Ubuntu 22.04.1 LTS
Image hash:     4247b344c976 (Ubuntu 22.04 LTS)
Load:           0.11 0.13 0.16
Disk usage:     4.8G out of 38.6G
Memory usage:   539.1M out of 3.8G
Mounts:         /Users/obliviousharmony => /Users/obliviousharmony
                    UID map: 501:default
                    GID map: 20:default
```
```
qemu
```

luis4a0:
Fixes #2707.

### DIFF
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -58,13 +58,6 @@ namespace
 constexpr auto category = "utils";
 constexpr auto scrypt_hash_size{64};
 
-auto quote_for(const std::string& arg, mp::utils::QuoteType quote_type)
-{
-    if (quote_type == mp::utils::QuoteType::no_quotes)
-        return "";
-    return arg.find('\'') == std::string::npos ? "'" : "\"";
-}
-
 QString find_autostart_target(const QString& subdir, const QString& autostart_filename)
 {
     const auto target_subpath = QDir{subdir}.filePath(autostart_filename);
@@ -223,7 +216,8 @@ std::string mp::utils::to_cmd(const std::vector<std::string>& args, QuoteType qu
     fmt::memory_buffer buf;
     for (auto const& arg : args)
     {
-        fmt::format_to(std::back_inserter(buf), "{0}{1}{0} ", quote_for(arg, quote_type), arg);
+        fmt::format_to(std::back_inserter(buf), "{0} ",
+                       quote_type == QuoteType::no_quotes ? arg : escape_for_shell(arg));
     }
 
     // Remove the last space inserted
--- a/tests/test_cli_client.cpp
+++ b/tests/test_cli_client.cpp
@@ -1585,7 +1585,7 @@ TEST_P(SSHClientReturnTest, execCmdWithDirWorks)
 
 INSTANTIATE_TEST_SUITE_P(Client, SSHClientReturnTest, Values(0, -1, 1, 127));
 
-TEST_F(Client, execCmdWithDirPrependsCd)
+/*TEST_F(Client, execCmdWithDirPrependsCd)
 {
     std::string dir{"/home/ubuntu/"};
     std::string cmd{"pwd"};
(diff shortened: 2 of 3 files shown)


townsend2010 (approved):
The code looks good and it's cool a function get dropped for this :)

I couldn't detect any regressions in my testing either.

LGTM!  Thanks!

bors merge

PR #2985, opened 2023-03-21
[daemon] Refactor instance selection

ricab:
Refactor how we select instances given instance name args, as well as how we react when there is a problem with those args.

PR #3056, opened 2023-04-25
[daemon] Filter duplicate instances with set

### LINKED ISSUE #3055: Multipass crashes on repeated cli arguments
**Describe the bug**
Passing repeated arguments causes the daemon to crash on `multipass delete`

**To Reproduce**
1. `multipass launch foo`
2. `multipass delete foo foo`

**Expected behavior**
The instance will be deleted

sharder996:
There are a number of different ways of only performing operations on unique instances. Chose to use a set to avoid iterating over `vm_instance_specs` multiple times. This is just a temporary fix if #2985 is merged in, as the issue is reproducible there as well.

Fixes #3055

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -718,17 +718,17 @@ auto find_requested_instances(const Instances& instances, const InstanceMap& vms
 template <typename Instances, typename InstanceMap>
 auto find_instances_to_delete(const Instances& instances, const InstanceMap& operational_vms,
                               const InstanceMap& trashed_vms)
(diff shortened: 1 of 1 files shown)


ricab:
Thank you @sharder996. I guess a conflict will show on whichever PR we try to merge later and the implementation needs adapting then. Since the daemon is such a central piece of code, and the refactoring affects a good part of it, I would like for it to be merged relatively soon (hopefully shortly after we release).

sharder996:
Closed as the issue will be solved in #3063

PR #3058, opened 2023-04-29
[url_downloader] Add HTTP UserAgent header info

### LINKED ISSUE #3049: Add Multipass-specific HTTP UserAgent header info
Add Multipass-specific HTTP UserAgent info when initiating a network connection.

townsend2010:
Fixes #3049

---

@nathan-j-hart This will be in the form of something like:
`Multipass/1.11.1 (macos-11.3; arm64)`

Let me know what you think.

### DIFF
--- a/src/network/url_downloader.cpp
+++ b/src/network/url_downloader.cpp
@@ -22,6 +22,8 @@
 #include <multipass/file_ops.h>
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
+#include <multipass/platform.h>
+#include <multipass/version.h>
 
 #include <QDir>
 #include <QEventLoop>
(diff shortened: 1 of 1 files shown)


nathan-j-hart:
Those are the must-have data points, as long as it is consumable by graylog / influxDB (presumably we can parse this however we want) then it should be good.

townsend2010:
I'm pretty sure graylog will just capture this as a string and then when we import it into influxDB, we'll have to parse things in the string how we see fit, but should be doable.

townsend2010:
Ok, after rebooting the integration test machine, this passed: https://github.com/canonical/multipass/actions/runs/4859933146

PR #3060, opened 2023-05-02
[build] Add Qt5::Concurrent to required libraries

rtgill82:


luis4a0:
Hi @rtgill82, thanks for the PR. Can you please explain why this PR is relevant, what does it do or fixes? Thanks!

rtgill82:
I'm building multipass on Gentoo which has Qt5 separated into a number of differernt packages (dev-qt/qtconccurent is the relevant package in this case). I just thought it would be helpful if a required dependency was caught earlier in the build process.

luis4a0:
Hi @rtgill82, thanks for the explanation. Looks good, let me just ask if that's the only necessary change to build in Gentoo.

rtgill82:
Yeah, that was the only necessary change for building. I was planning on providing a patch to allow falling back to using syslog for logging if systemd is unavailable.

luis4a0:
Ok, so let's merge this now. Thanks!

bors r+

PR #3066, opened 2023-05-05
Added ccache to cmakefile and updated the corresponding patches.

georgeliao:
This is a small improvement for developers who use ccache to accelerate the multipass build. With the change, the developer only has to install ccache and does not have to manually append the cmake arguments because it is taken care of by the cmake script itself.  
Reference: https://crascit.com/2016/04/09/using-ccache-with-cmake/


townsend2010 (approved):
Hey @georgeliao!

LGTM, thanks!

bors merge

georgeliao:
bors merge

PR #3068, opened 2023-05-09
[platform/logger] Fallback on syslog when systemd is unavailable

rtgill82:
Allow syslog to be used for logging when systemd is not available.


luis4a0:
Hi @rtgill82, thanks for the contribution. Please note that you are required to sign Canonical's CLA in order to contribute code to the company's projects; as explained [here](https://ubuntu.com/legal/contributors).

Speaking about the code, there is a linter error in your commit. We format our diffs through `clang-format` and continuous integration fails if the code doesn't follow that format. In order to check and fix indentation on your commit, you can `git add` the files and then run `git clang-format <files>`.

In the meantime, I'll check your code works. Thanks again!

luis4a0:
Strictly speaking about the code, the first thing I'd propose is to let the user decide whether to use systemd or syslog (defaulting to systemd because it's the current behavior). If the user chooses systemd but it's not detected, then CMake would fail and suggest to use syslog (this will make the user be aware of the logging facility they are using).

I'll continue the review after compiling and testing on my computer.

luis4a0:
Hi again @rtgill82!

After testing on my computer, I can confirm that the syslog logger works correctly. Summing up then, we'll be done with the three things I mentioned:
- signing the CLA
- indenting the code as requested by `clang-format`
- making the logger to use optional.

Thanks!

rtgill82:
I'll work on adding a CMake flag, so I suppose don't merge this yet. Also, I created a Launchpad account and signed the contributor agreement, but the cla-check doesn't seem to be passing yet.

rtgill82:
Would it be preferred if the variable be called `USE_SYSLOG` or `ENABLE_SYSLOG`?

luis4a0:
Hi @rtgill82!

About the CLA, maybe it's a matter of caching. Just to be sure: did you sign using the same mail address you're using in Github?

About the variable name, I'd prefer `MULTIPASS_ENABLE_SYSLOG`, in order to be consistent with other CMake variables names.

Thanks!

luis4a0:
In the check logs, I can see the following lines:
```
Checking the following users on Launchpad:
- rtgill82@gmail.com ✕ (has no Launchpad account)
```
Can you please confirm you used this address in your launchpad account?

luis4a0:
On the other hand, what is your Launchpad account? We can try to track down what happened.

rtgill82:
Ok, it looks like I created an account with Ubuntu One, but didn't link it with launchpad.net. Maybe it will work now.

luis4a0:
Great, it passed now! So I'll continue with the review. Thanks!

luis4a0:
Hi @rtgill82! I did another pass on your code. Although it works when compiling by hand, I am concerned now about the snap. On one hand, the journalling method would be chosen at run time, so the users of all distros can use the same snap. On the other hand, I don't think syslog is enabled by default in the snaps.

My questions thus are related to this. Are you planning to use Multipass from a compiled snap? Is your use case something I didn't consider? If you give us more insight on your use case, we can drive a solution togheter which works for us all. Thanks!

rtgill82:
No, I only intend to use a manually compiled version. The Snap version should be compiled with systemd by default, which I believe fits the majority of users' needs. I may be misunderstanding your question though, I don't have any experience with Snap.

luis4a0:
Hi @rtgill82, thanks for the answer. The packaging system we support is the snap, but it is completely valid to compile it by hand. The only caveat would be that, if something changes in the logging system in the future, we might not have the time to integrate the changes to the syslog logger. That said, we welcome your contribution!

src/platform/logger/syslog_logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+        return LOG_DEBUG;
+    case mpl::Level::error:
+        return LOG_ERR;
+    case mpl::Level::info:
+        return LOG_INFO;
+    case mpl::Level::warning:
+        return LOG_WARNING;
+    }
+    return 42;
+}
```
luis4a0, line 40: Since this is repeated in both loggers, what about putting this function in a new file? It ca be a method in a class `LinuxLogger`, inheriting from `Logger`? This way, `SyslogLogger` and `JournaldLogger` inherit from `LinuxLogger`.

src/platform/CMakeLists.txt
```diff
@@ -53,3 +55,7 @@ add_subdirectory(client)
 add_subdirectory(console)
 add_subdirectory(logger)
 add_subdirectory(update)
+
+if (JOURNALD_FOUND AND NOT MULTIPASS_ENABLE_SYSLOG)
+  add_definitions(-DJOURNALD_ENABLED)
```
luis4a0, line 60: I'd also prefix this definition with `MULTIPASS_`, so `MULTIPASS_JOURNALD_ENABLED` is coherent with most of the definitions added here.

luis4a0 (changes requested):
Hi @rtgill82, almost there. See two comments inline and we are ready to go.

rtgill82:
I've completed the requested changes. Everything compiles and seems to run as expected on my end. I'm less familiar with C++ than with C, so hopefully this is implemented as you expect.

rtgill82:
Looks like I have to implement tests.

luis4a0 (approved):
LGTM, thanks!

luis4a0:
bors retry

PR #3072, opened 2023-05-12
[image_host] Switch to using buildd Simplestreams for snapcraft remote

### LINKED ISSUE #3069: Use Simplestreams for getting Snapcraft buildd images
Now that Simplestreams is enabled for buildd daily images, we need to use that instead of the hard coded information.

townsend2010:
Fixes #3069

---

Corresponding private PR is at canonical/multipass-private#523.

There a couple of things to keep in mind with this PR:
1. The `snapcraft:` remote will no longer show when doing a `multipass find`.  I think this is fine since the `snapcraft` remote is really only intended for Snapcraft usage and should really have never been shown by default in the first place.  That said, `multipass find snapcraft:` will show the supported images.
2. To go along with `multipass find snapcraft:`, there are a couple of extra aliases that are shown from Simplestreams that we don't support in Multipass.  It would require a pretty big refactor to properly filter those out and now is not the time for that.  I think since the `snapcraft` remote would no longer be shown by default, this isn't a big deal and probably would never be noticed by a user.

### DIFF
--- a/include/multipass/constants.h
+++ b/include/multipass/constants.h
@@ -28,6 +28,8 @@ namespace multipass
 constexpr auto client_name = "multipass";
 constexpr auto daemon_name = "multipassd";
 
+constexpr auto snapcraft_remote = "snapcraft";
+
 constexpr auto min_memory_size = "128M";
 constexpr auto min_disk_size = "512M";
 constexpr auto min_cpu_cores = "1";
--- a/include/multipass/vm_image_info.h
+++ b/include/multipass/vm_image_info.h
@@ -40,5 +40,13 @@ public:
     int64_t size;
     bool verify;
 };
+
+inline bool operator==(const VMImageInfo& a, const VMImageInfo& b)
+{
+    return std::tie(a.aliases, a.os, a.release, a.release_title, a.supported, a.image_location, a.kernel_location,
+                    a.initrd_location, a.id, a.stream_location, a.version, a.size, a.verify) ==
+           std::tie(b.aliases, b.os, b.release, b.release_title, b.supported, b.image_location, b.kernel_location,
+                    b.initrd_location, b.id, b.stream_location, b.version, b.size, b.verify);
+}
 }
 #endif // MULTIPASS_VM_IMAGE_INFO_H
--- a/tests/test_custom_image_host.cpp
+++ b/tests/test_custom_image_host.cpp
@@ -49,12 +49,7 @@ constexpr auto sha256_sums =
     "934d52e4251537ee3bd8c500f212ae4c34992447e7d40d94f00bc7c21f72ceb7 *ubuntu-core-16-amd64.img.xz\n"
     "1ffea8a9caf5a4dcba4f73f9144cb4afe1e4fc1987f4ab43bed4c02fad9f087f *ubuntu-core-18-amd64.img.xz\n"
     "52a4606b0b3b28e4cb64e2c2595ef8fdbb4170bfd3596f4e0b84f4d84511b614 *ubuntu-core-20-amd64.img.xz\n"
-    "6378b1fa3db76cdf18c905c8282ebc97401951a9338722486f653dbf16eb7915 *ubuntu-core-22-amd64.img.xz\n"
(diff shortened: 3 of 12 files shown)


include/multipass/constants.h
```diff
@@ -28,6 +28,8 @@ namespace multipass
 constexpr auto client_name = "multipass";
 constexpr auto daemon_name = "multipassd";
 
+constexpr auto snapcraft_remote = "snapcraft";
+
```
georgeliao, line 32: Two doubts here:
1. There other three remotes "release", "daily", "appliance" strings are defined in ubuntu_image_host.h, but "snapcraft" is defined in constants.h, is there a reason behind that? 
2. I know that the codebase has been using auto to catch string literal, which makes the type char*. What about C++17 feature std::string_view, it might result in some client code changes, but it has its upsides, like more public methods and better performance (has length info, no longer a null-terminated which requires traversal for every operation).

include/multipass/vm_image_info.h
```diff
@@ -40,5 +40,13 @@ class VMImageInfo
     int64_t size;
     bool verify;
 };
+
```
georgeliao, line 43: C++20 can do default three way comparison `auto operator<=>(const VMImageInfo& other) const = default;` to save us from the boilerplate thing. However, since we are not there yet, this is the best we can do.

tests/test_custom_image_host.cpp
```diff
@@ -337,14 +289,23 @@ TEST_F(CustomImageHost, info_for_unsupported_remote_throws)
...
     MP_EXPECT_THROW_THAT(host.info_for(make_query("xenial", unsupported_remote)), mp::UnsupportedRemoteException,
                          mpt::match_what(HasSubstr(fmt::format(
                              "Remote \'{}\' is not a supported remote for this platform.", unsupported_remote))));
 }
 
+TEST_F(CustomImageHost, info_for_full_hash_returns_empty_image_info)
+{
+    mp::CustomVMImageHost host{"x86_64", &mock_url_downloader, default_ttl};
+
+    auto info = host.info_for_full_hash("934d52e4251537ee3bd8c500f212ae4c34992447e7d40d94f00bc7c21f72ceb7");
```
georgeliao, line 304: May adding const to `auto info`, auto catch things by value and non-const by default, since` info_for_full_hash` returns by value, then catching by value is fine; but const is something we have to add manually.

georgeliao:
@townsend2010 , lgtm, just left a few minor comments.

include/multipass/constants.h
townsend2010: Hey @georgeliao!

1. I want to start moving those constants to, well, `constants.h`, so since I'm modifying things for the `snapcraft` remote in this PR, this is a good place to start.  I'm reluctant to change the other remotes in this PR since that wouldn't be really part of what this PR is accomplishing.
2. I think this is a good idea in changing these string literals to be `string_view` instead of `auto`, but I don't think this is the PR to do it.  I *could* make this particular change be that though.

include/multipass/vm_image_info.h
townsend2010: Yup, exactly.  And there are a few other classes that use `std::tie()` as well.

tests/test_custom_image_host.cpp
townsend2010: Sure, makes sense and will do that.

include/multipass/constants.h
townsend2010: Ah, wait, yes, I'm going to revert the `string_view` change since it's breaking the other remotes.  I think if we want to change string literal constants to `string_view`, that should be a separate PR with all of the changes necessary in the code.

georgeliao: I think the client code for snapcraft_remote variable has to change accordingly, because char* can be converted to std::string implicitly, but std::string_view can not, because the `explicit std::string(std::string_view)` constructor is explicit. So std::string(snapcraft_remote) needs to be called to fix the compilation error. Anyway, it is a minor thing though, it is also ok to skip this for now.

georgeliao:
bors merge

PR #3075, opened 2023-05-16
[doc] Fix docs regarding exceptions in top_catch_all

ricab:
- [doc] Delete wrong comment in top-catch helper
- [doc] Clarify exception behavior in top-catch-all


townsend2010 (approved):
Ok, makes sense to me.  Thanks!

bors merge

PR #3082, opened 2023-05-19
[snap] Update to qemu 8.0

townsend2010:


luis4a0 (commented):
Hey @townsend2010! The snap code looks good and, since evaluating the diff of a diff is difficult, I just read the file in the repo and tested functionally the patch. Everything looks OK, so I am good to approve this. Just a question before that: why there is no machine for Ubuntu 23.04?

luis4a0:
I also tested the replacement for `virtiofsd`, works good!

townsend2010:
> I also tested the replacement for `virtiofsd`, works good!

Well, there isn't a replacement for it in this PR.  `virtiofsd` is now a separate Rust-based project that will have to be added whenever we support that mount transport type.  So in this PR, it's *only* `virtfs`, ie, 9p.

andrei-toterman (approved):
Alright, looks good to me!

PR #3098, opened 2023-05-25
Create dynamic libraries with C interfaces for Dart FFI

andrei-toterman:
This PR adds a new shared library called `dart_ffi` which contains C interfaces of some of our C++ functions which we'd like to call from Dart using FFI.
Some additional refactoring was needed in order decouple the generation of certificates from the creation of the gRPC channel, so that certificate data can be more easily retrieved using a C interface.

### DIFF
--- a/3rd-party/CMakeLists.txt
+++ b/3rd-party/CMakeLists.txt
@@ -86,6 +86,7 @@ option(YAML_CPP_BUILD_CONTRIB OFF)
 option(YAML_CPP_BUILD_TESTS OFF)
 
 add_subdirectory(yaml-cpp EXCLUDE_FROM_ALL)
+set_target_properties(yaml-cpp PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 add_library(yaml INTERFACE)
 
@@ -121,6 +122,7 @@ set(BUILD_SHARED_LIBS OFF)
 add_subdirectory(semver EXCLUDE_FROM_ALL)
 target_include_directories(semver INTERFACE
   ${CMAKE_CURRENT_SOURCE_DIR}/semver/include)
+set_target_properties(semver PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 # xz-decoder config
 add_subdirectory(xz-decoder EXCLUDE_FROM_ALL)
--- a/src/client/cli/client.cpp
+++ b/src/client/cli/client.cpp
@@ -74,7 +74,7 @@ auto make_handler_unregisterer(mp::SettingsHandler* handler)
 } // namespace
 
 mp::Client::Client(ClientConfig& config)
-    : stub{mp::Rpc::NewStub(mp::client::make_channel(config.server_address, config.cert_provider.get()))},
+    : stub{mp::Rpc::NewStub(mp::client::make_channel(config.server_address, *config.cert_provider))},
       term{config.term},
       aliases{config.term}
 {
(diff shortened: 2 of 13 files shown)


src/client/desktop_gui/ffi/dart_ffi.cpp
```diff
@@ -0,0 +1,70 @@
...
+catch (const std::exception& e)
+{
+    mpl::log(mpl::Level::warning, category, fmt::format("failed retrieving certificate key pair: {}", e.what()));
+    return KeyCertificatePair{nullptr, nullptr};
+}
+catch (...)
+{
+    mpl::log(mpl::Level::warning, category, "failed retrieving certificate key pair");
+    return KeyCertificatePair{nullptr, nullptr};
+}
```
ricab, line 70: Hey @andrei-toterman you can use a git hook to avoid missing this sort of thing: https://paste.ubuntu.com/p/GZ7zyWdNGC/ . Ironically, that paste is missing a newline at the end (ubuntu pastebin apparently trims them).

3rd-party/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ option(YAML_CPP_BUILD_CONTRIB OFF)
 option(YAML_CPP_BUILD_TESTS OFF)
 
 add_subdirectory(yaml-cpp EXCLUDE_FROM_ALL)
+set_target_properties(yaml-cpp PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
luis4a0, line 89: Why do we need PIC here? Anyway, I'm OK with it, especially because it's third party. Just asking out of curiosity.

luis4a0 (approved):
Hey @andrei-toterman! Great job here. I see us using extensively this library in the future, so we'd need to be careful in the design. But until now, I don't see any flaws in design nor in code. So please rebase the base branch from main and make sure this will merge fine into it, answer please the few questions and we are good to go!

src/client/desktop_gui/ffi/dart_ffi.cpp
andrei-toterman: Thanks for the tip!

3rd-party/CMakeLists.txt
andrei-toterman: Without it, I was simply getting a compilation error that said it needs PIC to compile some functions in that library :)

src/client/cli/client.cpp
```diff
@@ -74,7 +74,7 @@ auto make_handler_unregisterer(mp::SettingsHandler* handler)
 } // namespace
 
 mp::Client::Client(ClientConfig& config)
-    : stub{mp::Rpc::NewStub(mp::client::make_channel(config.server_address, config.cert_provider.get()))},
+    : stub{mp::Rpc::NewStub(mp::client::make_channel(config.server_address, *config.cert_provider))},
```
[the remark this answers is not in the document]
andrei-toterman: I don't think so. We need to pass a reference to `make_channel` and `cert_provider` in this case is a `unique_ptr`, so we need `*` to get that reference.

src/client/common/client_common.cpp
```diff
@@ -191,51 +190,9 @@ void mp::client::register_global_settings_handlers()
 }
 
 std::shared_ptr<grpc::Channel> mp::client::make_channel(const std::string& server_address,
-                                                        mp::CertProvider* cert_provider)
+                                                        const mp::CertProvider& cert_provider)
```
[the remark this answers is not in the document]
andrei-toterman: Yeah, we could've kept the pointer but since after this refactoring we never pass a null pointer to this function, we can use a reference instead for more safety

luis4a0:
bors cancel

PR #3099, opened 2023-05-25
[vm] Default auto snapshot answer to yes on restore question

ricab:
Accept empty answer and default to yes on the snapshot question asking whether an automatic snapshot should be taken on restore. 

This is unlike what we do in the bridging question because here we have a natural, safe, default answer, unlike in that other one.


sharder996 (approved):
Looks good. Nice catch!

PR #3100, opened 2023-05-26
[snapshots] Add bash completions for mp snapshot and restore

sharder996:


ricab (commented):
Thanks @sharder996. I am not keen on completing instances in place of snapshots, but other than that this can go in. At some point we should fix offering multiple instances in commands that take only one, as well as repeating unrepeatable options, but that is for another time.

completions/bash/multipass
```diff
@@ -337,6 +343,9 @@ _multipass_complete()
             "recover")
                 _multipass_instances "Deleted"
             ;;
+            "snapshot"|"restore")
```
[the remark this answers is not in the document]
sharder996: This is ready for another review pass, @ricab. There should now be a two-stage auto-completion for `multipass restore` where firstly instances are suggested, and then once an instance has been matched, snapshots for that instance will be suggested.

completions/bash/multipass
```diff
@@ -19,7 +19,7 @@ _multipass_complete()
         local state=$1
 
         local cmd="multipass list --format=csv --no-ipv4"
-        [ -n "$state" ] && cmd="$cmd | \grep -E '$state'"
+        [ -n "$state" ] && cmd="$cmd | \awk -F',' '\$2 == \"$state\"'"
```
ricab, line 22: Good improvement. This is to avoid falling pray to instances called "Running-Rabbit", correct?

Is there any special need for this with snapshots? If not, could you cover it in a drive-by PR (the same as the other one would be fine).

ricab (commented):
Thanks @sharder996, this is very useful. And good job on the snapshot completion!

I am only uncertain about the 2-step approach. Why did you opt for it? It has the downside that you can't reuse it for `info` or `delete`, which accept both instances and snapshots. Why not bag it all from the get go? Let the user decide whether to add the period when both are acceptable. 

You could write a sibling to `_multipass_instances` called `_multipass_instances_and_snapshots`, also receiving and filtering by state, and then employ that in multiple commands.

ricab:
Hey @sharder996, from a first look at the code, I thought this was ready for another pass, but looking deeper it doesn't appear so. Can you confirm whether or not this is ready for another pass? At a first look I thought so, but looking a little deeper, you only rebased, correct?

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
+        if [[ $1 == "instances" ]]; then
+            backup=$instances
+            instances=""
+        elif [[ $1 == "snapshots" ]]; then
+            backup=$snapshots
+            snapshots=""
+        fi
 
         local found
+        _get_comp_words_by_ref -n := -w WORDS -i CWORD cur prev
```
ricab, line 66: I don't think this needs the colon and the equals. That incantation tells it not to consider those two characters as word splitters, for cases like instance:dir or key=val. Not really applicable here, I don't think.

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
+        if [[ "${cur}" =~ '.' ]]; then
+            _multipass_snapshots
+            _multipass_filter_array "snapshots"
+        fi
+
+        _multipass_instances
+        _multipass_filter_array "instances"
+    }
+
+    _multipass_filter_array()
```
ricab, line 53: What is the purpose of this function? Is this meant only to remove repetitions? If so, isn't that now being handled generically by #3156 ?

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
+                if [[ $1 == "instances" ]]; then
+                    instances="${instances} ${item}"
+                elif [[ $1 == "snapshots" ]]; then
+                    snapshots="${snapshots} ${item}"
+                fi
+            fi
+        done
+    }
 
+    _multipass_filter_snapshots()
```
ricab, line 85: How is this different from `_multipass_filter_array snapshots`?

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
+    {
+        _multipass_instances $1
+        _multipass_filter_array "instances"
+    }
+
+    _multipass_restorable_instances()
+    {
+        local retval=$( multipass info --all --no-runtime-information --format=csv \
+                           | \awk -F'("[^"]*")?,|\r?\n' '$2 == "Stopped" && $16 > 0' \
+                           | \grep -Ev '(\+--|Name)' \
```
ricab, line 115: No point filtering the title line after filtering stopped entries, so it seems to me that this should be the other way around. Again, better follow #3243 to filter the title line.

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
+    _multipass_instances_unique()
+    {
+        _multipass_instances $1
+        _multipass_filter_array "instances"
+    }
+
+    _multipass_restorable_instances()
+    {
+        local retval=$( multipass info --all --no-runtime-information --format=csv \
+                           | \awk -F'("[^"]*")?,|\r?\n' '$2 == "Stopped" && $16 > 0' \
```
ricab, line 114: Why do you need such a complex field separator? Things in between quotes are entries, not separators themselves, right? Are you trying to gobble quoted commas? But when would they come about? Doesn't just `-F',|\r?\n'` work?

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
         local cmd="multipass list --format=csv --no-ipv4"
         [ -n "$state" ] && cmd="$cmd | \awk -F',' '\$2 == \"$state\"'"
 
-        local instances=$( \eval $cmd | \grep -Ev '(\+--|Name)' | \cut -d',' -f 1 )
+        local retval=$( \eval $cmd | \grep -Ev '(\+--|Name)' | \cut -d',' -f 1 )
+
+        for instance in $retval; do
+            instances="${instances} ${instance}"
+        done
```
ricab, line 28: This is only to replace newlines with spaces, right? Wouldn't this work?

```
instances=${retval//?($'\r')$'\n'/ }
```

In any case, it is repeated in different places, so I would ask to extract into a function and call that (e.g. `_flatten_lines`).

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
                 fi
             done
             if [ ${found} == 0 ]; then
-                opts="${opts} ${instance}"
+                snapshots="${snapshots} ${snapshot}"
             fi
         done
     }
 
+    _multipass_instances_unique()
```
ricab, line 105: idem: filtering duplicates.

completions/bash/multipass
```diff
@@ -320,22 +408,35 @@ _multipass_complete()
...
+            ;;
+            "snapshot")
+                _multipass_instances_unique "Stopped"
+            ;;
+            "restore")
+                if [[ "${cur}" =~ '.' ]]; then
+                    _multipass_snapshots
+                else
+                    _multipass_restorable_instances
+                fi
```
ricab, line 439: Again, I don't understand the two-step approach. Is this motivated by `_get_comp_words_by_ref` splitting on dots? You can avoid that by passing `-n '.'`. See `/usr/share/bash-completion/bash_completion`.

completions/bash/multipass
```diff
@@ -21,25 +21,105 @@ _multipass_complete()
...
         done
     }
 
+    _multipass_instances_unique()
+    {
+        _multipass_instances $1
+        _multipass_filter_array "instances"
+    }
+
+    _multipass_restorable_instances()
```
ricab, line 111: If you drop the 2 step approach and instead just add all snapshots to the bag, you can just use `list --snapshots` and filter snapshots of stopped instances.

ricab (changes requested):
Hey @sharder996, thanks for splitting the code into separate units, I like that there are functions to complete snapshots, instances, and both. However, there are still a few things I don't understand. The most relevant is probably the 2-step approach of completing instance then snapshot. Could you explain a little why you went with that? The usual approach is to start with the full set of possibilities and prune it down as the user types.

completions/bash/multipass
sharder996: Yes, I found a way to generalize it and the same code is also being used there. What would be the best way to get the code here? Merge the other PR and rebase on main?

ricab: Yes, I think that is the best. To unblock this one you can create a temporary base branch out of snapshots that merges the unduplicating branch in, then rebase this on top of that, and turn it into a draft, with a note that it needs the other one to go in. That would allow you to deal with any conflicts early and otherwise progress here. Once the other one is merged, we'd rebase snapshots on main, rebase this on snapshots, and delete the temporary branch.

completions/bash/multipass
sharder996: The reason I used the 2 step approach was to avoid polluting the cli with an abundance of completions. For example, just by entering `multipass restore<TAB><TAB>` the script would otherwise give you all instances and all their snapshots as separate suggestions.

ricab: OK, my first inclination is to regard that as just the way tab-completion works. On further consideration, though, I recognize it is meant to be hierarchical too (as in completion of paths). So OK, as long as it doesn't complicate the code too much and instance/snapshot mixes work well, I am fine with that.

completions/bash/multipass
sharder996: Oh, I think this might have been because at one point I was expecting nested commas within a single field (eg. multiple ip addresses). Having `--no-runtime-information` makes this a non-issue though.

ricab (changes requested):
Hey @sharder996, I just tried this but it doesn't appear to be working yet. Here are some of the problems I found:

- some suggestions are missing for `list` (see inline).
- `multipass snapshot` suggests multiple instances, although only one is accepted:
```
$ multipass snapshot a [TAB][TAB]
b          --comment  --help     --name     --verbose
```
- `--comment` and `--name` keep being suggested. E.g.:
```
$ multipass snapshot a --comment [TAB][TAB]
b                --comment        fortuitous-rudd  --help           --name           --verbose
$ multipass snapshot a --comment "a comment" [TAB][TAB]
b                --comment        fortuitous-rudd  --help           --name           --verbose
```
- `multipass restore` auto-completes separate instances:
```
$ multipass restore for[TAB]
$ multipass restore fortuitous-rudd [CURSOR]
```
If I hit enter at that point, the command will fail. It should have auto-completed to `fortuitous-rudd.` if the instance had multiple snapshots, or to `fortuitous-rudd.lone-snapshot` if the instance had only that snapshot.
- snapshots not auto-completed by `restore`:
```
$ multipass list --snapshots  | grep fortui
fortuitous-rudd   s1          --          asdf
fortuitous-rudd   s2          s1          asdf
fortuitous-rudd   s3          s2          asdf
$ multipass restore fortuitous-rudd.[TAB][TAB]
[nothing]
$ multipass restore fortuitous-rudd.s[TAB][TAB]
[nothing]
```

I still think the 2-pass approach is asking for trouble... I haven't tried delete/info yet.

ricab (changes requested):
Getting closer! I found only a couple of misses:

##### Case 1

```
$ multipass info [TAB][TAB]
enormous-brambling  --format            --help              Name                --verbose
$ multipass info N[TAB]
$ multipass info Name [CURSOR]
```

"Name" isn't a valid suggestion. Same for `delete`.

##### Case 2

```
$ multipass list --snapshots
Instance             Snapshot   Parent   Comment
enormous-brambling   asdf       --       --
$ multipass info e[TAB]
$ multipass info enormous-brambling [CURSOR]
```

I would expect it to suggest the snapshot as well. Same for delete.

sharder996:
Hey @ricab, thanks for your patience with this one!

I fixed the first issue you mentioned. As for the second one, that is an artifact of having snapshots only suggested when you append a period to the end of an instance name. Also, since we filter out duplicate instances, if you enter an instance name for `info` or `delete`, that name will not be suggested again even though you could query info on its snapshots. I can change it so that it does the same thing as `restore` and suggest all instances and snapshots. I thought that would make the suggestions more cluttered and had wanted to go with something a bit more intelligent, but maybe bash completion isn't the place for unnecessary sophistication :sweat_smile:

ricab (approved):
Excellent! The only issue I found was that multiple-instance suggestion in single-instance commands. However, as you point out, that is a pre-existing generic issue that we can leave out of this PR.

PR #3105, opened 2023-05-31
[snapshots] QEMU-specific snapshot restoring

ricab:


sharder996 (approved):
LGTM!

Tested on both Mac and Linux

PR #3108, opened 2023-06-05
Upgrade deprecated gmock macro MOCK_METHOD<num> and MOCK_CONST_METHOD<num>

georgeliao:
---
**Describe the problem:**
If we take a close look at the two conversion examples: 
```
  MOCK_METHOD2(all_images_for, std::vector<VMImageInfo>(const std::string&, const bool));
->MOCK_METHOD(std::vector<VMImageInfo>, all_images_for, (const std::string&, const bool), (override));
```
```
  MOCK_CONST_METHOD2(make_firewall_config, FirewallConfig::UPtr(const QString&, const std::string&));
->MOCK_METHOD(FirewallConfig::UPtr, make_firewall_config, (const QString&, const std::string&), (const, override));
```
What essentially happened is that some substrings of the text moved, and some were replaced by other texts. So this is a typical regex search/replace problem. 
 
**Solve the problem:**
If we dissect the first example, what we want to do is to capture the substring `all_images_for`, `std::vector<VMImageInfo>` and `(const std::string&, const bool)` and re-position them. The regex feature capture group is perfect for this task. A capture group is part of the pattern enclosed in parentheses '()', it groups multiple tokens together and creates a group for extracting a substring and being used as a backreference. 

With a little bit of digging, we figured out the search and replace regex can be:
`MOCK_METHOD\d\((\w+), ([\w:<>&*, ]+)(\(.*\))\);` to `MOCK_METHOD(\2, \1, \3, (override));`
Here, if we applied the regex search to the first example, we will get `(\w+)` matches `all_images_for`, `([\w:<>&*, ]+)` matches `std::vector<VMImageInfo>` and `(\(.*\))\)` matches `(const std::string&, const bool)` (for more detailed dissection of the regex, please goto https://regexr.com/ and hover the mouse on the filled-in regex). These substrings have backreference numbers 1, 2,3 and will be re-positioned in the replacement regex. 

Similarly, `MOCK_CONST_METHOD` can also be replaced by regex: 
`MOCK_CONST_METHOD\d\((\w+), ([\w:<>&*, ]+)(\(.*\))\);` to `MOCK_METHOD(\2, \1, \3, (const, override));`

**A few minor issues were encountered during the execution:**

1. We have no `override` information from the original text, so it is impossible to know if the new one should be `override` or not, the approach we have taken is to mark everything `override` and manually fix the ones that did not compile, fortunately, there is only a handful of them. 
2. The new macro `MOCK_METHOD` has a small flaw, that is it uses commas as a delimiter to separate the arguments. What that means is the data type containing commas will be mistaken as two parts which cause the compilation to fail, adding parenthesis can fix that. One example is MOCK_METHOD(**(**std::vector<std::pair<std::string, VMImageInfo>>**)**, all_info_for, (const Query&), (override)); this was done manually. 
3. There are cases where it is multiple lines of text because of the line break. In these cases, regex can not match that, I also did not have a good idea to adjust regex since the line breakpoint can be different depending on where clang-format breaks it. This was done manually as well. One workaround can be changing clang-format `ColumnLimit` to a big number and formatting the whole project first, then applying the regex replacement and formatting the whole project back after that.

### DIFF
--- a/tests/mock_platform.h
+++ b/tests/mock_platform.h
@@ -35,7 +35,7 @@ public:
             .WillRepeatedly(testing::Return());
     };
 
-    MOCK_METHOD((std::map<std::string, NetworkInterfaceInfo)>, get_network_interfaces_info, (), (const, override));
+    MOCK_METHOD((std::map < std::string, NetworkInterfaceInfo) >, get_network_interfaces_info, (), (const, override));
     MOCK_METHOD(QString, get_blueprints_url_override, (), (const, override));
     MOCK_METHOD(bool, is_remote_supported, (const std::string&), (const, override));
     MOCK_METHOD(bool, is_backend_supported, (const QString&), (const, override));
--- a/tests/mock_qsettings.h
+++ b/tests/mock_qsettings.h
@@ -31,7 +31,8 @@ public:
     using WrappedQSettings::WrappedQSettings; // promote visibility
     MOCK_METHOD(QSettings::Status, status, (), (const, override));
     MOCK_METHOD(QString, fileName, (), (const, override));
-    MOCK_METHOD(QVariant, value_impl, (const QString& key, const QVariant& default_value), (const, override)); // promote visibility
+    MOCK_METHOD(QVariant, value_impl, (const QString& key, const QVariant& default_value),
+                (const, override)); // promote visibility
     MOCK_METHOD(void, setIniCodec, (const char* codec_name), (override));
     MOCK_METHOD(void, sync, (), (override));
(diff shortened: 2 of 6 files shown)


townsend2010:
Hey @georgeliao!

Could you please rebase on latest `main`?  Thanks!

townsend2010 (approved):
Hey @georgeliao!

Thanks for doing this!  LGTM!

bors merge

townsend2010:
Going to run the failed Windows integration tests again and if they pass, I'll merge this manually.

townsend2010:
Ok, the Windows integration tests passed, so will merge this now.

PR #3089, opened 2023-05-22
[qemu] Fix crash on concurrent shutdown

ricab:

PR #3114, opened 2023-06-06
Improve waiting for SSH

ricab:
Two main changes to improve the step of waiting for SSH, when starting/launching instances:

1. Increase the time allotted for each iteration of the loop to wait for SSH, to 1 second. Log each such iteration of the loop.
2. Retry only on specific exceptions, logging them (with `trace` verbosity, since this happens every second), and letting the rest through, in the hope that this may shed some light into cases where instances end up in unknown state.

### DIFF
--- a/include/multipass/exceptions/internal_timeout_exception.h
+++ b/include/multipass/exceptions/internal_timeout_exception.h
@@ -1,0 +1,28 @@
+//
+// Created by ricab on 06-06-2023.
+//
+
+#ifndef MULTIPASS_INTERNAL_TIMEOUT_EXCEPTION_H
+#define MULTIPASS_INTERNAL_TIMEOUT_EXCEPTION_H
+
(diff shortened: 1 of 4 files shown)


townsend2010 (approved):
:+1: thanks!

bors merge

townsend2010:
Hmm, there does seem to be an issue with this on Intel Macs: https://github.com/canonical/multipass/actions/runs/5398548958/jobs/9804491608#step:3:300

I tried running the integration test suite again for this PR and it still fails with the same issue.  The [previous PR](https://github.com/canonical/multipass/pull/3089) did not have this issue, so it seems it's related to this change.

ricab:
More likely it's a macOS specific exception that needs dedicated handling for retrying.

townsend2010 (approved):
Ok, works now.  Thanks!

bors merge

PR #3121, opened 2023-06-13
[snapshots] Fail preemptively before prompting user

sharder996:


src/daemon/daemon.cpp
```diff
@@ -2518,16 +2518,40 @@ try
...
+        {
+            return status_promise->set_value(
+                grpc::Status{grpc::NOT_FOUND, fmt::format("snapshot \"{}\" does not exist", request->snapshot())});
+        }
+
         const auto& vm_dir = instance_directory(instance_name, *config);
         if (!request->destructive())
         {
-            reply_msg(server, fmt::format("Taking snapshot before restoring {}", instance_name));
+            RestoreReply reply{};
```
ricab, line 2535: This shadows the other `reply` var.

src/daemon/daemon.cpp
```diff
@@ -2518,16 +2518,40 @@ try
         auto spec_it = vm_instance_specs.find(instance_name);
         assert(spec_it != vm_instance_specs.end() && "missing instance specs");
 
+        // TODO@snapshots need a way to query the existence of a snapshot before consuming it
+        try
+        {
+            auto snapshot = vm_ptr->get_snapshot(request->snapshot());
```
ricab, line 2524: You don't need this snapshot variable anywhere.

```suggestion
            vm_ptr->get_snapshot(request->snapshot());
```

src/client/cli/cmd/restore.cpp
```diff
@@ -53,8 +53,35 @@ mp::ReturnCode cmd::Restore::run(mp::ArgParser* parser)
...
+        if (const auto& msg = reply.reply_message(); !msg.empty())
+        {
+            spinner.stop();
+            spinner.start(msg);
+        }
+
+        if (reply.confirm_destruction())
+        {
+            spinner.stop();
+            RestoreRequest request;
```
ricab, line 71: Another shadow, this time for the field with the same name.

src/client/cli/cmd/restore.cpp
```diff
@@ -53,8 +53,35 @@ mp::ReturnCode cmd::Restore::run(mp::ArgParser* parser)
...
+
+            if (term->is_live())
+                request.set_destructive(confirm_destruction(reply.instance()));
```
ricab, line 74: If you rename the local request, you can get the instance from `this`, e.g.:

```cpp
                repeat_request.set_destructive(confirm_destruction(request.instance()));
```

src/rpc/multipass.proto
```diff
@@ -512,4 +512,6 @@ message RestoreRequest {
 message RestoreReply {
     string log_line = 1;
     string reply_message = 2;
+    string instance = 3;
```
ricab, line 515: I don't think you need this. See comment in restore.cpp.

src/client/cli/cmd/restore.cpp
```diff
@@ -104,40 +131,19 @@ mp::ParseCode cmd::Restore::parse_args(mp::ArgParser* parser)
     if (tokens.size() != 2 || tokens[0].isEmpty() || tokens[1].isEmpty())
     {
         cerr << "Invalid format. Please specify the instance to restore and snapshot to use in the form "
-                "<instance>.<spanshot>.\n";
```
ricab, line 107: heh

ricab (commented):
Thanks @sharder996, this is getting there. Only a few rough edges left.

src/daemon/daemon.cpp
```diff
@@ -2518,23 +2518,13 @@ try
...
-        try
-        {
-            auto snapshot = vm_ptr->get_snapshot(request->snapshot());
-        }
-        catch (const std::out_of_range&)
-        {
-            return status_promise->set_value(
-                grpc::Status{grpc::NOT_FOUND, fmt::format("snapshot \"{}\" does not exist", request->snapshot())});
-        }
+        vm_ptr->get_snapshot(request->snapshot());
```
ricab, line 2521: You removed the exception handing? I thought `NOT_FOUND` was appropriate. You can also place it together with the other catch at the end of the function. I would also add a comment to clarify why this is here.

src/client/cli/cmd/restore.cpp
```diff
@@ -65,18 +65,18 @@ mp::ReturnCode cmd::Restore::run(mp::ArgParser* parser)
...
-        if (reply.confirm_destruction())
+        if (reply.confirm_destructive())
         {
             spinner.stop();
-            RestoreRequest request;
+            RestoreRequest client_response;
 
             if (term->is_live())
-                request.set_destructive(confirm_destruction(reply.instance()));
+                client_response.set_destructive(confirm_destruction(this->request.instance()));
```
ricab, line 74: I don't think you need `this->` now that the names don't collide. But fine.

ricab (commented):
Thank you for addressing all of that so quickly @sharder996! One thing left to verify.

ricab (changes requested):
Hey @sharder996, I was going to say only conflicts missing, but I just realized `multipass restore` now hangs.

sharder996:
@ricab Are you sure you had the most recent changes? I noticed the command hanging as well, but had fixed it. Let me know if its still an issue for you, though.

ricab (approved):
Yeah, that's it! The snap is working for me now :slightly_smiling_face:

PR #3124, opened 2023-06-13
Add missing pieces for QEMU snapshot restoring

ricab:


sharder996:
@ricab Found an issue with mounts. If I create a mount on an instance and then revert back to a previous snapshot before creating the mount, it still persists in the instance despite not being reported in `multipass info` and causes required packages to be installed again after restoring the snapshot.

sharder996:
And then, after restarting the Multipass daemon and restoring a snapshot that had mounts, `multipass info` will report mounts are defined, but they will be missing in the instance.

ricab:
OK, thanks for catching these. I think I know what the issue is.

ricab:
Hey @sharder996, I believe this is finally ready for another pass.

sharder996 (approved):
This looks good now! I'm unable to break mounts in any way now.

PR #3126, opened 2023-06-14
[daemon] Fix mapping of instance/snapshot args

ricab:
Needs #3092

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1161,10 +1161,10 @@ bool is_ipv4_valid(const std::string& ipv4)
 }
 
 using InstanceSnapshotPairs = google::protobuf::RepeatedPtrField<mp::InstanceSnapshotPair>;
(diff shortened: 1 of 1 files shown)


src/daemon/daemon.cpp
```diff
@@ -1168,7 +1168,7 @@ std::unordered_map<std::string, std::unordered_set<std::string>> map_snapshots_t
     for (const auto& it : instances)
     {
         if (it.snapshot_name().empty())
-            instance_snapshots_map[it.instance_name()];
+            instance_snapshots_map[it.instance_name()].clear();
```
ricab, line 1171: Using clear to make it obvious that we're removing anything that is there already.

ricab:
I did this on top of your delete branch to avoid having to deal with conflicts there.

sharder996 (approved):
LGTM!

ricab:
Thanks for the review @sharder996. Did a clean rebase and merging.

PR #3128, opened 2023-06-16
Revert "[build] Use top commit when determing rc and dev versions"

townsend2010:
This reverts commit b9a097f6a5a959cf9df59e8ded78b2b44fcf59a3.

---

Seems the change I made before broke the current `dev` build and we get the wrong version for `edge`.  Let's revert this for now and re-evaluate if the `rc` builds get messed up again.


townsend2010:
bors retry

Issue #549, opened 2018-12-19
[delete] suspend and resume running instances when deleting


Saviq:
We're currently shutting down instances to be deleted, even if they're running. We could instead suspend them so that when recovering, the instance is back in the state it was deleted at, ensuring no data loss.

Whether to resume on `multipass recover` I'm not sure yet.

levkropp:
One important thing to look into with regards to this is if suspending an instance is faster than shutting it down, which is something that needs to be verified.
#138 is another related issue that discusses how `multipass delete` should behave with a suggestion for a `--wait` parameter to allow for a "slow" delete of an instance.

ricab:
I don't think we regard the instance trash (vs straight delete) a very valuable feature these days. It is not present in the GUI and we have contemplated removing it entirely. We could choose to keep it and implement this, but I am lowering priority.

PR #3130, opened 2023-06-19
[snapshots] resolve instance dir in vm

sharder996:
The motivation behind this change is to allow a vm to resolve the directory in which it and all it's relevant files are stored, which is needed for snapshots and retrospectively, native mounts on hyperv instances.

In order to achieve this, the idea was to take away the responsibility of creating/deleting instance directories from the vm image vault and give it to the vm factory. This conceptually makes more sense as the factory is the class responsible for creating/removing vm resources in the first place and is also the place where platform/backend specific directories locations are resolved. The vm image vault then queries the factory for a instance specific directories and using it as a download directory.

As this is not a small architectural change, I want to get some feedback on this and improve it.

Public side of [#549](https://github.com/canonical/multipass-private/pull/549)

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -154,6 +154,8 @@ public:
     // virtual machine helpers
     virtual void wait_for_cloud_init(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
                                      const SSHKeyProvider& key_provider) const;
+    virtual Path derive_instances_dir(const Path& data_dir, const Path& backend_directory_name,
+                                      const Path& instances_subdir) const;
 
     // system info helpers
     virtual std::string get_kernel_version() const;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -101,13 +101,14 @@ public:
 
     VirtualMachine::State state;
     const std::string vm_name;
-    const QDir instance_dir;
     std::condition_variable state_wait;
     std::mutex state_mutex;
     std::optional<IPAddress> management_ip;
     bool shutdown_while_starting{false};
 
 protected:
+    const QDir instance_dir;
+
     VirtualMachine(VirtualMachine::State state, const std::string& vm_name, const Path& instance_dir)
         : state{state}, vm_name{vm_name}, instance_dir{QDir{instance_dir}} {};
     VirtualMachine(const std::string& vm_name, const Path& instance_dir)
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -92,11 +92,11 @@ public:
     virtual int get_num_snapshots() const noexcept = 0;
     virtual std::shared_ptr<const Snapshot> get_snapshot(const std::string& name) const = 0;
     virtual std::shared_ptr<Snapshot> get_snapshot(const std::string& name) = 0;
-    virtual std::shared_ptr<const Snapshot> take_snapshot(const QDir& snapshot_dir, const VMSpecs& specs,
-                                                          const std::string& name, const std::string& comment) = 0;
-    virtual void delete_snapshot(const QDir& snapshot_dir, const std::string& name) = 0;
-    virtual void restore_snapshot(const QDir& snapshot_dir, const std::string& name, VMSpecs& specs) = 0;
-    virtual void load_snapshots(const QDir& snapshot_dir) = 0;
+    virtual std::shared_ptr<const Snapshot> take_snapshot(const VMSpecs& specs, const std::string& name,
+                                                          const std::string& comment) = 0;
+    virtual void delete_snapshot(const std::string& name) = 0;
+    virtual void restore_snapshot(const std::string& name, VMSpecs& specs) = 0;
+    virtual void load_snapshots() = 0;
     virtual std::vector<std::string> get_childrens_names(const Snapshot* parent) const = 0;
 
     VirtualMachine::State state;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -113,7 +113,6 @@ protected:
         : state{state}, vm_name{vm_name}, instance_dir{QDir{instance_dir}} {};
     VirtualMachine(const std::string& vm_name, const Path& instance_dir)
         : VirtualMachine(State::off, vm_name, instance_dir){};
-    VirtualMachine(const std::string& vm_name) : VirtualMachine(vm_name, ""){};
 };
 } // namespace multipass
 #endif // MULTIPASS_VIRTUAL_MACHINE_H
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -20,6 +20,7 @@
 
 #include "disabled_copy_move.h"
 #include "ip_address.h"
+#include "path.h"
 
 #include <QDir>
 #include <QJsonObject>
@@ -100,14 +101,18 @@ public:
 
     VirtualMachine::State state;
     const std::string vm_name;
+    const QDir instance_dir;
     std::condition_variable state_wait;
     std::mutex state_mutex;
     std::optional<IPAddress> management_ip;
     bool shutdown_while_starting{false};
 
 protected:
-    VirtualMachine(VirtualMachine::State state, const std::string& vm_name) : state{state}, vm_name{vm_name} {};
-    VirtualMachine(const std::string& vm_name) : VirtualMachine(State::off, vm_name){};
+    VirtualMachine(VirtualMachine::State state, const std::string& vm_name, const Path& instance_dir)
+        : state{state}, vm_name{vm_name}, instance_dir{QDir{instance_dir}} {};
+    VirtualMachine(const std::string& vm_name, const Path& instance_dir)
+        : VirtualMachine(State::off, vm_name, instance_dir){};
+    VirtualMachine(const std::string& vm_name) : VirtualMachine(vm_name, ""){};
 };
 } // namespace multipass
 #endif // MULTIPASS_VIRTUAL_MACHINE_H
--- a/include/multipass/virtual_machine_factory.h
+++ b/include/multipass/virtual_machine_factory.h
@@ -60,6 +60,7 @@ public:
     virtual void prepare_instance_image(const VMImage& instance_image, const VirtualMachineDescription& desc) = 0;
     virtual void hypervisor_health_check() = 0;
     virtual QString get_backend_directory_name() const = 0;
+    virtual QString get_instance_directory_name(const std::string& name) const = 0;
     virtual QString get_backend_version_string() const = 0;
     virtual VMImageVault::UPtr create_image_vault(std::vector<VMImageHost*> image_hosts, URLDownloader* downloader,
                                                   const Path& cache_dir_path, const Path& data_dir_path,
--- a/include/multipass/virtual_machine_factory.h
+++ b/include/multipass/virtual_machine_factory.h
@@ -59,8 +59,8 @@ public:
     virtual VMImage prepare_source_image(const VMImage& source_image) = 0;
     virtual void prepare_instance_image(const VMImage& instance_image, const VirtualMachineDescription& desc) = 0;
     virtual void hypervisor_health_check() = 0;
-    virtual QString get_backend_directory_name() = 0;
-    virtual QString get_backend_version_string() = 0;
+    virtual QString get_backend_directory_name() const = 0;
+    virtual QString get_backend_version_string() const = 0;
     virtual VMImageVault::UPtr create_image_vault(std::vector<VMImageHost*> image_hosts, URLDownloader* downloader,
                                                   const Path& cache_dir_path, const Path& data_dir_path,
                                                   const days& days_to_expire) = 0;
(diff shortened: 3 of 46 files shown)


ricab:
Hey @sharder996, it looks like this needs a rebase and conflict fix.

ricab:
Hey @sharder996, I pushed a new commit to account for new tests that are now present after rebasing. However, there are still a few other tests that failed when I tried them locally:

- ImageVault
    * returned_image_contains_instance_name
    * caches_prepared_images
    * remembers_prepared_images
- QemuBackend
    * get_backend_directory_name_calls_qemu_platform

 I am not sure if they broke with my rebase or if they were already failing. Would you be able to take a look?

sharder996:
Ho @ricab, thanks for fixing those tests! When I created this PR, I had not prioritized fixing tests since I'm anticipating there being review changes which would require me to fix tests again.

ricab:
Ah OK, if it's not caused by my rebase I am happy.

ricab (changes requested):
Alright, thanks @sharder996! It took me a little to wrap my head around this one, but I think I finally formed a cohesive opinion... 

This is tricky stuff but you made good progress here. I understand the option for creating the directory where it is derived (in `get_instance_directory_name()`), but I am not sure it's worth all the additional trips to disk (even to remove resources, the code tries to create the directory before removing it). We've been operating without that safety net for a long time and if we failed to create the directory now, that would be quickly noticed.

IIUC, the instance directory needs to be created only before it is used, i.e.:

  1. when fetching, just before placing the instance image there
  2. when launching with LXD (which does not place the image there), just before calling the instance constructor
  3. when starting up the daemon with LXD, on creation of pre-existing (LXD) VMs that still do not have a directory, just before calling the instance constructor (only needed the first time after upgrading)

Additionally, I don't see a need to create parent directories separately:
  - `.../vault` is the vault's business
  - `.../vault/instances` is not needed until there is an instance
  - creating `.../vault/instances/<instance>` with `MP_UTILS.make_dir` creates those parents

Therefore, it seems to me that, strictly speaking, the only new place we need to create a directory is in `LXDVMFactory::create_vm()`. Other backends already require the image to be in place when creating the VM, whichever way it got there (`DefaultVMImageVault` today, clone tomorrow, etc.). So there is no need to create the directory in those cases.

If this all makes sense to you, I think you could still keep the instance directory derivation, parameters, and fields where you added them, but actually create the directory only in `DefaultVMImageVault`, as before (`image_instance_from` and `extract_image_from`) and in the LXD factory.

WDYT?

ricab:

src/platform/backends/shared/base_virtual_machine_factory.h
```diff
@@ -86,8 +86,17 @@ class BaseVirtualMachineFactory : public VirtualMachineFactory
     virtual void prepare_interface(NetworkInterface& net, std::vector<NetworkInterfaceInfo>& host_nets,
                                    const std::string& bridge_type);
 
+    virtual void remove_resources_for_impl(const std::string& name) = 0;
```
ricab, line 89: That's it, you got it! Nice use of the template method pattern :wink:

include/multipass/virtual_machine.h
```diff
@@ -91,22 +92,26 @@ class VirtualMachine : private DisabledCopyMove
...
     bool shutdown_while_starting{false};
 
 protected:
-    VirtualMachine(VirtualMachine::State state, const std::string& vm_name) : state{state}, vm_name{vm_name} {};
-    VirtualMachine(const std::string& vm_name) : VirtualMachine(State::off, vm_name){};
+    VirtualMachine(VirtualMachine::State state, const std::string& vm_name, const Path& instance_dir)
+        : state{state}, vm_name{vm_name}, instance_dir{QDir{instance_dir}} {};
+    VirtualMachine(const std::string& vm_name, const Path& instance_dir)
+        : VirtualMachine(State::off, vm_name, instance_dir){};
+    VirtualMachine(const std::string& vm_name) : VirtualMachine(vm_name, ""){};
```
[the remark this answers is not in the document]
sharder996: This is here because of constructors in [`stub_virtual_machine.h`](https://github.com/canonical/multipass/pull/3130/files#diff-9083bcbd10dd40a4be12c75ca0c1035766f76711baa70cc4d4f276f4a1f26756L31-L37). The constructors in [`virtual_machine.h`](https://github.com/canonical/multipass/pull/3130/files#diff-d847b18fbc46d755306da3eb6cac52de71043c1cc87ff345cdf8b034ad141ffaL109-L110) are unchanged apart from the additional parameter for `instance_dir` making it a bit unsightly. Any ideas of how to make this look nicer?

src/platform/backends/shared/base_virtual_machine_factory.h
```diff
@@ -34,17 +35,23 @@ class BaseVirtualMachineFactory : public VirtualMachineFactory
 {
 public:
     BaseVirtualMachineFactory() = default;
```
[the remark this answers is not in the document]
sharder996: Are these here because of unit tests? I didn't touch them as I'm unfamiliar with how they work or why exactly they're required...

include/multipass/virtual_machine.h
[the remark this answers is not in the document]
ricab: The onus of providing a stub directory should be on the stub VM. instead of adding this constructor, the stub VM can call one of the other constructors, passing in a fake `instance_dir`. To be sure that tests don't mess around with disk in unproper places, the best would be to create a temporary directory (e.g. with `QTemporaryDir`).

src/daemon/default_vm_image_vault.cpp
```diff
@@ -241,7 +253,8 @@ mp::DefaultVMImageVault::~DefaultVMImageVault()
 
 mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type, const Query& query,
                                                  const PrepareAction& prepare, const ProgressMonitor& monitor,
-                                                 const bool unlock, const std::optional<std::string>& checksum)
+                                                 const bool unlock, const std::optional<std::string>& checksum,
+                                                 const mp::Path& download_dir)
```
ricab, line 257: This needs updating too.

tests/blueprint_test_lambdas.cpp
```diff
@@ -29,17 +29,19 @@
...
 std::function<mp::VMImage(const mp::FetchType&, const mp::Query&, const mp::VMImageVault::PrepareAction&,
-                          const mp::ProgressMonitor&, const bool, const std::optional<std::string>)>
+                          const mp::ProgressMonitor&, const bool, const std::optional<std::string>, const mp::Path&)>
 mpt::fetch_image_lambda(const std::string& release, const std::string& remote, const bool must_have_checksum)
 {
     return [&release, &remote, must_have_checksum](
                const mp::FetchType& fetch_type, const mp::Query& query, const mp::VMImageVault::PrepareAction& prepare,
-               const mp::ProgressMonitor& monitor, const bool unlock, const std::optional<std::string>& checksum) {
+               const mp::ProgressMonitor& monitor, const bool unlock, const std::optional<std::string>& checksum,
+               const mp::Path& download_dir) {
```
ricab, line 44: idem: save_dir

tests/lxd/test_lxd_image_vault.cpp
```diff
@@ -70,6 +70,7 @@ struct LXDImageVault : public Test
     mp::Query default_query{instance_name, "xenial", false, "", mp::Query::Type::Alias};
     mpt::StubURLDownloader stub_url_downloader;
     mpt::TempDir cache_dir;
+    mpt::TempDir download_dir;
```
ricab, line 73: idem: save_dir

tests/test_image_vault.cpp
```diff
@@ -168,16 +168,18 @@ struct ImageVault : public testing::Test
         [](const mp::VMImage& source_image) -> mp::VMImage { return source_image; }};
     mpt::TempDir cache_dir;
     mpt::TempDir data_dir;
+    mpt::TempDir download_dir;
```
ricab, line 171: idem: save_dir

include/multipass/vm_image_vault.h
```diff
@@ -82,7 +82,7 @@ class VMImageVault : private DisabledCopyMove
     virtual ~VMImageVault() = default;
     virtual VMImage fetch_image(const FetchType& fetch_type, const Query& query, const PrepareAction& prepare,
                                 const ProgressMonitor& monitor, const bool unlock,
-                                const std::optional<std::string>& checksum) = 0;
+                                const std::optional<std::string>& checksum, const Path& download_dir) = 0;
```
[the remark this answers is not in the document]
ricab: Thanks for changing this, but there are a few cases where the old name is still used.

src/platform/backends/shared/base_virtual_machine_factory.h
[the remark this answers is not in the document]
ricab: OK, this is another case like the previous one. Constructors are meant to ensure objects are properly initialized. On top of that, tests should not be using dedicated code paths that are separate from production. Uninitialized/stub bits are the responsibility of stub implementations, which should use the same constructors as anyone else.

This constructor was correct before, because there was nothing to initialize. But the parameter you now need should go in this constructor, not in a new one. 

I added a commit on top to get rid of most of these stub constructors. I left one (pre-existing) constructor in QEMU as an exercise for later.

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -34,20 +34,30 @@ namespace mpl = multipass::logging;
...
+    : BaseVirtualMachineFactory(derive_instances_dir(*qemu_platform, data_dir)), qemu_platform{std::move(qemu_platform)}
 {
 }
 
 mp::VirtualMachine::UPtr mp::QemuVirtualMachineFactory::create_virtual_machine(const VirtualMachineDescription& desc,
                                                                                VMStatusMonitor& monitor)
 {
-    return std::make_unique<mp::QemuVirtualMachine>(desc, qemu_platform.get(), monitor);
+    return std::make_unique<mp::QemuVirtualMachine>(desc, qemu_platform.get(), monitor,
+                                                    MP_UTILS.make_dir(get_instance_directory_name(desc.vm_name)));
```
ricab, line 57: idem: creating twice

include/multipass/virtual_machine_factory.h
```diff
@@ -59,8 +59,9 @@ class VirtualMachineFactory : private DisabledCopyMove
     virtual VMImage prepare_source_image(const VMImage& source_image) = 0;
     virtual void prepare_instance_image(const VMImage& instance_image, const VirtualMachineDescription& desc) = 0;
     virtual void hypervisor_health_check() = 0;
-    virtual QString get_backend_directory_name() = 0;
-    virtual QString get_backend_version_string() = 0;
+    virtual QString get_backend_directory_name() const = 0;
+    virtual QString get_instance_directory_name(const std::string& name) const = 0;
```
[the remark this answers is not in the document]
ricab: @sharder996 this still says "name" even though it returns a path. Did you forget to change?

ricab (changes requested):
OK, this is getting there, only a few loose ends left to tie (please see comments above too).

src/daemon/daemon.cpp
```diff
@@ -2662,8 +2656,8 @@ void mp::Daemon::persist_instances()
 
 void mp::Daemon::release_resources(const std::string& instance)
 {
-    config->factory->remove_resources_for(instance);
     config->vault->remove(instance);
+    config->factory->remove_resources_for(instance);
```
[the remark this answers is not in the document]
sharder996: Yes, tested on all linux backends and is working as expected.

src/utils/utils.cpp
```diff
@@ -674,3 +674,12 @@ void mp::utils::set_owner_for(mp::SSHSession& session, const std::string& root,
...
                                               vm_group, relative_target.substr(0, relative_target.find_first_of('/'))));
 }
+
+mp::Path mp::Utils::derive_instances_dir(const mp::Path& data_dir, const mp::Path& backend_directory_name,
+                                         const mp::Path& instances_subdir) const
+{
+    if (backend_directory_name.isEmpty())
+        return QDir(data_dir).filePath(instances_subdir);
+    else
+        return QDir(data_dir).filePath(QString("%1/%2").arg(backend_directory_name).arg(instances_subdir));
```
ricab, line 684: Better avoid the explicit separator and use `filePath` twice. I know it's a pain, because you have to create a new QDir... I wish `QDir` could return a subdir.

include/multipass/utils.h
```diff
@@ -154,6 +154,8 @@ class Utils : public Singleton<Utils>
     // virtual machine helpers
     virtual void wait_for_cloud_init(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
                                      const SSHKeyProvider& key_provider) const;
+    virtual Path derive_instances_dir(const Path& data_dir, const Path& backend_directory_name,
```
ricab, line 157: Perfect, well done adding this!

ricab (commented):
Getting real close, but something is fishy now:

```
$ multipass launch -n a
Launched: a
$ ll /var/snap/multipass/common/data/multipassd/vault/instances/a/
total 703784
drwxr-xr-x 3 root root      4096 set 20 11:43 .
drwxr-xr-x 3 root root      4096 set 20 11:43 ..
drwxr-xr-x 2 root root      4096 set 20 11:43 a
-rw-r--r-- 1 root root     53248 set 20 11:43 cloud-init-config.iso
-rw-r--r-- 1 root root 720699392 set 20 11:46 ubuntu-22.04-server-cloudimg-amd64.img
```

Notice the extra "a" subdir.

ricab (approved):
Great Scott :} Looking good on this side, so approving but postponing merging until the other side is reviewed too.

PR #3133, opened 2023-06-19
[snapshots] Implement generic part of deleting snapshots

ricab:
Implement snapshot deletion in Daemon and BaseVirtualMachine classes, along with necessary adaptations in Snapshot classes.


src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -108,7 +163,19 @@ auto BaseVirtualMachine::view_snapshots() const noexcept -> SnapshotVista
 std::shared_ptr<const Snapshot> BaseVirtualMachine::get_snapshot(const std::string& name) const
 {
     const std::unique_lock lock{snapshot_mutex};
-    return snapshots.at(name);
+    try
+    {
+        return snapshots.at(name);
+    }
+    catch (const std::out_of_range&)
```
sharder996, line 170: Catching here will nullify some code in the [daemon](https://github.com/canonical/multipass/blob/d7331c3a4a0d5fbb6b52a43839096c7b7627e423/src/daemon/daemon.cpp#L1781). Since you're changing this can you adjust that at the same time?

sharder996 (changes requested):
Looks good! Just one thing I'd like to see addressed so it doesn't get forgotten.

src/platform/backends/shared/base_virtual_machine.cpp
ricab: Like this?

sharder996: Yup, thanks!

sharder996 (approved):
LGTM!

PR #3134, opened 2023-06-19
[snapshots] Implement QEMU-specific part of snapshot deletion

ricab:


sharder996 (approved):
Looks good! Tested on Linux and macOS and is working as expected.

Issue #3116, opened 2023-06-09
[regression] Issue with 'multipass exec' in version 1.12.0


mkaapu:
**Describe the bug**
`multipass exec vm-instance -- bash -c "..."` behaves differently than with the previous version 1.11.0.

**To Reproduce**
Write and execute the following shell script:
```
multipass launch -n vm-instance
multipass exec vm-instance -- bash -c '
whoami
pwd
ls -al
'
```
the output will be:
```
bash: line 1: whoamipwdls: command not found
```

**Expected behavior**
With the previous multipass version 1.11.0, the output of the above script would be like:
```
ubuntu
/home/ubuntu
total 28
drwxr-x--- 4 ubuntu ubuntu 4096 Jun  9 17:29 .
drwxr-xr-x 3 root   root   4096 Jun  9 17:29 ..
-rw-r--r-- 1 ubuntu ubuntu  220 Jan  6  2022 .bash_logout
-rw-r--r-- 1 ubuntu ubuntu 3771 Jan  6  2022 .bashrc
drwx------ 2 ubuntu ubuntu 4096 Jun  9 17:29 .cache
-rw-r--r-- 1 ubuntu ubuntu  807 Jan  6  2022 .profile
drwx------ 2 ubuntu ubuntu 4096 Jun  9 17:29 .ssh
```
This is also what you would get if you open a shell prompt on the instance and run the shell script with the following content:
```
bash -c '
whoami
pwd
ls -al
'
```

**Logs**
[daemon_log.txt](https://github.com/canonical/multipass/files/11707590/daemon_log.txt)

**Additional info**
 - OS: Ubuntu 22.04 LTS
- `multipass version`:
```
multipass   1.12.0
multipassd  1.12.0
```
- `multipass info --all`
```
Name:           guest-vm
State:          Running
IPv4:           10.64.89.152
Release:        Ubuntu 22.04.2 LTS
Image hash:     604837370399 (Ubuntu 22.04 LTS)
CPU(s):         1
Load:           0.00 0.00 0.00
Disk usage:     1.4GiB out of 4.8GiB
Memory usage:   143.8MiB out of 460.6MiB
Mounts:         --
```
- `multipass get local.driver`: qemu

luis4a0:
Hi @mkaapu, thanks for reporting. This is indeed a regression. We should take care of it soon.

luis4a0:
This didn't fix the regression. Reopening.

PR #3135, opened 2023-06-20
Quote '\n' in strings which go through SSH.

luis4a0:
When we need to pass a string through SSH, we escape some characters. But we forgot about newlines, which are eliminated unless quoted. This PR just put newlines in double quotes, and leave escaping the rest of the characters untouched.

Fixes https://github.com/canonical/multipass/issues/3116.


townsend2010:
Bah, stupid ci...

bors retry

PR #3136, opened 2023-06-21
[snapshots] Output snapshot details in `multipass info`

sharder996:


src/client/cli/formatter/table_formatter.cpp
```diff
@@ -53,91 +53,150 @@ std::string to_usage(const std::string& usage, const std::string& total)
     return fmt::format("{} out of {}", mp::MemorySize{usage}.human_readable(), mp::MemorySize{total}.human_readable());
 }
 
-std::string generate_instance_info_report(const mp::InfoReply& reply)
+std::string generate_snapshot_details(const mp::DetailedInfoItem& item)
 {
     fmt::memory_buffer buf;
+    const auto& fundamentals = item.snapshot_info().fundamentals();
+
+    fmt::format_to(std::back_inserter(buf), "{:<16}{}\n", "Snapshot:", fundamentals.snapshot_name());
```
ricab, line 61: The 16-char width results in 4 spaces of separation here, unlike the 3 that the spec prescribed. However, looking at the output, 4 it is actually a better choice, because it maintains alignment with instance info, so :+1:

I have updated the spec accordingly.

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -53,91 +53,150 @@ std::string to_usage(const std::string& usage, const std::string& total)
...
+
+    auto mount_paths = item.mount_info().mount_paths();
+    fmt::format_to(std::back_inserter(buf), "{:<16}{}", "Mounts:", mount_paths.empty() ? "--\n" : "");
+    for (auto mount = mount_paths.cbegin(); mount != mount_paths.cend(); ++mount)
+    {
+        if (mount != mount_paths.cbegin())
+            fmt::format_to(std::back_inserter(buf), "{:<16}", "");
+        fmt::format_to(std::back_inserter(buf), "{:{}} => {}\n", mount->source_path(),
+                       item.mount_info().longest_path_len(), mount->target_path());
+    }
```
ricab, line 77: Having looked at it, I wonder if we should stick to the current format... WDYT?

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -53,91 +53,150 @@ std::string to_usage(const std::string& usage, const std::string& total)
...
+            fmt::format_to(std::back_inserter(buf), "{:<16}", "");
+        fmt::format_to(std::back_inserter(buf), "{}\n", *child);
+    }
 
-        fmt::format_to(std::back_inserter(buf), "{:<16}{}\n", "Name:", info.name());
-        fmt::format_to(std::back_inserter(buf), "{:<16}{}\n",
-                       "State:", mp::format::status_string_for(info.instance_status()));
-        fmt::format_to(std::back_inserter(buf), "{:<16}{}\n", "Snapshots:", instance_details.num_snapshots());
+    fmt::format_to(std::back_inserter(buf), "{:<16}{}\n",
+                   "Comment:", fundamentals.comment().empty() ? "--" : fundamentals.comment());
```
ricab, line 94: Can we pad extra lines with 16 spaces if the comment is multi-line, please?

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -53,91 +53,150 @@ std::string to_usage(const std::string& usage, const std::string& total)
...
+
+std::string generate_instance_info_report(const mp::InfoReply& reply)
+{
+    fmt::memory_buffer buf;
+
+    for (const auto& info : mp::format::sort_instances_and_snapshots(reply.detailed_report().details()))
+    {
+        if (info.has_instance_info())
+        {
+            fmt::format_to(std::back_inserter(buf), generate_instance_details(info));
```
ricab, line 194: Well done splitting these out, thank you!

ricab (changes requested):
Looking good, thanks! Only a few wrinkles to iron out.

include/multipass/cli/format_utils.h
```diff
@@ -109,6 +112,36 @@ Snapshots multipass::format::sort_snapshots(const Snapshots& snapshots)
...
+            return true;
+        else if (a.name() != petenv_name && b.name() == petenv_name)
+            return false;
+
+        if (a.has_instance_info())
+            return a.name() < b.name();
+        else
+            return TimeUtil::TimestampToNanoseconds(a.snapshot_info().fundamentals().creation_timestamp()) <
+                   TimeUtil::TimestampToNanoseconds(b.snapshot_info().fundamentals().creation_timestamp());
+    });
```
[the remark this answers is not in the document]
sharder996: Good catch!

include/multipass/cli/format_utils.h
```diff
@@ -109,6 +112,36 @@ Snapshots multipass::format::sort_snapshots(const Snapshots& snapshots)
     return ret;
 }
 
+template <typename Details>
+Details multipass::format::sort_instances_and_snapshots(const Details& details)
```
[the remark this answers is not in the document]
sharder996: I completely agree. Although, I wasn't aware of `std::enable_if` and its use case in this scenario. I'd like to take a look at it more though, but for the time being I'll add a todo for the sake of getting this merged (which will make some of the other `multipass info`/`list` changes easier).

src/client/cli/formatter/table_formatter.cpp
sharder996: The part that is omitted is the id mappings, correct? Personally, I think its fine if they're not shown in snapshot info as I would expect those finer details to only become relevant when inspecting the current state of an instance. Additionally, including id mappings will increase the relative length of the `mounts` section to everything else which has more of a terse feel to it.
These are my justifications for it, but I'm not opposed to including more details as `info` is supposed to be giving the user detailed information and they should be using `snapshot-overview` to get less details.

src/client/cli/formatter/table_formatter.cpp
sharder996: Does `fmt` support this? I'm unable to find any solution for this other than manually splitting the string based on terminal width and aligning each line separately.

include/multipass/cli/format_utils.h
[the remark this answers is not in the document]
ricab: OK, fair enough. This would be easier with concepts...

src/client/cli/formatter/table_formatter.cpp
ricab: OK, thanks. Let's leave it then.

src/client/cli/formatter/table_formatter.cpp
ricab: Hmm, I see what you mean. I don't think `fmt` can do that. What I had in mind was just to prepend spaces to existing newlines in the string. That can be achieved with substring replacement. To account for Windows, we need to replace `\r\n` and `\n` when not preceded by `\r`.

Longer lines will still wrap around, but perhaps that is OK for now? WDYT?

include/multipass/cli/format_utils.h
```diff
@@ -135,8 +136,15 @@ Details multipass::format::sort_instances_and_snapshots(const Details& details)
...
         else
+        {
+            if (a.name() < b.name())
+                return true;
+            else if (a.name() > b.name())
+                return false;
+
             return TimeUtil::TimestampToNanoseconds(a.snapshot_info().fundamentals().creation_timestamp()) <
                    TimeUtil::TimestampToNanoseconds(b.snapshot_info().fundamentals().creation_timestamp());
+        }
```
ricab: You can invert the logic to simplify a little:
```suggestion
        {
            if (a.name() == b.name())
                return TimeUtil::TimestampToNanoseconds(a.snapshot_info().fundamentals().creation_timestamp()) <
                   TimeUtil::TimestampToNanoseconds(b.snapshot_info().fundamentals().creation_timestamp());
            return a.name() < b.name();
        }
```

ricab (commented):
Hey @sharder996, thanks for the changes you did. Only a little simplification and the newline thing missing.

ricab (commented):
Almost there.

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -53,91 +55,154 @@ std::string to_usage(const std::string& usage, const std::string& total)
...
+        if (child != children.cbegin())
+            fmt::format_to(std::back_inserter(buf), "{:<16}", "");
+        fmt::format_to(std::back_inserter(buf), "{}\n", *child);
+    }
 
-        for (int i = 1; i < ipv4_size; ++i)
-            fmt::format_to(std::back_inserter(buf), "{:<16}{}\n", "", instance_details.ipv4(i));
+    // TODO@snapshots split and align string if it extends onto several lines
+    QString comment = QString::fromStdString(fundamentals.comment());
+    comment.replace(QRegularExpression("(\r\n|\n)"), "\\1" + QString(" ").repeated(16));
```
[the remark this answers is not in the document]
sharder996: This shouldn't be happening. I changed one of the tests to include some newlines (one of each type of line ending) and it's working correctly.

PR #3141, opened 2023-06-28
[snapshots] Improve purge wording

sharder996:
Added a confirmation action that a user must answer if purging snapshots and instances together in the same command. If you think about it, specifying both instances and snapshots together in a deletion command is a little bit odd and not something that user who is unfamiliar with the command would do. So, if they do happen to specify both, we ask them to confirm the action.

### DIFF
--- a/src/client/cli/cmd/delete.cpp
+++ b/src/client/cli/cmd/delete.cpp
@@ -92,7 +92,7 @@ mp::ParseCode cmd::Delete::parse_args(mp::ArgParser* parser)
                                   "<instance>[.snapshot] [<instance>[.snapshot] ...]");
 
     QCommandLineOption all_option(all_option_name, "Delete all instances and snapshots");
-    QCommandLineOption purge_option({"p", "purge"}, "Purge specified instances and snapshots immediately");
+    QCommandLineOption purge_option({"p", "purge"}, "Permanently delete specified instances and snapshots immediately");
     parser->addOptions({all_option, purge_option});
 
     auto status = parser->commandParse(this);
--- a/src/client/cli/cmd/delete.cpp
+++ b/src/client/cli/cmd/delete.cpp
@@ -24,6 +24,13 @@
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
 
+namespace
+{
+constexpr auto no_purge_base_error_msg =
+    "Snapshots can only be purged (after deletion, they cannot be recovered). Please use the `--purge` "
+    "flag if that is what you want";
+}
+
 mp::ReturnCode cmd::Delete::run(mp::ArgParser* parser)
 {
     auto ret = parse_args(parser);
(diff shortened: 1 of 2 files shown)


src/client/cli/cmd/delete.cpp
```diff
@@ -104,17 +105,44 @@ mp::ParseCode cmd::Delete::parse_args(mp::ArgParser* parser)
...
+                if (!confirm_action(term, "warning: Specifying both instances and snapshots together will also "
+                                          "permanently delete instances.\n"
+                                          "Are you sure you want to continue? (Yes/no)"))
+                    return mp::ParseCode::CommandFail;
+            }
+            catch (const mp::PromptException& e)
+            {
+                std::cerr << e.what() << std::endl;
+                return ParseCode::CommandLineError;
+            }
```
ricab, line 139: Why did you opt for a confirmation question on `--purge`? I see a few problems with this:
  - it would only be shown in interactive terminals, but succeed silently in scripts
  - it would warn when the user added `--purge` to begin with, on their own
  - it is another interactive question (I am a little wary of proliferating those)
  - the confirmation is asked only in a very specific case, making the usage asymmetric for no obvious reason (why ask for confirmation only if snapshots and instances are involved)?
  
IMO, the goal is only to avoid misguiding the user to purge instances inadvertently. If they choose to add the purge flag on their own, I don't think it makes a lot of sense to ask for confirmation.

ricab (changes requested):
Hey @sharder996, I had something a little different in mind. See my comment below and the suggestion I just added in the spec for this. Let me know what you think?

ricab:

src/client/cli/cmd/delete.cpp
```diff
@@ -104,17 +104,40 @@ mp::ParseCode cmd::Delete::parse_args(mp::ArgParser* parser)
...
+
+        if (instance_found)
+            cerr << fmt::format("{}:\n\n\tmultipass delete --purge {}\n\nYou can use a separate command to delete "
+                                "instances without purging them:\n\n\tmultipass delete {}\n",
+                                no_purge_base_error_msg, snapshots, instances);
+        else
+            cerr << fmt::format("{}.\n", no_purge_base_error_msg);
+
+        return mp::ParseCode::CommandLineError;
+    }
```
ricab, line 140: Could this be extracted into a separate method/function, please?

ricab (commented):
That's it. I have a simple ask to avoid the long function, but otherwise looks good.

src/client/cli/cmd/delete.cpp
sharder996: I moved the string to the unnamed namespace slimming this down. Does it make sense to create another function for just an `if` statement?

ricab: > I moved the string to the unnamed namespace slimming this down

Thanks, that makes it better indeed.

> Does it make sense to create another function for just an if statement?

I think it definitely makes sense to have a function with only an `if` statement. The reader most likely wants to see something like `refuse_unpurged_snapshots(...)` without having to read 15 lines of details (which can always be dug into). In any case, here I was actually referring to the whole block that takes care of instance/snapshot arguments, including the `for` loop above. If it was me, I would probably extract that whole block into a function and then the `if` into another function.

ricab (approved):
Perfect, that's it! See how you just made it look much nicer with just a little refactoring? :wink: 

I'll merge after fixing the last rebase I did, which lost merge commits, I just noticed...

PR #3143, opened 2023-06-29
[gui] call c code using dart ffi

andrei-toterman:
This PR add the necessary glue to be able to call C functions through FFI


luis4a0 (approved):
Hey @andrei-toterman, this looks good! I checked the functionality, checked that regexes are correct and read a bit about FFI to understand the code. So everything is OK for me.

Before merging, we would need rebasing. We'll take care ofif when all branches are ready. Thanks!

PR #3151, opened 2023-07-04
[snapshots] Snapshot and update QEMU mount args

ricab:
Keep reference to mount args in snapshot and update when applying. I am still not sure if this is necessary, since the `QemuMountHandler` updates the mount args upon construction and destruction.


ricab:
I couldn't find a use case that needed this after all. If we restored snapshots on suspended instances, we would need this [because](https://github.com/canonical/multipass/blob/4f6bf5cbac952112486686a5af92f7730419afa2/src/platform/backends/qemu/qemu_mount_handler.cpp#L43-L48). But we don't, so closing until we find some other reason.

PR #3152, opened 2023-07-05
[qemu] Allow more firmware directories in AppArmor

ricab:
Update the AppArmor profile for QEMU processes, to allow `/usr/local` based firmware directories as well as `qemu` "leave" directories. This caters to locally built QEMU needs.

### DIFF
--- a/src/platform/backends/qemu/qemu_vm_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vm_process_spec.cpp
@@ -203,7 +203,7 @@ profile %1 flags=(attach_disconnected) {
     catch (const mp::SnapEnvironmentException&)
     {
         signal_peer = "unconfined";
-        firmware = "/usr/share/{seabios,ovmf,qemu-efi}/*";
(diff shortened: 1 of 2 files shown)


townsend2010 (approved):
Hey @ricab!

Thanks for this!  I think this is a good fix for the time being, but as mentioned in the daily sync, I think the long term solution is that if `multipassd` is not running in a Snap, then any processes it spawns should not run with an Apparmor profile applied to it.  I will open an issue for that so we can track it.

bors merge

townsend2010:
Some issue w/ the Mac builder...

bors retry

ricab:
Thanks for the review @townsend2010!

Concerning AppArmor, I think I see your point. Although my first inclination is to reduce packaging-specific differences, I recognize they are necessary one way or the other in this case. And it doesn't make a lot of sense to maintain unnecessary AppArmor rules for files that are not going to be there in the snap. Some dedicated care and testing is always going to be needed for the snap and artificial profiles probably don't help much.

ricab:
bors retry

Issue #1637, opened 2020-07-10
[simplestreams] Fetching manifest data blocks daemon and can cause the client to appear slow


townsend2010:
Currently, there is a single shot `QTimer` that fires when an `ImageHost` is created on daemon start up.  With all non LXD backends, this is not a problem since the timer finally runs _after_ initialization is complete.  However, with LXD, when we do a `lxd_request` for querying information from LXD, Qt will decide the run the timer from the event queue when we have made a call into `NetworkAccessManager` which then delays initialization and blocks connections to the daemon with `EPERM` since the socket permissions are not yet set.

I have never really been satisfied with the `QTimer` approach here and would prefer to have a separate thread do manifest updates so as not to block the daemon during `update_manifest` operations.

ricab:
This happens with qemu too. I see ~10s elapsing while contacting the remotes here, in:

```
$ multipass version
multipass  1.7.0-rc.899+g38dcd724
multipassd 1.7.0-rc.899+g38dcd724
```

PR #3153, opened 2023-07-06
Improve the performance of manifests fetching

georgeliao:
This PR contains a few things. 
1. `update_manifests()` function will no longer be called internally in the `UbuntuImageHost` and `CustomImageHost` class, instead it becomes a public method and gets invoked by client code like `daemon.cpp`. 
2. The two `update_manifests()` calls are wrapped into one function and it is launched by the daemon constructor asynchronously and it will run periodically. 
3. Because of the first two, the `manifest_time_to_live` related codes are all removed since it is no longer relevant. 
4. Inside `UbuntuImageHost::update_manifests` and `CustomImageHost::update_manifests`, the parallelization of the manifests fetching is also applied in order to speed it up. 
---
[Epic link](https://warthogs.atlassian.net/browse/MULTI-254)
[issue link](https://github.com/canonical/multipass/issues/1637)

### DIFF
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -30,6 +30,22 @@ namespace mpl = multipass::logging;
 
 namespace multipass::utils
 {
+// Because AsyncPeriodicTask needs to be instantiated as data member of a class to work, and Class template argument
+// deduction (CTAD) does not work for data member, so that means we can not deduced type data member from the below code
+// snippet.
+
+// template <typename Callable, typename...Args>
+// class AsyncPeriodicTask
+// {
+// public:
+//     using ReturnType = std::invoke_result_t<std::decay_t<Callable>, std::decay_t<Args>...>;
+//     AsyncPeriodicTask(Callable&& func, Args&&... args);
+//     ...
+// };
+
+// Therefore, the user has to specify the ReturnType as template argument while constructing it.
+
+template <typename ReturnType = void>
 class AsyncPeriodicTask
 {
 public:
@@ -70,7 +86,7 @@ public:
 
 private:
     QTimer timer;
-    std::future<void> future;
+    std::future<ReturnType> future;
 };
 } // namespace multipass::utils
 
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -59,13 +59,13 @@ public:
         // mock_logger in the unit tests.
 
         // TODO, remove the launch_msg parameter once we have better class separation.
-        mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
+        mpl::log(mpl::Level::debug, "async task", std::string(launch_msg));
         future = std::async(std::launch::async, std::forward<Callable>(func), std::forward<Args>(args)...);
         QObject::connect(&timer, &QTimer::timeout, [launch_msg, this, func, args...]() -> void {
             // skip it if the previous one is still running
             if (future.wait_for(std::chrono::seconds(0)) == std::future_status::ready)
             {
-                mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
+                mpl::log(mpl::Level::debug, "async task", std::string(launch_msg));
                 future = std::async(std::launch::async, func, args...);
             }
         });
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -30,9 +30,9 @@ namespace mpl = multipass::logging;
 
 namespace multipass::utils
 {
-// Because AsyncPeriodicTask needs to be instantiated as data member of a class to work, and Class template argument
-// deduction (CTAD) does not work for data member, so that means we can not deduced type data member from the below code
-// snippet.
+// Because AsyncPeriodicTask needs to be instantiated as data member of a class to function properly, and Class template
+// argument deduction (CTAD) does not work for data member, so that means we can not deduced type data member from the
+// below code snippet.
 
 // template <typename Callable, typename...Args>
 // class AsyncPeriodicTask
@@ -50,7 +50,7 @@ class AsyncPeriodicTask
 {
 public:
     template <typename Callable, typename... Args>
-    void launch(std::string_view launch_msg, std::chrono::milliseconds msec, Callable&& func, Args&&... args)
+    AsyncPeriodicTask(std::string_view launch_msg, std::chrono::milliseconds msec, Callable&& func, Args&&... args)
     {
         // Log in a side thread will cause some unit tests (like launch_warns_when_overcommitting_disk) to have data
         // race on log, because the side thread log messes with the mock_logger. Because of that, we only allow the
@@ -71,6 +71,7 @@ public:
         });
         timer.start(msec);
     }
+
     void start_timer()
     {
         timer.start();
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_ASYNC_PERIODIC_TASK_H
+#define MULTIPASS_ASYNC_PERIODIC_TASK_H
+
+#include <multipass/logging/log.h>
+
+#include <chrono>
+#include <future>
+#include <string_view>
+
+#include <QTimer>
+
+namespace mpl = multipass::logging;
+
+namespace multipass::utils
+{
+class AsyncPeriodicTask
+{
+public:
+    template <typename Callable, typename... Args>
+    void launch(std::string_view launch_msg, std::chrono::milliseconds msec, Callable&& func, Args&&... args)
+    {
+        // Log in a side thread will cause some unit tests (like launch_warns_when_overcommitting_disk) to have data
+        // race on log, because the side thread log messes with the mock_logger. Because of that, we only allow the
+        // main thread log for now. That is why launch_msg parameter is here. A long-term solution would be a better
+        // separation of classes and mock the corresponding class function, so it will not log and mess with the
+        // mock_logger in the unit tests.
+
+        // TODO, remove the launch_msg parameter once we have better class separation.
+        mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
+        future = std::async(std::launch::async, std::forward<Callable>(func), std::forward<Args>(args)...);
+        QObject::connect(&timer, &QTimer::timeout, [launch_msg, this, func, args...]() -> void {
+            // skip it if the previous one is still running
+            if (future.wait_for(std::chrono::seconds(0)) == std::future_status::ready)
+            {
+                mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
+                future = std::async(std::launch::async, func, args...);
+            }
+        });
+        timer.start(msec);
+    }
+    void start_timer()
+    {
+        timer.start();
+    }
+    void stop_timer()
+    {
+        timer.stop();
+    }
+    void wait_ongoing_task_finish() const
+    {
+        future.wait();
+    }
+
+private:
+    QTimer timer;
+    std::future<void> future;
+};
+} // namespace multipass::utils
+
+#endif // MULTIPASS_TIMER_H
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -142,11 +142,15 @@ bool is_default_constructed(const T& input_type)
 
 // simplified parallel transform, it takes a std container and a unary operation and
 // returns a std::vector<OutputValueType> where the OutputValueType is the unary operation return type
+// There are two options of the return types, one is the one below and the other one is auto. Eventually, I went with
+// the std::invoke_result_t based one because it makes the function signature more expressive despite the fact that it
+// makes std::invoke_result_t<std::decay_t<UnaryOperation>, InputValueType> code duplicate.
 template <typename Container, typename UnaryOperation>
-auto parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
+std::vector<std::invoke_result_t<std::decay_t<UnaryOperation>, typename Container::value_type>>
+parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
 {
     using InputValueType = typename Container::value_type;
-    using OutputValueType = std::invoke_result_t<UnaryOperation, InputValueType>;
+    using OutputValueType = std::invoke_result_t<std::decay_t<UnaryOperation>, InputValueType>;
     const auto num_elements = input_container.size();
 
     // Pre-allocate space for futures
@@ -155,7 +159,7 @@ auto parallel_transform(const Container& input_container, UnaryOperation&& unary
     int index = 0;
     for (const auto& item : input_container)
     {
-        futures[index++] = std::async(std::launch::async, unary_op, item);
+        futures[index++] = std::async(std::launch::async, unary_op, std::cref(item));
     }
 
     std::vector<OutputValueType> results;
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -171,6 +171,23 @@ auto parallel_transform(const Container& input_container, UnaryOperation&& unary
     return results;
 }
 
+template <typename Container, typename UnaryOperation>
+void parallel_for_each(Container& input_container, UnaryOperation&& unary_op)
+{
+    const auto num_elements = input_container.size();
+    std::vector<std::future<void>> empty_futures;
+    empty_futures.reserve(num_elements);
+
+    for (auto& item : input_container)
+    {
+        empty_futures.emplace_back(std::async(std::launch::async, unary_op, std::ref(item)));
+    }
+
+    for (auto& empty_future : empty_futures)
+    {
+        empty_future.get();
+    }
+}
 } // namespace utils
 
 class Utils : public Singleton<Utils>
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -112,7 +112,7 @@ std::string match_line_for(const std::string& output, const std::string& matcher
 // virtual machine helpers
 bool is_running(const VirtualMachine::State& state);
 void wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
-    std::function<void()> const& ensure_vm_is_running = []() {});
+                       std::function<void()> const& ensure_vm_is_running = []() {});
 std::string run_in_ssh_session(SSHSession& session, const std::string& cmd);
 
 // yaml helpers
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -28,6 +28,7 @@
 
 #include <chrono>
 #include <functional>
+#include <future>
 #include <sstream>
 #include <string>
 #include <thread>
@@ -110,8 +111,9 @@ std::string match_line_for(const std::string& output, const std::string& matcher
 
 // virtual machine helpers
 bool is_running(const VirtualMachine::State& state);
-void wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
-                       std::function<void()> const& ensure_vm_is_running = []() {});
+void wait_until_ssh_up(
+    VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
+    std::function<void()> const& ensure_vm_is_running = []() {});
 std::string run_in_ssh_session(SSHSession& session, const std::string& cmd);
 
 // yaml helpers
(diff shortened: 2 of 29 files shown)


src/daemon/custom_image_host.cpp
```diff
@@ -91,56 +97,77 @@ auto map_aliases_to_vm_info_for(const std::vector<mp::VMImageInfo>& images)
...
-                                        base_image_info.hash, // id
-                                        "",
-                                        base_image_info.last_modified, // version
-                                        0,
-                                        true};
-
-        default_images.push_back(full_image_info);
+        empty_futures.emplace_back(std::async(std::launch::async, fetch_one_image_info_and_write_to_index, index,
+                                              url_downloader, image_info_pair));
+        ++index;
```
townsend2010, line 151: I *believe* it [should be defined behavior](https://stackoverflow.com/a/598150) if you do
```
 empty_futures.emplace_back(std::async(std::launch::async, fetch_one_image_info_and_write_to_index, index++,
                                              url_downloader, image_info_pair));
```
which will clean this up a bit.

src/daemon/custom_image_host.cpp
```diff
@@ -206,18 +233,23 @@ std::vector<std::string> mp::CustomVMImageHost::supported_remotes()
     return remotes;
 }
 
-void mp::CustomVMImageHost::fetch_manifests()
+void mp::CustomVMImageHost::fetch_manifests(bool is_force_update_from_network)
```
townsend2010, line 236: Perhaps `const bool`?

src/daemon/custom_image_host.h
```diff
@@ -36,25 +36,25 @@ struct CustomManifest
...
     std::vector<std::pair<std::string, VMImageInfo>> all_info_for(const Query& query) override;
     std::vector<VMImageInfo> all_images_for(const std::string& remote_name, const bool allow_unsupported) override;
     std::vector<std::string> supported_remotes() override;
 
-protected:
+private:
     void for_each_entry_do_impl(const Action& action) override;
     VMImageInfo info_for_full_hash_impl(const std::string& full_hash) override;
-    void fetch_manifests() override;
+    void fetch_manifests(bool is_force_update_from_network) override;
```
townsend2010, line 56: Maybe `const bool` here too?

src/daemon/daemon.cpp
```diff
@@ -3019,3 +3022,30 @@ void mp::Daemon::finish_async_operation(const std::string& async_future_key)
     if (async_op_result.status_promise)
         async_op_result.status_promise->set_value(async_op_result.status);
 }
+
+void mp::Daemon::update_manifests_all(bool is_force_update_from_network)
```
townsend2010, line 3026: Again, perhaps `const bool`?

src/daemon/daemon.cpp
```diff
@@ -3019,3 +3022,30 @@ void mp::Daemon::finish_async_operation(const std::string& async_future_key)
...
+{
+    std::vector<std::future<void>> empty_futures;
+    empty_futures.reserve(config->image_hosts.size());
```
townsend2010, line 3029: Same question as before, would
```
std::vector<std::future<void>> empty_futures(config->image_hosts.size());
```
work instead?

src/daemon/daemon.cpp
```diff
@@ -3019,3 +3022,30 @@ void mp::Daemon::finish_async_operation(const std::string& async_future_key)
...
+        empty_futures.emplace_back(std::async(std::launch::async, &VMImageHost::update_manifests, image_host.get(),
+                                              is_force_update_from_network));
+    }
+    for (auto& empty_future : empty_futures)
+    {
+        empty_future.get();
+    }
+}
+
+void mp::Daemon::wait_update_manifests_all_and_optionally_applied_force(bool force_manifest_network_download)
```
townsend2010, line 3041: That is a mouthful for a function name :wink: 

But, `const bool`?

src/daemon/daemon.h
```diff
@@ -164,6 +165,9 @@ public slots:
                              std::promise<grpc::Status>* status_promise, const std::string& errors);
     void finish_async_operation(const std::string& async_future_key);
     QFutureWatcher<AsyncOperationStatus>* create_future_watcher(std::function<void()> const& finished_op = []() {});
+    void update_manifests_all(bool is_force_update_from_network = false);
```
townsend2010, line 168: `const bool`?

src/daemon/daemon.h
```diff
@@ -164,6 +165,9 @@ public slots:
...
+    // it is applied in Daemon::find wherever the image info fetching is involved, aka non-only-blueprints case
+    void wait_update_manifests_all_and_optionally_applied_force(bool force_manifest_network_download);
```
townsend2010, line 170: `const bool`?

src/daemon/ubuntu_image_host.cpp
```diff
@@ -42,27 +43,27 @@ namespace
...
 constexpr auto index_path = "streams/v1/index.json";
 
-auto download_manifest(const QString& host_url, mp::URLDownloader* url_downloader)
+bool is_default_constructed(const std::pair<std::string, std::unique_ptr<mp::SimpleStreamsManifest>>& manifest_pair)
 {
-    auto json_index = url_downloader->download({host_url + index_path});
+    return manifest_pair == std::pair<std::string, std::unique_ptr<mp::SimpleStreamsManifest>>{};
+}
+
+auto download_manifest(const QString& host_url, mp::URLDownloader* url_downloader, bool is_force_update_from_network)
```
townsend2010, line 51: I think you get the idea- `const bool`

src/daemon/ubuntu_image_host.cpp
```diff
@@ -42,27 +43,27 @@ namespace
...
-            host_url + info.image_location,
-            info.id,
-            info.stream_location,
-            info.version,
-            info.size,
+    return {info.aliases,   info.os,
+            info.release,   info.release_title,
+            info.supported, host_url + info.image_location,
+            info.id,        info.stream_location,
+            info.version,   info.size,
```
townsend2010, line 66: Ugh, I assume this is `clang-format` doing this, right?

src/daemon/ubuntu_image_host.cpp
```diff
@@ -262,11 +267,38 @@ void mp::UbuntuVMImageHost::fetch_manifests()
...
+    {
+        empty_futures.emplace_back(
+            std::async(std::launch::async, fetch_one_remote_and_write_to_index, index, remote_name, remote_info));
+        ++index;
+    }
+
+    for (auto& empty_future : empty_futures)
+    {
+        empty_future.get(); // use get instead of wait to retain the exception throwing
+    }
```
townsend2010, line 294: This really seems like repeated code from earlier like in `custom_image_host.cpp` and `daemon.cpp`.  Any way to refactor this into a common way of doing this to avoid the code duplication?

townsend2010 (changes requested):
Hi @georgeliao!

Thanks for this!  Good work!  I have some feedback below and still need to review the unit tests, but thought I would go ahead and give you what I've found so far.

include/multipass/async_periodic_task.h
```diff
@@ -0,0 +1,94 @@
...
+        // mock_logger in the unit tests.
+
+        // TODO, remove the launch_msg parameter once we have better class separation.
+        mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
+        future = std::async(std::launch::async, std::forward<Callable>(func), std::forward<Args>(args)...);
+        QObject::connect(&timer, &QTimer::timeout, [launch_msg, this, func, args...]() -> void {
+            // skip it if the previous one is still running
+            if (future.wait_for(std::chrono::seconds(0)) == std::future_status::ready)
+            {
+                mpl::log(mpl::Level::info, "async task", std::string(launch_msg));
```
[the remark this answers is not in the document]
georgeliao: Hmm, I thought this should be seen even by the user when they look into the log, not just for us.

src/daemon/custom_image_host.cpp
```diff
@@ -67,11 +68,16 @@ const QMap<QString, QMap<QString, CustomImageInfo>> multipass_image_info{
...
        {"https://cdimage.ubuntu.com/ubuntu-core/22/stable/current/", {"core22"}, "Ubuntu", "core-22", "Core 22"}}}}};
 
+bool is_default_constructed(const mp::VMImageInfo& image_info)
+{
+    return image_info == mp::VMImageInfo{};
+}
+
 auto base_image_info_for(mp::URLDownloader* url_downloader, const QString& image_url, const QString& hash_url,
-                         const QString& image_file)
+                         const QString& image_file, bool is_force_update_from_network = false)
```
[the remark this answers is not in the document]
georgeliao: Yes, something I overlooked.

src/daemon/daemon.h
```diff
@@ -173,6 +177,8 @@ public slots:
     std::unordered_set<std::string> allocated_mac_addrs;
     DaemonRpc daemon_rpc;
     QTimer source_images_maintenance_task;
+    multipass::utils::AsyncPeriodicTask<void> update_manifests_all_task{
+        "fetch manifest periodically", std::chrono::seconds(10), &Daemon::update_manifests_all, this, false};
```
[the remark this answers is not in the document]
georgeliao: Yep, I kept it like this for testing reason since we are still iterating; however, maybe it is better to change it now and apply local change to test.

src/daemon/ubuntu_image_host.cpp
georgeliao: Yep, actually, clang-format is the thing I have been meaning to discuss about. There are some parameters can be fine-tuned to format the file better, this code snippet is one example that can be improved.

src/daemon/ubuntu_image_host.cpp
georgeliao: I was aware of that. I have given some thoughts on the `parallel_for` thing, but did not find a good way to do it, I can have a look again.

src/daemon/daemon.cpp
georgeliao: Yep, but I also did not have a better name in mind which can describe what he does. Anyway, it is just a pre-process step that needs to be applied before retrieving images in the daemon::find function.

src/daemon/daemon.h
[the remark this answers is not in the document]
townsend2010: Yeah, for this review, let's change it to how we want it to work, so I can review the actual behavior.  Thanks!

src/daemon/custom_image_host.cpp
georgeliao: Yep, it can be done.

src/daemon/custom_image_host.cpp
```diff
@@ -91,56 +97,77 @@ auto map_aliases_to_vm_info_for(const std::vector<mp::VMImageInfo>& images)
...
+                                                base_image_info.hash, // id
+                                                "",
+                                                base_image_info.last_modified, // version
+                                                0,
+                                                true};
+    };
 
-    for (const auto& image_info : custom_image_info.toStdMap())
+    std::vector<std::future<void>> empty_futures;
+    empty_futures.reserve(custom_image_info.size());
```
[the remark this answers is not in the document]
georgeliao: That one is slightly different in terms of the behavior. `std::vector` constructor with size allocated `custom_image_info.size()` spaces and **defaultly constructed all the objects**, whereas `reserve` only allocate the raw spaces. This makes a difference to the later code `empty_futures.emplace_back(...);` if we go with `std::vector<std::future<void>> empty_futures(custom_image_info.size()));`, then the code will need to be changed to `empty_futures[index] = ...;` because `empty_futures` is already filled in with default objects, we only can overwrite it. I personally prefer the `reserve` here because the `std::future` object is lazily constructed.

include/multipass/async_periodic_task.h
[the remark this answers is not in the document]
townsend2010: I'm not sure if a user knowing that an asynchronous task started is something they should know about.  It seems more a debugging type thing, but perhaps others have an opinion.  @ricab, @luis4a0, et. al., any thoughts?

src/daemon/custom_image_host.cpp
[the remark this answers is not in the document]
townsend2010: Ah, ok, that makes sense.  Thanks for the explanation and I'm good with this.

src/daemon/daemon.cpp
townsend2010: The name is ok, just sayin'.  But the `const bool` stands. :slightly_smiling_face:

src/client/cli/cmd/find.cpp
```diff
@@ -90,6 +93,12 @@ mp::ParseCode cmd::Find::parse_args(mp::ArgParser* parser)
         return ParseCode::CommandLineError;
     }
 
+    if (parser->isSet(force_manifest_network_download) && parser->isSet(blueprintsOnlyOption))
+    {
+        cerr << "Force updating blueprints are not currently supported\n";
```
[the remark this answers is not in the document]
townsend2010: Yeah, you are right, it should be `is` instead of `are`.

include/multipass/async_periodic_task.h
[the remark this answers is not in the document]
ricab: Hmm, good question. I guess it's a little subjective... It depends on the task in question, on the frequency...

On the one hand, asynchronous tasks in general are an implementation detail. On the other, they probably involve a significant unit of work that the user may be interested in. But does it run every few seconds? We could parameterize the task class with how informative we want it to be, but that would feel like overkill here honestly. 

I would consider the concrete tasks this will be used for. Provided they are infrequent enough (15min sounds reasonable), and interesting enough (fetching image info arguably is), I would be fine with both `debug` and `info`. It can always be changed again if people complain.

Note: a period of 15min means 60*24/15 = 96 logs per day. IMO, that is perfectly within boundaries for a daemon like ours. It adds up to 34K/year or so, but that's just the nature of logging. Consumers should know that logs add up. Hopefully log facilities manage it by archiving, rotating, or something else.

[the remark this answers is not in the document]
luis4a0: Hey! I agree, this should be `debug`. The user doesn't need to know when the manifests are being fetched.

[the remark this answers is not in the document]
georgeliao: Thanks for all the inputs, I will change it to `debug`.

include/multipass/utils.h
```diff
@@ -110,8 +111,9 @@ std::string match_line_for(const std::string& output, const std::string& matcher
 
 // virtual machine helpers
 bool is_running(const VirtualMachine::State& state);
-void wait_until_ssh_up(VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
-                       std::function<void()> const& ensure_vm_is_running = []() {});
+void wait_until_ssh_up(
+    VirtualMachine* virtual_machine, std::chrono::milliseconds timeout,
+    std::function<void()> const& ensure_vm_is_running = []() {});
```
townsend2010, line 116: This looks like a `clang-format` change?  If so, please only run `clang-format` on the diff, not the whole file.  Unfortunately, it adds changes that really don't need to be reviewed.  Thanks!

include/multipass/utils.h
```diff
@@ -127,10 +129,69 @@ std::string qenum_to_string(RegisteredQtEnum val);
...
+    return input_type == T{};
+}
+
+// simplified parallel transform, it takes a std container and a unary operation and
+// returns a std::vector<OutputValueType> where the OutputValueType is the unary operation return type
+// There are two options of the return types, one is the one below and the other one is auto. Eventually, I went with
+// the std::invoke_result_t based one because it makes the function signature more expressive despite the fact that it
+// makes std::invoke_result_t<std::decay_t<UnaryOperation>, InputValueType> code duplicate.
+template <typename Container, typename UnaryOperation>
+std::vector<std::invoke_result_t<std::decay_t<UnaryOperation>, typename Container::value_type>>
```
townsend2010, line 149: It looks like this type, as you mentioned, is duplicated and is hard to read.  Any way to instead add the `using`s before this and use them here and in the function?

include/multipass/utils.h
```diff
@@ -127,10 +129,69 @@ std::string qenum_to_string(RegisteredQtEnum val);
...
+parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
+{
+    using InputValueType = typename Container::value_type;
```
townsend2010, line 152: I don't see `InputValueType` used anywhere else accept for the following `using OutputValueType = ...`.  I'm not sure there is any value in defining this.

include/multipass/utils.h
```diff
@@ -127,10 +129,69 @@ std::string qenum_to_string(RegisteredQtEnum val);
...
+}
+
+// simplified parallel transform, it takes a std container and a unary operation and
+// returns a std::vector<OutputValueType> where the OutputValueType is the unary operation return type
+// There are two options of the return types, one is the one below and the other one is auto. Eventually, I went with
+// the std::invoke_result_t based one because it makes the function signature more expressive despite the fact that it
+// makes std::invoke_result_t<std::decay_t<UnaryOperation>, InputValueType> code duplicate.
+template <typename Container, typename UnaryOperation>
+std::vector<std::invoke_result_t<std::decay_t<UnaryOperation>, typename Container::value_type>>
+parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
```
townsend2010, line 150: Nice, thanks for this!

src/daemon/ubuntu_image_host.cpp
```diff
@@ -34,6 +35,7 @@
 #include <QUrl>
 
 #include <algorithm>
+#include <future>
```
townsend2010, line 38: Again, after refactor, is this still needed?

src/daemon/custom_image_host.cpp
```diff
@@ -29,6 +30,7 @@
 #include <QMap>
 #include <QUrl>
 
+#include <future>
```
townsend2010, line 33: After the refactoring of the parallel request code, is this include still needed?

src/daemon/ubuntu_image_host.cpp
townsend2010: I really don't see a change, so ideally, as mentioned before, `clang-format` should only run on the diff, not the whole file to avoid these types of things that make it slightly harder to review.

src/network/url_downloader.cpp
```diff
@@ -74,18 +74,18 @@ void wait_for_reply(QNetworkReply* reply, QTimer& download_timeout)
 }
 
 template <typename ProgressAction, typename DownloadAction, typename ErrorAction, typename Time>
-QByteArray download(QNetworkAccessManager* manager, const Time& timeout, QUrl const& url, ProgressAction&& on_progress,
-                    DownloadAction&& on_download, ErrorAction&& on_error, const std::atomic_bool& abort_download,
-                    const bool force_cache = false)
+QByteArray
+download(QNetworkAccessManager* manager, const Time& timeout, QUrl const& url, ProgressAction&& on_progress,
+         DownloadAction&& on_download, ErrorAction&& on_error, const std::atomic_bool& abort_download,
+         QNetworkRequest::CacheLoadControl cache_load_control = QNetworkRequest::CacheLoadControl::PreferNetwork)
```
townsend2010, line 80: I suppose this can be `const`.

townsend2010 (changes requested):
Hey @georgeliao!

We're getting closer!  I did look through the unit tests this time.  Some more comments inline.  Thanks!

src/daemon/custom_image_host.cpp
georgeliao: Good catch, I always thought clang-tidy automatically detects unused include, but it turned out that it does not. I will search for other tools on linux which can do this.

include/multipass/utils.h
georgeliao: Good doubt you have raised, I am afraid there is a proper way to do this. The challenge here is that the `using` can only called in the function body in the scope of this function and we need that alias on the declaration of the function which is before the function body. 

However, there is a workaround. If we define the alias out of this scope by using template alias. We will have the result code like this:
```
template <typename Container, typename UnaryOperation>
using OutputValueType = std::invoke_result_t<std::decay_t<UnaryOperation>, typename Container::value_type>;

template <typename Container, typename UnaryOperation>
std::vector<OutputValueType<Container, UnaryOperation>>
parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
{
    using OutputValueType = OutputValueType<Container, UnaryOperation>;
    const auto num_elements = input_container.size();

    // Pre-allocate space for futures
    std::vector<std::future<OutputValueType>> futures(num_elements);
    ...
}
```
The downside here is that the duplication of the `template <typename Container, typename UnaryOperation>` scope and the scope of the alias `OutputValueType` is no longer within the `parallel_transform` function. Plus, the user of the alias has to call like `OutputValueType<Container, UnaryOperation>` with the specified template parameter. 

As to which one to choose, the original one is the same as `std::async`'s function's signature and it only has one template scope, on the other side, the new one does eliminate the duplication. I do not have a strong opinion on this. If you prefer the new one, then I will change it.

townsend2010: After considering what you've said, I think we should leave it as it is now.  Thanks!

townsend2010:
Hey @georgeliao!

I think we're really close on this.  I'd prefer the `clang-format` only changes to be fixed and then I think it's ready.

include/multipass/utils.h
townsend2010: @georgeliao :point_up:

georgeliao: Yep, this can be fixed. Originally, I thought `InputValueType` is more expressive and symmetric to `OutputValueType`. So It adds a small value even though it is only used once. However, putting `Container::value_type` in one line with `OutputValueType` might make things cleaner.

townsend2010: Ok, I think it's fine the way it is.

include/multipass/utils.h
```diff
@@ -76,7 +77,10 @@ void remove_directories(const std::vector<QString>& dirs);
 
 // filesystem mount helpers
 void make_target_dir(SSHSession& session, const std::string& root, const std::string& relative_target);
-void set_owner_for(SSHSession& session, const std::string& root, const std::string& relative_target, int vm_user,
+void set_owner_for(SSHSession& session,
+                   const std::string& root,
+                   const std::string& relative_target,
+                   int vm_user,
```
georgeliao, line 83: clang-format

tests/mock_image_host.h
```diff
@@ -110,17 +111,9 @@ class MockImageHost : public VMImageHost
...
-                                       true,
-                                       image.url(),
-                                       custom_image_id,
-                                       "",
-                                       custom_image_version,
-                                       1,
-                                       false};
+    VMImageInfo mock_custom_image_info{
+        {custom_alias},       "Ubuntu", "Custom Core", custom_release_info, true, image.url(), custom_image_id, "",
+        custom_image_version, 1,        false};
```
townsend2010, line 116: Ugh, another `clang-format` change it appears :slightly_frowning_face:

src/daemon/ubuntu_image_host.cpp
```diff
@@ -280,12 +295,11 @@ mp::SimpleStreamsManifest* mp::UbuntuVMImageHost::manifest_from(const std::strin
...
-
-    auto it = std::find_if(manifests.begin(), manifests.end(),
-                           [&remote](const std::pair<std::string, std::unique_ptr<SimpleStreamsManifest>>& element) {
-                               return element.first == remote;
-                           });
+    const auto it =
+        std::find_if(manifests.cbegin(), manifests.cend(),
+                     [&remote](const std::pair<std::string, std::unique_ptr<SimpleStreamsManifest>>& element) {
+                         return element.first == remote;
+                     });
```
townsend2010, line 302: Another `clang-format` only change...

townsend2010 (changes requested):
Hey @georgeliao!

I think I found 2 more `clang-format` only changes below.  Once those are fixed, I'll approve.

src/daemon/ubuntu_image_host.cpp
georgeliao: This one is fine, I prepended const (an actual code change) which makes different line breaks. The other one probably is.

tests/mock_image_host.h
georgeliao: This one can be fixed.

src/daemon/ubuntu_image_host.cpp
townsend2010: Ah, I see, a subtle change.  Thanks!

georgeliao:
@townsend2010 , sorry for that, thanks for the careful check. That was also reverted so now I think we are good.

tests/mock_image_host.h
townsend2010: Thanks!

townsend2010:
Hey @georgeliao!

Thanks and no worries :slightly_smiling_face: 

I should be approving soon!

townsend2010 (approved):
Ok, let's do it!  Thanks!

bors merge

townsend2010:
The integration tests did indeed pass: https://github.com/canonical/multipass/actions/runs/6346117580/job/17239174617

Going to merge manually!

PR #3155, opened 2023-07-11
[completions] autocomplete on state

sharder996:
If an instance contains a string which is used for an instance state, bash completions can suggest the instance in invalid situations. eg. `multipass stop <tab>` will suggest an instance named `Running-Rabbit` even if it's stopped because the name contains the string `Running`.


ricab (approved):
Yup, thanks for the separate PR!

bors r+

PR #3156, opened 2023-07-11
[bash] remove duplicates from completions

sharder996:
Bash completions will currently suggest repeated flags endlessly. This PR filters out previously seen options/flags with the assumption that no multipass currently accepts this pattern.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -198,7 +198,8 @@ _multipass_complete()
 
     if [[ "${multipass_cmds}" =~ " ${cmd} " || "${multipass_cmds}" =~ ^${cmd} || "${multipass_cmds}" =~ \ ${cmd}$ ]];
     then
-        opts="--help --verbose"
+        _add_nonrepeating_args "--help"
+        opts="${opts} --verbose"
     fi
 
     case "${cmd}" in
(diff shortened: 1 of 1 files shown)


ricab (changes requested):
Hmm, this doesn't seem to work for me:

```
$ multipass list --[TAB][TAB]
--format   --help     --verbose
$ multipass list --help --[TAB][TAB]
--format   --help     --verbose
$ multipass list --help --h[TAB]
$ multipass list --help --help
```

BTW, `--verbose` can be used multiple times... But I guess we can overlook that for the benefit here. People would use `-v` for that.

ricab:

ricab (changes requested):
OK, so that original issue I saw is fixed, but I have found just a little remaining problem.

In the stable channel, if I do `multipass info[TAB]`, `info` is completed, meaning that a space is added. With this PR, the cursor doesn't move. That's because `info` is already present so it is not suggested. This means that something like `multipass alias[TAB]` gets auto-completed to `aliases`, instead of offering the two options as in stable.

Is there a way you can exclude the current word from the list of existing words?

sharder996:
Hey @ricab, I believe the issue you addressed in your last comment is fixed. Ready for another review pass now.

ricab:
OK, thanks for letting me know. Enqueuing...

completions/bash/multipass
```diff
@@ -388,7 +388,12 @@ _multipass_complete()
     fi
 
     if [[ -n "${opts}" ]]; then
-        COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+        for word in "${COMP_WORDS[@]}"; do
+            if [[ "${COMP_LINE}" == *"${word} "* ]]; then
+                opts=("${opts[@]/$word}")
```
ricab, line 393: Hey @sharder996, sorry, I am afraid this won't do because it can remove strings where it shouldn't. For instance, I have instances called `a` and `b`. Here's what completion does:

```
$ multipass info [TAB][TAB]
a          --all      b          --format   --help     --verbose
$ multipass info a [TAB][TAB]
b          --format   --help     --ll       --verbose
```

It removed all "a"s from `opts`, so `--all` became `--ll`...

ricab (changes requested):
Hey Scott, I was about to approve when I noticed something :crying_cat_face:

completions/bash/multipass
ricab: I think this would work:

```suggestion
                opts=("${opts[@]/\b$word\b}")
```

ricab:
Hey @sharder996, thought you'd said you had something here, did you forget to push? Or did I misunderstand?

sharder996:
Hey @ricab, I've been working on it locally, but haven't been able to come up with an incantation that works yet.

ricab:
OK got it, thanks.

ricab (changes requested):
Hey @sharder996, something still amiss :confounded: 

```
$ multipass list
a                       Stopped           --               Ubuntu 18.04 LTS
b                       Stopped           --               Ubuntu 22.04 LTS
c                       Stopped           --               Ubuntu 23.04
$ multipass start [TAB][TAB]
--all      b          c          --help     --verbose
$ multipass shell [TAB][TAB]
b          c          --help     --verbose
```

Instance `a` is not in the suggestions. It is if I use stable.

sharder996:
Ugh, bash completions is just the gift that keeps on giving.

Anyways, this was an easy and should be working properly now.

ricab (approved):
Alright, appears to be good now. Thanks!

bors r+

ricab:
This is unrelated to mac and windows, so I am going ahead and merging.

PR #3161, opened 2023-07-13
[snapshots] populate snapshot details

sharder996:
Rebased with changes from #3136 to make testing easier

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1269,6 +1269,15 @@ void populate_snapshot_fundamentals(std::shared_ptr<const mp::Snapshot> snapshot
     timestamp->set_nanos(snapshot->get_creation_timestamp().time().msec() * 1'000'000);
 }
 
+void populate_snapshot_overview(const std::string& instance_name, std::shared_ptr<const mp::Snapshot> snapshot,
+                                mp::SnapshotOverviewInfoItem* overview)
+{
+    auto fundamentals = overview->mutable_fundamentals();
+
+    overview->set_instance_name(instance_name);
+    populate_snapshot_fundamentals(snapshot, fundamentals);
+}
+
 void populate_mount_info(const std::unordered_map<std::string, mp::VMMount>& mounts, mp::MountInfo* mount_info,
                          bool& have_mounts)
 {
@@ -1743,24 +1752,13 @@ try // clang-format on
                     vm.get_snapshot(snapshot); // verify validity of any snapshot name requested separately
 
                 for (const auto& snapshot : vm.view_snapshots())
-                {
-                    auto overview = response.mutable_snapshot_overview()->add_overview();
-                    auto fundamentals = overview->mutable_fundamentals();
-
-                    overview->set_instance_name(name);
-                    populate_snapshot_fundamentals(snapshot, fundamentals);
-                }
+                    populate_snapshot_overview(name, snapshot, response.mutable_snapshot_overview()->add_overview());
             }
             else
             {
                 for (const auto& snapshot : pick)
-                {
-                    auto overview = response.mutable_snapshot_overview()->add_overview();
-                    auto fundamentals = overview->mutable_fundamentals();
-
-                    overview->set_instance_name(name);
-                    populate_snapshot_fundamentals(vm.get_snapshot(snapshot), fundamentals);
-                }
+                    populate_snapshot_overview(name, vm.get_snapshot(snapshot),
+                                               response.mutable_snapshot_overview()->add_overview());
             }
         }
         catch (const NoSuchSnapshot& e)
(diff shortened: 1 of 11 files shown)


src/daemon/daemon.cpp
```diff
@@ -1627,59 +1627,33 @@ try // clang-format on
...
-            {
-                auto vm_image_info = config->image_hosts.back()->info_for_full_hash(vm_image.id);
-                original_release = vm_image_info.release_title.toStdString();
-            }
-            catch (const std::exception& e)
-            {
-                mpl::log(mpl::Level::warning, category, fmt::format("Cannot fetch image information: {}", e.what()));
-            }
-        }
+    auto populate_snapshot_fundamentals = [&](std::shared_ptr<const Snapshot> snapshot, auto& fundamentals) {
```
ricab, line 1638: This function is getting really big, with all the sub functions and all. I think this would be a good time to extract the individual parts into either private methods or unnamed namespace helpers.

ricab (changes requested):
Hey @sharder996, thanks for this work! This is only a superficial review, mainly to adjust the approach a little, but this seems to be working nicely.

src/daemon/daemon.cpp
```diff
@@ -1625,136 +1695,47 @@ try // clang-format on
...
-                original_release = vm_image_info.release_title.toStdString();
-            }
-            catch (const std::exception& e)
+            if (all)
             {
-                mpl::log(mpl::Level::warning, category, fmt::format("Cannot fetch image information: {}", e.what()));
+                populate_instance_info(vm, response.mutable_detailed_report()->add_details(),
+                                       request->no_runtime_information(), deleted, have_mounts);
+                for (const auto& snapshot : pick)
+                    vm.get_snapshot(snapshot); // verify validity of any snapshot name requested separately
```
ricab, line 1720: I don't think you need this here, since you are iterating over the individual snapshot arguments in the next few lines.

src/daemon/daemon.cpp
```diff
@@ -1625,136 +1695,47 @@ try // clang-format on
...
-        if (!vm_image.id.empty() && original_release.empty())
+        try
         {
-            try
-            {
-                auto vm_image_info = config->image_hosts.back()->info_for_full_hash(vm_image.id);
-                original_release = vm_image_info.release_title.toStdString();
-            }
-            catch (const std::exception& e)
+            if (all)
```
ricab, line 1715: This is now clearly a misnomer. Can we change it to `all_or_none`, here and in the `SnapshotPick`? Or something better you can think of.

src/daemon/daemon.cpp
```diff
@@ -3362,3 +3356,71 @@ void mp::Daemon::reply_msg(grpc::ServerReaderWriterInterface<Reply, Request>* se
...
+    instance_info->set_num_snapshots(vm.get_num_snapshots());
+    instance_info->set_image_release(original_release);
+    instance_info->set_id(vm_image.id);
+
+    auto vm_specs = vm_instance_specs[name];
+
+    auto mount_info = info->mutable_mount_info();
+    populate_mount_info(vm_specs.mounts, mount_info, have_mounts);
+
+    if (!runtime_info && mp::utils::is_running(present_state))
```
ricab, line 3397: `runtime_info` is the opposite of what this parameter currently means, isn't it? 

I agree it would make for sense to drop the negative, but then the logic needs inverting too, i.e. drop the `!` here and add it where you pass the argument from the grpc request.

src/daemon/daemon.cpp
```diff
@@ -1766,11 +1747,25 @@ try // clang-format on
...
             }
             else
+            {
                 for (const auto& snapshot : pick)
-                    get_snapshot_info(vm.get_snapshot(snapshot));
+                {
+                    auto overview = response.mutable_snapshot_overview()->add_overview();
+                    auto fundamentals = overview->mutable_fundamentals();
+
+                    overview->set_instance_name(name);
```
ricab, line 1765: Why didn't you include these lines in `populate_snapshot_fundamentals`, instead of repeating them?

ricab (commented):
Looking great! Thanks a lot for reworking this, it is much better :slightly_smiling_face: I have some more requests, but they should all be easy.

src/daemon/daemon.cpp
sharder996: I wanted to keep `populate_snapshot_fundamentals` aligned with the message type `SnapshotFundamentals` and not change anything outside of it. There is also [one situation](https://github.com/canonical/multipass/blob/ac8a96cb4d9819136f68f428151cdeb0f1359154/src/daemon/daemon.cpp#L1327) where the fundamentals are set and not `instance_name`.

ricab: I see. But then you can extract a helper employ here, similar to what happened already with `get_snapshot_info`.

ricab (approved):
Alright, LGTM!

bors r+

ricab:
duh, no need for bors on this one

bors cancel

ricab:
I'll just wait for CI before merging.

PR #3162, opened 2023-07-14
[snapshots] Fix daemon being blind to missing snapshots in `info` and `delete`

ricab:
Verify all snapshot arguments, even when the corresponding instance is also an argument. Do so while fetching each snapshot's info and before deleting any instance/snapshot.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1179,6 +1179,16 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
     return instance_snapshots_map;
 }
 
+void verify_snapshot_picks(const InstanceSelectionReport& report,
+                           const std::unordered_map<std::string, SnapshotPick>& snapshot_picks)
+{
+    for (const auto& selection : {report.deleted_selection, report.operative_selection})
+        for (const auto& vm_it : selection)
+            if (auto pick_it = snapshot_picks.find(vm_it->first); pick_it != snapshot_picks.end())
+                for (const auto& snapshot_name : pick_it->second.pick)
+                    vm_it->second->get_snapshot(snapshot_name); // throws if it doesn't exist
+}
+
 void add_aliases(google::protobuf::RepeatedPtrField<mp::FindReply_ImageInfo>* container, const std::string& remote_name,
                  const mp::VMImageInfo& info, const std::string& default_remote)
 {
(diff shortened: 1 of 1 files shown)


src/daemon/daemon.cpp
```diff
@@ -1742,25 +1757,26 @@ try // clang-format on
...
-                get_snapshot_info(snapshot);
-        }
-        else
+        const auto& it = instance_snapshots_map.find(name);
+        const auto& [pick, all] = it == instance_snapshots_map.end() ? SnapshotPick{{}, true} : it->second;
+
+        try
         {
-            for (const auto& snapshot : it->second)
+            if (all)
```
sharder996, line 1765: I think the if/else clauses need to be switched around, no?

src/daemon/daemon.cpp
```diff
@@ -1742,25 +1757,26 @@ try // clang-format on
...
-                try
-                {
-                    get_snapshot_info(vm.get_snapshot(snapshot));
-                }
-                catch (const NoSuchSnapshot& e)
-                {
-                    add_fmt_to(errors, e.what());
-                }
+                for (const auto& snapshot : pick)
+                    vm.get_snapshot(snapshot); // verify validity of any snapshot name requested separately
```
sharder996, line 1768: Why not use `verify_snapshot_picks()` in `info` as well?

sharder996 (changes requested):

src/daemon/daemon.cpp
ricab: Hmm, maybe I am overlooking something, but this seems correct to me. `all` is supposed to be true if we need to fetch all the snapshots for the instance called `name`. But before we do so, we need to verify that any arguments referring to individual snapshots of the same instance are valid, so that we complain on `multipass info --snapshot-overview x x.missing`.

src/daemon/daemon.cpp
ricab: Because we're fetching for an individual VM here, whereas `verify_snapshot_picks` checks all snapshot arguments. We could call it before calling this lambda, but we'd be calling `VM::get_snapshot` twice. 

Unlike `delete`, `info` does not change instances or snapshots, so it's fine if we detect a bad arg and abort the operation in the middle. I realize that fetching a bunch of info and then throwing it away is more wasteful than fetching each snapshot twice. However, the former only occurs when there is an error, whereas the latter would take place every time. I thought it preferable to optimize the successful path.

ricab:
Ugh why is there no CI at all for this PR?

src/daemon/daemon.cpp
sharder996: ah, I see. Alright, I think that's ok.

src/daemon/daemon.cpp
sharder996: ah, I see! Alright, I think that's ok.

sharder996 (approved):
Alright, looks good!

PR #3172, opened 2023-07-25
Refactor bridge prompt

luis4a0:
When adding a bridged interface to an existing instance, the user must be prompted to allow adding a network bridge disrupting connectivity. This is already implemented, but the code lies in the implementation of the `launch` command. This PR moves this code to the part common to all clients.

### DIFF
--- a/include/multipass/cli/prompters.h
+++ b/include/multipass/cli/prompters.h
@@ -18,7 +18,9 @@
 #include <multipass/disabled_copy_move.h>
 #include <multipass/terminal.h>
 
+#include <regex>
 #include <string>
+#include <vector>
 
 #ifndef MULTIPASS_CLI_PROMPTERS_H
 #define MULTIPASS_CLI_PROMPTERS_H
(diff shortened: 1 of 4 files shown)


include/multipass/cli/prompters.h
```diff
@@ -88,6 +90,26 @@ class NewPassphrasePrompter : public PassphrasePrompter
...
+
+private:
+    BridgePrompter() = default;
+
+    Terminal* term;
+
+    const std::regex yes{"y|yes", std::regex::icase | std::regex::optimize};
+    const std::regex no{"n|no", std::regex::icase | std::regex::optimize};
+    const std::regex later{"l|later", std::regex::icase | std::regex::optimize};
+    const std::regex show{"s|show", std::regex::icase | std::regex::optimize};
```
townsend2010, line 111: These need to be removed now due to https://github.com/canonical/multipass/pull/3173.

src/client/cli/cmd/launch.cpp
```diff
@@ -121,6 +107,7 @@ auto net_digest(const QString& options)
 
     return net;
 }
+
```
townsend2010, line 110: Why a newline here now?  `clang-format`?

townsend2010 (changes requested):
Hey @luis4a0!

Looks good overall, thanks!  Just a couple of inline comments.

include/multipass/cli/prompters.h
luis4a0: Right. Removed.

src/client/cli/cmd/launch.cpp
luis4a0: No, I don't think so. It just escaped my sight. Removed.

townsend2010 (changes requested):
Hey @luis4a0!

This is failing [this test](https://github.com/canonical/multipass-private/actions/runs/5739392718/job/15556926946#step:22:1190) on macOS CI.

luis4a0:
Hey @townsend2010! Death tests have indeed many subtleties. Fixed the failing test.

PR #3178, opened 2023-07-28
[iso] Complement error message on failure to open

ricab:
Add more information to the exception that is thrown when `QFile` fails to `open` to create a cloud-init ISO, in the hope that we can gain some insight into why this keeps failing in CI (e.g. https://github.com/canonical/multipass/actions/runs/5690481781/job/15423903919#step:13:101)


townsend2010 (approved):
Yep, makes sense.

bors merge

PR #3181, opened 2023-07-28
Fix relative paths in `launch`

### LINKED ISSUE #3180: Error using relative paths in launch with the CLI client
**Describe the bug**
When launching from a file using a relative path, like `multipass launch file://image.iso`, the path of the file is computed by the daemon, relative to its working folder.

**To Reproduce**
1. Download an image file an place it in the current folder, for instance `image.iso`.
2. Launch a new instance using that file, `multipass launch file://image.iso`. It will fail because the daemon will look for the file on its current folder.

**Expected behavior**
The absolute path must be computed with respect to the current folder.

**Logs**
Not needed.

**Additional info**
 - OS: Linux
- Multipass 1.12.1

luis4a0:
Fixes #3180.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -274,8 +274,15 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
     {
         auto remote_image_name = parser->positionalArguments().first();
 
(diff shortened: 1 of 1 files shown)


townsend2010:
Hrm, no bors...

bors merge

PR #3185, opened 2023-08-01
This adds HyperV to home edition

0m364:
this makes multipass run much more smoothly on win 11 home than using a 2nd 3rd party


townsend2010:
Hi @0m364,

Thank you for trying to improve the Multipass experience on Windows Home, but unfortunately, this is quite big grey area when it comes to this.  Microsoft clearly states that Hyper-V is supported on Windows Pro and higher and I don't think it would be in Canonical's best interest to have a script that circumvent's this.  We politely decline your contribution, but thank you again.

PR #3192, opened 2023-08-16
Flutter tray icon

andrei-toterman:
This PR adds a system tray icon menu to the app, similar to the existing Qt one.
Differences:
- a 'Toggle window' button which shows/hides the main app window
- removed 'Autostart at login' button, as this will be a setting in the main app window
- no 'Open shell' functionality yet, but will be added when we actually integrate a terminal into the app

Note: on KDE, there might be issues with submenu items not appearing. This is of no fault of our own, but it seems to be a bug in either `libayatana-appindicator` or `libdbusmenu`


andrei-toterman:

PR #3193, opened 2023-08-16
Add `local.instance.bridged` setting.

luis4a0:
This PR implements the `set` interface to add a bridged network to an existing instance.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -457,6 +457,19 @@ auto try_mem_size(const std::string& val) -> std::optional<mp::MemorySize>
     }
 }
 
+std::string get_bridged_interface_name()
+{
+    const auto bridged_id = MP_SETTINGS.get(mp::bridged_interface_key);
+
+    if (bridged_id == "")
+    {
+        throw std::runtime_error(fmt::format("You have to `multipass set {}=<name>` to use the \"bridged\" shortcut.",
+                                             mp::bridged_interface_key));
+    }
+
+    return bridged_id.toStdString();
+}
+
 std::vector<mp::NetworkInterface> validate_extra_interfaces(const mp::LaunchRequest* request,
                                                             const mp::VirtualMachineFactory& factory,
                                                             std::vector<std::string>& nets_need_bridging,
@@ -1170,10 +1183,11 @@ register_instance_mod(std::unordered_map<std::string, mp::VMSpecs>& vm_instance_
                       std::unordered_map<std::string, mp::VirtualMachine::ShPtr>& vm_instances,
                       const std::unordered_map<std::string, mp::VirtualMachine::ShPtr>& deleted_instances,
                       const std::unordered_set<std::string>& preparing_instances,
-                      std::function<void()> instance_persister)
+                      std::function<void()> instance_persister, std::function<std::string()> bridged_interface)
 {
     return MP_SETTINGS.register_handler(std::make_unique<mp::InstanceSettingsHandler>(
-        vm_instance_specs, vm_instances, deleted_instances, preparing_instances, std::move(instance_persister)));
+        vm_instance_specs, vm_instances, deleted_instances, preparing_instances, std::move(instance_persister),
+        std::move(bridged_interface)));
 }
 
 } // namespace
@@ -1184,8 +1198,9 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
           mp::utils::backend_directory_path(config->data_directory, config->factory->get_backend_directory_name()),
           mp::utils::backend_directory_path(config->cache_directory, config->factory->get_backend_directory_name()))},
       daemon_rpc{config->server_address, *config->cert_provider, config->client_cert_store.get()},
-      instance_mod_handler{register_instance_mod(vm_instance_specs, operative_instances, deleted_instances,
-                                                 preparing_instances, [this] { persist_instances(); })}
+      instance_mod_handler{register_instance_mod(
+          vm_instance_specs, operative_instances, deleted_instances, preparing_instances,
+          [this] { persist_instances(); }, get_bridged_interface_name)}
 {
     connect_rpc(daemon_rpc, *this);
     std::vector<std::string> invalid_specs;
--- a/src/daemon/instance_settings_handler.cpp
+++ b/src/daemon/instance_settings_handler.cpp
@@ -19,6 +19,7 @@
 
 #include <multipass/constants.h>
 #include <multipass/exceptions/invalid_memory_size_exception.h>
+#include <multipass/settings/bool_setting_spec.h>
 
 #include <QRegularExpression>
 #include <QStringList>
@@ -30,6 +31,7 @@ namespace
 constexpr auto cpus_suffix = "cpus";
 constexpr auto mem_suffix = "memory";
 constexpr auto disk_suffix = "disk";
+constexpr auto bridged_suffix = "bridged";
 
 enum class Operation
 {
@@ -46,7 +48,7 @@ QRegularExpression make_key_regex()
 {
     const auto instance_pattern = QStringLiteral("(?<instance>.+)");
     const auto prop_template = QStringLiteral("(?<property>%1)");
-    const auto either_prop = QStringList{cpus_suffix, mem_suffix, disk_suffix}.join("|");
+    const auto either_prop = QStringList{cpus_suffix, mem_suffix, disk_suffix, bridged_suffix}.join("|");
     const auto prop_pattern = prop_template.arg(either_prop);
 
     const auto key_template = QStringLiteral(R"(%1\.%2\.%3)");
@@ -153,6 +155,27 @@ void update_disk(const QString& key, const QString& val, mp::VirtualMachine& ins
     }
 }
 
+bool is_bridged(const mp::VMSpecs& spec)
+{
+    return std::any_of(spec.extra_interfaces.cbegin(), spec.extra_interfaces.cend(),
+                       [](const auto& network) -> bool { return network.auto_mode; });
+}
+
+void update_bridged(const QString& key, const QString& val, mp::VirtualMachine& instance, mp::VMSpecs& spec)
+{
+    auto bridged = mp::BoolSettingSpec{key, "false"}.interpret(val) == "true";
+
+    if (!bridged && is_bridged(spec))
+    {
+        throw mp::InvalidSettingException{key, val, "Bridged interface cannot be removed"};
+    }
+
+    if (bridged)
+    {
+        // TODO: add the interface
+    }
+}
+
 } // namespace
 
 mp::InstanceSettingsException::InstanceSettingsException(const std::string& reason, const std::string& instance,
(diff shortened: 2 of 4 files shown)


src/daemon/instance_settings_handler.cpp
```diff
@@ -153,6 +154,48 @@ void update_disk(const QString& key, const QString& val, mp::VirtualMachine& ins
     }
 }
 
+bool to_bool(const QString& key, const QString& val)
```
ricab, line 157: This is implemented in `BoolSettingSpec`. Instead of copy+pasting from there, why don't you instantiate a helper object and use that?

src/daemon/instance_settings_handler.cpp
```diff
@@ -216,13 +261,20 @@ void mp::InstanceSettingsHandler::set(const QString& key, const QString& val)
         update_cpus(key, val, instance, spec);
     else
     {
-        auto size = get_memory_size(key, val);
-        if (property == mem_suffix)
-            update_mem(key, val, instance, spec, size);
+        if (property == bridged_suffix)
```
ricab, line 264: `else if`? To avoid one extra nesting level.

ricab (changes requested):
Hey @luis4a0, I was curious and I just had a quick look. The code looks good, I like that you're splitting things into small self contained units :+1: I spotted only a couple of wrinkles in this very superficial pass.

src/daemon/instance_settings_handler.cpp
luis4a0: Hey @ricab, thanks for your quick review!

I thought in avoid copy+pasting, but I opted for not a `QString` and comparing against it. Just true or false. Do you still think it's OK to reuse the existing implementation?

src/daemon/instance_settings_handler.cpp
luis4a0: Ok, changed. I was hesitating whether to mix an inline `else if` with a structured `else { stmt; if ... }`. But right, we save four columns the way you mention.

src/daemon/instance_settings_handler.cpp
ricab: I see what you mean: you want to return a `bool` and compare against that in the code, right? But you can still achieve that with an extra step. For example, just something like this to replace the contents of `to_bool`: `return BoolSettingSpec{key, "false"}.interpret(val) == "true";`

src/daemon/instance_settings_handler.cpp
ricab: It's not for saving spaces, it's that it makes it easier to understand that they are alternatives at the same level, with no further shared logic (unlike the following 2 properties, which do share the conversion of the memory size).

src/daemon/instance_settings_handler.cpp
luis4a0: I wanted to avoid that string comparison, and throw directly instead of returning a value before throwing. But well, it is true that the code looks much better the way you propose. Since time is not critical here, I changed it. Thanks!

tests/test_instance_settings_handler.cpp
```diff
@@ -82,7 +82,9 @@ struct TestInstanceSettingsHandler : public Test
     std::unordered_map<std::string, mp::VirtualMachine::ShPtr> deleted_vms;
     std::unordered_set<std::string> preparing_vms;
     bool fake_persister_called = false;
-    inline static constexpr auto properties = std::array{"cpus", "disk", "memory"};
+    inline static constexpr auto numeric_properties = std::array{"cpus", "disk", "memory"};
+    inline static constexpr auto boolean_properties = std::array{"bridged"};
+    inline static constexpr auto properties = std::array{"cpus", "disk", "memory", "bridged"};
 };
```
georgeliao, line 88: I know it stems from the original code, but maybe change the copy initialization to direct initialization to make the code cleaner. 
```
    inline static constexpr std::array numeric_properties{"cpus", "disk", "memory"};
    inline static constexpr std::array boolean_properties{"bridged"};
    inline static constexpr std::array properties{"cpus", "disk", "memory", "bridged"};
```

src/daemon/instance_settings_handler.cpp
```diff
@@ -153,6 +155,32 @@ void update_disk(const QString& key, const QString& val, mp::VirtualMachine& ins
...
 
+bool is_bridged(const mp::VMSpecs& spec)
+{
+    for (const auto& network : spec.extra_interfaces)
+    {
+        if (network.auto_mode)
+            return true;
+    }
+
+    return false;
```
georgeliao, line 166: std algorithm based code can be another alternative, the code is shorter, the readability probably is the same
```
    return std::any_of(spec.extra_interfaces.cbegin(), spec.extra_interfaces.cend(),
                       [](const auto& network) -> bool { return network.auto_mode; });
```

src/daemon/instance_settings_handler.cpp
```diff
@@ -153,6 +155,32 @@ void update_disk(const QString& key, const QString& val, mp::VirtualMachine& ins
...
+}
+
+void update_bridged(const QString& key, const QString& val, mp::VirtualMachine& instance, mp::VMSpecs& spec)
+{
+    auto bridged = mp::BoolSettingSpec{key, "false"}.interpret(val) == "true";
+
+    if (!bridged && is_bridged(spec))
+    {
```
georgeliao, line 174: Is it possible to elaborate the role between bridged boolean (it looks like is checking user's input) and is_bridged(spec) (looks like checking if the spec is bridged)?

tests/test_instance_settings_handler.cpp
```diff
@@ -193,6 +195,26 @@ TEST_F(TestInstanceSettingsHandler, getReturnsMemorySizesInHumanReadableFormat)
...
+{
+    const auto bridged = GetParam();
+
+    constexpr auto target_instance_name = "lemmy";
+    specs.insert({{"mikkey", {}}, {"phil", {}}, {target_instance_name, {}}});
+
+    // An extra interface in auto mode will result in bridging.
+    specs[target_instance_name].extra_interfaces = {{"id", "52:54:00:12:34:56", bridged}};
+
+    auto got = make_handler().get(make_key(target_instance_name, "bridged"));
```
georgeliao, line 212: `const auto got` is better

src/daemon/instance_settings_handler.cpp
luis4a0: Oh, sorry if it is unclear. The idea was that adding a boolean to the VM specs data structure would tell us quickly if the VM has a bridged interface. But the boolean can be inferred from the `extra_interfaces` vector. If some of the elements of this vector had the `auto_mode` set to true, it means the VM has a bridged interface. Thus, we removed the boolean. This approach lets us to have many bridged interfaces at the same time.

src/daemon/instance_settings_handler.cpp
luis4a0: Looks nice, thanks! Changed.

tests/test_instance_settings_handler.cpp
luis4a0: Right, looks better that way. Changed.

tests/test_instance_settings_handler.cpp
luis4a0: Right, better. Changed.

src/daemon/instance_settings_handler.cpp
luis4a0: Oh, sorry if it wasn't clear. The `bridged` boolean is the value given by the user. The `is_bridged()` function returns a boolean stating if the VM has a bridged interface.

luis4a0:
Hey @georgeliao, thanks for the review! I addressed your concerns, let me know if everything looks better now.

src/daemon/instance_settings_handler.cpp
```diff
@@ -153,6 +155,27 @@ void update_disk(const QString& key, const QString& val, mp::VirtualMachine& ins
     }
 }
 
+bool is_bridged(const mp::VMSpecs& spec)
+{
+    return std::any_of(spec.extra_interfaces.cbegin(), spec.extra_interfaces.cend(),
+                       [](const auto& network) -> bool { return network.auto_mode; });
```
ricab, line 161: Hey @luis4a0, I think for our purposes we'd better consider the instance bridged if and only if it has an extra_interface corresponding to the network that is currently set for `local.bridged-network`. That would make it symmetric to `launch --bridged` and allow adding multiple interfaces.

src/daemon/instance_settings_handler.cpp
georgeliao: I see, thanks.

src/daemon/instance_settings_handler.cpp
luis4a0: Hey @ricab, thanks for your comment. You're right, I didn't think on that. I changed the code to reflect this behavior.

src/daemon/daemon.cpp
```diff
@@ -457,6 +457,19 @@ auto try_mem_size(const std::string& val) -> std::optional<mp::MemorySize>
     }
 }
 
+std::string get_bridged_interface_name()
+{
+    const auto bridged_id = MP_SETTINGS.get(mp::bridged_interface_key);
+
+    if (bridged_id == "")
+    {
+        throw std::runtime_error(fmt::format("You have to `multipass set {}=<name>` to use the `--bridged` shortcut.",
```
ricab, line 466: `--bridged` is not applicable in the case of settings, so the message needs to be generalized. Perhaps just:

```suggestion
        throw std::runtime_error(fmt::format("You have to `multipass set {}=<name>` to use the "bridged" shortcut.",
```

src/daemon/instance_settings_handler.cpp
ricab: Great, thanks!

src/daemon/daemon.cpp
```diff
@@ -1184,8 +1198,9 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
           mp::utils::backend_directory_path(config->data_directory, config->factory->get_backend_directory_name()),
           mp::utils::backend_directory_path(config->cache_directory, config->factory->get_backend_directory_name()))},
       daemon_rpc{config->server_address, *config->cert_provider, config->client_cert_store.get()},
-      instance_mod_handler{register_instance_mod(vm_instance_specs, operative_instances, deleted_instances,
-                                                 preparing_instances, [this] { persist_instances(); })}
+      instance_mod_handler{register_instance_mod(
+          vm_instance_specs, operative_instances, deleted_instances, preparing_instances,
+          [this] { persist_instances(); }, [] { return get_bridged_interface_name(); })}
```
ricab, line 1203: Do you really need the lambda here? Unlike to persist instances, there is no closure for the bridge, so you should be able to pass the function directly.

georgeliao: Yep, nice catch. `[this] { persist_instances(); }, get_bridged_interface_name)}` will do it.

luis4a0: Right, changed.

src/daemon/daemon.cpp
luis4a0: Right, changed.

src/daemon/daemon.cpp
ricab: :+1:

luis4a0:
Hey @ricab @georgeliao, thanks for the review! Let me note that all the uncovered code corresponds to the `get_bridged_interface()` function in `daemon.cpp`. It will be tested in the next PR, for doing it here would add lots of unnecessary complexity.

PR #3197, opened 2023-08-18
Add network interfaces on LXD backend

luis4a0:
This PR adds the necessary bits to implement adding network interfaces on LXD backend.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -251,7 +251,8 @@ std::vector<mp::NetworkInterface> read_extra_interfaces(const QJsonObject& recor
         {
             auto id = entry.toObject()["id"].toString().toStdString();
             auto mac_address = entry.toObject()["mac_address"].toString().toStdString();
-            if (!mpu::valid_mac_address(mac_address))
+            // Allow empty addresses (for nonconfigured interfaces).
+            if (!mac_address.empty() && !mpu::valid_mac_address(mac_address))
             {
                 throw std::runtime_error(fmt::format("Invalid MAC address {}", mac_address));
             }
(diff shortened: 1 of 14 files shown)


src/daemon/daemon.cpp
```diff
@@ -2859,6 +2872,33 @@ mp::MountHandler::UPtr mp::Daemon::make_mount(VirtualMachine* vm, const std::str
                : vm->make_native_mount_handler(config->ssh_key_provider.get(), target, mount);
 }
 
+void mp::Daemon::configure_new_interfaces(mp::VirtualMachine& vm, mp::VMSpecs& specs)
+{
+    std::vector<size_t> new_interfaces;
+
```
georgeliao, line 2878: It is kind weird that it you have to use empty mac_address as a flag and filter them out and generate mac address. I guess this is because generate_unused_mac_address only can be done in the daemon class and it holds allocated_mac_addrs. Ideally, these things should be utilities and decouple from the daemon so vm_instance_specs can access it easily. But anyway, that will result in a big refactoring. For now, the current code is ok.

georgeliao:
@luis4a0 it looks good in general. Just left a comment which does not require actual action.

src/daemon/daemon.cpp
luis4a0: Hey @georgeliao, thanks! I perfectly understand your point, because I thought in adding a flag indicating that a new MAC address needed to be created. However, I opted for not polluting more the data structures. On the other hand, I think having an empty MAC address is not bad in the sense that it will unequivocally imply that the interface was not configured and this must be done before going on.

georgeliao (approved):

georgeliao:
bors merge

PR #3204, opened 2023-08-23
Show image with `devel` alias regardless of its `supported` status

### LINKED ISSUE #3202: `multipass find` should show `daily:devel` by default
Right now, `daily:devel` usually has `supported: false` set in Simplestreams and as such, it is not shown by default in `multipass find` output.  Instead, we should key on the `devel` alias and show it by default regardless of how `supported` is set.

andrei-toterman:
fix #3202


townsend2010 (approved):
Yup, thanks for this!

bors merge

andrei-toterman:
bors retry

PR #3215, opened 2023-08-29
[build] Use vcpkg to build modules

townsend2010:
Use vcpkg to build and cache the grpc and poco libraries.  This will decrease build times in CI and locally as well.


luis4a0 (approved):
Hey @townsend2010! This is a huge piece of work, well done! It was hard to review, because there were many changes. I finally decided to check all the commits, one by one, to better understand what every change is doing. I don't have specific comments, I checked the code is good and that it compiles and works with Snapcraft and as well by hand, with gcc and with clang. I am a bit concerned by the amount of tweaks which must be done as patches, but I know this can't be done differently, so I am OK with them. Thanks for this!

luis4a0:
Hey @townsend2010! I think we need to address one comment on the private side. And it is possible that its fix go in this public part, so I'm not merging for the time being. Thanks!

luis4a0:
Hey @townsend2010! I see you force-pushed some changes, but I can't find a way of realizing what changed. In any case, I'll test it functionally. Thanks for this hard work!

townsend2010:
Hi @luis4a0!

Thanks for reviewing this!  The force push basically fixed some code duplication I saw in the top level `CMakeLists.txt` and a rebase on `main`.  The new change is seen at https://github.com/canonical/multipass/pull/3215/commits/185a920ceb408ce13eaac2ac6848d4f004ac9ff5.  This should address the issue you found on the Windows debug build.

luis4a0:
Thanks! That seems to fix the issue, I'll do some functional testing and we'll be done.

townsend2010:
Also,
> I am a bit concerned by the amount of tweaks which must be done as patches

These are basically taken from the upstream [grpc](https://github.com/microsoft/vcpkg/tree/master/ports/grpc) and [poco](https://github.com/microsoft/vcpkg/tree/master/ports/poco) vcpkg port with the main difference being to **only** build the libraries we need.

luis4a0:
> These are basically taken from the upstream [grpc](https://github.com/microsoft/vcpkg/tree/master/ports/grpc) and [poco](https://github.com/microsoft/vcpkg/tree/master/ports/poco) vcpkg ports with the main difference being to only build the libraries we need.

Perfect, that's what I needed to know to stay calm ;)

luis4a0 (approved):
Hey @townsend2010! I could finally test everything, on all the three platforms, with the different build types and now I am OK.Will wait for CI to merge. Thanks for this hell of a PR!

PR #3222, opened 2023-09-06
Remove qt tray icon

andrei-toterman:
This PR removes everything related to the Qt tray icon application, as it will be superseded by a Flutter implementation.

It also removes the code related to client-specific certificate directories in favor of using only the common certificate directory.


andrei-toterman:

PR #3237, opened 2023-09-25
[snapshots] use user locale in timestamp output

sharder996:


src/client/cli/formatter/format_utils.cpp
```diff
@@ -110,3 +110,19 @@ void mp::format::filter_aliases(google::protobuf::RepeatedPtrField<multipass::Fi
...
+{
+}
+
+std::string mp::FormatUtils::convert_to_localtime(const google::protobuf::Timestamp& timestamp) const
+{
+    std::ostringstream oss;
+    std::time_t t = google::protobuf::util::TimeUtil::TimestampToTimeT(timestamp);
+
+    // ISO 8601: YYYY-MM-DDThh:mm:ss±hhmm
+    oss << std::put_time(std::localtime(&t), "%FT%T%z");
```
sharder996, line 125: Currently have this set up to print out the UTC offset. Excluding `±hhmm` looks a bit more presentable, but then means that the datetime is ambiguous on whether it defines UTC time or localtime. See [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)).

ricab:
Thanks for this @sharder996. I will enqueue this with a little lower priority to focus on PRs that were already on flight. But hopefully we can still get it into the MVP.

ricab:
Hey @sharder, could you please rebase the relevant commits here? Thank you!

ricab (commented):
Hey @sharder996, thanks for this PR! It looks good overall, both structurally and in detail. However, you are using local time, but not a locale yet. The locale is important in this instance mainly to avoid ambiguity in the date (is 2022-11-02 the same as Nov 2nd or Feb 11th?); to a lesser extent, in the time too (is 1:15 am or pm?). 

In any case, this is certainly valuable as is. If it turns out to be difficult to use the locale, we can skip that and accept. But please have a look at my inline comment.

ricab (approved):
Thanks @sharder996! This doesn't work within a snap, but I think we can get it in and deal with that separately. I have only a tiny request inline, then we can merge.

townsend2010:
bors ping

townsend2010:
Ok, CI is all green as are [integration tests](https://github.com/canonical/multipass/actions/runs/6885616587).  Merging...

PR #3238, opened 2023-09-26
[daemon] Configure new bridged interfaces at boot.

luis4a0:


src/daemon/daemon.h
```diff
@@ -181,6 +181,7 @@ public slots:
     QFuture<void> image_update_future;
     SettingsHandler* instance_mod_handler;
     std::unordered_map<std::string, std::unordered_map<std::string, MountHandler::UPtr>> mounts;
+    std::unordered_map<std::string, std::vector<std::string>> run_at_boot;
```
georgeliao, line 184: So basically, the role of run_at_boot is holding the commands need to be executed at boot. I am surprised that adding network interface is the first one which requires this. But anyway, it is a good framework for future as well.

src/daemon/daemon.cpp
```diff
@@ -3025,6 +3048,27 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
         for (const auto& name : vms)
         {
             async_running_futures.erase(name);
+
+            try
+            {
+                const auto commands = run_at_boot.at(name);
+
+                auto vm = operative_instances[name];
+                auto vm_specs = vm_instance_specs[name];
```
georgeliao, line 3059: They can be `const auto vm` and `const auto vm_specs`

src/daemon/daemon.h
luis4a0: Indeed!

src/daemon/daemon.cpp
```diff
@@ -3025,6 +3048,27 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
...
+
+                for (const auto& command : commands)
+                {
+                    mpu::run_in_ssh_session(session, command);
+                }
+
+                run_at_boot.erase(name);
+            }
+            catch (const std::out_of_range&)
+            {
```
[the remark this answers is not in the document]
luis4a0: Yes, you're absolutely right. I changed the code as you propose, while waiting for C++20's `contains()`.

src/daemon/daemon.cpp
luis4a0: Ok, changed.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
georgeliao: Yep, `contains` method would be nice to have.

src/client/cli/cmd/start.cpp
```diff
@@ -59,8 +59,17 @@ mp::ReturnCode cmd::Start::run(mp::ArgParser* parser)
 
     auto on_success = [&spinner, this](mp::StartReply& reply) {
         spinner.stop();
+
         if (term->is_live() && update_available(reply.update_info()))
+        {
             cout << update_notice(reply.update_info());
+        }
+
+        if (reply.log_line() != "")
```
georgeliao, line 68: maybe change `reply.log_line() != ""` to `!reply.log_line().empty()`.

src/daemon/daemon.cpp
```diff
@@ -2996,7 +3019,7 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
                                      const std::vector<std::string>& vms, const std::chrono::seconds& timeout,
                                      std::promise<grpc::Status>* status_promise, const std::string& start_errors)
 {
-    fmt::memory_buffer errors;
+    fmt::memory_buffer errors, warnings;
     fmt::format_to(std::back_inserter(errors), "{}", start_errors);
```
georgeliao, line 3023: I know this also stems from the original code. But maybe move the `errors` and `warnings` variables closer to the code blocks that use them, so we can localize the variables better. 

 `fmt::memory_buffer warnings;` can be moved to the line before the code block. 
```
{
std::lock_guard<decltype(start_mutex)> lock{start_mutex};
```
and 
```
    fmt::memory_buffer errors;
    fmt::format_to(std::back_inserter(errors), "{}", start_errors);
```
can be moved to the line before 
`    for (const auto& future : start_synchronizer.futures())
`

src/daemon/daemon.cpp
```diff
@@ -3034,10 +3095,23 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
 
     if (server && std::is_same<Reply, StartReply>::value)
     {
+        bool write_reply{false};
```
georgeliao, line 3098: I kind see why you added the `write_reply` boolean variable. However, I would suggest just writing empty message into the `server` since the emptyness is handled automatically. That results in cleaner code

```
            Reply reply;
            if (config->update_prompt->is_time_to_show())
            {
                config->update_prompt->populate(reply.mutable_update_info());
            }

            if (warnings.size() > 0)
            {
                reply.set_log_line(fmt::to_string(warnings));
            }
            server->Write(reply);
```

georgeliao:
@luis4a0 , thanks for the first round iteration. The code looks good in general. I have a few other minor comments. I think we can merge after that.

src/client/cli/cmd/start.cpp
luis4a0: Right, changed.

src/daemon/daemon.cpp
luis4a0: Right, looks better this way. Changed.

georgeliao (approved):

PR #3242, opened 2023-09-29
clang-format improvement: remove bin pack arguments/parameters.

georgeliao:
suggestive clang-format improvement:

The function arguments and parameters are always bin-packed in the current clang-format style which can sometimes be hard to read. The improved clang-format formats arguments and parameters in a way that it either does not break the line because the line length is still within the column limit or it breaks each argument/parameter into a separate line. 

This PR contains the clang-format file code change and some sample files in the new format. If this PR is approved, then the sample files will be reverted and only the clang-format change will be merged.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -5,6 +5,11 @@ BasedOnStyle: LLVM
 Language: Cpp
 IndentWidth: 4
 BreakBeforeBraces: Custom
+BinPackArguments: false
+BinPackParameters: false
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+PenaltyReturnTypeOnItsOwnLine: 120
 BraceWrapping:
   AfterCaseLabel: true
   AfterClass: true


sharder996:
Personally, I don't have a strong preference either way, but there are a number of different formatting changes that have happened over time in the multipass repo. If we do decide as a team that this is an improvement, I would suggest only applying the change to new functions from this point on rather than arbitrarily editing some files and not others. Otherwise, making the switchover to a new style globally (or even whole files) is going to cause some annoying merge conflicts in feature branches, etc. Additionally, using a bot to do the formatting would make things like `git blame` easier to parse.

ricab (approved):
Personally, I am +1 on this (with the caveat of only merging .clang-format). I am a little unsure of the effects of `PenaltyReturnTypeOnItsOwnLine` but we can always tune further.

georgeliao:
@ricab , LLVM [default value](https://github.com/llvm/llvm-project/blob/53179129edbff13644b7c46336773fb8740899c9/clang/lib/Format/Format.cpp#L1578) of `PenaltyReturnTypeOnItsOwnLine` is 60, which makes many short return types break. Readability wise, I would say most of the return type should not break unless it is a very long name like 
```
template <typename Container, typename UnaryOperation>
std::vector<std::invoke_result_t<std::decay_t<UnaryOperation>, typename Container::value_type>>
parallel_transform(const Container& input_container, UnaryOperation&& unary_op)
{
'''
}
``` 
which pushes the function name to the edge of the panel. In this case, breaking helps. The bigger value of `PenaltyReturnTypeOnItsOwnLine` gets us closer to that break on the right type of return type behavior. 
Of course, we can always fine-tune it later.

townsend2010 (dismissed):
Yep, I think this looks much better.  I'm +1 with *just* the `.clang-format` change.

PR #3256, opened 2023-10-16
[snapshots] Unit tests

ricab:


ricab:
Can't change base branch to fix-snapshot-restore-rollback ATTOW, but no merge until rebased on main!

ricab:
Oh, GH doesn't let me reopen :thinking:

townsend2010:
Hey @ricab!

In order to be able to reopen this PR, the `snapshots-unit-tests` branch SHA needs to match what this PR thinks is the last commit, so in this case, the SHA would be `f6bfb48`.

ricab:
Hmm I see, thanks! Well I have since moved on from that so I guess I'll live with the replacement.

PR #3259, opened 2023-10-17
[snapshots] Implement `info --snapshots`

ricab:


sharder996 (changes requested):
If I execute `multipass info foo foo.snapshot1 --snapshots`, `snapshot1` will be included twice in the result.

ricab:
Good catch @sharder996. I hope it's working now.

sharder996 (approved):
Alright, looking good now!

PR #3260, opened 2023-10-18
Adapt output of `info` when no instances/snapshots are found

ricab:


src/rpc/multipass.proto
```diff
@@ -248,6 +248,7 @@ message DetailedInfoItem {
 
 message InfoReply {
     repeated DetailedInfoItem details = 1;
+    bool snapshots = 2; // useful to determine what entity (instance/snapshot) was absent when details are empty
```
ricab, line 251: @sharder996, a little disclaimer: I am not terribly happy about this. Unfortunately, alternatives I could think of were even worse: add a stub details entry with empty name; change the signature of `Formatter::format(const InfoReply& info)` to receive that boolean, but only use it in the `TableFormatter`. 

It's a close call between that last option and the one I went with. WDYT? Any better options? I think we had this problem already, didn't we? Do you remember what it was or what we chose?

sharder996: I really don't like the last option you mentioned, so I would probably go with what you did as well.

Yes, we've run into this problem before. You may be thinking of how we handle snapshots/instances in `ListReply`, which can be handled a bit differently because we don't mix instances and snapshots. Where this has been encountered before was in `FindReply` when adapting messages to queries for only blueprints or only images. I did something very similar there as well.

https://github.com/canonical/multipass/blob/ef40bb593eb7db4e76969be303f54733598d2aa5/src/rpc/multipass.proto#L152-L153

ricab: OK, great, thanks for your input. Let me know if you think we should change the name.

sharder996 (approved):

PR #3266, opened 2023-10-20
Add bridged interfaces in macOS QEMU

luis4a0:
This branch implements the Linux platform code for adding bridged network interfaces on QEMU. Since it will be available first in macOS, here the implementation just throws.


georgeliao:
@luis4a0 , thanks for the nice work, it looks good. I do not have comments which require action. Just one comment related to the next step of the development.

georgeliao (approved):

PR #3270, opened 2023-10-24
Replace Qt with std in critical parts of sftp_server

andrei-toterman:


townsend2010 (changes requested):
Hey @andrei-toterman!

Thanks for this!  Unfortunately, it is not working for me.  If I create a mount, `shell` into the instance, `cd` to the mounted directory, I get the following:
```
$ git clone https://github.com/canonical/multipass.git
Cloning into 'multipass'...
error: Unable to create '/home/ubuntu/test/multipass/.git/HEAD.lock': Operation not permitted
```

Also, I haven't full reviewed, but I did leave one comment below.

src/sshfs_mount/sftp_server.cpp
```diff
@@ -1071,3 +1064,9 @@ int mp::SftpServer::handle_extended(sftp_client_message msg)
 
     return reply_ok(msg);
 }
+
+template <typename T>
+T* multipass::SftpServer::get_handle(sftp_client_message msg)
+{
+    return reinterpret_cast<T*>(sftp_handle(msg->sftp, msg->handle));
```
georgeliao, line 1071: Maybe using `static_cast` which has some compile time checks instead of `reinterpret_cast`.

andrei-toterman: Yup, makes sense.

include/multipass/recursive_dir_iterator.h
```diff
@@ -249,6 +250,48 @@ class RecursiveDirIterator
...
+        return current = *iter++;
+    }
+
+    virtual ~DirIterator() = default;
+
+private:
+    std::optional<fs::path> self;
+    std::optional<fs::path> parent;
+    fs::directory_iterator iter;
+    DirectoryEntry current;
```
georgeliao, line 293: Nice wrapper utility class, besides that, I have a few small comments.
1. About the data member `current`, it holds the same information as `*iter`. I think you did this double bookkeeping because you need to return `DirectoryEntry` reference. I am wondering if it is gotta be better if you return by value and remove all the `current` related code.
2. There are some implicit constructions from `fs::directory_entry` to `DirectoryEntry`, maybe it is better to change the constructor `DirectoryEntry(const fs::directory_entry& _entry)` to explicit and explicitly convert these.

src/sshfs_mount/sftp_server.h
```diff
@@ -76,13 +78,16 @@ class SftpServer
...
+    T* get_handle(sftp_client_message msg);
+
     SSHSession ssh_session;
     SSHFSProcUptr sshfs_process;
     SftpSessionUptr sftp_server_session;
     const std::string source_path;
     const std::string target_path;
-    std::unordered_map<void*, std::unique_ptr<QFileInfoList>> open_dir_handles;
-    std::unordered_map<void*, std::unique_ptr<QFile>> open_file_handles;
+    std::unordered_map<void*, std::unique_ptr<NamedFd, void (*)(NamedFd*)>> open_file_handles;
```
georgeliao, line 89: Maybe change `void (*)(NamedFd*)` from `std::function<void(NamedFd*)`, so we have a more readable functor signature.

andrei-toterman: Good call. At first I avoided `std::function` because I thought it always involves a dynamic allocation, but after reading the docs better, I understand that in the case of free function pointers, small object optimization is guaranteed and no dynamic allocation takes place. So I'll change it.

include/multipass/recursive_dir_iterator.h
andrei-toterman: You are right, `current` is necessary for returning a `DirectoryEntry` reference. But the reason it returns a reference instead of a value is so that it can return a `MockDirectoryEntry` reference when writing unit tests.
As for the explicit constructors, sure, I will convert them.

src/utils/file_ops.cpp
```diff
@@ -147,6 +150,34 @@ bool mp::FileOps::commit(QSaveFile& file) const
...
 
+std::unique_ptr<mp::NamedFd, void (*)(mp::NamedFd*)> mp::FileOps::open_fd(const fs::path& path,
+                                                                          int flags,
+                                                                          int perms) const
+{
+    const auto fd = ::open(path.string().c_str(), flags | O_BINARY, perms);
+    const auto named_fd = new NamedFd{path, fd};
+    const auto deleter = [](NamedFd* named_fd) {
+        if (named_fd->second != -1)
+            ::close(named_fd->second);
```
georgeliao, line 161: I think it is the responsibility of the custom deleter to delete the memory explicitly. So a line `delete named_fd;` will have to be added after the if branch.         
```
        if (named_fd->second != -1)
            ::close(named_fd->second);
        delete named_fd;
```
Besides that, maybe move the `new NamedFd{path, fd};` into the return statement that makes `return {new NamedFd{path, fd}, deleter};
`  and remove the line `const auto named_fd = new NamedFd{path, fd};`.

andrei-toterman: Given that I made this mistake regarding the destructor, and the `std::fucntion` over simple function pointer comment, do you think that it would be worth it to change `NamedFd` from a simple `std::pair` to a proper struct that has its own destructor? That way, we can use a simple `std::unique_ptr<NamedFd>` and not worry about providing a correct deleter. Initially I used `std::pair` because I wanted to keep things simple, but maybe they are not as simple as I thought.

georgeliao: Actually, that is an even better idea. Having a proper struct and proper RAII and avoiding all the custom deleter hassle.

src/sshfs_mount/sftp_server.cpp
```diff
@@ -508,24 +470,26 @@ int mp::SftpServer::handle_mkdir(sftp_client_message msg)
...
     {
-        mpl::log(mpl::Level::trace, category, fmt::format("{}: mkdir failed for \'{}\'", __FUNCTION__, filename));
+        mpl::log(mpl::Level::trace, category, fmt::format("{}: mkdir failed for '{}'", __FUNCTION__, filename));
         return reply_failure(msg);
     }
 
-    QFile file(filename);
-    if (!MP_FILEOPS.setPermissions(file, to_qt_permissions(msg->attr->permissions)))
+    std::error_code err;
+    MP_FILEOPS.permissions(filename, static_cast<fs::perms>(msg->attr->permissions), err);
```
georgeliao, line 487: Maybe `static_cast<fs::perms>(msg->attr->permissions)` -> `fs::perms(msg->attr->permissions)`, both interpret uint32_t into bitmask, but the second one is cleaner.

andrei-toterman: Agreed

ricab: Hey guys, sorry I noticed this flying by. Why is a functional cast here preferable to a `static_cast`?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -622,124 +604,122 @@ int mp::SftpServer::handle_open(sftp_client_message msg)
...
-        const QFileInfo entry = dir_entries->takeFirst();
-        const auto filename = entry.fileName().toStdString();
+        const auto& entry = dir_iterator.next();
+        QFileInfo file_info{entry.path().string().c_str()};
         sftp_attributes_struct attr{};
-        if (entry.isSymLink())
+        if (entry.is_symlink())
         {
-            mp::platform::symlink_attr_from(entry.absoluteFilePath().toStdString().c_str(), &attr);
+            std::error_code _;
```
georgeliao, line 712: This `_` is not used. I think.

andrei-toterman: True. Must've forgotten it there from a previous attempt.

src/sshfs_mount/sftp_server.cpp
ricab: I guess `{}` is OK too if no narrowing is involved.

andrei-toterman: Yes, I think `{}` is the cleanest. I understand that a C-style cast is not really recommended and, if this was a conversion between more complex data types i.e. classes, I would also definitely prefer named casts. In this case, since it's just an integer to enum conversion, I initially agreed to the C-style cast since it would simply act like the `static_cast` that it replaced. But now that you mention `{}`, I think that is preferable. Thanks for pointing it out!

georgeliao: > I guess {} is OK too if no narrowing is involved.

That is what the item of the guideline said. On the top of that () and {} functional cast are also slightly different in terms of narrowing prevention. 

> Yes, I think {} is the cleanest. I understand that a C-style cast is not really recommended and,
I think you mean functional cast, C-style is the (T) var.

georgeliao:
@andrei-toterman , I have done the functional testing on various platforms and backends. The testing includes mounting the folder via classic mount and using the folder for creating/deleting files, git clone repo and delete the cloned folder, etc. They all work fine. @townsend2010 , I could not reproduce your permission denied error. Maybe that was based on the old version of the code.

townsend2010:
Hi @georgeliao!  Yes, that issue I ran into was with an older version of this branch and I have verified that it is now fixed as well.  Thanks!

townsend2010:
Hey @andrei-toterman,

Just a question about test coverage...this branch is not very well covered.  Is this due to a difficulty in testing the actual changes or is it an oversight?

src/sshfs_mount/sftp_server.cpp
ricab: Ahaa! And thus we just prevented a narrowing conversion.

tests/test_file_ops.cpp
```diff
@@ -19,6 +19,9 @@
 
 #include <multipass/file_ops.h>
 
+#include <fcntl.h>
+#include <streambuf>
```
georgeliao, line 23: I do not think we need the header streambuf.

georgeliao:
@andrei-toterman I have done another round of functional testing. It includes qemu and lxd on linux; Hyper-v and VirtualBox on virtualized windows; qemu on macOs. They all work fine. I think now it is ready to merge. @townsend2010 can you also remove the change request?

georgeliao (approved):

townsend2010 (approved):
LGTM now too, thanks!

PR #3276, opened 2023-10-28
Message only on SSH execution error

luis4a0:
The execution of commands through SSH used to throw `std::runtime_error` when the return value was not `0`. This PR adds a new exception to be thrown in that case, making better catching an execution failure. This permitted us to show the interface configuration message only when needed.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3081,7 +3081,7 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
                         mpu::run_in_ssh_session(session, command);
                     }
                 }
-                catch (const std::runtime_error&) // In case there is an error executing the command, report it.
+                catch (const mp::SSHExecFailure&) // In case there is an error executing the command, report it.
                 {
                     // Currently, the only use of running commands at boot is to configure networks. For this reason,
                     // the warning shown here refers to that use. In the future, in case of using the feature for other
--- a/include/multipass/exceptions/ssh_exception.h
+++ b/include/multipass/exceptions/ssh_exception.h
@@ -30,5 +30,13 @@ public:
     {
     }
 };
+
+class SSHExecFailure : public SSHException
+{
+public:
+    SSHExecFailure(const std::string& what_arg) : SSHException(what_arg)
+    {
+    }
+};
 } // namespace multipass
 #endif // MULTIPASS_SSH_EXCEPTION_H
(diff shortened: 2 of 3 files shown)


src/daemon/daemon.cpp
```diff
@@ -3081,7 +3081,7 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
                         mpu::run_in_ssh_session(session, command);
                     }
                 }
-                catch (const std::runtime_error&) // In case there is an error executing the command, report it.
+                catch (const mp::SSHExecFailure&) // In case there is an error executing the command, report it.
                 {
```
georgeliao, line 3085: It is already a big improvement. Just one small addition, maybe we can also add another branch below to catch `mp::SSHException` and return an error message because there is still a chance that the `mp::SSHSession` constructor throws.

luis4a0: Hey @georgeliao, thanks! That is a good idea but, what to do in the `catch SSHException&` branch? Just message a warning?

georgeliao: If that fails, then it can be still a warning, right? We suggest user to do it manually.

luis4a0: Ok, but in that case it will be the same as before...

luis4a0: Or almost

georgeliao: Hmm, indeed, then just catching mp::SSHException alone and append the exception message can be an option as well.

luis4a0: Done! Please take a look to the change.

src/daemon/daemon.cpp
```diff
@@ -3090,12 +3090,20 @@ mp::Daemon::async_wait_for_ready_all(grpc::ServerReaderWriterInterface<Reply, Re
...
                                    name);
                     }
 
                     warned_exec_failure = true;
                 }
+                catch (const SSHException&) // The SSH session could not be created.
+                {
+                    add_fmt_to(warnings,
+                               "Cannot create a SSH shell to execute commands on {}, you can configure new\n"
+                               "interfaces manually Netplan once logged to the instance.\n",
```
georgeliao, line 3104: Maybe "...manually **via** Netplan... "

src/daemon/daemon.cpp
georgeliao: Great, it looks way better now. just one small error message wording improvement.

src/daemon/daemon.cpp
luis4a0: Done! I also added some test for these newly added lines.

PR #3281, opened 2023-11-03
Gui instance list

andrei-toterman:
This PR adds the Instances page in the GUI, where a table shows all instances and their relevant info and allows bulk actions to be performed on them.


andrei-toterman:

PR #3285, opened 2023-11-06
Support Mantic compilers

luis4a0:
This fixes compilation with gcc-13 and clang-16.

### DIFF
--- a/3rd-party/submodule_info.md
+++ b/3rd-party/submodule_info.md
@@ -17,7 +17,7 @@ Version: 0.10.4 (+[our patches](https://github.com/CanonicalLtd/libssh/compare/l
 <https://www.libssh.org/>
 
 ### fmt
-Version: 9.1.0 |
+Version: 10.1.1 |
 <https://github.com/fmtlib/fmt.git> |
 <https://github.com/fmtlib/fmt/releases>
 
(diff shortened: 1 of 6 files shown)


luis4a0:
Still having issues with the `fmt` lib with gcc-13.

luis4a0:
Done.

luis4a0:
Tested it also works with clang-17.

3rd-party/fmt
townsend2010, line 1: It looks like this is an update to the `fmt` submodule.  Please update the `3rd-party/submodule_info.md` file to reflect this?

townsend2010 (changes requested):
Seems fine overall, but I have one inline comment and a general question.  Why the need to add those escaped quotes in some of the strings?

luis4a0:
Hey @townsend2010, thanks! I added those escape chars because, for some change in `fmt`, some strings which were outputted with double quotes don't have double quotes now. I wanted the output to be the same as before the change in `fmt`.

3rd-party/fmt
luis4a0: Done. Switched to version 10.1.1 (the last stable) and reflected it in `submodule_info.md`.

luis4a0:
Hey @townsend2010! The default formatting of `std::filesystem::path` changed in the last version of `fmtlib`. It used to quote the paths, and now it defaults to not quoting them. Thus I forced quoting where needed, so the tests are not affected by the change.

townsend2010 (approved):
Ok, LGTM, thanks!

PR #3288, opened 2023-11-07
Confirm snapshot deletion with prompt

ricab:
Confirm snapshot deletion with a prompt to users when they try to delete a snapshot without the `--purge` flag, as long as multipass is running interactively.

### DIFF
--- a/src/client/cli/cmd/delete.cpp
+++ b/src/client/cli/cmd/delete.cpp
@@ -75,12 +75,13 @@ mp::ReturnCode cmd::Delete::run(mp::ArgParser* parser)
         if (!reply.log_line().empty())
             cerr << reply.log_line();
 
+        // TODO refactor with bridging and restore prompts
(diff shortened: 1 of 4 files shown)


ricab:
Note that this is stacked on #3287 .

luis4a0:
Hey @ricab, thanks for this! Code looks nice, just a bit of manual testing and we're ready to go.

ricab:
Hey @luis4a0, thank you for the review!

We've been following the "convention" of using a capital letter for the default answer. With quotes because I am not sure how conventional it actually is, but I have seen it in different places. When there is no default, it is better to use just "yes/no" (as in the bridging question). Hope that makes sense :slightly_smiling_face:

luis4a0:
@ricab that makes complete sense, I didn't know about the convention. That's all I had. Thanks!

luis4a0 (approved):

PR #3289, opened 2023-11-07
Tweak handling of snapshots of deleted instances

ricab:
Stick with the approach of making properties of deleted instances readable but not writable.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1142,7 +1142,7 @@ bool verify_snapshot_picks(const InstanceSelectionReport& report,
             {
                 for (const auto& snapshot_name : pick_it->second.pick)
                 {
-                    if (selection == &report.deleted_selection && (!pick_it->second.all_or_none || !purge))
+                    if (selection == &report.deleted_selection && !(pick_it->second.all_or_none && purge))
                         snapshots_of_deleted_instances.push_back(fmt::format("{}.{}", vm_it->first, snapshot_name));
 
                     vm_it->second->get_snapshot(snapshot_name); // throws if missing
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1151,7 +1151,7 @@ bool verify_snapshot_picks(const InstanceSelectionReport& report,
             }
         }
 
-        if (snapshots_of_deleted_instances.size())
+        if (!snapshots_of_deleted_instances.empty())
             throw std::runtime_error{fmt::format(
                 "Cannot delete snapshots of deleted instances: {}",
                 fmt::join(snapshots_of_deleted_instances.begin(), snapshots_of_deleted_instances.end(), ", "))};
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1841,7 +1841,7 @@ try // clang-format on
     auto cmd = request->snapshots() ? std::function(fetch_snapshot) : std::function(fetch_instance);
 
     auto status = cmd_vms(select_all(operative_instances), cmd);
-    if (status.ok() && !request->snapshots())
+    if (status.ok())
     {
         deleted = true;
         status = cmd_vms(select_all(deleted_instances), cmd);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1129,14 +1129,29 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
 }
 
 bool verify_snapshot_picks(const InstanceSelectionReport& report,
-                           const std::unordered_map<std::string, SnapshotPick>& snapshot_picks)
+                           const std::unordered_map<std::string, SnapshotPick>& snapshot_picks,
+                           bool purge)
 {
     auto any_snapshot = false;
     for (const auto* selection : {&report.deleted_selection, &report.operative_selection})
+    {
         for (const auto& vm_it : *selection)
+        {
             if (auto pick_it = snapshot_picks.find(vm_it->first); pick_it != snapshot_picks.end())
+            {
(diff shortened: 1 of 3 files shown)


ricab:
Note that this is stacked on #3288.

src/daemon/daemon.cpp
```diff
@@ -1129,14 +1129,33 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
...
+        {
             if (auto pick_it = snapshot_picks.find(vm_it->first); pick_it != snapshot_picks.end())
+            {
                 for (const auto& snapshot_name : pick_it->second.pick)
-                    any_snapshot = true, vm_it->second->get_snapshot(snapshot_name); // throws if missing
+                {
+                    if (selection == &report.deleted_selection && (!pick_it->second.all_or_none || !purge))
+                        snapshots_of_deleted_instances.push_back(fmt::format("{}.{}", vm_it->first, snapshot_name));
+
+                    vm_it->second->get_snapshot(snapshot_name); // throws if missing
```
georgeliao, line 1148: I think the original line `any_snapshot = true, vm_it->second->get_snapshot(snapshot_name);` equals 
```
any_snapshot = true; 
vm_it->second->get_snapshot(snapshot_name);
```
Should it be the same order in order to keep the same behavior since get_snapshot has a chance to throw?

src/daemon/daemon.cpp
```diff
@@ -1129,14 +1129,33 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
 }
 
 bool verify_snapshot_picks(const InstanceSelectionReport& report,
-                           const std::unordered_map<std::string, SnapshotPick>& snapshot_picks)
+                           const std::unordered_map<std::string, SnapshotPick>& snapshot_picks,
+                           bool purge)
 {
     auto any_snapshot = false;
+    std::vector<std::string> snapshots_of_deleted_instances{};
     for (const auto* selection : {&report.deleted_selection, &report.operative_selection})
```
georgeliao, line 1137: Basically, `snapshots_of_deleted_instances` and the `if (selection == &report.deleted_selection && (!pick_it->second.all_or_none || !purge))` and the throwing later are only for `report.deleted_selection`. If we split the loop into two code blocks, then it will be more clear and have more localized variables. However, that causes more code duplication. I guess it is a trade-off and we can leave it like this.

src/daemon/daemon.cpp
```diff
@@ -1129,14 +1129,33 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
...
+                    if (selection == &report.deleted_selection && (!pick_it->second.all_or_none || !purge))
+                        snapshots_of_deleted_instances.push_back(fmt::format("{}.{}", vm_it->first, snapshot_name));
+
+                    vm_it->second->get_snapshot(snapshot_name); // throws if missing
+                    any_snapshot = true;
+                }
+            }
+        }
+
+        if (snapshots_of_deleted_instances.size())
```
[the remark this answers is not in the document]
ricab: Sure, I'll do that.

src/daemon/daemon.cpp
ricab: You're right that this changes the order, but I don't think it matters much. That boolean is supposed to be returned, but it won't be if the function exits with an exception. The previous order was also pretty much arbitrary.

src/daemon/daemon.cpp
```diff
@@ -1129,14 +1129,33 @@ InstanceSnapshotsMap map_snapshots_to_instances(const InstanceSnapshotPairs& ins
...
     for (const auto* selection : {&report.deleted_selection, &report.operative_selection})
+    {
         for (const auto& vm_it : *selection)
+        {
             if (auto pick_it = snapshot_picks.find(vm_it->first); pick_it != snapshot_picks.end())
+            {
                 for (const auto& snapshot_name : pick_it->second.pick)
-                    any_snapshot = true, vm_it->second->get_snapshot(snapshot_name); // throws if missing
+                {
+                    if (selection == &report.deleted_selection && (!pick_it->second.all_or_none || !purge))
```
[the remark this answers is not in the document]
ricab: That's right. If the user tries to delete a snapshot of an instance and to purge-delete the instance in the same command, then the command should succeed. For example:

```
$ multipass delete my-vm
$ multipass delete -p my-vm.snapshot my-vm
```

src/daemon/daemon.cpp
ricab: Yes, the rationale is indeed less code duplication. If I had two blocks, you would still need to read them to understand that they are mostly the same. This makes common code immediate and still separates the difference in that inner `if`. It is not ideal, but it seems to me to be the "least bad" of the two options.

georgeliao:
@ricab  , most things are addressed, one last thing is the small tweak of the conditions. Let me know if that can be done. If so, maybe we can push that commit and get the build ready tonight. So I can do the functional testing tomorrow morning.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
ricab: Sure, I can do that. I think I had it that way but thought it better to "simplify" all the way. But if you think that's more understandable, I'll change it.

ricab:
Hey @georgeliao, I just pushed that last change. Please let me know if there is anything else you need.

georgeliao:
@ricab All comments are addressed. Besides that, I have run extensive functional tests on various backends and platforms, it includes qemu on linux, hyper-v on windows and qemu on macos. They all work fine. I think it is ready to merge.

georgeliao (approved):

ricab:
Great, thanks a lot for the review @georgeliao!

PR #3298, opened 2023-11-10
[tests] Try fixing the random ISO generation test failure in CI

townsend2010:


ricab (approved):
Thanks! Let's hope this does it :crossed_fingers: 

bors r+

townsend2010:
Kicking this to get the changes on `full` regarding the Qt install timeout.

townsend2010:
Ok, got a good CI run.  No need for an integration test run since this only affects a unit test.  Will mege.

PR #3305, opened 2023-11-16
ci: drop bors from linux.yml

Saviq:


Saviq:
@townsend2010 you will need to review-approve this. Also, since I pushed, it needs "Merge when ready" again.

.github/workflows/linux.yml
```diff
@@ -243,19 +238,6 @@ jobs:
...
-    needs:
-    - BuildAndTest
-
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Report failure
-      if: ${{ needs.BuildAndTest.result != 'success' }}
-      run: exit 1
-
```
Saviq, line 258: @townsend2010 you'll want to drop the bare "Linux" check from required checks. This was only here so we needn't list all the build types in `bors.toml`.

townsend2010: Hey @Saviq, I don't think the bare Linux check is a required check from what I can tell.

townsend2010 (dismissed):
Hey @Saviq!

Cool, thanks!  This makes sense to me.

Saviq:
@townsend2010 probably want to add Lint, though:

townsend2010:
@Saviq Yep, thanks, added that and cla-check as well.

Saviq:
Huuh?

```
 -block-labels = [ "no-merge" ]
Error:  Forbidden string(s) found in the diff ^ (matched "(?i)\bno[_-]?merge\b")
```

@townsend2010 you may want to drop Lint for now after all :P

townsend2010:
Strange, I don't even understand what it's complaining about here, but ok, whatever, I will remove it for now and enable it again after this is merged.

Saviq:
> I don't even understand what it's complaining about here

Yeah, about the file that I just deleted ¯\_(ツ)_/¯

ricab:
Hi guys, where did that show up? I think it is just noticing the "forbidden string" in the diff and complaining because of that. It should probably only do so for added lines...

ricab:
`bors.toml` would normally have been excluded, but that is also changing with this diff. Do you mind if I commit something on top for fix the "no-merge" issue here?

Saviq:
Please don't commit here. I think we should just ignore Lint on this one.

ricab:
You're removing the exclusion for bors.toml, I thought that could justify replacing it with a more generic exclusion. But OK, I will make a separate PR.

townsend2010 (dismissed):
Let's go for this again!

townsend2010 (dismissed):
LGTM, thanks!

townsend2010 (dismissed):
Let's try this

townsend2010 (approved):
Ok, LGTM.  Will merge manually after the CI run.

townsend2010 (approved):
:+1:

townsend2010:
Ok, seems this is ok and will merge manually.

PR #3307, opened 2023-11-16
[ci] Forbid only additions of unwanted strings

ricab:
Allow "no-merge" variations to appear in the diff as long as they are
only removals.


townsend2010 (approved):
Ok, seems to work on https://github.com/canonical/multipass/pull/3305.  Thanks!

PR #3315, opened 2023-11-24
Merge blueprint vendor data

### LINKED ISSUE #3247: Passed in `cloud-init` with `write-files` defined overwrites pollinate data
If any passed in `cloud-init`, be it from `--cloud-init` or a Blueprint, has `write-files` defined, it will overwrite the data we generate for pollinate when getting entropy.

We should instead be smart about this and try as best as we can to merge entries where possible.

andrei-toterman:
This PR makes it so that cloud-init vendor data coming from a blueprint is merged with the vendor data that we normally provide. The merging follows these rules:
- If a key refers to a scalar value in both our vendor data and the blueprint vendor data, the blueprint is rejected
- If a key refers to values of different types in our vendor data and the blueprint vendor data, the blueprint is rejected
- If an entry from the blueprint vendor data does not exist in our own vendor data, simply add it
- If a key refers to a sequence in both our vendor data and the blueprint vendor data, append the items from the blueprint sequence to the sequence from our data
- If a key refers to a map in both our vendor data and the blueprint vendor data, recurse this algorithm on each key from the blueprint map

Fixes #3247

### DIFF
--- a/src/blueprint_provider/default_vm_blueprint_provider.cpp
+++ b/src/blueprint_provider/default_vm_blueprint_provider.cpp
@@ -168,6 +168,41 @@ std::string get_blueprint_version(const YAML::Node& blueprint_instance)
     return "v1";
 }
 
+void merge_yaml_entries(YAML::Node& ours, const YAML::Node& theirs, const std::string& key, bool override_null = false)
+{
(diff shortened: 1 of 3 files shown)


townsend2010 (approved):
Hey @andrei-toterman!

Thanks for this.  LGTM and works as advertised!

PR #3349, opened 2024-01-02
[mounts] only allow one to one id mappings

### LINKED ISSUE #3332: [mounts] Allow only one-to-one uid/gid mappings for mounts
**Describe the bug**
When specifying uid/gid mappings for a mount, you can use the same uid/gid in more than one mapping. This creates confusion as it is not clear which map should be used.

**To Reproduce**
How, and what happened?
1. `multipass mount dir instance:dir -u 1000:2000 -u 1000:3000 -u 3000:2000`
1. Now it's not clear if a host file owned by 1000 should be owned 2000 or 3000 in the VM or if a remote file owned by 2000 should be owned by 1000 or 3000 on the host.

**Expected behavior**
Throw an error when an attempted `mount` does not have a one-to-one correspondence between its mapped uids/gids.

sharder996:
Fixes #3332

Public side of [#641](https://github.com/canonical/multipass-private/pull/641)

### DIFF
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -33,24 +33,20 @@ namespace multipass
 {
 using id_mappings = std::vector<std::pair<int, int>>;
 
-inline id_mappings unique_id_mappings(const id_mappings& xid_mappings)
+inline auto unique_id_mappings(id_mappings& xid_mappings)
 {
-    id_mappings ret;
-    std::set<std::pair<int, int>> id_set;
+    std::set<int> id_set;
+    std::set<int> rev_id_set;
 
-    auto is_mapping_repeated = [&id_set](const auto& m) {
-        if (id_set.insert(m).second)
+    auto is_mapping_repeated = [&id_set, &rev_id_set](const auto& m) {
+        if (id_set.insert(m.first).second && rev_id_set.insert(m.second).second)
             return false;
 
         mpl::log(mpl::Level::debug, "id_mappings", fmt::format("Dropping repeated mapping {}:{}", m.first, m.second));
-
         return true;
     };
 
-    std::remove_copy_if(xid_mappings.cbegin(), xid_mappings.cend(), std::back_insert_iterator(ret),
-                        is_mapping_repeated);
-
-    return ret;
+    return std::remove_if(xid_mappings.begin(), xid_mappings.end(), is_mapping_repeated);
 }
 } // namespace multipass
 
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -52,7 +52,7 @@ inline auto unique_id_mappings(id_mappings& xid_mappings)
             mpl::log(mpl::Level::debug,
                      "id_mappings",
                      fmt::format("Dropping repeated mapping {}:{}", it->first, it->second));
-            xid_mappings.erase(it);
+            it = xid_mappings.erase(it);
         }
         else
         {
@@ -60,7 +60,7 @@ inline auto unique_id_mappings(id_mappings& xid_mappings)
         }
     }
 
-    auto filter_map = [](auto& map) {
+    auto filter_non_repeating = [](auto& map) {
         for (auto it = map.begin(); it != map.end();)
         {
             if (it->second.size() <= 1)
@@ -70,8 +70,8 @@ inline auto unique_id_mappings(id_mappings& xid_mappings)
         }
     };
 
-    filter_map(dup_id_map);
-    filter_map(dup_rev_id_map);
+    filter_non_repeating(dup_id_map);
+    filter_non_repeating(dup_rev_id_map);
 
     return std::make_pair(dup_id_map, dup_rev_id_map);
 }
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <iterator>
 #include <set>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
--- a/include/multipass/id_mappings.h
+++ b/include/multipass/id_mappings.h
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <iterator>
 #include <set>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -35,18 +36,44 @@ using id_mappings = std::vector<std::pair<int, int>>;
 
 inline auto unique_id_mappings(id_mappings& xid_mappings)
 {
-    std::set<int> id_set;
-    std::set<int> rev_id_set;
+    std::unordered_map<int, std::unordered_set<int>> dup_id_map;
+    std::unordered_map<int, std::unordered_set<int>> dup_rev_id_map;
 
-    auto is_mapping_repeated = [&id_set, &rev_id_set](const auto& m) {
-        if (id_set.insert(m.first).second && rev_id_set.insert(m.second).second)
-            return false;
+    for (auto it = xid_mappings.begin(); it != xid_mappings.end();)
+    {
+        bool duplicate =
+            dup_id_map.find(it->first) != dup_id_map.end() || dup_rev_id_map.find(it->second) != dup_rev_id_map.end();
 
-        mpl::log(mpl::Level::debug, "id_mappings", fmt::format("Dropping repeated mapping {}:{}", m.first, m.second));
-        return true;
+        dup_id_map[it->first].insert(it->second);
+        dup_rev_id_map[it->second].insert(it->first);
+
+        if (duplicate)
+        {
+            mpl::log(mpl::Level::debug,
+                     "id_mappings",
+                     fmt::format("Dropping repeated mapping {}:{}", it->first, it->second));
+            xid_mappings.erase(it);
+        }
+        else
+        {
+            ++it;
+        }
+    }
+
+    auto filter_map = [](auto& map) {
+        for (auto it = map.begin(); it != map.end();)
+        {
+            if (it->second.size() <= 1)
+                it = map.erase(it);
+            else
+                ++it;
+        }
     };
 
-    return std::remove_if(xid_mappings.begin(), xid_mappings.end(), is_mapping_repeated);
+    filter_map(dup_id_map);
+    filter_map(dup_rev_id_map);
+
+    return std::make_pair(dup_id_map, dup_rev_id_map);
 }
 } // namespace multipass
 
--- a/include/multipass/vm_mount.h
+++ b/include/multipass/vm_mount.h
@@ -26,8 +26,9 @@
 
 namespace multipass
 {
-struct VMMount
+class VMMount
 {
+public:
     enum class MountType : int
     {
         Classic = 0,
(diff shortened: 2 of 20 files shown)


andrei-toterman:
This is not a review, I just glanced over it out of curiosity. And I'm not sure I understand the whole `BackInserter` thing, so I would like to open up a discussion on this. Couldn't we use something like
```cpp
auto unique_iter_end = std::unique(mappings.begin(), mappings.end(), [](auto& a, auto& b) {
    return a.first == b.first || a.second == b.second;
});
```
Then `unique_iter_end != mappings.end()` to check if there were duplicates or `mappings.erase(unique_iter_end, mappings.end())` to remove the duplicates.

sharder996:
Hey @andrei-toterman,

The idea of the BackInserter was so that insertions could be made into the returned vector from either an existing vector (as used in `vm_mount`) or from a repeated grpc object of type `IdMap` (as used in the `daemon`). Other motivations for changes in `id_mappings` was also to return a vector of duplicate mappings so that they could be reported back to the user. Also, templating the `unique_id_mappings()` function was done so that it could accept multiple types of containers.

Using `std::unique()` looks promising except for that it only finds groups of equivalent elements in consecutive groups.

andrei-toterman:
Oops, you're right about the `std::unique`, I misread the documentation. We have to filter the duplicates manually.

Anyway, when trying to do a mount with repeated mappings, the errors are indeed reported, but the mount still occurs. Shouldn't we abort the operation entirely?
For example, if I do this
```shell
mp mount dir instance:dir -u 1000:1000 -u 1000:2000
```
I get the following error message
```
cannot apply mapping 1000:2000 with duplicate uids
```
But the mount still happens, but only with the `1000:1000` mapping.

What about checking for duplicates only inside the `MountHandler` base class constructor and throwing if they occur. That way it would be impossible to create an invalid `MountHandler`. Or maybe do the check inside `VMMount`'s constructor? Anyway, I think we should make invalid states unrepresentable. What do you think?

sharder996:
Yes, I had considered that. For the some of the "errors" that can happen when creating mounts (eg. incorrect paths, incorrect instances name) we simply ignore those items and continue with establish as many of the mounts as possible. That is why I treated mappings the same way.

I like the idea of checking mapping inside of `VMMount` and not so much in the `MountHandler` since it need to dive inside the `VMMount` passed to it. Rather do that directly when the `VMMount` is created.

src/utils/vm_mount.cpp
```diff
@@ -58,6 +58,22 @@ mp::VMMount::VMMount(const std::string& sourcePath,
...
+        while (iter_begin != iter_end)
+        {
+            retval += fmt::format("{}:{}; ", (*iter_begin).first, (*iter_begin).second);
+            ++iter_begin;
+        }
+        return retval.substr(0, retval.size() - 2);
+    };
+
+    if (auto unique_iter_end = mp::unique_id_mappings(gid_mappings); unique_iter_end != gid_mappings.end())
+        throw std::runtime_error(fmt::format("Mount cannot apply mapping with duplicate gids: {}",
```
townsend2010, line 72: I don't think the message is very clear here.  It doesn't reference the first mapping found with the duplicate.  For example, if I passed in `-g 1000:1000 -g 1000:1001 -g 1000:10002`, it will only print out the last two.  I think it should print *all* of the duplicate mappings and let the user see what they are and decide how to fix it.

Also, I'm kind of on the fence on this, but I wonder if we should be a little more specific *which* gid is the duplicate meaning either the host gid or the instance gid that is the offender.  Any opinions on this?

src/utils/vm_mount.cpp
```diff
@@ -58,6 +58,22 @@ mp::VMMount::VMMount(const std::string& sourcePath,
...
+                                             print_mappings(unique_iter_end, gid_mappings.end())));
+    if (auto unique_iter_end = mp::unique_id_mappings(uid_mappings); unique_iter_end != uid_mappings.end())
+        throw std::runtime_error(fmt::format("Mount cannot apply mapping with duplicate uids: {}",
+                                             print_mappings(unique_iter_end, uid_mappings.end())));
```
townsend2010, line 76: Idem: Error message is a bit unclear as I pointed out above.

townsend2010 (changes requested):
Hey @sharder996!

Thanks for this!  I'm not done reviewing, but I did want to go ahead and give you my feedback below.

src/utils/vm_mount.cpp
sharder996: I agree that the error message could be more descriptive, but it require the filtering algorithm to change quite a bit so that the offending id can be identified as well as it's associated duplicate id mappings. I haven't thought of a good way to do this yet and seeing as how other things are higher priority I haven't done anything yet.

townsend2010: Ok, makes sense.  We can put this on the back burner for now until the other items you are working on are more complete.

src/utils/vm_mount.cpp
```diff
@@ -41,8 +41,8 @@ mp::VMMount parse_json(const QJsonObject& json)
             {gid_entry.toObject()["host_gid"].toInt(), gid_entry.toObject()["instance_gid"].toInt()});
     }
 
-    uid_mappings = mp::unique_id_mappings(uid_mappings);
-    gid_mappings = mp::unique_id_mappings(gid_mappings);
+    mp::unique_id_mappings(uid_mappings), uid_mappings.end();
+    mp::unique_id_mappings(gid_mappings), gid_mappings.end();
```
andrei-toterman, line 45: What do those `.end()` do here?

include/multipass/vm_mount.h
```diff
@@ -26,33 +26,63 @@
...
 
     QJsonObject serialize() const;
 
+    const std::string& get_source_path() const noexcept;
+    const id_mappings& get_gid_mappings() const noexcept;
+    const id_mappings& get_uid_mappings() const noexcept;
+    MountType get_mount_type() const noexcept;
+
+    friend bool operator==(const VMMount& a, const VMMount& b) noexcept;
+    friend bool operator!=(const VMMount& a, const VMMount& b) noexcept;
```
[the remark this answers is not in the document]
sharder996: They are not member functions, so they can't be CV-qualified.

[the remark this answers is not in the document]
andrei-toterman: Oh, right. I guess they could be member functions, though. But it's fine like this.

tests/test_vm_mount.cpp
```diff
@@ -20,44 +20,123 @@
...
+
+const mp::VMMount TestVMMount::a_mount{"asdf", {{1, 2}, {2, 4}}, {{8, 4}, {6, 3}}, mp::VMMount::MountType::Classic};
+
+TEST_P(TestUnequalVMMountsTestSuite, CompareMountsUnequal)
+{
+    auto [mount_a, mount_b] = GetParam();
+
+    // Force use of overloaded operator== and operator!=
+    EXPECT_TRUE(mount_a != mount_b);
+    EXPECT_FALSE(mount_a == mount_b);
```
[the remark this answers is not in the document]
sharder996: This is to force the use of the `operator==()` and the `operator!=()` functions and cover them both. Because `a_mount` and `b_mount` are supposed to be unequal, there's no way to use `EXPECT_EQ` and force the use of `operator==()`.

include/multipass/id_mappings.h
```diff
@@ -33,24 +34,46 @@ namespace multipass
...
+        dup_id_map[it->first].insert(it->second);
+        dup_rev_id_map[it->second].insert(it->first);
 
-        return true;
+        if (duplicate)
+        {
+            mpl::log(mpl::Level::debug,
+                     "id_mappings",
+                     fmt::format("Dropping repeated mapping {}:{}", it->first, it->second));
+            xid_mappings.erase(it);
```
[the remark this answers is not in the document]
sharder996: Right, missed that.

sharder996:
@andrei-toterman Thanks for the review! I think I've addressed all your comments.

tests/test_vm_mount.cpp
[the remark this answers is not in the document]
andrei-toterman: Ah, I understand. You could always do something like `EXPECT_EQ(mount_a, mount_a)`, since that's pretty much an essential test for `operator==()` to pass. But I think this is a trivial thing, so we'll keep it as it is.

andrei-toterman:
Thanks, @sharder996. It's looking good. I just want your opinion on a last detail:
```
$ multipass mount -u 1000:1001 -u 1002:1001 -g 1000:1000 -g 1001:1000 ~ test:bla      
mount failed: Mount cannot apply mapping with duplicate uids: 1001: [1002:1001, 1000:1001]
Mount cannot apply mapping with duplicate gids: 1000: [1001:1000, 1000:1000]
```
As you can see, the first line reporting duplicate uids is prefixed by the `mount failed: ` thing. How do you think this would look better? Maybe putting both messages on new lines? Or maybe somehow integrating them into one line? Or maybe something else?

andrei-toterman:
@sharder996, sure, I think that looks great. Thanks!

andrei-toterman:
Alright, this should be ready to be merged. But it's failing to build on CI on mac. Seems like it's a missing include https://github.com/canonical/multipass-private/actions/runs/9411867391/job/25931752655#step:26:3601

src/utils/vm_mount.cpp
```diff
@@ -58,6 +58,51 @@ mp::VMMount::VMMount(const std::string& sourcePath,
       uid_mappings(std::move(uidMappings)),
       mount_type(mountType)
 {
+    fmt::memory_buffer errors;
+
+    auto print_mappings = [](auto& dup_id_map, const auto& dup_rev_id_map) {
```
ricab, line 63: Why a lambda? You're not capturing anything, so no closure. I like that you have a function, but I would prefer to see it in the unnamed namespace for this file. The reader of this constructor (or any function) will most often want to see _what_ it is doing, not _how_ it is doing it. If they want to see the _what_ part, they can always zone in on what interests them.

ricab (commented):
Nice, good code. Thank you Scott! I have only one request inline.

include/multipass/id_mappings.h
```diff
@@ -33,24 +35,46 @@ namespace multipass
 {
 using id_mappings = std::vector<std::pair<int, int>>;
 
-inline id_mappings unique_id_mappings(const id_mappings& xid_mappings)
+inline auto unique_id_mappings(id_mappings& xid_mappings)
```
ricab, line 38: I am not a fan of this in a header, but this whole file shouldn't even exist (this function is an implementation detail of `VMMount` exclusively). Fixing that is out of scope here, so fine.

ricab (approved):
Thank you @sharder996, this looks good to me! 

@andrei-toterman, as primary reviewer, I'll leave it to you to give the final go-ahead. Thanks.

andrei-toterman (approved):

andrei-toterman (changes requested):
@sharder996. There is an error with building on Windows. I think you'll need a private PR. Sorry I didn't catch this earlier!

andrei-toterman (approved):

PR #3350, opened 2024-01-02
[sftp] set files without a mapping as nobody:nogroup

### LINKED ISSUE #3333: [mounts] Map unknown uid/gid to `nobody:nobody`
**Describe the bug**
When mounting a directory that contains files owned by uids/gids that are not mapped to any uids/gids on the VM, no mapping occurs at all and inside the VM they appear to be owned by the same uids/gids as on the host. This can lead to confusion, since we deny any operation on a file that does not have its uid and gid mapped.

**To Reproduce**
How, and what happened?
1. Provide irrelevant mappings to override the default
```
multipass mount dir instance:dir -u 1234:1234 -g 1234:1234
```
2. Have a file owned by the uid and gid of `ubuntu:ubuntu` (`1000:1000` in this case)
```
sudo chown 1000:1000 dir/file
```
3. From inside the VM, it appears as if the file is owned by `ubuntu:ubuntu`
```
ls -l dir/file
-rw-r--r-- 1 ubuntu ubuntu 0 Dec 14 17:00 dir/file
```
4. But we can't use it
```
cat dir/file
cat: dir/file: Permission denied
```

**Expected behavior**
When no mapping is provided for a uid/gid, it should be mapped to `65534` (the `nobody` user/group).

sharder996:
Fixes #3333

### DIFF
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -215,22 +215,14 @@ auto create_sshfs_process(mp::SSHSession& session, const std::string& sshfs_exec
     return std::make_unique<mp::SSHProcess>(std::move(sshfs_process));
 }
 
-int mapped_id_for(const mp::id_mappings& id_maps, const int id, const int id_if_not_found)
+int mapped_id_for(const mp::id_mappings& id_maps, const int id, const int default_id)
 {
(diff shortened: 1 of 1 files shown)


townsend2010 (approved):
Hey @sharder996!

LGTM, thanks!

PR #3360, opened 2024-01-08
Catch timeout on commands at boot

luis4a0:
Some executions can give a timeout for several reasons, but does not necessarily mean there was an error. However, the daemon used to fail on timeouts as it failed in severe cases like not being able to create a SSH connection. This PR makes the daemon catch the timeout exception, avoiding an error be shown to the user.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3671,6 +3671,12 @@ void mp::Daemon::run_commands_at_boot_on_instance(const std::string& name, fmt::
 
             warned_exec_failure = true;
         }
+        catch (const ExitlessSSHProcessException& e) // Timeout does not mean there is a configuration error.
+        {
+            mpl::log(mpl::Level::warning,
+                     category,
+                     fmt::format("Exception when executing command at boot in {}: {}", name, e.what()));
+        }
         catch (const SSHException&) // The SSH session could not be created.
         {
             add_fmt_to(warnings,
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -3671,6 +3673,10 @@ void mp::Daemon::run_commands_at_boot_on_instance(const std::string& name, fmt::
 
             warned_exec_failure = true;
         }
+        catch (const ExitlessSSHProcessException&) // Timeout does not mean there is a configuration error.
```
townsend2010, line 3676: Why not just use the string returned in `what()` here and not have to capture the current command above?

src/daemon/daemon.cpp
```diff
@@ -3671,6 +3673,10 @@ void mp::Daemon::run_commands_at_boot_on_instance(const std::string& name, fmt::
...
+        {
+            add_fmt_to(warnings, "Timeout executing command \"{}\" in instance {}.\n", *current_command, name);
```
townsend2010, line 3678: Not sure if this really needs to be communicated back to the client.  I think logging the warning would be better.

townsend2010 (changes requested):
Hey @luis4a0!

Thanks for this!  I have a few comments below that I think need to be addressed.

luis4a0:
Thanks for the review @townsend2010! I addressed your two points, so this is ready for a new round.

townsend2010 (approved):
Hey @luis4a0!

Looks much better now, thanks!  :+1:

PR #3367, opened 2024-01-11
[daemon] Avoid fixing state from off to stopped

ricab:
Avoid fixing the state of deleted instances from off to stopped, along with incoherent log "<instance> is deleted but has incompatible state 0, resetting state to 0 (stopped)".

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1480,11 +1480,14 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         allocated_mac_addrs = std::move(new_macs); // Add the new macs to the daemon's list only if we got this far
 
         // FIXME: somehow we're writing contradictory state to disk.
-        if (spec.deleted && spec.state != VirtualMachine::State::stopped)
+        if (spec.deleted && spec.state != VirtualMachine::State::stopped && spec.state != VirtualMachine::State::off)
         {
-            mpl::log(mpl::Level::warning, category,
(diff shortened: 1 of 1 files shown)


townsend2010:
Hey @sharder996, could you please review this one?  Thanks!

src/daemon/daemon.cpp
```diff
@@ -1480,11 +1480,14 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         allocated_mac_addrs = std::move(new_macs); // Add the new macs to the daemon's list only if we got this far
 
         // FIXME: somehow we're writing contradictory state to disk.
-        if (spec.deleted && spec.state != VirtualMachine::State::stopped)
-        {
-            mpl::log(mpl::Level::warning, category,
-                     fmt::format("{} is deleted but has incompatible state {}, resetting state to 0 (stopped)", name,
-                                 static_cast<int>(spec.state)));
+        if (spec.deleted && spec.state > VirtualMachine::State::stopped)
```
sharder996, line 1483: This is now reliant on the relative positioning of states in the State enum. Wouldn't checking for the Off state be better?

The enum order might be why the incoherent log was being printed in the first place.

sharder996 (changes requested):

src/daemon/daemon.cpp
ricab: Probably, yeah... OK, I'll fix it.

ricab: Done!

sharder996 (approved):
Looks good!

PR #3371, opened 2024-01-12
Filter new extra interfaces before instance start

luis4a0:
Adding interfaces after the instance was created introduced the possibility of some kinds of clashes when the instance starts. To fix this, we check first if the backend is able to add each one of the network interfaces requested by the user. The daemon then filters out the failing interfaces, letting only the interfaces which work be added to the instance.

This PR fixes issue 4 of [this review](https://github.com/canonical/multipass/pull/3195#pullrequestreview-1765804396).

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2179,7 +2179,7 @@ try // clang-format on
     std::vector<std::string> starting_vms{};
     starting_vms.reserve(instance_selection.operative_selection.size());
 
-    fmt::memory_buffer start_errors;
+    fmt::memory_buffer start_errors, start_warnings;
     for (auto& vm_it : instance_selection.operative_selection)
     {
         std::lock_guard lock{start_mutex};
@@ -2216,8 +2216,8 @@ try // clang-format on
 
             if (!failed_interfaces.empty())
             {
-                fmt::format_to(std::back_inserter(start_errors),
-                               "Cannot bridge {} for instance {}",
+                fmt::format_to(std::back_inserter(start_warnings),
+                               "Cannot bridge {} for instance {}. Please see the logs for more details.\n",
                                fmt::join(failed_interfaces, ", "),
                                name);
             }
@@ -2229,9 +2229,14 @@ try // clang-format on
     }
 
     auto future_watcher = create_future_watcher();
-    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<StartReply, StartRequest>, this,
-                                                server, starting_vms, timeout, status_promise,
-                                                fmt::to_string(start_errors)));
+    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<StartReply, StartRequest>,
+                                                this,
+                                                server,
+                                                starting_vms,
+                                                timeout,
+                                                status_promise,
+                                                fmt::to_string(start_errors),
+                                                fmt::to_string(start_warnings)));
 }
 catch (const std::exception& e)
 {
@@ -2333,8 +2338,13 @@ try // clang-format on
     }
 
     auto future_watcher = create_future_watcher();
-    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<RestartReply, RestartRequest>, this,
-                                                server, names_from(instance_targets), timeout, status_promise,
+    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<RestartReply, RestartRequest>,
+                                                this,
+                                                server,
+                                                names_from(instance_targets),
+                                                timeout,
+                                                status_promise,
+                                                std::string(),
                                                 std::string()));
 }
 catch (const std::exception& e)
@@ -2759,8 +2769,13 @@ void mp::Daemon::on_restart(const std::string& name)
         virtual_machine->state = VirtualMachine::State::running;
         virtual_machine->update_state();
     });
-    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<StartReply, StartRequest>, this,
-                                                nullptr, std::vector<std::string>{name}, mp::default_timeout, nullptr,
+    future_watcher->setFuture(QtConcurrent::run(&Daemon::async_wait_for_ready_all<StartReply, StartRequest>,
+                                                this,
+                                                nullptr,
+                                                std::vector<std::string>{name},
+                                                mp::default_timeout,
+                                                nullptr,
+                                                std::string(),
                                                 std::string()));
 }
 
@@ -2931,8 +2946,14 @@ void mp::Daemon::create_vm(const CreateRequest* request,
                         server->Write(reply);
                     });
                     future_watcher->setFuture(
-                        QtConcurrent::run(&Daemon::async_wait_for_ready_all<LaunchReply, LaunchRequest>, this, server,
-                                          std::vector<std::string>{name}, timeout, status_promise, std::string()));
+                        QtConcurrent::run(&Daemon::async_wait_for_ready_all<LaunchReply, LaunchRequest>,
+                                          this,
+                                          server,
+                                          std::vector<std::string>{name},
+                                          timeout,
+                                          status_promise,
+                                          std::string(),
+                                          std::string()));
                 }
                 else
                 {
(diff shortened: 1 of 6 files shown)


townsend2010:
Kicking this to have it build with the proper private branch.

townsend2010 (changes requested):
Hey @luis4a0!

I've tested this and I think we're getting closer. One thing though is the warning output is confusing, especially to end users. First, it *looks* like an error although it's not.  Second, the message is not clear *what* the problem is.

If I look at the available networks as instructed in the docs, I get the following:
```
$ multipass networks
Name     Type     Description
lxdbr0   bridge   Network bridge
mpbr0    bridge   Network bridge for Multipass
virbr0   bridge   Network bridge
```
That to me would mean `mpbr0` would be a candidate for setting as the bridged network.  However, with this fix and following Ricardo's issue #4 instructions, I of course get the following:
```
$ multipass start test-net
Cannot bridge mpbr0 for instance test-net
```
There isn't any explanation as to *why* this is problematic.  I think saying something to the effect of:
`Warning: Cannot bridge mpbr0 for instance test-net: Bridge mpbr0 already used by internal network`
or similar would be a bit more helpful.

As an aside and not related to this PR, I think showing `mpbr0` in `multipass networks` is a bug.

luis4a0:
Hey @townsend2010!

It's true that the message you propose looks better. But, for this, we should distinguish `mpbr0` from the rest of the bridges which cannot be used. Moreover, the LXD bridge can be used only once per instance. And the other bridges, like `virbr0`, can be used any number of times. So I think with one message like `Warning: cannot bridge mpbr0, it is already being used for instance test-net." will suffice, without distinguishing between LXD and Multipass bridges (honestly, I don't know what do they have which makes them impossible to use twice). What do you think?

src/daemon/daemon.cpp
```diff
@@ -3315,37 +3344,62 @@ mp::MountHandler::UPtr mp::Daemon::make_mount(VirtualMachine* vm, const std::str
...
-        for (const auto i : new_interfaces)
+                ++j;
+            }
+            else
+            {
+                // The generated MAC will not be used, so we remove it from the list of used addresses.
+                // TODO: avoid adding the new MAC to allocated_mac_addrs if the interface is bad.
+                allocated_mac_addrs.erase(iface.mac_address);
+
+                bad_bridges.insert(iface.id);
```
townsend2010, line 3384: You could add a `continue` in the failure block and tighten up the code here. Something like this:
```c++
    for (auto& iface : specs.extra_interfaces)
    {
        if (iface.mac_address.empty()) // An empty MAC address means the interface needs to be configured.
        {
            iface.mac_address = generate_unused_mac_address(allocated_mac_addrs);

            if (vm.add_network_interface(j, iface))
            {
                added_good_interfaces = true;

                commands.push_back(generate_netplan_script(j, iface.mac_address));
            }
            else
            {
                // The generated MAC will not be used, so we remove it from the list of used addresses.
                // TODO: avoid adding the new MAC to allocated_mac_addrs if the interface is bad.
                allocated_mac_addrs.erase(iface.mac_address);

                bad_bridges.insert(iface.id);

                continue;
            }
        }

        filtered_interfaces.push_back(iface);

        j++;
    }
```

And if it's possible to use the `try`/`catch` I have proposed, it would look something like:
```c++
    for (auto& iface : specs.extra_interfaces)
    {
        if (iface.mac_address.empty()) // An empty MAC address means the interface needs to be configured.
        {
            iface.mac_address = generate_unused_mac_address(allocated_mac_addrs);

            try
            {
                vm.add_network_interface(j, iface);
   
                added_good_interfaces = true;

                commands.push_back(generate_netplan_script(j, iface.mac_address));
            }
            catch (const LXDNetworkError&)
            {
                // The generated MAC will not be used, so we remove it from the list of used addresses.
                // TODO: avoid adding the new MAC to allocated_mac_addrs if the interface is bad.
                allocated_mac_addrs.erase(iface.mac_address);

                bad_bridges.insert(iface.id);

                continue;
            }
        }

        filtered_interfaces.push_back(iface);

        j++;
    }
```

src/daemon/daemon.cpp
```diff
@@ -3315,37 +3344,62 @@ mp::MountHandler::UPtr mp::Daemon::make_mount(VirtualMachine* vm, const std::str
                : vm->make_native_mount_handler(config->ssh_key_provider.get(), target, mount);
 }
 
-void mp::Daemon::configure_new_interfaces(const std::string& name, mp::VirtualMachine& vm, mp::VMSpecs& specs)
+// This function configures the network interfaces whose MAC address is empty. They will stay in specs.extra_interfaces
+// only if the backend is able to add them to the virtual machine. If not, they will be removed.
+std::unordered_set<std::string> mp::Daemon::configure_new_interfaces(const std::string& name,
```
townsend2010, line 3349: I think the `std::unordered_set<std::string>` return type should also be commented to indicated it's a set of interfaces that could not be configured and added.  Otherwise, without really reading the code, it would seem it would be a set of the interfaces that were configured.

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -487,7 +487,16 @@ void mp::LXDVirtualMachine::add_network_interface(int index, const mp::NetworkIn
 
     QJsonObject patch_json{{"devices", QJsonObject{{net_name, net_config}}}};
 
-    lxd_request(manager, "PATCH", url(), patch_json);
```
townsend2010, line 490: To go with my `try`/`catch` idea, I think letting the exception through here like before would be more desirable unless I'm missing something.

townsend2010 (changes requested):
Hey @luis4a0!

I've made some comments and suggestions below.

Also, regarding the warning message, I see what you mean.  I think we should even be more general here.

So maybe something like this:
```
Warning: Cannot bridge mpbr0 for instance test-net. Please see the logs for more details.
```
This way, we don't have to try to capture any backend specific error strings nor make generalizations that *may not* apply to other backends.

include/multipass/virtual_machine.h
```diff
@@ -84,7 +84,7 @@ class VirtualMachine : private DisabledCopyMove
     virtual void update_cpus(int num_cores) = 0;
     virtual void resize_memory(const MemorySize& new_size) = 0;
     virtual void resize_disk(const MemorySize& new_size) = 0;
-    virtual void add_network_interface(int index, const NetworkInterface& net) = 0;
+    virtual bool add_network_interface(int index, const NetworkInterface& net) = 0; // Returns true iff successful.
```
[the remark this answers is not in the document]
luis4a0: Yes, it's a possibility. The original motivation was that exceptions can be backend-specific, so we can't rely on which exception to catch, so a bool is the same thing. But OK, you're right in the sense that catching an exception simplifies the calling code. So I changed it back to `void`.

src/daemon/daemon.cpp
```diff
@@ -3315,37 +3344,62 @@ mp::MountHandler::UPtr mp::Daemon::make_mount(VirtualMachine* vm, const std::str
...
+    {
+        if (iface.mac_address.empty()) // An empty MAC address means the interface needs to be configured.
         {
-            new_interfaces.push_back(i);
+            iface.mac_address = generate_unused_mac_address(allocated_mac_addrs);
 
-            interface.mac_address = generate_unused_mac_address(allocated_mac_addrs);
-        }
-    }
+            if (vm.add_network_interface(j, iface))
```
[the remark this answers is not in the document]
luis4a0: The problem here is that we can't catch specifically the `LXDNetworkError` because it's a backend-specific exception. But yes, I changed that.

src/daemon/daemon.cpp
luis4a0: That looks much better. Changed. Thanks!

src/daemon/daemon.cpp
luis4a0: Right, done.

src/platform/backends/lxd/lxd_virtual_machine.cpp
luis4a0: Definitely. Reverted.

luis4a0:
Hey @townsend2010, thanks for the review! I addressed all your comments and implemented your suggestions.

src/daemon/daemon.cpp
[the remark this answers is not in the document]
townsend2010: Ok, just so I'm clear, *any* error that occurs when a backend configures a network is going to be treated basically as a warning and the user will just get the generic message that the network couldn't be configured and they should check the logs and the instance start won't be a failure, right?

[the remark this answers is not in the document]
luis4a0: Right, that's what's implemented now.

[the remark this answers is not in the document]
townsend2010: Ok, thanks for confirming.

townsend2010 (approved):
Hey @luis4a0!  This is much better, thank you!

Just a note that isn't related to this PR, but the log message that gets put in the journal when the interface isn't added is not complete/corrupted.  This is a snippet of what I see:
```
Jan 29 09:12:24 Razor-Crest systemd[5145]: Started snap.multipass.multipass-27753e77-b2fd-42e3-8275-436f01f59392.scope.
Jan 29 09:12:24 Razor-Crest multipassd[120858]: Using the 'default' storage pool.
 - Failed add validation for device "eth1": Instance DNS name "test1" conflict between "eth1" and "eth0" because both are connected to same networkroject=multipass: Conflict
Jan 29 09:12:24 Razor-Crest dnsmasq[4063]: read /etc/hosts - 7 addresses
```
As you can see, the warning is incomplete and some of the text is partially overwritten. This will need to be fixed since we are directing users to consult the log when they see the warning.

luis4a0:
Hey @townsend2010, thanks for reviewing again and approving!

Indeed, the message is incorrectly logged. I suspect a newline is lacking somewhere. I'll investigate it and create a new PR.

PR #3374, opened 2024-01-16
[Snapshots] Unit tests (take 2)

ricab:
New PR to see if it clears some codecov confusion.

### DIFF
--- a/src/platform/backends/shared/base_snapshot.cpp
+++ b/src/platform/backends/shared/base_snapshot.cpp
@@ -238,7 +238,7 @@ auto mp::BaseSnapshot::erase_helper()
     auto deleting_filepath = tmp_dir->filePath(snapshot_filename);
 
     QFile snapshot_file{snapshot_filepath};
-    if (!MP_FILEOPS.rename(snapshot_file, deleting_filepath))
+    if (!MP_FILEOPS.rename(snapshot_file, deleting_filepath) && snapshot_file.exists())
         throw std::runtime_error{
             fmt::format("Failed to move snapshot file to temporary destination: {}", deleting_filepath)};
 
--- a/tests/daemon_test_fixture.cpp
+++ b/tests/daemon_test_fixture.cpp
@@ -525,6 +525,31 @@ bool mpt::DaemonTestFixture::is_ready(std::future<R> const& f)
     return f.wait_for(std::chrono::seconds(5)) == std::future_status::ready;
 }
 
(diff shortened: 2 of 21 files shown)


ricab:
Didn't help codecov, so going back to the earlier one to keep history.

ricab:
Can't reopen the oher one...

tests/daemon_test_fixture.cpp
```diff
@@ -525,29 +525,6 @@ bool mpt::DaemonTestFixture::is_ready(std::future<R> const& f)
...
-    thread->start();
-
-    EXPECT_TRUE(is_ready(status_future));
-
-    thread->quit();
-    thread->wait();
-
-    return status_future.get();
-}
-
```
townsend2010, line 550: Why the move to the header?  There are the actual templates still defined below.  This was *supposed* to help with compiling *only* for the types we need, not every single type that could possibly be.

src/platform/backends/shared/base_snapshot.cpp
```diff
@@ -60,8 +60,8 @@ QJsonObject read_snapshot_json(const QString& filename)
 
     if (const auto json = QJsonDocument::fromJson(data, &parse_error).object(); parse_error.error)
         throw std::runtime_error{fmt::format("Could not parse snapshot JSON; error: {}; file: {}",
-                                             file.fileName(),
-                                             parse_error.errorString())};
+                                             parse_error.errorString(),
+                                             file.fileName())};
```
townsend2010, line 64: Oops :slightly_smiling_face:

townsend2010 (commented):
Hey @ricab!

Thanks for this- very nice!  I enjoyed the references to pirate stuff, The Matrix, etc :laughing: 

I just really have one question below and once we get that squared away, I think this is good.

tests/daemon_test_fixture.cpp
ricab: Hi Chris, I kept wondering how this could have worked before, I totally missed those declarations below. The template would still only be instantiated for the types we used it with, but that could happen multiple times for the same types, which the current scheme avoids. I'll revert and adhere to it as well instead.

townsend2010: Right, thanks!

ricab:
@townsend2010, it should be good now. Thanks!

townsend2010 (approved):
Hey @ricab!

Great work!  LGTM :+1:

PR #3388, opened 2024-01-29
Fix showing alias message on Blueprint creation

luis4a0:
The alias message was not shown when launching a Blueprint which defined aliases. The issue was that it was shown only when the size of the alias dictionary was one, and in fact it would be two when the new alias is in a context different to `default`.

This bug was reported [here](https://discourse.ubuntu.com/t/running-a-container-with-the-docker-blueprint-in-multipass/26806/12).


townsend2010 (approved):
Hey @luis4a0!

Thanks for this!  :+1:

PR #3393, opened 2024-01-30
[daemon] Add pollinate package to internal cloud-init

### LINKED ISSUE #3390: Images used for `snapcraft` do not report pollinate data
The buildd images used for the `snapcraft` lack to pollinate package.  We should seed `pollinate` in our `cloud-init` in order to get the pollinate data.

townsend2010:
This will add the pollinate package to the snapcraft buildd images. It is no-op on the regular cloud images since it already exists.

Fixes #3390

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -143,6 +143,7 @@ auto make_cloud_init_vendor_config(const mp::SSHKeyProvider& key_provider, const
     config["ssh_authorized_keys"].push_back(ssh_key_line);
     config["timezone"] = request->time_zone();
     config["system_info"]["default_user"]["name"] = username;
+    config["packages"].push_back("pollinate");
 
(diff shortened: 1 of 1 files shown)


townsend2010:
Hey @sharder996, could you please review this at your earliest convenience?  Thanks!

sharder996 (approved):
LGTM!

mytchh:
Is there an option available to not force pollinate package in merged cloud-init?

I'm not sure if valid use-case... but I think this breaks install of other packages when image deployed via multipass does not have package manager with sources for pollinate.

E.g. deploying alpine imagine with cloud-init, I can define apk packages to install, but now pollinate is added, causing all other packages to fail install.

ricab:
Hi @mytchh, I am afraid that only ubuntu images are currently supported. To support other images, we would need to generalize this and other things in the code.

Issue #3399, opened 2024-02-05
[regression] Unhandled exception when converting qcow2 images for snapshot support


kyanny:
**Describe the bug**

`multipass` command stopped working. (was working before)

**To Reproduce**

```
❯ multipass list
list failed: cannot connect to the multipass socket

❯ multipass info
info failed: cannot connect to the multipass socket
```

**Expected behavior**

`multipass list` displays list of VM names, and so on.

**Logs**
Please provide logs from the daemon, see [accessing logs](https://multipass.run/docs/accessing-logs) on where to find them on your platform.

[multipassd.log](https://github.com/canonical/multipass/files/14164074/multipassd.log)

**Additional info**
 - OS: macOS 14.3
   - Processor `2.3 GHz Quad-Core Intel Core i7`
   - Graphics `Intel Iris Plus Graphics 1536 MB`
   - Memory `32 GB 3733 MHz LPDDR4X`
- `multipass version` `1.13.0+mac`
- `multipass info --all`

    ```
    ❯ multipass info --all
    Warning: the `--all` flag for the `info` command is deprecated. Please use `info` with no positional arguments for the same effect.
    info failed: cannot connect to the multipass socket
    ```

- `multipass get local.driver`

    ```
    ❯ multipass get local.driver
    get failed: cannot connect to the multipass socket
    ```

**Additional context**

I installed Multipass via Homebrew.
I have uninstalled/reinstalled Multipass but the problem still persists.

townsend2010:
Hi @kyanny!

Sorry you are having this issue.  Looking at the log you provided, it seems there is an issue with your `primary` instance image and we aren't handling this correctly.  This was added in the 1.13 release, so I'm marking this a regression and it will be fixed in 1.13.1.

townsend2010:
Hi @kyanny,

Could you please confirm that the `primary` instance existed before you upgrade to 1.13.0 and then after the upgrade, this stopped working?  Thank you!

ricab:
Hi @kyanny, could you please paste the output of the following command?

```
qemu-img info "/var/root/Library/Application Support/multipassd/qemu/vault/instances/primary/ubuntu-22.04-server-cloudimg-amd64.img"
```

If that also fails, do you remember how large the instance was (more or less)?

kyanny:
Hi @townsend2010 thank you for your response.

> Could you please confirm that the `primary` instance existed before you upgrade to 1.13.0 and then after the upgrade, this stopped working? Thank you!

[...]

PR #3400, opened 2024-02-06
Handle failure to convert to QCOW2 v3

ricab:
Addresses #3399.

I am not super happy about this approach, as it goes against *fail fast*, breaking preconditions for other operations (snapshots+resizing) and leaving a VM around in a bad state. A better approach would be to put any VMs whose constructor threw into a special quarantine state, where only a select few operations would be allowed. But we don't have time for that right now and the operation in question is transitory anyway, so I guess this will do.


townsend2010 (approved):
Hey @ricab!

Thanks for this!  I agree it's not the most desirable approach, but it does the job.  :+1:

PR #3404, opened 2024-02-12
Snapshot extra interfaces

luis4a0:
This PR adds the `extra_interfaces` data structure to the snapshots infrastructure. This is needed to snapshot instance with interfaces added after creation.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -658,6 +658,13 @@ mp::QemuVirtualMachine::MountArgs& mp::QemuVirtualMachine::modifiable_mount_args
     return mount_args;
 }
 
+void mp::QemuVirtualMachine::restore_snapshot(const std::string& name, VMSpecs& specs)
+{
+    BaseVirtualMachine::restore_snapshot(name, specs);
+
+    desc.extra_interfaces = specs.extra_interfaces;
+}
+
 auto mp::QemuVirtualMachine::make_specific_snapshot(const std::string& snapshot_name,
                                                     const std::string& comment,
                                                     const VMSpecs& specs,
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -658,13 +658,6 @@ mp::QemuVirtualMachine::MountArgs& mp::QemuVirtualMachine::modifiable_mount_args
     return mount_args;
 }
 
-void mp::QemuVirtualMachine::restore_snapshot(const std::string& name, VMSpecs& specs)
-{
-    BaseVirtualMachine::restore_snapshot(name, specs);
-
-    desc.extra_interfaces = specs.extra_interfaces;
-}
-
 auto mp::QemuVirtualMachine::make_specific_snapshot(const std::string& snapshot_name,
                                                     const std::string& comment,
                                                     const VMSpecs& specs,
--- a/src/platform/backends/qemu/qemu_virtual_machine.h
+++ b/src/platform/backends/qemu/qemu_virtual_machine.h
@@ -69,6 +69,7 @@ public:
     virtual MountArgs& modifiable_mount_args();
     std::unique_ptr<MountHandler> make_native_mount_handler(const SSHKeyProvider* ssh_key_provider,
                                                             const std::string& target, const VMMount& mount) override;
+    void restore_snapshot(const std::string& name, VMSpecs& specs) override;
 
 signals:
     void on_delete_memory_snapshot();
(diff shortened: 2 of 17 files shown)


src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -515,6 +515,11 @@ void BaseVirtualMachine::restore_rollback_helper(const Path& head_path,
 }
 
 void BaseVirtualMachine::restore_snapshot(const std::string& name, VMSpecs& specs)
+{
+    restore_snapshot_impl(name, specs);
+}
+
+void BaseVirtualMachine::restore_snapshot_impl(const std::string& name, VMSpecs& specs)
```
ricab, line 522: Why did you choose to add this? I understand that extra interfaces aren't supported on all backends, but on those that don't support it, they would just be empty, so the generic handling would apply (with most likely insignificant detriment to perf), no?

ricab (commented):
Hey @luis4a0, good work :slightly_smiling_face: I did just a quick pass and it looks good to me overall. One question inline.

src/platform/backends/shared/base_virtual_machine.cpp
luis4a0: Hey @ricab, I did this because QEMU backend does exactly what this function does, plus modifying an internal data structure. It was the easiest way of reusing the code in other backend.

ricab: Hey @luis4a0, I still don't understand this. You could just call `BaseVM::restore` from `QemuVM::restore`, you don't need a separate impl method for that. However, I don't think you need even that. You can just update extra interfaces together with other `desc` fields [in the snapshot](https://github.com/canonical/multipass/blob/acf850a59f5bee58ae0992fe4f069ed101e37e31/src/platform/backends/qemu/qemu_snapshot.cpp#L101).

ricab: In fact you already update the `desc` in the `QemuSnapshot::apply`, which is [called from restore](https://github.com/canonical/multipass/blob/922646dc81ee127032ba0501bcce36e41138cd1d/src/platform/backends/shared/base_virtual_machine.cpp#L542).

ricab:
Hey @luis4a0, not sure if it affects this PR, but here is one more thing to consider: how do you plan to handle existing snapshots? For instance, suppose I do this today:

```
multipass launch --network eth0 -n foo
multipass launch -n bar
multipass snapshot -n s1 foo
multipass snapshot -n s1 bar
```

After we release bridge addition, I then do

```
multipass restore foo.s1
multipass restore bar.s1
```

How will you ensure that `foo` will have `eth0`, but not `bar`?

src/platform/backends/shared/base_virtual_machine.cpp
luis4a0: Hey @ricab, thanks! Right, I removed the implementation of `QemuVirtualMachine::restore_snapshot()`. I am currently testing to see if I'm not forgetting something. Will let you know when I'm OK with the code.

luis4a0:
@ricab, right. This needs to be considered, but in another PR. I'd say we can merge this one and start testing and addressing this other issue. Thanks!

luis4a0:
Hey @ricab, thanks for the review! The case you mention is exactly the one I am considering for the next PR. To solve that, two things should be done:
1. snapshot `run_at_boot`,
2. run `configure_new_interfaces`.
The second point is still obscure to me, it should be ran at start but it is not. The proof it's not ran is that the backend tries to launch QEMU with an empty MAC address, what means that the interface was not configured.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -30,6 +30,7 @@
 #include <multipass/platform.h>
 #include <multipass/utils.h>
 #include <multipass/vm_mount.h>
+#include <multipass/vm_specs.h>
```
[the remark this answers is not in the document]
luis4a0: Godd catch, removed.

src/platform/backends/qemu/qemu_virtual_machine.h
```diff
@@ -68,7 +68,8 @@ class QemuVirtualMachine : public QObject, public BaseVirtualMachine
     virtual void add_network_interface(int index, const NetworkInterface& net) override;
     virtual MountArgs& modifiable_mount_args();
     std::unique_ptr<MountHandler> make_native_mount_handler(const SSHKeyProvider* ssh_key_provider,
-                                                            const std::string& target, const VMMount& mount) override;
+                                                            const std::string& target,
+                                                            const VMMount& mount) override;
```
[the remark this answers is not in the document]
luis4a0: Good catch, reverted.

ricab (approved):
Alright, apart from the half-state issue (which is going to get fixed elsewhere), I think this is good to go. Thank you @luis4a0!

PR #3405, opened 2024-02-12
[snap] Improve snap packaging

townsend2010:
This will quiet the errors and linter.  The only warning now is for libvirt.so which is still needed since it is dlopen()'d.  This reduces the Snap size by ~45%!

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -321,7 +321,6 @@ parts:
       - dmz-cursor-theme
       - libgdk-pixbuf2.0-bin # GDK pixbuf
       - fonts-freefont-ttf
-      - libglib2.0-0
     override-build: |
       craftctl default
       update-mime-database $CRAFT_PART_INSTALL/usr/share/mime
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -13,9 +13,15 @@ adopt-info: multipass
 confinement: strict
 base: core22
 
+architectures:
```
sharder996, line 16: What determines architecture support? Is it based on the backend drivers that we use?

snap/snapcraft.yaml
```diff
@@ -262,6 +302,16 @@ parts:
     override-pull: ""
     stage-packages:
     - on amd64: [msr-tools]
+    stage:
+    - -usr/share/doc
+    - -usr/share/man
+
+  compile-schemas:
```
sharder996, line 309: What is compile-schemas? What does it do?

snap/snapcraft.yaml
```diff
@@ -122,21 +135,47 @@ parts:
     - qt6-base-dev-tools
     stage-packages:
     - apparmor
-    - on amd64: [libgl1]
```
sharder996, line 125: How did determine that some packages were not needed?

snap/snapcraft.yaml
```diff
@@ -122,21 +135,47 @@ parts:
...
+    - -usr/lib/*/libGLX_indirect.so.*
+    - -usr/lib/*/libXxf86vm.so.*
+    - -usr/lib/*/libxcb-dri2.so.*
+    - -usr/lib/*/libxcb-present.so.*
+    - -usr/lib/*/libxshmfence.so.*
+    - -usr/lib/*/libicuio.so.*
+    - -usr/lib/*/libicutest.so.*
+    - -usr/lib/*/libicutu.so.*
+    - -usr/lib/*/libLLVM-15.so*
+    - -usr/lib/*/libdrm*
```
sharder996, line 165: How did you find the different parts of a package? And then know which parts we actually need?

sharder996 (approved):
This looks good to me!

I just have a few questions for my own learning.

snap/snapcraft.yaml
townsend2010: It's based on what Multipass can actually build on.  Without this, it would try to build on the default architectures on Launchpad (amd64, arm64, armhf, ppc64el, s390x), but it is broken on armhf due to the lack of `kvm` support there.  So now I made it explicit so it won't try to build on armhf and always fail.

snap/snapcraft.yaml
townsend2010: I basically split this out from the `terminal` part in order to not pull in all of libglib.  It's easier to just pull this out of `terminal` and *only* specify the one file that is needed instead of leaving it in `terminal` and having to specify **all** of the files we **don't** want.

At the heart of it, [`glib-compile-schemas`](https://manpages.ubuntu.com/manpages/trusty/en/man1/glib-compile-schemas.1.html) is needed for `xfce4-terminal` to operate correctly.

snap/snapcraft.yaml
townsend2010: Well, the Snapcraft linter helped with some of it and then a bunch of trial and error.

snap/snapcraft.yaml
townsend2010: Again, it was the Snapcraft linter giving some clues along with trial and error.

sharder996:
@townsend2010, Thanks for replying to all my comments!

PR #3424, opened 2024-03-04
Snapshot boot commands

luis4a0:
This PR adds the ability to snapshot the commands to be run at boot on each instance that the daemon keeps.


townsend2010:
Hi @luis4a0!

I don't remember now, but did we decide not to do this?

PR #3433, opened 2024-03-12
Snapshot restore applies cloud init change

georgeliao:
This PR is about applying the extra interface changes to cloud-init-config.iso file when the snapshot retore command is called.

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -85,8 +85,8 @@ public:
     virtual void resize_memory(const MemorySize& new_size) = 0;
     virtual void resize_disk(const MemorySize& new_size) = 0;
     virtual void add_network_interface(int index, const NetworkInterface& net) = 0;
-    virtual void add_extra_interfaces_to_cloud_init(const std::string& default_mac_addr,
-                                                    const std::vector<NetworkInterface>& extra_interfaces) = 0;
+    virtual void apply_extra_interfaces_to_cloud_init(const std::string& default_mac_addr,
+                                                      const std::vector<NetworkInterface>& extra_interfaces) = 0;
     virtual std::unique_ptr<MountHandler> make_native_mount_handler(const SSHKeyProvider* ssh_key_provider,
                                                                     const std::string& target,
                                                                     const VMMount& mount) = 0;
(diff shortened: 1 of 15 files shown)


include/multipass/virtual_machine.h
```diff
@@ -85,8 +85,8 @@ class VirtualMachine : private DisabledCopyMove
     virtual void resize_memory(const MemorySize& new_size) = 0;
     virtual void resize_disk(const MemorySize& new_size) = 0;
     virtual void add_network_interface(int index, const NetworkInterface& net) = 0;
-    virtual void add_extra_interfaces_to_cloud_init(const std::string& default_mac_addr,
-                                                    const std::vector<NetworkInterface>& extra_interfaces) = 0;
+    virtual void apply_extra_interfaces_to_cloud_init(const std::string& default_mac_addr,
+                                                      const std::vector<NetworkInterface>& extra_interfaces) = 0;
```
georgeliao, line 89: Name prefix changed from add to apply which better reflects what the function is doing.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -541,6 +541,7 @@ void BaseVirtualMachine::restore_snapshot(const std::string& name, VMSpecs& spec
     specs.num_cores = snapshot->get_num_cores();
     specs.mem_size = snapshot->get_mem_size();
     specs.disk_space = snapshot->get_disk_space();
+    const bool is_extra_interfaces_different = specs.extra_interfaces != snapshot->get_extra_interfaces();
```
luis4a0, line 544: I'd change this to `are_extra_interfaces_different`. I know it's a data structure, but it sounds better to me to have the plural.

luis4a0 (changes requested):
Hey @georgeliao! This looks good. I've made a minor comment inline about a variable name change. Then, I'd like you to rebase on the current code of `cloud-init-extra-interfaces-config_plus_snapshot`, so we have the current code OK. Finally, I have to make some more functional tests, but everything seems OK. Thanks for this!

src/platform/backends/shared/base_virtual_machine.cpp
georgeliao: Yep, that is a better name. Will do.

luis4a0 (approved):
Alright @georgeliao, this works as intended, does what it is supposed to so we are ready to merge. Thanks for this!

PR #3437, opened 2024-03-14
Fix handling of `~` in Blueprint launch from file.

### LINKED ISSUE #3421: Incomplete handling of file URLs
**Describe the bug**
The CLI cannot expand paths containing `~`, or handle hidden folders in file URLs.

**To Reproduce**
Create a minimal blueprint `~/blueprint.yaml` .

Using `~`:
```bash
multipass launch file://~/blueprint.yaml

launch failed: Wrong file '/home/alex/~/blueprint.yaml'
```

Hidden file
```bash
mv blueprint.yaml .blueprint.yaml
multipass launch file://.blueprint.yaml

[2024-02-29T10:14:52.896] [error] [blueprint provider] Invalid Blueprint name '': must be a valid host name
launch failed: Invalid Blueprint name '': must be a valid host name
```

Hidden folder
```bash
mkdir .directory
mv .blueprint.yaml .directory/blueprint.yaml
multipass launch file://.directory/blueprint.yaml

launch failed: Wrong file '/home/alex/.directory/blueprint.yaml'
```

**Logs**

<details><summary>Multipass logs</summary>

```text
Feb 29 09:05:27 alex-frame multipassd[1887]: Starting Multipass 1.13.1
Feb 29 09:05:27 alex-frame multipassd[1887]: Daemon arguments: /snap/multipass/11732/bin/multipassd --v>
Feb 29 09:05:32 alex-frame dnsmasq[3172]: reading /etc/resolv.conf
Feb 29 09:05:32 alex-frame dnsmasq[3172]: using nameserver 127.0.0.53#53
Feb 29 09:05:32 alex-frame dnsmasq[3172]: using only locally-known addresses for multipass
Feb 29 09:05:32 alex-frame dnsmasq[3172]: reading /etc/resolv.conf
Feb 29 09:05:32 alex-frame dnsmasq[3172]: using nameserver 127.0.0.53#53
Feb 29 09:05:32 alex-frame dnsmasq[3172]: using only locally-known addresses for multipass
Feb 29 09:20:28 alex-frame multipassd[1887]: fetch manifest periodically
Feb 29 09:35:28 alex-frame multipassd[1887]: fetch manifest periodically
Feb 29 09:50:28 alex-frame multipassd[1887]: fetch manifest periodically
Feb 29 10:05:28 alex-frame multipassd[1887]: fetch manifest periodically
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "anbox-cloud-appliance" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "charm-dev" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "docker" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "jellyfin" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "minikube" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "ros-noetic" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: Loading "ros2-humble" v1
Feb 29 10:14:22 alex-frame multipassd[1887]: [92399] started: qemu-system-x86_64 --version
Feb 29 10:14:22 alex-frame multipassd[1887]: Reading Blueprint 'blueprint' from file /home/alex/~/bluep>
Feb 29 10:14:52 alex-frame multipassd[1887]: [93182] started: qemu-system-x86_64 --version
Feb 29 10:14:52 alex-frame multipassd[1887]: Reading Blueprint '.blueprint' from file /home/alex/.bluep>
Feb 29 10:14:52 alex-frame multipassd[1887]: Invalid Blueprint name '': must be a valid host name
Feb 29 10:16:28 alex-frame multipassd[1887]: [95638] started: qemu-system-x86_64 --version
Feb 29 10:16:28 alex-frame multipassd[1887]: Reading Blueprint '.blueprint' from file /home/alex/direct>
Feb 29 10:16:28 alex-frame multipassd[1887]: Invalid Blueprint name '': must be a valid host name
Feb 29 10:16:54 alex-frame multipassd[1887]: [96352] started: qemu-system-x86_64 --version
Feb 29 10:16:54 alex-frame multipassd[1887]: Reading Blueprint '.blueprint' from file /home/alex/.direc>
Feb 29 10:16:54 alex-frame multipassd[1887]: Invalid Blueprint name '': must be a valid host name
Feb 29 10:17:19 alex-frame multipassd[1887]: [96953] started: qemu-system-x86_64 --version
Feb 29 10:17:19 alex-frame multipassd[1887]: Reading Blueprint 'blueprint' from file /home/alex/.direct>
Feb 29 10:20:28 alex-frame multipassd[1887]: fetch manifest periodically
Feb 29 10:25:59 alex-frame multipassd[1887]: Returning setting local.driver=qemu
lines 10323-10381/10381 (END)
```

</details>

**Additional info**
 - OS: Ubuntu 23.10 (Mantic)
- `multipass version`
multipass   1.13.1  
multipassd  1.13.1
- `multipass info`
No instances found.
qemu

luis4a0:
`QFileInfo does` not expand `~`, so we need to do it by hand when necessary.

Fixes #3421.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -262,7 +262,11 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
         if (remote_image_name.startsWith("file://"))
         {
             // Convert to absolute because the daemon doesn't know where the client is being executed.
-            auto file_info = QFileInfo(remote_image_name.remove(0, 7));
+            if (remote_image_name.remove(0, 7)[0] == '~')
+            {
+                remote_image_name.replace(0, 1, QDir::homePath());
+            }
+            auto file_info = QFileInfo(remote_image_name);
             auto absolute_file_path = file_info.absoluteFilePath();
 
             request.set_image("file://" + absolute_file_path.toStdString());


src/client/cli/cmd/launch.cpp
```diff
@@ -262,7 +262,11 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
         if (remote_image_name.startsWith("file://"))
         {
             // Convert to absolute because the daemon doesn't know where the client is being executed.
-            auto file_info = QFileInfo(remote_image_name.remove(0, 7));
+            if (remote_image_name.remove(0, 7)[0] == '~')
+            {
+                remote_image_name.replace(0, 1, QDir::homePath());
+            }
+            auto file_info = QFileInfo(remote_image_name);
```
ricab, line 269: Can't we use an existing utility to parse `file://` urls? There will be other things that this does not cover and we shouldn't have to implement all of them.

luis4a0: Hey @ricab, thanks for reviewing this! Right, it would be cleaner to do that. I investigated on this, maybe `POCO::Uri` can help us with `file://`. But, on the other hand, `~` is usually expanded directly by the shell, and having `file://` in front of it makes the shell think it is not used as a home folder. Maybe that's the reason why Qt and `std::filesystem` do not implement `~` expansion.

So moving to an external utility to handle `file://` would not solve our original problem, the `~` expansion. I think making special code to handle this only case (i.e., that the character after `file://` is a `~`) is the way to go. But maybe there is something I am missing. WDYT?

luis4a0: That said, let me point out that `QUrl` can handle `file://` and recognize if a given string is a local file or not. See [here](https://doc.qt.io/qt-6/qurl.html#isLocalFile).

luis4a0: Ohhh, no. `QUrl` expects `file:` instead of `file://`: we are on the same situation than before.

ricab: Hmm, honestly I think we should delegate on something that behaves (mostly) correctly and not try to support anything further. If Qt and Poco do not expand `~`, do we really need to? We would then need to [go dig](https://url.spec.whatwg.org/) to find exactly what is and isn't a valid URL. That is a complex standard and we should not be in the business of handling that type of detail.

luis4a0: I agree that can leave this to the shell and if the shell doesn't expand it, then it's not our problem. For example, on Firefox, writing in the address bar `file://~/multipass-coverage/index.html ` does not expand `~`. So leaving to the shell would be correct.

OTOH, I think every Linux user (and I am speaking about Linux because for the time being launching blueprints from file is only supported in Linux) expects `~` to be expanded as the home folder. So deciding whether or not to implement this expansion is a matter of being friendly to users or being conservative. I don't have a strong opinion on this, honestly ;)

luis4a0: Oh, I answered your first comment without reading the second one (it just arrived). Yeah, I'm seeing there are some options to read better the file://, but the point is to decide whether we want to expand `~`. If not, we can leave everything as it is now and forget about this PR.

luis4a0: @ricab, something more about this: if we choose to expand `~`, we'll have to expand also `~user`, and I think that's not what we are supposed to do. So I think we shouldn't expand `~`, but leave it to the shell. And, considering the second and third issues the user reported are not bugs, I vote for closing this PR. WDYT?

ricab: OK, I am fine with that. I am not sure how `~` and `~user` are supposed to be handled in the URL standard to be honest. I wouldn't mind seeing us delegating URL parsing, but we don't have to do it here or now.

luis4a0:
Right, closing for now. Thanks for your input!

PR #3439, opened 2024-03-15
Handle legacy snapshots

luis4a0:
Snapshots taken with old versions of Multipass did not store the `extra_interfaces` field. If an interface is added after instance creation, it will overwrite the extra interfaces added at creation. This PR fixes that behavior.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -320,16 +320,17 @@ std::unordered_map<std::string, mp::VMSpecs> load_db(const mp::Path& data_path, 
             mounts[json["target_path"].toString().toStdString()] = mp::VMMount{json};
         }
 
-        reconstructed_records[key] = {num_cores,
-                                      mp::MemorySize{mem_size.empty() ? mp::default_memory_size : mem_size},
-                                      mp::MemorySize{disk_space.empty() ? mp::default_disk_size : disk_space},
-                                      default_mac_address,
-                                      MP_JSONUTILS.read_extra_interfaces(record),
-                                      ssh_username,
-                                      static_cast<mp::VirtualMachine::State>(state),
-                                      mounts,
-                                      deleted,
-                                      metadata};
+        reconstructed_records[key] = {
+            num_cores,
+            mp::MemorySize{mem_size.empty() ? mp::default_memory_size : mem_size},
+            mp::MemorySize{disk_space.empty() ? mp::default_disk_size : disk_space},
+            default_mac_address,
+            MP_JSONUTILS.read_extra_interfaces(record).value_or(std::vector<mp::NetworkInterface>{}),
+            ssh_username,
+            static_cast<mp::VirtualMachine::State>(state),
+            mounts,
+            deleted,
+            metadata};
     }
     return reconstructed_records;
 }
--- a/src/platform/backends/shared/base_snapshot.cpp
+++ b/src/platform/backends/shared/base_snapshot.cpp
@@ -166,27 +166,27 @@ mp::BaseSnapshot::BaseSnapshot(const std::string& name,
     assert(index > 0 && "snapshot indices need to start at 1");
 }
 
-mp::BaseSnapshot::BaseSnapshot(const QString& filename, VirtualMachine& vm)
-    : BaseSnapshot{read_snapshot_json(filename), vm}
+mp::BaseSnapshot::BaseSnapshot(const QString& filename, VirtualMachine& vm, VirtualMachineDescription& desc)
+    : BaseSnapshot{read_snapshot_json(filename), vm, desc}
 {
 }
 
-mp::BaseSnapshot::BaseSnapshot(const QJsonObject& json, VirtualMachine& vm)
+mp::BaseSnapshot::BaseSnapshot(const QJsonObject& json, VirtualMachine& vm, VirtualMachineDescription& desc)
     : BaseSnapshot{
-          json["name"].toString().toStdString(),                                                  // name
-          json["comment"].toString().toStdString(),                                               // comment
-          find_parent(json, vm),                                                                  // parent
-          json["index"].toInt(),                                                                  // index
-          QDateTime::fromString(json["creation_timestamp"].toString(), Qt::ISODateWithMs),        // creation_timestamp
-          json["num_cores"].toInt(),                                                              // num_cores
-          MemorySize{json["mem_size"].toString().toStdString()},                                  // mem_size
-          MemorySize{json["disk_space"].toString().toStdString()},                                // disk_space
-          MP_JSONUTILS.read_extra_interfaces(json).value_or(std::vector<mp::NetworkInterface>{}), // extra_interfaces
-          static_cast<mp::VirtualMachine::State>(json["state"].toInt()),                          // state
-          load_mounts(json["mounts"].toArray()),                                                  // mounts
-          json["metadata"].toObject(),                                                            // metadata
-          vm.instance_directory(),                                                                // storage_dir
-          true}                                                                                   // captured
+          json["name"].toString().toStdString(),                                           // name
+          json["comment"].toString().toStdString(),                                        // comment
+          find_parent(json, vm),                                                           // parent
+          json["index"].toInt(),                                                           // index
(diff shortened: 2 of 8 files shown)


src/platform/backends/shared/base_snapshot.cpp
```diff
@@ -166,12 +166,12 @@ mp::BaseSnapshot::BaseSnapshot(const std::string& name,
...
 
-mp::BaseSnapshot::BaseSnapshot(const QString& filename, VirtualMachine& vm)
-    : BaseSnapshot{read_snapshot_json(filename), vm}
+mp::BaseSnapshot::BaseSnapshot(const QString& filename, VirtualMachine& vm, VirtualMachineDescription& desc)
+    : BaseSnapshot{read_snapshot_json(filename), vm, desc}
 {
 }
 
-mp::BaseSnapshot::BaseSnapshot(const QJsonObject& json, VirtualMachine& vm)
+mp::BaseSnapshot::BaseSnapshot(const QJsonObject& json, VirtualMachine& vm, VirtualMachineDescription& desc)
```
ricab, line 174: You should be able to make the new parameter `const`.

src/platform/backends/shared/base_snapshot.h
```diff
@@ -21,6 +21,7 @@
 #include <multipass/snapshot.h>
 
 #include <multipass/memory_size.h>
+#include <multipass/virtual_machine_description.h>
```
ricab, line 24: Small detail, but you should be able to fwd declare `VirtualMachineDescription` here and include only in the `cpp`.

src/platform/backends/shared/base_snapshot.cpp
```diff
@@ -181,7 +181,7 @@ mp::BaseSnapshot::BaseSnapshot(const QJsonObject& json, VirtualMachine& vm)
           json["num_cores"].toInt(),                                                       // num_cores
           MemorySize{json["mem_size"].toString().toStdString()},                           // mem_size
           MemorySize{json["disk_space"].toString().toStdString()},                         // disk_space
-          MP_JSONUTILS.read_extra_interfaces(json),                                        // extra_interfaces
+          MP_JSONUTILS.read_extra_interfaces(json).value_or(desc.extra_interfaces),        // extra_interfaces
```
ricab, line 184: There is only one step missing: you need to call `BaseSnapshot::persist` if the networks were empty. 

I think the most elegant way to achieve that is in the constructor being called here, i.e. the one in line 99. You would receive the optional there and call `value_or` in that other constructor. Then, in the body of the constructor, check if empty, assert captured and persist. Please let me know if that makes sense.

ricab (changes requested):
Hey @luis4a0, thank you for this work! I haven't manually tested but looking at the code I think it is almost good to go.

There is one crucial step missing: we need to persist the snapshot when we fill in the networks from the VM. Otherwise, we'd use the networks from the VM the following time too (the user could have changed them by then).

Other than that, only a couple of small details.

src/platform/backends/shared/base_snapshot.cpp
luis4a0: Right, done.

src/platform/backends/shared/base_snapshot.h
luis4a0: Right, done.

src/platform/backends/shared/base_snapshot.cpp
luis4a0: Yes, that makes sense. Just to be clear, in your first statement: we need to call `persist()` if the networks were `nullopt`, not an empty vector, right?

luis4a0: The only problem with doing that is that the constructor in line 99 (now 100) does not have information to infer the extra interfaces. Why not leaving that as it is in the current constructor, and calling `persist()` from the current constructor?

Adding a mew parameter to the constructor in line 99 would alter the rest of the code, because it is also called from another constructor which does not have the information needed.

ricab: > Just to be clear, in your first statement: we need to call persist() if the networks were nullopt, not an empty vector, right?

Yep, that's right.

> The only problem with doing that is that the constructor in line 99 (now 100) does not have information to infer the extra interfaces. Why not leaving that as it is in the current constructor, and calling persist() from the current constructor?

Good point. OK, I guess you'll need to default-initialize the field and then add logic in the body of the constructor for this. I was trying to avoid that, but the alternatives aren't pretty either. I agree we shouldn't pass the new parameter to the other constructor and we should not add an another intermediate constructor for this either. So I think your proposal is a good compromise here.

src/platform/backends/shared/base_snapshot.h
```diff
@@ -40,7 +42,7 @@ class BaseSnapshot : public Snapshot
                  std::shared_ptr<Snapshot> parent,
                  const VMSpecs& specs,
                  const VirtualMachine& vm);
-    BaseSnapshot(const QString& filename, VirtualMachine& vm);
+    BaseSnapshot(const QString& filename, VirtualMachine& vm, VirtualMachineDescription& desc);
```
ricab, line 45: `const` missing here too.

luis4a0: Right, done.

src/platform/backends/shared/base_snapshot.cpp
luis4a0: Good. What about what I implemented [here](https://github.com/canonical/multipass/pull/3439/commits/76b6b58243dcd675d8aedc888b320c642f98373b)?

ricab: Oh OK. That's two json interrogations for the same field, but the alternative involves a 2-step "initialization" of the vector field, with possible growth, so your option is probably better :+1:

ricab (commented):
Thanks for the changes @luis4a0! I still need to test, but I think this is good now. I'll submit a new review after testing.

luis4a0:
Great, thanks @ricab! I'll rebase the private branch, there is some Hyper-V code also. Thanks!

src/daemon/daemon.cpp
```diff
@@ -320,16 +320,17 @@ std::unordered_map<std::string, mp::VMSpecs> load_db(const mp::Path& data_path,
...
-                                      static_cast<mp::VirtualMachine::State>(state),
-                                      mounts,
-                                      deleted,
-                                      metadata};
+        reconstructed_records[key] = {
+            num_cores,
+            mp::MemorySize{mem_size.empty() ? mp::default_memory_size : mem_size},
+            mp::MemorySize{disk_space.empty() ? mp::default_disk_size : disk_space},
+            default_mac_address,
+            MP_JSONUTILS.read_extra_interfaces(record).value_or(std::vector<mp::NetworkInterface>{}),
```
georgeliao, line 328: I think the origninal `MP_JSONUTILS.read_extra_interfaces` function which returns `std::vector<mp::NetworkInterface>` instead of `std::optional<std::vector<mp::NetworkInterface>>` does this already, right? When `record.contains("extra_interfaces")` is false, then the function returns an empty vector, which is essentially the `std::optional` + `value_or(std::vector<mp::NetworkInterface>{})` does.

luis4a0: Thanks @georgeliao! The point is that `read_extra_interfaces()` was modified to distinguish both cases (empty vector and absent value). This is needed in the snapshot code. I have chosen to do it this way because I didn't want two functions in `json_utils.cpp`.

georgeliao: In the case of snapshot, I think the ternary if can do the job based on the original `read_extra_interfaces` function, that makes `record.contains("extra_interfaces") ? MP_JSONUTILS.read_extra_interfaces(json) : desc.extra_interfaces`. However, there is an additional check of `record.contains("extra_interfaces") `, so in summary, maybe the `std::optional` approach is fine here.

ricab (approved):
Alright, I have finished testing this and it is working great on macOS :smiley: Thanks for your work here!

I will review the private side and then we can merge.

townsend2010:
This is going to fail the merge queue due to the private side.  Please merge manually when ready. :slightly_smiling_face:

ricab:
Right, done!

PR #3446, opened 2024-03-21
Make snapshot and restore track the instance id

georgeliao:
A few things are done in this PR. 
1. Removed `make_cloud_init_meta_config` and `make_cloud_init_network_config` in daemon.cpp which is the rebase fix forgotten to apply before. 
2. Made all the cloudinit free functions into singleton and established the unit test mocking framework for that. 
3. Toggle instance-id scheme was changed back to the simple appending "_e" since `snapshot_restore` will restore the instance-id so we will not have the long appending "_e" chain. 
4. Added `add_extra_interface_to_cloud_init` utility function into `CloudInitFileOps`, so it can be used by `base_virtual_machine::add_network_interface` and be mocked easily. 
5. `apply_extra_interfaces_to_cloud_init` function lxd part was removed for now because this utility function is only used in `restore_snapshot` and lxd does not support that yet. 
6. `apply_extra_interfaces_to_cloud_init` call from daemon is removed and each backend `add_network_interface` implementation is appended with a `add_extra_interface_to_instance_cloud_init` call. 
7. At last and the most important, the tracking instance id code in `basesnapshot` is added, including data member and `get_instance_id` function.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3591,7 +3591,7 @@ void mp::Daemon::add_bridged_interface(const std::string& instance_name)
     mpl::log(mpl::Level::trace, category, "Adding new interface to instance");
     try
     {
-        instance->add_network_interface(spec.extra_interfaces.size() - 1, new_if);
+        instance->add_network_interface(spec.extra_interfaces.size() - 1, spec.default_mac_address, new_if);
         mpl::log(mpl::Level::trace, category, "Done adding");
 
         // Configure the new interface via cloud-init (this won't throw).
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -45,7 +45,6 @@
 #include <multipass/ssh/ssh_session.h>
 #include <multipass/sshfs_mount/sshfs_mount_handler.h>
 #include <multipass/top_catch_all.h>
-#include <multipass/utils.h>
 #include <multipass/version.h>
 #include <multipass/virtual_machine.h>
 #include <multipass/virtual_machine_description.h>
@@ -53,6 +52,7 @@
 #include <multipass/vm_image.h>
 #include <multipass/vm_image_host.h>
 #include <multipass/vm_image_vault.h>
+#include <multipass/yaml_node_utils.h>
 
 #include <scope_guard.hpp>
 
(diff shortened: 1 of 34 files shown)


src/platform/backends/shared/base_snapshot.h
```diff
@@ -37,6 +37,7 @@ class BaseSnapshot : public Snapshot
 public:
     BaseSnapshot(const std::string& name,
                  const std::string& comment,
+                 const std::string& instance_id,
```
ricab, line 40: Can we call this `cloud_init_id` instead? `instance_id` is fine in the context of cloud init code, but here it is a little enigmatic. Readers would wonder "how does this differ from the name?" and they would have to read code to understand. Better make the name immediately obvious (here and in accessors).

ricab: Yep, sounds good to me.

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -472,22 +472,36 @@ void mp::LXDVirtualMachine::resize_disk(const MemorySize& new_size)
     lxd_request(manager, "PATCH", url(), patch_json);
 }
 
-void mp::LXDVirtualMachine::add_network_interface(int index, const mp::NetworkInterface& net)
+void mp::LXDVirtualMachine::add_network_interface(int index,
+                                                  const std::string& default_mac_addr,
+                                                  const mp::NetworkInterface& extra_interface)
```
georgeliao, line 477: Changed all variable name `net` to `extra_interface` to be consistent.

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -472,22 +472,36 @@ void mp::LXDVirtualMachine::resize_disk(const MemorySize& new_size)
...
 
     QJsonObject patch_json{{"devices", QJsonObject{{net_name, net_config}}}};
 
     lxd_request(manager, "PATCH", url(), patch_json);
+    add_extra_interface_to_instance_cloud_init(default_mac_addr, extra_interface);
 }
 
-void mp::LXDVirtualMachine::apply_extra_interfaces_to_cloud_init(const std::string& default_mac_addr,
-                                                                 const std::vector<NetworkInterface>& extra_interfaces)
+std::unique_ptr<mp::MountHandler> mp::LXDVirtualMachine::make_native_mount_handler(const std::string& target,
```
georgeliao, line 492: This function appears here because of being moved up, there is no change of it.

georgeliao:
@ricab @luis4a0 , the public side of the PR is ready for review, please read the intro comment to have an overview of what has been done.

src/daemon/daemon.cpp
```diff
@@ -3635,12 +3591,7 @@ void mp::Daemon::add_bridged_interface(const std::string& instance_name)
     mpl::log(mpl::Level::trace, category, "Adding new interface to instance");
     try
     {
-        instance->add_network_interface(spec.extra_interfaces.size() - 1, new_if);
-        mpl::log(mpl::Level::trace, category, "Done adding");
-
-        // Configure the new interface via cloud-init (this won't throw).
-        mpl::log(mpl::Level::trace, category, "Adding new interface to cloud-init");
-        instance->apply_extra_interfaces_to_cloud_init(spec.default_mac_address, spec.extra_interfaces);
+        instance->add_network_interface(spec.extra_interfaces.size() - 1, spec.default_mac_address, new_if);
```
ricab, line 3594: Ahh, thank you!

src/utils/yaml_node_utils.cpp
```diff
@@ -114,15 +128,54 @@ YAML::Node mp::utils::make_cloud_init_network_config(const std::string& default_
...
+        const std::string extra_interface_name = "extra0";
+        network_data["ethernets"][extra_interface_name] =
+            create_extra_interface_node(extra_interface_name, extra_interface.mac_address);
+
+        return network_data;
+    }
+
+    YAML::Node network_data = YAML::Load(network_config_file_content);
+
+    int i = 0;
```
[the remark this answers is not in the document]
georgeliao: That is what the code below does, `while (true)` loop keeps probing the next available number, if extra0 exists already then try extra1, until it does not exist. Then we create the extra interface item and add to the `YAML::Node`.

tests/test_base_virtual_machine.cpp
```diff
@@ -19,10 +19,12 @@
 #include "dummy_ssh_key_provider.h"
 #include "file_operations.h"
 #include "mock_cloud_init_file_ops.h"
+#include "mock_file_ops.h"
 #include "mock_logger.h"
 #include "mock_snapshot.h"
 #include "mock_ssh_test_fixture.h"
 #include "mock_utils.h"
+
```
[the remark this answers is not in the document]
georgeliao: Yep, this was an oversight, will fix it.

georgeliao:
@luis4a0 , I think all the comments are addressed. Maybe you can have a look of the [private side](https://github.com/canonical/multipass-private/pull/599).

luis4a0 (approved):
Hey @georgeliao! Thanks for this! The code looks good to me and this works as expected. So we are ready to merge!

PR #3448, opened 2024-03-22
Shutdown grpc server before quitting the Qt event loop

andrei-toterman:
If the Qt event loop exited before the gRPC server was done processing ongoing requests, the gRPC server became stuck. This PR makes sure that before we actually quit the Qt event loop, we tell the gRPC server to shut down and wait until it is done with what it was doing.

### DIFF
--- a/src/daemon/daemon_main.cpp
+++ b/src/daemon/daemon_main.cpp
@@ -84,6 +84,11 @@ int main_impl(int argc, char* argv[])
 
     mp::daemon::monitor_and_quit_on_settings_change(); // TODO replace with async restart in relevant settings handlers
     mp::Daemon daemon(std::move(config));
+    QObject::connect(&app,
+                     &QCoreApplication::aboutToQuit,
+                     &daemon,
+                     &mp::Daemon::shutdown_grpc_server,
+                     Qt::DirectConnection);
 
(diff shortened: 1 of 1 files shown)


src/daemon/daemon_main.cpp
```diff
@@ -84,6 +84,11 @@ int main_impl(int argc, char* argv[])
 
     mp::daemon::monitor_and_quit_on_settings_change(); // TODO replace with async restart in relevant settings handlers
     mp::Daemon daemon(std::move(config));
+    QObject::connect(&app,
+                     &QCoreApplication::aboutToQuit,
+                     &daemon,
+                     &mp::Daemon::shutdown_grpc_server,
```
ricab, line 90: For my understanding, why not connect to `daemon_rpc` directly?

andrei-toterman: Because we don't have access to it from here. It is a private member of `Daemon`. We could expose a getter to it, and then we could connect to it directly from here. Would that be preferable?

ricab: Oh duh, I read this as if it was in `daemon.cpp`. Nevermind

townsend2010 (approved):
Hey @andrei-toterman!

LGTM, thanks!!!

PR #3457, opened 2024-04-02
Thread-safe SSH session and process

ricab:


ricab:
OK I am declaring this ready for review, but I would like to hear opinions on something. 

As I have it, an `SSHSession` locks when creating an `SSHProcess` and passes the lock to the process. The process only releases the lock when someone asks for the `exit_code()` or when it is destroyed (or if a friend [`SftpServer`] asks to release the channel). 

If someone asks for the exit code again, and if we got it the first time, the known code is returned. However, if we timeout obtaining the exit code the first time, the lock is released and the session can be used by other threads, even though the exit code can be requested again. At that point, the session is no longer thread safe. 

I could instead hold the lock until the exit code is successfully obtained and provide a knob for client code to release the lock manually on timeout. But if the client failed to do this, the program would deadlock when trying to destroy the session. I could also *invalidate* the ssh process and refuse to look for the exit code again, but then we couldn't recover it. What do you think is the best approach for this?

townsend2010:
Hey @ricab!

Looks like you're going to need a private side PR: https://github.com/canonical/multipass-private/actions/runs/8543123264/job/23406233450#step:22:627

ricab:
Ah shoot... OK, I'll try to do that soon.

townsend2010:
Hey @ricab!

I've given your question some thought here and IMO, we shouldn't really allow calling `exit_status()` again after a timeout.  A timeout to me means that the `SSHProcess` is no longer valid and the whole execution should be retried in a new object.  And really, I think once `exit_status()` is called, that should be it regardless if it returned a value or timeout out.

I think if the `SSHProcess` object is still around after calling `exit_status()` and tries to call `exit_status()` again, `exit_status()` should check if it no longer owns the lock and if not, should throw an exception.

We can talk more about about this if this isn't really making sense or if you disagree :slightly_smiling_face:

ricab:
Hey @townsend2010, thanks for the feedback. OK, I think that equates to what I had in mind with my last option (_[...]_ invalidate _the ssh process and refuse to look for the exit code again [...]_). The principle makes sense to me, but I would still like to discuss the details and I need to see how to fold that with the `SFTPServer` (I think it asks for the `exit_code` twice today).

ricab:
Hey @townsend2010, I did something along the lines we discussed, although instead of "invalidating" subsequent `exit_code` calls I just return the result from the first call, be that an actual exit code or one of two new Exitless exceptions. This is more in line with the principle of not discarding results of computations. It avoids yet another exception while preventing any attempts to actually fetch the exit code again. The session is unlocked by then and SSH isn't used again. The new "peeking" method also saves the exit code if it gets one (but not exceptions).

There is one commit that I am quite unhappy about: https://github.com/canonical/multipass/pull/3457/commits/becb2c58a635dcc6ed95140497c6f1b0dea6db8c. Doing better would require rewriting many sftp/sshfs tests. That would deviate too much from what we're after here. Unless you find an easier way to adapt? I found that `ExitStateMock` does not have "enough resolution", and mocking `MP_UTILS.run_in_ssh_session` on top of the exinting premocks doesn't work without reworking what is there, let along mocking `SSHProcess`/`SSHSession` directly.

Let me know what you think.

include/multipass/ssh/ssh_process.h
```diff
@@ -45,13 +68,16 @@ class SSHProcess
...
-    ssh_channel release_channel();
+    ssh_channel release_channel(); // releases the lock on the session; callers are on their own to ensure thread safety
 
+    std::unique_lock<std::mutex> session_lock; // do not attempt to re-lock, as this is moved from
     ssh_session session;
-    const std::string cmd;
+    std::string cmd;
     ChannelUPtr channel;
-    std::optional<int> exit_status;
+    std::variant<std::monostate, int, std::exception_ptr> exit_result;
```
townsend2010, line 80: Nice!

src/sshfs_mount/sftp_server.cpp
```diff
@@ -194,15 +194,13 @@ auto validate_path(const std::string& source_path, const std::string& current_pa
 
 void check_sshfs_status(mp::SSHSession& session, mp::SSHProcess& sshfs_process)
 {
-    try
+    if (sshfs_process.exit_recognized(250ms))
     {
-        if (sshfs_process.exit_code(250ms) != 0)
+        // This `if` is artificial and should not really be here. However there is a complex arrangement of Sftp and
+        // SshfsMount tests depending on this.
+        if (sshfs_process.exit_code(250ms) != 0) // TODO remove
```
townsend2010, line 201: I'm fine with this and understand the conundrum.  We sometimes have to add little quirks in the main code for testing- this is just one of those cases.

townsend2010 (commented):
Hey @ricab!

Really nice work here!  The code looks good to me.  I've made a few inline comments, but nothing to act on.

I'd like to test this with the latest `main` and built as a snap, so could you please rebase this on `main` and then I will test?  Thanks!

ricab:
Thanks for the review @townsend2010! 

> I'd like to test this with the latest main and built as a snap, so could you please rebase this on main and then I will test? Thanks!

I am trying, but having trouble with Flutter at the moment (@andrei-toterman build instructions are sorely missed :wink: )

In the meantime, doesn't the snap build on a main merge?

townsend2010:
> In the meantime, doesn't the snap build on a main merge?

Yeah, it is, so it's probably ok.

ricab:
Alright, clean rebase @townsend2010 :point_up_2:

townsend2010:
Ok, cool, thanks @ricab!

townsend2010 (approved):
Hey @ricab!

Good work here! Testing went fine for me, 👍

townsend2010:
Right, need to merge manually...

PR #3463, opened 2024-04-05
Bridging on Linux QEMU

luis4a0:
This PR implements the bridging feature on the QEMU backend on Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,7 +219,9 @@ else()
       add_definitions(-DMULTIPASS_COMPILER_GCC)
     endif()
 
-    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds -Wnon-virtual-dtor)
+    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds)
+
+    string(APPEND CMAKE_CXX_FLAGS " -Wnon-virtual-dtor")
 
     if(NOT ${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "^arm")
       add_compile_options(-Wcast-align)
--- a/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
+++ b/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
@@ -238,8 +238,7 @@ auto mp::LXDVirtualMachineFactory::networks() const -> std::vector<NetworkInterf
                 ret.push_back(std::move(network));
 
         for (auto& net : ret)
-            if (net.needs_authorization && find_bridge_with(ret, net.id) != ret.cend())
-                net.needs_authorization = false;
+            net.needs_authorization &= (find_bridge_with(ret, net.id) == ret.cend());
     }
 
     return ret;
--- a/src/platform/backends/qemu/linux/bridge_helper.c
+++ b/src/platform/backends/qemu/linux/bridge_helper.c
@@ -1,0 +1,331 @@
+/*
+ * QEMU Bridge Helper
+ *
+ * Copyright IBM, Corp. 2011
+ * Copyright (C) Canonical, Ltd.
+ *
+ * Authors:
+ * Anthony Liguori   <aliguori@us.ibm.com>
+ * Richa Marwaha     <rmarwah@linux.vnet.ibm.com>
+ * Corey Bryant      <coreyb@linux.vnet.ibm.com>
+ * Luis Peñaranda    <luis.penaranda@canonical.com>
+ *
+ * This work is licensed under the terms of the GNU GPL, version 2.  See
+ * the COPYING file in the top-level directory.
+ */
+
+/* This version of the bridge helper was adapted for use with Multipass.
+   The changes are:
+   - the authorization via ACL was removed;
+   - dependencies on other QEMU functions were replaced by common includes;
+   - functionality was wrapped inside a function.
+ */
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include <sys/ioctl.h>
+#include <sys/prctl.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+
+#include <fcntl.h>
+#include <linux/if_tun.h>
+#include <net/if.h>
+
+#include <linux/sockios.h>
+
+#ifndef SIOCBRADDIF
+#include <linux/if_bridge.h>
+#endif
+
+#ifdef CONFIG_LIBCAP_NG
+#include <cap-ng.h>
+#endif
+
+static void usage(void)
+{
+    fprintf(stderr, "Usage: bridge_helper [--use-vnet] --br=bridge --fd=unixfd\n");
+}
+
+static bool has_vnet_hdr(int fd)
+{
+    unsigned int features = 0;
+
+    if (ioctl(fd, TUNGETFEATURES, &features) == -1)
+    {
+        return false;
+    }
+
+    if (!(features & IFF_VNET_HDR))
+    {
+        return false;
+    }
+
+    return true;
+}
+
+static void prep_ifreq(struct ifreq* ifr, const char* ifname)
+{
+    memset(ifr, 0, sizeof(*ifr));
+    snprintf(ifr->ifr_name, IFNAMSIZ, "%s", ifname);
+}
+
+static int send_fd(int c, int fd)
+{
+    char msgbuf[CMSG_SPACE(sizeof(fd))];
+    struct msghdr msg = {
+        .msg_control = msgbuf,
+        .msg_controllen = sizeof(msgbuf),
+    };
+    struct cmsghdr* cmsg;
+    struct iovec iov;
+    char req[1] = {0x00};
+
+    cmsg = CMSG_FIRSTHDR(&msg);
+    cmsg->cmsg_level = SOL_SOCKET;
+    cmsg->cmsg_type = SCM_RIGHTS;
+    cmsg->cmsg_len = CMSG_LEN(sizeof(fd));
+    msg.msg_controllen = cmsg->cmsg_len;
+
+    iov.iov_base = req;
+    iov.iov_len = sizeof(req);
+
+    msg.msg_iov = &iov;
+    msg.msg_iovlen = 1;
+    memcpy(CMSG_DATA(cmsg), &fd, sizeof(fd));
+
+    return sendmsg(c, &msg, 0);
+}
+
+#ifdef CONFIG_LIBCAP_NG
+static int drop_privileges(void)
+{
+    /* clear all capabilities */
+    capng_clear(CAPNG_SELECT_BOTH);
+
+    if (capng_update(CAPNG_ADD, CAPNG_EFFECTIVE | CAPNG_PERMITTED, CAP_NET_ADMIN) < 0)
+    {
+        return -1;
+    }
+
+    /* change to calling user's real uid and gid, retaining supplemental
+     * groups and CAP_NET_ADMIN */
+    if (capng_change_id(getuid(), getgid(), CAPNG_CLEAR_BOUNDING))
+    {
+        return -1;
+    }
+
+    return 0;
+}
+#endif
+
+int bridge_helper(const char* bridge, const int unixfd, const int use_vnet)
+{
+    struct ifreq ifr;
+#ifndef SIOCBRADDIF
+    unsigned long ifargs[4];
+#endif
+    int ifindex;
+    int fd = -1, ctlfd = -1;
+    int mtu;
+    char iface[IFNAMSIZ];
+    int ret = EXIT_SUCCESS;
+
+#ifdef CONFIG_LIBCAP_NG
+    /* if we're run from an suid binary, immediately drop privileges preserving
+     * cap_net_admin */
+    if (geteuid() == 0 && getuid() != geteuid())
+    {
+        if (drop_privileges() == -1)
+        {
+            fprintf(stderr, "failed to drop privileges\n");
+            return 1;
+        }
+    }
+#endif
+
+    if (bridge == NULL || unixfd == -1)
+    {
+        usage();
+        return EXIT_FAILURE;
+    }
+    if (strlen(bridge) >= IFNAMSIZ)
+    {
+        fprintf(stderr, "name `%s' too long: %zu\n", bridge, strlen(bridge));
+        return EXIT_FAILURE;
+    }
+
+    /* open a socket to use to control the network interfaces */
+    ctlfd = socket(AF_INET, SOCK_STREAM, 0);
+    if (ctlfd == -1)
+    {
+        fprintf(stderr, "failed to open control socket: %s\n", strerror(errno));
+        ret = EXIT_FAILURE;
+        goto cleanup;
+    }
+
+    /* open the tap device */
+    fd = open("/dev/net/tun", O_RDWR);
+    if (fd == -1)
+    {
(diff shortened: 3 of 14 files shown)


src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -247,7 +238,7 @@ auto mp::LXDVirtualMachineFactory::networks() const -> std::vector<NetworkInterf
                 ret.push_back(std::move(network));
 
         for (auto& net : ret)
-            if (net.needs_authorization && find_bridge_with(ret, net.id) != ret.cend())
+            if (net.needs_authorization && MP_BACKEND.find_bridge_with(ret, net.id) != ret.cend())
                 net.needs_authorization = false;
```
sharder996, line 242: Change this to match https://github.com/canonical/multipass/pull/3463/files#diff-77064bb2fbd2589b9a34679c3e6bde02dc7ad8b0b22931195d4d687e8aef2720R232? It makes it easier to see that the logic is the same in both places.

src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -266,6 +267,17 @@ std::string mp::Backend::get_subnet(const mp::Path& network_dir, const QString&
     return new_subnet;
 }
 
+std::vector<mp::NetworkInterfaceInfo>::const_iterator mp::Backend::find_bridge_with(
```
sharder996, line 270: There is some duplicate happening between here and [here](https://github.com/canonical/multipass/blob/71d73f38f0724c5899ad68b01919eae686c87ba2/src/platform/backends/shared/base_virtual_machine_factory.cpp#L35). I see the duplicate implementation being modified in #3513, but it would be nice to see these combined into a common implementation in the this utils file.

sharder996 (changes requested):

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,22 +194,59 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
 {
-    // TODO: Do not uncomment the following line when implementing bridging in Linux QEMU. Since implementing it would
-    // yield the same exact implementation than in macOS, we need to move the code out of here, and put it only once
-    // in src/platform/backends/qemu/qemu_virtual_machine.[h|cpp].
-    // desc.extra_interfaces.push_back(net);
+    auto platform_ifs_info = MP_PLATFORM.get_network_interfaces_info();
+
+    std::vector<NetworkInterfaceInfo> networks;
+
+    for (auto& ifs_info : platform_ifs_info)
```
sharder996, line 217: I think this can be declared `const`

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,22 +194,59 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
+    {
+        auto& info = ifs_info.second;
```
sharder996, line 219: idem

sharder996 (changes requested):

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,22 +194,59 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
 }
 
 mp::QemuPlatform::UPtr mp::QemuPlatformFactory::make_qemu_platform(const Path& data_dir) const
 {
     return std::make_unique<mp::QemuPlatformDetail>(data_dir);
 }
+
+std::string mp::QemuPlatformDetail::bridge_name_for(const std::string& iface_name) const
+{
+    return MP_BACKEND.bridge_name(iface_name);
```
ricab, line 246: I don't really understand what is going on here. This `bridge_name_for` thing has become pretty complex. This is what I see:

1. the daemon calls `VirtualMachineFactory::bridge_name_for`. This is implemented to:
    1. return an empty string by default. 
        1. this covers libvirt and VirtualBox, which have no specific implementation
    2. In LXD, it returns `mp::Backend::bridge_name()`. 
    3. In QEMU, it delegates on `QemuPlatform::bridge_name_for`, which:
        1. returns an empty string by default (covers macOS, as there is no specific implementation there)
        2. again calls `mp::Backend::bridge_name()` on Linux 
    4. On Hyper-V, it incorrectly returns a hardcoded "External Virtual Switch" string (bug we've discussed)

This complexity is complemented by tests and mocks. But the only functional bit here is `mp::Backend::bridge_name()`, on Linux, where we create bridges. However, and this is the crucial part, IIANM this function shouldn't be used at all in the first place! The daemon should not be looking for the bridge based on the name at all, but based on its members, as we've established when we discussed the bug in Hyper-V. 

Note that information on bridge members is readily available on our return from `VirtualMachineFactory::networks()`. We just need to run a find that is the same for all backends. And this find is implemented. It's the one you just moved [here](https://github.com/canonical/multipass/pull/3463/files#diff-d7e8137b54f1c26de443d2d3364309d322dea72c5ca4299068f7fb23504680b7R270). 

So you could just replace this call [https://github.com/canonical/multipass/blob/main/src/daemon/daemon.cpp#L3523] with a call to `find_bridge_with`. That would fix the behavior and get rid of all this code... Am I missing something??

ricab (changes requested):

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
luis4a0: Hey @ricab, thanks! Well, what you are proposing is exactly what I am implementing as the fix to the Hyper-V bug [here](https://github.com/canonical/multipass/pull/3513). What do you think in finishing that fix, merging and then rebasing this branch on top of that? That way, the history will reflect exactly what change does what and this PR will only add bridging to QEMU on Linux.

Otherwise, I can fix everything here. The resulting code will be the same but, you know, the change here can be seen as a drive-by change.

WDYT?

luis4a0: Or, I can leave that here, merge this and fix later on the other PR. In any case, both PR's will go to 1.14 IIRC.

CMakeLists.txt
```diff
@@ -219,7 +219,9 @@ else()
       add_definitions(-DMULTIPASS_COMPILER_GCC)
     endif()
 
-    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds -Wnon-virtual-dtor)
+    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds)
+
+    string(APPEND CMAKE_CXX_FLAGS " -Wnon-virtual-dtor")
```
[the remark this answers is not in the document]
luis4a0: Hey @sharder996, thanks for the review! The problem here is that the C compiler rejects the flags which are only meant for C++.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
ricab: Ah, I looked at that PR but did not see this. And since this one is adding to the `bridge_name` stuff... But OK, yes, what you propose is perfectly fine! As long as it does the right thing and removes the obsolete code in the end.

luis4a0: Of course, I'll let you know when I commit to the other PR. Thanks!

CMakeLists.txt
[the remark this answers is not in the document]
sharder996: OK, that was what I was suspecting!

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
luis4a0: Right, changed in the base branch.

src/platform/backends/shared/linux/backend_utils.cpp
luis4a0: Right but, in any case, this is outdated.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
luis4a0: Ok, this was all addressed in the other PR. I changed the base of this PR, so now it's on top of the `get` fix.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
luis4a0: Right, changed.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
luis4a0: Right, changed.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
sharder996: Missed this one

luis4a0: Ouch. Done (and force-pushed).

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
sharder996: This still isn't changed

src/platform/backends/shared/linux/backend_utils.cpp
sharder996: Ok, this is fine now, but...

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,29 +194,65 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
-    throw NotImplementedOnThisBackendException("add interfaces");
+    return networks;
 }
 
 std::vector<mp::NetworkInterfaceInfo>::const_iterator mp::QemuPlatformDetail::find_bridge_with(
     const std::vector<mp::NetworkInterfaceInfo>& networks,
     const std::string& member_network) const
 {
-    return networks.cend();
+    return std::find_if(std::cbegin(networks),
```
sharder996, line 243: There is duplication happening here now with the duplicate being found in `base_virtual_machine_factory.cpp`.

What if you pulled out 
```

    for (auto& net : networks)
    {
        if (net.type == "ethernet")
        {
            net.needs_authorization = (find_bridge_with(networks, net.id) == networks.cend());
        }
    }
```

from `networks()` above and moved it to `networks()` in `qemu_virtual_machine_factory`. That way you take advantage of the inherited `find_bridge_with()` method.

src/platform/backends/qemu/qemu_platform.h
```diff
@@ -54,11 +54,11 @@ class QemuPlatform : private DisabledCopyMove
     {
         return {};
     };
-    virtual std::vector<NetworkInterfaceInfo> networks() const
+    virtual std::vector<NetworkInterfaceInfo> networks() const = 0;
+    virtual std::string create_bridge_with(const std::string& interface) const
     {
-        throw NotImplementedOnThisBackendException("networks");
+        return interface;
```
sharder996, line 60: This just echoes the string because the bridge gets handled by the qemu bridge helper, right?

sharder996 (changes requested):
Ok, just a few more questions. Otherwise, the code looks good, but I am unable to functionally test since my laptop running Linux died and I haven't setup a replacement yet.

src/platform/backends/qemu/qemu_platform.h
luis4a0: No, in fact, the simple implementation is not used on Linux. I've done the function pure virtual.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
luis4a0: The problem with doing that is that we need to ask for authorization only on Linux. If we move it, then we'll be setting the need for authorization on macOS also. I thought a lot about this when implementing, and concluded that this is the easiest way of doing it. But please let me know if I am missing something.

luis4a0:
Hey @sharder996! I addressed your two points. this is ready for another round of review. Thanks!

townsend2010 (changes requested):
Hey @luis4a0,

Just a couple of copyright things inline below. :) Thanks!

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
luis4a0: Oh, sorry, I didn't get it. I changed it as you propose, but I used `&=` instead of `=` to maintain exactly the same logic (i.e., the same exact truth table).

src/platform/backends/qemu/linux/bridge_helper.c
```diff
@@ -0,0 +1,329 @@
+/*
+ * QEMU Bridge Helper
+ *
+ * Copyright IBM, Corp. 2011
+ *
```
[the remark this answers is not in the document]
luis4a0: Right, added.

src/platform/backends/qemu/linux/bridge_helper.c
```diff
@@ -0,0 +1,329 @@
...
+ * Authors:
+ * Anthony Liguori   <aliguori@us.ibm.com>
+ * Richa Marwaha     <rmarwah@linux.vnet.ibm.com>
+ * Corey Bryant      <coreyb@linux.vnet.ibm.com>
```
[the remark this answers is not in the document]
luis4a0: Well, I didn't make functional changes. But I added it in order to blame someone if something goes wrong ;)

luis4a0:
@sharder996 I addressed the point I didn't address in the last round, so this is ready again for review. Thanks!

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
sharder996: Ah, ok. I forgot about the private side. Yes, then this makes sense. At least there isn't much for duplicated logic, just a loop through a list.

sharder996:
@luis4a0 Got my PC up and dual-booting with Ubuntu so I did some more functional testing. There are a couple of things that I noticed that seem a bit odd:

1. If I take a snapshot of an instance before and after setting `local.foo.bridged=true`, then restore the first snapshot (before adding the bridge) and then restore the second snapshot (after adding the bridge), when I start the instance the bridge does not get initialized. Only after restarting the instance does the bridge come up.

2. Also, since I recently moved I am on a different network and had to create a new bridge for that network. As such, I have an extra bridge in `multipass networks` which was for the old network and does nothing. Deleting the bridge with `ip link` will make it go away, but when I restart the Multipass daemon it comes back.

luis4a0:
Hey @sharder996, good to hear you have a working Linux again! Thanks for the new review, I'll answer your two points:
1. The problem here comes from cloud-init configuration. The configuration at Multipass level worked and I don't think we can do better (maybe increasing the timeout is one option).
2. This failure is not specific to QEMU, it's due to the way bridging on Linux works in Multipass. To avoid that child bridge being created, we need to remove from `extra_interfaces` the bridged interface entry. This will be trivial to do when we implement removing bridges (in the future, not on this PR).
Thanks!

PR #3471, opened 2024-04-07
Added zip as a dependency to debian/control

nielsreijers:
I tried building multipass in a clean 22.04 vm (started using multipass) following the instructions in the README.md.

This failed on the `cmake ../` step:

```
ubuntu@mptest:~/multipass/build$ cmake ../
-- Bootstrapping vcpkg before install
-- Bootstrapping vcpkg before install - failed
CMake Error at 3rd-party/vcpkg/scripts/buildsystems/vcpkg.cmake:895 (message):
  vcpkg install failed.  See logs for more information:
  /home/ubuntu/multipass/build/vcpkg-bootstrap.log
```

The mentioned log file was pretty clear about what was missing:

```
Could not find zip. Please install it (and other dependencies) with:
On Debian and Ubuntu derivatives:
  sudo apt-get install curl zip unzip tar
```

I figured adding it to debian/control so `mk-build-deps` will install it if it's missing would be the most user friendly way to fix it.
But if there's a good reason not to want it in there, I guess the README.md should be updated to add `zip` to the `apt install devscripts equivs` line.


ricab (approved):
Makes sense, thanks!

PR #3473, opened 2024-04-08
[actions] Continue on Coverage build errors

townsend2010:
This will help allow builds to continue while the issue of tokenless uploads to Codecov is worked out.

See https://github.com/codecov/codecov-action/issues/1359 for more details.


ricab (approved):
As clarified elsewhere, the fact that the run shows green with `continue-on-error` is a GH issue, and this is only a temporary workaround to unblock us until Codecov fixes things, so +1.

PR #3475, opened 2024-04-09
Added libgl1-mesa-dev as a dependency to debian/control

nielsreijers:
Similar to PR [3471](https://github.com/canonical/multipass/pull/3471)

I tried building multipass in a clean 22.04 vm (started using multipass) following the instructions in the README.md.

This failed on the cmake ../ step:

```
-- Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR) 
-- Could NOT find WrapOpenGL (missing: WrapOpenGL_FOUND) 
CMake Warning at 3rd-party/vcpkg/scripts/buildsystems/vcpkg.cmake:855 (_find_package):
  Found package configuration file:

    /usr/lib/x86_64-linux-gnu/cmake/Qt6Gui/Qt6GuiConfig.cmake

  but it set Qt6Gui_FOUND to FALSE so package "Qt6Gui" is considered to be
  NOT FOUND.  Reason given by package:

  Qt6Gui could not be found because dependency WrapOpenGL could not be found.
```

This in turns causes `Qt6Widgets` and `Qt6` to be NOT FOUND and the script to fail.

Adding libgl1-mesa-dev fixes this.


townsend2010:
Hey @nielsreijers!

Thanks for this contribution!  Just as you add this, we will be removing the dependency on `Qt6Widgets`/`Qt6Gui` very soon in https://github.com/canonical/multipass/pull/3282.  That said, I don't think this will be relevant anymore once that is merged.  I'll leave this open for now just in case this is actually still needed once we merge that other PR.

andrei-toterman:
This was resolved in #3282 as part of moving to the new Flutter GUI. We don't need Qt Widgets anymore.

PR #3482, opened 2024-04-12
Fix assert condition in QEMU VMs

ricab:
Fix an assert condition when making a new snapshot of a QEMU VM.


townsend2010 (approved):
Oops!  LGTM, thanks!

ricab:
Yeah, oops indeed... good thing it doesn't go in releases :disappointed_relieved: 

Thanks for the review! :)

PR #3483, opened 2024-04-12
Introduce a Stopping state

ricab:


ricab:
Obsolete now. May be used as a reference if we ever want to do this again.

PR #3487, opened 2024-04-16
Retry on download failure

### LINKED ISSUE #3464: launch failed: Remote "" is unknown or unreachable — then `multipass launch` magically starts working 15min after boot!
**Describe the bug**

~Since December/January, several people have confirmed that the Edge channel version of multipass can no longer run `multipass launch` after every boot.~  `multipass launch` fails when run within 15min of boot.  Several have confirmed this, and we each assumed the bug would go away.  But 4+ months later it has not (hence this bug report!)

The error message is: `launch failed: Remote "" is unknown or unreachable.`

One workaround (annoying, but it works) is to run: `snap restart multipass`

**Additional info**

Host PC details:

```
# cat /etc/issue
Ubuntu 22.04.4 LTS \n \l

# multipass version
multipass   1.14.0-dev.1600+g2701b66b
multipassd  1.14.0-dev.1600+g2701b66b

# multipass get local.driver
lxd

# multipass get local.bridged-network
enp1s0

# lxd --version
5.20

# snap info lxd
...
installed:          5.20-f3dd836              (27049) 155MB -

# uname -a
Linux 126-u2204-desk 6.5.0-26-generic #26~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Mar 12 10:22:43 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

**Additional context**

ASIDE: Will we be able test this on pre-release builds of Multipass on Windows in future, as hinted at... ?

- https://github.com/canonical/multipass/issues/3346#issuecomment-1879094428

Likely Related:

- #1853
- #3074

georgeliao:
fix #3464 

This PR adds [binary exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) based retry mechanism to async periodic download tasks, the checking exception (download failure an exception) in the main thread is done in a non-blocking fashion via `QFutureWatcher`. 

1. Changed `std::async` + `std::future` pair to `QtConcurrent::run` + `QFuture` in order to use some QFuture specific features.
2. Added  `QFutureWatcher` to watch `QFuture` and react correspondingly in the successful and failed cases when QFuture is finished. 
3. Changed `DownloadException` to a QException-based class so it can cross the qt thread boundary. 
4. Made the CRTP `BaseQException` class to reuse the boilerplate code for all the QException derived classes. 

To test the behavior, you can modify the default delay time in `daemon.h` line 222 to a small number, so the download will run more frequently. After that, the wifi network can be manually turned off in ubuntu settings, which in a way causes download failure.

### DIFF
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -68,27 +68,28 @@ public:
         mpl::log(mpl::Level::debug, "async task", std::string(launch_msg));
         future = QtConcurrent::run(std::forward<Callable>(func), std::forward<Args>(args)...);
 
-        QObject::connect(&future_watcher, &QFutureWatcher<ReturnType>::finished, [retry_start_delay_time, this] {
+        auto event_handler_on_sucess_and_failure = [retry_start_delay_time, this]() -> void {
             try
             {
                 // rethrow exception
                 future.waitForFinished();
-                mpl::log(mpl::Level::debug, "async task", "no Exception caught. ");
-                // success case, we reset the time interval for timer and the retry dayly time value
+                mpl::log(mpl::Level::debug, "async task", "The download succeeded.");
+                // success case, we reset the time interval for timer and the reset the retry delay time value
                 timer.start(default_delay_time);
                 retry_current_delay_time = retry_start_delay_time;
             }
             catch (const multipass::DownloadException& e)
             {
-                mpl::log(mpl::Level::info,
+                mpl::log(mpl::Level::debug,
                          "async task",
                          fmt::format("QFutureWatcher caught DownloadException {}", e.what()));
                 // failure case, trigger or continue the retry mechanism
                 timer.start(retry_current_delay_time);
                 retry_current_delay_time = std::min(2 * retry_current_delay_time, default_delay_time);
             }
-            mpl::log(mpl::Level::debug, "async task", "QFutureWatcher finished. ");
-        });
+        };
+
+        QObject::connect(&future_watcher, &QFutureWatcher<ReturnType>::finished, event_handler_on_sucess_and_failure);
         future_watcher.setFuture(future);
 
         QObject::connect(&timer, &QTimer::timeout, [launch_msg, this, func, args...]() -> void {
--- a/include/multipass/async_periodic_task.h
+++ b/include/multipass/async_periodic_task.h
@@ -31,31 +31,31 @@ namespace mpl = multipass::logging;
 
 namespace multipass::utils
 {
-// Because AsyncPeriodicTask needs to be instantiated as data member of a class to function properly, and Class template
-// argument deduction (CTAD) does not work for data member, so that means we can not deduced type data member from the
-// below code snippet.
+// Because AsyncPeriodicDownloadTask needs to be instantiated as data member of a class to function properly, and Class
+// template argument deduction (CTAD) does not work for data member, so that means we can not deduced type data member
+// from the below code snippet.
 
 // template <typename Callable, typename...Args>
-// class AsyncPeriodicTask
+// class AsyncPeriodicDownloadTask
 // {
 // public:
 //     using ReturnType = std::invoke_result_t<std::decay_t<Callable>, std::decay_t<Args>...>;
-//     AsyncPeriodicTask(Callable&& func, Args&&... args);
+//     AsyncPeriodicDownloadTask(Callable&& func, Args&&... args);
 //     ...
 // };
 
 // Therefore, the user has to specify the ReturnType as template argument while constructing it.
 
 template <typename ReturnType = void>
-class AsyncPeriodicTask
+class AsyncPeriodicDownloadTask
 {
 public:
     template <typename Callable, typename... Args>
-    AsyncPeriodicTask(std::string_view launch_msg,
-                      std::chrono::milliseconds normal_delay_time,
-                      std::chrono::milliseconds retry_start_delay_time,
-                      Callable&& func,
-                      Args&&... args)
+    AsyncPeriodicDownloadTask(std::string_view launch_msg,
+                              std::chrono::milliseconds normal_delay_time,
+                              std::chrono::milliseconds retry_start_delay_time,
+                              Callable&& func,
+                              Args&&... args)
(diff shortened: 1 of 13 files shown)


sharder996:
Hey @georgeliao, thanks for the work on this!

Would you be able to edit the PR description to link the github issues that this fixes in the format 'Fixes #xxxx'? That way I can get some more context on what behavior this is fixing.

Thanks!

georgeliao:
@sharder996 , sorry that I forgot it, the link is added now.

include/multipass/exceptions/base_qexception.h
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    void raise() const override
+    {
+        throw static_cast<const DerivedException&>(*this);
+    }
+
+    BaseQException* clone() const override
+    {
+        return new DerivedException(static_cast<const DerivedException&>(*this));
```
ricab, line 46: Hey @georgeliao, while this is a very clever idea, it is a little dangerous. It works well for the first derived class, but it will [break on the second](https://godbolt.org/z/b9E5PfvjE).

That said, I am fine with keeping it as long as:
- we document that this template should be instantiated with the ultimate concrete type for `DerivedException` (not an intermediate type)
- explain that means that any derived class should either be final, be itself templated, or override `raise` and `clone` (which defeats the purpose)
- follow that recipe in any derived exception today

georgeliao: That is nice the observation of CRTP, I think generally CRTP is used in one level of inheritance. So labeling the derived classes to final and document it should be more appropriate.

include/multipass/exceptions/base_qexception.h
```diff
@@ -25,13 +25,14 @@
 namespace multipass
 {
 // CRTP template base class that is the boilerplate code of QException derived classes
+// The usage of it should be CRTP derive class, that is inheriting the base calass and label itself final.
```
ricab, line 28:
```suggestion
// Derived classes should either be final or be themselves templated with the ultimate leaf type.
```

georgeliao: @ricab , indeed, we discussed two ways of using this base class. Later I thought about it again, since we are suggesting users about how to use CRTP base class here, the templated derived class is no longer a CRTP and it does not make sense here. That is why I ditched that (only keep the final derived class possibility). Tell me what you think about it.

ricab: IMO we shouldn't opt out of something just because it goes a little beyond the strict textbook concept that the term CRTP describes. Patterns often need to be adapted when employed. You could call it "generalized CRTP" if you want. 

Exceptions very often require multiple levels of inheritance. This comment precludes that, even though it would work just fine. That may cause unnecessary hurdles to future devs, instead of giving them the solution right away.

That said, this is a small matter and I am not going to block on it (but I would probably just do it myself in a PR if I came across it :wink: ). I will request that you at least fix the typos in the message, please.

include/multipass/async_periodic_task.h
```diff
@@ -60,16 +66,41 @@ class AsyncPeriodicTask
...
+        auto event_handler_on_sucess_and_failure = [retry_start_delay_time, this]() -> void {
+            try
+            {
+                // rethrow exception
+                future.waitForFinished();
+                // success case, we reset the time interval for timer and the reset the retry delay time value
+                timer.start(default_delay_time);
+                retry_current_delay_time = retry_start_delay_time;
+            }
+            catch (const multipass::DownloadException& e)
```
sharder996, line 80: If `async_periodic_task` is meant to be generic, shouldn't we instead catch something a bit more generic so that this isn't tied to only doing download tasks?

sharder996 (changes requested):
Looks good @georgeliao!

Just a few nitpicks and questions.

include/multipass/exceptions/base_qexception.h
```diff
@@ -0,0 +1,61 @@
...
+{
+// CRTP template base class that is the boilerplate code of QException derived classes
+// Derived classes should either be final or be themselves templated with the ultimate leaf type.
+template <typename DerivedException>
+class BaseQException : public QException
+{
+public:
+    BaseQException(const std::string& err) : error_string{err}
+    {
+        // use concepts instead of static_assert to apply the contraint once C++ 20 arrives.
```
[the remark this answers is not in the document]
georgeliao: Good idea, I will use the format `TODO@C++20/23 ` for this, it can distinguish itself from the major TODO tasks.

include/multipass/async_periodic_task.h
georgeliao: Excellent question and a nice observation, but I think retry should only act on the DownloadException and not act on other types of exceptions. 

On the other hand, `async_periodic_task` is a generic utility class that handles both download and non-download types of tasks; in the case of non-download type classes, we may not need this retry mechanism. Then maybe it is better to decouple the retry part code from the class. That will be some work. 

For now, there are only async periodic download tasks that occur, so changing this class name to `AsyncPeriodicDownloadTask` and making it only launches download tasks is also a good solution.

ricab: Just to add my two cents, you both have good points and I think renaming is a good compromise for the time being.

sharder996: I'm fine with the renaming as well.

I think in the future if we needed to create an `AsyncPeriodicTask` without a retry timer we would add another constructor with `retry_start_delay_time` omitted that simply uses `INT_MAX`, or something of the like, as a default delay.

sharder996: @georgeliao The filename is still unchanged

georgeliao: Nice catch, overlooked that,

sharder996: Ok, I think this is good for the time being, however I would like to see it become more generic in the future. @ricab, if you are fine with this, then I think it can go in.

ricab: Thanks @sharder996, I did my review already so feel free to merge when you approve.

sharder996 (approved):

PR #3493, opened 2024-04-17
Permit only one bridge

luis4a0:
If `local.instance.bridged` is `true`, no more interfaces bridging the current network will be allowed.

### DIFF
--- a/src/daemon/instance_settings_handler.cpp
+++ b/src/daemon/instance_settings_handler.cpp
@@ -21,15 +21,19 @@
 #include <multipass/constants.h>
 #include <multipass/exceptions/invalid_memory_size_exception.h>
 #include <multipass/exceptions/not_implemented_on_this_backend_exception.h>
+#include <multipass/logging/log.h>
 #include <multipass/settings/bool_setting_spec.h>
 
 #include <QRegularExpression>
 #include <QStringList>
 
 namespace mp = multipass;
+namespace mpl = multipass::logging;
 
 namespace
 {
+constexpr auto category = "instance settings";
+
 constexpr auto cpus_suffix = "cpus";
 constexpr auto mem_suffix = "memory";
 constexpr auto disk_suffix = "disk";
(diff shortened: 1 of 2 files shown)


src/daemon/instance_settings_handler.cpp
```diff
@@ -166,14 +166,19 @@ void update_bridged(const QString& key,
...
+        {
+            throw mp::InvalidSettingException{key, val, "Bridged interface cannot be removed"};
+        }
     }
-
-    if (bridged)
+    else
     {
-        add_interface(instance_name);
+        if (bridged)
```
ricab, line 178: How about `else if`?

src/daemon/instance_settings_handler.cpp
```diff
@@ -166,14 +166,19 @@ void update_bridged(const QString& key,
...
     auto bridged = mp::BoolSettingSpec{key, "false"}.interpret(val) == "true";
 
-    if (!bridged && is_bridged(instance_name))
+    if (is_bridged(instance_name))
     {
-        throw mp::InvalidSettingException{key, val, "Bridged interface cannot be removed"};
+        if (!bridged)
+        {
+            throw mp::InvalidSettingException{key, val, "Bridged interface cannot be removed"};
+        }
```
ricab, line 174: I think it deserves an `info` log saying that the instance is already bridged.

ricab (commented):
Tested and it is almost there, I have only a couple of very small requests.

src/daemon/instance_settings_handler.cpp
luis4a0: Right, done.

src/daemon/instance_settings_handler.cpp
luis4a0: Good, done.

src/daemon/instance_settings_handler.cpp
```diff
@@ -166,12 +170,18 @@ void update_bridged(const QString& key,
     // This is the user parameter, true or false.
     auto bridged = mp::BoolSettingSpec{key, "false"}.interpret(val) == "true";
 
-    if (!bridged && is_bridged(instance_name))
+    if (is_bridged(instance_name))
     {
-        throw mp::InvalidSettingException{key, val, "Bridged interface cannot be removed"};
+        if (bridged)
+        {
+            mpl::log(mpl::Level::warning, category, fmt::format("{} is already bridged", instance_name));
```
ricab, line 177: I would have thought `info` more appropriate here, but I won't block on that.

ricab (approved):
Alright, I am sending this in! Thanks Luis.

PR #3517, opened 2024-04-30
[unit test] use nicemock to remove the excessive gmock warnings. 

georgeliao:
Remove the excessive gmock warnings from unit tests.


sharder996 (approved):
Since the `MockCertProvider` is a dependency used in each of the test classes and we aren't testing the CertProvider specifically here, I think this generally is an okay situation to ignore uninteresting calls.

PR #3523, opened 2024-05-08
Fix issues with runtime info

ricab:
- Add back runtime info items that got lost in a rebase.
- Fix logic to decide whether to execute cmds in parallel
- Fix command quoting to avoid collapsing a double space on cpu_times


townsend2010 (changes requested):
Hey @ricab,

Although this looks correct, it is not working as it should.  It shows a constant 100% CPU utilization.  It's not apparent to me why though at this time.

townsend2010 (approved):
Hey @ricab!

LGTM :+1:  Thanks!

PR #3526, opened 2024-05-13
[actions] Enable Linux integration tests when using merge_queue

townsend2010:


ricab:
I don't think this worked, so closing, but let's come back to this as a reference.

PR #3529, opened 2024-05-16
[cli] Added -a flag to the stop command

rogkne:
Adds the _-a_ flag in addition to the already existing _--all_ flag to the sub command _stop_.

This provides a more consistent help output and it is handy for regular use:
```
Options:
  -h, --help         Displays help on commandline options
  -v, --verbose      Increase logging verbosity. Repeat the 'v' in the short
                     option for more detail. Maximum verbosity is obtained with
                     4 (or more) v's, i.e. -vvvv.
  -a, --all          Stop all instances
  -t, --time <time>  Time from now, in minutes, to delay shutdown of the
                     instance
  -c, --cancel       Cancel a pending delayed shutdown
```

### DIFF
--- a/src/client/cli/cmd/stop.cpp
+++ b/src/client/cli/cmd/stop.cpp
@@ -80,7 +80,7 @@ mp::ParseCode cmd::Stop::parse_args(mp::ArgParser* parser)
 
     parser->addPositionalArgument("name", description, syntax);
 
-    QCommandLineOption all_option(all_option_name, "Stop all instances");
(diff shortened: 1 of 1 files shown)


ricab (commented):
Hi @rogkne, thank you for the PR and your interest in Multipass.

The absence of `-a` is intentional here. I know it is a little bit more effort, but the goal is to require a more conscious decision on the part of the user to affect all instances at once. And we want the flag to be consistent across commands (e.g. start, delete)

rogkne:
Thanks @ricab for the clarification. I close the PR.

PR #3538, opened 2024-05-30
Flutter submodule

andrei-toterman:
This PR adds git submodules for Flutter and the Dart protoc plugin.
Before this, we relied on Flutter and Dart to be manually installed and available in the `PATH` in order to build the GUI. We also required a specific version of Flutter, which might've created a conflict if another version of Flutter was already installed on the system.
Now we can handle all of that inside our project.
Of course, this PR also updates the build scripts in order to use those submodules.


snap/snapcraft.yaml
```diff
@@ -212,12 +232,12 @@ parts:
       cd ${CRAFT_PART_SRC}/3rd-party
       find . -name 'LICENSE*' -print0 | xargs -0 -I FILE install --mode=644 -D --no-target-directory FILE ${CRAFT_PART_INSTALL}/licenses/FILE
 
-      # Stage protoc and protobuf files for use in the gui part
-      PROTOBUF_INSTALL_DIR=${CRAFT_PART_INSTALL}/protobuf
-      mkdir -p ${PROTOBUF_INSTALL_DIR}/google/protobuf
-      cp -a ${CRAFT_PART_BUILD}/vcpkg_installed/*/tools/protobuf/protoc* ${PROTOBUF_INSTALL_DIR}
-      cp -a ${CRAFT_PART_BUILD}/vcpkg_installed/*/include/google ${PROTOBUF_INSTALL_DIR}
-      cp -a ${CRAFT_PART_SRC}/src/rpc/multipass.proto ${PROTOBUF_INSTALL_DIR}
+      # # Stage protoc and protobuf files for use in the gui part
```
georgeliao, line 235: Why does this part remain commented? Should we just remove them?

andrei-toterman: True. I commented them to test something but then I forgot to remove them. My bad.

georgeliao:
Hi @andrei-toterman 
Excellent work, it looks good to me. I just have some minor comments. 

Besides this, I have tried to build `multipass.gui` target on ubuntu 24.04 and I have found myself needing to install `libkeybinder-3.0-dev`, `libgtkmm-3.0-dev`, and `libayatana-appindicator3-dev`. My question here is that do we have a complete list of apt packages multipass.gui build (the flutter and dart tool chain) depends on? Later on, maybe we can have a section in multipass readme for building multipass.gui so we can added the `apt install <fluter dependency list>` there to clarify the building steps.

andrei-toterman:
Hey, @georgeliao! Indeed, there are some dependencies needed for our GUI. But they were added to the `debian/control` file in an earlier PR, so the existing build instruction of running `mk-build-deps -s sudo -i` would take care of them as well.

georgeliao:
@andrei-toterman 
I see, then that is even better. Thanks for clarifying.

georgeliao (approved):

PR #3544, opened 2024-06-06
[utils] Use a consistent namespace alias for utils

ricab:
Normalize uses of `mu` to `mpu`, which was already in use and is more in line with the three-letter norm for multipass namespace aliases.


sharder996 (approved):
This looks like a sane change to me

sharder996:
Oh, yes. I thought this had been merged. I will add it to the queue.

Issue #3489, opened 2024-04-16
Multipass does not recognize Hyper-V instances as bridged


ricab:
**Describe the bug**
The Multipass daemon keeps thinking that bridged instances aren't bridged, when the `bridged-network` is set to an non-bridge interface, in Hyper-V. If asked to set `bridge=true` again, it will then duplicate the bridged interface.

The problem is that the bridge is not associated with the interface [here](https://github.com/canonical/multipass-private/blob/275fbe13789163966c1613e6d99ecdfcf4b85d6f/src/platform/backends/hyperv/hyperv_virtual_machine_factory.h#L48-L51). This appears to be a forgotten placeholder and shouldn't be hard to fix.

VirtualBox doesn't have a dedicated implementation, but I think we don't need bridges there and just pass the end interface, so this shouldn't apply there.

**To Reproduce**

```
C:\> multipass get local.bridged-network
Ethernet Instance 0
C:\> multipass get local.glad-menhaden.bridged
false
C:\> multipass set local.glad-menhaden.bridged=true
C:\> multipass get local.glad-menhaden.bridged
false
C:\> multipass set -vvvv local.glad-menhaden.bridged=true
C:\> multipass set -vvvv local.glad-menhaden.bridged=true
```

The instance now has 3 new bridged networks.

**Expected behavior**
The bridge would be recognized and a second attempt to make the setting true would be a no op.

**Additional info**
 - OS: Windows
- `multipass version`
- `multipass get local.driver=hyper-v`

PR #3545, opened 2024-06-06
Fix detection of whether an instance is bridged, take2

ricab:
An instance is considered bridged, at any point in time, *iff* the configured preferred network at that time is:
  - part of the corresponding `VMSpecs::extra_networks`, or
  - linked to a bridge/switch (depending on the platform) which is part of the instance's `extra_networks`.

This PR implements that predicate in `mp::Daemon::is_bridged` (and `is_bridged_impl`), removing any previous logic relying on bridge names to relate bridges with other interfaces, and instead using readily available `NetworkInterfaceInfo::links`. To achieve this, the predicate calls `find_bridge_with`, which was refactored into a backend-agnostic utility.

Because Windows uses "switches" instead of "bridges", a platform function is added to obtain the correct nomenclature, replacing hard-coded uses in backend factories. Since the factories no longer need to call a base `prepare_networking_guts` with the correct nomenclature, that method is removed and its implementation moved to `BaseVirtualMachine::prepare_networking`, which is overridden to do nothing in VirtualBox.

The separation between `mp::Daemon::is_bridged` (and `is_bridged_impl`) is meant to avoid inspecting host networks or the preferred bridge setting twice when adding a bridge. To that end, the responsibility of checking whether and instance is already bridged, when attempting to bridge it, is split from the `InstanceSettingsHandler` to the `Daemon`, depending on the use case. 

This scheme of having different entities checking if the instance is bridged is a compromise that avoids passing yet more lambdas to the settings handler, which would otherwise be required to obtain networks or settings only once from that level. A better implementation would instead extract all required functionality (all "lambdas") into a separate class that could be passed to the settings handler instead of disconnected callables. But that would require more time.

Finally, existing tests were adapted, removed where obsolete, and a couple of tests were added to cover everything. Codecov seems to be unhappy that I've removed previously covered, but wrong/unnecessary, code.

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -99,15 +99,16 @@ bool valid_hostname(const std::string& name_string);
 std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
 template <typename NetworkContainer>
-auto find_bridge_with(const NetworkContainer& networks,
-                      const std::string& member_network,
-                      const std::string& bridge_type)
+std::optional<NetworkInterfaceInfo> find_bridge_with(const NetworkContainer& networks,
+                                                     const std::string& member_network,
+                                                     const std::string& bridge_type)
 {
-    return std::find_if(std::cbegin(networks),
-                        std::cend(networks),
-                        [&member_network, &bridge_type](const NetworkInterfaceInfo& info) {
-                            return info.type == bridge_type && info.has_link(member_network);
-                        });
+    const auto it = std::find_if(std::cbegin(networks),
+                                 std::cend(networks),
+                                 [&member_network, &bridge_type](const NetworkInterfaceInfo& info) {
+                                     return info.type == bridge_type && info.has_link(member_network);
+                                 });
+    return it == std::cend(networks) ? std::nullopt : std::make_optional(*it);
 }
 
 // string helpers
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -100,13 +100,13 @@ std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
 template <typename NetworkContainer>
 std::optional<NetworkInterfaceInfo> find_bridge_with(const NetworkContainer& networks,
-                                                     const std::string& member_network,
+                                                     const std::string& target_network,
                                                      const std::string& bridge_type)
 {
     const auto it = std::find_if(std::cbegin(networks),
                                  std::cend(networks),
-                                 [&member_network, &bridge_type](const NetworkInterfaceInfo& info) {
-                                     return info.type == bridge_type && info.has_link(member_network);
+                                 [&target_network, &bridge_type](const NetworkInterfaceInfo& info) {
+                                     return info.type == bridge_type && info.has_link(target_network);
                                  });
     return it == std::cend(networks) ? std::nullopt : std::make_optional(*it);
 }
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -19,6 +19,7 @@
 #define MULTIPASS_UTILS_H
 
 #include <multipass/logging/level.h>
+#include <multipass/network_interface_info.h>
 #include <multipass/path.h>
 #include <multipass/singleton.h>
 #include <multipass/ssh/ssh_session.h>
@@ -97,6 +98,17 @@ void validate_server_address(const std::string& value);
 bool valid_hostname(const std::string& name_string);
 std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
+template <typename NetworkContainer>
+auto find_bridge_with(const NetworkContainer& networks,
+                      const std::string& member_network,
+                      const std::string& bridge_type)
+{
+    return std::find_if(std::cbegin(networks),
+                        std::cend(networks),
+                        [&member_network, &bridge_type](const NetworkInterfaceInfo& info) {
+                            return info.type == bridge_type && info.has_link(member_network);
+                        });
+}
 
 // string helpers
 bool has_only_digits(const std::string& value);
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -98,18 +98,10 @@ void validate_server_address(const std::string& value);
 bool valid_hostname(const std::string& name_string);
 std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
-template <typename NetworkContainer>
-std::optional<NetworkInterfaceInfo> find_bridge_with(const NetworkContainer& networks,
+
+std::optional<NetworkInterfaceInfo> find_bridge_with(const std::vector<NetworkInterfaceInfo>& networks,
                                                      const std::string& target_network,
-                                                     const std::string& bridge_type)
-{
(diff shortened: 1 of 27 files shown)


ricab:
Bahh freakin CI...

include/multipass/utils.h
```diff
@@ -97,6 +98,18 @@ void validate_server_address(const std::string& value);
 bool valid_hostname(const std::string& name_string);
 std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
+template <typename NetworkContainer>
```
georgeliao, line 101: What is the point of making this template parameter here where all the callers pass `std::vector<NetworkInterfaceInfo>` type? Is there an envision for `std::list` or `std::deque`?

include/multipass/utils.h
```diff
@@ -97,6 +98,18 @@ void validate_server_address(const std::string& value);
...
+std::optional<NetworkInterfaceInfo> find_bridge_with(const NetworkContainer& networks,
+                                                     const std::string& member_network,
+                                                     const std::string& bridge_type)
+{
```
georgeliao, line 105: About the 3rd parameter `bridge_type`, since the `MP_PLATFORM.bridge_nomenclature()` is a singleton class member function. Can we remove this parameter and retrieve the bridge_type inside the function?

include/multipass/utils.h
```diff
@@ -97,6 +98,18 @@ void validate_server_address(const std::string& value);
 bool valid_hostname(const std::string& name_string);
 std::string generate_mac_address();
 bool valid_mac_address(const std::string& mac);
+template <typename NetworkContainer>
+std::optional<NetworkInterfaceInfo> find_bridge_with(const NetworkContainer& networks,
+                                                     const std::string& member_network,
```
georgeliao, line 103: The `member_network` seems a little weird. What this function does is to find any networks who has a link to the target network, so maybe `target_network`. Not sure whether it is a more expressive name. Besides this, since it is a `std::string`, so maybe it can be sufixed with `_id` or `_name` to emphasize the difference to `networks` name where network means a `NetworkInterfaceInfo` object.

src/daemon/daemon.cpp
```diff
@@ -362,6 +362,18 @@ std::string get_bridged_interface_name()
     return bridged_id.toStdString();
 }
 
+bool is_bridged_impl(const mp::VMSpecs& specs,
+                     const std::vector<mp::NetworkInterfaceInfo>& host_nets,
+                     const std::string& preferred_net)
```
georgeliao, line 367: Same like the previous, maybe `preferred_net`->`prefered_net_name`.

src/daemon/daemon.cpp
```diff
@@ -3516,65 +3528,64 @@ void mp::Daemon::populate_instance_info(VirtualMachine& vm,
...
+                           config->factory->networks(),
+                           get_bridged_interface_name());
 }
 
 void mp::Daemon::add_bridged_interface(const std::string& instance_name)
 {
-    // These two use operator[] to access elements because this function is called after existence was checked.
-    mp::VMSpecs& spec = vm_instance_specs[instance_name];
-    mp::VirtualMachine::ShPtr instance = operative_instances[instance_name];
+    mp::VMSpecs& specs = vm_instance_specs.at(instance_name);
```
georgeliao, line 3540: Nice `[]` -> `at` switch.

georgeliao:
@ricab ,
Thanks for the good work.
It looks good in general, just a few minor comments to address.

include/multipass/utils.h
ricab: Hmm, I am not sure to be honest. Notice this is essentially moving existing code. I suppose it was just a generic implementation that ended up being used with vectors only. Do you think we should change it?

include/multipass/utils.h
ricab: Yeah, `target_network` would be better. I'll update it

Not so sure about id/name. There is a balance between how much information a variable name provides and how large it is. Code looks a lot better with shorter names, especially in an already verbose language like C++. In the context of this function, the fact that the type is a string already tells you that this is a designator of some sort.

src/daemon/daemon.cpp
ricab: See reply above. I don't think this needs to be tied to the type. "name" is something that I would append if I also had a variable with the actual thing (e.g. file and file_name). In this context, I think it is pretty redundant and only makes the code longer and harder to read.

include/multipass/utils.h
ricab: OK, I have changed this.

include/multipass/utils.h
ricab: Also, callers already need to interrogate the platform for this and other stuff. For example: https://github.com/canonical/multipass/blob/ac1629cff7c677a1573f557b8b7844d2825ca3e0/src/platform/backends/shared/base_virtual_machine_factory.cpp#L71 and https://github.com/canonical/multipass/blob/46afc1cc266e33e20b13cb163eaef98fd7da85d4/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp#L236

Don't you think it is better for "utilities" to be independent of the platform? Let me know your thoughts.

include/multipass/utils.h
ricab: I've updated it to `target_network`. How does that sound?

ricab:
Hey @georgeliao, thanks for the review. I have answered your comments and changed a few things. Let me know what you think!

include/multipass/utils.h
georgeliao: Hmm, indeed. making `utilities` depends on `MP_PLATFORM` may not be ideal. It looks like all callers are either `virtual_machine_factory` or `daemon`. Then moving the `find_bridge_with` function from `utilities` to `base_virtual_machine_factory.h` and make it a free function there might be a better choice. However, this requires more code change, I think we can leave it for now.

src/daemon/daemon.cpp
georgeliao: Yep, it is really a minor thing. I think we can leave it like this.

include/multipass/utils.h
ricab: OK, thanks. I reconsidered moving it there, but notice that this is totally independent from the factory or the backends. It really is a generic utility: just a find over a container for a certain network. We could have it as a static in the factory, but that would kind of break single-responsibility, locality...

ricab:
Hey @georgeliao, I have replied back in here. Thanks again.

include/multipass/utils.h
georgeliao: > We could have it as a static in the factory,

Do you mean `BaseVirtualMachineFactory::find_bridge_with` static function? That is similar with a free function in `base_virtual_machine_factory.h` but with access to data member of the BaseVirtualMachineFactory which is not needed.

georgeliao: Yep, the static function only accesses static data members. I have mixed that with friend function. As to free function might pollute file `base_virtual_machine_factory.h`, indeed, this utility is not really coupled with the `base_virtual_machine_factory` class. Overral, I think we can leave it for now.

georgeliao (approved):

ricab:
Fixed #3489

PR #3549, opened 2024-06-21
[ssh] Reduce log level to avoid log spam with GUI

ricab:
Reduce logging of ssh executions inside instances for `list` and `info` to trace level, while keeping the rest at debug.

### DIFF
--- a/tests/mock_virtual_machine.h
+++ b/tests/mock_virtual_machine.h
@@ -66,7 +66,13 @@ struct MockVirtualMachineT : public T
     MOCK_METHOD(std::string, management_ipv4, (), (override));
     MOCK_METHOD(std::vector<std::string>, get_all_ipv4, (), (override));
     MOCK_METHOD(std::string, ipv6, (), (override));
-    MOCK_METHOD(std::string, ssh_exec, (const std::string& cmd), (override));
+
+    MOCK_METHOD(std::string, ssh_exec_mock, (const std::string& cmd, bool whisper));
+    std::string ssh_exec(const std::string& cmd, bool whisper = false) override
+    {
+        return ssh_exec_mock(cmd, whisper);
+    }
+
     MOCK_METHOD(void, ensure_vm_is_running, (), (override));
     MOCK_METHOD(void, wait_until_ssh_up, (std::chrono::milliseconds), (override));
     MOCK_METHOD(void, wait_for_cloud_init, (std::chrono::milliseconds), (override));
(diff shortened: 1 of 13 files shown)


tests/mock_virtual_machine.h
```diff
@@ -66,7 +66,13 @@ struct MockVirtualMachineT : public T
     MOCK_METHOD(std::string, management_ipv4, (), (override));
     MOCK_METHOD(std::vector<std::string>, get_all_ipv4, (), (override));
     MOCK_METHOD(std::string, ipv6, (), (override));
-    MOCK_METHOD(std::string, ssh_exec, (const std::string& cmd), (override));
+
+    MOCK_METHOD(std::string, ssh_exec_mock, (const std::string& cmd, bool whisper));
+    std::string ssh_exec(const std::string& cmd, bool whisper = false) override
```
sharder996, line 71: Can you explain to me the purpose of this extra implementation of `ssh_exec()` and why you've added `ssh_exec_mock` instead of just expanding the argument list?

ricab: Hmm good thing you highlighted this, I don't think I was thinking straight... 

Google mock doesn't support having a default argument in `MOCK_METHOD` calls and normally, when overriding a virtual method with a default argument, one should make sure that the same default is used. Otherwise, different defaults could take effect depending on the concrete type of the object that the method is called on. Or it may be impossible to call the default version. When mocking, I suppose it would be OK to just ignore the default argument... In this case we call the method directly via the mock in a couple of places: https://github.com/canonical/multipass/blob/6719804fb179062cee288b1a9bdd50e675f34622/tests/test_base_virtual_machine.cpp#L1372)

But I guess we could just add a boolean to the call. Or we could add an ssh_exec without the second argument. But it may be better to avoid the default argument entirely to begin with. Let me revisit this.

ricab: Thanks for calling my attention to this @sharder996, it should be better now.

sharder996: Thanks, It looks much better now!

The private side will need to be updated as well.

sharder996 (approved):

tests/mock_virtual_machine.h
ricab: Ah indeed.

ricab: Actually, I don't think we even need a private side anymore, do we?

sharder996: Right, since it was just for a function name change.

PR #3550, opened 2024-06-24
[bash] `get` and `set` bash completions

sharder996:


andrei-toterman (approved):
Seems to be working really well. Thanks, @sharder996!

PR #3556, opened 2024-07-02
Fix broken launch on QEMU

ricab:
On backends that do not implement networks, `prepare_networking()` should be a NOOP, rather than throw. We throw NotImplemented elsewhere.


andrei-toterman (approved):
Yup, this is exactly what I did locally in order to make it work, so I think we can consider it already tested. Thanks for this, @ricab!

sharder996 (approved):

PR #3567, opened 2024-07-08
Ignore build and package files

kianmeng:


ricab (approved):
Thanks @kianmeng, this makes sense indeed.

Issue #647, opened 2019-02-28
[libvirt] multipass instances time out after snap refresh


ricab:
When using the libvirt backend, a multipass refresh sometimes fails to fully recover previously running instances. In particular, they get no IP (virsh shows them running, but no DHCP leases). This does not happen all the time though.

Steps to reproduce:
```sh
snap install --beta --classic multipass
snap set multipass driver=LIBVIRT
multipass launch
snap refresh --edge multipass
multipass list
```

Recovering of suspended instances after a reboot of the host was also reported to suffer from the same issue.

Saviq:
I believe this is because we rely on `dnsmasq`'s leases file that gets reset and so we don't know the IP of the instance that resumed. We need to cache the last known IP per instance or not destroy the leases file.

townsend2010:
Since we have recently moved to using host based `libvirt`, I wonder how we should handle the restart of `multipassd` wrt to suspending instances when using the libvirt backend?

Saviq:
If we can tell multipassd that it's shutting down for a restart (as opposed to system shutdown), we can probably keep them going.

townsend2010:
Yeah, good point.  We need to know _why_ `multipassd` is shutting down.

At any rate, I think this particular bug goes away since `libvirtd` and its `dnsmasq` will no longer be restarting on a multipass snap refresh.  I'll do some tests to see and close if I can't reproduce.

townsend2010:
I have been unable to reproduce this since moving `libvirt` out of the snap.  I do think we should try to be more "nice" and not suspend `libvirt` based instances when the multipass snap refreshes.  I'll open a separate bug for that though.

PR #3573, opened 2024-07-09
[linux/qemu] Qemu bridging v2

sharder996:
Public side of [#647](https://github.com/canonical/multipass-private/pull/647)

The spec for this feature can be found [here](https://docs.google.com/document/d/1UAZgCHHjdY_RmYCIir-p9sDpa2M7uRP7P81Tx2qM0c8/edit).

This second version of QEMU network bridging was required because of the change in implementation of `find_bridge_with()`. Since most of bridged networks was already available on macOS/QEMU and implemented, this PR is mostly just for enabling and wiring up the function calls needed to create a network interface on Linux.

The key difference on Linux is the presence of `bridge_helper.c`. Below is an excerpt from the spec on the purpose of the bridge helper:

> Finally, to use bridges on QEMU, on Linux, we will use a bridge helper. The QEMU bridge helper,  shipped with QEMU, is a simple program which checks a file in `/etc/` to see if using a certain bridge is allowed; then it sets up the devices needed for QEMU to use the bridge. Access to this file would be denied in our snap. Instead, our approach will be to fork the source of the bridge helper and remove the need to check for this file, thus allowing all the bridges we create. Then, from the QEMU command line, we must specify the path to our binary, for avoiding executing the default bridge helper.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,7 +240,9 @@ else()
       add_definitions(-DMULTIPASS_COMPILER_GCC)
     endif()
 
-    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds -Wnon-virtual-dtor)
+    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds)
+
+    string(APPEND CMAKE_CXX_FLAGS " -Wnon-virtual-dtor")
 
     if(NOT ${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "^arm")
       add_compile_options(-Wcast-align)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,9 +240,7 @@ else()
       add_definitions(-DMULTIPASS_COMPILER_GCC)
     endif()
 
-    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds)
-
-    string(APPEND CMAKE_CXX_FLAGS " -Wnon-virtual-dtor")
+    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds $<$<COMPILE_LANGUAGE:CXX>:-Wnon-virtual-dtor>)
 
     if(NOT ${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "^arm")
       add_compile_options(-Wcast-align)
--- a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
+++ b/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
@@ -221,3 +221,9 @@ void mp::QemuPlatformDetail::prepare_networking(std::vector<NetworkInterface>& /
 {
     // Nothing to do here until we implement networking on this backend
 }
+
+std::string mp::QemuPlatformDetail::create_bridge_with(const NetworkInterfaceInfo& interface) const
+{
+    assert(interface.type == "ethernet");
+    return MP_BACKEND.create_bridge_with(interface.id);
+}
--- a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
+++ b/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
@@ -195,7 +195,8 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
          << QString::fromStdString(fmt::format("tap,ifname={},script=no,downscript=no,model=virtio-net-pci,mac={}",
                                                tap_device_name, vm_desc.default_mac_address));
 
-    const auto bridge_helper_exec_path = QDir(QCoreApplication::applicationDirPath()).filePath("bridge_helper");
+    const auto bridge_helper_exec_path =
+        QDir(QCoreApplication::applicationDirPath()).filePath(BRIDGE_HELPER_EXEC_NAME_CPP);
 
     for (const auto& extra_interface : vm_desc.extra_interfaces)
     {
--- a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
+++ b/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
@@ -229,9 +229,8 @@ void mp::QemuPlatformDetail::set_authorization(std::vector<NetworkInterfaceInfo>
     const auto& br_nomenclature = MP_PLATFORM.bridge_nomenclature();
 
     for (auto& net : networks)
-        if (net.needs_authorization && net.type == "ethernet" &&
-            mpu::find_bridge_with(networks, net.id, br_nomenclature))
-            net.needs_authorization = false;
+        if (net.type == "ethernet" && !mpu::find_bridge_with(networks, net.id, br_nomenclature))
+            net.needs_authorization = true;
 }
 
 std::string mp::QemuPlatformDetail::create_bridge_with(const NetworkInterfaceInfo& interface) const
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -240,7 +240,9 @@ else()
       add_definitions(-DMULTIPASS_COMPILER_GCC)
     endif()
 
-    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds -Wnon-virtual-dtor)
+    add_compile_options(-Wextra -Wempty-body -Wformat-security -Winit-self -Warray-bounds)
```
ricab, line 243: Why is this needed?

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,27 +195,47 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
 }
 
-mp::QemuPlatform::UPtr mp::QemuPlatformFactory::make_qemu_platform(const Path& data_dir) const
+void mp::QemuPlatformDetail::set_authorization(std::vector<NetworkInterfaceInfo>& networks)
 {
-    return std::make_unique<mp::QemuPlatformDetail>(data_dir);
+    const auto& br_nomenclature = MP_PLATFORM.bridge_nomenclature();
+
+    for (auto& net : networks)
+        if (net.type == "ethernet" && net.needs_authorization &&
```
ricab, line 232: Tiny irrelevant detail: better place the check for `net.needs_authorization` first and avoid the string comparison in many cases.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,27 +195,47 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
...
+            mpu::find_bridge_with(networks, net.id, br_nomenclature))
+            net.needs_authorization = false;
```
ricab, line 234: But where is this ever set to true? 

In LXD, it is set in [here](https://github.com/canonical/multipass/blob/1f67f00527092c7a571529a86986fed4ae7d7add/src/platform/backends/lxd/lxd_virtual_machine_factory.cpp#L64-L67). Seeing as this is really a platform-level thing, I am thinking that should instead be set immediately when reading networks from the system in the Platform. WDYT?

ricab (commented):
Hey @sharder996, thank you for your work on this. I did a quick pass and I have a couple of questions.

CMakeLists.txt
sharder996: The `non-virtual-dtor` is not a valid flag when compiling the `bridge_helper.c` which is why it's explicitly added to `CMAKE_CXX_FLAGS` a few lines below.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
sharder996: Like setting it [here](https://github.com/canonical/multipass/blob/1f67f00527092c7a571529a86986fed4ae7d7add/src/platform/platform_linux.cpp#L152), then?

ricab: That is what I was thinking, but now I am not sure anymore, because not all backends need that. Namely Virtualbox, which can just connect to an interface without a dedicated switch. I guess the easiest for now is to just add the logic to necessary in here — but set it to `true` as necessary, not to `false` (unless I am missing something).

CMakeLists.txt
ricab: Hmm, couldn't we just compile the bridge helper as cpp? It was already adapted for us, right?

ricab: Meh, nevermind. Leave it

georgeliao:
@sharder996 
Since I was not in the discussions, it would be nice to have a context introduction. Something like why bridge_helper executable is needed on linux but not on macos. What was the state of the code before and why did you do some refactoring?

sharder996:
Does clang-format format `test_qemu_linux_platform.cpp` the same as the linter in CI for anyone else? Don't know why is doesn't format the same for me.

georgeliao:
@sharder996 
Check the clang-format executable version, check the clang-format configuration file your local IDE uses is the in-repo one by changing the file and see if re-apply formatting makes a difference.

georgeliao (changes requested):
@sharder996 
It looks very good. I only have a few minor comments. The unit test coverage is not high enough, it is easy to add these tests as well. However, on the other side, the uncovered part does not have any actual logic, so it is ok to skip them. I will leave it to you whether you want to push the test coverage up.  

Funcitonal testing was done as well, it works fine on linux and macos.

sharder996:
@georgeliao As for the missing test line coverage. Having no logic to actually test was the reason why I didn't pursue it.

tests/qemu/test_qemu_backend.cpp
```diff
@@ -801,15 +802,24 @@ TEST_F(QemuBackend, createsQemuSnapshotsFromJsonFile)
...
+    auto [mock_platform, guard] = mpt::MockPlatform::inject();
+    EXPECT_CALL(*mock_platform, get_network_interfaces_info)
+        .WillOnce(Return(std::map<std::string, mp::NetworkInterfaceInfo>{
+            {"lxdbr0", {"lxdbr0", "bridge", "gobbledygook"}},
+            {"mpbr0", {"mpbr0", "bridge", "gobbledygook"}},
+            {"virbr0", {"virbr0", "bridge", "gobbledygook"}},
+            {"mpqemubr0", {"mpqemubr0", "bridge", "gobbledygook"}},
+            {"enxe4b97a832426", {"enxe4b97a832426", "ethernet", "gobbledygook"}}}));
+
+    EXPECT_NO_THROW(backend.networks());
```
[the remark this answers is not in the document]
georgeliao: I do not think what you are trying to achieve is possible. 

The syntax `mock_qemu_platform->QemuPlatform::is_network_supported(arg)` is using the derived class (`MockQemuPlatform`) instance to call the base class (`QemuPlatform`) function implementation because of function hiding. This is useful when you want to call the base class function from a derived class instance. However, this only works if the base class has the function implementation and it is not the case here because `QemuPlatform::is_network_supported` is pure virtual. 

And there is no way it can somehow dispatch to `QemuPlatformDetail::is_network_supported`, it is a different derived class of `QemuPlatform` and `MockQemuPlatform` knows nothing about it, he only knows himself and the base class. 

Feel free to play with the [code snippet](https://godbolt.org/z/cd7qYnT14) to understand better.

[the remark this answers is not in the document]
sharder996: Yeah, I kinda thought as much. I wanted to do something like this in order to get some meaningful coverage on some of the new functions in the qemu platform.

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -192,27 +195,46 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
          << QString::fromStdString(fmt::format("tap,ifname={},script=no,downscript=no,model=virtio-net-pci,mac={}",
                                                tap_device_name, vm_desc.default_mac_address));
 
+    auto bridge_helper_path = QCoreApplication::applicationDirPath() + "/bridge_helper";
```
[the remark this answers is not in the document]
ricab: If it is not difficult to do, I would say yes. If you start running into trouble and needing to invest more time for any reason, I would say leave it.

georgeliao (dismissed):

ricab (approved):
Good stuff, well done!

Issue #3569, opened 2024-07-09
Multipass info does not work right after launching a instance


georgeliao:
**Describe the bug**
`multipass info` does not work right after launching an instance, it worked later without changing anything. 
**To Reproduce**
How, and what happened?
Just simply run `multipass launch -n vm4 bionic && multipass info vm4 -vvvv`

**Expected behavior**
`multipass info vm4 -vvvv` should work in regardless of how soon after launch finishes

**Logs**
```
Launched: vm4
[2024-07-09T15:33:00.576] [trace] [ssh session] Executing 'echo loadavg: "$(eval "cat /proc/loadavg | cut -d ' ' -f1-3")"; echo mem_usage: "$(eval "free -b | grep 'Mem:' | awk '{printf \$3}'")"; echo mem_total: "$(eval "free -b | grep 'Mem:' | awk '{printf \$2}'")"; echo disk_usage: "$(eval "df -t ext4 -t vfat --total -B1 --output=used | tail -n 1")"; echo disk_total: "$(eval "df -t ext4 -t vfat --total -B1 --output=size | tail -n 1")"; echo cpus: "$(eval "nproc")"; echo cpu_times: "$(eval "head -n1 /proc/stat")"; echo uptime: "$(eval "uptime -p | tail -c+4")"; echo current_release: "$(eval "cat /etc/os-release | grep 'PRETTY_NAME' | cut -d \\\" -f2")"'
[2024-07-09T15:33:00.594] [trace] [ssh process] /home/georgel/projects/canonical_projects/multipass/src/ssh/ssh_process.cpp:167 read_stream(type = 0, timeout = -1):
[2024-07-09T15:33:00.594] [trace] [ssh process] /home/georgel/projects/canonical_projects/multipass/src/ssh/ssh_process.cpp:189 read_stream(): num_bytes = 211
[2024-07-09T15:33:00.594] [trace] [ssh process] /home/georgel/projects/canonical_projects/multipass/src/ssh/ssh_process.cpp:189 read_stream(): num_bytes = 0
info failed: yaml-cpp: error at line 9, column 1: bad conversion
```

**Additional info**
 - OS: [Ubuntu 24.04]
- `multipass version`
- 1.15.0-dev.2327+gcbafad3fc and 1.14.0-rc5
- `multipass get local.driver`
- qemu

**Additional context**
[...]

PR #3574, opened 2024-07-10
Improve the uptime check isnull

georgeliao:


src/daemon/runtime_instance_info_helper.cpp
```diff
@@ -94,7 +94,9 @@ void mp::RuntimeInstanceInfoHelper::populate_runtime_info(mp::VirtualMachine& vm
     info->set_disk_total(results[Keys::disk_total_key].as<std::string>());
     info->set_cpu_count(results[Keys::cpus_key].as<std::string>());
     instance_info->set_cpu_times(results[Keys::cpu_times_key].as<std::string>());
-    instance_info->set_uptime(results[Keys::uptime_key].as<std::string>());
+    // Some older versions Ubuntu might return nothing from "uptime -p" command, which makes the
+    // results[Keys::uptime_key] null, so calling "T Node::as(const S& fallback)" makes code more robust here
+    instance_info->set_uptime(results[Keys::uptime_key].as<std::string>("0 minutes"));
```
ricab, line 99: How about this?

```suggestion
    // In some older versions of Ubuntu "uptime -p" prints only "up" right after startup. In those cases,
    // results[Keys::uptime_key] would be null.
    instance_info->set_uptime(results[Keys::uptime_key].as<std::string>(/* fallback = */ "0 minutes"));
```

ricab:
Improved fix for #3569

andrei-toterman (dismissed):
Tested it and seems to be working just fine.

ricab (approved):

ricab (approved):

PR #3585, opened 2024-07-17
LXD deadlock quickfix

andrei-toterman:
The `ssh_exec` method used to acquire the `state` mutex for its entire duration, which might've led to a deadlock if the method entered the `renew_ssh_session` flow, since that also tries to get the mutex. Now `ssh_exec` locks for more discrete periods of time.

This is not a final solution to the underlying problem. But it does seems to bring back the behavior observed so far. An `info` call might still block while a VM is launching, but at least it does not deadlock the entire daemon, and only gives an error and can be retried.

### DIFF
--- a/src/platform/backends/shared/base_virtual_machine.cpp
+++ b/src/platform/backends/shared/base_virtual_machine.cpp
@@ -221,7 +221,7 @@ void mp::BaseVirtualMachine::check_state_for_shutdown(bool force)
 
 std::string mp::BaseVirtualMachine::ssh_exec(const std::string& cmd, bool whisper)
 {
-    const std::unique_lock lock{state_mutex};
+    std::unique_lock lock{state_mutex};
 
     std::optional<std::string> log_details = std::nullopt;
     bool reconnect = true;
@@ -235,7 +235,9 @@ std::string mp::BaseVirtualMachine::ssh_exec(const std::string& cmd, bool whispe
             mpl::log(logging::Level::info, vm_name, msg);
 
             reconnect = false; // once only
+            lock.unlock();
             renew_ssh_session();
+            lock.lock();
         }
 
         try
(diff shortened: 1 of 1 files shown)


andrei-toterman:
@ricab, given what we recently discussed about this not protecting the `ssh_session` variable and that it would be undesirable to add another locking of the mutex around that, what do you think if we just keep the original lock at the top of the function, but manually unlock it before entering `renew_ssh_session` and locking after getting out of it? Then everything that happens in `ssh_exec` happens while having the lock, with the exception of `renew_ssh_session` so that it avoids the deadlock?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -258,8 +260,11 @@ std::string mp::BaseVirtualMachine::ssh_exec(const std::string& cmd, bool whispe
 
 void mp::BaseVirtualMachine::renew_ssh_session()
 {
-    if (!MP_UTILS.is_running(current_state())) // spend time updating state only if we need a new session
-        throw SSHException{fmt::format("SSH unavailable on instance {}: not running", vm_name)};
+    {
+        const std::unique_lock lock{state_mutex};
+        if (!MP_UTILS.is_running(current_state())) // spend time updating state only if we need a new session
+            throw SSHException{fmt::format("SSH unavailable on instance {}: not running", vm_name)};
+    }
```
ricab, line 267: Changes to the `ssh_session` below are still unsafe (if not super likely to cause issues).

ricab (approved):
Thank you @andrei-toterman! 

As discussed, I am taking this in as a stop-gap measure to allow the release to go through, with a view to a proper fix later on (probably based around only changing the state to Running after cloud-init finishes, as in other backends).

PR #3601, opened 2024-07-24
[workflows] reduce retention time of snap packages

sharder996:
Multipass is currently using a lot of storage in GitHub workflows. This can easily be addressed by reducing the number of days that artifacts are retained.

Additionally, built packages are also uploaded to the snapcraft store which persist for a period of 30 days from the point of the last update to the channel.

https://snapcraft.io/docs/channels


ricab (commented):
Hey @sharder996, could you please explain the rationale for this one briefly? Thanks!

ricab (approved):
Thank you!

PR #3619, opened 2024-08-01
[gui] when clicking the select all checkbox, respect instance filters

### LINKED ISSUE #3618: [GUI] Bulk selection affects invisible instances
**Describe the bug**
In the instance list page, when "show only running instances" is toggled, the checkbox at the top selects non-running instances too, although they are invisible. Bulk actions then affect all of them.

**Expected behavior**
Only running instances would be selected.

`multipass version`: 1.14.0

andrei-toterman:
Fix #3618


ricab (approved):
LGTM, thanks!

PR #3621, opened 2024-08-02
Show in-app notification if an update is available

andrei-toterman:
This PR adds a notification if there is a Multipass update available.
There is also a refactoring for the update entry in the settings page.

### DIFF
--- a/src/client/cli/cmd/info.cpp
+++ b/src/client/cli/cmd/info.cpp
@@ -35,6 +35,9 @@ mp::ReturnCode cmd::Info::run(mp::ArgParser* parser)
     auto on_success = [this](mp::InfoReply& reply) {
         cout << chosen_formatter->format(reply);
 
+        if (term->is_live() && update_available(reply.update_info()))
+            cout << update_notice(reply.update_info());
+
         return ReturnCode::Ok;
     };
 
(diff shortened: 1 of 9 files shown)


sharder996 (changes requested):
Tested and I see the update notification. However, it only shows up on launching the GUI. If I leave the GUI running long enough for `DefaultUpdatePrompt::is_time_to_show()` to return true again, I don't get another notification telling me about a newer version.

sharder996:
As a follow-up, I'm not sure which is the preferred behaviour. Showing an update notification once on launch, or as per the update prompt timer. Multipass doesn't update that often so getting "live" update notifications might not be an issue.

ricab:
Live notifications would be better, to quickly get to people who just suspend their laptop and keep the GUI running.

sharder996 (dismissed):
Looks good!

Noticed that the update notification is not showing up in `start` and `restart` via the CLI even though we are piping the update info through the grpc reply. Addressing that is outside the scope of this PR, so I will create a bug report for it.

andrei-toterman:
@sharder996 sorry to bother, but I just realized I was over-complicating things a little when extracting the update info from the rpc reply, so I pushed an update. Again, sorry for interrupting the merge :sweat_smile:

sharder996 (approved):

PR #3625, opened 2024-08-05
Fix the bug where VirtualMachine::shutdown function throw affects multipass delete

### LINKED ISSUE #3606: Installed multipass 1.14.0 on OSX but don't have --force option
**Describe the bug**
I installed multipass 1.14.0 on OSX (Sonoma 14.0) and have a suspended multipass instance. I tried to use the --force command and receive "Unknown option 'force'."

**To Reproduce**
How, and what happened?

```
X@X-MBP iiab-calibre % multipass list              
box-20240225            Suspended         --               Ubuntu 22.04 LTS
....
X@X-MBP iiab-calibre % multipass delete box-20240225
delete failed: Cannot stop suspended instance. Use --force to override.
X@X-MBP iiab-calibre % multipass --force delete box-20240225
Unknown option 'force'.
```

**Expected behavior**
I expected the --force option to work and the instance to be deleted.

**Logs**

https://dpaste.com/5MA59LH62

**Additional info**
 - OS: OSX Sonoma 14.0
- `multipass version` 
  - multipass   1.14.0+mac
  - multipassd  1.14.0+mac
- `multipass info` 
This is also odd. I didn't see this behavior with 1.14.0-rc6: 

avni@Avnis-MBP ~ % multipass info
info failed: ssh connection failed: 'Socket error: Connection reset by peer'

- `multipass get local.driver` qemu

**Additional context**
I can delete other non-suspended instances and purge them.

### LINKED ISSUE #3624: multipass delete on suspended vm cause weird error message
**Describe the bug**
`multipass delete` on suspended vm cause weird error message

**To Reproduce**
How, and what happened?
1. `multipass launch -n vm1`
2. `multipass suspend vm1`
3. `multipass delete vm1`
delete failed: Cannot stop suspended instance. Use --force to override.

**Expected behavior**
`multipass delete` should delete the suspended VM and later it can be recovered . 

**Additional info**
 - OS: ubuntu 24.04
- `multipass version`: 1.14.0

**Additional context**
This only occurs on the suspended VM and not on the running VM or stopped VM.

georgeliao:
close #3624
close #3606

A few things were done based on our team discussion

1. skip_states check in `daemon.cpp` was removed because it is the responsibility of `VirtualMachine::shutdown` function to check the state. 
2. delegate the cmd suggestion part of the error message to CLI client. The daemon maps `VMStateInvalidException` exception to `grpc::StatusCode::FAILED_PRECONDITION`, and the CLI side interprets this code respectively. 
3. The force boolean input parameter of function `VirtualMachine::shutdown` is replaced by a 3 values enum `ShutdownPolicy`, corresponding refactor was performed as well.

Functional testing:
Please at least cover the following scenarios:

1. `multipass stop` on a deleted or non-existing VM does not trigger the `--force` suggestion
2. `multipass stop` on a VM with invalid states does trigger the `--force` suggestion. Invalid states include suspended, suspending, starting and restarting, the transtioning states require running `multipass stop` in 2nd cmd panel. 
3. `multipass delete` on a deleted or non-existing VM does not trigger the `--purge` suggestion
4. `multipass delete` on a suspended and stopped vm can retain the state meaning the recovered vm should have the same state. 
5. `multipass delete` on arunning vm will shutdown the vm gracefully and recover it to the stopped state.  
6. `multipass delete` on a VM with invalid states does trigger the `--purge` suggestion. Invalid states are only transtioning states like suspending, starting and restarting, they all require the 2nd cmd panel to run. 
7. Run these two commands in bulk with mixed state of the vms and see if they behave right. 
8. Check whether GUI encounters these exceptions. 

Note:
1. The private side is not done yet, but it will be a simple adaptation due to the `shutdown` and `check_state_for_shutdown` interface change. I would rather waiting the public PR review finish and change that later.
2. The unit tests on daemon::stop daemon::delete are lacking due to lacking frameworks which are similar to `test_daemon_snapshot_restore.cpp`. It is not a big problem to create the `stop` and `delete` version of this, but it might take more time.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -349,7 +349,7 @@ void mp::QemuVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
 {
     std::unique_lock<std::mutex> lock{state_mutex};
 
-    force_shutdown = force;
+    force_shutdown = (shutdown_policy == ShutdownPolicy::Poweroff);
 
     try
     {
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -345,7 +345,7 @@ void mp::QemuVirtualMachine::start()
     vm_process->write(qmp_execute_json("qmp_capabilities"));
 }
 
-void mp::QemuVirtualMachine::shutdown(bool force)
+void mp::QemuVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
 {
     std::unique_lock<std::mutex> lock{state_mutex};
 
@@ -353,7 +353,7 @@ void mp::QemuVirtualMachine::shutdown(bool force)
 
     try
     {
-        check_state_for_shutdown(force);
+        check_state_for_shutdown(shutdown_policy);
     }
     catch (const VMStateIdempotentException& e)
     {
@@ -361,7 +361,7 @@ void mp::QemuVirtualMachine::shutdown(bool force)
         return;
     }
 
-    if (force)
+    if (shutdown_policy == ShutdownPolicy::Poweroff)
     {
         mpl::log(mpl::Level::info, vm_name, "Forcing shutdown");
 
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -349,8 +349,6 @@ void mp::QemuVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
 {
     std::unique_lock<std::mutex> lock{state_mutex};
 
-    force_shutdown = (shutdown_policy == ShutdownPolicy::Poweroff);
-
     try
     {
         check_state_for_shutdown(shutdown_policy);
@@ -368,6 +366,7 @@ void mp::QemuVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
         if (vm_process)
         {
             mpl::log(mpl::Level::info, vm_name, "Killing process");
+            force_shutdown = true;
             lock.unlock();
             vm_process->kill();
             if (vm_process != nullptr && !vm_process->wait_for_finished(kill_process_timeout))
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -60,9 +60,10 @@ public:
 
     enum class ShutdownPolicy
     {
-        Powerdown,
-        Poweroff,
-        Halt
+        Powerdown, // gracefully shut down the vm
+        Poweroff,  // forcefully shut down the vm
+        Halt // halt the vm to non-running state. More specifically. suspended and stopped state will remain the same
+             // and running state will be shut down to stopped state
     };
 
     using UPtr = std::unique_ptr<VirtualMachine>;
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -58,12 +58,19 @@ public:
         unknown
     };
 
+    enum class ShutdownPolicy
+    {
+        Powerdown,
+        Poweroff,
+        Halt
+    };
+
     using UPtr = std::unique_ptr<VirtualMachine>;
     using ShPtr = std::shared_ptr<VirtualMachine>;
 
     virtual ~VirtualMachine() = default;
     virtual void start() = 0;
-    virtual void shutdown(bool force = false) = 0;
+    virtual void shutdown(ShutdownPolicy shutdown_policy = ShutdownPolicy::Powerdown) = 0;
     virtual void suspend() = 0;
     virtual State current_state() = 0;
     virtual int ssh_port() = 0;
--- a/src/client/cli/cmd/delete.cpp
+++ b/src/client/cli/cmd/delete.cpp
@@ -69,8 +69,8 @@ mp::ReturnCode cmd::Delete::run(mp::ArgParser* parser)
     };
 
     auto on_failure = [this](grpc::Status& status) {
-        // grpc::StatusCode::INVALID_ARGUMENT matches mp::VMStateInvalidException
-        return status.error_code() == grpc::StatusCode::INVALID_ARGUMENT
+        // grpc::StatusCode::FAILED_PRECONDITION matches mp::VMStateInvalidException
+        return status.error_code() == grpc::StatusCode::FAILED_PRECONDITION
                    ? standard_failure_handler_for(name(), cerr, status, "Use --purge to forcefully delete it.")
                    : standard_failure_handler_for(name(), cerr, status);
     };
--- a/src/client/cli/cmd/delete.cpp
+++ b/src/client/cli/cmd/delete.cpp
@@ -68,7 +68,12 @@ mp::ReturnCode cmd::Delete::run(mp::ArgParser* parser)
         return mp::ReturnCode::Ok;
     };
 
-    auto on_failure = [this](grpc::Status& status) { return standard_failure_handler_for(name(), cerr, status); };
+    auto on_failure = [this](grpc::Status& status) {
+        // grpc::StatusCode::INVALID_ARGUMENT matches mp::VMStateInvalidException
+        return status.error_code() == grpc::StatusCode::INVALID_ARGUMENT
+                   ? standard_failure_handler_for(name(), cerr, status, "Use --purge to forcefully delete it.")
+                   : standard_failure_handler_for(name(), cerr, status);
+    };
 
     using Client = grpc::ClientReaderWriterInterface<DeleteRequest, DeleteReply>;
     auto streaming_callback = [this](const mp::DeleteReply& reply, Client* client) {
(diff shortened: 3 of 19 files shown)


sharder996 (changes requested):
Firstly, I don't think this is the proper way of fixing this. It's more of a hack than an actually fix. I think we need to go back to the question of "What is non-purge delete supposed to do?". Moving `check_state_for_shutdown()` to the daemon isn't the correct proper fix either in my opinion. Instead we should be evaluating what stop is supposed to do to a suspended vm:

1. **Do nothing**: `check_state_for_shutdown()` needs to change to not throw an exception.
2. **Put the machine in a stopped state**: do the same thing as force stop.

Secondly, it would be nice if the commit history got cleaned up a bit. Currently, only [f34db7f](https://github.com/canonical/multipass/pull/3625/commits/f34db7f4e8a72a5fe27310b29bc46cb69977152e) is relevant to the final diff, so the rest can be dropped.

georgeliao:
> Firstly, I don't think this is the proper way of fixing this. It's more of a hack than an actually fix. I think we need to go back to the question of "What is non-purge delete supposed to do?". Moving check_state_for_shutdown() to the daemon isn't the correct proper fix either in my opinion. Instead we should be evaluating what stop is supposed to do in the context of a delete of a suspended VM.

To answer the question "What is non-purge delete supposed to do?", let's delve into the expected behavior of each state. 

Major states:
1. **Suspended**: Shutdown should be skipped, and the VM is moved into deleted_instance. When the VM is recovered, the VM state is still suspended. That is what the current code does. 
2. **Running**: non-force shutdown is applied, the state changes to stopped and the VM is moved into deleted_instance. When the VM is recovered, the VM state is stopped. That is what the current code does. 
3. **Stopped**: Shutdown should be skipped, and the VM is moved into deleted_instance. When the VM is recovered, the VM state is still stopped. That is what the current code does. 

So in summary, the catch exception and do nothing did the skip shutdown thing. In my opinion, this is a reasonable thing to do while the `check_state_for_shutdown` function is still in the `void mp::VirtualMachine::shutdown(bool force)` and it throws exception for both `multipass stop` and `multipass delete`. Feel free to share any other code design idea you might have. 

There are other minor states like **starting** or **restarting**, these states on cmd blocks the daemon. The user will have to open another terminal to delete the starting VM. The ideal behavior would be waiting for the starting process finish and shut it down gracefully from the running state. However, I do not think we have good thread safety for this in the first place, so tackling these states would be beyond the scope of this PR.

georgeliao:
> Secondly, it would be nice if the commit history got cleaned up a bit. Currently, only [f34db7f](https://github.com/canonical/multipass/pull/3625/commits/f34db7f4e8a72a5fe27310b29bc46cb69977152e) is relevant to the final diff, so the rest of the commits can be dropped.

This can be done, sorry about that. Initially I was thinking about put two fixes into one PR.

ricab:
> There are other minor states like **starting** or **restarting**, these states on cmd blocks the daemon. The user will have to open another terminal to delete the starting VM. The ideal behavior would be waiting for the starting process finish and shut it down gracefully from the running state. However, I do not think we have good thread safety for this in the first place, so tackling these states would be beyond the scope of this PR.

Hmm, I would not say that those states are out of scope, but I think we could simplify and not wait for a graceful shutdown either. Instead, `multipass delete` could still refuse to delete instances in those states (although with a different error message, not mentioning `--force`). The user could still use `multipass delete --purge` to force-shutdown and then delete. So, I think "suspended" is the only state where, to maintain behavior, we'd need to skip shutdown. 

Here is an option to achieve that: `mp::VirtualMachine::shutdown` could receive an enum instead of a bool, to accept a third possible value: `accepted_suspended` (or a better name). So, on `Daemon::delete_vm`, we'd now call `shutdown(purge ? OurEnum::force : OurEnum::accept_suspended)`. On stop, we'd call instead `shutdown(force ? OurEnum::force : OurEnum::plain)` or something of this sort.

georgeliao:
@sharder996 
There are two items in email message that were edited later maybe. Regardless, just clarify them

>   1. Do nothing: check_state_for_shutdown() needs to change to not throw an exception.
>   2. Put the machine in a stopped state: do the same thing as force stop.

1. The throw is intended for `multipass stop -> shutdown(force) ` and now it affects `multipass delete -> shutdown(purge) ` we can not just simply change that throw. 
2. I think doing nothing and recovering it back to suspended is more reasonable, force stop is the one can changing from suspended to stop state.

ricab:
This just occurred to me and I thought I'd note it down. We need to remember to test what happens when we pass multiple instance arguments or use `--all`. What happens when a VM in the wrong state and we do stuff like:

- `multipass delete a b c`
- `multipass delete --all`
- `multipass delete --all --purge`
- `multipass stop --force a b c`
- `multipass stop --all`
- `multipass stop --all --force`

Do we do the right thing in all those cases?

georgeliao:
> This just occurred to me and I thought I'd note it down. We need to remember to test what happens when we pass multiple instance arguments or use --all. What happens when a VM in the wrong state and we do stuff like:
> 
> multipass delete vm1 vm2 vm3
> multipass delete --all
> multipass delete --all --purge
> multipass stop --force vm1 vm2 vm3
> multipass stop --all
> multipass stop --all --force
> Do we do the right thing in all those cases?

I changed the name of the VM to vm1, vm2 and vm3 for eaiser illustration. 

It is thoughtful that you mentioned the bulk operations.  This boils down to how the command behaves when one of the VM throws during this operation. Based on [this line](https://github.com/canonical/multipass/blob/b685dbf331818a030a1c6c290bd6c90eab343b5b/src/daemon/daemon.cpp#L868) of code, it should exit on the first throwing VM and report the error to the cli and to the user. For example, if vm1 is running, vm2 is suspended and v,3 is stopped, then `multipass stop vm1 vm2 vm3` should stop vm1 and throw on vm2 and leave vm3 intact.  See the below command line output. 

```
georgel@georgel-XPS-15:~/projects/canonical_projects/multipass_build/bin$
./multipass list
vm1                     Running           10.74.98.248     Ubuntu 24.04 LTS
vm2                     Suspended         --               Ubuntu 24.04 LTS
vm3                     Stopped           --               Ubuntu 24.04 LTS
georgel@georgel-XPS-15:~/projects/canonical_projects/multipass_build/bin$
./multipass stop vm1 vm2 vm3

stop failed: Cannot shut down suspended instance vm2.
Use --force to power it off.
georgel@georgel-XPS-15:~/projects/canonical_projects/multipass_build/bin$
./multipass list
vm1                     Stopped           --               Ubuntu 24.04 LTS
vm2                     Suspended         --               Ubuntu 24.04 LTS
vm3                     Stopped           --               Ubuntu 24.04 LTS
```

Regarding these three lines, 
> multipass delete --all --purge
> multipass stop --force vm1 vm2 vm3
> multipass stop --all --force

They do not throw, so they can apply the operation to every VMs.

ricab:
Hi @georgeliao, thanks for clarifying. I did not quite understand your last comment though. Do you mean that those commands do not go through the line of code you linked to earlier?

In any case, I still think this should be manually tested, just to verify expectations.

georgeliao:
> Hi @georgeliao, thanks for clarifying. I did not quite understand your last comment though. Do you mean that those commands do not go through the line of code you linked to earlier?

What I was trying to say is that things are simpler without throwing. The loop will not exit early. Every VM in that selection will get the operation applied or skipped. It is not that different from applying the operation to a single VM.

ricab:
In some cases it will still need to throw, right? Anyway, just trying to make sure the behavior is verified experimentally.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -345,23 +345,23 @@ void mp::QemuVirtualMachine::start()
     vm_process->write(qmp_execute_json("qmp_capabilities"));
 }
 
-void mp::QemuVirtualMachine::shutdown(bool force)
+void mp::QemuVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
 {
     std::unique_lock<std::mutex> lock{state_mutex};
 
-    force_shutdown = force;
+    force_shutdown = (shutdown_policy == ShutdownPolicy::Poweroff);
```
georgeliao, line 353: @sharder996 I just noticed that `force_shutdown` value should be set to true later in order to truely pair up with [this line](https://github.com/canonical/multipass/blob/b685dbf331818a030a1c6c290bd6c90eab343b5b/src/platform/backends/qemu/qemu_virtual_machine.cpp#L663),  `check_state_for_shutdown` has a chance to throw and leaves `force_shutdown` to true without actually executing forceful shutdown. Something we overlooked in the silencing error PR. I think the right place to set it might be after line 370, `mpl::log(mpl::Level::info, vm_name, "Killing process");`. What do you think? That makes 

```
            mpl::log(mpl::Level::info, vm_name, "Killing process");
            force_shutdown = true;
```

georgeliao:
@ricab @sharder996 , the followup codework is done based on our team discussion, so it is ready another round of review. Please read the intro description first to get an overview of the steps and the functional testing cases.

sharder996 (approved):
LGTM! Thanks for the work in to make this better, Jia!

ricab (approved):
Quick review and LGTM, thanks Jia!

Leaving it with @sharder996.

sharder996:
@georgeliao Just needs the private side now 😄

sharder996 (dismissed):

ricab (approved):
LGTM, thanks Jia!

There is a commit message that I find confusing, but I am not going to block on that. Merge away!

PR #3628, opened 2024-08-05
update flutter to 3.22.3 and our deps

andrei-toterman:


andrei-toterman:
Seems like 3.22 is broken in the snap. Related to flutter/flutter#152297.
Since this update is not urgent, I'd rather wait for the issue to be solved than to upgrade only to 3.19.6.

andrei-toterman:
superseded by #3655

PR #3632, opened 2024-08-09
correct the condition check of suspended vm on force stop

### LINKED ISSUE #3626: multipass stop --force can not stop suspended vm if it does not have a suspend tag
**Describe the bug**
`multipass stop --force` can not stop suspended vm if it does not have a suspend tag

**To Reproduce**
How, and what happened?
1. `multipass launch -n vm1`
2. Manually delete the snapshot tag of the qemu image. `qemu-img snapshot -d suspend <image_path_of_the_vm>`
3. Run `multipass stop vm1 --force` or `multipass delete vm1 --purge`
stop failed: Failed to delete suspend image: qemu-img failed (Process returned exit code: 1) with output:
qemu-img: Could not delete snapshot 'suspend': snapshot not found

**Expected behavior**
Despite the faulty state (lacking a suspend tag but suspended) the VM is in, it should still delete or stop the VM forcefully. 

**Additional info**
 - OS: ubuntu 24.04
- `multipass version`: 1.14.0

**Additional context**
It is a very unusual state to get into, however, it did happen. Thus, to make the code more robust, it would be nice to handle the state gracefully.

georgeliao:
close #3626

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -380,13 +380,6 @@ void mp::QemuVirtualMachine::shutdown(bool force)
             mpl::log(mpl::Level::debug, vm_name, "No process to kill");
         }
 
-        if (state == State::suspended && !mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
-        {
-            mpl::log(mpl::Level::warning,
-                     vm_name,
-                     fmt::format("{} is suspended, but the image does not have the suspension snapshot", vm_name));
-        }
-
         if (mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
         {
             if (state != State::suspended)
@@ -399,6 +392,12 @@ void mp::QemuVirtualMachine::shutdown(bool force)
 
             mpl::log(mpl::Level::info, vm_name, "Deleting suspend image");
             mp::backend::delete_instance_suspend_image(desc.image.image_path, suspend_tag);
+        }
+        else if (state == State::suspended)
+        {
+            mpl::log(mpl::Level::warning,
+                     vm_name,
+                     fmt::format("{} is suspended, but the image does not have the suspension snapshot", vm_name));
         }
 
         state = State::off;
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -384,7 +384,7 @@ void mp::QemuVirtualMachine::shutdown(bool force)
         {
             mpl::log(mpl::Level::warning,
                      vm_name,
-                     fmt::format("{} is suspended, but the image does not have the suspension snapshot.", vm_name));
+                     fmt::format("{} is suspended, but the image does not have the suspension snapshot", vm_name));
         }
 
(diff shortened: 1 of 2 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -380,7 +380,7 @@ void mp::QemuVirtualMachine::shutdown(bool force)
             mpl::log(mpl::Level::debug, vm_name, "No process to kill");
         }
 
-        if (state == State::suspended || mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
+        if (state == State::suspended && mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
```
sharder996, line 383: Correct me if I'm wrong, but wouldn't always deleting the suspend snapshot be the correct behaviour for force stop? Force stop is supposed to "wipe the slate clean", so getting rid of the snapshot seems like the expected result.

```suggestion
        if (mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
```

georgeliao: That is a very good question actually. I think the state transition of VM guarantees that the image should only have the suspend_tag when the VM is in suspended state. To be more specific, the suspend operation from a running vm creates the suspend_tag, both start and force-stop consume the suspend_tag. 

However, this change might not make a difference in practice, but definitely make things more robust in case the user gets the VM into some unexpected situation.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -380,7 +380,7 @@ void mp::QemuVirtualMachine::shutdown(bool force)
             mpl::log(mpl::Level::debug, vm_name, "No process to kill");
         }
 
-        if (state == State::suspended || mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
+        if (mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
         {
             mpl::log(mpl::Level::info, vm_name, "Deleting suspend image");
```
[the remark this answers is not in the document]
ricab: And vice-versa, actually. If it is suspended but does not have a suspend tag, I would also like to know.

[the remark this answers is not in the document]
georgeliao: Actually, I like this warning log suggestion. It is definitely an improvement.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -380,8 +380,23 @@ void mp::QemuVirtualMachine::shutdown(bool force)
             mpl::log(mpl::Level::debug, vm_name, "No process to kill");
         }
 
-        if (state == State::suspended || mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
+        if (state == State::suspended && !mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
```
sharder996, line 383: You can put this into an else block of the below conditional to avoid calling qemu-img unnecessarily.

sharder996 (changes requested):

sharder996 (dismissed):

PR #3636, opened 2024-08-15
[qemu] silence force stop logs

### LINKED ISSUE #3591: [qemu] error level logs are outputted when force stopping an instance
**Describe the bug**
When you force stop an instance or delete purge it, Multipass outputs some scary logs that make it appear as if something went wrong. Happens on any platform running QEMU.

**To Reproduce**
`multipass stop --force foo` or `multipass delete --purge foo`

```
% multipass stop --force bar
[2024-07-16T21:31:50.536] [error] [bar] process error occurred Crashed program: qemu-system-aarch64; error: Process crashed
[2024-07-16T21:31:50.538] [error] [bar] error: program: qemu-system-aarch64; error: Process crashed
```

**Expected behavior**
No logs are outputted as the command has completed successfully

sharder996:
Downgrades log levels when forcing a shutdown. This way we still get logs from the vm process, but they are not shown to the user through the CLI.

Fixes #3591

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -626,7 +626,7 @@ void mp::QemuVirtualMachine::initialize_vm_process()
             // out any scary error messages for this state
             if (update_shutdown_status)
             {
-                auto log_level = force_shutdown ? mpl::Level::info : mpl::Level::error;
+                const auto log_level = force_shutdown ? mpl::Level::info : mpl::Level::error;
                 mpl::log(log_level,
                          vm_name,
                          fmt::format("process error occurred {} {}", utils::qenum_to_string(error), error_string));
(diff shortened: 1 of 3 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -646,7 +650,8 @@ void mp::QemuVirtualMachine::initialize_vm_process()
             }
             else
             {
-                mpl::log(mpl::Level::error, vm_name, fmt::format("error: {}", process_state.error->message));
+                auto log_level = force_shutdown ? mpl::Level::info : mpl::Level::error;
+                mpl::log(log_level, vm_name, fmt::format("error: {}", process_state.error->message));
```
georgeliao, line 654: I think we should reset `force_shutdown` value to false after a force shutdown, otherwise the started VM will have `force_shutdown` value true without forcefully shutdown. Then the operations should report the message as an error will have it silenced. (By the way, I am not sure which exact operation should trigger that, be free to share if you know. )

If I did not understand wrong, `Process::error_occurred` happens first, then `Process::finished`, so maybe add `force_shutdown = false;` here which is the end of the sequence of actions.

georgeliao (changes requested):
@sharder996 

Thanks for the good work. It looks good in general. Just a few minor comments.

georgeliao (approved):
The PR looks good to me. 
@ricab In case you want to give a scan, otherwise we can go ahead and merge.

ricab (approved):
LGTM, thank you!

PR #3644, opened 2024-08-21
On Windows, document about winget

menuet:


ricab (commented):
Winget is not maintained by us and we'll soon move from our current package to an MSI one. The winget recipe will need to be updated at that point.

After that, if all is good, we can add it to our docs.

ricab:
Hi @menuet, we ended up deciding to [document](https://multipass.run/docs/install-multipass#install) the official approach and then "package managers" in general (since they are not under our control).

PR #3649, opened 2024-08-28
clarify behavior of `multipass shell` in short_help and description

### LINKED ISSUE #1133: `multipass shell` launching a VM is unexpected
I just typed `multipass shell` and accidentally hit Return before typing my instance name in, and was very surprised to see multipass launch a new VM. This was particularly confusing given that the short help says "Open a shell on a _running_ instance" (emphasis mine, of course), so that's what I assumed it did.

Once I'd seen this happen, I looked at the full help text, and I think we also have some confusing wording there. Specifically, "If the instance is not running, an attempt is made to start it" is inaccurate, as `multipass` did not attempt to `start` it, it instead `launch`ed a completely new VM.

levkropp:
close #1133

### DIFF
--- a/src/client/cli/cmd/shell.cpp
+++ b/src/client/cli/cmd/shell.cpp
@@ -128,7 +128,7 @@ QString cmd::Shell::short_help() const
 QString cmd::Shell::description() const
 {
     return QStringLiteral(
-        "Open a shell prompt on the instance. If the instance is not running, it will automatically be started.");
+        "Open a shell prompt on the instance. If the instance is not running, it will be started automatically.");
 }
 
 mp::ParseCode cmd::Shell::parse_args(mp::ArgParser* parser)
--- a/src/client/cli/cmd/shell.cpp
+++ b/src/client/cli/cmd/shell.cpp
@@ -122,7 +122,7 @@ std::vector<std::string> cmd::Shell::aliases() const
 
 QString cmd::Shell::short_help() const
 {
-    return QStringLiteral("Open a shell on the primary or specified instance");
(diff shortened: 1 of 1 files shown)


ricab:
Hi @giuliazanchi, how does this change look to you?

src/client/cli/cmd/shell.cpp
```diff
@@ -117,12 +117,12 @@ std::vector<std::string> cmd::Shell::aliases() const
 
 QString cmd::Shell::short_help() const
 {
-    return QStringLiteral("Open a shell on a running instance");
+    return QStringLiteral("Open a shell on the primary or specified instance");
```
giuliazanchi, line 120: I'd simplify the short help further by just saying: 
```suggestion
    return QStringLiteral("Open a shell on an instance");
```

src/client/cli/cmd/shell.cpp
```diff
@@ -117,12 +117,12 @@ std::vector<std::string> cmd::Shell::aliases() const
...
 }
 
 QString cmd::Shell::description() const
 {
-    return QStringLiteral("Open a shell prompt on the instance.");
+    return QStringLiteral("Open a shell prompt on the instance. If the instance is not running, it will automatically be started.");
```
giuliazanchi, line 125: Good idea to mention this! Just a minor rephrasing suggestion:

```suggestion
    return QStringLiteral("Open a shell prompt on the instance. If the instance is not running, it will be started automatically.");
```

giuliazanchi (dismissed):

ricab (approved):
Looks good @levkropp, thanks!

PR #3653, opened 2024-08-30
[qemu] Make logging code terser

ricab:
Condense code to log mismatches between suspension snapshot and VM state. Just trying to reduce the space that is taken by logging code, which dilutes actual logic for the reader.

This could be further improved by adding a formatter for the State enum to `multipass/format.h`. That would be useful in other places too, but we should then probably also move the corresponding logic from the CLI formatters, so leaving that for now.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -380,24 +380,16 @@ void mp::QemuVirtualMachine::shutdown(bool force)
             mpl::log(mpl::Level::debug, vm_name, "No process to kill");
         }
 
-        if (mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag))
-        {
-            if (state != State::suspended)
-            {
-                mpl::log(
-                    mpl::Level::warning,
-                    vm_name,
-                    fmt::format("{} has the image suspension snapshot, but it is not in the suspended state", vm_name));
-            }
+        const auto has_suspend_snapshot = mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag);
+        if (has_suspend_snapshot != (state == State::suspended)) // clang-format off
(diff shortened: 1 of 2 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -380,25 +380,17 @@ void mp::QemuVirtualMachine::shutdown(bool force)
...
-        {
-            if (state != State::suspended)
-            {
-                mpl::log(
-                    mpl::Level::warning,
-                    vm_name,
-                    fmt::format("{} has the image suspension snapshot, but it is not in the suspended state", vm_name));
-            }
+        const auto has_suspend_snapshot = mp::backend::instance_image_has_snapshot(desc.image.image_path, suspend_tag);
+        if (has_suspend_snapshot != (state == State::suspended)) // clang-format off
```
georgeliao, line 384: Nice trick to utilize the symmetry.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -380,25 +380,17 @@ void mp::QemuVirtualMachine::shutdown(bool force)
...
+            mpl::log(mpl::Level::warning, vm_name, fmt::format("Image has {} suspension snapshot, but the state is {}",
+                                                               has_suspend_snapshot ? "a" : "no",
+                                                               static_cast<short>(state))); // clang-format on
```
georgeliao, line 388: The only problem of this is the integer value returns to us via `static_cast<short>(state)`, however, enum -> string map or a switch statement may not worth it.

georgeliao (commented):
@ricab , it looks good, only comments on the code, no requested change.

src/platform/backends/qemu/qemu_virtual_machine.cpp
ricab: Right, this is what my comment above about formatters is about. Ideally, we'd write a Formatter for VM::State... see above.

PR #3654, opened 2024-08-30
intercept `val` in driver_interpreter and remove capitals

### LINKED ISSUE #3565: Accept local.driver values regardless of the case
For instance, `multipass set local.driver=VirtualBox` should work as well as `multipass set local.driver=virtualbox`. Also, we should probably accept "Hyper-V"/"hyper-v" along with "hyperv".

levkropp:
close #3565

### DIFF
--- a/src/daemon/daemon_init_settings.cpp
+++ b/src/daemon/daemon_init_settings.cpp
@@ -54,6 +54,13 @@ QString persistent_settings_filename()
 QString driver_interpreter(QString val)
 {
     val = val.toLower();
+
+    if (val == QStringLiteral("hyper-v"))
+        val = QStringLiteral("hyperv");
+
+    if (val == QStringLiteral("vbox"))
+        val = QStringLiteral("virtualbox");
+
     if (!MP_PLATFORM.is_backend_supported(val))
         throw mp::InvalidSettingException(mp::driver_key, val, "Invalid driver");
 
--- a/src/daemon/daemon_init_settings.cpp
+++ b/src/daemon/daemon_init_settings.cpp
@@ -57,8 +57,7 @@ QString driver_interpreter(QString val)
 
     if (val == QStringLiteral("hyper-v"))
         val = QStringLiteral("hyperv");
-
-    if (val == QStringLiteral("vbox"))
+    else if (val == QStringLiteral("vbox"))
         val = QStringLiteral("virtualbox");
 
     if (!MP_PLATFORM.is_backend_supported(val))
--- a/src/daemon/daemon_init_settings.cpp
+++ b/src/daemon/daemon_init_settings.cpp
@@ -53,6 +53,7 @@ QString persistent_settings_filename()
 
 QString driver_interpreter(QString val)
 {
+    val = val.toLower();
     if (!MP_PLATFORM.is_backend_supported(val))
         throw mp::InvalidSettingException(mp::driver_key, val, "Invalid driver");
 
--- a/tests/test_global_settings_handlers.cpp
+++ b/tests/test_global_settings_handlers.cpp
@@ -267,6 +267,34 @@ TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatAcceptsValidBackend
     ASSERT_NO_THROW(handler->set(key, val));
 }
 
+TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatTransformsHyperVDriver)
+{
+    const auto key = mp::driver_key;
+    const auto val = "hyper-v";
+    const auto transformed_val = "hyperv";
+
+    mp::daemon::register_global_settings_handlers();
+
+    EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val))).Times(1);
+    inject_mock_qsettings();
+
+    ASSERT_NO_THROW(handler->set(key, val));
+}
+
+TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatTransformsVBoxDriver)
+{
+    const auto key = mp::driver_key;
+    const auto val = "vbox";
+    const auto transformed_val = "virtualbox";
+
+    mp::daemon::register_global_settings_handlers();
+
+    EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val))).Times(1);
+    inject_mock_qsettings();
+
+    ASSERT_NO_THROW(handler->set(key, val));
+}
+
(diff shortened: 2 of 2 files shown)


src/daemon/daemon_init_settings.cpp
```diff
@@ -53,6 +53,7 @@ QString persistent_settings_filename()
 
 QString driver_interpreter(QString val)
 {
+    val = val.toLower().remove("-");
```
georgeliao, line 56: This also accepts something like `HyPeR-----V`, right? Is that ok?  I think we might need a regex format check on the input val first and then interpret.

levkropp: > Is that ok?

Since hyper-v is the only hypervisor that contains dashes IIRC, it might be better to only apply toLower() to `val` and then whenever there is a check for `if (driver == "hyperv")` to also check for `if (driver == "hyperv" || driver == "hyper-v")`. What do you think about this solution instead?

ricab: I would like to add a third option :slightly_smiling_face:  How about we do only `toLower()` here and convert hyper-v to hyperv on the Windows platform?

ricab: That could even be done in a separate PR, to avoid having to sync branches.

georgeliao: > Since hyper-v is the only hypervisor that contains dashes IIRC, it might be better to only apply toLower() to val and then whenever there is a check for if (driver == "hyperv") to also check for if (driver == "hyperv" || driver == "hyper-v"). What do you think about this solution instead?

This sounds good. One more hard-coded string is not that bad in this scenario.

ricab (dismissed):
Thank you Lev, I haven't tested but this looks good to me. Fine with me if @georgeliao approves.

georgeliao (dismissed):

georgeliao:
> Thank you Lev, I haven't tested but this looks good to me. Fine with me if @georgeliao approves.

The only concern of mine is that we will accept all the partial upper case name cases like, `hYPER-V`, `vBOx`, `virtUALBox`, because we convert the whole word to lower cases. However, I think these names should be also acceptable.

src/daemon/daemon_init_settings.cpp
```diff
@@ -53,6 +53,14 @@ QString persistent_settings_filename()
 
 QString driver_interpreter(QString val)
 {
+    val = val.toLower();
+
+    if (val == QStringLiteral("hyper-v"))
+        val = QStringLiteral("hyperv");
+
+    if (val == QStringLiteral("vbox"))
```
ricab, line 61: Small detail that later occurred to me:

```suggestion
    else if (val == QStringLiteral("vbox"))
```

ricab:
Thanks @georgeliao. I think we can indeed accept that, since no ambiguity should ever come from it.

@levkropp small detail inline if you want, then we can merge. Thanks!

ricab (commented):
Oh, tests missing. We missed Codecov before. This shouldn't be too hard: look for the tests that cover the surrounding code and it should be easy to replicate/generalize.

tests/test_global_settings_handlers.cpp
```diff
@@ -267,6 +267,32 @@ TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatAcceptsValidBackend
     ASSERT_NO_THROW(handler->set(key, val));
 }
 
+TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatTransformsHyperVDriver)
+{
+    auto key = mp::driver_key, val = "hyper-v";
```
georgeliao, line 272: Two minor things here: 

First one is the multiple variables declaration in one line. I know we have other code doing this and the code here is correct, but this is not a good practice in general in terms of readability and error proneness. There is also a c++ core guideline [item](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es10-declare-one-name-only-per-declaration) for this. 

Second one is that `const` can be added to the three to make these variables read only. That makes the result code: 
```
    const auto key = mp::driver_key;
    const auto val = "hyper-v";
    const auto transformed_val = "hyperv";
```

tests/test_global_settings_handlers.cpp
```diff
@@ -267,6 +267,32 @@ TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatAcceptsValidBackend
...
+
+    EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val)));
+    inject_mock_qsettings();
+
+    ASSERT_NO_THROW(handler->set(key, val));
+}
+
+TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatTransformsVBoxDriver)
+{
+    auto key = mp::driver_key, val = "vbox";
```
georgeliao, line 285: Idem

tests/test_global_settings_handlers.cpp
```diff
@@ -267,6 +267,32 @@ TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatAcceptsValidBackend
...
+    auto transformed_val = "virtualbox";
+
+    mp::daemon::register_global_settings_handlers();
+
+    EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val)));
```
georgeliao, line 290: Idem

tests/test_global_settings_handlers.cpp
```diff
@@ -267,6 +267,32 @@ TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatAcceptsValidBackend
...
 }
 
+TEST_F(TestGlobalSettingsHandlers, daemonRegistersHandlerThatTransformsHyperVDriver)
+{
+    auto key = mp::driver_key, val = "hyper-v";
+    auto transformed_val = "hyperv";
+
+    mp::daemon::register_global_settings_handlers();
+
+    EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val)));
```
georgeliao, line 277: Another minor thing. 

Without specifying the called times, the default expected called number is at least once. Maybe we can make it more accurate by specifying times once since it is only called once. That makes

`EXPECT_CALL(*mock_qsettings, setValue(Eq(key), Eq(transformed_val))).Times(1)`

georgeliao (changes requested):
@levkropp 
Thanks for the good work. 

Functional testing was done on windows, it worked as expected. 

Both the interpreter code and unit tests look good. I only had a few minor comments on unit tests for refinement.

georgeliao (approved):

PR #3655, opened 2024-08-30
update flutter to 3.24.1 and our deps

andrei-toterman:


andrei-toterman:
Welp, this can't be right...
image
image
It only happens in the snap, not if I run the locally built GUI.
Also, the app flickers regularly...

ricab:
Is that an April's fools build?

ricab:
I gotta say, it still looks pretty good :dark_sunglasses:

andrei-toterman:
Seems to be a case of the good old RGB/BGR confusion. If I swap those color channels in the screenshot, it all looks good. I shall report it.

levkropp (approved):
Looks good to me. Builds, passes tests and GUI looks correct on Ubuntu 24.04 + KDE

ricab:
Hey @andrei-toterman , the release packages are getting signed, so I think we can go ahead with this. Could you please resolve the new conflicts?

andrei-toterman:
Superseded by #3713

PR #3663, opened 2024-09-05
change window minimumSize to 750x450, and default size based on screen resolution 

### LINKED ISSUE #3590: GUI opens larger than screen on lower resolutions
**Describe the bug**

out-of-bounds

**Expected behavior**
We should probably coerce the size to the screen maximum. We may also want to save the last size?

levkropp:
close #3590 

We can use the window_size flutter plugin from google's flutter desktop embeddings repository https://github.com/google/flutter-desktop-embedding/tree/main/plugins/window_size
to check the size of the screen, and 
- if the size of the screen is below 1600x900, open a small 750x450 window
- if the size of the screen is larger then open the 1400x822 window that multipass is currently set to by default. We can also think of more reactive functionality than this but I don't think it's necessary. 

We'll need to add this plugin to pubspec.yaml, but it's ~an official plugin from Google~ the closest-to-official Flutter plugin that handles screen size and it seems that the only reason that this plugin needs to be added at all is because Google has not yet added this functionality to the Flutter framework itself

### DIFF
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -3,6 +3,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:hotkey_manager/hotkey_manager.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:window_manager/window_manager.dart';
+import 'package:window_size/window_size.dart';
 
 import 'before_quit_dialog.dart';
 import 'catalogue/catalogue.dart';
@@ -21,15 +22,24 @@ import 'vm_table/vm_table_screen.dart';
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  await setupLogger();
+  // Get the current screen size
+  final screen = await getCurrentScreen();
+  Size windowSize;
 
-  await windowManager.ensureInitialized();
-  const windowOptions = WindowOptions(
+  if (screen != null && screen.frame.size.width >= 1600
+      && screen.frame.size.height >= 900) {
+    windowSize = const Size(1400, 822);  // For screens larger than 1600x900
+  } else {
+    windowSize = const Size(750, 450);   // Default window size
+  }
+
+  final windowOptions = WindowOptions(
     center: true,
-    minimumSize: Size(1000, 600),
-    size: Size(1400, 822),
+    minimumSize: const Size(750, 450),
+    size: windowSize,
     title: 'Multipass',
   );
+
   await windowManager.waitUntilReadyToShow(windowOptions, () async {
     await windowManager.show();
     await windowManager.focus();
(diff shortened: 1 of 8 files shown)


src/client/gui/lib/main.dart
```diff
@@ -21,15 +22,24 @@ import 'vm_table/vm_table_screen.dart';
...
 
-  await setupLogger();
+  // Get the current screen size
+  final screen = await getCurrentScreen();
+  Size windowSize;
 
-  await windowManager.ensureInitialized();
-  const windowOptions = WindowOptions(
+  if (screen != null && screen.frame.size.width > 1600
+      && screen.frame.size.height > 900) {
```
ricab, line 30: Resolutions tend to stick to round numbers, so maybe use `>=`.

ricab (approved):
LGTM, modulo the inline suggestion.

src/client/gui/lib/main.dart
```diff
@@ -21,15 +22,24 @@ import 'vm_table/vm_table_screen.dart';
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  await setupLogger();
+  // Get the current screen size
+  final screen = await getCurrentScreen();
```
andrei-toterman, line 26: Why is this removing the logger setup?

Also, since we only care about the screen size, not the entire screen object, you could do something like
```dart
final screenSize = await getCurrentScreen().then((screen) {
    return screen?.frame.size;
});
```
so that you don't have to do the `.frame.size` thing multiple times

src/client/gui/lib/main.dart
```diff
@@ -21,15 +22,24 @@ import 'vm_table/vm_table_screen.dart';
...
+  Size windowSize;
 
-  await windowManager.ensureInitialized();
-  const windowOptions = WindowOptions(
+  if (screen != null && screen.frame.size.width >= 1600
+      && screen.frame.size.height >= 900) {
+    windowSize = const Size(1400, 822);  // For screens larger than 1600x900
+  } else {
+    windowSize = const Size(750, 450);   // Default window size
+  }
```
andrei-toterman, line 34: Why is this removing the initialization of the window manager?

Also, you could make `windowSize` final and initialize it with a ternary operator instead of the `if-else`.

andrei-toterman (changes requested):

andrei-toterman (changes requested):
Hey, @levkropp! It looks much better now. But I noticed something is missing. You added a plugin, which also modified the plugin registry files for each platform. And those changes must also be committed.

So you'll have to do something like this in the `src/client/gui` directory:
```
flutter config --enable-macos-desktop --enable-windows-desktop
flutter pub get
```
This should modify some files inside the `linux`, `macos` and `windows` directories related to registered plugins. Please commit those as well.

If you don't already have flutter in your PATH, you can add the `multipass/3rd-party/flutter/bin` directory to your path. (And by this I mean of course the entire path of that directory)

andrei-toterman (approved):
Alright, LGTM. Thanks, @levkropp!

PR #3671, opened 2024-09-12
call windowManager.show() and windowManager.restore() together as an extension

### LINKED ISSUE #3667: [GUI] Tray icon doesn't bring the window up if minimized or hidden behind another window
**Describe the bug**

When the GUI is running but the windows is hidden behind other windows, or minimized, clicking the tray icon does not bring it up front.

**Additional info**
 - OS: Linux
- `multipass version`: 1.14

levkropp:
resolve #3667 

`windowManager.restore()` calls gtk_window_deiconify as well as gtk_window_present https://docs.gtk.org/gtk3/method.Window.deiconify.html

In my testing, this has made it so "Open in Multipass" for an instance either
- opens the Multipass window (if keyboard focus is not stolen), or 
- shows the " is ready" dialog if typing focus is on another app 

100% of the time, and fixes the issues seen from #3667 without needing to fork the flutter window_manager plugin

### DIFF
--- a/src/client/gui/lib/tray_menu.dart
+++ b/src/client/gui/lib/tray_menu.dart
@@ -19,6 +19,13 @@ import 'vm_action.dart';
 import 'vm_details/terminal_tabs.dart';
 import 'vm_details/vm_details.dart';
 
+extension WindowManagerExtensions on WindowManager {
+  Future<void> showAndRestore() async {
+    await show();
+    await restore();
+  }
+}
+
 final trayMenuDataProvider = Provider.autoDispose((ref) {
   return ref.watch(daemonAvailableProvider)
       ? ref.watch(vmStatusesProvider)
@@ -57,7 +64,7 @@ Future<void> setupTrayMenu(ProviderContainer providerContainer) async {
       label: 'Toggle window',
       callback: (_, __) async => await windowManager.isVisible()
           ? windowManager.hide()
-          : windowManager.show(),
+          : windowManager.showAndRestore(),
     );
   }
 
@@ -207,7 +214,7 @@ Future<void> _updateTrayMenu(
           if (providerContainer.exists(provider)) {
             providerContainer.read(provider.notifier).start();
           }
-          windowManager.show();
+          windowManager.showAndRestore();
         },
       );
     } else {
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -129,7 +129,7 @@ class _AppState extends ConsumerState<App> with WindowListener {
     final primary = ref.read(clientSettingProvider(primaryNameKey));
     if (vms.contains(primary)) {
       ref.read(sidebarKeyProvider.notifier).set('vm-$primary');
-      windowManager.show();
+      windowManager.showAndRestore();
     }
   }
 


andrei-toterman (changes requested):
Good that you found this method, @levkropp! But there are multiple places where we call `show`, and I think those might need to be replaced by `restore` as well :)

andrei-toterman:
@levkropp check `main.dart` again ;)

andrei-toterman (changes requested):
Unfortunately, replacing `show` with `restore` breaks things on Windows, since the window no longer appears at all after being hidden. Looking at the `window_manager` source code, it seems that the Windows implementation of `restore` does not call some of the functions called by `show`. So there are several options. We could call `show` and `restore` together all the timer. We could fork the plugin and make it do what we need in one function call. Or we could introduce a platform-specific method that calls the appropriate `window_manager` methods when needed i.e. `show` for Windows and `restore` for the rest. Opinions, @levkropp and @ricab?

ricab:
Do you foresee any issues with just calling show + restore? Would it perhaps make things slow? If not, I guess that would be an easy fix. Otherwise, if we find that there is something wrong with the `window_manager` implementation that should be fixed, I would say fork+upstream. If it is not something wrong overall but just a specific need of ours, perhaps a platform method.

andrei-toterman:
No, I don't think there would be any issues with calling both all the time, and that would indeed be an easy fix. Just wanted to hear other opinions :)

andrei-toterman (changes requested):
@levkropp I think we first need `show` and then `restore`. On Windows, `restore` has no effect until `show` is called (which might not be a problem, but better safe than sorry). Also, you could group those two method calls into a single function, so you don't have to actually call them together all the time.
You could look into creating an extension method for the `WindowManager` class called `showAndRestore` or something like that. Make sure the method is async and awaits both `show` and `restore` ;)
You can place that extension in the tray menu file, since that's the most relevant place I can think of...

levkropp:
> You can place that extension in the tray menu file, since that's the most relevant place I can think of...

Thank you for the feedback @andrei-toterman ! I like the idea of making an extension for window_manager to handle this as well

If we place the extension in the tray_menu file then I believe we'll have to import tray_menu.dart to access it if we want to use it for goToPrimary in main.dart

We could make a new .dart file to hold the extension, or perhaps not import the extension at all for the main.dart case since goToPrimary is the only place where restore() is needed

andrei-toterman:
@levkropp we already import `tray_menu.dart` into `main.dart` since that's where we call `setupTrayMenu` which creates the tray menu itself, so that would not be an issue. But now that I look again, we have an `extensions.dart` file where there are various other extensions that couldn't find a better place. If you put it there, then we'd have to import that into `main.dart`, but that is also perfectly ok. Take your pick!

src/client/gui/lib/tray_menu.dart
```diff
@@ -55,9 +62,7 @@ Future<void> setupTrayMenu(ProviderContainer providerContainer) async {
     await TrayMenu.instance.addLabel(
       'toggle-window',
       label: 'Toggle window',
-      callback: (_, __) async => await windowManager.isVisible()
-          ? windowManager.hide()
-          : windowManager.show(),
+      callback: (_, __) async => await windowManager.showAndRestore(),
```
andrei-toterman, line 65: What is the purpose of this change? 'Toggle window' is supposed to hide the window if it's visible and show it if it's not.

PR #3677, opened 2024-09-16
add `curl` to build dependencies in debian/control

levkropp:
`curl` is required by `vcpkg` to function correctly, and without it installed on the Linux system, vcpkg fails at the `cmake ../` step. curl is not installed by default on e.g. a fresh Ubuntu 22.04 minimal installation.


andrei-toterman (approved):
Yup, seems legit. Thanks @levkropp!

PR #3681, opened 2024-09-18
small qemu improvements

sharder996:
Small cleanup:
- Moved common stub monitor to `test_qemu_backend` constructor
- Some tests were using a mock object but were not mocking function calls, so replaced with stub object
- Removed a couple instance variables from the qemu virtual machine class as they are simply copies of members from the already saved `VirtualMachineDescription`


ricab (approved):
LGTM, thanks!

Feel free to merge if/when it passes CI.

PR #3684, opened 2024-09-18
Fix `stop` succeeding when instance isn't stopped

### LINKED ISSUE #2907: `stop` command succeeds even when the instance isn't stopped
**Describe the bug**
Stop does not report failures to stop an instance and exits with 0.

**To Reproduce**
1. `multipass shell`
2. inside the instance:
    1. `sudo fsfreeze --freeze /`
    2. `sudo ip link set ens3 down`
3. back outside, `multipass stop`
4. Observe the command exits with success and without reporting any error, even though the instance is still running.

**Expected behavior**
The `stop` command would report the failure to stop the instance and exit with an error code.

**Additional info**
 - OS: Linux
- `multipass version` 1.12.0-dev.123+gdc4cbef6
- `multipass get local.driver` QEMU

**Additional context**
Related to #2906

Sploder12:
resolves #2907 

When running `stop` on an instance that could not be stopped no error would be reported. This was because `QemuVirtualMachine::shutdown(force=false)` ignored the result of `vm_process->wait_for_finished(shutdown_timeout)` which indicates if a timeout occurred or not. This issue does not seem to be present in other VM backends.

This PR makes it so the timeout is not ignored and the timeout is properly reported.

### DIFF
--- a/tests/qemu/test_qemu_backend.cpp
+++ b/tests/qemu/test_qemu_backend.cpp
@@ -302,6 +302,9 @@ TEST_F(QemuBackend, throws_when_shutdown_while_starting)
 
 TEST_F(QemuBackend, throws_on_shutdown_timeout)
 {
+    static const std::string sub_error_msg1{"The QEMU process did not finish within "};
+    static const std::string sub_error_msg2{"seconds after being shutdown"};
+
     mpt::MockProcess* vmproc = nullptr;
     process_factory->register_callback([&vmproc](mpt::MockProcess* process) {
         if (process->program().startsWith("qemu-system-") &&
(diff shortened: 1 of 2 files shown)


ricab:
Hey @Sploder12, thanks! Would you mind rebasing? Although there doesn't seem to be a conflict, GH still shows a diff to the old version here:

Screenshot from 2024-09-19 12-40-51

tests/qemu/test_qemu_backend.cpp
```diff
@@ -300,6 +300,38 @@ TEST_F(QemuBackend, throws_when_shutdown_while_starting)
...
+
+    machine->start();
+
+    ASSERT_TRUE(vmproc);
+    EXPECT_CALL(*vmproc, wait_for_finished).WillOnce(Return(false)).WillRepeatedly(Return(true));
+    EXPECT_CALL(*vmproc, running).WillOnce(Return(true)).WillRepeatedly(Return(false));
+
+    machine->state = mp::VirtualMachine::State::running;
+
+    EXPECT_THROW(machine->shutdown(), std::runtime_error);
```
andrei-toterman, line 331: You could look into our `MP_EXPECT_THROW_THAT` macro, which we use to add a matcher for the exception message, so that it's more clear about what you actually expect, not just any random `std::runtime_error`.

andrei-toterman (approved):
Alright, code looks good and now the `stop` command fails as it should. Thanks, @Sploder12!

Issue #3672, opened 2024-09-12
New Multipass version not reported in `start` and `restart`


sharder996:
**Describe the bug**
`multipass restart` and `multipass start` do not report a new Multipass version. The `update_info` field is present in both replies, and seemingly gets populated by the daemon, but is not shown in by the CLI.

**To Reproduce**
`multipass start foo`
OR
`mutlipass restart foo`

**Expected behavior**
Need to determine if the update notification should be appended to the output of these commands. If not, some code can be removed.

sharder996:
Fixed by #3686

PR #3686, opened 2024-09-19
Fix `restart` not showing new version

Sploder12:
Partial fix for #3672, unable to reproduce `start` behavior. 

When running `restart` with a new version of Multipass available, no notification would be shown but the field is present in the gRPC reply.

This PR makes it so the daemon properly populates that field for `restart` and has the CLI respond to it.

### DIFF
--- a/src/client/cli/cmd/restart.cpp
+++ b/src/client/cli/cmd/restart.cpp
@@ -39,9 +39,14 @@ mp::ReturnCode cmd::Restart::run(mp::ArgParser* parser)
     if (ret != ParseCode::Ok)
         return parser->returnCodeFrom(ret);
 
-    auto on_success = [](mp::RestartReply& reply) { return ReturnCode::Ok; };
-
     AnimatedSpinner spinner{cout};
+    auto on_success = [this, &spinner](mp::RestartReply& reply) {
+        spinner.stop();
+        if (term->is_live() && update_available(reply.update_info()))
+            cout << update_notice(reply.update_info());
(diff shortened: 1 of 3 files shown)


sharder996 (approved):
This looks good to me as is and as such I'm going to approve. 

However, I think some of the logic can be improved. 
- `write_reply` is unnecessary since writing an empty reply is idempotent
- The `config->update_prompt->is_time_to_show()` check and `config->update_prompt->populate()` can be simplified with a single call to `populate_if_time_to_show()`. `populate()` displays the new version info every time the command is called, which is probably a bit too spammy. `populate_if_time_to_show()` will only display new version info every `notify_user_frequency` which is inline with what the other commands do.

ricab:
Hmm, I would rather avoid being too naggy. @Sploder12 WDYT?

Sploder12:
> Hmm, I would rather avoid being too naggy. @Sploder12 WDYT?

I agree, I'd want to avoid that too @ricab.

ricab:
Alright, let's wait for it then, thanks.

sharder996 (approved):
LGTM! Will wait for CI before merging.

BTW, if you use one of these [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) along with the issue number, the issue will be closed automatically when the PR gets merged.

sharder996:
Fixes #3672

PR #3689, opened 2024-09-20
Increase robustness of driver factory tests

Sploder12:
Previously the following tests could fail when a user did not have permissions for ip related processes
- `PlatformLinux.test_default_driver_produces_correct_factory`
- `PlatformLinux.test_explicit_qemu_driver_produces_correct_factory`
- `PlatformLinux.test_libvirt_in_env_var_is_ignored`

Since these tests are not meant to fail for this case, these failures are unwanted. To resolve the issue, mocking the DNSMasqServerFactory allows the tests to focus more on testing the driver factories.

### DIFF
--- a/tests/linux/test_platform_linux.cpp
+++ b/tests/linux/test_platform_linux.cpp
@@ -71,13 +71,8 @@ struct PlatformLinux : public mpt::TestWithMockedBinPath
     template <typename VMFactoryType>
     void aux_test_driver_factory(const QString& driver)
     {
-        mpt::MockDNSMasqServerFactory::GuardedMock dnsmasq_server_factory_attr{
+        const mpt::MockDNSMasqServerFactory::GuardedMock dnsmasq_server_factory_attr{
             mpt::MockDNSMasqServerFactory::inject<NiceMock>()};
-        mpt::MockDNSMasqServerFactory* mock_dnsmasq_server_factory = dnsmasq_server_factory_attr.first;
-
-        ON_CALL(*mock_dnsmasq_server_factory, make_dnsmasq_server).WillByDefault([this](auto...) {
-            return std::make_unique<mpt::MockDNSMasqServer>();
-        });
 
         auto factory = mpt::MockProcessFactory::Inject();
         setup_driver_settings(driver);
--- a/tests/linux/test_platform_linux.cpp
+++ b/tests/linux/test_platform_linux.cpp
@@ -24,6 +24,7 @@
 #include "tests/mock_settings.h"
 #include "tests/mock_standard_paths.h"
 #include "tests/mock_utils.h"
+#include "tests/qemu/linux/mock_dnsmasq_server.h"
 #include "tests/temp_dir.h"
 #include "tests/test_with_mocked_bin_path.h"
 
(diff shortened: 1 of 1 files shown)


tests/linux/test_platform_linux.cpp
```diff
@@ -70,6 +71,14 @@ struct PlatformLinux : public mpt::TestWithMockedBinPath
     template <typename VMFactoryType>
     void aux_test_driver_factory(const QString& driver)
     {
+        mpt::MockDNSMasqServerFactory::GuardedMock dnsmasq_server_factory_attr{
```
georgeliao, line 74: Very minor thing, maybe prepend `const`.

tests/linux/test_platform_linux.cpp
```diff
@@ -70,6 +71,14 @@ struct PlatformLinux : public mpt::TestWithMockedBinPath
...
+            mpt::MockDNSMasqServerFactory::inject<NiceMock>()};
+        mpt::MockDNSMasqServerFactory* mock_dnsmasq_server_factory = dnsmasq_server_factory_attr.first;
+
+        ON_CALL(*mock_dnsmasq_server_factory, make_dnsmasq_server).WillByDefault([this](auto...) {
+            return std::make_unique<mpt::MockDNSMasqServer>();
+        });
+
```
georgeliao, line 81:
```
        mpt::MockDNSMasqServerFactory* mock_dnsmasq_server_factory = dnsmasq_server_factory_attr.first;

        ON_CALL(*mock_dnsmasq_server_factory, make_dnsmasq_server).WillByDefault([this](auto...) {
            return std::make_unique<mpt::MockDNSMasqServer>();
        });
```
I do not think this part is needed, basically `QemuVirtualMachineFactory` object has `qemu_platform`, and `qemu_platform_details` which is the instance of `qemu_platform` on linux has `DNSMasqServer` pointer. If we just mock the `DNSMasqServerFactory` and do nothing, then this [line](https://github.com/canonical/multipass/blob/f11f181fd7d8b46ec3b39bd94c9ffd3aa14ebc9a/src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp#L108) call returns the default value of the return type which is `nulltpr`. The code above makes it returns an pointer of `mpt::MockDNSMasqServer`, but the unit tests in this file did not use any of his mocked behavior. As a matter of fact, only the `QemuVirtualMachineFactory` constructor gets called, so it is ok to leave `DNSMasqServer` pointer empty.

georgeliao (changes requested):
Hi @Sploder12 
Thanks for the good work, mocking the `DNSMasqServerFactory` is defintely the right way to isolate the unit test and circumvent any dependency related issues like the permissions you mentioned. I just have some minor comments to refine the code.

georgeliao:
@Sploder12 
Another thing is the functional testing. 
Is there a simple way to set the permission so the run will fail on the original implementation?

Sploder12:
> @Sploder12 Another thing is the functional testing. Is there a simple way to set the permission so the run will fail on the original implementation?

Changing your user permissions for `/proc/sys/net/ipv4/ip_forward` to read only should work.

georgeliao:
> Changing your user permissions for /proc/sys/net/ipv4/ip_forward to read only should work.

It is a root owned file and it is ready only for other users already. But my unit test was able to run on the original code. I only got a warning about can not open that file, but the unit test did not fail. 
```
[       OK ] PlatformLinux.test_default_driver_produces_correct_factory (7 ms)
[ RUN      ] PlatformLinux.test_explicit_qemu_driver_produces_correct_factory
[warning] [apparmor] AppArmor disabled by environment variable
[warning] [qemu platform] Unable to open /proc/sys/net/ipv4/ip_forward
[debug] [dnsmasq] Starting dnsmasq
[info] [firewall] Using iptables-nft for firewall rules.
[debug] [dnsmasq] terminating
[info] [dnsmasq] failed to terminate nicely, killing
[warning] [dnsmasq] failed to kill
```

georgeliao (approved):

PR #3695, opened 2024-09-24
[daemon] let daemon choose mount target if not specified

andrei-toterman:
This PR adds the functionality for the daemon to choose a mount target path if it is not given. Right now the logic is to take the last directory name in the source path and append it to `/home/ubuntu` e.g. given the source path `/home/andrei/projects/multipass`, the default target path would be `/home/ubuntu/multipass`

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -235,6 +235,8 @@ public:
     virtual QString make_uuid(const std::optional<std::string>& seed = std::nullopt) const;
     virtual void sleep_for(const std::chrono::milliseconds& ms) const;
     virtual bool is_ipv4_valid(const std::string& ipv4) const;
+
+    virtual Path default_mount_target(const Path& source) const;
 };
 } // namespace multipass
 
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -582,6 +582,11 @@ bool mp::Utils::is_ipv4_valid(const std::string& ipv4) const
     return true;
 }
 
+mp::Path mp::Utils::default_mount_target(const Path& source) const
+{
+    return source.isEmpty() ? "" : QDir{QDir::cleanPath(source)}.dirName().prepend("/home/ubuntu/");
+}
+
 auto mp::utils::find_bridge_with(const std::vector<mp::NetworkInterfaceInfo>& networks,
                                  const std::string& target_network,
                                  const std::string& bridge_type) -> std::optional<mp::NetworkInterfaceInfo>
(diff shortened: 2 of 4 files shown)


include/multipass/utils.h
```diff
@@ -235,6 +235,8 @@ class Utils : public Singleton<Utils>
     virtual QString make_uuid(const std::optional<std::string>& seed = std::nullopt) const;
     virtual void sleep_for(const std::chrono::milliseconds& ms) const;
     virtual bool is_ipv4_valid(const std::string& ipv4) const;
+
+    virtual Path default_mount_target_from_source(const Path& source) const;
```
ricab, line 239: Could we have a shorter name, please? C++ is verbose enough as it is, and source is already mentioned in the parameter, so perhaps just:

```suggestion
    virtual Path default_mount_target(const Path& source) const;
```

src/utils/utils.cpp
```diff
@@ -582,6 +582,11 @@ bool mp::Utils::is_ipv4_valid(const std::string& ipv4) const
     return true;
 }
 
+mp::Path mp::Utils::default_mount_target_from_source(const Path& source) const
+{
+    return source.isEmpty() ? "" : QDir{QDir::cleanPath(source)}.dirName().prepend("/home/ubuntu/");
```
ricab, line 587: Shouldn't we throw if this is empty? You probably want to do that in this function's client (daemon) and make it a pre-condition here that the directory isn't empty, perhaps with an assert.

ricab (commented):
Thanks Andrei, just a quick review pass.

src/utils/utils.cpp
andrei-toterman: There is a purpose for returning an empty path if given an empty path, which is the GUI target preview. If the source field in the GUI is left empty, the target should also be empty. So this just simplifies that process. But I understand that taken in isolation, the better behavior for this method is to error if the source is empty.

include/multipass/utils.h
andrei-toterman: Sure, makes sense. I'll change that everywhere.

src/utils/utils.cpp
ricab: Hmm, well that could be done in the GUI itself, no? Anyway, I am OK keeping it like this as long as the daemon itself handles the case.

ricab (approved):
Thanks @andrei-toterman! Let's wait for a primary reviewer, but LGTM on secondary review.

Sploder12 (approved):
LGTM! `default_mount_target` returning `""` is handled by the `invalid_target_path` check a little later so that edge case is properly handled by the daemon.

ricab (approved):
Alright, this is going in, thanks!

PR #3696, opened 2024-09-24
[cli/mount] remove default target if not specified

andrei-toterman:
This PR simply removes the responsibility of the CLI to choose a mount target path if it is not given, since that will now be the responsibility of the daemon. Thus, the CLI can now pass an empty target path to the daemon.


Sploder12 (approved):
LGTM

ricab (approved):
And so is this...

ricab:
... as soon as the base finishes.

ricab (approved):

PR #3697, opened 2024-09-24
[gui] add ffi for getting mount target from source

andrei-toterman:
This PR makes the logic of computing a mount target path from a source path available in the FFI library that we use from Dart. This is so that the GUI can show a preview of what a target path will be before actually sending the request to the daemon.

### DIFF
--- a/include/multipass/dart_ffi.h
+++ b/include/multipass/dart_ffi.h
@@ -40,6 +40,8 @@ int gid();
 int default_id();
 
 long long memory_in_bytes(char* value);
+
+char* default_mount_target(char* source);
 }
 
 #endif // MULTIPASS_DART_FFI_H
--- a/include/multipass/dart_ffi.h
+++ b/include/multipass/dart_ffi.h
@@ -7,9 +7,9 @@ extern "C"
 // clang-format on
 const char* multipass_version();
 
-const char* generate_petname();
+char* generate_petname();
 
-const char* get_server_address();
+char* get_server_address();
 
 struct KeyCertificatePair
 {
@@ -27,11 +27,11 @@ enum SettingResult
     UnexpectedError,
 };
 
-const char* settings_file();
+char* settings_file();
 
-enum SettingResult get_setting(const char* key, const char** output);
+enum SettingResult get_setting(char* key, char** output);
 
-enum SettingResult set_setting(const char* key, const char* value, const char** output);
+enum SettingResult set_setting(char* key, char* value, char** output);
 
 int uid();
 
@@ -39,7 +39,7 @@ int gid();
 
 int default_id();
 
-long long memory_in_bytes(const char* value);
+long long memory_in_bytes(char* value);
 }
 
 #endif // MULTIPASS_DART_FFI_H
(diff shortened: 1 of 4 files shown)


levkropp (approved):
LGTM. Builds & tests pass!

src/client/gui/ffi/dart_ffi.cpp
```diff
@@ -214,4 +215,23 @@ catch (...)
     mpl::log(mpl::Level::warning, category, "failed converting memory to bytes");
     return -1;
 }
+
+const char* default_mount_target(const char* source)
+try
+{
+    const QString q_source{source};
+    free((void*)source);
```
ricab, line 223: I am a little surprised that this takes ownership of the parameter and deletes memory that is supposed to hold const chars. I suppose the intention is that the client doesn't have to free after each function call, but in that case, don't you think we should remove the `const`? As it stands, it is easy to get into UB (e.g. call with a compile-time constant).

I realize this pattern is repeated all over this file, but it should be straightforward to modify everywhere (provided it works).

src/client/gui/ffi/dart_ffi.cpp
```diff
@@ -214,4 +215,23 @@ catch (...)
...
+    return strdup(target.c_str());
+}
+catch (const std::exception& e)
+{
+    mpl::log(mpl::Level::warning, category, fmt::format("failed retrieving default mount target: {}", e.what()));
+    return nullptr;
+}
+catch (...)
+{
+    mpl::log(mpl::Level::warning, category, "failed retrieving default mount target");
```
ricab, line 234: Could you please factor out the error message? Can't use function try any longer, but that should be fine. You can have a `static constexpr char *` at the beginning.

include/multipass/dart_ffi.h
```diff
@@ -40,6 +40,8 @@ int gid();
 int default_id();
 
 long long memory_in_bytes(const char* value);
+
+const char* default_mount_target(const char* source);
```
ricab, line 44: Here too. This transfers ownership, so the `const` doesn't look right.

ricab (commented):
Hey @andrei-toterman, thanks for modularising it in this way, it is great that we don't need to repeat the logic. I have a couple of matters for discussion below.

src/client/gui/ffi/dart_ffi.cpp
andrei-toterman: That is a very good point, thanks. That is indeed a critical mistake. I'll fix it everywhere ASAP. I'll make a new PR for fixing those issues and set that PR as the base for this one. Is that ok?

src/client/gui/ffi/dart_ffi.cpp
andrei-toterman: Yes, sure. It just sucks that it adds another level of indentation for the whole body of the function...

src/client/gui/ffi/dart_ffi.cpp
ricab: Sure, that sounds perfectly fine :smiley:

ricab (dismissed):
LGTM, thanks!

ricab (approved):
Merging this to main directly to speed things up.

PR #3698, opened 2024-09-24
Mount redesign

andrei-toterman:
This PR implements the redesign of the mounting interface, as well as some cosmetic changes to the pages that use the mounting interface, namely the launch panel and the individual instance page.

See https://warthogs.atlassian.net/browse/MULTI-894 and https://warthogs.atlassian.net/browse/MULTI-895 for the details of the changes.

There was also some refactoring of adjacent code that is not necessarily mount-related (e.g. splitting some files into multiple files), so that the modification and addition of the mount code is easier


src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -42,9 +59,9 @@ class LaunchForm extends ConsumerWidget {
     );
 
     final nameInput = SpecInput(
+      label: 'Name',
       autofocus: true,
-      helper:
-          'Names are immutable. If no names are provided, Multipass will randomly generate a name.',
+      helper: 'Namess cannot be changed one an instance is created',
```
levkropp, line 64: Typo in "Names" and "once"

levkropp (changes requested):
Well done!! The mount section looks good and is intuitive to use. I also liked how you turned ConfirmationDialog into a reusable widget 😃

levkropp (dismissed):

ricab (changes requested):
Hi Andrei,

Thanks, looking very nice. Most of what I tried worked as it should, but I did find a few things to look into: 

- Default mount path points to the inside of the snap, whereas design specifies it should start at the home directory
Screenshot from 2024-10-22 11-29-49
- Placeholder/square char in the dialogue to delete a mount 
Screenshot from 2024-10-22 11-35-04
- After deleting the last couple of mounts, the "cancel" button lingers and "add mount" is no longer present - it should perhaps go back to the initial state, with "add mount" on the right
Screenshot from 2024-10-22 11-41-00
- very long paths are not "truncated in the middle" when in saved state 
Screenshot from 2024-10-22 12-00-18

To discuss with Design:
- Save being disabled when new mount empty - not sure what disabled state means
- file chooser not opening in Home - opening in recent instead
    - if a path is already filled in, should it start from there?
- when I delete a mount, leaving others behind, it is a little weird to see a cancel button, since the action cannot be reverted any longer. Should it say "Save" instead? But save without cancel is weird. Should it just go back to "saved mount state", as when adding a mount? Or maybe the text should be changed to "Done"... On the same topic, if I add a mount, then select "cancel", the mount remains.

andrei-toterman:
Regarding the 3rd point with the 'add mount' button not being present, I don't think it should appear on the right, but on the left where it appears when there are mounts present. 'Add mount' is on the right when we are not in the 'Configure' mode of the form.

And I agree that 'Cancel' is a mid misleading since it doesn't really cancel stuff. What about just 'Close' for now?

ricab:
Only that last permissions thing missing, otherwise looking good.

PR #3709, opened 2024-10-01
Reduce spurious CI test failures

Sploder12:
Previously CI could fail while writing to JSON files, especially in snapshot tests. Now, snapshot tests and many daemon tests minimally rely on the filesystem so the chance of spurious filesystem failures should be reduced.

MULTI-1230

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -138,6 +138,9 @@ struct Daemon : public mpt::DaemonTestFixture
 
     mpt::MockSettings::GuardedMock mock_settings_injection = mpt::MockSettings::inject<StrictMock>();
     mpt::MockSettings& mock_settings = *mock_settings_injection.first;
+
+    const mpt::MockJsonUtils::GuardedMock mock_json_utils_injection = mpt::MockJsonUtils::inject<NiceMock>();
+    mpt::MockJsonUtils& mock_json_utils = *mock_json_utils_injection.first;
 };
 
 TEST_F(Daemon, receives_commands_and_calls_corresponding_slot)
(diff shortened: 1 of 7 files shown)


andrei-toterman (approved):
Thanks, @Sploder12! Looks good to me!

tests/test_daemon.cpp
```diff
@@ -375,6 +378,18 @@ struct DaemonCreateLaunchAliasTestSuite : public Daemon, public FakeAliasConfig,
     {
         EXPECT_CALL(mpt::MockStandardPaths::mock_instance(), writableLocation(_))
             .WillRepeatedly(Return(fake_alias_dir.path()));
+
+        EXPECT_CALL(mock_json_utils, write_json(_, _)).WillRepeatedly([](auto&& a, auto&& b) {
+            // this try catch is necessary since propagating the error hangs the test.
+            try
+            {
+                MP_JSONUTILS.JsonUtils::write_json(a, b);
```
ricab, line 386: could use perfect forwarding here:

```suggestion
                MP_JSONUTILS.JsonUtils::write_json(std::forward<decltype(a)>(a), std::forward<decltype(b)>(b));
```

tests/test_daemon.cpp
```diff
@@ -375,6 +378,18 @@ struct DaemonCreateLaunchAliasTestSuite : public Daemon, public FakeAliasConfig,
     {
         EXPECT_CALL(mpt::MockStandardPaths::mock_instance(), writableLocation(_))
             .WillRepeatedly(Return(fake_alias_dir.path()));
+
+        EXPECT_CALL(mock_json_utils, write_json(_, _)).WillRepeatedly([](auto&& a, auto&& b) {
+            // this try catch is necessary since propagating the error hangs the test.
```
ricab, line 383: Hmm interesting... Did you check if that is a generic problem with the test setup or even with the daemon? 

For my understanding: why does throwing from here hang the test? How is that different from other throws?

ricab (commented):
Hey Trevor, thank you for this, it looks great! A couple of things inline...

ricab (approved):

PR #3713, opened 2024-10-03
Update flutter 3.24.3

andrei-toterman:


ricab:
Hi @levkropp, we would need just a sanity check review and confirm this is the same as the old PR before we can merge.

levkropp (approved):
LGTM! Builds, tests pass and GUI is working

PR #3715, opened 2024-10-04
Change file permissions on instances

### LINKED ISSUE #3866: Multipass service does not start properly on Windows (Installed in French language!)
**Describe the bug**
After Multipass setup, after first restart, if I try any command I get the error "cannot connect to the multipass socket" and if I check the status of the "Multipass service" it's turned off.

(Please see below Additional context for my own investigation about that issue and a way to circumvent it !) 

**To Reproduce**
How, and what happened?
1. Install Multipass on Windows (OS installed in French language!)
2. Restart as expected by the setup process
3. `multipass info`

**Expected behavior**

Multipass commands should run correctly after setup and first reboot (even when the OS is installed in French)

**Logs**
Please provide logs from the daemon, see [accessing logs](https://canonical.com/multipass/docs/accessing-logs) on where to find them on your platform.

[powershell.exe] [8556] started: powershell.exe takeown /a /r /d Y /f "C:\ProgramData\Multipass"

[chown] [8556] PowerShell finished abnormally: Process returned exit code: 1

[chown] [21032] stderr: Erreurï¿½: Syntaxe incorrecte. La valeur ï¿½ï¿½Yï¿½ï¿½ n'est pas autorisï¿½e pour l'option ï¿½ï¿½/dï¿½ï¿½.
Entrez "TAKEOWN /?" pour afficher la syntaxe.

**Additional info**
 - OS: Windows 11 pro (installed in French)
- `multipass version`  1.15.0+win
- `multipass info`
- `multipass get local.driver` hyperv

**Additional context**

When I tried to execute "takeown /?" I've found that in French the "/d" option expects "O" or "N" values (instead of "Y" or "N" in english)
So I executed the command myself : takeown /a /r /d O /f "C:\ProgramData\Multipass"
It ran successfully and after I was able to start the Multipass service without issues ! And of course after that I got no errors when trying some multipass commands !

Sploder12:
This PR changes file permissions on Multipass cache and data directories to be read/write only by root. Previously the files could be read by all.

Fixes #3866 

MULTI-1403
MULTI-1723


levkropp (changes requested):
Looks good, it seems that codecov wants to see test coverage for some lines:
  -  if (!MP_PLATFORM.set_root_as_owner(path)) in `write_to`
  - `extract_image` doesn't seem to have tests for it at all?

Sploder12:
I have decided that the vm_image_vault tests are out of the scope of this PR. This is because writing tests would require a significant refactor of unrelated code (see 1653 in Jira).

sharder996 (changes requested):
Thanks for your work on this @Sploder12! Tackling these parts of the code can be tricky :)

Since this PR is closely related with #3782, there might be some overlap. But, here are some thoughts I had reviewing this:

1. Existing instance directories/files do not have their permissions changed, only new instances.
2. Is there a reason to apply permissions specifically to `cloud_init_iso` and instance directories and not to the other files stored within the `data_directory` and `cache_directory`?
3. Would it be better/easier to apply blanket permissions on the `storage_directory`/`cache_directory`/`data_directory` in `daemon_config`? That would solve 1 if it doesn't interfere with 2.
4. Instead of the name `set_root_as_owner()`, how about `takeown()`? If need be, it could default to root/admin with an option to set ownership to the current user. That would follow pretty closely the Windows definition of `takeown`.

Sploder12:
@sharder996 

1. Good point! Updating the permissions when the daemon starts is what I'm thinking to fix that, WDYT?
2. Some files aren't as sensitive, but it might be worth covering those too, @ricab might have more thoughts on this.
3. That might work, I'll give it a try
4. I don't like `takeown` personally, reminds me too much of runic C string functions like `wcsxfrm`. But something like `take_ownership` or `set_owner` could be nice for that. WDYT?

ricab:
> Some files aren't as sensitive, but it might be worth covering those too, @ricab might have more thoughts on this.

I guess covering the whole directory would be safer for the future and agree with principle of least privilege better. We can always expose any thing we need to on a case by case basis.

src/utils/permission_utils.cpp
```diff
@@ -0,0 +1,107 @@
...
+namespace mp = multipass;
+namespace fs = mp::fs;
+
+using Path = mp::PermissionUtils::Path;
+
+namespace
+{
+void set_single_permissions(const Path& path, const QFileDevice::Permissions& permissions)
+{
+    QString qpath = QString::fromUtf8(path.u8string());
```
sharder996, line 32: There's some funny conversion going on here. `QString` to `u16String` to `u8string` to `QString`.

sharder996 (changes requested):
Initial review pass and I noticed some things that I think could be improved. Will make a more detailed review afterwards.

By setting permissions on the top level data/cache directories, all sub-directories should inherit those permissions, correct? If so, some of the code could be cleaned up.

src/daemon/daemon_config.cpp
```diff
@@ -189,6 +190,16 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
     }
 
+    if (!storage_path.isEmpty())
+    {
+        MP_PERMISSIONS.restrict_permissions(storage_path.toStdU16String());
+    }
+    else
+    {
+        MP_PERMISSIONS.restrict_permissions(data_directory.toStdU16String());
+        MP_PERMISSIONS.restrict_permissions(cache_directory.toStdU16String());
```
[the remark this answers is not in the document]
Sploder12: yep, [`std::filesystem::path`](https://en.cppreference.com/w/cpp/filesystem/path/path)'s constructor from `std::string` is assumed to be the "native narrow encoding", which is not guaranteed to be UTF-8 (I think). But the `char16_t` constructor does handle UTF-16 conversion.

src/utils/permission_utils.cpp
Sploder12: yeah, `QString` -> `u16String` -> `std::filesystem::path` -> `u8string` -> `QString` -> `std::string` -> `char*`. I could probably get it to `QString` -> `u16String` -> `std::filesystem::path` -> `std::string` -> `char*` but any better than that would require some pretty massive changes to everything.

include/multipass/utils/permission_utils.h
```diff
@@ -0,0 +1,46 @@
...
+
+#define MP_PERMISSIONS multipass::PermissionUtils::instance()
+
+namespace multipass
+{
+
+class PermissionUtils : public Singleton<PermissionUtils>
+{
+public:
+    using Path = std::filesystem::path;
```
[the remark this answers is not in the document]
Sploder12: `namespace fs = std::filesystem` is already declared elsewhere as well and has other limitations that `using` avoids. I prefer this since it's contained in the `PermissionUtils` class. Of course it'd be nice to just have one `Path` but the codebase is split between `QString` and `fs::path`.

ricab:
I meant to review this but I am a little bogged down with reviews right now. This has 2 reviewers already, so I'll leave it up to them.

sharder996 (approved):
Looks much better now. Thanks @Sploder12!

include/multipass/platform.h
```diff
@@ -58,7 +58,11 @@ class Platform : public Singleton<Platform>
     virtual bool is_remote_supported(const std::string& remote) const;
     virtual bool is_backend_supported(const QString& backend) const; // temporary (?)
     virtual int chown(const char* path, unsigned int uid, unsigned int gid) const;
-    virtual bool set_permissions(const std::filesystem::path& path, std::filesystem::perms permissions) const;
+    virtual bool set_permissions(const std::filesystem::path& path,
+                                 std::filesystem::perms permissions,
+                                 bool try_inherit = false) const;
+    virtual bool take_ownership(const std::filesystem::path& path) const;
```
ricab, line 64: I was looking into our current code for this because of a support question and it occurred to me that a `nodiscard` attribute would help ensure we handle errors appropriately.

src/daemon/daemon_config.cpp
```diff
@@ -193,14 +192,27 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
+        const std::filesystem::path storage = storage_path.toStdU16String();
+        restrict_result &= MP_PLATFORM.set_permissions(storage, std::filesystem::perms::owner_all);
+        restrict_result &= MP_PLATFORM.take_ownership(storage);
     }
     else
     {
-        MP_PERMISSIONS.restrict_permissions(data_directory.toStdU16String());
-        MP_PERMISSIONS.restrict_permissions(cache_directory.toStdU16String());
+        const std::filesystem::path data = data_directory.toStdU16String();
+        restrict_result &= MP_PLATFORM.set_permissions(data, std::filesystem::perms::owner_all);
```
georgeliao, line 205: But the permission will be relaxed on the standard grpc mTLS PR.

src/daemon/daemon_config.cpp
```diff
@@ -193,14 +192,27 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
+        restrict_result &= MP_PLATFORM.take_ownership(data);
+
+        const std::filesystem::path cache = cache_directory.toStdU16String();
```
georgeliao, line 208: Can we keep cache folder permission to recursive since it does not interfere the image file?

ricab:
@georgeliao @levkropp I believe this needs a green check mark to proceed.

levkropp (approved):

PR #3724, opened 2024-10-11
[flutter] add local_notifier plugin & show local notification on update available

levkropp:


src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +17,39 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
     final updateNotificationExists = ref.read(notificationsProvider).any((n) {
       return n is UpdateAvailableNotification && n.updateInfo == updateInfo;
     });
     if (updateNotificationExists) return;
-    ref
-        .read(notificationsProvider.notifier)
-        .add(UpdateAvailableNotification(updateInfo));
+
+    // Existing in-app notification
+    ref.read(notificationsProvider.notifier).add(UpdateAvailableNotification(updateInfo));
```
andrei-toterman, line 27: Was this re-formatted automatically? If I run `dart format` on the file, it reverts it to the way it was before.
If your IDE automatically formats code, make sure it uses `dart format` instead of its own rules.

src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +17,39 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
+    }
+
+    final notification = LocalNotification(
+      title: 'Multipass Update Available',
+      body: 'Version ${updateInfo.version} is available. Click to upgrade now.',
+    );
+
+    notification.onClick = () async {
+      await launchInstallUrl();
+    };
```
andrei-toterman, line 50: I think you can make this just `notification.onClick = launchInstallUrl;`

src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +17,39 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
+
     final updateNotificationExists = ref.read(notificationsProvider).any((n) {
       return n is UpdateAvailableNotification && n.updateInfo == updateInfo;
     });
     if (updateNotificationExists) return;
-    ref
-        .read(notificationsProvider.notifier)
-        .add(UpdateAvailableNotification(updateInfo));
+
+    // Existing in-app notification
```
andrei-toterman, line 26: I think just `In-app notification` would be enough for this comment. Specifying that it's `existing` does not clarify much.

src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +17,39 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
+    // Update the state
     state = updateInfo;
+
+    // Create and show a local notification
+    _showLocalNotification(updateInfo);
+  }
+
+  void _showLocalNotification(UpdateInfo updateInfo) {
+
+    if (Platform.isLinux) {
```
andrei-toterman, line 38: We generally try to avoid using `Platform.is<OS>` directly. Take a look at the `MpPlatform` class. You can add a bool getter there and override it in each derived class. Then you can use here the `mpPlatform` singleton to check the condition. It is a little more cumbersome, but it keeps all platform specific behavior centralized.

src/client/gui/pubspec.yaml
```diff
@@ -26,6 +26,7 @@ dependencies:
   grpc: ^4.0.1
   hotkey_manager: ^0.2.3
   intl: ^0.19.0
+  local_notifier: ^0.1.6
```
andrei-toterman, line 29: I see that this plugin requires `libnotify-dev` in order to build on Linux. Even though we don't need its functionality on Linux, it's still needed in order to build. So you'll need to add `libnotify-dev` to our `debian/control`. Surprisingly, I see that the snap builds fine, so something must be bringing in that library.

andrei-toterman (changes requested):

src/client/gui/lib/update_available.dart
```diff
@@ -1,7 +1,10 @@
+import 'dart:io';
 import 'package:basics/basics.dart';
 import 'package:flutter/material.dart';
+import 'package:local_notifier/local_notifier.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_svg/svg.dart';
+import 'package:multipass_gui/platform/platform.dart';
```
andrei-toterman, line 7: for our own files, it's preferable you use relative imports e.g. `import 'platform/platform.dart'` just so it's separated from the 3rd party stuff

src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +18,38 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
+    // Update the state
     state = updateInfo;
+
+    // Create and show a local notification
+    _showLocalNotification(updateInfo);
+  }
+
+  void _showLocalNotification(UpdateInfo updateInfo) {
+    if (!mpPlatform.showLocalUpdateNotifications) {
+      // Do not show notification on Linux
```
andrei-toterman, line 41: I think you can remove this comment now, since it is no longer the responsibility of this method to decide on which platform the notification is shown. Then you can one-line `if (...) return;`

andrei-toterman (changes requested):
Just a few small nitpicks :)

src/client/gui/lib/update_available.dart
```diff
@@ -1,7 +1,10 @@
+import 'dart:io';
```
andrei-toterman, line 1: Do you still need this import?

andrei-toterman (changes requested):

src/client/gui/lib/update_available.dart
```diff
@@ -15,14 +17,35 @@ class UpdateNotifier extends Notifier<UpdateInfo> {
...
+
+  void _showLocalNotification(UpdateInfo updateInfo) {
+    if (!mpPlatform.showLocalUpdateNotifications) return;
+
+    final notification = LocalNotification(
+      title: 'Multipass Update Available',
+      body: 'Version ${updateInfo.version} is available. Click to upgrade now.',
+    );
+
+    notification.onClick = launchInstallUrl;
```
andrei-toterman, line 46: Oof, I thought that the notification will automatically close when you click it, but it doesn't, so it must be closed manually.
```suggestion
    notification.onClick = () async {
      await launchInstallUrl();
      await notification.close();
    };
```

andrei-toterman (changes requested):

andrei-toterman (approved):
Perfect, thanks @levkropp!

PR #3725, opened 2024-10-11
[gui] guard against launching an instance with the name of a deleted instance

andrei-toterman:


ricab (dismissed):
Looks good, thanks!

ricab:
Merging together with #3698.

PR #3729, opened 2024-10-14
Vm resource sliders

andrei-toterman:
This PR introduces a new way to input VM specs using sliders.
See https://warthogs.atlassian.net/browse/MULTI-959 for reference.

### DIFF
--- a/include/multipass/memory_size.h
+++ b/include/multipass/memory_size.h
@@ -39,9 +39,12 @@ public:
     long long in_megabytes() const noexcept;
     long long in_gigabytes() const noexcept;
 
-    std::string human_readable() const;
+    std::string human_readable(unsigned int precision = 1) const;
+
+    static MemorySize from_bytes(long long bytes) noexcept;
 
 private:
+    explicit MemorySize(long long bytes) noexcept;
     long long bytes;
 };
 
--- a/include/multipass/memory_size.h
+++ b/include/multipass/memory_size.h
@@ -39,7 +39,7 @@ public:
     long long in_megabytes() const noexcept;
     long long in_gigabytes() const noexcept;
 
-    std::string human_readable(unsigned int precision = 1) const;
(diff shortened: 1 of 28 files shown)


include/multipass/memory_size.h
```diff
@@ -39,9 +39,12 @@ class MemorySize
     long long in_megabytes() const noexcept;
     long long in_gigabytes() const noexcept;
 
-    std::string human_readable() const;
+    std::string human_readable(unsigned int precision = 1) const;
+
+    static MemorySize from_bytes(long long bytes) noexcept;
```
ricab, line 44: Very good choice. The static function with this name makes it clear what the unit is, which would no be the case with a constructor :+1:

ricab (changes requested):
OK, this is looking really nice. I'll give it to you, you hit the nail with the slider non-linearity. Well done!

A few things need attention. High priority:

- middle and right slider both say "Memory"
- I get a maximum disk of 16G for some reason

Middle priority:

- most of the time, I can't input digits on the text (number) fields for mem and disk (e.g. try to erase the number and type a new one, nothing gets typed)

Nice to have:

- It would be nice if round numbers like 512.00 showed up as 512. 512.00 indicates an imprecision lower than += 0.005. 0 is only one such imprecision. I can see that the minimum I can select lines up to the byte (which indeed makes most sense), so better just say 512MiB.

Screenshot from 2024-10-22 21-41-50

andrei-toterman:
Regarding not being able to input values into the text field, can you give some examples? There are some restrictions in place i.e. you are not allowed to enter a value that falls outside the range or that cannot be parsed into a number. This behavior was again inspired by VirtualBox, but I can understand how it can be a little confusing or cumbersome, and I have some other ideas that I can experiment with.

include/multipass/memory_size.h
andrei-toterman: Just bringing some Rust habits into C++ :wink:

PR #3730, opened 2024-10-14
Fix dart ffi

andrei-toterman:
This PR removes const qualifiers in the case of ownership transfer.
It also extracts common error messages of each function to static variables.

### DIFF
--- a/include/multipass/dart_ffi.h
+++ b/include/multipass/dart_ffi.h
@@ -7,9 +7,9 @@ extern "C"
 // clang-format on
 const char* multipass_version();
 
-const char* generate_petname();
+char* generate_petname();
 
-const char* get_server_address();
+char* get_server_address();
 
 struct KeyCertificatePair
 {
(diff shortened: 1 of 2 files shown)


ricab (commented):
Much better!

I only just noticed though that almost all of those try-catches are basically `mp::top_catch_all`. If you prefer to do that later to avoid rebases, that's fine.

andrei-toterman:
Basically, they are `mp::top_catch_all`. But looking through some of the exceptions thrown by those functions, I think their messages lack some specificity. Now, all of the logs contain the exception message preceded by some other message that adds more context that is relevant for the function. If I were to use `top_catch_all`, all the logs would contain 'Caught an unhandled/unknown exception'. And that would make the logs a little too vague IMO. Of course, I could catch those exception and rethrow them with more context in their message, but that completely undermines the point of using `top_catch_all`.

ricab:
Ah, you're right. I was imagining we supported a prefix message.

ricab (approved):
Yup, thanks!

ricab:
This was merged together with #3697.

PR #3734, opened 2024-10-16
Ensure terminal doesn't render partial lines & Snap terminal scrolling to line height

levkropp:
Via a fork of the flutter xterm.dart plugin, editing ui/render.dart

### DIFF
--- a/src/client/gui/pubspec.yaml
+++ b/src/client/gui/pubspec.yaml
@@ -52,7 +52,10 @@ dependencies:
       url: https://github.com/google/flutter-desktop-embedding.git
       path: plugins/window_size
       ref: 6c66ad2
-  xterm: ^4.0.0
+  xterm:
+    git:
+      url: https://github.com/levkropp/xterm.dart
+      ref: 4.0.0+mp
 
 dependency_overrides:
   hotkey_manager_linux:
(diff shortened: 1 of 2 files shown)


andrei-toterman:
Nice, the terminals scrolls nicely and there are no partial lines. One issue though is that now the terminal does not autoscroll anymore when new lines appear. For example, if I do `sudo apt update && sudo apt upgrade` there are many new lines written to the terminal, but I am not automatically taken to the bottom of them as they appear.

andrei-toterman (approved):
It works really nicely now :smile: Thanks, @levkropp!

ricab:
Oh, I meant to secondary-review this...

src/client/gui/pubspec.yaml
```diff
@@ -52,7 +52,10 @@ dependencies:
       url: https://github.com/google/flutter-desktop-embedding.git
       path: plugins/window_size
       ref: 6c66ad2
-  xterm: ^4.0.0
+  xterm:
+    git:
+      url: https://github.com/levkropp/xterm.dart
```
ricab, line 57: This should live in the Canonical org, but I see there are similar cases here. I'll propose fixing all that on an upcoming pulse.

ricab (commented):
Late approval :heavy_check_mark:

PR #3735, opened 2024-10-16
Update libssh to 0.11.1

### LINKED ISSUE #3745: SSHFS watchdog doesn't notice thread exit
**Describe the bug**
The SSHFS server's watchdog doesn't have a way of verifying that the SFTP thread it watches is still alive. This causes the SSHFS server process to continue executing even if it is no longer functional. This causes mounted folders to hang the shell since the SSH session is still alive but SFTP is not.

**To Reproduce**
1. Create a SSHFS mount.
2. Cause `SftpServer::run` to exit by some means.
3. Observe the SSHFS server process running with no errors logged and not functional.

**Expected behavior**
The SSHFS watchdog notices the thread exit and takes appropriate actions so that there are logs and the shell doesn't hang.

Sploder12:
Sets the timeout to infinite since no messages for extended periods of time is expected behavior. Libssh updated to fix timeout being ignored.
Sets the pty mode explicitly to default since Libssh updated to inherit settings from `stdin`.

Fixes #3745

MULTI-1528

### DIFF
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -96,6 +96,7 @@ int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 // Creates a function that will wait for signals or until the passed function returns false.
 // The passed function is checked every `period` milliseconds.
 // If a signal is received the optional contains it, otherwise the optional is empty.
+// `make_quit_watchdog` should only be called once.
 std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
     const std::chrono::milliseconds& period); // call while single-threaded; call result later, in dedicated thread
 
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -93,7 +93,9 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
-std::function<int()> make_quit_watchdog(); // call while single-threaded; call result later, in dedicated thread
+std::function<int()> make_quit_watchdog(
+    const std::chrono::milliseconds& timeout,
+    const std::function<bool()>& condition); // call while single-threaded; call result later, in dedicated thread
 
 std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
 
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -93,6 +93,9 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
+// Creates a function that will wait for signals or until the passed function returns false.
+// The passed function is checked every timeout milliseconds.
+// If a signal is received the optional contains it, otherwise the optional is empty.
 std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
     const std::chrono::milliseconds& timeout); // call while single-threaded; call result later, in dedicated thread
 
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -94,10 +94,10 @@ std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spe
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
 // Creates a function that will wait for signals or until the passed function returns false.
-// The passed function is checked every timeout milliseconds.
+// The passed function is checked every `period` milliseconds.
 // If a signal is received the optional contains it, otherwise the optional is empty.
 std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
-    const std::chrono::milliseconds& timeout); // call while single-threaded; call result later, in dedicated thread
+    const std::chrono::milliseconds& period); // call while single-threaded; call result later, in dedicated thread
 
 std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
 
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -93,9 +93,8 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
-std::function<int()> make_quit_watchdog(
-    const std::chrono::milliseconds& timeout,
-    const std::function<bool()>& condition); // call while single-threaded; call result later, in dedicated thread
+std::function<int(const std::function<bool()>&)> make_quit_watchdog(
+    const std::chrono::milliseconds& timeout); // call while single-threaded; call result later, in dedicated thread
 
 std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
 
--- a/include/multipass/platform.h
+++ b/include/multipass/platform.h
@@ -93,7 +93,7 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
-std::function<int(const std::function<bool()>&)> make_quit_watchdog(
+std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
     const std::chrono::milliseconds& timeout); // call while single-threaded; call result later, in dedicated thread
 
 std::string reinterpret_interface_id(const std::string& ux_id); // give platforms a chance to reinterpret network IDs
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -71,21 +71,15 @@ mp::UnixConsole::UnixConsole(ssh_channel channel, UnixTerminal* term) : term{ter
 
     if (term->is_live())
     {
-        setup_console();
-
         const char* term_type = std::getenv("TERM");
         term_type = (term_type == nullptr) ? "xterm" : term_type;
 
         update_local_pty_size(term->cout_fd());
 
-        // do not inherit settings from stdin
-        constexpr unsigned char modes[1] = {0};
-        ssh_channel_request_pty_size_modes(channel,
-                                           term_type,
-                                           local_pty_size.columns,
-                                           local_pty_size.rows,
-                                           modes,
-                                           sizeof(modes));
+        ssh_channel_request_pty_size(channel, term_type, local_pty_size.columns, local_pty_size.rows);
+
+        // set stdin to Raw Mode after libssh inherits sane settings from stdin.
+        setup_console();
     }
 }
 
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -77,7 +77,10 @@ mp::UnixConsole::UnixConsole(ssh_channel channel, UnixTerminal* term) : term{ter
         term_type = (term_type == nullptr) ? "xterm" : term_type;
 
         update_local_pty_size(term->cout_fd());
-        ssh_channel_request_pty_size(channel, term_type, local_pty_size.columns, local_pty_size.rows);
+
+        // do not inherit settings from stdin
+        constexpr unsigned char modes[1] = {0};
+        ssh_channel_request_pty_size_modes(channel, term_type, local_pty_size.columns, local_pty_size.rows, modes, sizeof(modes));
     }
 }
 
--- a/src/platform/console/unix_console.cpp
+++ b/src/platform/console/unix_console.cpp
@@ -80,7 +80,12 @@ mp::UnixConsole::UnixConsole(ssh_channel channel, UnixTerminal* term) : term{ter
 
         // do not inherit settings from stdin
         constexpr unsigned char modes[1] = {0};
-        ssh_channel_request_pty_size_modes(channel, term_type, local_pty_size.columns, local_pty_size.rows, modes, sizeof(modes));
+        ssh_channel_request_pty_size_modes(channel,
+                                           term_type,
+                                           local_pty_size.columns,
+                                           local_pty_size.rows,
+                                           modes,
+                                           sizeof(modes));
     }
 }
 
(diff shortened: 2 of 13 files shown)


ricab (commented):
Nice work Trevor, thanks. I ask only for some unit tests for the new code. Other than that, +1 on my secondary review.

georgeliao (commented):
Hi @Sploder12 

Thanks for the good work, the fix seems to be solid. I only have a few minor comments and questions.

src/ssh/ssh_session.cpp
```diff
@@ -33,14 +33,13 @@ namespace mpl = multipass::logging;
...
-                           const SSHKeyProvider& key_provider,
-                           const std::chrono::milliseconds timeout)
+                           const SSHKeyProvider& key_provider)
     : session{ssh_new(), ssh_free}, mut{}
 {
     if (session == nullptr)
         throw mp::SSHException("could not allocate ssh session");
 
-    const long timeout_secs = std::chrono::duration_cast<std::chrono::seconds>(timeout).count();
+    const long timeout_secs = std::numeric_limits<long>::max();
```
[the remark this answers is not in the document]
ricab: Wasn't this a keep-alive timeout? A better var name might help here.

ricab (commented):
Great PR @Sploder12! And nice review as well @georgeliao :smiley:

A few comments on my secondary review.

src/platform/console/unix_console.cpp
```diff
@@ -71,13 +71,14 @@ mp::UnixConsole::UnixConsole(ssh_channel channel, UnixTerminal* term) : term{ter
...
     {
-        setup_console();
-
         const char* term_type = std::getenv("TERM");
         term_type = (term_type == nullptr) ? "xterm" : term_type;
 
         update_local_pty_size(term->cout_fd());
         ssh_channel_request_pty_size(channel, term_type, local_pty_size.columns, local_pty_size.rows);
+
+        // set stdin to Raw Mode after libssh inherits sane settings from stdin.
```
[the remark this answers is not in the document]
Sploder12: yep

src/ssh/ssh_session.cpp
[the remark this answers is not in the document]
Sploder12: The program will not behave correctly with a reasonable timeout unfortunately. After the timeout expires, the sshfs_server process will exit, the watchdog fix makes it so that it doesn't hang. It behaves like a keep-alive timeout, but libssh has no keep-alive messages.

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -151,21 +151,28 @@ auto make_sftp_server(mp::SSHSession&& session, const std::string& source, const
...
 
-mp::SshfsMount::SshfsMount(SSHSession&& session, const std::string& source, const std::string& target,
-                           const mp::id_mappings& gid_mappings, const mp::id_mappings& uid_mappings)
-    : sftp_server{make_sftp_server(std::move(session), source, target, gid_mappings, uid_mappings)},
+mp::SshfsMount::SshfsMount(SSHSession&& session,
+                           const std::string& source,
+                           const std::string& target,
+                           const mp::id_mappings& gid_mappings,
+                           const mp::id_mappings& uid_mappings)
+    : running{true},
```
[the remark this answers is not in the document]
Sploder12: Ricardo's belief is correct. The watchdog would immediately exit if the thread takes a while to start. I think the enum approach would work well though.

src/ssh/ssh_session.cpp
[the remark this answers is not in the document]
georgeliao: ok, so it looks like a keep-alive timeout as opposed to ssh session connection establishment wait time, interesting. It is a weird option offered to users though.

src/sshfs_mount/sshfs_mount.cpp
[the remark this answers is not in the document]
georgeliao: ok, i see, then the original setting is fine.

src/ssh/ssh_session.cpp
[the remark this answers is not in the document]
Sploder12: `multipass shell` doesn't hang since it's setup to automatically reconnect. You can see this behavior in https://github.com/canonical/multipass/issues/3810 where the SSH session is restarting every 20 seconds. Also with normal use it's much more likely there is some SSH traffic within the 20 seconds compared to SFTP which is likely to not get any traffic for long periods of time.

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,53 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
 {
-    return [sigset = make_and_block_signals({SIGQUIT, SIGTERM, SIGHUP})]() {
-        int sig = -1;
-        sigwait(&sigset, &sig);
-        return sig;
+    const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(duration);
+    return timespec{seconds.count(), std::chrono::duration_cast<std::chrono::nanoseconds>(duration - seconds).count()};
+}
+
+#include <iostream>
```
georgeliao, line 187: This header inclusion definitely can be removed.

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,53 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+    const std::chrono::milliseconds& timeout)
+{
+    return [sigset = make_and_block_signals({SIGQUIT, SIGTERM, SIGHUP, SIGUSR2}),
+            timeout](const std::function<bool()>& condition) -> std::optional<int> {
+        std::mutex sig_mtx;
+        std::condition_variable sig_cv;
+        int sig = SIGUSR2;
+
+        // A signal generator that triggers after `timeout`
+        AutoJoinThread signaler([&sig_mtx, &sig_cv, &sig, &timeout, signalee = pthread_self()] {
```
georgeliao, line 199: Ok, we need to replace this in house utility when std::jthread is available.

src/platform/platform_unix.cpp
Sploder12: oops, didn't fully clean up my print debugging.

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,51 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+            std::unique_lock lock(sig_mtx);
+            while (sig == SIGUSR2)
+            {
+                auto status = sig_cv.wait_for(lock, timeout);
```
georgeliao, line 201: I think you can use the `wait_for` overloaded function with the predicate to check the `sig != SIGUSR2`, which can simplify the code here.

Sploder12: Good catch, so something like this should work
```c++
while (!sig_cv.wait_for(lock, timeout, [&sig]() { return sig != SIGUSR2; })) {
    pthread_kill(signalee, SIGUSR2);
}
```

include/multipass/platform.h
```diff
@@ -91,7 +91,11 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
-std::function<int()> make_quit_watchdog(); // call while single-threaded; call result later, in dedicated thread
+// Creates a function that will wait for signals or until the passed function returns false.
+// The passed function is checked every timeout milliseconds.
+// If a signal is received the optional contains it, otherwise the optional is empty.
+std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
+    const std::chrono::milliseconds& timeout); // call while single-threaded; call result later, in dedicated thread
```
ricab, line 98: This parameter is used as a timeout in the implementation, but it's really a period for the caller, right? Should the name reflect that?

include/multipass/platform.h
```diff
@@ -91,7 +91,11 @@ std::unique_ptr<Process> make_sshfs_server_process(const SSHFSServerConfig& conf
 std::unique_ptr<Process> make_process(std::unique_ptr<ProcessSpec>&& process_spec);
 int symlink_attr_from(const char* path, sftp_attributes_struct* attr);
 
-std::function<int()> make_quit_watchdog(); // call while single-threaded; call result later, in dedicated thread
+// Creates a function that will wait for signals or until the passed function returns false.
+// The passed function is checked every timeout milliseconds.
+// If a signal is received the optional contains it, otherwise the optional is empty.
+std::function<std::optional<int>(const std::function<bool()>&)> make_quit_watchdog(
```
ricab, line 97: Is there a reason to ever call this and then pass different functions to the result? If not, wouldn't it be clearer to tighten the closure as as much as possible early on and receive the periodic predicate as a parameter in this function? 

I mean, instead of a function that returns a function that receives a function, having a function that receives a function and returns a function. In essence, turn this into an adaptor, rather than something that returns an adaptor. WDYT?

Sploder12: We don't ever pass different predicates and probably never will. But we can't pass the predicate at creation since this function has the precondition of being the only thread and the predicate we use only exists after we make another thread. Though it is possible, it'd be ugly

include/multipass/platform.h
Sploder12: Good point!

include/multipass/platform.h
ricab: Oh OK, got it.

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,46 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+        AutoJoinThread signaler([&sig_mtx, &sig_cv, &sig, &period, signalee = pthread_self()] {
+            std::unique_lock lock(sig_mtx);
+            while (!sig_cv.wait_for(lock, period, [&sig] { return sig != SIGUSR2; }))
+            {
+                pthread_kill(signalee, SIGUSR2);
+            }
+        });
+
+        // wait on signals and condition
+        int ret = SIGUSR2;
```
ricab, line 206: Can this name be improved a little, please? This is not returned anywhere. I know it is used for the output of `sigwait`, but I would expect ret to be reserved for something we return, not the return of a function we call. Perhaps `latest_sig` or something of the sort. Feel free to come up with something better.

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,46 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+            sigwait(&sigset, &ret);
+        }
+
+        {
+            std::lock_guard lock(sig_mtx);
+            sig = ret == SIGUSR2 ? 0 : ret;
+        }
+        sig_cv.notify_all();
+
+        // if sig is SIGUSR2 then we know we're exiting because condition() was false
```
ricab, line 219: We've already used this reasoning in line 215, so should the comment be moved there?

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,46 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+        return ret == SIGUSR2 ? std::nullopt : std::make_optional(sig);
```
ricab, line 220: And then we could use a single variable here to simplify this slightly:

```suggestion
        return sig ? std::make_optional(sig) : std::nullopt;
```

src/platform/platform_unix.cpp
```diff
@@ -175,11 +177,46 @@ sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
...
+    const std::chrono::milliseconds& period)
+{
+    return [sigset = make_and_block_signals({SIGQUIT, SIGTERM, SIGHUP, SIGUSR2}),
+            period](const std::function<bool()>& condition) -> std::optional<int> {
+        std::mutex sig_mtx;
+        std::condition_variable sig_cv;
+        int sig = SIGUSR2;
+
+        // A signal generator that triggers after `timeout`
+        AutoJoinThread signaler([&sig_mtx, &sig_cv, &sig, &period, signalee = pthread_self()] {
```
ricab, line 197: I wonder if we should just use our existing [`Timer`](https://github.com/canonical/multipass/blob/43bd53f5eb6c7e9c31b6fa51c2330e50543625a2/include/multipass/timer.h#L40)...

src/platform/platform_unix.cpp
Sploder12: They have different purposes so probably not. But adding a comment to 215 would probably be good

Sploder12:
@ricab I can try to write unit tests. Timer won't help much since the main issue is the signals. Integration tests would be needed to properly test this too, sending SIGTERM is a bit too dangerous for a unit test. Where/How would I write an integration test?

src/platform/platform_unix.cpp
Sploder12: Nevermind, I didn't think long or hard enough about it. `Timer` works well here (I think)

georgeliao (dismissed):

src/platform/platform_unix.cpp
```diff
@@ -157,6 +158,25 @@ int mp::platform::symlink_attr_from(const char* path, sftp_attributes_struct* at
...
+
+int mp::platform::SignalWrapper::send(pthread_t target, int signal) const
+{
+    return pthread_kill(target, signal);
+}
+
+int mp::platform::SignalWrapper::wait(const sigset_t& sigset, int& got) const
+{
+    return sigwait(std::addressof(sigset), std::addressof(got));
+}
```
ricab, line 178: For the record, it is fine for these to go untested by unit tests, since they are themselves mere wrappers to enable testing (mocking) elsewhere.

ricab (commented):
Good stuff Trevor, thanks! This is pretty much there, I have only a couple last renaming requests. They're not a big deal... I am hoping the IDE can rename everything for you, but let me know if this is troublesome for any reason.

georgeliao:
@Sploder12 since the windows build fix is in, maybe do a rebase on the main and trigger a total rebuild.

georgeliao (approved):

ricab (approved):
Perfect, thanks!

src/platform/platform_unix.cpp
```diff
@@ -171,15 +191,39 @@ sigset_t mp::platform::make_sigset(const std::vector<int>& sigs)
...
 {
     auto sigset{make_sigset(sigs)};
-    pthread_sigmask(SIG_BLOCK, &sigset, nullptr);
+    MP_POSIX_SIGNAL.pthread_sigmask(SIG_BLOCK, &sigset, nullptr);
     return sigset;
 }
 
-std::function<int()> mp::platform::make_quit_watchdog()
+template <class Rep, class Period>
+timespec make_timespec(std::chrono::duration<Rep, Period> duration)
```
ricab, line 199: @Sploder12, it looks like this is no longer used and was forgotten?

Sploder12: ah, good catch

ricab (approved):
LGTM!

ricab:
This is going to fail because it needs the other side. After we confirm that is the only issue, we can merge manually.

PR #3754, opened 2024-10-23
[gui] make info providers update even if data is equal

andrei-toterman:
Providers only update on new data. Built collections provide deep equality comparison. But in our case, we want vm info providers to update even if the data is the same. For example, a stopped vm will emit the same info, so CPU sparkline does not update because of that. So this PR makes providers emit regular collections, which do not provide deep equality, so this makes the GUI update even if the data that comes from a vm is the same. This could also be optimized with more granular updates in the needed providers, if we notice the need for more performance.


levkropp (approved):
LGTM! Well done!

PR #3768, opened 2024-11-07
Scale VMStats in GUI details according to platform 

ricab:
Scale VMStats according to the text scale in use. Also extract a magic number. Thanks for the tip @andrei-toterman !

[MULTI-1571]

[MULTI-1571]: https://warthogs.atlassian.net/browse/MULTI-1571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/gui/lib/vm_details/vm_details_general.dart
+++ b/src/client/gui/lib/vm_details/vm_details_general.dart
@@ -130,7 +130,7 @@ class VmStat extends StatelessWidget {
 
 class GeneralDetails extends ConsumerWidget {
   final String name;
-  static const double preferredHeight = 50;
+  static const double baseVmStatHeight = 50;
 
   const GeneralDetails(this.name, {super.key});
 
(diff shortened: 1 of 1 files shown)


src/client/gui/lib/vm_details/vm_details_general.dart
```diff
@@ -129,6 +130,7 @@ class VmStat extends StatelessWidget {
 
 class GeneralDetails extends ConsumerWidget {
   final String name;
+  static const double preferredHeight = 50;
```
andrei-toterman, line 133: Perhaps this can benefit from a better name. Something like `defaultHeightForEntry` to make it clear which height it is referencing. Also, specifying the type is not really needed if it can be inferred, so you could do something like `static const variable = 50.0;`

andrei-toterman (changes requested):
Thanks, @ricab! I tested it and those widgets seem to scale well now. There are still many places in the GUI that need to be taken care of for this, but this is a very good start! I just have a tiny rename request :)

andrei-toterman:
Oops, I pressed 'Add single comment' instead of 'Start a review' for that change request. Sorry!

src/client/gui/lib/vm_details/vm_details_general.dart
ricab: Based on our discussion elsewhere, I updated to baseVmStatsHeight. I ended up not modifying the name of the parameter itself since, as you say, the better approach would be for the widget to just derive the height from the child. Let me know if this is what you had in mind.

levkropp (approved):

PR #3775, opened 2024-11-13
Fix typo in launch success notification

Sploder12:
Fixes launch success notification typo. Previously it said `You can start using using it now`. Now it says `You can start using it now`.


andrei-toterman (approved):
Thanks, @Sploder12!

PR #3781, opened 2024-11-18
[ci] make the renovate bot run only every 2 weeks

andrei-toterman:
This PR makes the renovate bot run only every two weeks on Monday, and it also moves the config file to JSON5 so that it can support trailing commas.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,4 +3,5 @@
   "extends": [
     "config:recommended"
   ],
+  "schedule": "every 2 weeks on Monday",
 }
--- a/renovate.json5
+++ b/renovate.json5
@@ -2,5 +2,5 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:recommended"
-  ]
+  ],
 }


sharder996 (approved):
LGTM!

ricab (commented):
Did you test this somehow? And why the move to JSON5 (for my understanding)?

andrei-toterman:
@ricab I updated the PR description to mention that the move to JSON5 was just so that the file can support trailing commas. Not a necessity, but IMO it would be nice in the future if we need to add more config options, to not have to worry about that last comma.

As for testing, I ran `npx --yes --package renovate -- renovate-config-validator` and it passed the validation. Of course, I also ran that validation with a small mistake in the file to make sure it actually accepts only the right stuff :) But I can't really be sure how the actual bot will behave now, we'll just have to wait to see, as I don't really know if and how we can test the actual bot.

andrei-toterman:
Also, in the docs they mention that they use the `@breejs/later` npm package to parse that schedule string, and I played with it locally a bit and it could parse my string.

ricab (approved):
Perfect, thanks for the clarifications!

ricab:
> @ricab https://docs.renovatebot.com/presets-schedule/#schedulequarterly
> 
> As a similar example

Yep, I'd seen that, but I wasn't sure if it could do "on monday".

PR #3785, opened 2024-11-19
[cmake] Fix exclusion of build files from coverage

ricab:
Fix the pattern for excluding build dir files from the coverage report. Without this, nothing matches the build dir pattern given to `lcov` (at least with my version).

I suppose Codecov knows to exclude anything that is not part of the repo, which would explain why its results are not impacted.

Local test recipe, from build directory:

1. `cmake -DCMAKE_BUILD_TYPE=Coverage ...`
2. `cmake --build . --target covreport`
3. `xdg-open coverage/index.html`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,7 +295,7 @@ if(cmake_build_type_lower MATCHES "coverage")
         '/usr/*'
         ${CMAKE_SOURCE_DIR}'/3rd-party/*'
         ${CMAKE_SOURCE_DIR}'/tests/*'
-        ${CMAKE_BINARY_DIR}
+        ${CMAKE_BINARY_DIR}'/*'
         --output-file coverage.cleaned
       COMMAND ${CMAKE_COMMAND} -E remove coverage.info
       COMMAND ${GENHTML} -o coverage coverage.cleaned


CMakeLists.txt
```diff
@@ -295,7 +295,7 @@ if(cmake_build_type_lower MATCHES "coverage")
         '/usr/*'
         ${CMAKE_SOURCE_DIR}'/3rd-party/*'
         ${CMAKE_SOURCE_DIR}'/tests/*'
-        ${CMAKE_BINARY_DIR}
+        ${CMAKE_BINARY_DIR}'/*'
```
georgeliao, line 298: Makes sense to me even though it does not change the behavior.

georgeliao (approved):

CMakeLists.txt
ricab: Doesn't it change behavior for you? It does for me, locally. It may have to do with tool versions...

georgeliao: > I suppose Codecov knows to exclude anything that is not part of the repo, which would explain why its results are not impacted.

I am talking about this one. Meaning that Codecov does not count anything in to calculate the code coverage.

ricab: Ah OK. Right so this is meant to fix coverage in general, not just codecov.

PR #3789, opened 2024-11-22
Fix basic process tests on Windows

### LINKED ISSUE #3787: Basic Process tests crash/fail on Windows
**Describe the bug**
When running tests locally on Windows, basic process tests that test for crashing will fail. This is from a combination of issues, the first being trying to get the value of a `std::nullopt`, the second being a difference in how QProcess and Windows handles `abort`.

**To Reproduce**
1. Run tests on Windows
2. Observe the test suite crash or have basic process tests fail

**Expected behavior**
The tests pass and other tests continue

**Additional info**
 - OS: Windows 10 Home/Pro

**Additional context**
These tests seem to pass in CI but not locally for me. They fail on my dedicated Windows 10 Home machine and my Windows 10 Pro VM. Related to debug vs release building.

Sploder12:
fixes #3787 

Some `EXPECT`s should've been `ASSERT`s because of getting value of `std::nullopt`
Windows compiled with debug has different behavior from release builds, this has been fixed for the tests.

MULTI-1652

### DIFF
--- a/tests/test_basic_process.cpp
+++ b/tests/test_basic_process.cpp
@@ -38,7 +38,7 @@ TEST_F(BasicProcessTest, execute_missing_command)
     EXPECT_FALSE(process_state.completed_successfully());
     EXPECT_FALSE(process_state.exit_code);
 
-    EXPECT_TRUE(process_state.error);
+    ASSERT_TRUE(process_state.error);
     EXPECT_EQ(QProcess::ProcessError::FailedToStart, process_state.error->state);
 }
 
@@ -50,7 +50,7 @@ TEST_F(BasicProcessTest, execute_crashing_command)
     EXPECT_FALSE(process_state.completed_successfully());
     EXPECT_FALSE(process_state.exit_code);
 
-    EXPECT_TRUE(process_state.error);
+    ASSERT_TRUE(process_state.error);
     EXPECT_EQ(QProcess::ProcessError::Crashed, process_state.error->state);
 }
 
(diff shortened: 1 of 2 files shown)


levkropp (approved):

tests/test_basic_process.cpp
```diff
@@ -38,7 +38,7 @@ TEST_F(BasicProcessTest, execute_missing_command)
     EXPECT_FALSE(process_state.completed_successfully());
     EXPECT_FALSE(process_state.exit_code);
 
-    EXPECT_TRUE(process_state.error);
+    ASSERT_TRUE(process_state.error);
```
andrei-toterman, line 41: I'm not sure I understand why this was needed. Yes, `error` is a `std::optional`, and we want to check if it has a value. But what does the assert do differently from the expect?

Sploder12: Assert will abort the test so if there is no value in the optional we don't crash from trying to access it

andrei-toterman: ah, i wasn't reading the line just below this. apologies.
but this raises the question: if the error optional does not have a value, why do we assert/expect it to have one. is the content of `error` not deterministic? shouldn't we have the test always pass?

also, not requesting this as a change, but what's your opinion on something like
```cpp
const auto process_has_error = process_state.error.has_value();
EXPECT_TRUE(process_has_error);
if (process_has_error) {
    EXPECT_EQ(..., process_state.error->state);
}
```
I'm asking this because I think (please correct me) that if an EXPECT fails, the test will report all of the other failed EXPECTs, but if an ASSERT fails, the test won't report any other failed EXPECTs.

Sploder12: The test is making sure that the optional does have a value, if the unit under test has bug then the optional might not have a value and the test should fail. Before this PR the unit had a bug and instead of failing it would crash. 

I think that change would be good if we had more assets/expects after we check the value, assuming those checks make sense if there is no value. ASSERT will abort the test so any other expects after won't be reported, but any EXPECT before the ASSERT will still be reported

andrei-toterman: Ok, perfect. Makes sense. Thanks for the details!

andrei-toterman (approved):

PR #3797, opened 2024-11-26
[cli] update messages to clarify default mount target

andrei-toterman:
Now we use a different way of deriving the target path of a mount based on the source path, if the target path is not explicitly specified. We take the directory name of the source and use it under `/home/ubuntu` in the instance.
For example, for a source path `/abc/def/ghi` the corresponding default target is `/home/ubuntu/ghi`


giuliazanchi (approved):
LGTM!

ricab (approved):
:+1:

PR #3800, opened 2024-11-29
[gui] set version to 1.15.0

andrei-toterman:
GUI version must be updated manually, unfortunately...

### DIFF
--- a/src/client/gui/pubspec.yaml
+++ b/src/client/gui/pubspec.yaml
@@ -2,7 +2,7 @@ name: multipass_gui
 description: A new Flutter project.
 publish_to: 'none'
 
-version: 1.14.1
+version: 1.15.0
 
 environment:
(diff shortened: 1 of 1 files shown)


ricab:
Oh... We'll need RC2 then.

ricab (approved):
Hey @andrei-toterman, I am approving, but where does this show in the GUI? Also, how difficult would it be to make it pick up from the build, as the CLI and daemon do?

andrei-toterman:
It is shown on macOS in the app info
image
As for making CMake fill that value in at configure time, it should be entirely possible, but not using CMake's configure_file, since that would introduce some difficulties. We'd have to use some custom regex stuff, but it shouldn't be too hard.

ricab:
OK thanks, I opened #3803 for that.

PR #3805, opened 2024-12-02
[renovatebot] Auto assign random reviewer from Multipass GitHub team

sharder996:
Unfortunately, this means that @jibel will get the occasional review request. 

But, I think this is a better approach than manually listing all Multipass developers and having to maintain it when developers join/leave.

P.S. and @giuliazanchi!


ricab (approved):
LGTM

PR #3806, opened 2024-12-03
Automatic gui version

andrei-toterman:
Set the GUI version using the --build-name and --build-number args, using the MULTIPASS_VERSION CMake var. This way, we don't have to manually update the version all the time.
For the --build-number, the argument must only contain digits, so we extract the digits out of the commit hash respective to the current version.

MULTI-1702


src/client/gui/CMakeLists.txt.linux
```diff
@@ -29,7 +29,7 @@ set(MULTIPASS_GUI_EXECUTABLE_FULL_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${MULTI
 
 file(GLOB_RECURSE MULTIPASS_GUI_SOURCE_FILES ${MULTIPASS_GUI_SOURCE_DIR}/lib/*.dart)
 
-set(FLUTTER_BUILD_ARGS --release --suppress-analytics --verbose)
+set(FLUTTER_BUILD_ARGS --release --suppress-analytics --verbose --build-name=${MULTIPASS_VERSION})
```
ricab, line 32: Shouldn't we split between build name and number?

ricab (commented):
Thank you @andrei-toterman!

src/client/gui/CMakeLists.txt.linux
andrei-toterman: Ah, yes we should!

andrei-toterman: Since the --build-number can only contain digits, I made it so that only the digits are extracted from the version string. So from `1.15.0-...`, all the digits are extracted. If there are none, it defaults to 0. This already happens by default on macOS, so I made it the same on all platforms.

CMakeLists.txt
```diff
@@ -208,14 +208,33 @@ function(determine_version OUTPUT_VARIABLE)
...
   endif()
+
+  string(REPLACE "-" ";" MULTIPASS_VERSION_COMPONENTS ${VERSION_STRING})
+  list(GET MULTIPASS_VERSION_COMPONENTS 0 MULTIPASS_SEMANTIC_VERSION)
+  set(MULTIPASS_SEMANTIC_VERSION ${MULTIPASS_SEMANTIC_VERSION})
+  set(MULTIPASS_BUILD_VERSION 0)
+  list(LENGTH MULTIPASS_VERSION_COMPONENTS MULTIPASS_VERSION_COMPONENTS_LENGTH)
+  if (MULTIPASS_VERSION_COMPONENTS_LENGTH GREATER 1)
+    list(GET MULTIPASS_VERSION_COMPONENTS 1 MULTIPASS_BUILD_VERSION_STR)
+    string(REGEX REPLACE "[^0-9]" "" MULTIPASS_BUILD_VERSION_NUMBER "${MULTIPASS_BUILD_VERSION_STR}")
```
sharder996, line 224: The CI runs sets the build version to `Setting build version to: 76638061597231`, cutting out the 1.16.0. IIUC release builds will then have a build version of `0`.

sharder996 (changes requested):

CMakeLists.txt
```diff
@@ -208,14 +208,33 @@ function(determine_version OUTPUT_VARIABLE)
...
+    if(NOT "${MULTIPASS_BUILD_VERSION_NUMBER}" STREQUAL "")
+        set(MULTIPASS_BUILD_VERSION "${MULTIPASS_BUILD_VERSION_NUMBER}")
+    endif()
+  endif()
+  set(MULTIPASS_SEMANTIC_VERSION ${MULTIPASS_SEMANTIC_VERSION} PARENT_SCOPE)
+  set(MULTIPASS_BUILD_VERSION ${MULTIPASS_BUILD_VERSION} PARENT_SCOPE)
```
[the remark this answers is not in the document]
andrei-toterman: Yep, my bad.

src/client/gui/CMakeLists.txt.linux
ricab: For my understanding, what would we get in these examples?

- 1.15.0-rc3.748+ga9411c0e9
- 1.15.0-rc4+win.3531+g65170ab8a.win
- 1.16.0-dev.3557.pr698+g9bae52f58.win
- 1.16.0-dev.3550.pr3811+g2c48add28.mac
- 1.16.0-dev.764.ci7330+gfa725c28c

ricab: And didn't you say that we could have the entire thing in the build name? Also, notice that we don't actually have a build number anywhere in there. It's all deterministic regardless of the build.

andrei-toterman: `1.15.0-rc3.748+ga9411c0e9` => `--build-name=1.15.0 --build-number=3748941109`
`1.15.0-rc4+win.3531+g65170ab8a.win` => `--build-name=1.15.0 --build-number=43531651708`
`1.16.0-dev.3557.pr698+g9bae52f58.win` => `--build-name=1.16.0 --build-number=355769895258`
`1.16.0-dev.3550.pr3811+g2c48add28.mac` => `--build-name=1.16.0 --build-number=3350381124828`
`1.16.0-dev.764.ci7330+gfa725c28c` => `--build-name=1.16.0 --build+-number=764733072528`

I was wrong about being able to pass something like `--build-name=1.16.0-dev.764.ci7330+gfa725c28c`. It worked on macos, firstly because for them the version is just a single string, not 3 numbers, and secondly because they do what I did up there aka removing all of the non-digits, so `--build-name=1.16.0-dev.764.ci7330+gfa725c28c` gets transformed into `--build-name=1.16.0.764.733072528`, which I don't think it's right, since the build name should be just 1.16.0 in this case, and the rest (or a part of it) should be the build number.
On Linux, I thought those flags are meaningless, but they do actually do something: they set the CMake variables `FLUTTER_VERSION_(MAJOR|MINOR|PATCH|BUILD)`, which are 4 independent numbers. To be honest, I don't know exactly the purpose of those, but its better to be safe then sorry and have those set correctly. And here, passing something like `--build-name=1.16.0 --build-number=dev.764.ci7330+gfa725c28c` simply prints a warning that the version could not be parsed and set those values to 1.0.0.0.
So the safe approach would be to pass correct arguments. For `--build-name`, the correct value is clear, it should be `1.16.0`. The question is then what number to pass to `--build-number`? I agree that my approach is not at all intuitive, and I did it just because that's what macos did. But what unique number could we derive for each build then?

ricab: I am actually more concerned with the build name. Something like `1.16.0-dev.3557.pr698+g9bae52f58.win` is a *pre-release*, so it is actually lower than "1.16.0" (see https://semver.org/#spec-item-9 and the last example in https://semver.org/#spec-item-11). With this approach, that is lost entirely. I wonder if we should just set the version to 1.15.0 then...

These tools... not even support for semantic versioning :dotted_line_face:

andrei-toterman: For posterity, after understanding better the meaning of the semantic version components, what we need is `--build-name=1.16.0-dev.3557.pr698` and `--build-number=decimal(9bae52f58)` which means that we take the commit hash and turn it from hexadecimal to decimal. This aligns most closely with semantic versioning, where stuff after `+` is considered build metadata. But since flutter only accepts a decimal number for that, converting the commit hash to decimal was a good compromise.

CMakeLists.txt
```diff
@@ -208,14 +208,33 @@ function(determine_version OUTPUT_VARIABLE)
...
   if(VERSION_MATCH)
     set(${OUTPUT_VARIABLE} ${CMAKE_MATCH_1} PARENT_SCOPE)
+    set(VERSION_STRING ${CMAKE_MATCH_1})
+    
   else()
     message(FATAL_ERROR "Invalid tag detected: ${NEW_VERSION}")
   endif()
+
+  string(REPLACE "-" ";" MULTIPASS_VERSION_COMPONENTS ${VERSION_STRING})
+  list(GET MULTIPASS_VERSION_COMPONENTS 0 MULTIPASS_SEMANTIC_VERSION)
```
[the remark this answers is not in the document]
andrei-toterman: We went with a different approach, where everything before a '+' is the version and we derive the build number from the commit hash after the '+'. I tried making it work with regex but I couldn't, so I did it with just simple find+substring.

CMakeLists.txt
andrei-toterman: So now, the build number will the the decimal representation of the commit hash, or 0 otherwise.

CMakeLists.txt
```diff
@@ -208,14 +208,34 @@ function(determine_version OUTPUT_VARIABLE)
...
     message(FATAL_ERROR "Invalid tag detected: ${NEW_VERSION}")
   endif()
+
+  # stuff before + is the version and after the + is the build metadata
+  string(FIND "${VERSION_STRING}" "+" PLUS_POS)
+  string(SUBSTRING "${VERSION_STRING}" 0 ${PLUS_POS} MULTIPASS_SEMANTIC_VERSION)
+  if (PLUS_POS GREATER -1)
+      string(SUBSTRING "${VERSION_STRING}" ${PLUS_POS} -1 MULTIPASS_BUILD_STRING)
+      # if the build metadata starts with a g, the hexadecimal chars after that are a commit hash
+      # otherwise, we do not derive any build string
```
ricab, line 222: Just for my understanding, did you find any valid case with something other than a 'g' after the plus?

CMakeLists.txt
```diff
@@ -208,14 +208,34 @@ function(determine_version OUTPUT_VARIABLE)
 
   if(VERSION_MATCH)
     set(${OUTPUT_VARIABLE} ${CMAKE_MATCH_1} PARENT_SCOPE)
+    set(VERSION_STRING ${CMAKE_MATCH_1})
   else()
     message(FATAL_ERROR "Invalid tag detected: ${NEW_VERSION}")
   endif()
+
```
ricab, line 215: This whole block could be run as a separate step, right? Could we have a separate function for it instead, and call it only where relevant (i.e. Flutter build, possibly MSI stuff later)?

ricab (commented):
Hey Andrei. The approach looks solid. Just a couple things inline.

CMakeLists.txt
ricab: Yes, that is what I meant. Should have been clearer. BTW, I like that you are printing the vars that are getting derived. It would be nice if you could keep that.

CMakeLists.txt
ricab: Yeah, nahhh. It's fine as you have it :) Thanks for clarifying.

ricab (approved):
Approving on the same grounds as on the other side.

CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ find_package(Qt6 COMPONENTS Core Concurrent Network REQUIRED)
 # POCO config
 find_package(Poco REQUIRED COMPONENTS Foundation Zip)
 
-function(determine_version OUTPUT_VARIABLE)
+function(determine_version OUTPUT_VARIABLE SEMANTIC_VERSION BUILD_NUMBER)
```
sharder996, line 140: `SEMANTIC_VERSION BUILD_NUMBER` is unnecessary now, right?

CMakeLists.txt
```diff
@@ -213,7 +213,26 @@ function(determine_version OUTPUT_VARIABLE)
...
+  endif()
+  # convert the hexadecimal commit hash to decimal. if none was extracted, this defaults to 0
+  # we need it to be decimal because flutter --build-number accepts only decimal chars
+  math(EXPR MULTIPASS_BUILD_NUMBER "0x0${BUILD_NUMBER_HEX}")
+
+  set(${SEMANTIC_VERSION} ${MULTIPASS_SEMANTIC_VERSION} PARENT_SCOPE)
+  set(${BUILD_NUMBER} ${MULTIPASS_BUILD_NUMBER} PARENT_SCOPE)
+endfunction()
+
+determine_version(MULTIPASS_VERSION MULTIPASS_SEMANTIC_VERSION MULTIPASS_BUILD_NUMBER)
```
sharder996, line 235: idem

sharder996 (changes requested):

CMakeLists.txt
andrei-toterman: yup, good catch

CMakeLists.txt
andrei-toterman: yep, thanks

sharder996 (approved):
LGTM

PR #3808, opened 2024-12-03
[ci] Check whitespace on PRs

ricab:


Sploder12:
clang-format provides [InsertNewlineAtEOF](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertnewlineateof) which could be useful, but it's from version 16 so we'd have to update our versions.

andrei-toterman (commented):
I see this runs only on the Linux workflow. When this runs on the private repo, will it work?

.github/workflows/linux.yml
```diff
@@ -28,6 +28,12 @@ jobs:
...
         fetch-depth: 2
 
+    - name: Check whitespace
+      uses: ./.github/actions/whitespace
+
+    - name: Check final newline
+      uses: qwrtln/check-final-newline@v1.0.0
+
     - name: Lint the code
       uses: ./.github/actions/lint
```
andrei-toterman, line 38: Instead of adding these as new actions, can't they be integrated into the `./github/actions/lint`?

ricab: Yep, good idea. Done.

andrei-toterman (approved):
So the public side is all good and I am ready to approve. However, I see that on the private side the lint fails on the `packaging/windows/template.rtf` file because it does not have a new line at the end of the file. So I will approve this PR, but we need a private PR which fixes that thing as well.

PR #3814, opened 2024-12-04
[gui] remember window size + better default sizes

### LINKED ISSUE #3812: Derive the GUI size as a percentage of the screen, when within the [min,max] range
Resize the GUI to a sensible percentage of the screen when within a common range.

For screen resolutions in the range ]1024x576, 1600x900[ (exclusive on both ends), resize the GUI to the largest rectangle that is the same ratio of the screen (e.g. 16x9) and enclosed in 80% of the screen. For resolutions outside of those bounds, resize the GUI to same size as today (750x450 when below, 1400x822 when above).

When the screen size cannot be determined, resize the GUI to 750x450.

In the above, the resolution is *in the range* iff the corresponding rectangle encloses the lower bound and is enclosed by the higher bound (with no overlaps).

**Additional context**
In the future, we could consider varying the aspect ratio according to the screen, but we'd need special handling for ultra wide. Actual percentages and sizes open for adjustment (e.g. to minimize how much smaller we go when crossing the high bound).

### LINKED ISSUE #3813: Reapply last GUI size on startup
After a user manually changes the size of the GUI's window, closes the GUI, and reopens it, it should retain the latest (manual) size.

Save the window size when there is a scale event. Reapply it on startup.

andrei-toterman:
fix #3812
fix #3813

MULTI-1696
MULTI-1697

### DIFF
--- a/src/client/gui/lib/extensions.dart
+++ b/src/client/gui/lib/extensions.dart
@@ -107,3 +107,19 @@ extension NullableMap<T> on T? {
     }
   }
 }
+
+// needed because in release mode Flutter does not emit the actual code for toString for some classes
+// instead the returned strings are of type "Instance of '<Type>'"
+// this is done to reduce binary size, and it cannot be turned off :face-with-rolling-eyes:
+// see https://api.flutter.dev/flutter/dart-ui/keepToString-constant.html for more info
+extension SizeActualString on Size {
+  String s() {
+    return 'Size(${width.toStringAsFixed(1)}, ${height.toStringAsFixed(1)})';
+  }
+}
+
+extension RectActualString on Rect {
+  String s() {
+    return 'Rect.fromLTRB(${left.toStringAsFixed(1)}, ${top.toStringAsFixed(1)}, ${right.toStringAsFixed(1)}, ${bottom.toStringAsFixed(1)})';
+  }
+}
--- a/src/client/gui/lib/extensions.dart
+++ b/src/client/gui/lib/extensions.dart
@@ -107,19 +107,3 @@ extension NullableMap<T> on T? {
     }
   }
 }
-
-// needed because in release mode Flutter does not emit the actual code for toString for some classes
-// instead the returned strings are of type "Instance of '<Type>'"
-// this is done to reduce binary size, and it cannot be turned off :face-with-rolling-eyes:
-// see https://api.flutter.dev/flutter/dart-ui/keepToString-constant.html for more info
-extension SizeActualString on Size {
-  String s() {
-    return 'Size(${width.toStringAsFixed(1)}, ${height.toStringAsFixed(1)})';
-  }
-}
-
-extension RectActualString on Rect {
-  String s() {
-    return 'Rect.fromLTRB(${left.toStringAsFixed(1)}, ${top.toStringAsFixed(1)}, ${right.toStringAsFixed(1)}, ${bottom.toStringAsFixed(1)})';
-  }
-}
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -17,6 +17,9 @@ import 'extensions.dart';
 import 'help.dart';
 import 'logger.dart';
 import 'notifications.dart';
+import 'platform/linux.dart';
+import 'platform/platform.dart';
+import 'platform/windows.dart';
 import 'providers.dart';
 import 'settings/hotkey.dart';
 import 'settings/settings.dart';
@@ -234,10 +237,20 @@ Future<Size?> getCurrentScreenSize() async {
   try {
     final screen = await getCurrentScreen();
     if (screen == null) throw Exception('Screen instance is null');
+
+    final scaleFactor = screen.scaleFactor;
+    var size = screen.visibleFrame.size;
+
+    // Adjust size based on the platform
+    if (mpPlatform is LinuxPlatform || mpPlatform is WindowsPlatform) {
+      size = Size(size.width * scaleFactor, size.height * scaleFactor);
+    }
+
     logger.d(
-      'Got Screen{frame: ${screen.frame.s()}, scaleFactor: ${screen.scaleFactor}, visibleFrame: ${screen.visibleFrame.s()}}',
+      'Got Screen{frame: ${screen.frame.s()}, scaleFactor: $scaleFactor, visibleFrame: ${screen.visibleFrame.s()}, adjustedSize: ${size.s()}}',
     );
-    return screen.visibleFrame.size;
+
+    return size;
   } catch (e) {
     logger.w('Failed to get current screen information: $e');
     return null;
@@ -252,10 +265,9 @@ Size? getLastWindowSize(SharedPreferences sharedPreferences, Size? screenSize) {
       : null;
   logger.d('Got last window size: ${size?.s()}');
   if (size == null) return null;
-  final clampedSize = Size(
-    min(size.width, screenSize?.width ?? size.width),
-    min(size.height, screenSize?.height ?? size.height),
-  );
+  final clampedSize = size.width >= screenSize!.width || size.height >= screenSize.height
+      ? computeDefaultWindowSize(screenSize)
+      : size;
   logger.d('Using clamped window size: ${clampedSize.s()}');
   return clampedSize;
 }
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -13,6 +13,7 @@ import 'package:window_size/window_size.dart';
 import 'before_quit_dialog.dart';
 import 'catalogue/catalogue.dart';
 import 'daemon_unavailable.dart';
+import 'extensions.dart';
 import 'help.dart';
 import 'logger.dart';
 import 'notifications.dart';
@@ -38,7 +39,7 @@ void main() async {
   final sharedPreferences = await SharedPreferences.getInstance();
   final screenSize = await getCurrentScreen().then((screen) {
     logger.d(
-      'Got Screen{frame: ${screen?.frame}, scaleFactor: ${screen?.scaleFactor}, visibleFrame: {${screen?.visibleFrame}}}',
+      'Got Screen{frame: ${screen?.frame.s()}, scaleFactor: ${screen?.scaleFactor}, visibleFrame: ${screen?.visibleFrame.s()}}',
     );
     return screen?.frame.size;
   });
@@ -170,7 +171,7 @@ class _AppState extends ConsumerState<App> with WindowListener {
   final saveWindowSizeTimer = RestartableTimer(1.seconds, () async {
     final currentSize = await windowManager.getSize();
     final sharedPreferences = await SharedPreferences.getInstance();
-    logger.d('Saving screen size: $currentSize');
+    logger.d('Saving screen size: ${currentSize.s()}');
     sharedPreferences.setDouble(windowWidthKey, currentSize.width);
     sharedPreferences.setDouble(windowHeightKey, currentSize.height);
   });
@@ -240,13 +241,13 @@ Size? getLastWindowSize(SharedPreferences sharedPreferences, Size? screenSize) {
   final size = lastWindowWidth != null && lastWindowHeight != null
       ? Size(lastWindowWidth, lastWindowHeight)
       : null;
-  logger.d('Got last window size: $size');
+  logger.d('Got last window size: ${size?.s()}');
   if (size == null) return null;
   final clampedSize = Size(
     min(size.width, screenSize?.width ?? size.width),
     min(size.height, screenSize?.height ?? size.height),
   );
-  logger.d('Using clamped window size: $clampedSize');
+  logger.d('Using clamped window size: ${clampedSize.s()}');
   return clampedSize;
 }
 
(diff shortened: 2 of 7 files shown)


levkropp (commented):
LGTM! Multipass now remembers my window size on KDE and resizes based on resolution.

Just checking: What happens if I'm on 1080p and I resize my window to almost fullscreen, and then I change my resolution to something smaller like 1280x720? Does Flutter automatically detect that its trying to open a window size larger than the current resolution or do we need to check for this?

andrei-toterman:
Ah, good point. It does not do any of that. It just uses the last size exactly as it was. We could indeed clamp it by the screen size.

levkropp (commented):
I'm getting behavior on my Framework laptop display (2256x1404) on KDE where if I open Multipass, resize it, and then close it, when I open it again it doesn't remember the window size properly anymore
[framework_doesnt_remember_window_size_levkropp.webm](https://github.com/user-attachments/assets/97f45f64-8b22-4480-a3bc-3b301b0a6c5e)

resetting to the previous commit has the GUI remembering the window size correctly

andrei-toterman:
Interesting. For me it still works fine :thinking: I'll add some logging to help with debugging.

levkropp:
I've tested on my desktop machine which is 1920x1080, Kubuntu 24.10, and I don't get the regression for remembering window size that I did on my Framework. However, the "clamping" does not appear to work: if I resize a window to 1680x1050 when on 1080p then set my display resolution to 1440x900, it opens up a window that is larger than my screen resolution. Additionally, the logs do not provide the actual numbers they just say "Instance of Size" etc, but I'm not sure if that's intended or not on your end
image
(this is my desktop even though the username is 'frame')

levkropp:
The clamping behavior works on my 1080p desktop but the clamping resolution ends up being the resolution of the entire screen so the GUI is cut off vertically. We should apply the better default size logic when clamping as well in my opinion
Screenshot_20241209_142739

The logging revealed the reason why the GUI was not remembering the window size on my Framework laptop: for some reason, flutter thinks my laptop's resolution is half of what it actually is, so when I size it larger than half of the screen it clamps to half of the screen. It correctly remembers the screen size for sizes smaller than 1128x752. We could just have this as a known bug for the framework laptop, or put in unique code for when the screen size 1128x752 is detected.

Screenshot_20241209_143357

andrei-toterman:
Interesting! I see that in your second screenshot, the screen has a `scaleFactor: 2.0`, so I think we need to take that into account when doing the calculation! It may be some fancy HiDPI stuff.
Also, yes, I agree that if the last window size is too big for the current screen, we should just apply the default size instead of this clamping. That makes more sense.

andrei-toterman:
Alright, I looked through the plugin code and I can confirm that `scaleFactor` is related to fancy HiDPI stuff. On most normal monitors, it is 1, but on HiDPI ones it can be more, often 2. I think that also explain why the screen size was behaving weirdly on macs, since I think they have HiDPI screens. Now I need to figure out if we need to divide or multiply by that number :laughing:

andrei-toterman:
@levkropp, apparently on macos this does not seem to be an issue (at least in my case). My mac also has a scaleFactor of 2.0, but the returned screen size matches the one from my system settings. So perhaps this is a Linux only issue? Could you perhaps take over this PR and see if you can make it work on your scaleFactor 2.0 screen? I don't have one, so I can't test it :sweat_smile:

andrei-toterman:
@levkropp just to clarify, did you have access to a windows laptop with a HiDPI screen in order to test this?
@ricab since you mentioned you have a linux laptop with a hidpi screen, could you please open this version of the gui in the terminal and send us the logs, along with the screen resolution that is set in your system settings?

levkropp:
> @levkropp just to clarify, did you have access to a windows laptop with a HiDPI screen in order to test this?

Sorry for not clarifying: I tested my commit on my Framework laptop which has a HiDPI screen running on Ubuntu. I haven't done testing on windows for this but plan on doing so today with my Framework

ricab:
Hey @andrei-toterman, here are the logs I get:

```
2024-12-12 16:57:54.480794 DEBUG Got Screen{frame: Rect.fromLTRB(0.0, 0.0, 3840.0, 2160.0), scaleFactor: 2.0, visibleFrame: Rect.fromLTRB(0.0, 0.0, 3840.0, 2160.0), scaledSize: Size(7680.0, 4320.0)}
2024-12-12 16:57:54.482362 DEBUG Got last window size: null
2024-12-12 16:57:54.482432 DEBUG Computed default window size: Size(1400.0, 822.0)
2024-12-12 16:57:54.482454 DEBUG Using clamped window size: Size(1400.0, 822.0)
2024-12-12 16:57:54.741920 INFO Sent FindRequest{showImages: true, showBlueprints: false}
2024-12-12 16:57:55.017374 INFO Sent VersionRequest{}
2024-12-12 16:57:55.018071 INFO FindRequest{showImages: true, showBlueprints: false} received FindReply{showImages: true, imagesInfo: [{os: Ubuntu, release: Core 16, version: 20200818, aliasesInfo: [{alias: core}, {alias: core16}], codename: Core 16}, {os: Ubuntu, release: Core 18, version: 20211124, aliasesInfo: [{alias: core18}], codename: Core 18}, {os: Ubuntu, release: Core 20, version: 20230119, aliasesInfo: [{alias: core20}], codename: Core 20}, {os: Ubuntu, release: Core 22, version: 20230717, aliasesInfo: [{alias: core22}], codename: Core 22}, {os: Ubuntu, release: Core 24, version: 20240603, aliasesInfo: [{alias: core24}], codename: Core 24}, {os: Ubuntu, release: 20.04 LTS, version: 20241112, aliasesInfo: [{alias: 20.04}, {alias: f}, {alias: focal}], codename: Focal Fossa}, {os: Ubuntu, release: 22.04 LTS, version: 20241206, aliasesInfo: [{alias: 22.04}, {alias: j}, {alias: jammy}], codename: Jammy Jellyfish}, {os: Ubuntu, release: 24.04 LTS, version: 20241119, aliasesInfo: [{alias: 24.04}, {alias: n}, {alias: noble}, {alias: lts}, {alias: default}], codename: Noble Numbat}, {os: Ubuntu, release: 24.10, version: 20241109, aliasesInfo: [{alias: 24.10}, {alias: o}, {alias: oracular}], codename: Oracular Oriole}, {os: Ubuntu, release: 25.04, version: 20241209, aliasesInfo: [{remoteName: daily, alias: 25.04}, {remoteName: daily, alias: p}, {remoteName: daily, alias: plucky}, {remoteName: daily, alias: devel}], codename: Plucky Puffin}, {os: Ubuntu, release: AdGuard Home Appliance, version: 20200812, aliasesInfo: [{remoteName: appliance, alias: adguard-home}], codename: Core 18}, {os: Ubuntu, release: Mosquitto Appliance, version: 20200812, aliasesInfo: [{remoteName: appliance, alias: mosquitto}], codename: Core 18}, {os: Ubuntu, release: Nextcloud Appliance, version: 20200812, aliasesInfo: [{remoteName: appliance, alias: nextcloud}], codename: Core 18}, {os: Ubuntu, release: openHAB Home Appliance, version: 20200812, aliasesInfo: [{remoteName: appliance, alias: openhab}], codename: Core 18}, {os: Ubuntu, release: Plex Media Server Appliance, version: 20200812, aliasesInfo: [{remoteName: appliance, alias: plexmediaserver}], codename: Core 18}]}
2024-12-12 16:57:55.018589 INFO FindRequest{showImages: true, showBlueprints: false} is done
2024-12-12 16:57:55.018960 INFO VersionRequest{} received VersionReply{version: 1.16.0-dev.42.pr3814+g600faa1c2, updateInfo: {}}
2024-12-12 16:57:55.019027 INFO VersionRequest{} is done
2024-12-12 16:57:55.558845 DEBUG Saving screen size: Size(1400.0, 822.0)
2024-12-12 16:58:04.346991 DEBUG Saving screen size: Size(1400.0, 822.0)
2024-12-12 16:58:15.732012 DEBUG Saving screen size: Size(1400.0, 822.0)
2024-12-12 17:03:16.020323 DEBUG Saving screen size: Size(1400.0, 822.0)
```

This with a dual screen setup:

image
image

Notice that I have different scales on the two displays. The primary display is display 2 in that setup, which is where the GUI opened.

andrei-toterman:
Ugh, this is what I feared. In your case, the screen size retrieved by the flutter plugin matches the one you have in your system settings, so we shouldn't multiply by the scale factor here. So, @levkropp, in the best case, the disparity is present only on the framework laptop, or in the worst case, it just depends on some unknown stuff for different monitors. So, IMO, since we can't be sure that the screen size we get is the same one as specified in the system settings, we should take it as it is and hope that it's good. Again, on mac it also returns the right one without having to multiply by the scale factor. And apparently on Linux it can return the right size as well. And I have a hunch that on Windows it will also vary from monitor to monitor. Opinions?

Also, thanks @ricab for checking!

andrei-toterman:
Well, the plugin calls gdk_monitor_get_geometry. So maybe :shrug:
KDE is using Qt underneath, so @levkropp what does the following give you
```python
from PyQt5.QtWidgets import QApplication
from PyQt5.QtGui import QGuiApplication
import sys

app = QApplication(sys.argv)
screen = QGuiApplication.primaryScreen()
print(f"{screen.geometry()}")
```

levkropp:
image
It looks like the issue might just be with KDE/Qt handling the screen on the Framework laptop. On GNOME, the GUI reports the proper resolution as well as a scaleFactor of 1.0

andrei-toterman:
Well, the behavior we get is just too unpredictable to make any good decision. My suggestion is that we trust the window size that we get as it is. Apparently we can't really know for sure if it's the right one, but it sometimes is, so what else could we do?

I've also been playing with some other apps (Dolphin on KDE) to see how they react. And what they do is they remember the last window size, per resolution. If I am on one resolution and set the window to be really wide, the app will remember that size when I open it again. If I go to another resolution and make the app really thin, it will remember that. But if I go back to the previous resolution, it will open the app very wide again, as it was before. So it remembers the last window size for each resolution, and if there is no last window size for a resolution, it opens it to a default size that fits.
And perhaps we could add this for us as well. Opinions?

levkropp:
> the behavior we get is just too unpredictable to make any good decision. My suggestion is that we trust the window size that we get as it is.

I agree with this. The screen size we get from the flutter plugin has been correct in all of our testing except specifically the Framework Laptop on specifically KDE, which appears to be from how Qt handles screen size and scaling

>it remembers the last window size for each resolution, and if there is no last window size for a resolution, it opens it to a default size that fits. And perhaps we could add this for us as well. Opinions?

This would be useful for people who connect their laptops to external monitors e.g. in an office. If it is not too much work to implement/maintain then I am all for it

levkropp (commented):
Testing it again and getting weird behavior both on my desktop and framework laptop, both on KDE :/
not saving window size
on my desktop (above) it was not remembering the window size even though it said "saving window size" and remembered a previous size

Screenshot_20241216_190526(1)
on my framework laptop it was not clamping when reading the resolution on the normal size screen resolution (when before it would glitch as shown in the video in [this comment](https://github.com/canonical/multipass/pull/3814#pullrequestreview-2483018929)) and it seemed to forget the last window size entirely when changing the screen resolution in the display options. We've already discussed how we shouldn't cater to the unique situation of the framework laptop, and if these issues that I am seeing on my desktop aren't reproducible on your or @ricab's machines then it's probably safe to assume that it is just an issue with my setup or hardware or display drivers or configuration etc

levkropp (approved):
Oh I see, so in fact the screenshots show that the correct behavior is being applied in all cases 😁 In that case everything looks good to me!

Sploder12 (commented):
The code looks good, but there is some _interesting_ behavior that I noticed that might be related but I have no clue. Some seem to be Flutter bugs.

It seems sometimes the window opens on one monitor, the size is calculated, and then the window moves to the other monitor (very quickly). This seems to be the case without these changes but it's hard to tell without the logging.

Which monitor the window opens on is extremely inconsistent. (not new)

The size is saved on window focus and unfocus, it seems the window losing/gaining a shadow triggers a resize event but the reported size is the same.

`minimumSize` is sometimes ignored. `minimumSize` always breaks when the window spawns on multiple monitors. 

`center` seems to center the framebuffer (not the window), but doesn't work most of the time.

It seems resizing the window to the minimum will save the size as being smaller than the minimum.
 
Manually setting the saved width/height disregards minimumSize (unless maybe this is caused by above?).

It is possible to make the saved size much larger than the size of the window by having the window span multiple screens. (But maybe that's fine)

andrei-toterman:
Hey, @Sploder12! Thanks for the in-depth feedback! I looked some more into the `window_manager` plugin that we are using and some of those issues are unfortunately caused by its implementation. IIUC, the issue with the minimum window size comes from the fact that they are not using the GTK API as they should (didn't check for Windows and macOS). I think the issue of the app not opening consistently on the correct display comes from the fact that we want to center the window when it starts, and the plugin gets confused about what is the screen on which it should center. WDYT if we didn't center the app, but also remembered its last position on screen? Or what if we didn't do anything to it and just let it open as it wants? What do other apps that you use do?

Sploder12 (approved):
Since the issues seem to be unrelated to these changes, LGTM!

Other apps on Ubuntu tend to open in random, but consistent per application, locations on the monitor I open them on. Sometimes it's top-left, sometimes it's center, sometimes they remember their last location. But fairly inconsistent overall.

Most Windows windows tend to save location and size. But on occasion an app will open centered or in some fixed random location.

Since the centering is terribly broken as is, it'd probably be better to let the OS handle it. Saving the position could be a nice change too though (for a later PR).

PR #3825, opened 2024-12-09
Terminal zoom

andrei-toterman:
This PR adds the ability to change the font size in the GUI terminals, using common shortcuts like `Ctrl`+`+` to increase, `Ctrl`+`-` to decrease and `Ctrl`+`0` to reset. `Command` is used instead of `Ctrl` on macOS.

It also include two small, drive-by fixes, which change the deprecated `MaterialStateColor` with the newer `WidgetStateColor`. Those were done because they were in the same files as the other, relevant changes.

MULTI-1678

### DIFF
--- a/src/client/gui/lib/platform/linux.dart
+++ b/src/client/gui/lib/platform/linux.dart
@@ -35,7 +35,15 @@ class LinuxPlatform extends MpPlatform {
             PasteTextIntent(SelectionChangedCause.keyboard),
         SingleActivator(LogicalKeyboardKey.equal, control: true):
             IncreaseTerminalFontIntent(),
+        SingleActivator(LogicalKeyboardKey.equal, control: true, shift: true):
+            IncreaseTerminalFontIntent(),
+        SingleActivator(LogicalKeyboardKey.add, control: true, shift: true):
+            IncreaseTerminalFontIntent(),
+        SingleActivator(LogicalKeyboardKey.numpadAdd, control: true):
+            IncreaseTerminalFontIntent(),
         SingleActivator(LogicalKeyboardKey.minus, control: true):
+            DecreaseTerminalFontIntent(),
+        SingleActivator(LogicalKeyboardKey.numpadSubtract, control: true):
             DecreaseTerminalFontIntent(),
         SingleActivator(LogicalKeyboardKey.digit0, control: true):
             ResetTerminalFontIntent(),
--- a/src/client/gui/lib/platform/linux.dart
+++ b/src/client/gui/lib/platform/linux.dart
@@ -4,6 +4,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
 import '../settings/autostart_notifiers.dart';
+import '../vm_details/terminal.dart';
 import 'platform.dart';
 
 class LinuxPlatform extends MpPlatform {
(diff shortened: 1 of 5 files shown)


src/client/gui/lib/platform/linux.dart
```diff
@@ -32,6 +33,12 @@ class LinuxPlatform extends MpPlatform {
             CopySelectionTextIntent.copy,
         SingleActivator(LogicalKeyboardKey.keyV, control: true, shift: true):
             PasteTextIntent(SelectionChangedCause.keyboard),
+        SingleActivator(LogicalKeyboardKey.equal, control: true):
+            IncreaseTerminalFontIntent(),
```
levkropp, line 37: These 4 intents exist for each platform. Maybe it would be better to define default terminal shortcuts in mpPlatform?

```
  // Define default terminal shortcuts (platform.dart)
  Map<SingleActivator, Intent> get terminalShortcuts => {
        SingleActivator(LogicalKeyboardKey.keyC, control: true): CopySelectionTextIntent.copy,
        SingleActivator(LogicalKeyboardKey.keyV, control: true, shift: true): PasteTextIntent(SelectionChangedCause.keyboard),
        SingleActivator(LogicalKeyboardKey.equal, control: true): IncreaseTerminalFontIntent(),
        SingleActivator(LogicalKeyboardKey.minus, control: true): DecreaseTerminalFontIntent(),
        SingleActivator(LogicalKeyboardKey.digit0, control: true): ResetTerminalFontIntent(),
      };
```

levkropp (dismissed):
LGTM! Just one suggestion regarding DRY

src/client/gui/lib/platform/linux.dart
andrei-toterman: Well, they are not really the same, because macos uses Command (`meta` in this case) instead of Control for those shortcuts. And the fact that linux and windows share the shortcuts is more of a coincidence than a duplication. Since changing one does not imply that the other has to change as well, it is not duplication IMO.

src/client/gui/lib/platform/linux.dart
```diff
@@ -32,6 +33,12 @@ class LinuxPlatform extends MpPlatform {
             CopySelectionTextIntent.copy,
         SingleActivator(LogicalKeyboardKey.keyV, control: true, shift: true):
             PasteTextIntent(SelectionChangedCause.keyboard),
+        SingleActivator(LogicalKeyboardKey.equal, control: true):
```
Sploder12, line 36: I was going to say this should be `LogicalKeyboardKey.plus` (which doesn't exist) since this breaks for certain keyboard layouts, but it seems that Flutter doesn't provide an easy way to handle this. Do you think this is worth fixing?

Example broken keyboard layout:
1. Set your keyboard layout to Japanese
2. Notice `+` and `=` are on separate keys
3. Notice `ctrl`+`+` and `ctrl`+`=`, don't work since shift is pressed

Sploder12 (commented):
Looks mostly good to me! Just a question/issue about alternative keyboard layouts.

src/client/gui/lib/platform/linux.dart
andrei-toterman: Good catch! Yes, I do think it is worth fixing. I checked other terminals and they handle it correctly, so we should as well.

andrei-toterman: Alright, so from my testing it seems that other apps handle both `+` and `=` as zoom increase, even if they are on separate keys and both require shift. And also they seem to handle numpad plus and minus. So I added those.

ricab: Well spotted, this would never have occurred to me. Good job guys!

Sploder12 (approved):
LGTM! Seems i missed `LogicalKeyboardKey.add`, oops. Good job adding the numpad support too!

Sploder12 (approved):
LGTM!

PR #3826, opened 2024-12-09
Fix console creation

### LINKED ISSUE #3788: CLI client tests overwrite console output settings on Windows
**Describe the bug**
Running tests from `test_cli_client` on Windows will overwrite and not restore the console output settings. This causes all output during and after the test to not appear.

**To Reproduce**
1. Run tests on Windows
2. Observe no tests after `execCmdWithDirPrependsCd` output to the console

**Expected behavior**
Later tests output their result.

**Additional info**
 - OS: Windows 10 Home/Pro

This seems to not be an issue in CI

Sploder12:
Fixes #3788 

Previously the test output would disappear since `StubTerminal*` was being cast to `WindowsTerminal*` and then `cin`/`cout`/`cerr` would get deleted. Now `StubTerminal*` is not cast at all and `Console` creation is handled by their respective Terminals.

MULTI-1676

### DIFF
--- a/src/client/cli/cmd/shell.cpp
+++ b/src/client/cli/cmd/shell.cpp
@@ -74,7 +74,7 @@ mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)
 
         try
         {
-            auto console_creator = [this](auto channel) { return Console::make_console(channel, term); };
+            auto console_creator = [this](auto channel) { return term->make_console(channel); };
             mp::SSHClient ssh_client{host, port, username, priv_key_blob, console_creator};
             ssh_client.connect();
         }
--- a/tests/stub_terminal.h
+++ b/tests/stub_terminal.h
@@ -20,6 +20,8 @@
 
 #include <multipass/terminal.h>
 
+#include "stub_console.h"
+
 namespace multipass
 {
 namespace test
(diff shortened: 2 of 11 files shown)


tests/stub_terminal.h
```diff
@@ -61,10 +63,26 @@ class StubTerminal : public multipass::Terminal
...
+
+    ConsolePtr make_console(ssh_channel channel) override
+    {
+        if (!console)
+        {
+            return std::make_unique<StubConsole>();
+        }
+
+        return std::exchange(console, nullptr);
+    }
```
ricab, line 79: Hmm, this smells a little of over-engineering :nose: Is `set_console` even used anywhere? If it ever was, wouldn't a mock be more appropriate than a stub? Why don't you simplify to just this?

```suggestion
    ConsolePtr make_console(ssh_channel channel) override
    {
        return std::make_unique<StubConsole>();
    }
```

src/client/cli/cmd/shell.cpp
```diff
@@ -74,7 +74,7 @@ mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)
 
         try
         {
-            auto console_creator = [this](auto channel) { return Console::make_console(channel, term); };
+            auto console_creator = [this](auto channel) { return term->make_console(channel); };
```
ricab, line 77: It shouldn't be hard to get this covered. Do you want to give it a shot?

ricab (commented):
Thanks, this is better indeed :smile: I have only a couple of requests.

tests/stub_terminal.h
Sploder12: Good point

ricab (approved):
Great, thanks Trevor!

sharder996 (approved):
LGTM

ricab:
This needs the other side, so merging manually.

PR #3828, opened 2024-12-09
Escape mount sources and targets

### LINKED ISSUE #3759: Mount targets are not validated
**Describe the bug**
The mount target is not checked/validated to be a valid path. This allows for _interesting_ targets to be used.

**To Reproduce**
1. launch multipass GUI
2. create a mount with a valid source but invalid target such as `" & mkdir hello & "`
3. observe mount fails for some reason, that may not be "bad mount point"
4. shell into the instance
5. observe `~/hello` exists

Also works in CLI but less interesting.

**Expected behavior**
The daemon prevents the usage of such targets without allowing side effects.

**Additional context**
Mount source is verified to be an existing file but, it might make sense to also validate the source as well. Windows paths are important to consider for both source and target since `C:\foo\` is a valid path.

Sploder12:
Fixes #3759 

Previously, mount targets and sources were not escaped. This caused strange targets and mounts like `" & mkdir hello & "` to be able to execute commands in the instance. Now they are properly escaped so strangely named directories like `" & mkdir hello & "` will work as source or target. 

MULTI-1675

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -118,7 +118,6 @@ Str&& trim(Str&& s, Filter&& filter);
 template <typename Str>
 Str&& trim(Str&& s);
 std::string& trim_newline(std::string& s);
-std::string escape_char(const std::string& s, char c);
 std::string escape_for_shell(const std::string& s);
 std::vector<std::string> split(const std::string& string, const std::string& delimiter);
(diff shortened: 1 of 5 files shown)


andrei-toterman (approved):
Well that was a horrible bug. :clap::clap::clap: for @Sploder12!

sharder996 (approved):

PR #3830, opened 2024-12-10
`Open shell` button starts the VM if needed

andrei-toterman:
Now, if the VM is in a state from which it can be started (`Stopped` or `Suspended`), the `Open shell` button will be enabled and it will start the VM before shelling into it.

MULTI-1681


Sploder12 (approved):
LGTM! Tests pass and functions as expected

levkropp (dismissed):

Sploder12 (approved):
LGTM!

PR #3831, opened 2024-12-10
make some files that were using crlf use just lf

andrei-toterman:
Those files mess with the new whitespace checker, reporting CR as trailing whitespace. All the other files in our repo seem to be using plain LF, so these were just some accidents.


ricab (approved):
Yup, thanks!

ricab:
Nothing like this on the private side?

andrei-toterman:
I checked but couldn't find anything.

PR #3835, opened 2024-12-12
[gui] ask before closing running terminal

andrei-toterman:
If a terminal is running, show a confirmation dialog that tells the user the state of the terminal will be lost.
The user can also choose to not be asked again about this.

There is also a drive-by fix of removing the `onClose` callback on the dialog shown before quitting Multipass, which was no longer needed due to an earlier refactoring.

MULTI-1682


ricab:
Was this designed? I am not used to terminals asking for confirmation to close, unless there are children processes running.

levkropp (approved):
LGTM! Tested and working on my machine

src/client/gui/lib/close_terminal_dialog.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return ConfirmationDialog(
+      title: 'Are you sure you want to close this terminal?',
+      body: Column(crossAxisAlignment: CrossAxisAlignment.start, children: [
+        const Text('Its current state will be lost.'),
+        const SizedBox(height: 12),
+        Switch(
+          value: doNotAsk,
+          label: 'Do not ask me again.',
```
ricab, line 34: For another time: I would love to be able to click on the label/area before a switch to have it toggled (not have to place the mouse exactly on top of th widget). This would apply to other switches, check boxes... radio buttons when enabling... WDYT? Does Flutter have a mechanism to achieve that easily?

src/client/gui/lib/vm_details/terminal_tabs.dart
```diff
@@ -144,7 +148,32 @@ class TerminalTabs extends ConsumerWidget {
           title: 'Shell ${shellId.id}',
           selected: index == currentIndex,
           onTap: () => ref.read(notifier).setCurrent(index),
-          onClose: () => ref.read(notifier).remove(index),
+          onClose: () {
```
ricab, line 151: Should we ask also when closing the entire GUI? I know we already ask about stopping instances, so perhaps there's a way we could fold this question in there somehow... Something like changing the text of that last popup to also warn about open terminals and offer a cancel button?

And what about "stopping" an instance? Should we ask there too?

ricab (commented):
Thanks @andrei-toterman, this looks good and works well. Only one matter I would like to clarify before closing, along with a side suggestion for later.

src/client/gui/lib/close_terminal_dialog.dart
andrei-toterman: Yes, that can be easily achieved. And in places like this popup, it would indeed make a lot of sense.
But I'm not sure about the settings page, because there you also have dropdowns and text fields, for which I don't know if this pattern makes sense, and having different actionable areas based on the type of setting would also not be nice. Of course, we can also make this aspect configurable, and have places where the whole label is clickable and places where its not, but then that would be introducing an inconsistency for the sake of removing another, so I don't know...

src/client/gui/lib/vm_details/terminal_tabs.dart
andrei-toterman: Hmm, when closing the entire GUI, what should we do with the 'Do not ask again' switch? Show it under the 'Keep instances running' option? And how would a user indicate that they want to close their terminals, but keep the instances running in the background? Should we show the popups one after the other?

About having the popup when you want to stop an instance, yes, I think we can have that.

src/client/gui/lib/close_terminal_dialog.dart
ricab: Hmm I see what you mean. I think I am used to being able to click on checkbox labels, not sure about the others. Maybe in the settings we'd just need a bigger switch button. Anyway, just something to keep in the back of our heads for later I guess.

src/client/gui/lib/vm_details/terminal_tabs.dart
ricab: Hmm, but do you think it makes sense to ask to close each terminal but not when closing all terminals? It feels a bit like double standards to me.

Perhaps we could add just a notice that there were running terminals that would be closed. The question could remain about stopping instances, such that the "do not ask again" kept referring to it. 

Perhaps something like [this](https://excalidraw.com/#json=H2sfIpcV4kUQDbJgoK5FG,sQVAlwJ-jRh0gTngfbI1hA): 
image

The text could be improved. Suggestions welcome, of course.

And the cancel button is actually already there in the form of a cross... I wonder if it should be a proper button too, regardless of terminals, but that is a separate matter.

src/client/gui/lib/close_terminal_dialog.dart
ricab, line 34: I just noticed: the exit dialog says "Remember this setting". Shouldn't we say the same here?

andrei-toterman: I think the semantics do not match entirely. For example:
There is a terminal running. I click 'X' on its tab. The pop-up comes up and says 'Are you sure you want to close this terminal?'. I check 'Remember this setting' and then click 'No', so that the terminal does not get closed. Now I click 'X' again. What is the setting that should be remembered? That I do not want to close this terminal?
If the label says 'Do not ask me again', I think it is more clear that the 'X' button will close the terminal, and it will either ask me for confirmation or it will not ask me for confirmation.
In the exit dialog, the setting that we remember is more tied to what to do with the instances. In the terminal dialog, what we remember is not whether or not the terminal should be closed, but whether or not we should ask the user for confirmation.
WDYT?

andrei-toterman: One solution that perhaps would be more clear, but less pretty would be to not have that switch, and have 3 buttons for the following actions:
- Don't close the terminal, then always ask me again when I want to close it again
- Close the terminal, then always ask me again when I want to close it again
- Close the terminal, then do not ask me again when I want to close it again

Right now, we permit the action of 'Don't close the terminal, but then do not ask me again when I want to close it again', which IMO is a bit weird.

andrei-toterman: I think they are not the same thing.
If we want to do the same thing from now on, and when asked if I am sure that I want to close the terminal, I say 'No' + 'Remember this setting', what should we do from now on?

To clarify how I understand the situation:
In the case of the exit dialog, the question asked is: what to do with running instances? And it has two answers: stop them or leave them running. Either of those answers can be chosen manually or automatically (if remembered).

In the case of the terminal dialog, the question asked is: should the terminal be closed? And it has the answers 'Yes' and 'No'. Both of those answers can be chosen manually, but I think only 'Yes' should be able to be chosen automatically. Choosing 'No' automatically would mean that the 'X' button on a running terminal will become useless.

ricab: Or maybe two radio buttons, such that the second one only shows when the user selects close on the first one. 

Just throwing ideas up. Maybe we just merge now and do a new design iteration for this. WDYT?

ricab: But 3 buttons would do I guess: "Don't close", "close", "always close". Perhaps with tooltips...

ricab: Bah, crossed messages again. We should chat...

andrei-toterman: Welp, I just tested this in Konsole, perhaps you can test in the Gnome terminal as well (or any terminal, like Alacritty). I had a running process, I clicked 'X', I got the pop-up, I checked 'Do not ask again', then clicked 'Cancel'. The terminal stays open, but then whenever I click 'X' again, I do not get the popup and the terminal does not close, making the X button useless. I could only fix this by going in the settings and making the pop-up show up again. IMO this behavior is not nice at all, and is precisely what I was trying to avoid in our case.

ricab: Alacritty just didn't ask. Gnome-terminal always asks (no "don't ask me again").

src/client/gui/lib/vm_details/terminal_tabs.dart
ricab: For the record, as discussed elsewhere...

The proposal above would still have the issue that, if the user selected "remember this setting" here but "always ask" on the terminal question, terminals would still be closed with no question or indication when closing the GUI.

As for asking when stopping, we'd need to decide what to do when batch closing.

PR #3844, opened 2024-12-16
[ci] make renovate ask for users instead of team for reviews

andrei-toterman:
Apparently, when asking for a team, it asks for the whole team, not just users within that team. The good news is that now we can specify exactly which users from the team qualify for reviews. The bad news is that we must update this list as part of the onboarding/offboarding process :shrug:


ricab (approved):
LGTM, thanks.

PR #3846, opened 2024-12-16
Refactor image vault utils

Sploder12:
This PR refactors several functions related to image vault utilities. All util functions now have unit test coverage and can be mocked to ease testing for components that use the utilities. Some functions have been removed or moved to file_ops since they already exist or fit better there.

MULTI-1653

### DIFF
--- a/include/multipass/vm_image_vault_utils.h
+++ b/include/multipass/vm_image_vault_utils.h
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_VM_IMAGE_VAULT_UTILS_H
+#define MULTIPASS_VM_IMAGE_VAULT_UTILS_H
+
+#include "file_ops.h"
+#include "xz_image_decoder.h"
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#define MP_IMAGE_VAULT_UTILS                                                                                           \
+    multipass::ImageVaultUtils                                                                                         \
+    {                                                                                                                  \
+    }
+
+namespace multipass
+{
+class VMImageHost;
+
+class ImageVaultUtils
+{
+public:
+    using DefaultDecoder = XzImageDecoder;
+
+    static QString copy_to_dir(const QString& file, const QDir& output_dir);
+    static QString compute_hash(QIODevice& device);
+    static QString compute_file_hash(const QString& file);
+
+    template <class Hasher = ImageVaultUtils>
+    static void verify_file_hash(const QString& file, const QString& hash, const Hasher& hasher = ImageVaultUtils{});
+
+    template <class Decoder = DefaultDecoder>
+    static QString extract_file(const QString& file,
+                                const ProgressMonitor& monitor,
+                                bool delete_original = false,
+                                const Decoder& decoder = DefaultDecoder{});
+
+    using HostMap = std::unordered_map<std::string, VMImageHost*>;
+    using Hosts = std::vector<VMImageHost*>;
+
+    static HostMap configure_image_host_map(const Hosts& image_hosts);
+};
+
+template <class Hasher>
+void ImageVaultUtils::verify_file_hash(const QString& file, const QString& hash, const Hasher& hasher)
+{
+    auto file_hash = hasher.compute_file_hash(file);
+}
+
+template <class Decoder>
+QString ImageVaultUtils::extract_file(const QString& file,
+                                      const ProgressMonitor& monitor,
+                                      bool delete_original,
+                                      const Decoder& decoder)
+{
+    auto new_path = MP_FILEOPS.remove_extension(file);
+    decoder.decode_to(file, new_path, monitor);
+
+    if (delete_original)
+    {
+        QFile qfile{file};
+        MP_FILEOPS.remove(qfile);
+    }
+
+    return new_path;
+}
+
+} // namespace multipass
+
+#endif // MULTIPASS_VM_IMAGE_VAULT_UTILS_H
--- a/include/multipass/vm_image_vault_utils.h
+++ b/include/multipass/vm_image_vault_utils.h
@@ -25,10 +25,7 @@
 #include <unordered_map>
 #include <vector>
 
-#define MP_IMAGE_VAULT_UTILS                                                                                           \
-    multipass::ImageVaultUtils                                                                                         \
-    {                                                                                                                  \
-    }
+#define MP_IMAGE_VAULT_UTILS multipass::ImageVaultUtils()
 
 namespace multipass
 {
@@ -62,6 +59,9 @@ template <class Hasher>
 void ImageVaultUtils::verify_file_hash(const QString& file, const QString& hash, const Hasher& hasher)
 {
     auto file_hash = hasher.compute_file_hash(file);
+
+    if (file_hash != hash)
+        throw std::runtime_error(fmt::format("Hash of {} does not match {}", file, hash));
 }
 
 template <class Decoder>
--- a/include/multipass/vm_image_vault_utils.h
+++ b/include/multipass/vm_image_vault_utils.h
@@ -21,65 +21,56 @@
 #include "file_ops.h"
 #include "xz_image_decoder.h"
 
+#include <functional>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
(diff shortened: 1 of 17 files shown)


tests/test_image_vault_utils.cpp
```diff
@@ -83,7 +83,7 @@ TEST_F(TestImageVaultUtils, copy_to_dir_copys_to_dir)
 TEST_F(TestImageVaultUtils, compute_hash_throws_when_cant_read)
 {
     QBuffer buffer{}; // note: buffer is not opened
-    MP_EXPECT_THROW_THAT(MP_IMAGE_VAULT_UTILS.compute_hash(buffer),
+    MP_EXPECT_THROW_THAT(auto _ = MP_IMAGE_VAULT_UTILS.compute_hash(buffer),
```
ricab, line 86: You could use `std::ignore` for this, but really, isn't this a signal that the function should not be `[nodiscard]`? I know recommendations sometimes ignore exceptions (including clang-tidy), but we don't need to :slightly_smiling_face:

IMO, if a function returns and throws, and if it is a reasonable use for the client to just call to check for validity, then it shouldn't be nodiscard.

Sploder12: `std::ignore` would be good here. I agree with your general opinion but for this specific case I'm going to disagree. I don't think it's reasonable for a user to call this function without caring about the return. There are much much much better ways to check validity, and I'm not sure why you'd only check validity (other than for the sake of testing).

ricab: OK, makes sense.

src/utils/file_ops.cpp
```diff
@@ -170,6 +170,18 @@ bool mp::FileOps::flush(QFile& file) const
     return file.flush();
 }
 
+QString mp::FileOps::remove_extension(const QString& path) const
```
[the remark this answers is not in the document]
sharder996: or `QFileInfo::completeBaseName()`

sharder996 (changes requested):
Looks good @Sploder12 

Thanks for tackling this. This has needed to be cleaned up for a while.

src/utils/file_ops.cpp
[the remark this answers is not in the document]
Sploder12: It'd also have to use `QFileInfo::dir` and `QDir::fileName` since base name doesn't preserve the path. But `QFileInfo::dir` has some unwanted behavior for an empty path so it'd end up being more code to handle that edge-case. WDYT?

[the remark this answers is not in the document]
sharder996: What about using the std::filesystem? Otherwise, I guess its ok, just a bit nasty with the explicit string manipulation.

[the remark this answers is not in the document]
Sploder12: `std::filesystem` would be much better in almost every way but requires a conversion from `QString` to `path`. Sounds good to me!

sharder996 (approved):

tests/mock_image_decoder.h
```diff
@@ -0,0 +1,37 @@
...
+#include "common.h"
+
+#include <multipass/path.h>
+#include <multipass/progress_monitor.h>
+
+namespace multipass::test
+{
+
+class MockImageDecoder
+{
```
georgeliao, line 30: Nice standalone mock class only for template parameter use.

src/utils/vm_image_vault_utils.cpp
```diff
@@ -27,79 +28,72 @@
...
+    const auto file_hash = compute_file_hash(file);
+
+    if (file_hash != hash)
+        throw std::runtime_error(fmt::format("Hash of {} does not match {}", file, hash));
+}
 
-    new_image_path.remove(".xz");
+QString mp::ImageVaultUtils::extract_file(const QString& file, const Decoder& decoder, bool delete_original) const
+{
+    const auto fs_new_path = MP_FILEOPS.remove_extension(file.toStdU16String());
```
georgeliao, line 80: Why exactly do we need the `toStdU16String` function as opposed to `toStdString`? I thought the `toStdString` checks the value type and return the corresponding string so it should be safe on windows, maybe I have missed something.  Same question goes to `fs_new_path.u8string()` called later.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -356,7 +359,9 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type,
                                        last_modified.toString(),
                                        0,
                                        checksum.has_value()};
-                const auto image_filename = mp::vault::filename_for(image_url.path());
+
+                QFileInfo file_info{image_url.path()};
+                const auto image_filename = file_info.fileName();
```
georgeliao, line 364: Can we do one liner `const auto image_filename = QFileInfo{image_url.path()}.fileName();`? Which is cleaner. Same goes to other occurrences of this pattern in the PR.

src/utils/vm_image_vault_utils.cpp
```diff
@@ -27,79 +28,72 @@
...
+    if (!MP_FILEOPS.exists(info))
+        throw std::runtime_error(fmt::format("File {} not found", file));
 
-    QFileInfo info{file_name};
-    const auto source_name = info.fileName();
-    auto new_path = output_dir.filePath(source_name);
-    QFile::copy(file_name, new_path);
-    return new_path;
-}
+    auto new_location = output_dir.filePath(info.fileName());
```
georgeliao, line 45: A small `const` maybe can be added to the `auto new_location`  variable.

georgeliao (commented):
@Sploder12 
Well done, I only have a few minor comments and questions.

tests/test_image_vault_utils.cpp
```diff
@@ -0,0 +1,236 @@
...
+namespace mpt = multipass::test;
+
+using namespace testing;
+
+namespace
+{
+
+struct TestImageVaultUtils : public ::testing::Test
+{
+    const mpt::MockFileOps::GuardedMock mock_file_ops_guard{StrictMock<mpt::MockFileOps>::inject()};
```
georgeliao, line 38: About this syntax, there are `mpt::MockFileOps::inject<StrictMock>():` vs `StrictMock<mpt::MockFileOps>::inject()`, I do not think the second one does what it supposed to do, because I still see the warning from the unit test and the warning did not fail the unit test. The first one seems to work, probably this is related to the implementation of `inject` method which is multipass code instead of gmock code. Besides that, I feel we should use `NiceMock` here to omit the uninteresting call warning.

include/multipass/vm_image_vault_utils.h
```diff
@@ -0,0 +1,78 @@
...
+class ImageVaultUtils : public Singleton<ImageVaultUtils>
+{
+public:
+    ImageVaultUtils(const PrivatePass&) noexcept;
+
+    using Decoder = std::function<void(const QString&, const QString&)>;
+    using DefaultDecoderT = XzImageDecoder;
+
+    virtual QString copy_to_dir(const QString& file, const QDir& output_dir) const;
+    [[nodiscard]] virtual QString compute_hash(QIODevice& device) const;
```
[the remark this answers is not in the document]
Sploder12: It could be made protected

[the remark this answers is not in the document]
georgeliao: Anyway, I do not have a strong opinion on this. Feel free to either way.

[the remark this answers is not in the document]
Sploder12: My reasoning for making it was to make `compute_file_hash` easier to unit test, since otherwise mocking would require a wrapper around the Qt hashing functions. Or it could use a real file for testing, but that tends to cause issues in CI so I wanted to avoid it.

src/daemon/default_vm_image_vault.cpp
Sploder12: I did so for where I could. The places with `remove` need an lvalue reference so a one liner won't work unless we change the API for the fileop.

tests/test_image_vault_utils.cpp
Sploder12: huh, I'm a bit surprised that doesn't work, odd. But yeah, using `NiceMock` makes more sense.

src/utils/vm_image_vault_utils.cpp
georgeliao: Another wondering is the `QString::toStdU16String`'s behavior on linux, `std::u16string` is the input string to construct `fs::path`, will that be safe considering linux use UTF-8?

Sploder12: The path constructor overload for `std::u16string` says "conversion from UTF-16 to native filesystem encoding is used" so it should be safe. While the `std::string` constructor says "is assumed to be the native narrow encoding". This whole issue is probably why they added the `std::u8string` overload in C++20 which has the same spec as `std::u16string`, but with UTF-8

include/multipass/vm_image_vault_utils.h
```diff
@@ -0,0 +1,78 @@
...
+    [[nodiscard]] virtual QString compute_file_hash(const QString& file) const;
+
+    virtual void verify_file_hash(const QString& file, const QString& hash) const;
+
+    virtual QString extract_file(const QString& file, const Decoder& decoder, bool delete_original = false) const;
```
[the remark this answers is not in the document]
Sploder12: Same response, in this case `const Decoder& decoder` (a function object) is used as an injection point.

georgeliao (approved):

sharder996 (approved):
+1 from me!

PR #3847, opened 2024-12-17
Unix drive-by tweaks

ricab:
Small drive by fixes that I noticed while reviewing #3735 (which this is based on).

### DIFF
--- a/src/platform/platform_unix.cpp
+++ b/src/platform/platform_unix.cpp
@@ -199,7 +199,9 @@ sigset_t mp::platform::make_sigset(const std::vector<int>& sigs)
 sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
 {
     auto sigset{make_sigset(sigs)};
-    MP_POSIX_SIGNAL.pthread_sigmask(SIG_BLOCK, &sigset, nullptr);
+    if (const auto ec = MP_POSIX_SIGNAL.pthread_sigmask(SIG_BLOCK, &sigset, nullptr); ec)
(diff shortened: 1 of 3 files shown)


sharder996 (approved):
✅

src/platform/platform_unix.cpp
```diff
@@ -198,8 +199,10 @@ sigset_t mp::platform::make_sigset(const std::vector<int>& sigs)
 
 sigset_t mp::platform::make_and_block_signals(const std::vector<int>& sigs)
 {
-    auto sigset{make_sigset(sigs)};
-    MP_POSIX_SIGNAL.pthread_sigmask(SIG_BLOCK, &sigset, nullptr);
+    const auto sigset{make_sigset(sigs)};
+    if (const auto ec = MP_POSIX_SIGNAL.pthread_sigmask(SIG_BLOCK, &sigset, nullptr); ec)
+        throw std::runtime_error(fmt::format("Failed to block signals: {}", strerror(ec)));
```
levkropp, line 204: Codecov says that this line is not covered by tests. Is this function tested in the current test suite?

levkropp (commented):
Very good! Builds and all tests pass. Just one question about test coverage

src/platform/platform_unix.cpp
ricab: No, I meant to do that indeed.

ricab: Should be good now @levkropp, thanks for pointing it out.

sharder996 (approved):

ricab:
Heh, it's a tiny thing, but thanks 😄

PR #3850, opened 2024-12-17
Slight updates to the security doc

ricab:


giuliazanchi (approved):
LGTM

ricab:
Thank you for your review @giuliazanchi! This is a very small PR so I think we can go without secondary review, unless you would prefer otherwise. 

If you're alright with it, you can click "Merge when ready". Please let me know if you have any question.

PR #3852, opened 2024-12-18
use SelectableText in GUI for instance info

### LINKED ISSUE #3821: Make text selectable
Currently it's not possible to select and copy text in GUI. Main areas of concern are the All instances table (names, state, image, public & private IP addresses), the details page (same items as in the table + created, uptime, CPU, memory, disk, bridged network status, mounts).

Make all values listed above selectable. I **cannot** think of a strong use case for making headers, table titles and catalogue cards selectable too. So, only the values.

**Additional context**
We can keep the default actions 'copy' and 'select all', as long as 'select all' acts within a single value on the page. 
Let me know if I can provide more input, in case any of the defaults in the SelectableText don't work.

levkropp:
closes #3821 

This PR changes most `Text` widgets in the Instances and VM Details pages to `SelectableText`
`SelectableText` does not support `overflow: ellipsis` so I have set `maxLines: 1` to ensure that table cells are still 1 line max.

I did not make the instance's name selectable in the Instances page because the name column has unique behavior implemented with `VMNameLink` where clicking it links to the instance's shell. Instead, the name of the instance on the Instance details/shell page is now selectable.

[Screencast_20250115_131745.webm](https://github.com/user-attachments/assets/b6780eb4-f88b-43c9-b30e-9da6c20b04c2)

### DIFF
--- a/src/client/gui/lib/selectable_text.dart
+++ b/src/client/gui/lib/selectable_text.dart
@@ -1,0 +1,43 @@
+import 'package:flutter/material.dart' as fl;
+
+class SelectableText extends fl.StatelessWidget {
+  final String text;
+  final fl.TextStyle? style;
+  final fl.TextAlign? textAlign;
+  final int? maxLines;
+
+  const SelectableText(
+      this.text, {
+        super.key,
+        this.style,
+        this.textAlign,
+        this.maxLines,
+      });
+
+  @override
+  fl.Widget build(fl.BuildContext context) {
+    final textButtonStyle = fl.Theme.of(context).textButtonTheme.style?.copyWith(
+      backgroundColor: const fl.WidgetStatePropertyAll(fl.Colors.transparent),
+    );
+
+    return fl.SelectableText(
+      text,
+      style: style?.copyWith(overflow: fl.TextOverflow.ellipsis) ??
+          const fl.TextStyle(overflow: fl.TextOverflow.ellipsis),
+      textAlign: textAlign,
+      maxLines: maxLines,
+      contextMenuBuilder: (context, editableTextState) {
+        return fl.TapRegion(
+          onTapOutside: (_) => fl.ContextMenuController.removeAny(),
+          child: fl.TextButtonTheme(
+            data: fl.TextButtonThemeData(style: textButtonStyle),
+            child: fl.AdaptiveTextSelectionToolbar.buttonItems(
+              anchors: editableTextState.contextMenuAnchors,
+              buttonItems: editableTextState.contextMenuButtonItems,
+            ),
+          ),
+        );
+      },
+    );
+  }
+}
--- a/src/client/gui/lib/vm_details/ip_addresses.dart
+++ b/src/client/gui/lib/vm_details/ip_addresses.dart
@@ -1,6 +1,7 @@
-import 'package:flutter/material.dart' hide Tooltip;
+import 'package:flutter/material.dart' hide Tooltip, SelectableText;
 
 import '../extensions.dart';
+import '../selectable_text.dart';
 import '../tooltip.dart';
 
 class IpAddresses extends StatelessWidget {
--- a/src/client/gui/lib/vm_details/ip_addresses.dart
+++ b/src/client/gui/lib/vm_details/ip_addresses.dart
@@ -17,7 +17,7 @@ class IpAddresses extends StatelessWidget {
       Expanded(
         child: Tooltip(
           message: firstIp,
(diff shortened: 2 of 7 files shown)


src/client/gui/lib/extensions.dart
```diff
@@ -107,3 +107,45 @@ extension NullableMap<T> on T? {
...
+class WhitePopupMenuSelectableText extends StatelessWidget {
+  final String text;
+  final TextStyle? style;
+  final TextAlign? textAlign;
+  final int? maxLines;
+
+  const WhitePopupMenuSelectableText({
+    super.key,
+    required this.text,
+    this.style,
```
andrei-toterman, line 121: Just to keep things in line with the usual text widgets, how about you make `text` a positional parameter instead of a required named parameter?

src/client/gui/lib/extensions.dart
```diff
@@ -107,3 +107,45 @@ extension NullableMap<T> on T? {
     }
   }
 }
+
+/// A custom wrapper for SelectableText with a white popup (right-click) menu.
+class WhitePopupMenuSelectableText extends StatelessWidget {
```
andrei-toterman, line 112: That's kind of a lengthy name IMO. You can name it just `SelectableText` and use `import 'package:flutter/material.dart' hide SelectableText;` so that you can use yours without interfering with Flutter's.
Also, you can move this to its own file. `extensions.dart` is more for defining useful extension methods on types rather that for widgets.

andrei-toterman (changes requested):

src/client/gui/lib/extensions.dart
levkropp: > You can name it just `SelectableText` and use `import 'package:flutter/material.dart' hide SelectableText;` so that you can use yours without interfering with Flutter's.

When I tried to do this I wasn't able to define the contextMenuBuilder parameter: `The contextMenuBuilder parameter is not defined in the SelectableText widget.` 
So instead I moved the class into its own file at selectable_text.dart and renamed it to WhiteSelectableText

andrei-toterman: Yeah, it was because your `SelectableText` was conflicting with Flutter's here. You can circumvent this by importing Flutter under an alias in just this file. Take a look [here](https://github.com/canonical/multipass/blob/main/src/client/gui/lib/tooltip.dart) where there is a custom `Tooltip` widget that wraps Flutter's `Tooltip`.

andrei-toterman: I understand if this feels weird, and maybe it creates confusion. But IMO it's better than coming up with more elaborate names for widgets just because Flutter used them first. I'm open to criticism though!

ricab: Reusing standard names can be confusing. How about a shorter name for middle ground? Something like `MySelectableText`, `CustomSelectableText`, `MPSelectableText`? Or would `PopupSelectableText` be accurate enough?

levkropp: > Reusing standard names can be confusing.

It seems like we already have precedent for overwriting standard widgets in the GUI (e.g. Tooltip in [src/client/gui/lib/tooltip.dart](https://github.com/canonical/multipass/blob/4d27d5c84fad6b24f6ae48a9ee1a828b91b06066/src/client/gui/lib/tooltip.dart) and Switch in [src/client/gui/lib/switch.dart](https://github.com/canonical/multipass/blob/4d27d5c84fad6b24f6ae48a9ee1a828b91b06066/src/client/gui/lib/switch.dart)), but we can think about refactoring to a unified naming convention for all of these instead of overwriting the widgets entirely. Thoughts on this @andrei-toterman ?

andrei-toterman: My vote is for `MpSelectableText` or for something like
```dart
import '../selectable_text.dart' as mp;

final selectableText = mp.SelectableText(...);
```

levkropp: > something like
> 
> ```dart
> import '../selectable_text.dart' as mp;
> 
> final selectableText = mp.SelectableText(...);
> ```

I think this solution is the most elegant one because we only have to make changes to the imports and calls, and we don't have to change selectable_text.dart which keeps its implementation the same for other material.dart widget overrides we are doing (in switch.dart and tooltip.dart)

Whatever we choose, we also don't need to `hide SelectableText` when importing material.dart anymore. but maybe we should keep that for clarity?

levkropp:
We have chosen to implement CopyableText (click-to-copy functionality) in #3878 instead of SelectableText to allow users to copy instance information to the clipboard in the GUI

PR #3853, opened 2024-12-19
[gui] Open and close terminal tabs keyboard shortcuts

sharder996:
Tried playing around with the GUI and adding a change I'd personally like to see. Behaviour still isn't exactly how I'd like it. @andrei-toterman, a little help if you have time :)

Tabs open and close as expected with Ctrl+t/Ctrl+w or Cmd+t/Cmd+w when the vm is running and ssh connections can be made. However, if the vm is stopped (Terminal tabs are blank with the "Open shell" button), my keyboard shortcuts don't work. Maybe the intents/callback actions need to be in `terminal.dart` instead of `terminal_tabs.dart`, I'm not sure.

### DIFF
--- a/src/client/gui/lib/platform/linux.dart
+++ b/src/client/gui/lib/platform/linux.dart
@@ -5,6 +5,7 @@ import 'package:flutter/widgets.dart';
 
 import '../settings/autostart_notifiers.dart';
 import '../vm_details/terminal.dart';
+import '../vm_details/terminal_tabs.dart';
 import 'platform.dart';
 
 class LinuxPlatform extends MpPlatform {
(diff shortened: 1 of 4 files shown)


andrei-toterman (changes requested):
Hey, @sharder996! Indeed if the terminal is opened, the shortcuts work. The reason for this is because the shortcuts in that map that you extended are registered only on the widget for the actual terminal view, not for the widget that displays that 'Open shell' button. So if the widget for the terminal view is not built, it won't pick up the shortcuts. I think these shortcuts and intents are not related directly to a single terminal, but to the whole tab collection. So the place where you put the `Actions` widget is perfect. What I think needs to change is to also introduce a `Shortcuts` widget to `terminal_tabs.dart`, so that they are associated to the widget responsible for managing all the tabs, not to every individual terminals.

PR #3877, opened 2025-01-15
add Visual Studio and VSCode project configuration directories to gitignore

levkropp:
This PR updates the .gitignore file to include the `.vs` and `.vscode` folders. These folders contain IDE-specific configuration files created by Visual Studio and Visual Studio Code, similar to the `.idea` folder for JetBrains IDEs that is already currently in the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .idea
+.vs
+.vscode
 cmake-build-*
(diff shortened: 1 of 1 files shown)


ricab (approved):
Yup :+1:

PR #3896, opened 2025-01-24
Add a CPU architecture/model field to the bug-report template

ricab:
I find that it would often be useful to have this information from the beginning.


andrei-toterman (approved):
Yup, makes perfect sense.

PR #3897, opened 2025-01-24
Fix gui launch bridged error

### LINKED ISSUE #3881: Create instance with bridged network: instance <name> does not exist
**Describe the bug**
Configured an instance with bridged network -> errors out with "instance <name> does not exist".

**To Reproduce**
How, and what happened?
Fire up the GUI, choose Catalog > 24.04 > cog-wheel to configure. Choose name, cpu's, memory, *enable bridged network* and add a mount. Then click Launch

Result: error "instance <name> does not exist" where <name> is the name I chose.

SOLUTION: After I configured a network interface in settings, this error did not occur anymore. The error reproduced above is very unclear.

**Expected behavior**
It just works. OR an error: "you have not configured a network interface for bridged network, go to Settings and choose a network interface"

**Additional info**
 - OS: macOS 14.6.1
- multipass   1.15.0+mac
- `multipass info`
-  qemu

andrei-toterman:
This PR introduces the following changes:
- when launching an instance with some mounts specified, if the reply stream of the launch RPC yielded an error, the launch operation would not end early, and it would proceed to attempt the mounts. now, instead of yielding directly from the launch reply stream, we use `await for` to iterate over the stream and manually yield the replies. this turns errors from the stream into exceptions that are thrown, interrupting the launch process early, so mounts are no longer attempted
- we now allow our custom `Switch` to be conditionally disabled. there were also some minor changes to it, related to using newer styling properties in place of deprecated ones, with no effect to visuals
- in the launch form, we check if a bridged network was set correctly, and if not, we disable the switch for connecting to the bridged network, with an appropriate message
- we now allow un-setting the bridged network from the Settings page, by introducing a 'None' option, equivalent to `multipass set local.bridged-network=''`

fix #3881
MULTI-1764

### DIFF
--- a/src/client/gui/lib/catalogue/launch_form.dart
+++ b/src/client/gui/lib/catalogue/launch_form.dart
@@ -56,12 +56,16 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
     super.dispose();
   }
 
+  final bridgedNetworkProvider = daemonSettingProvider('local.bridged-network');
+
   @override
   Widget build(BuildContext context) {
     final imageInfo = ref.watch(launchingImageProvider);
     final randomName = ref.watch(randomNameProvider);
     final vmNames = ref.watch(vmNamesProvider);
     final deletedVms = ref.watch(deletedVmsProvider);
+    final networks = ref.watch(networksProvider);
+    final bridgedNetworkSetting = ref.watch(bridgedNetworkProvider).valueOrNull;
 
     final closeButton = IconButton(
       icon: const Icon(Icons.close),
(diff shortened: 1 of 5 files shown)


levkropp (dismissed):
Looks good after changing
`Once connection is established, you won't be able to unset it.` 
to
`Once established, you won't be able to unset it.` 
or
`You won't be able to unset the connection once it is established.` 
or something along those lines
settings work on my end!

georgeliao:
The behavior changes on the preemptive abortion during launch and adding bridged network `None` option makes sense to me. 
The code change also looks sane to me. 
I also did some functional testing on the macos. Everything works as expected.

georgeliao (approved):

levkropp (approved):
Good on my end now! `multipassd` crashes when switching the bridged network but when installed from the snap it reconnects in a couple of seconds. Launching and shelling into an instance with a bridged network set up works successfully!

ricab:
FWIW, it is not a crash. The daemon exits whenever its configuration changes, relying on the platform to restart the service.

PR #3905, opened 2025-01-28
[gui] update tray menu dep so that it has a license

andrei-toterman:
Before, this dependency had no license at all. Now it has a LGPL3 license, so we should include that as well.


ricab (approved):
Yep :+1:

PR #3909, opened 2025-01-30
Standard grpc mTLS 

georgeliao:
[MULTI-1765](https://warthogs.atlassian.net/browse/MULTI-1765)
Public side of https://github.com/canonical/multipass-private/pull/719

**A few things to note**
The `make_cert_key_pair` function originally handles the generation of both server and client key-certificate pairs . The `server_name` parameter determines the use case, when `server_name` is an empty string, it indicated the client key-certificate case. In the standard gRPC TLS scheme, the root key-certificate pair is added. 

Now, the role of the `make_cert_key_pair` function is still generating and holding server and client key-certificate pairs. However, the server certificate has evolved into a signed server certificate. As the result, the server side branch must first generate root key-certificate pair and use it to sign the newly created server certificate. 

Previously the constructor of  `X509Cert` only handled server and client certifcate generation, distinguishing between them based on whether `server_address` was empty. Now, with the introduction of root, server and client certifcate, the dispatch is managed by using the `CertType` enumeration. 

Additionally, the `X509Cert` constructor has been refined to generate certificates in a standard format. Key differences between before and after include adjustments to the serial number format and the inclusion of X509v3 extensions. To inspect a certificate, you can use the following command:  `openssl x509 -in <cert_path> -noout -text` 

The certificate paths are as follows:

- Server certificate:`/root/.local/share/multipassd/certificates/localhost.pem` 
- Client certificate: `/home/<user name>/.local/share/multipass-client-certificate/multipass_cert.pem` 
- Root certificate: `/usr/local/share/ca-certificates/multipass_root_cert.pem`

**Snap environment considerations**
In the snap environment, the root certificate is stored at `/var/snap/multipass/common/data/multipassd/certificates/multipass_root_cert.pem` 
Unlike `/root/`, `/var/snap/` directory allows other users view its files, making this setup feasible. 

**Certificate regeneration and migration**
The root certificate and server key-certificate pairs area automatically regenerated if either is missing. This mechanism ensures a smooth server key-certificate pair migration when updating Multipass. Upon upgrading, the server startup process will automatically generates root key-certificate pair and use it to sign a fresh server certificate. Consequently, the original server key-certificate pair will be overwritten, enabling successful verification under the new standard grpc TLS. 

The Multipass upgrade process should be included in the functional testing as well, both the cmd and gui clients should be tested. 

**Unit test adaptations**
The unit tests have been modified to accommodate changes in the gRPC TLS verification process. The key adjustments include:
1. Mocking the `get_root_cert_path` to allow usage of string-based certificates. 
2. Merge the two server and client key-certificate pairs into one,  With `MockCertProvider` being used on both server side and client to provide key-certificate pair. In the unit testing environment, they can be the same. 

Following the commit history and messages is also a helpful way to understand changes that have been made. 

[MULTI-1765]: https://warthogs.atlassian.net/browse/MULTI-1765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -196,7 +196,7 @@ void parallel_for_each(Container& input_container, UnaryOperation&& unary_op)
     }
 }
 
-// utility function for checking return code or pointer from C-apis
+// utility function for checking return code or raw pointer from C-apis
 // TODO: constrain T to int or raw pointer once C++20 concepts is available
 template <typename T>
 void check(T result, const std::string& errorMessage)
@@ -210,7 +210,7 @@ void check(T result, const std::string& errorMessage)
     }
     else
     {
-        if (result != 1)
+        if (result <= 0)
         {
             throw std::runtime_error(errorMessage);
         }
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -195,27 +195,6 @@ void parallel_for_each(Container& input_container, UnaryOperation&& unary_op)
         empty_future.get();
     }
 }
-
-// utility function for checking return code or raw pointer from C-apis
-// TODO: constrain T to int or raw pointer once C++20 concepts is available
-template <typename T>
-void check(T result, const std::string& errorMessage)
-{
-    if constexpr (std::is_pointer_v<T>)
-    {
-        if (result == nullptr)
-        {
-            throw std::runtime_error(errorMessage);
-        }
-    }
-    else
-    {
-        if (result <= 0)
-        {
-            throw std::runtime_error(errorMessage);
-        }
-    }
-}
 } // namespace utils
 
 class Utils : public Singleton<Utils>
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -195,6 +195,27 @@ void parallel_for_each(Container& input_container, UnaryOperation&& unary_op)
         empty_future.get();
     }
 }
+
+// utility function for checking return code or pointer from C-apis
+// TODO: constrain T to int or raw pointer once C++20 concepts is available
+template <typename T>
+void check(T result, const std::string& errorMessage)
+{
+    if constexpr (std::is_pointer_v<T>)
+    {
+        if (result == nullptr)
+        {
+            throw std::runtime_error(errorMessage);
+        }
+    }
+    else
+    {
+        if (result != 1)
+        {
+            throw std::runtime_error(errorMessage);
+        }
+    }
+}
 } // namespace utils
 
 class Utils : public Singleton<Utils>
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -41,12 +41,14 @@ class WritableFile
 public:
     explicit WritableFile(const QString& file_path)
     {
-        const std::string file_path_str = file_path.toStdString();
-        std::filesystem::remove(std::filesystem::path{file_path_str});
+        const std::filesystem::path file_path_std{file_path.toStdString()};
+        std::filesystem::remove(file_path_std);
         // The key files are read-only for the owner, meaning that opening them with fopen in wb mode typically fails,
         // preventing the overwrite functionality. Therefore, fs::remove is called beforehand, as it depeneds on the
         // parent directory's write permission rather than the file's write permission.
-        const auto raw_fp = fopen(file_path_str.c_str(), "wb");
+        std::filesystem::create_directories(file_path_std.parent_path());
+        // make sure the parent directory exist
+        const auto raw_fp = fopen(file_path_std.c_str(), "wb");
 
         if (raw_fp == nullptr)
             throw std::runtime_error(
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -151,7 +151,7 @@ void set_random_serial_number(X509* cert)
     serial_bytes[0] &= 0x7F;
 
     // Convert bytes to an BIGNUM, an arbitrary-precision integer type
-    std::unique_ptr<BIGNUM, decltype(&BN_free)> bn(BN_bin2bn(serial_bytes.data(), serial_bytes.size(), NULL), BN_free);
+    std::unique_ptr<BIGNUM, decltype(&BN_free)> bn(BN_bin2bn(serial_bytes.data(), serial_bytes.size(), nullptr), BN_free);
     if (!bn)
     {
         throw std::runtime_error("Failed to convert serial bytes to BIGNUM\n");
@@ -160,7 +160,7 @@ void set_random_serial_number(X509* cert)
 
     // Convert BIGNUM to ASN1_INTEGER and set it as the certificate serial number
     // ASN1 is a standard binary format for encoding data like serial numbers in X.509 certificates
-    ASN1_INTEGER* serial = BN_to_ASN1_INTEGER(bn.get(), NULL);
+    ASN1_INTEGER* serial = BN_to_ASN1_INTEGER(bn.get(), nullptr);
     if (!serial)
     {
         throw std::runtime_error("Failed to convert serial bytes to BIGNUM\n");
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -41,11 +41,12 @@ class WritableFile
 public:
     explicit WritableFile(const QString& file_path)
     {
-        std::filesystem::remove(file_path.toStdString().c_str());
+        const std::string file_path_str = file_path.toStdString();
+        std::filesystem::remove(std::filesystem::path{file_path_str});
         // The key files are read-only for the owner, meaning that opening them with fopen in wb mode typically fails,
         // preventing the overwrite functionality. Therefore, fs::remove is called beforehand, as it depeneds on the
         // parent directory's write permission rather than the file's write permission.
-        const auto raw_fp = fopen(file_path.toStdString().c_str(), "wb");
+        const auto raw_fp = fopen(file_path_str.c_str(), "wb");
 
         if (raw_fp == nullptr)
             throw std::runtime_error(
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -152,11 +152,11 @@ void set_random_serial_number(X509* cert)
 
     // Convert bytes to an BIGNUM, an arbitrary-precision integer type
     std::unique_ptr<BIGNUM, decltype(&BN_free)> bn(BN_bin2bn(serial_bytes.data(), serial_bytes.size(), NULL), BN_free);
-
     if (!bn)
     {
         throw std::runtime_error("Failed to convert serial bytes to BIGNUM\n");
     }
+    assert(!BN_is_negative(bn.get()));
 
     // Convert BIGNUM to ASN1_INTEGER and set it as the certificate serial number
     // ASN1 is a standard binary format for encoding data like serial numbers in X.509 certificates
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -194,9 +194,11 @@ public:
 
         set_random_serial_number(cert.get());
         X509_gmtime_adj(X509_get_notBefore(cert.get()), 0); // Start time: now
-        const long valid_duration_sec = cert_type == CertType::Root ? 3650L * 24L * 60L * 60L : 365L * 24L * 60L * 60L;
-        // 10 years for root certicicate and 1 year for server and client certificate
-        X509_gmtime_adj(X509_get_notAfter(cert.get()), valid_duration_sec);
+
+        constexpr std::chrono::seconds one_year = std::chrono::hours{24} * 365;
+        constexpr std::chrono::seconds ten_years = one_year * 10;
+        const auto valid_duration = cert_type == CertType::Root ? ten_years : one_year;
+        X509_gmtime_adj(X509_get_notAfter(cert.get()), valid_duration.count());
 
         constexpr int APPEND_ENTRY{-1};
         constexpr int ADD_RDN{0};
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -45,12 +45,37 @@ FILE* open_file(const QString& file_path)
     const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
 
     if (raw_fp == nullptr)
-        throw std::runtime_error(
-            fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
+        throw std::runtime_error(fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
 
     return raw_fp;
 }
 
+EVP_PKEY* create_raw_key_pointer()
+{
+    std::unique_ptr<EVP_PKEY_CTX, decltype(&EVP_PKEY_CTX_free)> ctx(EVP_PKEY_CTX_new_from_name(nullptr, "EC", nullptr),
+                                                                    EVP_PKEY_CTX_free);
+
+    if (!ctx || EVP_PKEY_keygen_init(ctx.get()) <= 0)
+    {
+        throw std::runtime_error("Failed to initialize key generation");
+    }
+
+    // Set EC curve (P-256)
+    const std::array<OSSL_PARAM, 2> params = {
+        // the 3rd argument is length of the buffer, which is 0 in the case of static buffer like "P-256"
+        OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_GROUP_NAME, const_cast<char*>("P-256"), 0),
+        OSSL_PARAM_construct_end()};
+    EVP_PKEY_CTX_set_params(ctx.get(), params.data());
+
+    // Generate the key
+    EVP_PKEY* raw_key = nullptr;
+    if (EVP_PKEY_generate(ctx.get(), &raw_key) <= 0)
+    {
+        throw std::runtime_error("Failed to generate EC key");
+    }
+
+    return raw_key;
+}
 
 class WritableFile
 {
@@ -73,33 +98,8 @@ private:
 class EVPKey
 {
 public:
-    EVPKey()
+    EVPKey() : key{create_raw_key_pointer(), EVP_PKEY_free}
     {
-        std::unique_ptr<EVP_PKEY_CTX, decltype(&EVP_PKEY_CTX_free)> ctx(
-            EVP_PKEY_CTX_new_from_name(nullptr, "EC", nullptr),
-            EVP_PKEY_CTX_free);
-
-        if (!ctx || EVP_PKEY_keygen_init(ctx.get()) <= 0)
-        {
-            throw std::runtime_error("Failed to initialize key generation");
-        }
-
-        // Set EC curve (P-256)
-        const std::array<OSSL_PARAM, 2> params = {
-            // the 3rd argument is length of the buffer, which is 0 in the case of static buffer like "P-256"
-            OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_GROUP_NAME, const_cast<char*>("P-256"), 0),
-            OSSL_PARAM_construct_end()};
-        EVP_PKEY_CTX_set_params(ctx.get(), params.data());
-
-        // Generate the key
-        EVP_PKEY* raw_key = nullptr;
-        if (EVP_PKEY_generate(ctx.get(), &raw_key) <= 0)
-        {
-            throw std::runtime_error("Failed to generate EC key");
-        }
-
-        // Assign generated key to unique_ptr for RAII management
-        key.reset(raw_key);
     }
 
     std::string as_pem() const
@@ -134,7 +134,7 @@ public:
     }
 
 private:
-    std::unique_ptr<EVP_PKEY, decltype(&EVP_PKEY_free)> key{nullptr, EVP_PKEY_free};
+    std::unique_ptr<EVP_PKEY, decltype(&EVP_PKEY_free)> key;
 };
 
 std::vector<unsigned char> as_vector(const std::string& v)
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -36,22 +36,27 @@ namespace mp = multipass;
 
 namespace
 {
+FILE* open_file(const QString& file_path)
+{
+    const std::filesystem::path file_path_std{file_path.toStdString()};
+    std::filesystem::create_directories(file_path_std.parent_path());
+    // make sure the parent directory exist
+
+    const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
+
+    if (raw_fp == nullptr)
+        throw std::runtime_error(
+            fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
+
+    return raw_fp;
+}
+
+
 class WritableFile
 {
 public:
-    explicit WritableFile(const QString& file_path)
+    explicit WritableFile(const QString& file_path) : fp{open_file(file_path), fclose}
     {
-        const std::filesystem::path file_path_std{file_path.toStdString()};
-        std::filesystem::create_directories(file_path_std.parent_path());
-        // make sure the parent directory exist
-
-        const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
-
-        if (raw_fp == nullptr)
-            throw std::runtime_error(
-                fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
-
-        fp.reset(raw_fp);
     }
 
     FILE* get() const
@@ -62,7 +67,7 @@ public:
 private:
     // decltype(&fclose) does not preserve these some extra function attributes of fclose, leads to warning and
     // compilation error
-    std::unique_ptr<FILE, int (*)(FILE*)> fp{nullptr, fclose};
+    std::unique_ptr<FILE, int (*)(FILE*)> fp;
 };
 
 class EVPKey
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -48,7 +48,7 @@ public:
         // parent directory's write permission rather than the file's write permission.
         std::filesystem::create_directories(file_path_std.parent_path());
         // make sure the parent directory exist
-        const auto raw_fp = fopen(file_path_std.c_str(), "wb");
+        const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
 
         if (raw_fp == nullptr)
             throw std::runtime_error(
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -42,12 +42,15 @@ public:
     explicit WritableFile(const QString& file_path)
     {
         const std::filesystem::path file_path_std{file_path.toStdString()};
-        std::filesystem::remove(file_path_std);
-        // The key files are read-only for the owner, meaning that opening them with fopen in wb mode typically fails,
-        // preventing the overwrite functionality. Therefore, fs::remove is called beforehand, as it depeneds on the
-        // parent directory's write permission rather than the file's write permission.
         std::filesystem::create_directories(file_path_std.parent_path());
         // make sure the parent directory exist
+        if (std::filesystem::exists(file_path_std))
+        {
+            // enable fopen with wb mode
+            MP_PLATFORM.set_permissions(file_path_std,
+                                        std::filesystem::perms::owner_read | std::filesystem::perms::owner_write);
+        }
+
         const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
 
         if (raw_fp == nullptr)
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -44,12 +44,6 @@ public:
         const std::filesystem::path file_path_std{file_path.toStdString()};
         std::filesystem::create_directories(file_path_std.parent_path());
         // make sure the parent directory exist
-        if (std::filesystem::exists(file_path_std))
-        {
-            // enable fopen with wb mode
-            MP_PLATFORM.set_permissions(file_path_std,
-                                        std::filesystem::perms::owner_read | std::filesystem::perms::owner_write);
-        }
 
         const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
 
@@ -114,6 +108,14 @@ public:
 
     void write(const QString& key_path) const
     {
+        const std::filesystem::path key_path_std = key_path.toStdU16String();
+        if (std::filesystem::exists(key_path_std))
+        {
+            // enable fopen in WritableFile with wb mode
+            MP_PLATFORM.set_permissions(key_path_std,
+                                        std::filesystem::perms::owner_read | std::filesystem::perms::owner_write);
+        }
+
         WritableFile file{key_path};
         if (!PEM_write_PrivateKey(file.get(), key.get(), nullptr, nullptr, 0, nullptr, nullptr))
             throw std::runtime_error(fmt::format("Failed writing certificate private key to file '{}'", key_path));
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -178,8 +178,9 @@ public:
         if (cert == nullptr)
             throw std::runtime_error("Failed to allocate x509 cert structure");
 
-        set_random_serial_number(cert.get());
+        X509_set_version(cert.get(), 2); // 0 index based, 2 amounts to 3
 
+        set_random_serial_number(cert.get());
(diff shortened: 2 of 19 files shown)


src/platform/platform_unix.cpp
```diff
@@ -241,3 +242,13 @@ long long mp::platform::Platform::get_total_ram() const
 {
     return static_cast<long long>(sysconf(_SC_PHYS_PAGES)) * sysconf(_SC_PAGESIZE);
 }
+
+std::filesystem::path mp::platform::Platform::get_root_cert_path() const
+{
+    constexpr auto* root_cert_file_name = "multipass_root_cert.pem";
+
+    return mp::utils::in_multipass_snap()
+               ? std::filesystem::path{QString{mp::utils::snap_common_dir()}.toStdString().c_str()} /
```
andrei-toterman, line 251:
```suggestion
               ? std::filesystem::path{mp::utils::snap_common_dir().toStdString()} /
```

andrei-toterman (changes requested):
Just a quick pass with some nitpicks, didn't test functionally yet.

src/platform/platform_linux.cpp
```diff
@@ -470,3 +479,11 @@ std::string multipass::platform::host_version()
     return mpu::in_multipass_snap() ? multipass::platform::detail::read_os_release()
                                     : fmt::format("{}-{}", QSysInfo::productType(), QSysInfo::productVersion());
 }
+
+std::filesystem::path mp::platform::Platform::get_root_cert_path() const
+{
+    constexpr auto* root_cert_file_name = "multipass_root_cert.pem";
```
xmkg, line 485: Is there a reason for using `auto*` instead of just plain `auto`? Both should result in `const char*`.

xmkg (dismissed):
LGTM, except for one small nit.

src/platform/platform_linux.cpp
georgeliao: Yes, they are equivalent in this case. But normally I prefer append `*` or `&` to add a bit expressiveness of data type.

andrei-toterman (changes requested):
Hey, @georgeliao! So now, in the local environment, the GUI works fine both when upgrading from the old certificates and when starting with no certificates. But in the case of the CLI, it works when starting with no certificates, but if I upgrade from old certificates, I get the following error
```
E0211 14:23:49.406977684   36809 ssl_transport_security.cc:1316]       Handshake failed with fatal error SSL_ERROR_SSL: error:0A000086:SSL routines::certificate verify failed.
```
And in the snap, both when starting fresh or when upgrading, I get the following
```
[error] [client] Caught an unhandled exception: failed to open file '/var/snap/multipass/commondata/multipassd/certificates/multipass_root_cert.pem': No such file or directory(2)
```

andrei-toterman:
Yeah, for the dev environment, I don't think it's worth doing the effort. We already have to remove stuff manually, so this is just an extra thing to remove.

andrei-toterman (approved):
Great, now both the CLI and the GUI work as expected, both locally and in the snap, both when upgrading or installing fresh. Thanks, @georgeliao!

Now, one thing to discuss with everyone: the purpose of these changes was so that we could remove our grpc patches, and that goal is accomplished by this PR. And now the CLI verifies the server certificate, which couldn't be avoided without the patches. But the GUI can avoid verifying the server certificate, using the plain vanilla grpc dart library. My question is: should the GUI also verify the server certificate, to be in line with the CLI, or should we keep the existing, functioning behavior of not having the GUI verify the server certificate?

ricab:
As discussed elsewhere, let's have the GUI also adhere to the new scheme, preferably in a separate PR.

src/platform/platform_linux.cpp
```diff
@@ -470,3 +478,11 @@ std::string multipass::platform::host_version()
     return mpu::in_multipass_snap() ? multipass::platform::detail::read_os_release()
                                     : fmt::format("{}-{}", QSysInfo::productType(), QSysInfo::productVersion());
 }
+
+std::filesystem::path mp::platform::Platform::get_root_cert_path() const
```
georgeliao, line 482: @ricab Can you review the implementation of `get_root_cert_path` function, especially its sub-function `multipass_final_storage_location()`, make sure the path looks correct to you. Besides that, please also have a look of the whole private PR [diff](https://github.com/canonical/multipass-private/pull/719). Thanks.

src/cert/ssl_cert_provider.cpp
```diff
@@ -103,128 +129,216 @@ class EVPKey
...
 std::vector<unsigned char> as_vector(const std::string& v)
 {
     return {v.begin(), v.end()};
 }
 
-std::string cn_name_from(const std::string& server_name)
+void set_random_serial_number(X509* cert)
 {
-    if (server_name.empty())
-        return mp::utils::make_uuid().toStdString();
```
ricab, line 117: Was this case only ever hit in clients, before?

src/cert/ssl_cert_provider.cpp
```diff
@@ -41,10 +39,19 @@ namespace
 class WritableFile
 {
 public:
-    explicit WritableFile(const QString& name) : fp{fopen(name.toStdString().c_str(), "wb"), fclose}
+    explicit WritableFile(const QString& file_path)
```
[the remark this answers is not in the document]
georgeliao: That makes sense. What about a small variation of that, meaning makes the `helper` function lambda and invoke it right away. In this case, the exclusivity and localization of the functor is better. 
That makes the result code:
```
class WritableFile
{
public:
    explicit WritableFile(const QString& file_path)
        : fp{[](const QString& file_path) {
                 const std::filesystem::path file_path_std{file_path.toStdString()};
                 std::filesystem::create_directories(file_path_std.parent_path());
                 // make sure the parent directory exist

                 const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");

                 if (raw_fp == nullptr)
                     throw std::runtime_error(
                         fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
                 return raw_fp;
             }(file_path),
             fclose}
    {
    }

    FILE* get() const
    {
        return fp.get();
    }

private:
    // decltype(&fclose) does not preserve these some extra function attributes of fclose, leads to warning and
    // compilation error
    std::unique_ptr<FILE, int (*)(FILE*)> fp;
};
```

src/daemon/daemon_config.cpp
```diff
@@ -167,8 +167,8 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
     if (ssh_key_provider == nullptr)
         ssh_key_provider = std::make_unique<OpenSSHKeyProvider>(data_directory);
     if (cert_provider == nullptr)
-        cert_provider = std::make_unique<mp::SSLCertProvider>(MP_UTILS.make_dir(data_directory, "certificates"),
```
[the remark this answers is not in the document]
georgeliao: This stems from the change I have made in the `Writable` class constructor, the line  `std::filesystem::create_directories(file_path_std.parent_path());` was added, meaning that the parent directory of the certificate and key is guaranteed to exist. That includes server certificate directory (that is this one), client certificate directory, and root certificate directory. As a result, the `make_dir` call is no longer needed.

src/cert/ssl_cert_provider.cpp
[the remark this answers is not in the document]
ricab: Sure, I am not opposed to that. I would personally prefer a well-named function, but I think that lambda is also fine.

src/daemon/daemon_config.cpp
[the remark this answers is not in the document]
ricab: Got it, thanks.

src/cert/ssl_cert_provider.cpp
```diff
@@ -103,128 +129,216 @@ class EVPKey
...
-    X509_add1_ext_i2d(c, NID_subject_alt_name, gens, 0, X509V3_ADD_DEFAULT);
+    // Convert BIGNUM to ASN1_INTEGER and set it as the certificate serial number
+    // ASN1 is a standard binary format for encoding data like serial numbers in X.509 certificates
+    ASN1_INTEGER* serial = BN_to_ASN1_INTEGER(bn.get(), NULL);
+    if (!serial)
+    {
+        throw std::runtime_error("Failed to convert serial bytes to BIGNUM\n");
+    }
 
-    GENERAL_NAMES_free(gens);
```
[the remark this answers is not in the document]
georgeliao: Which code? `GENERAL_NAMES_free` function is a deprecated api. The whole `set_san_name` function was re-written based on the modern openssl c-apis. Particularly, this call and the implementation of `add_extension` does that, 
```
        if (cert_type == CertType::Server)
        {
            add_extension(ctx, NID_subject_alt_name, ("DNS:" + server_name).c_str());
        }
```
feel

src/cert/ssl_cert_provider.cpp
```diff
@@ -103,128 +129,216 @@ class EVPKey
...
-
-        X509_set_version(x509.get(), 2);
+        X509_set_version(cert.get(), 2); // 0 index based, 2 amounts to 3
 
-        ASN1_INTEGER_set(X509_get_serialNumber(x509.get()), big_num);
-        X509_gmtime_adj(X509_get_notBefore(x509.get()), 0);
-        X509_gmtime_adj(X509_get_notAfter(x509.get()), 31536000L);
+        set_random_serial_number(cert.get());
+        X509_gmtime_adj(X509_get_notBefore(cert.get()), 0); // Start time: now
+        const long valid_duration_sec = cert_type == CertType::Root ? 3650L * 24L * 60L * 60L : 365L * 24L * 60L * 60L;
```
xmkg, line 192: We can get rid of the magic numbers here in favor of something more readable:

```C++
    constexpr std::chrono::seconds one_year = std::chrono::hours{24} * 365; 
    constexpr std::chrono::seconds ten_years = one_year * 10; 
    const auto valid_duration = cert_type == CertType::Root ? ten_years : one_year;
    X509_gmtime_adj(X509_get_notAfter(cert.get()), valid_duration.count());
```

We can also simplify this even further when we switch to C++20 with the help of "std::chrono::years".

georgeliao: Yep, it is definitely better with chrono time units and automatic unit conversion.

georgeliao: It is a bit funny that `hours` is the biggest unit before c++20.

src/cert/ssl_cert_provider.cpp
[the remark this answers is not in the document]
georgeliao: Indeed, named function is better.

src/cert/ssl_cert_provider.cpp
```diff
@@ -53,30 +60,41 @@ class WritableFile
     }
 
 private:
-    std::unique_ptr<FILE, std::function<int(FILE*)>> fp;
+    // decltype(&fclose) does not preserve these some extra function attributes of fclose, leads to warning and
+    // compilation error
+    std::unique_ptr<FILE, int (*)(FILE*)> fp{nullptr, fclose};
```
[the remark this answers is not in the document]
georgeliao: done

src/cert/ssl_cert_provider.cpp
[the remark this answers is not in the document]
georgeliao: done

src/cert/ssl_cert_provider.cpp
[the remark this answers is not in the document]
ricab: OK, thanks for clarifying.

src/cert/ssl_cert_provider.cpp
ricab: OK, it might be good to add a comment next to the signature of the `SSLCertProvider` constructor. Something like: `// leave server_name empty for clients; choose a (non-empty) name for servers.`

src/cert/ssl_cert_provider.cpp
```diff
@@ -103,128 +129,216 @@ class EVPKey
...
-{
-    std::string san_dns = server_name;
-    GENERAL_NAMES* gens = sk_GENERAL_NAME_new_null();
-    GENERAL_NAME* gen = GENERAL_NAME_new();
-    ASN1_IA5STRING* ia5 = ASN1_IA5STRING_new();
-    ASN1_STRING_set(ia5, san_dns.data(), san_dns.length());
-    GENERAL_NAME_set0_value(gen, GEN_DNS, ia5);
-    sk_GENERAL_NAME_push(gens, gen);
+    // Convert bytes to an BIGNUM, an arbitrary-precision integer type
+    std::unique_ptr<BIGNUM, decltype(&BN_free)> bn(BN_bin2bn(serial_bytes.data(), serial_bytes.size(), NULL), BN_free);
```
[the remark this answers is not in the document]
georgeliao: That indeed an good addition, and you are in luck. openssl has an utility function `BN_is_negative` for this.

src/platform/platform_linux.cpp
ricab: > 1\. How to correctly check the readability of the file to other users?

I think we could skip the verification of the parent directories. We know what directories can be returned on each platform and we know what their permissions are supposed to be. In practice, we know that `C:/ProgramData`, `/Library/Application\ Support` and `/usr/local/share/` are all readable by all. IOW, I think it is safe to assume that the first `GenericDataLocation` entry that isn't bound to a user is readable by all.

> I am not even sure whether this check is compatible with windows.

That would need to be verified first but it would be weird if it wasn't. Qt is supposed to abstract the platform.

> 2\. How can we make the client locates the server generated data folder?
The [locate function](https://doc.qt.io/qt-6/qstandardpaths.html#locate) still requires `QStandardPaths::AppDataLocation` as an input and that will be resolved to a client user based path.

`AppDataLocation` is bound to the current app (paths include `<APPNAME>` and `<APPDIR>`), so we would need to locate `multipassd/.../root.cert` in `GenericDataLocation`. As discussed, the locate function tries each possible location in turn so, as long as we choose a unique name for the root certificate, it would eventually look in the second entry (the one that does not depend on the user, e.g. ` /usr/local/share`) and find it there.

> Indeed, the restriction is enforced by Multipass instead of C:/ProgramData/ itself. The exact line does it is [here](https://github.com/canonical/multipass/blob/24247990441d5d4c95e0c16a9411467e31d2b0df/src/daemon/daemon_config.cpp#L113).

OK, that is as I suspected. It's a pretty recent change. You'd need to fix it to be more granular.

> That also means the client process will use `MP_PLATFORM.multipass_storage_location()` as opposed to `QStandardPaths::AppDataLocation` to locate the path if we want to go down this route.

Yes, we'd always need to use it if we wanted to `MULTIPASS_STORAGE` for the root certificate.

xmkg (changes requested):
Looks good overall, a couple of nitpicks and a proposal for handling all the error codes returned by the OpenSSL API.

georgeliao:
@ricab @Sploder12 
Please review this file permission related [part](https://github.com/canonical/multipass-private/compare/1351dd1cd821154ad6b7757b4cc2a487bf7fc47c...d70bb0304d9db6a9c21d02231cd0fc67aaec6033) which is based on our discussion. The current permission hierarchy in the storage location looks like the list below. 

```
drwx-----x 7 root root 4096 mrt  3 15:17 .
drwxr-xr-x 3 root root 4096 mrt  3 13:58 ..
drwx------ 2 root root 4096 mrt  3 13:58 authenticated-certs
drwxr-xr-x 2 root root 4096 mrt  3 13:58 certificates
-rwx------ 1 root root 1528 mrt  3 15:17 multipassd-vm-instances.json
drwxr-xr-x 2 root root 4096 mrt  3 13:58 lxd
drwx------ 2 root root 4096 mrt  3 13:58 network
drwx------ 2 root root 4096 mrt  3 13:58 ssh-keys
drwx------ 3 root root 4096 mrt  3 15:16 vault
drwxr-xr-x 3 root root 4096 mrt  3 15:16 virtualbox
```
`certificates` directory is untouched based on our discussion, and `lxd` and `virtualbox` directories are about to be removed so I also did not bother to set their permission. 

One thing to note is that the current implementation did not set the permissions recursively, which is enough based on the permission mechanism. If we really want the recursive permission, that involves more changes. It includes changing the `restrict_permissions` to accept a `fs::perm` argument in order to interface with the `make_dir` function, changing its name since the role of the function is changed, and adapting a bunch of other things.

src/daemon/daemon_config.cpp
```diff
@@ -192,14 +192,15 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
                 std::make_unique<DefaultVMBlueprintProvider>(url_downloader.get(), cache_directory, manifest_ttl);
     }
 
-    // restrict permissions for all existing files and folders
+    // restrict permissions for all existing files and folders in cache directory, the data directory opens execute
```
Sploder12, line 195: should the comment say "list" instead of "execute" since it's being applied to a directory?

src/daemon/daemon_config.cpp
```diff
@@ -192,14 +192,15 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
+    // permission for other users, the sub-directories of it will have a granular permissions setting
     if (!storage_path.isEmpty())
     {
-        MP_PERMISSIONS.restrict_permissions(storage_path.toStdU16String());
+        MP_PLATFORM.set_permissions(storage_path.toStdU16String(), fs::perms::owner_all | fs::perms::others_exec);
     }
     else
     {
-        MP_PERMISSIONS.restrict_permissions(data_directory.toStdU16String());
+        MP_PLATFORM.set_permissions(data_directory.toStdU16String(), fs::perms::owner_all | fs::perms::others_exec);
```
Sploder12, line 203: Should we consider the `group` permissions here as well?

Sploder12 (commented):
Looks pretty good, just a little issue with windows string to path conversion, and some questions

src/daemon/daemon_config.cpp
georgeliao: But there is no "list" permission, right? There is only a read permission on directory and that let you read the directory (list the items in it). However, that permission is not relevant to this use case because we want other users to view some files in the sub-directory. That requires the permission to enter all the parent directories and that is execute permission on directory is about.

src/daemon/daemon_config.cpp
```diff
@@ -109,7 +109,7 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
     auto multiplexing_logger = std::make_shared<mpl::MultiplexingLogger>(std::move(logger));
     mpl::set_logger(multiplexing_logger);
 
-    MP_PLATFORM.setup_permission_inheritance();
+    MP_PLATFORM.setup_permission_inheritance(false);
```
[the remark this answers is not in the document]
georgeliao: Yep, I was kind of aware that removing it and setting it false are more or less the same. Thanks for pointing out the subtle difference.

src/daemon/daemon_config.cpp
georgeliao: I think we only want to let the client process view the file and the client process is run from a normal user. Meanwhile, we also want to keep the directory as restrictive as possible. So just adding the other users execute on the directory is exactly what we need, right?

src/daemon/daemon.cpp
```diff
@@ -2867,7 +2867,10 @@ void mp::Daemon::persist_instances()
     }
     QDir data_dir{
         mp::utils::backend_directory_path(config->data_directory, config->factory->get_backend_directory_name())};
-    MP_JSONUTILS.write_json(instance_records_json, data_dir.filePath(instance_db_name));
+
+    const QString instance_db_path = data_dir.filePath(instance_db_name);
+    MP_JSONUTILS.write_json(instance_records_json, instance_db_path);
+    MP_PLATFORM.set_permissions(fs::path{instance_db_path.toStdString()}, fs::perms::owner_all);
```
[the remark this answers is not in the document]
georgeliao: Will do

src/cert/ssl_cert_provider.cpp
```diff
@@ -103,128 +134,219 @@ class EVPKey
...
-    auto cert_path = cert_dir.filePath(prefix + ".pem");
+    const auto priv_key_path = cert_dir.filePath(prefix + "_key.pem");
+    const auto cert_path = cert_dir.filePath(prefix + ".pem");
 
-    if (QFile::exists(priv_key_path) && QFile::exists(cert_path))
+    if (!server_name.empty())
     {
-        return {mp::utils::contents_of(cert_path), mp::utils::contents_of(priv_key_path)};
+        const std::filesystem::path root_cert_path = MP_PLATFORM.get_root_cert_path();
+        if (std::filesystem::exists(root_cert_path) && QFile::exists(priv_key_path) && QFile::exists(cert_path))
```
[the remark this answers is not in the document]
georgeliao: `if (std::filesystem::exists(root_cert_path)` is newly added code and the other two are legacy. I would prefer to replace all qt file system utilities in this file if we want to do that. But that will be another PR.

src/daemon/daemon_config.cpp
Sploder12: You're right, sometimes i get read/execute mixed up on directories.

src/cert/ssl_cert_provider.cpp
```diff
@@ -38,13 +36,52 @@ namespace mp = multipass;
...
+
+    const auto raw_fp = fopen(file_path_std.u8string().c_str(), "wb");
+
+    if (raw_fp == nullptr)
+        throw std::runtime_error(fmt::format("failed to open file '{}': {}({})", file_path, strerror(errno), errno));
+
+    return raw_fp;
+}
+
+EVP_PKEY* create_raw_key_pointer()
```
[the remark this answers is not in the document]
georgeliao: Yes, indeed. We have avoid delegating the deallocation responsibility to the client code.

tests/test_client_common.cpp
```diff
@@ -44,12 +44,22 @@ struct TestClientCommon : public mpt::DaemonTestFixture
     {
         ON_CALL(mpt::MockStandardPaths::mock_instance(), writableLocation(mp::StandardPaths::GenericDataLocation))
             .WillByDefault(Return(temp_dir.path()));
+        ON_CALL(*mock_utils, contents_of(_)).WillByDefault(Return(mpt::root_cert));
+        // delegate some functions to the orignal implementation
```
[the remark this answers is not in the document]
georgeliao: Let me sort out the whole story. These changes are added because 
```
    const mpt::MockUtils::GuardedMock utils_attr{mpt::MockUtils::inject<NiceMock>()};
    const mpt::MockUtils* mock_utils = utils_attr.first;
``` 
is added down there to override the `Utils` singleton class. As a result of that, some unit tests failed because MockUtils without mock implementation does nothing in the function body. In order to restore the original behavior, the delegation of these two functions is needed. As to whether the original code alters the filesystem, I think yes, but it happens in a confined temporary directory and the directory got removed after the unit test run.

include/multipass/utils.h
```diff
@@ -195,6 +195,27 @@ void parallel_for_each(Container& input_container, UnaryOperation&& unary_op)
...
+    if constexpr (std::is_pointer_v<T>)
+    {
+        if (result == nullptr)
+        {
+            throw std::runtime_error(errorMessage);
+        }
+    }
+    else
+    {
+        if (result <= 0)
```
xmkg, line 213: I think this check will depend on the C API being used. Some of the C API's use 0 to indicate success. We can decouple the checker implementation to have more flexibility:

```c++
// tag type to distinguish checker type
struct checker_base{};

// generic enough, can be placed in utils
struct nullptr_checker : private checker_base {
    bool operator()(const void* v){
        return v == nullptr;
    }
};

// openssl specific checker, better to be placed in a more localized file
struct ossl_checker : private checker_base{
    bool operator()(int v){
        return v == 1;
    }
};

template <typename CheckerT = nullptr_checker, typename T>
auto check(T result, const std::string& errorMessage) -> typename std::enable_if<std::is_base_of_v<checker_base, CheckerT>, void>::type
{
    if(!CheckerT{}(result)){
        throw std::runtime_error{errorMessage};
    }
}
```

usage:

```c++
int test(){
    return 5;
}

int main(void){
    check<ossl_checker>(test(), "");
    // check<int>(test(), ""); -> wont compile
}

```

xmkg: It'd also be good for troubleshooting if we could incorporate the error code into the thrown error.

tests/test_client_common.cpp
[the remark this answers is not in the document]
xmkg: I say we should put this as a comment as well, so we won't get confused later on.

include/multipass/utils.h
georgeliao: The error code can be added. However, about custom checker injection. It becomes that the user have to specify the checker even within the openssl c-api return pointer or integers cases. That defies one of the purpose of the utility function, abstraction. For now, I will move the utility function into `ssl_cert_provider.cpp` as a dedicated openssl c-api checker instead of trying to be a global c-api checker. Making a general checker for all c-apis (openssl, hyperv, libssh) can be done in another PR.

georgeliao: It is a very concern you brought up.

xmkg: > It becomes that the user have to specify the checker even within the openssl c-api return pointer or integers cases

We can create [local aliases](https://gcc.godbolt.org/z/b3eb4E11o) for such cases, but the gist of it is that we have to leave a customization point for the condition (e.g., a predicate or a dedicated type) if we want it to be a global utility, so different API result codes can be accommodated too. 


> For now, I will move the utility function into `ssl_cert_provider.cpp` as a dedicated openssl c-api checker instead of trying to be a global c-api checker.

I'm OK with that, we can migrate to the global one later if we decide to implement one.

xmkg: ... or we can adopt a rather simple approach:

```c++
struct PreconditionFailureError : std::runtime_error {
    using std::runtime_error::runtime_error;
};

void expects(bool what, const std::string& failure_msg) {
    if (!what) {
        throw PreconditionFailureError(failure_msg);
    }
}

int test() { return 5; }
int main(void) { 
    expects(test() == 4, "test failed."); 
}
```

src/daemon/daemon_config.cpp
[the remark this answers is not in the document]
ricab: Why keep a line of code that has no effect though? 

Also, couldn't we keep this true and override selectively?

src/daemon/daemon_config.cpp
ricab, line 203: What happens to the "takeown" part of the replaced calls? Is it no longer necessary? is it applied somewhere else?

ricab (commented):
Hey @georgeliao, I've reviewed the permissions part and it looks mostly good to me, provided you're in sync with @Sploder12 on it. Only a couple of things inline

include/multipass/utils.h
georgeliao: Yep, that definitely can be a candidate for global C-api checker.

src/daemon/daemon_config.cpp
georgeliao: I did not get this question. What is "takeown" part ? You mean setting owner only permission?

src/daemon/daemon_config.cpp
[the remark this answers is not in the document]
ricab: OK, that makes sense. @Sploder12 WDYT? Should we put it back in a later PR?

src/daemon/daemon_config.cpp
ricab: The calls to `restrict_permissions`, which are now replaced, used to take ownership of the files (beside setting permissions). I wonder if you really meant to skip that and, if so, why. Do we no longer need it, or are we doing the same somewhere else?

src/daemon/daemon_config.cpp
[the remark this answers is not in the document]
Sploder12: Putting it back in a later PR would be good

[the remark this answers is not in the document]
georgeliao: Hi, I have done a quick prototype of the `MP_PLATFORM.setup_permission_inheritance(true);` based implementation which is the last commits. It is simpler than the original one where we set permission for each sub-directory one by one. With the new one, we only set the default permission of file/folder creation to owner only and overrides the permission of storage location, the certificates folder, and the certificate files. We get the desired permission for root certificate by this way as well. However, there is only one problem. It only works for newly created directory and files, not existing ones from the users.

src/daemon/daemon_config.cpp
georgeliao: Ok, now I get the question. That is a very good point got brought up. I am also confused the fact that why we need it in the first place. Were all files and folders in the storage location owned by root user already? @Sploder12 Maybe you can share some context?

tests/test_client_common.cpp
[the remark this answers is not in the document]
georgeliao: But that will be comment of code transitioning, the code might evolve later and the connotation will be lost.  I would say just leave it like this. As long as we know we do have temporary folder based unit tests, then we should be able to understand the code.

src/daemon/daemon_config.cpp
Sploder12: If multipassd created the files and folders, they'd be owned by the root user already. So it's usually redundant and probably unneeded

src/daemon/daemon_config.cpp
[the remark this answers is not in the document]
georgeliao: @ricab the new `daemon_config.cpp` code change is ready for review. Another concern of yours is whether the certificate permission will overwrite if the certificate already exists. The answer is no based on our current code. However, we are still fine with the migration because this PR guarantees the user's certificate recreation due to the missing root certificate.

xmkg (dismissed):
LGTM :+1: Good work!

src/cert/ssl_cert_provider.cpp
```diff
@@ -280,6 +280,9 @@ class X509Cert
 
         openssl_check(PEM_write_X509(file.get(), cert.get()),
                       fmt::format("Failed writing certificate to file '{}'", cert_path));
+        const std::filesystem::path cert_path_std = cert_path.toStdU16String();
+        MP_PLATFORM.set_permissions(cert_path_std,
+                                    std::filesystem::perms::owner_all | std::filesystem::perms::others_read);
```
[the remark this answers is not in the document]
georgeliao: After looking into some literature, I think I have a misconception and the group read permission can be added.

ricab (approved):
Approving the latest changes as well (secondary review, no testing).

georgeliao:
Now it is time for a final round functional testing. Please make sure the grpc authentication works (`multipass list` can run and the GUI can connect to daemon) on migration and fresh installation. @Sploder12 can you check this on the linux platform especially the snap version? Maybe also check the directory permissions in this case. @andrei-toterman Can you test this on windows and macos? Also double check the uninstaller removes everything. Thanks.

Sploder12 (dismissed):
Functional testing on Ubuntu (both snap and not) revealed no issues and permissions on the directories are as expected. Looks good to me, great work!

georgeliao:
@Sploder12 can you do another round functional testing on linux just as you did before? I will test the windows and macos part.

Sploder12 (approved):
LGTM, everything works as expected on Ubuntu!

georgeliao:
@Sploder12

Can you do a review on the last two commits? Basically, the recursive permission setting on the data folder overwrites the root cert permission to be not accessible to client process. The exact behavior is that the server restart causes the client process unable to read the root cert due to permission issue. The last two commits are to fix that. 

By the way, also feel free to review the file permission related code, very little code in `daemon_config.cpp`. The whole idea is calling `restrict_permissions` on the data folder first, and overwrite the permissions of root cert related files and folders. It includes data folder itself, `certificates` sub folder and `multipass_root_cert.pem` file itself. 

The functional testing is the same, includes multipass upgrade, new installation and plus the server restart.

Sploder12 (dismissed):
LGTM, good catch with the restart issue! This time I tested the restart as well. Everything looks good to me on the Linux side of things!

src/cert/ssl_cert_provider.cpp
```diff
@@ -312,6 +312,13 @@ mp::SSLCertProvider::KeyCertificatePair make_cert_key_pair(const QDir& cert_dir,
         const std::filesystem::path root_cert_path = MP_PLATFORM.get_root_cert_path();
         if (std::filesystem::exists(root_cert_path) && QFile::exists(priv_key_path) && QFile::exists(cert_path))
         {
+            // In cases where the root certificate exists within the data folder (such as in Snap or Windows
+            // environments), its file permissions is unintentionally overwritten due to recursive permission
+            // settings applied to the entire folder. To ensure the client process has access, we need to explicitly
+            // reset the file permissions on the root certificate.
```
ricab, line 318: Hmm I don't think we need all these comments. They are valuable for the PR, not so much for future readers of the code IMO. I mean, permissions are no longer "unintentionally overwritten" once this is merged, right? How about something simpler, like

```suggestion
            // Unlike other daemon files, the root certificate needs to be accessible by everyone
```

ricab (approved):
Thanks Jia, just a small comment, but otherwise looks good. Approving modulo suggestion

xmkg (approved):
Re-approving the latest changes.

xmkg:
The merge queue pipeline should be fixed once https://github.com/canonical/multipass/pull/4017 lands.

PR #3913, opened 2025-01-31
[format] Ensure proper ending of the last line

ricab:
Tiny PR to ensure that we end files with a newline. See the [POSIX definition of a line](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_185).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,41 +2,41 @@
 # We'll use defaults from the LLVM style, but with 4 columns indentation.
 BasedOnStyle: LLVM
 ---
-Language: Cpp
-IndentWidth: 4
-IndentExternBlock: NoIndent
-BreakBeforeBraces: Custom
-BinPackArguments: false
-BinPackParameters: false
+AccessModifierOffset: -4
 AllowAllArgumentsOnNextLine: false
 AllowAllParametersOfDeclarationOnNextLine: false
-PenaltyReturnTypeOnItsOwnLine: 120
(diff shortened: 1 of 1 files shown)


andrei-toterman (dismissed):
Yup, tested and it works as expected. Thanks, @ricab!
A question for the future: wouldn't it be nicer if the config options were sorted alphabetically?

ricab:
It would indeed. I thought about it but resisted the perfectionism here for a change :sweat_smile:

ricab:
I think that is it, but I wouldn't mind another set of eyes, as I just pieced it together with `sort`.

ricab:
I wonder if we should place Language and Standard first in the section though, with a newline in between.

.clang-format
```diff
@@ -3,39 +3,41 @@
 BasedOnStyle: LLVM
 ---
 Language: Cpp
-IndentWidth: 4
-IndentExternBlock: NoIndent
-BreakBeforeBraces: Custom
-BinPackArguments: false
-BinPackParameters: false
+Standard: Cpp11
```
xmkg, line 6: Unrelated to this PR's changes but, is there a reason why this is still set to Cpp11 instead of Cpp17?

ricab: Not that I know of. I'll make a separate PR for that.

xmkg (approved):

PR #3922, opened 2025-02-07
add daemon gRPC boilerplate for AZs

andrei-toterman:
This PR adds AZ information to RPC messages that need it, as well as methods for listing AZs and changing their state.

### DIFF
--- a/src/daemon/daemon_rpc.cpp
+++ b/src/daemon/daemon_rpc.cpp
@@ -418,6 +418,28 @@ grpc::Status mp::DaemonRpc::daemon_info(grpc::ServerContext* context,
         client_cert_from(context));
 }
 
+grpc::Status mp::DaemonRpc::zones(grpc::ServerContext* context,
+                                  grpc::ServerReaderWriter<ZonesReply, ZonesRequest>* server)
+{
+    ZonesRequest request;
+    server->Read(&request);
+
+    return verify_client_and_dispatch_operation(
+        std::bind(&DaemonRpc::on_zones, this, &request, server, std::placeholders::_1),
+        client_cert_from(context));
+}
+
+grpc::Status mp::DaemonRpc::zones_state(grpc::ServerContext* context,
+                                        grpc::ServerReaderWriter<ZonesStateReply, ZonesStateRequest>* server)
+{
+    ZonesStateRequest request;
+    server->Read(&request);
+
+    return verify_client_and_dispatch_operation(
+        std::bind(&DaemonRpc::on_zones_state, this, &request, server, std::placeholders::_1),
+        client_cert_from(context));
+}
+
 template <typename OperationSignal>
 grpc::Status mp::DaemonRpc::verify_client_and_dispatch_operation(OperationSignal signal, const std::string& client_cert)
 {
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -579,6 +579,8 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
     QObject::connect(&rpc, &mp::DaemonRpc::on_snapshot, &daemon, &mp::Daemon::snapshot);
     QObject::connect(&rpc, &mp::DaemonRpc::on_restore, &daemon, &mp::Daemon::restore);
     QObject::connect(&rpc, &mp::DaemonRpc::on_daemon_info, &daemon, &mp::Daemon::daemon_info);
+    QObject::connect(&rpc, &mp::DaemonRpc::on_zones, &daemon, &mp::Daemon::zones);
+    QObject::connect(&rpc, &mp::DaemonRpc::on_zones_state, &daemon, &mp::Daemon::zones_state);
 }
 
 enum class InstanceGroup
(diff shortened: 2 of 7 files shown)


src/daemon/daemon_rpc.cpp
```diff
@@ -418,6 +418,28 @@ grpc::Status mp::DaemonRpc::daemon_info(grpc::ServerContext* context,
...
 }
 
+grpc::Status mp::DaemonRpc::zones(grpc::ServerContext* context,
+                                  grpc::ServerReaderWriter<ZonesReply, ZonesRequest>* server)
+{
+    ZonesRequest request;
+    server->Read(&request);
+
+    return verify_client_and_dispatch_operation(
+        std::bind(&DaemonRpc::on_zones, this, &request, server, std::placeholders::_1),
```
xmkg, line 428: Is there a reason why we favor std::bind over lambda here?

andrei-toterman: Yup, that's the only reason. Just following the beaten path.

andrei-toterman: +1 for the idea that `std::bind` expresses intent better. with lambdas, you always have to read it to understand what it does. with `std::bind`, you just know immediately you're doing partial function application

xmkg: I agree that `std::bind` might express the intent better for a trained eye, but it has so many subtle gotcha's that outweigh it. The most prominent one is the following (quoting from [abseil/TOTW#108](https://abseil.io/tips/108)) :

```C++
void DoStuffAsync(std::function<void(absl::Status)> cb);

class MyClass {
  void Start() {
    DoStuffAsync(std::bind(&MyClass::OnDone, this));
  }
  void OnDone();  // No absl::Status here.
};
```

Here, the code compiles fine even though `OnDone` has no arguments (except for the implicit `this`). There are a couple more examples in the post for those interested.

> With a lambda, it's mixed in the body, which you have to read to find out what else there is to it

If we're strictly talking about a lambda replacing the std::bind equivalent, that's not the case. In nearly all of the cases the body is just a one-liner. I also argue that std::bind is more confusing-looking, and less known by the newer audiences since many projects stopped using it altogether. Some projects still rely on std::bind, but it's often for the unavailability of the lambda features needed (e.g. C++11 compat mode). For me, it's a relic from the past.

> I hear lambdas are more efficient, which is mildly surprising since lambdas could be doing many more things

Lambdas are just anonymous structs. It's easy for the compiler to reason about, optimize them, and detect errors. std::bind's inner workings might hinder the compiler's ability to detect errors or optimize the code.

> It is probably also easier for subtle changes to go unnoticed (e.g. reference types and the like)

I think if that is a concern, you can be explicit in the lambda's capture list to eliminate that possibility.

> Again, if that is the case, shouldn't that be a concern of the standard library and implementers rather than the user

I agree completely, but the reality we live in is different. In the C++ realm once something is in place, it's most often hard to replace it with a better one, even when that means *zero* external API change. There are many things the language has to carry along just to prevent ABI breakage and other shenanigans (see: std::regex). I'm a strong believer in "leaving things behind", "setting bars" or "replacing the broken", and I think C++ would be a better, safer, and cleaner language if we could ditch the things that are simply wrong... but it's easier said than done. See the whole ordeal with the auto_ptr -- it took 6 years to deprecate and remove it. 

> We pass unnamed arguments all the time.

I advocate against passing unnamed arguments as well. Explicit is almost always better than implicit for the readers. The compiler is indifferent to an alias, but it'll make a human's reasoning easier.

> I wonder, if for loops were introduced after std::copy, if people would now be advocating replacing copy calls with plain fors

Well, "for" loops are a low-level language construct and not exactly an apples-to-apples comparison with the lambdas, but I see your point. I agree that we shouldn't blindly ditch things and start using new shiny things because they're said to be *new and better*.  I've used std::bind and lambdas, and I can't recall a single instance that I thought "hmm, std::bind could be a better choice here", but that's just my opinion.

I like lambdas because it's explicit. You have to specify the capture list, the parameters, and precisely what to do with them. I also know that the lambdas are just anonymous structs with a call operator -- the capture list goes to the constructor, arguments go to the call operator;  easy to visualize what's going on Sure, it might burden the reader more than the `std::bind` at first glance, but if the code follows a pattern or uses a certain kind of API (e.g. gRPC), the reader can easily reason about the function of the lambda. Consider the following change:

```C++
return verify_client_and_dispatch_operation(
         // with std::bind:
         std::bind(&DaemonRpc::on_zones, this, &request, server, std::placeholders::_1),
         // with a lambda:
         [this](ArgT arg){ on_zones(&request, server, arg); },
         /*...*/
``` 
Breakdown:        
```C++                
         // Explicit capture of "this", expresses what exactly is going to be used from the other scopes
         [this]
         // Clear list of arguments with proper type and name, instead of placeholders. 
         (ArgT arg)
         // No confusing member function binding syntax, no placeholders. just calling a function with
         // the available arguments at the scope
         { on_zones(&request, server, arg); }
```

IMO, the lambda version is; 

a) shorter
b) explicit
c) easy to reason about
d) has much less potential gotcha's

Back to our case; I don't have any strong opinions about using std::bind, and rewriting the code in lambda terms might prove little benefit at this point. I think there's a middle-ground and it's called "std::bind_front", but unfortunately it's a C++20 feature. Maybe we can discuss it again when we make the switch to C++20.

ricab: OK, you're right that bind has a few shortcomings that I did after all not pay enough attention to, the one you highlight being the most concerning. Thanks for the abseil link BTW :slightly_smiling_face: 

I still don't quite agree on the readability front, for the reason that Andrei put so succinctly. I am all for `bind_front` though, when we get it (shouldn't be long now, since we're moving to newer Apple versions). Its ability to perfect-forward I believe will tip that scale clearly in its favor. See [this lambda-based case in the same file](https://github.com/canonical/multipass/blob/e63104e3a88e842466a7e02270f0c57257b34fd2/src/daemon/daemon_rpc.cpp#L209-L211) for comparison. That is neither shorter nor easier to reason about, I don't think.

A larger point that I would like to address though: I also tend to favor explicit vs implicit, but only for a certain interpretation of "explicit", and then not at all costs. It has limits. One is being synthetic. If you always avoid passing rvalues you end up with a bunch of artificial code, diluting the important bits. Information density is important too. Sometimes it is better to extract a variable and name it, of course, but often it is not. In essence, these values are not absolute, so we still need to find that crafty balance (which includes a degree of subjectivity). Surely you wouldn't argue for not writing functions and having all the code inline because it is "explicit". That's not the sort of "explicit" we're after. You wouldn't defend reimplementing things that the standard library offers either (at least not in general). Not inlining the "partial function application" is sort of the same. In a sense, `bind` can be seen as *more* explicit, just like `copy` vs a plain `for`. It is explicit about the intent — only binding/copying and nothing else — unlike with the lower-level constructs (lambdas/fors), which afford many other uses. That is the sense of my earlier comparison. They're all apples in this regard (they are not the same thing, but if logical argument was limited to comparing identical things, comparisons wouldn't be very useful. It's is enough that they share the features that the comparison illustrates). 

Anyway, I concede the main point: we should probably avoid bind by default, because of its pitfalls. In the present case, I think either way is fine. We could perhaps add a TODO to change to `bind_first` once we get C++20.

Sploder12 (approved):
LGTM!

xmkg:
Hi @andrei-toterman, I see that you've rebased this branch to main. Could you also rebase the availability-zones so the diff would only show the changes introduced in this branch?

andrei-toterman:
ah, right, good catch. thanks for that @xmkg! there are so many stacked branches here that i sometimes forget the proper ordering :sweat_smile:

xmkg:
> ah, right, good catch. thanks for that @xmkg! there are so many stacked branches here that i sometimes forget the proper ordering 😅

You and me, both :) Trust me, I learned it the hard way..

ricab:
@andrei-toterman, is this good to go? Might as well merge, no?

andrei-toterman:
Yes :pray:

PR #3923, opened 2025-02-07
Az definitions

andrei-toterman:
This PR adds interfaces for AZs and for an AZ manager, as well as some AZ specific exceptions.

### DIFF
--- a/include/multipass/availability_zone.h
+++ b/include/multipass/availability_zone.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_AVAILABILITY_ZONE_H
+#define MULTIPASS_AVAILABILITY_ZONE_H
+
+#include "disabled_copy_move.h"
+#include "virtual_machine.h"
+
+#include <memory>
+
+namespace multipass
+{
+class AvailabilityZone : private DisabledCopyMove
+{
+public:
+    using UPtr = std::unique_ptr<AvailabilityZone>;
+    using ShPtr = std::shared_ptr<AvailabilityZone>;
+
+    virtual ~AvailabilityZone() = default;
+
+    [[nodiscard]] virtual const std::string& get_name() const = 0;
+    [[nodiscard]] virtual const std::string& get_subnet() const = 0;
+    [[nodiscard]] virtual bool is_available() const = 0;
+    virtual void set_available(bool new_available) = 0;
+    virtual void add_vm(VirtualMachine& vm) = 0;
+    virtual void remove_vm(VirtualMachine& vm) = 0;
+};
+} // namespace multipass
+
+#endif // MULTIPASS_AVAILABILITY_ZONE_H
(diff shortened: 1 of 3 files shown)


include/multipass/exceptions/availability_zone_exceptions.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include <multipass/format.h>
+
+namespace multipass
+{
+struct AvailabilityZoneError : std::runtime_error
+{
+    template <typename... Args>
+    explicit AvailabilityZoneError(const char* fmt, Args&&... args)
+        : runtime_error(fmt::format(fmt, std::forward<Args>(args)...))
```
xmkg, line 32: Just so you know, moving fmt::format to here disables the compile-time format argument validation, e.g. type or count in c++20 and above.

xmkg (approved):

include/multipass/exceptions/availability_zone_exceptions.h
xmkg: It boils down to fmt::format's "fmt" argument not getting a string literal directly, i.e. via a "const char *". Therefore, it cannot decide whether the format string is a compile-time constant or a runtime one. Therefore, fmt would require a `fmt::runtime()` wrap around the fmt parameter(=> c++20), which means runtime evaluation of the format string (which is the default in '< c++20').

fmt in pre-c++20 supports compile-time format string evaluation by several macro helpers, such as `FMT_STRING` and `FMT_COMPILE`:

https://fmt.dev/11.1/api/#legacy-compile-time-checks
https://fmt.dev/11.1/api/#format-string-compilation

While researching the topic I found a solution that works for both pre/post c++20:

https://github.com/fmtlib/fmt/issues/2391

So, replacing the `const char* fmt` with `fmt::format_string<Args...> fmt` would future-proof this code, allowing it to benefit from compile-time checks when c++20 arrives. It should behave exactly as if it's a "const char*" for c++17.

andrei-toterman: Thanks for the details! That's very cool!

levkropp (approved):
Looks good to me! Good defs and good comments explaining the difference between `get_automatic_zone_name` and `get_default_zone_name`  😀

include/multipass/availability_zone.h
```diff
@@ -0,0 +1,45 @@
...
+#define MULTIPASS_AVAILABILITY_ZONE_H
+
+#include "disabled_copy_move.h"
+#include "virtual_machine.h"
+
+#include <memory>
+
+namespace multipass
+{
+class AvailabilityZone : DisabledCopyMove
```
ricab, line 28: Not a review, I just noticed this: private inheritance would be preferable in this case (and in the manager).

ricab: https://www.boost.org/doc/libs/1_87_0/libs/core/doc/html/core/noncopyable.html

andrei-toterman: Isn't `private` the default in this case?

xmkg: surprisingly, [nope](https://gcc.godbolt.org/z/nbPGEc4r8).

andrei-toterman: @xmkg but we're dealing with classes here, not with structs https://gcc.godbolt.org/z/o5Mdbnsxs

include/multipass/availability_zone_manager.h
```diff
@@ -0,0 +1,44 @@
...
+class AvailabilityZoneManager : DisabledCopyMove
+{
+public:
+    using UPtr = std::unique_ptr<AvailabilityZoneManager>;
+    using ShPtr = std::shared_ptr<AvailabilityZoneManager>;
+
+    virtual ~AvailabilityZoneManager() = default;
+
+    virtual AvailabilityZone& get_zone(const std::string& name) = 0;
+    virtual std::vector<std::reference_wrapper<const AvailabilityZone>> get_zones() = 0;
```
ricab, line 34: I wouldn't mind an include for the reference_wrapper, nor for vector, string, and the smart ptrs while you're at it.

include/multipass/exceptions/availability_zone_exceptions.h
xmkg: We can now use the [FormattedExceptionBase<>](https://github.com/canonical/multipass/blob/f32e3a0cb3c0bbddcb64219e4244de72156752e3/include/multipass/exceptions/formatted_exception_base.h#L43) as a base class to this.

xmkg (approved):

PR #3925, opened 2025-02-07
[cli] add retrieval of zones

andrei-toterman:
This PR adds the `zones` CLI command, for retrieving zones and their status.
It also adds AZ information to the `list` and `info` commands.
There is support for all the expected `--format` options.

There is also a change to the base `Command` class, in order to allow it to receive gRPC callbacks with `const` parameters.


tests/test_output_formatter.cpp
```diff
@@ -80,11 +83,17 @@ auto construct_multiple_instances_list_reply()
     list_entry->mutable_instance_status()->set_status(mp::InstanceStatus::RUNNING);
     list_entry->set_current_release("16.04 LTS");
     list_entry->add_ipv4("10.21.124.56");
+    auto zone = list_entry->mutable_zone();
+    zone->set_name("zone");
+    zone->set_available(true);
```
xmkg, line 88: I see that set_available is true for all the cases. Can we also make one of them false to verify the "unavailable" scenario?

xmkg (changes requested):

xmkg (approved):
LGTM

levkropp (approved):
LGTM. I pushed a change to test_client_cli that adds coverage for zones.cpp

```
Processing file src/client/cli/cmd/zones.cpp
  lines=31 hit=30 functions=8 hit=7
```

ricab:
@levkropp this can now be merged, right?

levkropp:
@ricab still waiting for your review on https://github.com/canonical/multipass/pull/3926 then yes, everything except the GUI is all ready! :smile:

ricab:
I just got back to #3926, then I noticed this. Do you think in this case we should merge that one into this first? Normally we merge in sequence.

PR #3926, opened 2025-02-07
[cli] add enable/disable-zones commands

andrei-toterman:
This PR adds the CLI commands `enable-zones` and `disable-zones` which send the gRPC calls for making an AZ available or unavailable.


src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    ZonesStateRequest request{};
+    for (const auto& zone_name : zone_names)
+        request.add_zones(zone_name);
+    request.set_available(false);
+    request.set_verbosity_level(parser->verbosityLevel());
+
+    const auto on_success = [&](const ZonesStateReply&) {
+        spinner.stop();
+        cout << (zone_names.size() == 1 ? "Zone" : "Zones") << " disabled successfully" << std::endl;
```
xmkg, line 57: We can also use something like `<N> zone(s) disabled successfully.`, so the user can verify that it matches the amount of zones requested.

xmkg (approved):
LGTM

src/client/cli/cmd/disable_zones.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+    using Command::Command;
+    ReturnCode run(ArgParser* parser) override;
+    std::string name() const override;
+    QString short_help() const override;
+    QString description() const override;
+
+private:
+    bool ask_for_confirmation = true;
+    std::vector<std::string> zone_names{};
```
ricab, line 36: Why the separate fields?

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+QString EnableZones::short_help() const
+{
+    return QStringLiteral("Make zones available");
+}
+
+QString EnableZones::description() const
+{
+    return QStringLiteral("Makes the requests availability zones available.");
```
ricab, line 73:
```suggestion
    return QStringLiteral("Makes the given availability zones available.");
```

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,110 @@
...
+    ZonesStateRequest request{};
+    for (const auto& zone_name : zone_names)
+        request.add_zones(zone_name);
+    request.set_available(false);
+    request.set_verbosity_level(parser->verbosityLevel());
+
+    const auto on_success = [&](const ZonesStateReply&) {
+        spinner.stop();
+        const auto n_zones = zone_names.size();
+        cout << n_zones << (n_zones == 1 ? " zone" : " zones") << " disabled successfully" << std::endl;
```
ricab, line 58: I tried a few commands to see what they say on success:

1. `start`, `stop`, `delete`, `recover` or `purge`, `set`, `mount`, `umount` don't print anything.
2. `launch` prints `Launched: <instance>`. 
3. `snapshot` says `Snapshot taken: <instance>.<snapshot>`.
4. `restore` prints `Snapshot restored: <instance>.<snapshot>`

In 2 and 3, the success message is used to inform the user of the name that was chosen. 4 falls a little off of this scheme. 

I don't mind being a little more informative. It may come in handy if we ever implement wildcards. But could we stick to the same format, i.e. something like `[Zone[s]] Disabled: <zones>`?

ricab (commented):
Hey @andrei-toterman, looks good overall. Only small comments and nitpicks.

src/client/cli/cmd/disable_zones.h
andrei-toterman: I'm not sure I understand what you mean. Those fields are used to store the result of parsing the CLI args.

src/client/cli/cmd/disable_zones.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#ifndef MULTIPASS_DISABLE_ZONES_H
+#define MULTIPASS_DISABLE_ZONES_H
+
+#include <multipass/cli/command.h>
+
+namespace multipass::cmd
+{
+class DisableZones final : public Command
```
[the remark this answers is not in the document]
ricab: Almost, but not quite :smirk: 

- https://github.com/canonical/multipass/blob/3991d38ae578c1843631c2266d158e18a885b497/src/client/cli/cmd/snapshot.h#L25 
- https://github.com/canonical/multipass/blob/f4f5e9bd25d0dd5a5f3e4542b2a71719d22eebe3/src/client/cli/cmd/restore.h#L25

I know other commands have it, but I don't see a reason for it, so decided to ask. Not a big deal, but FWIW https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-final

src/client/cli/cmd/disable_zones.h
andrei-toterman: Ah, I see. My logic was that the parsing code should not be concerned with how the arguments are used, so it should not know about any gRPC stuff, and just create an intermediate representation that can then be used by `run` to do the gRPC stuff. I realize that this is too pedantic, so I'll simplify it.

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,110 @@
...
+        return Ok;
+    };
+
+    const auto on_failure = [this, &spinner](const grpc::Status& status) {
+        spinner.stop();
+        return standard_failure_handler_for(name(), cerr, status);
```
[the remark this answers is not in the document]
andrei-toterman: I would say with which VMs failed to comply. Making an AZ (un)available involves two steps:
- toggling the AZ status between available and unavailable. there shouldn't be any points of failure here (besides writing the status to disk, which of course should be reported, but i don't expect that to be common failure point)
- telling the VMs to become (un)available, which involves stopping/starting them as needed, and here each VM could fail individually, so we can report which failed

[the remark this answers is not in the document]
ricab: How about AZs that failed to fully disable in the status, with VMs in the details?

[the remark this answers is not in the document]
andrei-toterman: Sure, sounds good.

ricab (commented):
Looking good. Asking only for some polish.

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+QString DisableZones::description() const
+{
+    return QStringLiteral("Makes the given availability zones unavailable.");
+}
+
+ParseCode DisableZones::parse_args(ArgParser* parser)
+{
+    parser->addPositionalArgument("zone", "Name of the zones to make unavailable", "<zone> [<zone> ...]");
```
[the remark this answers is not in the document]
levkropp:
```
$ ./bin/multipass zones
Name    State
zone1   Available
zone2   Available
zone3   Available
$ ./bin/multipass disable-zones 
Name argument or --all is required
$ ./bin/multipass disable-zones --all
This operation will forcefully stop the VMs in zone1, zone2 and zone3. Proceed? (Yes/No): No 
$ ./bin/multipass disable-zones --all --force
All zones disabled
$ ./bin/multipass zones
Name    State
zone1   Unavailable
zone2   Unavailable
zone3   Unavailable
```

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+        };
+
+        auto zones_on_failure = [this](const grpc::Status& status) {
+            return standard_failure_handler_for(name(), cerr, status);
+        };
+
+        // First, get all zones
+        if (const auto ret = dispatch(&RpcMethod::zones, zones_request, zones_on_success, zones_on_failure); ret != Ok)
+            return ret;
+    }
```
ricab: Hmm, I don't think we need the back and forth. In other places where we have `--all` we signal that to the daemon through gRPC, usually by leaving the list of targets empty. The daemon just needs to know to targets all things if no zones are specified. See for example `Daemon::start` and the instance selection code.

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (const auto ret = parse_args(parser); ret != ParseCode::Ok)
+        return parser->returnCodeFrom(ret);
+
+    // If no zones exist, just inform the user and exit successfully
+    if (request.zones_size() == 0)
+    {
+        cout << "No zones exist" << std::endl;
+        return ReturnCode::CommandFail;
+    }
+
```
ricab: I find that "no zones exist" is a little misleading. This just means that no zones were specified on the command line. Which, by the way, is perfectly acceptable with `--all`.

ricab: And commands with neither zone arguments nor `--all` should result in a parsing error, so better handle that in the parsing method.

src/client/cli/cmd/enable_zones.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+    using Command::Command;
+    ReturnCode run(ArgParser* parser) override;
+    std::string name() const override;
+    QString short_help() const override;
+    QString description() const override;
+
+private:
+    ZonesStateRequest request{};
+    bool use_all_zones{false};
```
ricab: If we follow the approach of signaling all to the daemon, this field can be removed. The request needs to have all the info by the time parsing finishes.

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+        };
+
+        // First, get all zones
+        if (const auto ret = dispatch(&RpcMethod::zones, zones_request, zones_on_success, zones_on_failure); ret != Ok)
+            return ret;
+    }
+
+    AnimatedSpinner spinner{cout};
+    const auto message =
+        use_all_zones ? "Enabling all zones" : fmt::format("Enabling {}", fmt::join(request.zones(), ", "));
```
ricab, line 38: After dropping `use_all_zones`, you can interrogate `request.zones()` instead.

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+QString EnableZones::short_help() const
+{
+    return QStringLiteral("Make zones available");
+}
+
+QString EnableZones::description() const
+{
+    return QStringLiteral(
+        "Makes the given availability zones available (any previously running VMs are started again).");
```
ricab:
```suggestion
        "Makes the given availability zones available. Instances therein are started if they were running when their zone was last disabled.");
```

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "enable_zones.h"
+
+#include "animated_spinner.h"
+#include "common_callbacks.h"
+#include "common_cli.h"
+#include "multipass/cli/argparser.h"
+#include "multipass/cli/client_common.h"
+
+#include <QCommandLineOption>
```
ricab, line 26: This comes with argparser.h above.

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    parser->addPositionalArgument("zone", "Name of the zones to make available", "<zone> [<zone> ...]");
+
+    QCommandLineOption all_option(all_option_name, "Enable all zones");
+    parser->addOption(all_option);
+
+    if (const auto status = parser->commandParse(this); status != ParseCode::Ok)
+        return status;
+
+    if (const auto status = check_for_name_and_all_option_conflict(parser, cerr); status != ParseCode::Ok)
```
ricab, line 87: :+1:

src/client/cli/cmd/enable_zones.cpp
ricab: Oh that's taken care of by `check_for_name_and_all_option_conflict`.

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,170 @@
...
+    const auto format_zones = [this] {
+        if (request.zones_size() == 1)
+            return request.zones(0);
+
+        const auto last_zone = request.zones_size() - 1;
+        return fmt::format("{} and {}",
+                           fmt::join(request.zones().begin(), request.zones().begin() + last_zone, ", "),
+                           request.zones(last_zone));
+    };
+    const auto message = "This operation will forcefully stop the VMs in " + format_zones() + ". Proceed? (Yes/No)";
```
ricab: The capital usually indicates the default option. Keeping consistency with `delete`, this could say
```suggestion
    const auto message = "This operation will forcefully stop the VMs in " + format_zones() + ". Are you sure you want to continue? (Yes/no)";
```

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    const PlainPrompter prompter{term};
+    auto answer = prompter.prompt(message);
+    while (!std::regex_match(answer, client::yes_answer) && !std::regex_match(answer, client::no_answer))
+        answer = prompter.prompt("Please answer (Yes/No)");
```
ricab, line 137: And accept an empty answer to mean Yes (same as `delete`).

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+namespace multipass::cmd
+{
+ReturnCode DisableZones::run(ArgParser* parser)
+{
+    if (const auto ret = parse_args(parser); ret != ParseCode::Ok)
+        return parser->returnCodeFrom(ret);
+
+    // If no zones exist, just inform the user and exit successfully
+    if (request.zones_size() == 0)
```
ricab: idem

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+        cout << "No zones exist" << std::endl;
+        return ReturnCode::CommandFail;
+    }
+
+    // If --all was specified, we need to fetch all zone names first
+    if (use_all_zones)
```
ricab: idem

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,170 @@
...
+QString DisableZones::short_help() const
+{
+    return QStringLiteral("Make zones unavailable");
+}
+
+QString DisableZones::description() const
+{
+    return QStringLiteral("Makes the given availability zones unavailable (all VMs are switched off immediately and "
+                          "forcefully and cannot be interacted with until the zone is available again, simulating a "
+                          "loss of availability on a cloud provider).");
```
ricab:
```suggestion
    return QStringLiteral("Makes the given availability zones unavailable. Instances therein are forcefully switched off and remain unavailable until their zone is re-enabled (simulating a "
                          "loss of availability on a cloud provider).");
```

src/client/cli/cmd/enable_zones.cpp
levkropp: ~~This is not checking for arguments in the command line, but checking how many zones there actually are defined in the az_manager.~~

:bell: I was wrong :bell:

src/client/cli/cmd/disable_zones.cpp
levkropp: This suggested change results in a line that is 190 characters long. I'll switch the wording and add the parentheses but it will need to stay as 3 lines

src/client/cli/cmd/disable_zones.cpp
levkropp: This suggestion results in a line too large for `clang-format-16`, it wants to see

```
    const auto message = "This operation will forcefully stop the VMs in " + format_zones() +
                         ". Are you sure you want to continue? (Yes/no)";
```
Is this acceptable as well?

src/client/cli/cmd/disable_zones.cpp
ricab: I know, I skipped the formatting. Thanks.

src/client/cli/cmd/disable_zones.cpp
ricab: Yes of course :)

src/client/cli/cmd/enable_zones.cpp
ricab: Could you please explain how `request.zones_size()` tells you that?

Even if it did, this doesn't make much sense to me, because:

- that's not a concern for the client
- there are always going to be zones (we established 3)
- multipass commands are idempotent, so if no zones need to be enabled (i.e. they are all enabled as it is) the command should succeed.

levkropp: Not only was I confidently incorrect (shame on me :bell:), but this code is never even reached because of the `check_for_name_and_all_option_conflict` check you mentioned. I'll remove it from both enable_ and disable_zones

src/client/cli/cmd/enable_zones.cpp
levkropp: I chose to keep it around as a local variable because we can use it in both the pre-success and post-success message conditionals

```c++
    const auto use_all_zones = request.zones().empty();
    const auto message =
        use_all_zones ? "Disabling all zones" : fmt::format("Disabling {}", fmt::join(request.zones(), ", "));
    spinner.start(message);

    const auto on_success = [&](const ZonesStateReply&) {
        spinner.stop();
        const auto output_message = use_all_zones ? "All zones disabled"
                                                  : fmt::format("Zone{} disabled: {}",
                                                                request.zones_size() == 1 ? "" : "s",
                                                                fmt::join(request.zones(), ", "));
        cout << output_message << std::endl;
        return Ok;
    };
```

But we can use `request.zones().empty();` in both places if you prefer

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,141 @@
...
+        for (const auto& zone_name : parser->positionalArguments())
+            request.add_zones(zone_name.toStdString());
+    }
+
+    ask_for_confirmation = !parser->isSet(forceOption);
+
+    return ParseCode::Ok;
+}
+
+bool DisableZones::confirm()
```
ricab: This could be `const`, right?

src/daemon/daemon.cpp
```diff
@@ -2918,9 +2918,20 @@ try // clang-format on
 {
     mpl::ClientLogger logger{mpl::level_from(request->verbosity_level()), *config->logger, server};
 
-    for (const auto& zone_name : request->zones())
+    if (request->zones().empty())
     {
-        config->az_manager->get_zone(zone_name).set_available(request->available());
+        auto& az_manager = *config->az_manager;
```
ricab: If you want to keep this, better extract it outside the if and use it in both branches homogeneously.

src/client/cli/cmd/disable_zones.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+QString DisableZones::short_help() const
+{
+    return QStringLiteral("Make zones unavailable");
+}
+
+QString DisableZones::description() const
+{
+    return QStringLiteral("Makes the given availability zones unavailable. Instances therein are"
```
ricab:
```suggestion
    return QStringLiteral("Makes the given availability zones unavailable. Instances therein are "
```

src/client/cli/cmd/enable_zones.cpp
```diff
@@ -0,0 +1,101 @@
...
+                                                  : fmt::format("Zone{} enabled: {}",
+                                                                request.zones_size() == 1 ? "" : "s",
+                                                                fmt::join(request.zones(), ", "));
+        cout << output_message << std::endl;
+        return Ok;
+    };
+
+    const auto on_failure = [this, &spinner](const grpc::Status& status) {
+        spinner.stop();
+        return standard_failure_handler_for(name(), cerr, status);
```
ricab, line 53: If I use a bad AZ, I get this error message:

```
$ multipass enable-zones asdf
enable-zones failed: no AZ with name "asdf" found
```

Could we have another PR to avoid "AZ" in messages to the user? Let's stick to "zone" as a single nomenclature. "Availability zone" in the help text and docs is fine though IMO.

PR #3931, opened 2025-02-11
Migrate to org GitHub to Jira application

didrocks:
We have a new GitHub to Jira global organization application. The previous gh workflow action is now deprecated.

UDENG-5960


.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,7 @@
+settings:
+  jira_project_key: MULTI
+  labels:
+    - jira
+  status_mapping:
+    opened: Untriaged
+    closed: Done
```
ricab, line 7: I guess `not_planned: rejected` would make sense too?

ricab, line 7: Perhaps the label mapping from the example in the readme would make sense too?

```
  label_mapping:
    enhancement: Story
```

.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,7 @@
+settings:
```
ricab, line 1: Could we perhaps have a top comment with some reference to docs or, failing that, [the GH](https://github.com/canonical/gh-jira-sync-bot)? Just so that future readers know what this is about.

ricab (commented):
Thank you Didier! This looks reasonable to me. I have only a few questions below and small suggestions inline.

- Out of curiosity, what is the motivation for the change?
- I don't suppose there is a way we could test this before merging?
- Is this all there is to it, or is anything else needed on Jira's court? How about "GitHub App installation"?
- I suppose after this we can remove `secrets.JIRA_WEBHOOK_URL`, right? I didn't follow the current setup very closely TBH... is there anything else that we should be cleaning up, do you know?

.github/.jira_sync_config.yaml
didrocks: see my global comment.

didrocks: see my global comment.

.github/.jira_sync_config.yaml
didrocks: That’s a good suggestion, Let me add a top comment with the repo URL.

ricab (approved):
Looking good, thanks again Didier!

PR #3932, opened 2025-02-11
[snapcraft] Scrape files with LICENSE or full path contains LICENSE

sharder996:
Scrape license files that either:
- Have a name starting with LICENSE, or
- Have a full path that contains /LICENSE/

MULTI-1763


andrei-toterman (approved):
Yup, tested it and it finds the expected license files. Thanks, @sharder996

ricab (approved):
Perfect, thank you so much @sharder996!

Issue #3802, opened 2024-12-02
Issues with 3rd party compilation order


lars18th:
**Describe the bug**
Manually compiling Multipass the build process stops at some point.

**To Reproduce**
Inside one Ubuntu 20.04 container, the build process compiling with CLang stops with this error:
```
-- Looking for openssl/des.h - not found
CMake Error at 3rd-party/libssh/libssh/ConfigureChecks.cmake:80 (message):
  Could not detect openssl/des.h
Call Stack (most recent call first):
  3rd-party/libssh/CMakeLists.txt:51 (include)
```

**Expected behavior**
The problem seems to be related to the _order_ of the 3rd party components:
https://github.com/canonical/multipass/blob/6c4b8b4f23462a58c81b708808782e682627bf9b/CMakeLists.txt#L125-L138

If you check the pre-builded `vcpkg-ports` 3rd party package, it includes custom `grpc` and `poco` packages as well. However, the configuration is not using these libs to compile other 3rd party packages. And the same is true for the `libssh`.

The idea is then: compile first custom `libssh`, `grpc` and `poco`; and after use _only_ these versions to compile all the other dependencies.

I hope you want to improve the compilation process to reduce dependencies and use _only_ the custom build 3rd party libraries. This could facilitate compilation in other environments and reduce compilation time.

giuliazanchi:
Hi @lars18th , thanks for bringing this to our attention, it's definitely worth exploring further to simplify and improve the build process and, possibly, the build instructions. 
Since you seem to have investigated the problem and possible solutions, feel free to open a PR and suggest a fix, we're always happy to review and incorporate contributions from the community.

ljonathan0719:
Hello! Is there any way I could contribute to this issue?

ricab:
Hi @ljonathan0719. Yes, we welcome community contributions, so you can submit a pull request! I suggest starting by finding a simple recipe to reproduce the issue and take it from there (if you haven't yet). Thank you for your interest and let us know if you need further guidance.

ricab:
@ljonathan0719 do you mean to pursue #3939 any further? If not, I guess this would be free @SuhasHareesh.

zizeca:
I think you need to use `add_dependencies(...)` to build in a specific order
[add_dependencies](https://cmake.org/cmake/help/latest/command/add_dependencies.html)

vicdeveloperr:
I don't see movement here for a while.
I'm working on this.

[...]

PR #3939, opened 2025-02-18
Modified the 3rd party compilation order

ljonathan0719:
This pull request addresses Issue #3802 where manually building Multipass fails due to OpenSSL header detection errors in libssh. The issue was caused due to an incorrect dependency order in the CMakeLists.txt file. This resulted in the system's OpenSSL installation verison being used instead of the custom built 3rd party libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,11 +122,18 @@ if(cmake_build_type_lower MATCHES "coverage")
   endif()
 endif()
 
-# OpenSSL config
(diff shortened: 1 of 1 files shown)


Sploder12 (changes requested):
Thank you for your interest and contribution @ljonathan0719! Unfortunately, there are some issues preventing building.
 
Also, would you mind [signing the CLA](https://ubuntu.com/legal/contributors) using the same GitHub handle / same email?

ricab:
@ljonathan0719 do you mean to pursue this any further?

PR #3943, opened 2025-02-20
[tests] add log level to string unit tests

xmkg:
This MP adds log level to string unit test case to ensure all cases are covered.

The MP also includes the commits from https://github.com/canonical/multipass/pull/3944 to resolve cyclic dependency between the two branches causing a CI failure

[giuliazanchi] Fixed link to Multipass Discourse landing page

MULTI-1831

### DIFF
--- a/docs/how-to-guides/customise-multipass/build-multipass-images-with-packer.md
+++ b/docs/how-to-guides/customise-multipass/build-multipass-images-with-packer.md
@@ -58,7 +58,7 @@ Anything you do here will be reflected in the resulting image. You can install p
 
 Go to [Packer's documentation](https://developer.hashicorp.com/packer/docs) to learn about the QEMU builder, other provisioners and their configurations as well as everything else that might come in handy. Alternatively, you could extend `user-data` with other `cloud-init` [modules](https://cloudinit.readthedocs.io/en/latest/reference/modules.html) to provision your image.
 
-Join the discussion on the [Multipass forum](https://discourse.ubuntu.com/c/multipass/) and let us know about your images!
+Join the discussion on the [Multipass forum](https://discourse.ubuntu.com/c/project/multipass/21/) and let us know about your images!
 
 ---
 
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -683,7 +683,7 @@ Congratulations! You can now use Multipass proficiently.
 
 There's more to learn about Multipass and its capabilities. Check out our [How-to guides](/how-to-guides/index) for ideas and help with your project. In our [Explanation](/explanation/index) and [Reference](/reference/index) pages you can find definitions of key concepts, a complete CLI command reference, settings options and more.
 
-Join the discussion on the [Multipass forum](https://discourse.ubuntu.com/c/multipass/) and let us know what you are doing with your instances!
+Join the discussion on the [Multipass forum](https://discourse.ubuntu.com/c/project/multipass/21/) and let us know what you are doing with your instances!
 
 ---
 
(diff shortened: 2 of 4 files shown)


xmkg:
Cherry-picked the commits from https://github.com/canonical/multipass/pull/3944 to resolve the cyclic CI failure between the two branches.

Sploder12 (approved):
LGTM! 

Not really relating to this PR but `mpl::as_string` returning `CString` is interesting, `std::string_view` seems more appropriate.

georgeliao:
It looks good to me as well. `CString` is a in-house utility class which should be removed.

georgeliao (approved):

xmkg:
> LGTM!
> 
> Not really relating to this PR but `mpl::as_string` returning `CString` is interesting, `std::string_view` seems more appropriate.

Yeah, I'm not fond of CString as well. The only reason it works is the [expression temporary lifetime extension](https://en.cppreference.com/w/cpp/language/lifetime):

```c++
void foo(CString f){ }

foo(fmt::format("test")); 
// fmt::format returns a std::string, which has a temporary lifetime. 
// CString has a const std::string& constructor which grabs the return value 
// of .c_str() to assign it to a private const char* member.
// The lifetime of the temporary is extended just because of the expression `foo(...)`.
// The following is UB:

CString x = fmt::format("test");
foo(x);

```

Ideally, the log function should take perfectly forwarded parameters and bind them to something that is *not* temporary.

ricab:
+1 on the CString stuff, but I would prefer avoiding `string_view`s as well. They're unfortunately pretty broken/dangerous in C++. For example, they suffer from the exact same problem that @xmkg pointed out.

Sploder12:
> It's not uncommon that I see std::string_view.c_str() being passed around for a format parameter to "%s", for example, where the correct way to format a string view is to use `%.*s`, which might not be available as a format specifier depending on the platform.

`std::string_view` doesn't have `c_str` for that very reason, but I've seen `data` used for that purpose. Perfect forwarding to the log function would be nice but then it wouldn't be able to be polymorphic. I'd argue the "correct" way to format a string view is to not use a C API to format it.

xmkg:
> `std::string_view` doesn't have `c_str` for that very reason

ah yes, I mean `data()`.

> I'd argue the "correct" way to format a string view is to not use a C API to format it

Well, unfortunately, the format specifiers are language-agnostic. I'd rather prefer an explicit length specifier `%.*s` and not have a `%s` specifier at all. The whole `NUL` Terminator thing always felt like a hack to me.

xmkg:
> Perfect forwarding to the log function would be nice but then it wouldn't be able to be polymorphic

Do we need it to be, though? The perfect forwarding function would act as an interim step that produces the final formatted string and calls the concrete, polymorphic log() function, which would do the actual logging. It should be good enough for mocking purposes.

Sploder12:
> Do we need it to be, though? The perfect forwarding function would act as an interim step that produces the final formatted string and call the concrete, polymorphic log() function, which would do the actual logging. It should be good enough for mocking purposes.

Good point, I think that'd work really well! +1 from me for that approach!

xmkg:
> Good point, I think that'd work really well! +1 from me for that approach!

I might give that a stab on the next pulse -- I need to fiddle with the log() API to add Unicode support for Windows anyway.

xmkg:
> > Good point, I think that'd work really well! +1 from me for that approach!
> 
> I might give that a stab on the next pulse -- I need to fiddle with the log() API to add Unicode support for Windows anyway.

https://github.com/canonical/multipass/pull/3957

PR #3945, opened 2025-02-21
Remove old Discourse footer from all pages

giuliazanchi:
This PR removes the Discourse footer from all pages, linking the Multipass docs feedback google form and the GitHub issue creator link. Since moving docs to RTD, these are no longer necessary. 

EDIT - I also commented out the manual list of contributor that was added to a small set of pages on Discourse, as it was only experimental and will not be maintained further.


xmkg (commented):
I noticed that the contributors are commented out in some parts. Is that intentional?

giuliazanchi:
> I noticed that the contributors are commented out in some parts. Is that intentional?

Hi @xmkg , yes, the reason I hid them is because only some pages included the list of contributors on Discourse, and it was added manually, so it's by no means complete throughout the docs. The original contributors will still be visible on the source Discourse topics and the new contributors will be automatically included in the new Sphinx docs.

I didn't delete them but simply commented these out so that if, for any reason, you want to be able to still see who contributed to those pages, it will remain visible in the source files.

xmkg (approved):

PR #3953, opened 2025-02-24
[CI] run tics on weekly schedule

sharder996:
Adds Tiobe's TICS github action to the Multipass project
- Runs on a weekly basis (the Desktop team [SSDLC guide](https://docs.google.com/document/d/11BLKA6XC_HmrHVJO5qDhurnkq3xYTQBODTL5oEVmoVE/edit?tab=t.gewyardc36dz) suggests nightly)
- Each analysis takes about 4+hrs to run (may vary depending on the number of changed files between runs)
- The TICS custom action has some specific requirements which are not always clear
  - The coverage report must be at `${{ github.workspace }}/coverage/coverage.xml`
  - The project must be built beforehand with standard build tools and build directories
    - Build files should be in `${{ github.workspace }}/build`
    - `make` must be used for compilation (using `ninja` generator will fail the action)
- TICS analysis now includes and analyzes the `tests/` subdirectory (this is inline with similar projects such as Mir)
- Current metrics from test runs can be found [here](https://canonical.tiobe.com/tiobeweb/TICS/TqiDashboard.html#axes=Project(multipass),Sub()&metric=tqi)

---

MULTI-1759

---

Closes #3962

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -17,6 +17,8 @@ jobs:
       VCPKG_FORCE_SYSTEM_BINARIES: 1
       BUILD_DIR: ${{ github.workspace }}/build
       COVERAGE_DIR: ${{ github.workspace }}/coverage
+      NEEDRESTART_SUSPEND: yes
+      DEBIAN_FRONTEND: noninteractive
 
     steps:
     - name: Checkout code
@@ -32,7 +34,7 @@ jobs:
     - name: Install build dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install --yes devscripts equivs lcov ninja-build python3-venv python3-pip
+        sudo apt-get --preserve-env install --yes devscripts equivs lcov ninja-build python3-venv python3-pip
         mk-build-deps -s sudo -i
 
     - name: Configure
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -3,7 +3,7 @@ name: Linux
 on:
     schedule:
     # 0:00 UTC every Sunday
-    - cron: 0 0 * * 0
+    - cron: 11 4 * * 0
 
 jobs:
   build:
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Install build dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install --yes devscripts equivs lcov ninja-build
+        sudo apt-get install --yes devscripts equivs lcov ninja-build python3-venv python3-pip
         mk-build-deps -s sudo -i
 
     - name: Configure
@@ -46,8 +46,18 @@ jobs:
         cmake --build ${{ env.BUILD_DIR }} \
               --target covreport \
               -j$(nproc)
+
+    - name: Convert lcov to Cobertura XML
+      run: |
+        python -m venv .venv
+        source .venv/bin/activate
+        pip install gcovr
+
         mkdir -p ${{ env.COVERAGE_DIR }}
-        mv ${{ env.BUILD_DIR }}/coverage/index.html ${{ env.COVERAGE_DIR }}/coverage.html
+        gcovr -r ${{ github.workspace }} \
+              --object-directory ${{ env.BUILD_DIR }} \
+              --cobertura ${{ env.COVERAGE_DIR }}/coverage.xml \
+              --gcov-ignore-parse-errors
 
     - name: Run TICS Analysis
       uses: tiobe/tics-github-action@v3
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build:
-    runs-on: [self-hosted, reactive, amd64, large-tiobe, noble]
+    runs-on: [self-hosted, reactive, amd64, 2xlarge, noble]
 
     outputs:
       label: ${{ steps.build-params.outputs.label }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tics.yml
```diff
@@ -0,0 +1,72 @@
+name: Linux
+
+on:
+    schedule:
+    # 0:00 UTC every Sunday
+    - cron: 0 0 * * 0
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
Saviq, line 10: You might want to opt for our hosted runners. Should be quicker and more reliable.

```suggestion
    runs-on: [self-hosted, reactive, amd64, large-tiobe, noble]
```

Some other tweaks I needed / added over past weeks:

https://github.com/canonical/mir/compare/0a737ee2c1a1186266391dd8d26096e492587e0f...main#diff-6f8e29420dabe4694a23e1b4d06801de1b54ea481b379216a51d41213296bdf5

ricab:
Hey Scott, it looks like there is an issue with tools folder: https://canonical.tiobe.com/tiobeweb/TICS/TqiDashboard.html#axes=Project(multipass),Sub()&metric=tqiTestCoverage&sel=Sub(tools)

sharder996:
@ricab I think its out of my control. I can see the same "error" in the initial TICS run that was run by Tiobe staff. Comparing between the different time stamps, I've noticed some times slight changes between runs even though the source code was the same. I also Tiobe is actively developing the backend of TICS and analysis results may be subject to changes. The TICS github action used here is simply a client that calls some backend API endpoint.

https://canonical.tiobe.com/tiobeweb/TICS/TqiDashboard.html#axes=Project(multipass),Date(1715703041),Sub()&metric=tqiTestCoverage&sort=T(1,worst)

.github/workflows/tics.yml
ricab: Hey @Saviq, thanks for the heads-up! The diff concerns mainly running in PRs, right? Is there anything else we should note?

@sharder996, did you mean to change the runner in a separate PR?

.github/workflows/tics.yml
```diff
@@ -0,0 +1,59 @@
+name: Linux
+
+on:
+    schedule:
+    # 0:00 UTC every Sunday
+    - cron: 0 0 * * 0
```
ricab, line 6: Can we phase this a little off the round hour? For example `11 4 * * 0`? Still the night from Saturday to Sunday, but 4 hours later and irregular minutes.

sharder996: Sure. Can I ask why though?

ricab: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

> The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. If the load is sufficiently high enough, some queued jobs may be dropped. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

.github/workflows/tics.yml
sharder996: I can change the runner here and any other small tweaks you see.

Saviq: Yes, the most relevant here would be `runs-on` (I actually [tried `2xlarge` since](https://github.com/canonical/mir/actions/runs/14515905858) and it halved the analysis time).

The `--yes`es are required on our self hosted runners - GH probably has tweaked APT config.

.github/workflows/tics.yml
```diff
@@ -31,9 +31,8 @@ jobs:
 
     - name: Install build dependencies
       run: |
-        sudo apt-get update
-        sudo apt-get install -y devscripts equivs lcov
-        mk-build-deps -s sudo -i
+        sudo --preserve-env apt-get install --yes devscripts equivs lcov ninja-build
```
ricab, line 34: Could you please explain why `--preserve-env` was necessary?

ricab: And why not update? Are these runner's apt cache guaranteed to be up to date?

.github/workflows/tics.yml
```diff
@@ -31,9 +31,8 @@ jobs:
...
+        sudo --preserve-env apt-get build-dep --yes ./
```
ricab, line 35: Is this approach preferable to `mk-builds-deps`? Only here or in general? Why?

ricab (commented):
Thanks Scott, I have a few questions.

.github/workflows/tics.yml
sharder996: This is how Mir had it set up so I copied it, but I just tested and it isn't necessary. The variables that `--preserve-env` is protecting isn't related to building Multipass anyways.

.github/workflows/tics.yml
sharder996: Again, how Mir was doing it. Tested and it doesn't make a difference either way. Reverting to stay consistent with the Multipass `README.md`.

ricab: Got it, thanks.

.github/workflows/tics.yml
ricab: OK, makes sense. How about the missing `apt update`?

Saviq: We have this b/c:
https://github.com/canonical/mir/blob/df6674bd3686286aa0e422fdc26664e1fbef5adf/.github/workflows/tics.yml#L24-L25

You might get hit with some dependency that requires configuration, and the install will seem stuck b/c it will wait on user input.

sharder996: No reason to exclude it so I added it back.

.github/workflows/tics.yml
Saviq: It's just avoiding a step, using `apt` itself rather than an intermediary.

Using `mk-build-deps` makes more sense on user systems, as it makes it easier to uninstall. But in CI, it's just a waste.

.github/workflows/tics.yml
ricab: I see, interesting. Thanks @Saviq, is this only a concern in these runners? I suppose GH could be addressing that under the hood.

@sharder996, might be a good idea to do the same then (i.e. avoid apt-get asking what services should be restarted.)

ricab (approved):
Thanks @sharder996, this looks good to me now!

As discussed, we have a little margin to speed up with ccache if we need to in the future. Otherwise just needs a rebase.

PR #3955, opened 2025-02-25
[ci/clang-format] fix the language standard string

xmkg:
The CI lint action does not detect invalid syntax in the .clang-format file if the pull request does not contain any code changes.

This patch corrects the `Standard` value and adds a `clang-format --dump-config` run to ensure that the .clang-format is syntatically valid.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,7 +3,7 @@
 BasedOnStyle: LLVM
 ---
 Language: Cpp
-Standard: Cpp17
+Standard: c++17
 
 AccessModifierOffset: -4
 AllowAllArgumentsOnNextLine: false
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -3,7 +3,7 @@
 BasedOnStyle: LLVM
 ---
 Language: Cpp
-Standard: Cpp17
+Standard: c++17
```
ricab, line 6: [Ouch](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#standard):

> `Cpp11` is a deprecated alias for `Latest`...

.github/actions/lint/action.yml
```diff
@@ -36,6 +36,9 @@ runs:
     run: |
       set -e -o pipefail
 
+      # Ensure that the .clang-format configuration is valid
+      clang-format-16  --dump-config > /dev/null
```
ricab, line 40: Good idea!

ricab (approved):
Indeed @xmkg, good catch!

PR #3957, opened 2025-02-25
[log] introduce log function w/ formatting

xmkg:
the most usual case for a log() call is to produce a log message using `fmt::format()`. the new log() variadic template function streamlines that by providing the formatting by default with minimal intrusion. The old log() API is preserved as-is -- the new code can opt in using the new log() API by passing a format string and format arguments directly:

before
```c++
mpl::log(lvl::debug, "cat", fmt::format("{} {}", 1, 2)); // uses the existing log() overload
```

after

```c++
mpl::log(lvl::debug, "cat", "{} {}", 1 ,2); // uses the new variadic overload
```

The new log() function also uses fmt::format_string<...> to capture the format string argument which means the code will get compile time format string checks for free when the codebase switches to C++20.

MULTI-1844

### DIFF
--- a/include/multipass/logging/log.h
+++ b/include/multipass/logging/log.h
@@ -22,10 +22,43 @@
 #include <multipass/logging/level.h>
 #include <multipass/logging/logger.h>
 
+#include <fmt/std.h>   // standard library formatters
+#include <fmt/xchar.h> // char-type agnostic formatting
+
 namespace multipass
 {
 namespace logging
 {
+
+/**
+ * Log with formatting support
+ *
+ * The old (legacy) log function and this overload are distinguished
+ * via presence of a format argument. When a format argument is absent
+ * both the old function and the new one are valid candidates for a
+ * `log(Level::debug, "test", "test")` call. The ambiguity is resolved
+ * by the fact that a non-templated overload is a better fit.
+ *
+ * @ref https://en.cppreference.com/w/cpp/language/overload_resolution#Best_viable_function
+ *
+ * @tparam Args Type of the format arguments
+ * @param level Log level
+ * @param category Log category
+ * @param fmt Format string
+ * @param args Format arguments
+ */
+template <typename... Args>
+constexpr void log(Level level, const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
+{
+    const auto formatted_log_msg = fmt::format(fmt, std::forward<Args>(args)...);
+    // CString{} is necessary here. Without it, this function would infinitely recurse.
+    // The reason "why" is: the non-templated log() overload requires an implicit
+    // const std::string& -> CString conversion for the category argument, which
+    // makes it a "worse" overload than this templated function. We do the conversion
+    // here in order to prevent that.
+    log(level, CString{category}, formatted_log_msg);
+}
+
 void log(Level level, CString category, CString message);
 void set_logger(std::shared_ptr<Logger> logger);
 Level get_logging_level();
--- a/include/multipass/logging/log.h
+++ b/include/multipass/logging/log.h
@@ -34,29 +34,27 @@ namespace logging
  * Log with formatting support
  *
  * The old (legacy) log function and this overload are distinguished
- * via presence of a format argument. When a format argument is absent
- * both the old function and the new one are valid candidates for a
- * `log(Level::debug, "test", "test")` call. The ambiguity is resolved
- * by the fact that a non-templated overload is a better fit.
+ * via presence of a format argument. This is the reason why the 0th
+ * argument is taken explicitly. The overload resolution rules of C++
+ * makes it complicated to make it work reliably in the codebase, so
+ * the code relies on explicity here.
  *
  * @ref https://en.cppreference.com/w/cpp/language/overload_resolution#Best_viable_function
  *
- * @tparam Args Type of the format arguments
+ * @tparam Arg0 Type of the first format argument
+ * @tparam Args Type of the rest of the format arguments
  * @param level Log level
  * @param category Log category
  * @param fmt Format string
- * @param args Format arguments
+ * @param arg0 The first format argument
+ * @param args Rest of the format arguments
  */
-template <typename... Args>
-constexpr void log(Level level, const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
+template <typename Arg0, typename... Args>
+constexpr void
+log(Level level, const std::string& category, fmt::format_string<Arg0, Args...> fmt, Arg0&& arg0, Args&&... args)
 {
-    const auto formatted_log_msg = fmt::format(fmt, std::forward<Args>(args)...);
-    // CString{} is necessary here. Without it, this function would infinitely recurse.
-    // The reason "why" is: the non-templated log() overload requires an implicit
-    // const std::string& -> CString conversion for the category argument, which
-    // makes it a "worse" overload than this templated function. We do the conversion
-    // here in order to prevent that.
-    log(level, CString{category}, formatted_log_msg);
+    const auto formatted_log_msg = fmt::format(fmt, std::forward<Arg0>(arg0), std::forward<Args>(args)...);
+    log(level, category, formatted_log_msg);
 }
 
 void log(Level level, CString category, CString message);
--- a/tests/test_log.cpp
+++ b/tests/test_log.cpp
@@ -15,13 +15,17 @@
  *
  */
 
+#include "mock_logger.h"
+
 #include <gtest/gtest.h>
 #include <multipass/logging/level.h>
 
 namespace mpl = multipass::logging;
+namespace mpt = multipass::test;
 
 struct log_tests : ::testing::Test
 {
+    mpt::MockLogger::Scope logger_scope = mpt::MockLogger::inject();
 };
 
 TEST_F(log_tests, test_levels_as_string)
@@ -33,4 +37,36 @@ TEST_F(log_tests, test_levels_as_string)
     ASSERT_STREQ(mpl::as_string(mpl::Level::trace).c_str(), "trace");
     ASSERT_STREQ(mpl::as_string(static_cast<mpl::Level>(-1)).c_str(), "unknown");
     ASSERT_STREQ(mpl::as_string(static_cast<mpl::Level>(5)).c_str(), "unknown");
+}
+
+TEST_F(log_tests, test_non_format_overload)
+{
+    logger_scope.mock_logger->expect_log(mpl::Level::error, "no format whatsoever {}");
+    mpl::log(mpl::Level::error, "test_category", "no format whatsoever {}");
+}
+
+TEST_F(log_tests, test_format_overload_single_arg)
+{
+    logger_scope.mock_logger->expect_log(mpl::Level::error, "with formatting 1");
+    mpl::log(mpl::Level::error, "test_category", "with formatting {}", 1);
+}
+
+TEST_F(log_tests, test_format_overload_multiple_args)
+{
+    logger_scope.mock_logger->expect_log(mpl::Level::error, "with formatting 1 test");
+    mpl::log(mpl::Level::error, "test_category", "with formatting {} {}", 1, "test");
+}
+
+TEST_F(log_tests, test_format_overload_multiple_args_superfluous)
+{
+    logger_scope.mock_logger->expect_log(mpl::Level::error, "with formatting 1 test");
+    // Superfluous arguments are ignored. This should be an error with C++20
+    mpl::log(mpl::Level::error, "test_category", "with formatting {} {}", 1, "test", "superfluous");
+}
+
+TEST_F(log_tests, test_format_overload_multiple_args_missing)
+{
+    // Missing arguments are runtime error in C++17. This should be a compile
+    // time error with the C++20
+    EXPECT_THROW({ mpl::log(mpl::Level::error, "test_category", "with formatting {} {}", 1); }, fmt::format_error);
 }


include/multipass/logging/log.h
```diff
@@ -22,10 +22,37 @@
...
+
 namespace multipass
 {
 namespace logging
 {
+
+/**
+ * Log with formatting support
+ *
+ * The old (legacy) log function and this overload is distinguished
```
Sploder12, line 36:
```suggestion
 * The old (legacy) log function and this overload are distinguished
```

include/multipass/logging/log.h
```diff
@@ -22,10 +22,37 @@
...
+ * @tparam Arg0 The type of the first argument
+ * @tparam Args Type of the rest of the arguments
+ * @param level Log level
+ * @param category Log category
+ * @param fmt Format string
+ * @param arg0 Format argument #0
+ * @param args Rest of the format arguments
+ */
+template <typename Arg0, typename... Args>
+inline constexpr void
```
Sploder12, line 49: `inline` is implicit here, but nothing wrong with being explicit!

tests/test_log.cpp
```diff
@@ -14,14 +14,17 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
+#include "fmt/format.h"
+#include "mock_logger.h"
```
Sploder12, line 18: include formatting is a bit inconsistent with existing style. Specifically separating local `""` includes and external `<>` includes with a newline.
```suggestion

#include "fmt/format.h"
#include "mock_logger.h"

```

Sploder12 (commented):
Great work! Only a couple nitpicks from me. I like that the parts that would change in C++20 are marked with comments, that should make transitioning easier in the future!

tests/test_log.cpp
xmkg: Ah, that one's clangd's handiwork. I forgot to pass `-header-insertion=never`, so it just adds `missing` header files as it sees fit to help(!) me.

xmkg:
> Great work! Only a couple nitpicks from me. I like that the parts that would change in C++20 are marked with comments, that should make transitioning easier in the future!

Thanks for the review, @Sploder12! I've made the suggested changes.

xmkg:
Also added @georgeliao and @andrei-toterman to get more opinions.

Sploder12 (dismissed):

andrei-toterman (approved):
It looks great!
I wanted something like this for such a long time, so thank you @xmkg!

Not relevant directly to this PR, but IMO another nice addition would be to have specific logging functions for each logging level, as is popular in many other logging libraries.
For example, instead of `mpl::log(mpl::Level::warning, ...)` it would be nice to have `mpl::warning(...)`. WDYT?

xmkg:
> It looks great! I wanted something like this for such a long time, so thank you @xmkg!
> 
> Not relevant directly to this PR, but IMO another nice addition would be to have specific logging functions for each logging level, as is popular in many other logging libraries. For example, instead of `mpl::log(mpl::Level::warning, ...)` it would be nice to have `mpl::warning(...)`. WDYT?

Yeah, I'd love that. Another thing I want to do is the "lazy evaluation" of the format arguments -- maybe we can merge those two. Assume the following:

```C++
mpl::log(lvl::debug, "cat", fmt::format("{} {}", foo(), bar()));
// ... or mpl::log(lvl::debug, "cat", "{} {}", foo(),baz()); 
```

The foo() and bar() will be invoked and the format string will be produced regardless of whether the log level is set to debug or not. So, what I'd like to do is defer the evaluation of the arguments and produce the formatted string if and only if the log level is set to the log call's level, meaning we need to do sth like this:

```
if(logger_level == log_level){
 // Do the formatting and stuff
}
```

We can avoid producing the format string this way, but if we also want to avoid evaluating the format arguments altogether, we have to rely on macros:

```
#define MPL_LOG(/*varargs*/) \
 if(check log level) \
   fmt::format(/* paste macro args as-is here */)
```

I'd like to avoid macros here but [lazy evaluation is not a thing in core C++ yet](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0927r0.pdf).

Therefore, I'd like to propose introducing log-level macros, like `MPL_WARNING()` and `MPL_TRACE()` to get all the said benefits. What do you think?

andrei-toterman:
I like that. Sounds like a good approach, so I'm +1 on that.

include/multipass/logging/log.h
```diff
@@ -22,10 +22,37 @@
...
 {
 namespace logging
 {
+
+/**
+ * Log with formatting support
+ *
+ * The old (legacy) log function and this overload are distinguished
+ * via presence of a format argument. Hence, the first argument of
+ * the variadic list is taken explicitly.
```
georgeliao, line 38: I think there is a better way than this workaround. The problem here is that the `log(level, category, formatted_log_msg);` matched with the template function because the 2nd parameter type is `const std::string&`. If you want it to match with the non-template `log` function, you can simply convert the `std::string` to `CString`. That eventually leads to the result code
```
template <typename... Args>
constexpr void log(Level level, const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
{
    const auto formatted_log_msg = fmt::format(fmt, std::forward<Args>(args)...);
    log(level, CString{category}, formatted_log_msg);
}

```

georgeliao (commented):
@xmkg Thanks for the contribution. I am happy with this enhancement. Just left one minor comment for improvement.

include/multipass/logging/log.h
xmkg: I believe

> I think there is a better way than this workaround. The problem here is that the `log(level, category, formatted_log_msg);` matched with the template function because the 2nd parameter type is `const std::string&`. If you want it to match with the non-template `log` function, you can simply convert the `std::string` to `CString`. That eventually leads to the result code
> 
> ```
> template <typename... Args>
> constexpr void log(Level level, const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
> {
>     const auto formatted_log_msg = fmt::format(fmt, std::forward<Args>(args)...);
>     log(level, CString{category}, formatted_log_msg);
> }
> ```
Good catch, I believe you're right! The correct (old) function is called even without the explicit Arg0, but that's not because of the `const std::string&`. The compiler picks the right candidate [even when the category arg of the template has the type CString](https://gcc.godbolt.org/z/To3TeKr31). Both functions are good candidates for a call `log(Level::debug, "test", "test")`, but according to the C++ overload resolution order a non-templated function is a [better match than a templated one](https://en.cppreference.com/w/cpp/language/overload_resolution#Best_viable_function), so the non-templated overload is picked and the call is not ambiguous, which makes the explicit Arg0 redundant.

I'll make the change.

xmkg: Done :+1:

georgeliao: If it were `CString` on the template function, I think it comes down to the 3rd argument to parameter conversion, right? More specifically, converting `formatted_log_msg` to `CString` or `fmt::format_string`, and the compiler decides that converting to `CString` is easier. Template function vs non-template function comes into play later if this is equal.

xmkg: > If it were `CString` on the template function, I think it comes down to the 3rd argument to parameter conversion, right? More specifically, converting `formatted_log_msg` to `CString` or `fmt::format_string`, and the compiler decides that converting to `CString` is easier. Template function vs non-template function comes into play later if this is equal.

I think the `fmt::format_string` and CString have equal weight because both have to perform a conversion from `const char*`. fmt::format_string<> is a materialized type at the point of evaluation. If the argument type being a template would weigh the overload down, I'd expect std::string to have the same effect because std::string is just an alias to `basic_string<char>`, which is a templated type.

FWIW, the following results in ambiguity:

```
void log(Level level, CString category, fmt::format_string<void> args){}
void log(Level level, CString category, CString message) {}

int main(void) { 
    log(Level::debug, "test", "test");
}
```

https://gcc.godbolt.org/z/55o71Yc95

georgeliao: Interesting, I have changed the call to 
```
    const std::string str = "test_str";
    log(Level::debug, CString{"test"}, str);
```
That still results in ambiguity. Therefore your previous claim is right. It did fall back to rank 4 comparison which is template vs non-template function.

georgeliao (dismissed):

xmkg:
Reverted to the initial implementation due to the concerns I've raised [here](https://github.com/canonical/multipass/pull/3957#discussion_r1975330507). @andrei-toterman @Sploder12 This is the initial implementation you've approved, so there are no changes at this time. @georgeliao, feel free to check if I'm missing anything. I think the explicit argument route is the safest way to go because overload resolution rules are too broad to be all accounted for.

Sploder12 (approved):

ricab (approved):
This is a *quaternary* review only, so I only glanced at the code, but thank you @xmkg for a very welcome change!

georgeliao (approved):

xmkg:
> Re: macros and lazy evaluation: as discussed elsewhere, that one I am not very keen on (although it is a clever idea) because, IMO:
> 
> 1. The only real benefit would be performance, which would only be significant in hot paths that were not already dominated much slower operations, like launching VMs, ssh'ing somewhere, resolving IPs, etc.; there shouldn't be many of those in Multipass.
> 2. Where we find such cases, I am totally for optimizing the logs. However, it would then be best to avoid slowing things down regardless of the level. Furthermore, we retain the ability to add a local `if` on the level anyway. The macro shortcut for such rare cases wouldn't be very useful.
> 3. It could counterintuitive — when people call something, they expect it to be evaluated.
> 4. Most importantly, I think we should strive for the logging level to influence program behavior as little as possible. We don't want to increase the log level to debug and find that the issue fails to reproduce because it doesn't manifest.


> The only real benefit would be performance
> It could be counterintuitive — when people call something, they expect it to be evaluated.

Well, for the logging I'd expect the opposite. I'm mainly pursuing this idea because of the program correctness -- performance is a nice side benefit. If I'm flagging a log line as "trace", I wouldn't expect it to have any side effects when the log level is set to "info" (i.e., a no-op), but that's not the case at the moment. Assume that there's an expensive function that produces a whopping 1G file and returns its path, which is only meant to be used for runtime debugging purposes. Right now, the file would be produced regardless of the log level if that function is called as a part of a logging call -- which is not an unreasonable thing to do. IMHO, the logging calls should not have any side effects at all. They're simply an observation point, nothing more.

> Most importantly, I think we should strive for the logging level to influence program behavior as little as possible. 

Exactly! Logging should not do anything else than logging. Changing the logging level should not change how the program behaves because the program shouldn't rely on logging calls to do its job. I think if there are instances of this, we should find and fix them. It shouldn't be hard to detect such log calls since it must be an invocation, either by operator() or a function such as apply, invoke, etc. I'm optimistic that there aren't many instances of this, but it might be a reason to shy away from the idea if it's a common practice.

I'll try to find such calls as extensively as possible and write a feasibility report. Then, we can discuss this while going through the potential impact and drawbacks.

ricab:
> Well, for the logging I'd expect the opposite.

Why? It is C++, if you do `f(foo(), bar())`, you expect `foo`, `bar`, and `f` to be called. Are you saying you would expect otherwise if `f` happens to be a log call?

> Exactly! Logging should not do anything else than logging. Changing the logging level should not change how the program behaves because the program shouldn't rely on logging calls to do its job.

It's not the logging calls, it is the calls you make before them. The side effects of `foo` or `bar` in the example above are not side effects of the logging call. Your proposal would simulate them being part of the log call though.

> I think if there are instances of this, we should find and fix them.

How do you know in advance what is relevant? Just differences in timing may be enough to change outcomes (we should try to *minimize* effects on program behavior, but we should know that we can't *eliminate* them completely). If there is a bug in a log call, it is because it wasn't identified in advance. The inability to reproduce it with different log levels can only hinder its revelation.

Suppose we implemented your proposal. If you found this code: 

```c++
LOG("{}{}{}", deadlocking_fun(), slow_fun(), some_map["absent"].getter());
```

How would we minimize the impact on program behavior? Well, we could, fix `deadlocking_fun()` not to deadlock, fix `slow_fun()` to be fast, and make sure the element was in the map, but the point is that you don't always know about these problems a priori. Hence bugs exist!

Unless you are proposing that we be very strict as a matter of practice and ask that all args be extracted to vars first, so it became

```c++
auto a = deadlocking_fun();
auto b = slow_fun();
auto c = some_map["absent"].getter();
LOG("{}{}{}", a, b, c);
```

But this achieves exactly what `mpl::log` does today! It evaluates things before the log call, just like everywhere else in C++, with the downside that it now takes 4 lines to log, diluting important code and polluting the scope with useless vars (unless you add 2 more lines for curly braces (or adopt questionable formatting to have them inline)). 

By replacing `mpl::log` with `LOG` you threw away an automatic guarantee that all that stuff was evaluated first. You could reinstate it by other means (like some sanitizer or something) but that would just be going in circles.

I think you're confused on this one @xmkg, but maybe I am. I would appreciate if others could share their input.

xmkg:
> Why? It is C++, if you do `f(foo(), bar())`, you expect `foo`, `bar`, and `f` to be called. Are you saying you would expect otherwise if `f` happens to be a log call?

We're not doing a `f(foo(), bar())`, are we? It's `f(level::debug, foo(), bar())`, and in an ideal world, I'd expect foo and bar not to be called at all. C++ core language lacks the mechanism to do so, therefore, the idea is to introduce a macro function to do that. Would you assume that a `MPL_WARNING(...)` call would or would not evaluate its arguments by default?  If you ask me, I'd not assume that because it's a *macro function*. That's the distinction. If you're calling the mpl::log(), sure, you'd get what you expect. Arguments are evaluated, and everything is fine and dandy. But if you're using the macro function -- and I'd like to underline that it's an *opt-in choice* -- you'll get a different behavior because you're calling a different thing. You might argue that we should make it obvious, and we can do that by giving a clear indication in the name, like `MPL_LAZY_LOG()`.

I stand by the point that if we're passing foo() and bar() directly to a log call, the only purpose of those calls must be to produce an argument for the format string. If foo() and bar() are functions with side effects and part of the core logic of the owning function, they should be explicitly made part of the owning function. The alternative is a code smell in my book.

> Suppose we implemented your proposal. If you found this code:
> 
> ```c++
> LOG("{}{}{}", deadlocking_fun(), slow_fun(), some_map["absent"].getter());
> ```
> 
> How would we minimize the impact on program behavior? Well, we could, fix `deadlocking_fun()` not to deadlock, fix `slow_fun()` to be fast, and make sure the element was in the map, but the point is that you don't always know about these problems a priori. Hence bugs exist!

> Unless you are proposing that we be very strict as a matter of practice and ask that all args be extracted to vars first, so it became
> 
> ```c++
> auto a = deadlocking_fun();
> auto b = slow_fun();
> auto c = some_map["absent"].getter();
> LOG("{}{}{}", a, b, c);
> ```

Yep, I'm exactly proposing this for all functions that are not just a basic getter or some kind of string massager. How often do you see code calling such functions in a log call? I'd say it's not a common practice. That's my opinion anyway, but that's not what I'm proposing for us to do. See below.

> By replacing `mpl::log` with `LOG` you threw away an automatic guarantee that all that stuff was evaluated first.

Again, I'm not suggesting that we should `replace` the mpl::log with macro calls. I'm just proposing that we can have macro log calls with said benefits so we can use it in new code or where it matters as an opt-in fashion. 

Let me know if I can clarify anything else for you.

Sploder12:
> Why? It is C++, if you do `f(foo(), bar())`, you expect `foo`, `bar`, and `f` to be called. Are you saying you would expect otherwise if `f` happens to be a log call?

Personally I agree with @ricab on this point. If I saw `f(foo(), bar())` I'd be very confused if `foo` or `bar` didn't happen. Although, if I saw `f(foo, bar)` (where `foo` and `bar` are functions), I wouldn't be confused since my expectations depend on `f` instead of the language.

ricab:
You won't let me go on weekened :wink: jk

(This whole bike shedding is way off-topic for the current PR, which I am again thankful for and glad to see merged! But better discuss and decide before more efforts and this just happens to be the forum where it was brought up.)

I also entirely agree that

> if we're passing foo() and bar() directly to a log call, the only purpose of those calls must be to produce an argument for the format string.

But the point is that `foo()` and `bar()` may have relevant side effects regardless of that *only purpose*. Those side effects can influence your ability to reproduce a bug.

In other words, and regardless of programmers' intentions, for any call of the form

```c++
xyz(lvl, "{}", apparently_simple_getter());
```

... one of the following must be the case:

1. `apparently_simple_getter()` has no relevant side effects that ever create or influence a bug
2. `apparently_simple_getter()` actually has relevant side effects that eventually create or influence a bug

We cannot postulate 2 never to be the case. Again, regardless of programmer intentions. You can recommend always extracting a variable (which would have the downsides I mentioned) to preclude code of that form entirely, but you cannot guarantee that happens.

Now, if you replace `xyz` with our `mpl::log`, `apparently_simple_getter` gets called regardless of the log level. So changing the log level, does not affect your ability to investigate bugs in either case 1 or in case 2. 

If you replace `xyz` with `MPL_LAZY_LOG`, however, you cannot say the same. You are safe in case 1, but not in case 2. For that reason, I am wary of providing the option and I don't see any benefit for correctness. As mentioned, I recognize a benefit for performance, of course, but I think that is less relevant.

ricab:
At the end of the day, I do not think it is that important. I have layed out my arguments, so I'll defer to whatwver the majority decides.

PR #3958, opened 2025-02-26
[ci/BuildAndTest]: upload test coredump to artifacts on crash

xmkg:
The BuıldAndTest(Debug) step crashes intermittently without any significant clue for us to pursue. The crash does not happen in the local environment so the best next thing is to grab the core dump & crashed executable and upload them to artifacts so we can investigate them.

MULTI-1845

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -166,16 +166,69 @@ jobs:
     - name: Clear CCache stats
       run: ccache --show-stats --zero-stats
 
+    - name: Set /proc/sys/kernel/core_pattern
+      if: ${{ matrix.build-type == 'Debug' }} || ${{ matrix.build-type == 'Coverage' }}
+      run : |
+
+        # The LXC container share the same "/proc/sys/kernel/core_pattern"
+        # with the host and it can't be overridden inside the container. Hence
+        # we need to override it at the runner level.
+        sudo bash -c 'echo "/coredump/%e.%p.%t" > /proc/sys/kernel/core_pattern'
+
+    # - name : Mock failure
+    #   id: mock-fail
+    #   if: ${{ matrix.build-type == 'Debug' }} || ${{ matrix.build-type == 'Coverage' }}
+    #   run: |
+
+    #     # Perform an intentional segfault
+    #     set +e
+    #     set -o xtrace
+    #     # mkdir -p /root/parts/multipass/build/bin/
+    #     instance_name=`/snap/bin/lxc --project snapcraft --format=csv --columns=n list | grep multipass`
+    #     /snap/bin/lxc --project snapcraft start $instance_name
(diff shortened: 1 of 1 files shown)


xmkg:
I've tested the crash dump capture and upload steps with [a mock step](https://github.com/canonical/multipass/actions/runs/13550769958/job/37873599126) (which has now been removed), and it seems to work properly. An example set of artifacts uploaded from the mock crash [can be found here](https://github.com/canonical/multipass/actions/runs/13550769958/artifacts/2658124881)

I've tried to retry the pipeline an unhealthy amount of times but could not trigger the real crash, but I think the code will work for the real thing as the mock is pretty close to the real thing (observable behavior-wise).

xmkg:
We discussed this out-of-band with @ricab, and he said that he's OK with the concept, so this PR needs some eyes to confirm that I haven't missed anything obvious.

.github/workflows/linux.yml
```diff
@@ -166,16 +166,34 @@ jobs:
     - name: Clear CCache stats
       run: ccache --show-stats --zero-stats
 
+    - name: Set /proc/sys/kernel/core_pattern
+      if: ${{ matrix.build-type == 'Debug' }} || ${{ matrix.build-type == 'Coverage' }}
```
Sploder12, line 170: It seems this is running in [Release](https://github.com/canonical/multipass/actions/runs/13585828619/job/37980523777?pr=3958) and [Clang](https://github.com/canonical/multipass/actions/runs/13585828619/job/37980523465?pr=3958), I think it might have to do with having two separate expressions. 
```suggestion
      if: ${{ matrix.build-type == 'Debug' || matrix.build-type == 'Coverage' }}
```

Sploder12 (commented):
Great work! Looks like there is a minor issue where an `if` isn't being evaluated as it should, but nothing major! 

Somewhat unrelated but for separate future task/PR(s): It'd be cool to have something similar for Windows and MacOS as well. I know Apple Clang tends to crash so retrieving that would be nice to have too.

.github/workflows/linux.yml
xmkg: Ah, good catch. It's still weird that it causes the pipeline to run for those, though.

xmkg: Made the change; let's see if it works as intended.

xmkg:
> Somewhat unrelated but for separate future task/PR(s): It'd be cool to have something similar for Windows and MacOS as well. I know Apple Clang tends to crash so retrieving that would be nice to have too.

That's a good idea. Let's use this patch as a pilot program, land it, and see how well it works. Then, we can convert it to a dedicated action that can be reused in different workflows. Maybe we can even register a coredump handler just like apport to have minimal intrusion on the jobs and still be able to capture the stuff we need to troubleshoot.

.github/workflows/linux.yml
xmkg: Yep, that was it. The step is only running for debug & coverage now.

Sploder12 (approved):

PR #3966, opened 2025-03-05
[log] introduce mpl::<level-name>(...) log functions

xmkg:
Added the following templated log functions: error, warn, info, debug, trace.

Added a templated log overload, which takes the log level as a template parameter. This function acts as a dispatcher for calls with/without a format.

Moved non-templated function declarations to the top.

Unit tests were added for the new log-level functions.

Prefixed all log() calls with logging:: namespace so log() from math library does not get pulled into overload resolution candidates.

MULTI-1853

### DIFF
--- a/include/multipass/logging/log.h
+++ b/include/multipass/logging/log.h
@@ -30,6 +30,11 @@ namespace multipass
 namespace logging
 {
 
+void log(Level level, CString category, CString message);
+void set_logger(std::shared_ptr<Logger> logger);
+Level get_logging_level();
+Logger* get_logger(); // for tests, don't rely on it lasting
+
 /**
  * Log with formatting support
  *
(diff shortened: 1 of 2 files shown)


tests/test_log.cpp
```diff
@@ -70,3 +70,67 @@ TEST_F(log_tests, test_format_overload_multiple_args_missing)
...
+    logger_scope.mock_logger->expect_log(mpl::Level::trace, "with formatting 1");
+    mpl::trace("test_category", "with formatting {}", 1);
+}
+
+// ------------------------------------------------------------------------------
+
+TEST_F(log_tests, test_log_error_function_noargs)
+{
+    logger_scope.mock_logger->expect_log(mpl::Level::error, "without formatting {}");
+    mpl::error("test_category", "without formatting {}");
```
andrei-toterman, line 111: Hmm, should this be the expected output? So something like `mpl::error("category", "{} {}")` will log `"{} {}"` just fine, but `mpl::error("category", "{} {}", 1)` will give an error? Why not always treat the string as a format string?

xmkg: Because the user's intention might legitimately include printing "{}" without formatting (e.g., printing a JSON document as-is) [which would cause fmt::format to throw an error](https://gcc.godbolt.org/z/8G53enW4E). There are several instances of this in the codebase, and we need to be backward-compatible. 

If we're not formatting, we shouldn't experience the side effects of formatting either.

andrei-toterman: ah, i see. for just printing {} with fmt::format, you could always use {{}}, but if this would break our existing code, it does make sense to have it like that. thanks for the explanation!

xmkg: I mean, these are new functions, and we can say that "these functions will always format, and you have to escape your braces", but that'd be inconsistent with the existing non-templated log overload.

```
mpl::error(category, "{}"); // runtime error, missing format argument
mpl::error(category, "{{}}"); // works fine, prints {}
mpl::log(Level::error, category, "{}"); // works fine, prints {}
```

andrei-toterman: i feel like that would be more confusing. those functions should just introduce the log level, not have any extra special behavior IMO

Sploder12 (approved):
LGTM! btw, what tool do you use for function comments? The comments are very nice.

xmkg:
> LGTM! btw, what tool do you use for function comments? The comments are very nice.

Thanks for the review, @Sploder12! I use the [doxdocgen](https://marketplace.visualstudio.com/items?itemName=cschlosser.doxdocgen) vscode extension; it's convenient.

include/multipass/logging/log.h
```diff
@@ -43,24 +48,116 @@ namespace logging
...
+template <Level level, typename... Args>
+constexpr void log(const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
+{
+    if constexpr (sizeof...(Args) > 0)
+    {
+        // Needs formatting
+        logging::log(level, category, fmt, std::forward<Args>(args)...);
+        return;
+    }
+    logging::log(level, category, CString{fmt.get().data()});
```
xmkg, line 88: I'm having second thoughts about this part. `.get()` returns a `fmt::string_view`, and `.data()` returns a `const char*` to a buffer, which is not guaranteed to be NUL-terminated. So, to ensure that the buffer is NUL-terminated, we have to construct a `std::string`. Alternatively, we can add a new log API function that can work with string views and handle them properly. I'm more inclined towards the second option.

xmkg: Maybe it's time to get rid of the `CString`.

Sploder12: I vote to get rid of the `CString`

xmkg: > I vote to get rid of the `CString`

:genie:   https://github.com/canonical/multipass/pull/3969

georgeliao:
Hi @xmkg 
Thanks for the good work. 
Just have one doubt, I would be happen to use 
```
template <Level level, typename... Args>
constexpr void log(const std::string& category, fmt::format_string<Args...> fmt, Args&&... args)
``` 
function where I specify the level via template argument. I am not sure we need the different function like `error`, `warn` wrap on that.

xmkg:
Let's hold this MP until https://github.com/canonical/multipass/pull/3969 gets merged.

xmkg:
> Let's hold this MP until #3969 gets merged.

The pending change I want to make could be done in #3969, so I'm merging this.

Issue #3965, opened 2025-03-03
[Windows] `multipass stop` hangs daemon when instance has mounts


Sploder12:
**Describe the bug**
On Windows, stopping an instance with a mount will cause the daemon to become unresponsive. No further commands will be processed. Restarting the daemon fixes the issue.

**To Reproduce**
1. Launch an instance with a mount
2. Stop the instance
3. Observe issue

**Expected behavior**
The daemon does not hang

**Logs**
Information: [valid-mandrill] SSH session disconnected
Error: [sftp server] sshfs in the instance appears to have exited unexpectedly. Trying to recover.
Error: [sshfs mount] Cause an unhandled exception: unable to create a channel for remote process: 'findmnt --source:C:\tmp -o TARGET -n', the SSH session is not connected.
Warning: [sshfs-mount-handler] Mount "/home/ubuntu/tmp/" in instance 'valid-mandrill' has stopped unsuccessfully: Process returned exit code: 1
Information: [valid-mandrill] SSH session disconnected: [ssh proc] failed to open session channel: 'Socket error: disconnected'

**Additional info**
- OS: Windows 10 Home
- CPU architecture or model: x86_64
- `multipass version`: 1.16.0-dev.3098+g63fffb483.win
- `multipass info`: never returns (or times out after a long time?)
- `multipass get local.driver`:  VirtualBox

ricab:
Ouch, does this happen with Hyper-V too, do you know? Can you reproduce the same thing with 1.15.1?

Sploder12:
> Ouch, does this happen with Hyper-V too, do you know? Can you reproduce the same thing with 1.15.1?

Not sure of Hyper-V. I can't reproduce on 1.15.1

PR #3971, opened 2025-03-07
[ssh/ssh_session] different timeouts for connect & established

xmkg:
Sometimes, ssh_connect blocks while a VM is booting up, where it won't return to the caller at all. Since the timeout is set to ~infinity the ssh_connect would block forever.

This patch fixes that by using two different timeout values for the different session connection states.

MULTI-1863

### DIFF
--- a/src/ssh/ssh_session.cpp
+++ b/src/ssh/ssh_session.cpp
@@ -48,7 +48,7 @@ mp::SSHSession::SSHSession(const std::string& host,
      * in order to prevent that from happening. The established timeout
      * is used afterward to prevent timing out in connected state.
      */
-    constexpr long kConnectTimeoutSecs = 10; // < how long to wait for ssh_connect
+    constexpr long kConnectTimeoutSecs = 5; // < how long to wait for ssh_connect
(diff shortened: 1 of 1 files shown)


Sploder12 (dismissed):
Looks good! sshfs mounts didn't give me any issues and other ssh features worked as expected.

xmkg:
Reduced the timeout from 10 to 5 as discussed. @ricab @andrei-toterman @georgeliao

Sploder12:
Interestingly, this indirectly fixes #3965

xmkg:
> Interestingly, this indirectly fixes #3965

Good to hear! Two birds with one stone 😉

georgeliao (approved):

xmkg:
> If we try to connect and wait infinitely, it should connect once the vm is up and sshd service is up instead of hanging there

That's the question. I think some quirks are going on regarding how libssh handles the situation. Maybe the server is disconnecting the connection (because it's too early for a client to connect), but libssh somehow not picking it up. The timeout is designed to handle situations like this, but I still believe there's a bug in ssh_connect that leads to this.

xmkg:
Thanks for the reviews @georgeliao, @Sploder12!

PR #3972, opened 2025-03-10
[test_sshfsmount] fix invalid optional access in helper function

xmkg:
make_exec_to_check_commands is not checking whether the optional object contains an optional or not. the patch fixes that.


Sploder12 (approved):
Approving on basis of I've done this same change several times before.

Although I approve of the fix, I think `make_exec_to_check_commands`'s `fail_invoked` parameter needs some attention since its usage and `std::optional<bool>&` is _interesting._

xmkg:
> `std::optional<bool>&` is _interesting._

Yep, indeed it is.

PR #3974, opened 2025-03-11
[ci/BuildAndTest] pull the coredump folder as a whole

xmkg:
one-by-one file pulling was breaking when the file included whitespace characters. this patch switches to pulling the whole coredumps folder to simplify the code. 

also added a fix for pull coredump step not triggering when the segfault happens in the coverage step.

MULTI-1875


xmkg:
Finally caught the crash, which also confirms that the changes made in this PR works: https://github.com/canonical/multipass/actions/runs/13810864158/job/38632040351

Sploder12 (approved):
LGTM! This will be very useful in reducing spurious CI failures!

PR #3979, opened 2025-03-12
az gui

levkropp:
Implement availability zones support in GUI

- Add ZONE column to VM instances table

- Add zone controls in bulk actions and filters

- Add zone selection to instance configuration

- Add Launch & Configure Next functionality


src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -68,6 +134,25 @@ class BulkActionsBar extends ConsumerWidget {
...
+                  ...actionButtons.gap(width: 8).toList(),
+                  const SizedBox(width: 8),
+                  const ZonesDropdown(),
+                ],
+              ),
+            ),
+          ),
+        ],
+      ),
+    );
```
andrei-toterman, line 156: Why was this necessary? It's a row that has a single expanded element inside it. And the items fit entirely even in a minimum size window, so the scroll view shouldn't be needed either?

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,72 @@ import '../providers.dart';
...
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final available = ref.watch(zoneStatusProvider(zoneName));
+    final client = ref.watch(grpcClientProvider);
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        Text('Zone $zoneName'),
```
andrei-toterman, line 68:
```suggestion
        Text(zoneName),
```

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,72 @@ import '../providers.dart';
...
+              ),
+            ),
+          ),
+        ];
+      },
+    );
+  }
+}
+
+class _ZoneToggleRow extends ConsumerWidget {
```
andrei-toterman, line 55: This is missing some parts of the design, namely the borders and the number of running VMs per zone.

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,72 @@ import '../providers.dart';
...
+class ZonesDropdown extends ConsumerWidget {
+  const ZonesDropdown({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    // Pre-fetch zones data
+    final zonesAsync = ref.watch(zonesProvider);
+    return PopupMenuButton(
+      tooltip: 'Change zone availability',
+      offset: const Offset(0, 40),
```
andrei-toterman, line 21:
```suggestion
      position: PopupMenuPosition.under,
```

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,72 @@ import '../providers.dart';
...
+      child: OutlinedButton(
+        onPressed: null,
+        child: const Text('Zones', style: TextStyle(color: Colors.black)),
+      ),
+      itemBuilder: (context) {
+        return [
+          PopupMenuItem(
```
andrei-toterman, line 28: This returns a list which contains a single item, and that item contains a column that contains all the zones-specific widgets. Why aren't they directly integrated into the returned list? Take a look at `HeaderSelection`, where something similar is done.

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +83,77 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+              ),
+              error: (_, __) => const Text('Error loading zones'),
+              data: (zones) {
+                final availableZones = zones.where((z) => z.available).toList();
+                final hasAvailableZones = availableZones.isNotEmpty;
+
+                return Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    DropdownButton<String>(
```
andrei-toterman, line 122: You could use our own `Dropdown` widget, which has made to have the right styling.

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -285,6 +362,13 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
             Row(children: [
               launchButton,
               const SizedBox(width: 16),
+              TextButton(
+                onPressed: hasAvailableZones
+                    ? () => launch(imageInfo, configureNext: true)
+                    : null,
+                child: const Text('Launch & Configure Next'),
+              ),
```
andrei-toterman, line 370: Could you extract this into a variable, just like `launchButton`?

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -317,7 +401,20 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+
+    if (configureNext) {
+      // Reset the form
+      formState.reset();
+      mountFormState?.reset();
+      setState(() {
+        mountRequests.clear();
+        addingMount = false;
+      });
+    }
```
andrei-toterman, line 413: Hmm, wasn't the whole point of 'Configure next' to keep your form intact so you can just modify what you want?

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +83,77 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+                      items: hasAvailableZones
+                          ? [
+                              const DropdownMenuItem(
+                                value: 'auto',
+                                child: Text('Auto (assign automatically)'),
+                              ),
+                              for (final zone in availableZones)
+                                DropdownMenuItem(
+                                  value: zone.name,
+                                  child: Text('Zone ${zone.name}'),
```
andrei-toterman, line 135:
```suggestion
                                  child: Text(zone.name),
```

andrei-toterman (changes requested):

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,126 @@ import '../providers.dart';
...
+                }).toList(),
+              ]),
+    );
+  }
+}
+
+class _ZoneToggleRow extends ConsumerWidget {
+  final String zoneName;
+  final bool available;
+  final int instanceCount;
```
andrei-toterman, line 94: I was wrong about my earlier advice of passing those in the constructor. You were right to watch providers here. Since those values are passed in, they won't update when the provider changes, since they are in a popup. So you need to go back to watching the providers directly in this widget, and have this widget receive just the zone name in the constructor. You'll need to add some code like this
```dart
final available = ref.watch(zonesProvider.select((zones) {
  return zones.firstWhere((z) => z.name == zoneName).available;
}));
final runningInstances = ref.watch(vmInfosProvider.select((infos) {
  return infos
      .where((vm) => vm.zone.name == zoneName)
      .where((vm) => vm.instanceStatus.status == Status.RUNNING)
      .length;
}));
```

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -9,6 +9,126 @@ import '../providers.dart';
...
+                      'Availability zones',
+                      style: TextStyle(
+                        fontSize: 24,
+                        fontWeight: FontWeight.w600,
+                        color: Colors.black,
+                      ),
+                    ),
+                  ),
+                ),
+                ...zones.map((zone) {
```
andrei-toterman, line 59: After you move the retrieval of `runningCount` to the `_ZoenToggleRow` widget, you could go to a `for (final zone in zones)` instead of `...zones.map`, which will look nicer.

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +84,69 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       width: 360,
     );
 
+    final zonesAsync = ref.watch(zonesProvider);
```
andrei-toterman, line 87: Could you move watching the provider to the top, where all the other watches are happening? Also, there you could add a `final availableZones = zones.where((z) => z.available).toList();`, since you need that in multiple places.

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +84,69 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Dropdown<String>(
+                      value: field.value,
+                      onChanged: field.didChange,
+                      items: {
+                        'auto': 'Auto (assign automatically)',
+                        for (final zone in availableZones) ...<String, String>{
+                          zone.name: 'Zone ${zone.name}',
+                        },
```
andrei-toterman, line 130:
```suggestion
                        for (final zone in availableZones)
                          zone.name: 'Zone ${zone.name}',
```

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +84,69 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+                      },
+                    ),
+                    if (!hasAvailableZones)
+                      const Padding(
+                        padding: EdgeInsets.only(top: 8),
+                        child: Text(
+                          'Cannot launch new instances - all zones are unavailable',
+                          style: TextStyle(color: Colors.red),
+                        ),
+                      ),
```
andrei-toterman, line 140: There is a much cleaner way to achieve this. Our `Dropdown` contains an `InputDecoration` which has an `errorText` property. You could add `final String? errorText` to our `Dropdown` and forward it to the `InputDecoration`. Then, you don't need to do all of this layout manually, you can just do something like
```dart
Dropdown<String>(
  ...
  errorText: availableZones.isEmpty ? 'All zones are unavailable' : null,
  ...
)
```
This will also allow you to get rid of the parent `Column`

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +84,69 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+                final hasAvailableZones = availableZones.isNotEmpty;
+
+                return Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Dropdown<String>(
+                      value: field.value,
+                      onChanged: field.didChange,
+                      items: {
+                        'auto': 'Auto (assign automatically)',
```
andrei-toterman, line 127: Perhaps it would be better to not display the auto option if there are no zones available
```suggestion
                         if (availableZones.isNotEmpty)
                           'auto': 'Auto (assign automatically)',
```

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -83,6 +84,69 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       width: 360,
     );
 
+    final zonesAsync = ref.watch(zonesProvider);
+    final zoneDropdown = FormField<String>(
+      initialValue: 'auto',
+      onSaved: (value) =>
+          launchRequest.zone = value == 'auto' ? '' : value ?? '',
+      builder: (field) {
+        return Column(
```
andrei-toterman, line 93: Perhaps all of this layout code could be encapsulated into a separate widget so that it doesn't pollute the rest of the launch form code. Ideally it should be just a FormField that wraps a single widget that knows how to render the whole dropdown

andrei-toterman (changes requested):

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -68,6 +188,15 @@ class BulkActionsBar extends ConsumerWidget {
...
     ];
 
-    return Row(children: actionButtons.gap(width: 8).toList());
+    return SizedBox(
+      height: 36,
+      child: Row(
+        children: [
+          ...actionButtons.gap(width: 8).toList(),
+          const SizedBox(width: 8),
+          const ZonesDropdown(),
```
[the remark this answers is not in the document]
levkropp: I think actionButtons is fine since the Zones button contains actions

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -68,6 +188,15 @@ class BulkActionsBar extends ConsumerWidget {
         ),
     ];
 
-    return Row(children: actionButtons.gap(width: 8).toList());
+    return SizedBox(
+      height: 36,
+      child: Row(
+        children: [
+          ...actionButtons.gap(width: 8).toList(),
```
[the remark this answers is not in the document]
levkropp: This change does not work, it only works with .toList()
```
[        ] [        ] lib/vm_table/bulk_actions.dart:193:31: Error: The argument type 'Iterable<Widget>' can't be assigned to the parameter type 'List<Widget>'.
```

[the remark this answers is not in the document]
andrei-toterman: Yeah, I made this comment before the comment about moving `ZonesDropdown()` to `actionButtons`. So if you made that move, then indeed `Row` needs to take a list. This comment was valid only if you kept the old code that had the spread operator inside another list.

src/client/gui/lib/providers.dart
```diff
@@ -138,6 +145,12 @@ final deletedVmsProvider = Provider((ref) {
       .toBuiltSet();
 });
 
+final zonesProvider = Provider<AsyncValue<BuiltList<Zone>>>((ref) {
+  return ref
+      .watch(pollingProvider)
+      .whenData((data) => data.zones.build() as BuiltList<Zone>);
+});
```
andrei-toterman, line 152: Again, as I mentioned in my previous comment, I think we should drop the `AsyncValue` stuff. Please refer to my original comment on this piece of code https://github.com/canonical/multipass/pull/3979#discussion_r2036062468. Of course, feel free to discuss this and provide counter-arguments :)
```suggestion
final zonesProvider = Provider<BuiltList<Zone>>((ref) {
  return ref.watch(pollingProvider).valueOrNull?.zones.build() ?? BuiltList();
});
```

src/client/gui/lib/vm_table/bulk_actions.dart
```diff
@@ -66,8 +67,11 @@ class BulkActionsBar extends ConsumerWidget {
           currentStatuses: statuses,
           function: () => function(action),
         ),
+      const ZonesDropdownButton(),
     ];
 
-    return Row(children: actionButtons.gap(width: 8).toList());
+    return Row(
+      children: actionButtons.gap(width: 8).toList(),
+    );
```
andrei-toterman, line 75:
```suggestion
    return Row(children: actionButtons.gap(width: 8).toList());
```

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -245,11 +261,22 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
+      onPressed: hasAvailableZones ? () => launch(imageInfo) : null,
       child: const Text('Launch'),
     );
 
+    final launchAndConfigureNextButton = TextButton(
+      onPressed: hasAvailableZones
+          ? () => launch(imageInfo, configureNext: true)
+          : null,
+      child: const Text('Launch & Configure Next'),
+    );
```
andrei-toterman, line 278: Since those 'Launch and configure next' changes are now part of another PR, I think they should be removed from this one.

src/client/gui/lib/vm_table/zones_dropdown_button.dart
```diff
@@ -0,0 +1,124 @@
...
+                fontSize: 12,
+                color: Colors.black,
+              ),
+            ),
+          ],
+        ),
+        Switch(
+          value: available,
+          onChanged: (value) => client.zonesState([zoneName], value),
+        ),
```
andrei-toterman, line 120: Isn't this still the Flutter Switch instead of our own Switch from `switch.dart`?

andrei-toterman (changes requested):
Hey, @levkropp. I see that many of my requests from the previous review round were not addressed. Perhaps they ended up in the collapsed part of the conversation? Please go again slowly and carefully through them. And of course, if you think something is not right, bring it up in the conversation!

src/client/gui/lib/catalogue/launch_form.dart
levkropp: This PR has updates to the logic of the launchAndConfigureNextButton however (hasAvailableZones check)

andrei-toterman: Ah, I see. But this will most likely lead to a merge conflict once its time for this PR to get into `main`. I guess that issue can be addressed later.

levkropp:
Hey @andrei-toterman , since there have been large reworks to the files, I'd appreciate it if you went through the new files and let me know what still needs to be done since (like we saw with the toList issue) the changes may be outdated and updates may be incorrect. I've made all the zoneAsync changes and the other changes you have mentioned in your latest round of review. Please let me know as of this latest force-push what still needs to be changed/implemented!

src/client/gui/lib/catalogue/zone_dropdown.dart
```diff
@@ -0,0 +1,55 @@
...
+          return Dropdown<String>(
+            value: value,
+            onChanged: onChanged,
+            items: {
+              'auto': 'Auto (assign automatically)',
+              for (final zone in availableZones) zone.name: 'Zone ${zone.name}',
+            },
+            errorText: hasAvailableZones
+                ? null
+                : 'Cannot launch new instances - all zones are unavailable',
```
andrei-toterman, line 49:
```suggestion
                : 'All zones are unavailable',
```

src/client/gui/lib/catalogue/zone_dropdown.dart
```diff
@@ -0,0 +1,55 @@
...
+        (() {
+          final zones = ref.watch(zonesProvider);
+          final availableZones = zones.where((z) => z.available).toList();
+          final hasAvailableZones = availableZones.isNotEmpty;
+
+          return Dropdown<String>(
+            value: value,
+            onChanged: onChanged,
+            items: {
+              'auto': 'Auto (assign automatically)',
```
andrei-toterman, line 44: Perhaps if there are no zones available, we shouldn't show the auto option. WDYT?
```suggestion
              if (hasAvailableZones)
                'auto': 'Auto (assign automatically)',
```

src/client/gui/lib/catalogue/zone_dropdown.dart
```diff
@@ -0,0 +1,55 @@
...
+              child:
+                  Icon(Icons.info_outline, size: 16, color: Colors.blue[700]),
+            ),
+          ],
+        ),
+        const SizedBox(height: 8),
+        (() {
+          final zones = ref.watch(zonesProvider);
+          final availableZones = zones.where((z) => z.available).toList();
+          final hasAvailableZones = availableZones.isNotEmpty;
```
andrei-toterman, line 38: you can move these at the top of the `build` method, then you don't need to do this whole anonymous function thing.

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -245,11 +261,21 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       ],
     );
 
+    final zones = ref.watch(zonesProvider);
+    final hasAvailableZones = zones.where((z) => z.available).isNotEmpty;
```
andrei-toterman, line 265: Now that you don't need `zones` directly in this widget, you can inline the thing. Could you also move this at the top of the `build` method, where all the other `ref.watch`es are happening?
```suggestion
    final hasAvailableZones = ref.watch(zonesProvider).any((z) => z.available);
```

src/client/gui/lib/catalogue/zone_dropdown.dart
levkropp: I have no strong opinions either way, but I don't want to show an empty dropdown either because that would look like a bug. So we could add an else condition that adds "No zones available", or just keep it as is IMO

src/client/gui/lib/vm_table/zones_dropdown_button.dart
```diff
@@ -0,0 +1,108 @@
...
+    final zones = ref.watch(zonesProvider);
+    final vmInfos = ref.watch(vmInfosProvider);
+
+    return PopupMenuButton(
+      tooltip: 'Change zone availability',
+      position: PopupMenuPosition.under,
+      child: OutlinedButton(
+        onPressed: null,
+        child: const Text('Zones', style: TextStyle(color: Colors.black)),
+      ),
```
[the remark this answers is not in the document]
levkropp: I have chosen Icons.dns
https://fonts.google.com/icons?selected=Material+Icons:dns:&icon.query=server&icon.size=24&icon.color=%231f1f1f&icon.platform=flutter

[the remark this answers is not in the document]
andrei-toterman: You can also download the SVG from the Figma design and add it to the `assets` directory. That's how we handle non-standard icons.

src/client/gui/lib/catalogue/zone_dropdown.dart
```diff
@@ -0,0 +1,54 @@
...
+            ),
+          ],
+        ),
+        const SizedBox(height: 8),
+        Dropdown<String>(
+          value: value,
+          onChanged: onChanged,
+          items: {
+            'auto': 'Auto (assign automatically)',
+            for (final zone in availableZones) zone.name: 'Zone ${zone.name}',
```
andrei-toterman, line 47: I think this will also lead to a `Zone zone1` situation
```suggestion
            for (final zone in availableZones) zone.name: zone.name,
```

src/client/gui/lib/vm_table/zones_dropdown_button.dart
```diff
@@ -0,0 +1,103 @@
...
+
+import '../providers.dart';
+
+class ZonesDropdownButton extends ConsumerWidget {
+  const ZonesDropdownButton({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final zones = ref.watch(zonesProvider);
+    final vmInfos = ref.watch(vmInfosProvider);
```
andrei-toterman, line 13: You don't need to watch this provider anymore here.

src/client/gui/lib/vm_table/zones_dropdown_button.dart
```diff
@@ -0,0 +1,103 @@
...
+
+    return PopupMenuButton(
+      tooltip: 'Change zone availability',
+      position: PopupMenuPosition.under,
+      child: OutlinedButton.icon(
+        onPressed: null,
+        icon: const Icon(Icons.dns, color: Colors.black, size: 16),
```
andrei-toterman, line 20: As mentioned in my other comment, you can get the SVG for the icon from Figma, then this will most likely look like this
```suggestion
        icon: SvgPicture.asset('assets/zones.svg'),
```

src/client/gui/lib/vm_table/zones_dropdown_button.dart
```diff
@@ -0,0 +1,103 @@
...
+      children: [
+        Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Text(zoneName,
+                style: TextStyle(
+                  fontSize: 16,
+                  fontWeight: FontWeight.w600,
+                  color: Colors.black,
+                )),
```
andrei-toterman, line 88: A comma here for better formatting results
```suggestion
                ),),
```

andrei-toterman (changes requested):
Alright, we're getting very close!

src/client/gui/lib/catalogue/zone_dropdown.dart
andrei-toterman: Yeah, I see your point. You can leave it as it is for now and perhaps discuss with design later. Another totally different approach could've been maybe to show all the zones, but the unavailable ones to be grayed-out and unselectable? Any way, that's an improvement that can come later.

src/client/gui/lib/vm_table/zones_dropdown_button.dart
levkropp: just VmInfos right? because we're using `zones` in `for (final zone in zones)`

andrei-toterman: Yes. I added the comment only to that line, but GH apparently chose to show the other lines as well.

andrei-toterman (changes requested):
Alright, everything looks good and works well! I was really ready to approve, but I got one last small improvement idea :sweat_smile:
In `providers.dart`, on line 163, we create a `DetailedInfoItem` from a `LaunchRequest`. This is done so that we can immediately display an instance as `launching` in the table even if the daemon cannot yet give us any info about that instance. That `DetailedInfoItem` could also take the zone from the launch request. So you'll just need to add this
```dart
DetailedInfoItem(
  ...
  zone: Zone(name: request.zone),
  ...
)
```
This should be the last request! :smile:

andrei-toterman (approved):
Alright, things are looking good! Good job, @levkropp!

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -5,12 +5,14 @@ import 'package:flutter/material.dart' hide Switch, ImageInfo;
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:rxdart/rxdart.dart';
 
+import '../dropdown.dart';
```
Sploder12, line 8: unused import

Sploder12 (commented):
Found a design oversight while testing. if you disable all zones and then delete all VMs, you can't launch VMs anymore since all zones are disabled and there is no way (from the GUI) to re-enable a zone without having at least 1 VM. wdyt?

andrei-toterman:
Ah, you're right @Sploder12! This needs to be brought to the attention of Design. IMO we should just remove the whole special 'No instances' screen.

levkropp:
> Found a design oversight while testing. if you disable all zones and then delete all VMs, you can't launch VMs anymore since all zones are disabled and there is no way (from the GUI) to re-enable a zone without having at least 1 VM. wdyt?

Great catch @Sploder12 ! I posted about this in the design channel to get feedback.

Sploder12 (approved):
Everything looks good and I didn't encounter any issues, good work!

vikorama:
Still haven't tested the package as I'm getting an error, but one change we definitely need to do is to remove Pseudo everywhere. This was the feedback from the Roadmap team-only.

vikorama:
Thank you @levkropp, a few things to look into:

- The switchers should have a hover state - the cursor changes to hover. We have [documentation](https://www.figma.com/design/Y0cqKbTG4rejU9xm2oh5pR/%F0%9F%92%A0-Vanilla---Core-component-library?node-id=2067-1958&t=1Am7OGc9cPdWTshL-0) for this component in web apps, maybe it'll help a bit to align Flutter.
- I have to click twice to switch a zone on and off. Steps to reproduce: 1. Expand the zone drop-down, 2. Point at a switcher and click, 3. Nothing happens, click again -> switcher reacts.
- Availability zones title and zone names should be 550 in weight (seems too bold at the moment). 
- Zone names font should be 16 px. Line height 24px.
- The subtitles' (# of running instances) font should be the default font. Now it looks like it doesn't have proper letter spacing. It should be the same as the VM names in the all instances table, also 16px. Line height 24px.
- The outer padding on the drop-down is 24px.
- The title should be left aligned.
- Remove border radius on the zone outlines. In our visual guidelines the border radius on buttons and all outlines is 0.
- The padding on the sides of the button feels uneven by 4-5px.
- [probably skip for now] [Button](https://canonical.github.io/react-components/?path=/docs/components-button--docs) outline should be the high-contrast color (707070) - we should either fix it for all button, or not touch it for now and fix later, cuz other button states are not consistent with our design system either.

vikorama:
I'm not sure if it is a part of this PR, but when one or more zones are disabled, we should indicate it to the user:
image

levkropp:
#### TLDR: Here is what it looks like now, with the Figma design for reference. Please let me know what you think!

image


Unfortunately, we cannot set a font weight of 550 to the best of my knowledge. If we really want, we could create/add a new font where one of the font values w100 w200 etc corresponds to a version of the font at 550 weight

---


Hi @vikorama , regarding your latest comments:

> The switchers should have a hover state - the cursor changes to hover. We have [documentation](https://www.figma.com/design/Y0cqKbTG4rejU9xm2oh5pR/%F0%9F%92%A0-Vanilla---Core-component-library?node-id=2067-1958&t=1Am7OGc9cPdWTshL-0) for this component in web apps, maybe it'll help a bit to align Flutter.


I do not have access to the documentation Figma you have linked. I have requested access. Do you mean something like the following? This is called a `SystemMouseCursors.click` in Flutter

image


> Availability zones title and zone names should be 550 in weight (seems too bold at the moment).

To the best of my knowledge, Flutter only supports specific font weights: w100, w200, w300, w400, w500, w600, w700, w800, w900. There is no w550. For a weight between w500 and w600, the closest valid options is FontWeight.w500 or FontWeight.w600. It's currently set to 600, but I can make it 500 if you prefer

> I have to click twice to switch a zone on and off. Steps to reproduce: 1. Expand the zone drop-down, 2. Point at a switcher and click, 3. Nothing happens, click again -> switcher reacts.

I am unable to reproduce this on Kubuntu 24.04. The switcher changes after a short delay for me after one click. There is a chance that this is also related to the incomplete implementation of availability/unavailiabilty in the branch, otherwise this may be a MacOS specific bug.

> Zone names font should be 16 px. Line height 24px.

The font size for zone names is set to 16px

```
Text(
  zoneName,
  style: TextStyle(
    fontSize: 16,
    fontWeight: FontWeight.w600,
    color: Colors.black,
  ),
),
```            
https://api.flutter.dev/flutter/painting/TextStyle/fontSize.html
There is a chance that this is being affected by a `textScaleFactor`. May I see a screenshot of what it looks like on your end?
I have added a `height` parameter of 1.5 to set the line height to 16 x 1.5 = 24 px
https://api.flutter.dev/flutter/painting/TextStyle/height.html

> * The subtitles' (# of running instances) font should be the default font. Now it looks like it doesn't have proper letter spacing. It should be the same as the VM names in the all instances table, also 16px. Line height 24px.

The subtitle is using the default font. There have been no changes made to kerning (letter spacing). I have updated it to now also be 16px with a line height of 24px

image


> * The outer padding on the drop-down is 24px.

:+1: 



> The padding on the sides of the button feels uneven by 4-5px.

:+1:
This was a result of the icons used. The padding has been made asymmetrical to account for this 



> * [probably skip for now] [Button](https://canonical.github.io/react-components/?path=/docs/components-button--docs) outline should be the high-contrast color (707070)

I agree that this should be in another PR

vikorama:
@levkropp Thank you for looking into all this! The look and the cursor are great! 

1. As I was making these visual comments, I realised I overdid it with the outlines. I simplified it and it looks lighter to me, without the unnecessary complexity: 

image

If you agree, could you please make these minor adjustments? 🙏

image

[Please see here for more details](https://www.figma.com/design/kY5ZFvQSyqIGXpIz7KNNju/AZs-first-iteration?node-id=103-6448&m=dev)

2. Could you please confirm if [my previous comment is included](https://github.com/canonical/multipass/pull/3979#issuecomment-3062029885) in this PR?

Issue #3991, opened 2025-03-21
Restrict docs and dev CI to changes in their respective domains


ricab:
To avoid failing unrelated PRs.

ricab:
Done.

PR #3993, opened 2025-03-21
[ci] Skip Linux CI on unrelated changes

ricab:
Addresses https://github.com/canonical/multipass/issues/3991. Not exhaustive, but at least filter out some paths.

MULTI-1899

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -9,6 +9,16 @@ on:
     - v[0-9]+.[0-9]+.[0-9]+
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/.jira_sync_config.yaml"
+      - ".github/workflows/automatic-doc-checks.yml"
+      - ".github/workflows/cla-check.yaml"
+      - "debian/**"
+      - "docs/**"
+      - "README.md"
+      - "renovate.json5"
+      - "SECURITY.md"
   merge_group:
     types: [checks_requested]
 


xmkg (approved):
LGTM

.github/workflows/linux.yml
```diff
@@ -9,6 +9,16 @@ on:
...
+    paths-ignore:
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/.jira_sync_config.yaml"
+      - ".github/workflows/automatic-doc-checks.yml"
+      - ".github/workflows/cla-check.yaml"
+      - "debian/**"
+      - "docs/**"
+      - "README.md"
+      - "renovate.json5"
+      - "SECURITY.md"
```
sharder996, line 21:
```suggestion
      - "SECURITY.md"
      - "LICENSE"
      - ".gitignore"
      - ".clang-format"
```

A few more if you're interested.

sharder996 (dismissed):

.github/workflows/linux.yml
ricab: I considered those too, but here's why I didn't include them:

- I think we include LICENSE in the package. It changes very infrequently anyway, so I didn't bother to confirm. I can include it if you're sure that is not the case.
- changes to `.gitignore` could cause builds to fail (for example, if they cause necessary source files to be uncommitted by mistake)
- changes to `.clang-format` could cause the Lint step to fail

Does this reasoning make sense to you?

sharder996:
Rebasing to include @xmkg changes for compatibility with CMake v4.0

sharder996 (approved):

xmkg (approved):

sharder996:
Kicking CI to run again.

ricab:
Ahh no, that's because the file really is different on the other side... I'll submit a private PR.

PR #3996, opened 2025-03-21
[ci] Fix description of custom checkout action

ricab:
The action claimed it checked out Multipass code, but that is actually
done elsewhere via the standard GitHub "checkout" action.


ricab:
I don't have any further fixes at this point, so may as well merge.

Sploder12 (approved):
Yep, that's more accurate!

sharder996 (approved):

PR #3998, opened 2025-03-22
Split tests into multiple executables

levkropp:
This pull request reorganizes the test suite in the `tests/CMakeLists.txt` file by splitting the tests into multiple categories and creating separate executables for each category.

Before:
`multipass_tests` = 3857 tests from 233 test suites ran. (42261 ms total)

Now:
`multipass_tests` = 1312 tests from 111 test suites ran. (12894 ms total)
`multipass_client_ssh_tests` = 820 tests from 56 test suites ran. (20483 ms total)
`multipass_daemon_tests` = 322 tests from 30 test suites ran. (2441 ms total)
`multipass_formatter_tests` = 373 tests from 9 test suites ran. (120 ms total)
`multipass_memory_tests` = 823 tests from 4 test suites ran. (36 ms total)
`multipass_network_tests` = 89 tests from 8 test suites ran. (2622 ms total)
`multipass_qemu_tests` = 131 tests from 16 test suites ran. (2271 ms total)

Reorganization of test suite:

* Created separate test source sets for SSH, Memory, Formatter, Network, Client, and QEMU tests, and updated the `add_executable` commands to create distinct test executables for each category.
* Added common include directories and linked libraries for the new test executables, ensuring that all necessary dependencies are included.
* Updated the `add_test` commands to include the new test executables, enabling them to be run as part of the test suite.
* Added mock binaries for testing and set dependencies for the new test executables to ensure the mock binaries are built before running the tests.


xmkg:
Hi @levkropp, that's looking great! I just have one topic to discuss while we're at it.

The different categories are already better than what we have; shall we go another step further and make each test source a separate executable? This way, each of the test suites is isolated at the process level, and we can also parallelize their execution if we want to. Also, we can have better control over what each test suite depends on. 

ctest is capable of curating tests from different executables, so the number of executables won't matter. It's designed to work in this way anyway.

@ricab @georgeliao opinions?

ricab:
> Agreed. Having separate executables will incur extra link time. However, if we introduce an intermediate dynamic library that bundles all the static ones, with the use of [--whole-archive](https://linux.die.net/man/1/ld) (or [/wholearchive](https://learn.microsoft.com/en-us/cpp/build/reference/wholearchive-include-all-library-object-files?view=msvc-170) for MSVC), we can avoid that completely.

But then we're back in the same situation as today, right? Every time we change something, we need to relink the world (now into that lib). My hope with separate executables was that we would need, for example, daemon stuff in only one executable, client stuff in another, ssh in yet another, etc. 

Parallelizing tests would be nice too, but a smaller gain in relative terms. As you point out, we don't need to split executables for that.

Let's perhaps discuss live when we can to align on what the goals are and how to best achieve them.

georgeliao:
> We can grab the suite names and then use them as a filter, i.e., .\build\bin\multipass_tests.exe --gtest_filter="MemorySize*" to execute them individually. This way, each of the suites still gets executed in a dedicated process, and we can parallelize, too -- all while without having to change how we build the tests.
> 
> The thing I've described is trivial enough to do, maybe even with CMake, so we don't need any sophisticated tool, but there's an unofficial Python tool by Google that does exactly this: https://github.com/google/gtest-parallel.

@xmkg About parallelization, currently we have ctest which is granular to executable. So on this PR, the 7 unit test executable can be parallelized. As to ctest parallel run vs gtest-parallel, I think gtest-parallel is granular to each test case, which is more fine-grained. Therefore, there is some value to it. Apart from this comparison,  the `ctest -j8` run (from the <multipass_build_folder>) on this PR shows some undefined behavior, sometimes it hangs, sometimes it failed some tests. Likely it is due to shared resources between several unit test executable, but this is probably not related to this PR as well.

ricab:
FWIW, 5 years ago I tried to [randomize tests](https://github.com/canonical/multipass/pull/1399) and found that they couldn't handle it. Unfortunately it didn't cross the priority threshold and 3 years later it remained open, so I closed. We had meagre resources back then, but that's finally changing now. I wouldn't be surprised with similar findings now...

xmkg:
@levkropp it seems like this one will need a private side PR, too: https://github.com/canonical/multipass-private/actions/runs/14360225482/job/40259517453

tests/CMakeLists.txt
```diff
@@ -30,42 +33,250 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+)
+
+# Base include directories needed by all test components
+set(BASE_INCLUDE_DIRS
+  ${CMAKE_SOURCE_DIR}
+  ${CMAKE_SOURCE_DIR}/src
+  ${CMAKE_SOURCE_DIR}/include # For multipass headers
+  ${CMAKE_SOURCE_DIR}/3rd-party/fmt/include # For fmt headers
+  ${CMAKE_SOURCE_DIR}/3rd-party/scope_guard # For scope_guard.hpp
+)
```
xmkg, line 63: Instead of listing those files and dependencies, we could create an INTERFACE library to encapsulate them properly:

```cmake

add_library(test_common_deps INTERFACE)

target_include_directories(test_common_deps INTERFACE
  ${gtest_SOURCE_DIR}/googlemock/include
  ${gtest_SOURCE_DIR}/googletest/include
  ${CMAKE_SOURCE_DIR}/src/ssh
  ${CMAKE_SOURCE_DIR}/src/sshfs_mount
)

target_link_libraries(test_common_deps INTERFACE
  Qt6::Core
  Qt6::Network
  ssh_test
  sftp_test
  ssh_client_test
  sshfs_mount_test
  gmock
  gtest
)

add_library(base_include_dirs INTERFACE)

target_include_directories(base_include_dirs INTERFACE
  ${CMAKE_SOURCE_DIR}
  ${CMAKE_SOURCE_DIR}/src
  ${CMAKE_SOURCE_DIR}/include # For multipass headers
  ${CMAKE_SOURCE_DIR}/3rd-party/fmt/include # For fmt headers
  ${CMAKE_SOURCE_DIR}/3rd-party/scope_guard # For scope_guard.hpp
)
```

This way, we can propagate all of this just by linking to it:

```cmake
add_library(test_settings_obj OBJECT
  test_global_settings_handlers.cpp
  test_instance_settings_handler.cpp
  test_persistent_settings_handler.cpp
  test_remote_settings_handler.cpp
  test_setting_specs.cpp
  test_settings.cpp
)
target_include_directories(test_settings_obj PRIVATE
  ${CMAKE_SOURCE_DIR}/src/settings
)
target_link_libraries(test_settings_obj PRIVATE
  settings
  test_common_deps
  base_include_dirs
)
```
I also noticed that TEST_COMMON_INCLUDE_PATHS and BASE_INCLUDE_DIRS are included together, so we can ditch the dedicated `base_include_dirs` target and merge it into `test_common_deps` instead.

tests/CMakeLists.txt
```diff
@@ -30,42 +33,250 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+target_link_libraries(test_network_obj PRIVATE ${TEST_COMMON_DEPS})
+
+add_library(test_daemon_obj OBJECT
+  daemon_test_fixture.cpp
+)
+target_include_directories(test_daemon_obj PRIVATE
+  ${BASE_INCLUDE_DIRS}
+  ${TEST_COMMON_INCLUDE_PATHS}
+  ${CMAKE_SOURCE_DIR}/src/daemon
```
xmkg, line 156: Listing includes paths per-target seems a little bit verbose to me. Because the includes will only hurt the build performance when they're actually included, we can just list all the include paths at the top:

```cmake
include_directories(/*list of all include paths used by the tests*/)
```

The downside would be that all the targets would be able to use any headers listed in the include path, which IMO is reasonable given that we don't have to micro-manage the include paths per target. What do you think?

tests/CMakeLists.txt
```diff
@@ -30,42 +33,250 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
-  test_base_snapshot.cpp
-  test_base_virtual_machine.cpp
-  test_base_virtual_machine_factory.cpp
-  test_basic_process.cpp
+  mock_sftp.cpp
+  mock_sftpserver.cpp
+)
+
+# Common source combinations using object libraries
+set(MINIMAL_TEST_OBJECTS
```
xmkg, line 189: I must say that I'm not thrilled about the idea of using lists for this. I'll think about it.

tests/CMakeLists.txt
```diff
@@ -220,10 +497,43 @@ set_target_properties(mock_process
...
 add_dependencies(multipass_tests mock_process)
+add_dependencies(multipass_daemon_tests mock_process)
+add_dependencies(multipass_client_ssh_tests mock_process)
+add_dependencies(multipass_memory_tests mock_process)
+add_dependencies(multipass_formatter_tests mock_process)
+add_dependencies(multipass_network_tests mock_process)
+
+if(MSVC)
+  add_definitions(-DGTEST_LANG_CXX11=1)
+  target_compile_options(multipass_tests PRIVATE /bigobj)
```
xmkg, line 508: Considering the fact that `/bigobj` is not a harmful flag in general, `add_compile_options(/bigobj)` would suffice here.

tests/CMakeLists.txt
```diff
@@ -220,10 +497,43 @@ set_target_properties(mock_process
...
+elseif(APPLE)
+  set(CMAKE_CXX_FLAGS "-Wno-inconsistent-missing-override")
+  add_definitions(-DHOST_ARCH="${HOST_ARCH}")
+endif()
 
-target_include_directories(multipass_tests
-  BEFORE
-    PRIVATE ${CMAKE_SOURCE_DIR}/src/platform/backends/shared/${MULTIPASS_PLATFORM}
+# Add platform include directory for each executable
+target_include_directories(multipass_tests BEFORE
```
xmkg, line 520: idem: target_include_directories

tests/CMakeLists.txt
```diff
@@ -151,43 +381,85 @@ target_compile_definitions(ssh_client_test PRIVATE
...
+target_link_libraries(multipass_client_ssh_tests ${CLIENT_SSH_TEST_LIBRARIES})
+target_link_libraries(multipass_memory_tests ${MEMORY_TEST_LIBRARIES})
+target_link_libraries(multipass_formatter_tests ${FORMATTER_TEST_LIBRARIES})
+target_link_libraries(multipass_network_tests ${NETWORK_TEST_LIBRARIES})
+target_link_libraries(multipass_tests ${DAEMON_TEST_LIBRARIES} poco_utils)
+
 # need https://github.com/google/googletest/commit/a09ea700d32bab83325aff9ff34d0582e50e3997 before this can be removed
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
   target_compile_options(multipass_tests PRIVATE -Wno-inconsistent-missing-override)
+  target_compile_options(multipass_daemon_tests PRIVATE -Wno-inconsistent-missing-override)
```
xmkg, line 441: idem: target_compile_options

tests/CMakeLists.txt
```diff
@@ -151,43 +381,85 @@ target_compile_definitions(ssh_client_test PRIVATE
...
+add_test(NAME multipass_client_ssh_tests COMMAND multipass_client_ssh_tests)
+add_test(NAME multipass_memory_tests COMMAND multipass_memory_tests)
+add_test(NAME multipass_formatter_tests COMMAND multipass_formatter_tests)
+add_test(NAME multipass_network_tests COMMAND multipass_network_tests)
+add_test(NAME multipass_tests COMMAND multipass_tests)
 
 foreach(BACKEND IN LISTS MULTIPASS_BACKENDS)
   string(TOUPPER ${BACKEND}_ENABLED DEF)
   target_compile_definitions(multipass_tests PRIVATE ${DEF})
+  target_compile_definitions(multipass_daemon_tests PRIVATE ${DEF})
```
xmkg, line 458: How about `add_compile_definitions()` to avoid the hassle?

tests/CMakeLists.txt
```diff
@@ -196,6 +468,11 @@ if (${MANIFEST_ARCH} STREQUAL "x86_64")
 endif()
 
 target_compile_definitions(multipass_tests PRIVATE -DMANIFEST_ARCH="${MANIFEST_ARCH}")
+target_compile_definitions(multipass_daemon_tests PRIVATE -DMANIFEST_ARCH="${MANIFEST_ARCH}")
```
xmkg, line 471: idem: target_compile_definitions

xmkg (changes requested):
Good initial version @levkropp! It needs refinement in some places.

tests/CMakeLists.txt
levkropp: I was running into issues with `add_compile_options(/bigobj)` because of how late the compiler flags were added in the CMakeLists. Moving this code to near the top of the file resolved the issue and I was able to use `add_compile_options(/bigobj)` successfully

tests/CMakeLists.txt
xmkg: How about ditching the lists and moving them to where they're being used?:

```cmake

add_library(process_mocks STATIC
  mock_process_factory.cpp
  stub_process_factory.cpp
  reset_process_factory.cpp
)
target_link_libraries(process_mocks 
  Qt6::Core
  gmock
  gtest
)

add_library(ssh_sftp_mocks STATIC 
  mock_ssh.cpp
  mock_ssh_client.cpp
  mock_sftp.cpp
  mock_sftpserver.cpp
)

target_link_libraries(ssh_sftp_mocks
  libssh
)

macro(declare_library_group NAME)
  add_library(${NAME} INTERFACE)
  target_link_libraries(${NAME} INTERFACE ${ARGN})
endmacro()


declare_library_group(minimal_test_objects test_minimal_obj test_openssl_obj)

# Daemon test executable
add_executable(multipass_daemon_tests
  test_daemon.cpp
  test_daemon_authenticate.cpp
  test_daemon_clone.cpp
  test_daemon_find.cpp
  test_daemon_launch.cpp
  test_daemon_mount.cpp
  test_daemon_snapshot_restore.cpp
  test_daemon_start.cpp
  test_daemon_suspend.cpp
  test_daemon_umount.cpp
)
target_link_libraries(multipass_daemon_tests
  minimal_test_objects process_mocks ssh_sftp_mocks test_fs_obj
  test_json_obj test_daemon_obj test_network_obj
)

# Client and SSH test executable
add_executable(multipass_client_ssh_tests
  test_sftp_client.cpp
  test_sftpserver.cpp
  test_ssh_client.cpp
  test_ssh_exec_failure.cpp
  test_ssh_key_provider.cpp
  test_ssh_process.cpp
  test_ssh_session.cpp
  test_sshfs_server_process_spec.cpp
  test_sshfsmount.cpp
  test_sshfs_mount_handler.cpp
  test_sftp_dir_iterator.cpp
  test_sftp_utils.cpp
  test_cli_client.cpp
  test_cli_prompters.cpp
  test_client_cert_store.cpp
  test_client_common.cpp
  test_client_logger.cpp
  test_alias_dict.cpp
)

target_link_libraries(multipass_client_ssh_tests
  minimal_test_objects process_mocks ssh_sftp_mocks test_fs_obj test_json_obj test_daemon_obj
)

# Memory test executable
add_executable(multipass_memory_tests
  test_memory_size.cpp
)

target_link_libraries(multipass_memory_tests
  minimal_test_objects
)

# Formatter test executable
add_executable(multipass_formatter_tests
  test_format_utils.cpp
  test_output_formatter.cpp
)
target_link_libraries(multipass_formatter_tests
  minimal_test_objects
)

# Network test executable
add_executable(multipass_network_tests
  test_url_downloader.cpp
  test_ip_address.cpp
  test_ubuntu_image_host.cpp
  test_custom_image_host.cpp
  test_new_release_monitor.cpp
)

target_link_libraries(multipass_network_tests
  minimal_test_objects test_network_obj test_fs_obj
)

# Main tests executable
add_executable(multipass_tests
  test_base_snapshot.cpp
  test_base_virtual_machine.cpp
  test_base_virtual_machine_factory.cpp
  test_basic_process.cpp
  test_cloud_init_iso.cpp
  test_common_callbacks.cpp
  test_constants.cpp
  test_delayed_shutdown.cpp
  test_disabled_copy_move.cpp
  test_id_mappings.cpp
  test_image_vault.cpp
  test_image_vault_utils.cpp
  test_json_utils.cpp
  test_mock_standard_paths.cpp
  test_mount_handler.cpp
  test_petname.cpp
  test_private_pass_provider.cpp
  test_qemuimg_process_spec.cpp
  test_simple_streams_index.cpp
  test_simple_streams_manifest.cpp
  test_singleton.cpp
  test_ssl_cert_provider.cpp
  test_timer.cpp
  test_top_catch_all.cpp
  test_utils.cpp
  test_yaml_node_utils.cpp
  test_vm_mount.cpp
  test_with_mocked_bin_path.cpp
  test_blueprint_provider.cpp
  test_file_ops.cpp
  test_recursive_dir_iter.cpp
  test_log.cpp
  test_exception.cpp
  test_permission_utils.cpp
  test_standard_logger.cpp
)

target_link_libraries(multipass_tests
  minimal_test_objects process_mocks ssh_sftp_mocks test_fs_obj test_json_obj
  test_daemon_obj test_network_obj test_network_parser_obj test_settings_obj
)

```

tests/CMakeLists.txt
```diff
@@ -151,51 +334,84 @@ target_compile_definitions(ssh_client_test PRIVATE
...
-  daemon
-  delayed_shutdown
-  gmock_main
-  gtest_main
-  ip_address
-  iso
-  network
-  petname
+# Base libraries needed by all tests
+set(BASE_TEST_LIBRARIES
```
xmkg, line 338: You can use the declare_library_group macro suggested above to bundle this.

tests/CMakeLists.txt
```diff
@@ -151,51 +334,84 @@ target_compile_definitions(ssh_client_test PRIVATE
...
+  client # For shared functionality
+  gmock_main # For mocking
+  gtest_main # For testing framework
+  premock # For mocking
+  scope_guard # For cleanup
+  utils_test # For utility functions
+)
+
+# Minimal networking libraries
+set(MEMORY_TEST_LIBRARIES
```
xmkg, line 348: To avoid clutter, we should move non-common library lists to their respective target_link_libraries instead.

tests/CMakeLists.txt
```diff
@@ -30,42 +30,219 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+  ssh_test
+  sftp_test
+  ssh_client_test
+  sshfs_mount_test
+  gmock
+  gtest
+)
+
+# Settings test object library
+add_library(test_settings_obj OBJECT
```
xmkg, line 76: After ditching the source lists, we can convert the object libraries to static.

Ref:
https://discourse.cmake.org/t/project-modularization-using-object-libraries/3825/11

tests/CMakeLists.txt
```diff
@@ -30,42 +30,219 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+  ${MINIMAL_TEST_OBJECTS}
+  ${TEST_PROCESS_SOURCES}
+  $<TARGET_OBJECTS:test_fs_obj>
+  ${TEST_SSH_SOURCES}
+  $<TARGET_OBJECTS:test_json_obj>
+  $<TARGET_OBJECTS:test_daemon_obj>
+)
+
+set(DAEMON_TEST_OBJECTS
+  ${MINIMAL_TEST_OBJECTS}
```
xmkg, line 176: I feel like we're trying to solve a problem which is already solved before. $<TARGET_OBJECTS:obj_lib_name> is effectively a static library, except that a static library also handles the linking. Moreover, the object libraries can be directly linked against and will have the same effect as expanding the objects into the source part.

xmkg (changes requested):
Right now, it's hard to reason about the CMake code since a target's nitty-gritty is listed in a bunch of different places. Overall, we should avoid clutter and keep things as neat as possible.

xmkg:
@levkropp you'll also need to update the CI pipeline since it's not using CTest to run the tests:

https://github.com/canonical/multipass/blob/fb14499de8be81f9b921b3195a1e9c347e743a45/.github/workflows/linux.yml#L206-L210

Also, there are a couple of direct references to the multipass_tests which will need to be updated to accommodate the new multiple-executable approach:

https://github.com/canonical/multipass/blob/fb14499de8be81f9b921b3195a1e9c347e743a45/.github/workflows/linux.yml#L264

tests/CMakeLists.txt
georgeliao: I personally feel the object library usage seems to be right here in terms of the unit does not needed to exported and avoid unnecessary linking. Can you elaborate on some pros static lib might possesses and cons object lib might have?

ricab: I wasn't familiar with CMake's idea of "object libraries" and I just read a little bit on it. Having a few extra symbols in the test executables sounds like a secondary concern to me, so it seems to me that this is basically a matter of the following trade-off: if we're iterating on a single test file, relinking from an existing static library is faster. But if we're changing something in "production" code that affects only a few objects, relinking the whole static library would be slower, if most of the executable targets are unaffected. Isn't that so?

xmkg: > But if we're changing something in "production" code that affects only a few objects, relinking the whole static library would be slower, if most of the executable targets are unaffected. Isn't that so?

Yes, but with a caveat. With object libraries, the linker would get all of the symbols in an object file without any filtering, so the linker's input size is larger, and a larger input size might increase the link time, too. That's not the case with the static libraries, as only the used symbols are included in the final archive file, so the symbol set is leaner.

Another aspect is that since the mold/lld patch is now landed, the link time is less of an issue now. The benefits of using one over the other might not be significant -- link speed-wise. In this case, I would be inclined towards using the static/shared over an object library.

So it boils down to what advantage the OBJECT library would provide in the status quo.

ricab: Hmm, there should be very few unused symbols I think. Again, I guess we need measurements. It might not even make sense to split any longer?

levkropp:
Testing with `touch tests/test_utils.cpp` and comparing between the current tests/CMakeLists in `main` vs the one in this repo that contains split test files and object binaries, only shows a 1-3% improvement in build time when using `mold`. When using `ld` the improvement in build time from these object libraries is a few seconds. `mold` is responsible for the majority of improvements to linking speed (re-linking with mold with no object binaries saves ~40%, relinking with ld with object binaries saves ~15%, relinking with mold with object binaries saves ~3% compared to not using object binaries with mold).

#### Old CMakeLists (current main) with default linker (ld)

Full build:
real    3m12.300s
user    24m45.839s
sys     1m44.742s

`touch tests/test_utils.cpp` and relinking `multipass_tests`
real    0m14.879s
user    0m13.226s
sys     0m3.505s

#### Old CMakeLists (current main) with mold
Full build:
real    3m1.819s
user    24m33.204s
sys     1m40.439s

`touch tests/test_utils.cpp` and relinking `multipass_tests`

real    0m9.007s
user    0m8.544s
sys     0m1.575s

#### New CMakeLists (split test executables, object libraries) with default linker (ld)
Full build:
real    3m9.086s
user    25m6.402s
sys     1m49.464s

`touch tests/test_utils.cpp` and relinking `multipass_tests`
real    0m12.733s
user    0m11.794s
sys     0m3.033s

#### New CMakeLists (split test executables, object libraries) with mold

Full build:
real    3m2.514s
user    24m51.858s
sys     1m44.613s

`touch tests/test_utils.cpp` and relinking `multipass_tests`
real    0m8.744s
user    0m8.730s
sys     0m1.763s

ricab:
Hey @levkropp, while this is hot in your terminal, what happens if you touch `utils.cpp` rather than `test_utils.cpp`? Just object libraries vs current main with mold would be enough.

levkropp:
> what happens if you touch `utils.cpp` rather than `test_utils.cpp`? Just object libraries vs current main with mold would be enough.

#### Old CMakeLists (current main) with mold

real    3m7.654s
user    25m24.351s
sys     1m44.390s

`touch src/utils/utils.cpp`, relinking the single multipass_test executable

real    0m7.873s
user    0m7.898s
sys     0m1.872s


#### New CMakeLists (split test executables, object libraries) with mold

real    3m7.205s
user    25m9.387s
sys     1m45.606s

`touch src/utils/utils.cpp`, relinking all test executables

real    0m7.694s
user    0m8.164s
sys     0m2.086s

`mold` links so quickly, that an extra potential time saving from object libraries, or a potential time loss from multiple test executables, seems negligible. I can do more rounds for both tests if you prefer

ricab:
OK, I think this has become moot really, at least as far as the original goal of speeding up iteration time. Let's discuss how to take it from here in the sprint perhaps

Issue #4005, opened 2025-03-26
Mounts won't get activated when daemon restarts for persistent backends


xmkg:
**Describe the bug**
For persistent backends (e.g., lxd) where the VM can continue executing while the daemon is shut down, the mounts won't get re-activated upon daemon start.

**To Reproduce**

```bash
# Set the driver to something persistent, e.g., lxd
 multipass set local.driver=lxd
# Launch a test VM
multipass launch noble --name mount-reactivate-test
# Mount a directory from host to VM
multipass mount --type classic /home/mkg/Downloads mount-reactivate-test:/home/ubuntu/mount-dir
# Check if mount is recognized:
multipass info mount-reactivate-test
# Outputs:
# Name:           mount-reactivate-test
# State:          Running
# IPv4:           10.211.127.92
# Release:        Ubuntu 24.04.2 LTS
# Image hash:     1e461e0e46d4 (Ubuntu 24.04 LTS)
# CPU(s):         1
# Load:           0.14 0.10 0.04
# Disk usage:     1.9GiB out of 9.6GiB
# Memory usage:   357.8MiB out of 946.3MiB
# Mounts:         /home/mkg/Downloads => /home/ubuntu/mount-dir
#                    UID map: 1000:default
#                   GID map: 1000:default

# Try to list the mounted directory:
multipass exec mount-reactivate-test ls /home/ubuntu/mount-dir
# Output: <non-empty list of files>

# Restart the daemon (this won't kill the VM):
sudo snap restart multipass

# Check if mount is still there:
multipass info mount-reactivate-test
# Outputs:
# Name:           mount-reactivate-test
# State:          Running
# IPv4:           10.211.127.92
# Release:        Ubuntu 24.04.2 LTS
# Image hash:     1e461e0e46d4 (Ubuntu 24.04 LTS)
# CPU(s):         1
# Load:           0.14 0.10 0.04
# Disk usage:     1.9GiB out of 9.6GiB
# Memory usage:   357.8MiB out of 946.3MiB
# Mounts:         /home/mkg/Downloads => /home/ubuntu/mount-dir
#                    UID map: 1000:default
#                   GID map: 1000:default

# Try to list directory contents:
multipass exec mount-reactivate-test ls /home/ubuntu/mount-dir
# Output: <empty>

```

**Expected behavior**
The mount should be re-activated upon restart and mount contents should become available.

**Workaround**
Remove the mount and attach it from scratch.

georgeliao:
[...]

PR #4006, opened 2025-03-26
[daemon] fix unchecked iterator access in async start mounts

xmkg:


xmkg:
Testing this has proven to be hard since the `async_wait_for_ssh_and_start_mounts_for` function is a templated non-virtual private function.  Also, triggering this condition depends on a synchronization failure or a race condition.

xmkg:
> Testing this has proven to be hard since the `async_wait_for_ssh_and_start_mounts_for` function is a templated non-virtual private function. Also, triggering this condition depends on a synchronization failure or a race condition.

I stumbled upon this one while trying to fix the #4005, specifically by calling the on_restart() function in the VM constructor.  That resulted in a crash since the operative_instances map is actively being populated during the constructor call:

https://github.com/canonical/multipass/blob/13f8025aed6953fe3e8296de56c0721898b091d0/src/daemon/daemon.cpp#L1370-L1371

Sploder12 (approved):
Good catch! There is still the issue of thread safety, but at least it might not break as bad hopefully if we're lucky.

src/daemon/daemon.cpp
```diff
@@ -3425,8 +3425,15 @@ mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::string& name, con
     fmt::memory_buffer errors;
     try
     {
-        auto it = operative_instances.find(name);
-        auto vm = it->second;
+        const auto& it = operative_instances.find(name);
```
georgeliao, line 3428: `const auto&` and `const auto` should not make a difference here since the `find` function returns value.

georgeliao (approved):
Nice catch, it looks good to me.

PR #4008, opened 2025-03-27
Fix invalid macro definition in CMakeLists.txt

rshrj:
This PR fixes an issue in `src/platform/CMakeLists.txt` where `add_compile_definitions()` is used with a `-D` prefix:

```cmake
add_compile_definitions(-DMULTIPASS_JOURNALD_ENABLED)
```

This results in a malformed macro definition `-D-DMULTIPASS_JOURNALD_ENABLED`, which causes a compiler error:

```shell
<command-line>: error: macro names must be identifiers
```

The fix is to pass only the macro name, as `add_compile_definitions()` automatically adds the `-D` prefix:

```cmake
add_compile_definitions(MULTIPASS_JOURNALD_ENABLED)
```

Tested in a clean Debian 12 environment using Make, where this issue was originally triggered.

### DIFF
--- a/src/platform/CMakeLists.txt
+++ b/src/platform/CMakeLists.txt
@@ -45,7 +45,7 @@ endfunction()
 option(MULTIPASS_ENABLE_SYSLOG "Use syslog for logging instead of journald" OFF)
 
 if(NOT MULTIPASS_ENABLE_SYSLOG)
-  add_compile_definitions(-DMULTIPASS_JOURNALD_ENABLED)
+  add_compile_definitions(MULTIPASS_JOURNALD_ENABLED)
 endif()
 
 add_target(platform)


xmkg:
Hi @rshrj, thanks for the contribution!

Good catch, I wonder why it hasn't caused any trouble before.

xmkg:
Hey @rshrj, could you sign the [CLA](https://ubuntu.com/legal/contributors) using the same GitHub username and email when you get a chance?

rshrj:
> Hey @rshrj, could you sign the [CLA](https://ubuntu.com/legal/contributors) using the same GitHub username and email when you get a chance?

Done!

rshrj:
> Hi @rshrj, thanks for the contribution!
> 
> Good catch, I wonder why it hasn't caused any trouble before.

Seems like `MULTIPASS_ENABLE_SYSLOG` is set true for most people

xmkg:
> Seems like `MULTIPASS_ENABLE_SYSLOG` is set true for most people

Actually, it's the opposite: journald is the default logger for Linux. 

According to the [CMake documentation for add_compile_definitions](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions), the command is supposed to strip the leading "-D" from the input, which is not happening in your case. I see that CMake 3.25.1 is the latest CMake version for the bookworm, so that figures.

It's a nice addition (or should I say removal) regardless :)

xmkg (approved):
LGTM.

ricab (approved):

PR #4010, opened 2025-03-27
[daemon] ensure all outstanding futures are done before exit

xmkg:
The daemon object hosts to QFuture and QFutureWatcher objects, which may be still active while the daemon is being destructed.

This patch ensures that the destructor call waits for all outstanding futures, and all the raised signals during the lifetime of the futures are delivered to the recipients.

MULTI-1911

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -38,6 +38,7 @@
 #include "mock_vm_blueprint_provider.h"
 #include "mock_vm_image_vault.h"
 #include "path.h"
+#include "signal.h"
 #include "stub_virtual_machine.h"
 #include "tracking_url_downloader.h"
 
@@ -386,15 +387,12 @@ TEST_F(Daemon, ensure_that_on_restart_future_completes)
         .WillOnce(Return(mp::VirtualMachine::State::stopped));
     EXPECT_CALL(*mock_vm, start).Times(1);
 
-    std::mutex mtx;
-    std::condition_variable cv;
-    bool ready = false;
+    mpt::Signal signal;
     // update_state is called by the finished() handler of the future. If it's called, then
     // everything's ok.
-    EXPECT_CALL(*mock_vm, update_state).WillOnce([&cv, &mtx, &ready] {
+    EXPECT_CALL(*mock_vm, update_state).WillOnce([&signal] {
         // Ensure that update_state is delayed until daemon's destructor call.
-        std::unique_lock lock(mtx);
-        cv.wait(lock, [&] { return ready; });
+        signal.wait();
         // Wait a bit to ensure that daemon's destructor has been run
         std::this_thread::sleep_for(std::chrono::milliseconds{50});
     });
@@ -403,11 +401,7 @@ TEST_F(Daemon, ensure_that_on_restart_future_completes)
 
     {
         mp::Daemon daemon{config_builder.build()};
-        {
-            std::lock_guard lock(mtx);
-            ready = true;
-        }
-        cv.notify_one();
+        signal.signal();
     }
 }
 


src/daemon/daemon.cpp
```diff
@@ -1465,6 +1465,25 @@ mp::Daemon::~Daemon()
...
+        /**
+         * Wait until all futures are finished, so there will
+         * be no outstanding requests left behind. Otherwise, the
+         * futures might be interrupted halfway and the actions
+         * associated with the futures are not going to be
+         * executed.
+         */
+        for (auto& [_, watcher] : async_future_watchers)
+        {
+            watcher->waitForFinished();
```
xmkg, line 1478: This line is not always going to be hit by the unit tests since it depends on the timing.

src/daemon/daemon.cpp
```diff
@@ -1465,6 +1465,37 @@ mp::Daemon::~Daemon()
...
+         * AsyncPeriodicDownloadTask maintain its own QFutureWatcher.
+         * Tell it to wrap things up.
+         *
+         * We would like to do that here explicitly instead of in a destructor
+         * since update_manifests_all_task might be using resources from the daemon
+         * and their destruction might be happening before the update_manifests_all_task
+         * object, which makes it troublesome to do this in the AsyncPeriodicDownloadTask
+         * destructor.
+         */
+        update_manifests_all_task.shutdown();
```
xmkg, line 1491: This fixes the notorious sporadic CI daemon unit test crash.

xmkg:
> > Caught another crash related to this. It happens when the daemon constructor throws during a VM construction:
> 
> @xmkg For reference only, one trick we can use is a private constructor which does not throw before the main body of the `daemon` constructor. In this case, if the daemon constructor throws, the destructor is called.

Hi @georgeliao, thanks for chiming in -- excellent idea!  It's a nice trick, and could work in our scenario. I tried it and it works like a charm :)

https://gcc.godbolt.org/z/oYPcvvaP1

tests/test_daemon.cpp
```diff
@@ -356,6 +359,52 @@ TEST_F(Daemon, blueprintsURLOverrideIsCorrect)
...
+    config_builder.vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
+
+    // This VM was running before, but not now.
+    auto mock_vm = std::make_unique<NiceMock<mpt::MockVirtualMachine>>("yakety-yak");
+    EXPECT_CALL(*mock_vm, current_state)
+        .WillOnce(Return(mp::VirtualMachine::State::stopped))
+        .WillOnce(Return(mp::VirtualMachine::State::stopped));
+    EXPECT_CALL(*mock_vm, start).Times(1);
+
+    mpt::Signal signal;
```
georgeliao, line 390: Nice interplay of the `mpt::Signal` in the main thread and the side thread.

georgeliao:
@xmkg 
Good work, the `timer.stop()`
 + `waitForFinished()` + `QCoreApplication::processEvents(QEventLoop::AllEvents);` in the destructor looks idiomatic QT to me.

georgeliao (approved):

xmkg:
@georgeliao thanks for the review!

Right now, I'm also looking into moving the future cleanup calls into their respective types/classes, but that could be handled in a separate PR -- maybe in the "handle cleanup properly when the daemon constructor throws" PR. Meanwhile, I'm happy to land this one.

@levkropp this one's all yours now. Let me know if you want me to make any changes, or explanation.

levkropp (approved):
LGTM! Testing on my local machine showed no regressions to the daemon functionality or test suite with these changes

xmkg:
@georgeliao @levkropp thanks folks!

PR #4016, opened 2025-03-28
[daemon] register the signal handler immediately on start

xmkg:
The signal handler should be the first thing to register since the controller (e.g. user, snapd) can send signals immediately. Right now, the daemon either freezes or crashes when it's being launched and then immediately killed via Ctrl-C, which can also observed via running `snap restart multipass` in quick succession.

The signal handler uses `QCoreApplication::quit()` to wrap everything up, but use of `QCoreApplication::quit()` requires proper initialization of the QT resources first, so the signal handler should wait until that's done.

In order to ensure that this patch does the following:

- signal.h is moved from test/ to multipass/
- Added a top-level mp::Signal object to facilitate communication between the signal handler thread and the main thread
- mp::top_catch_all can now accommodate a callable as a fallback_return, so the code can now execute additional logic when the wrappee throws
- Added `QThreadPool::globalInstance()->waitForDone()` to ensure that the tasks dispatched through QConcurrent::run() are waited for completion. (otherwise QCoreApplication's destructor might block)

MULTI-1918

### DIFF
--- a/include/multipass/top_catch_all.h
+++ b/include/multipass/top_catch_all.h
@@ -99,7 +99,14 @@ inline auto multipass::top_catch_all(std::string_view log_category,
         detail::error(log_category);
     }
 
-    return std::forward<decltype(fallback_return)>(fallback_return);
+    if constexpr (std::is_invocable_v<T>)
+    {
+        return std::invoke(std::forward<decltype(fallback_return)>(fallback_return));
+    }
+    else
+    {
+        return std::forward<decltype(fallback_return)>(fallback_return);
+    }
 }
 
 template <typename Fun, typename... Args>
--- a/src/daemon/daemon_main.cpp
+++ b/src/daemon/daemon_main.cpp
@@ -25,6 +25,7 @@
 #include <multipass/constants.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform_unix.h>
+#include <multipass/signal.h>
 #include <multipass/top_catch_all.h>
 #include <multipass/utils.h>
 #include <multipass/version.h>
@@ -41,11 +42,13 @@ namespace mpp = multipass::platform;
 
 namespace
 {
+
 class UnixSignalHandler
 {
 public:
-    UnixSignalHandler()
-        : signal_handling_thread{
+    UnixSignalHandler(mp::Signal& app_ready_signal)
+        : app_ready_signal(app_ready_signal),
+          signal_handling_thread{
               [this, sigs = mpp::make_and_block_signals({SIGTERM, SIGINT, SIGUSR1})] { monitor_signals(sigs); }}
     {
     }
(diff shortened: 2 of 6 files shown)


Sploder12 (dismissed):
LGTM, I was unable to hang the daemon during start!

xmkg:
@andrei-toterman @ricab @georgeliao this one needs a secondary review to land. Could I interest any of you :)

src/daemon/daemon_main.cpp
```diff
@@ -92,14 +102,42 @@ int main_impl(int argc, char* argv[])
...
     mpl::log(mpl::Level::info, "daemon", fmt::format("Daemon arguments: {}", app.arguments().join(" ")));
-    auto ret = QCoreApplication::exec();
 
+    // Signal the signal handler that app has completed its basic initialization, and
+    // ready to process signals.
+    app_ready_signal.signal();
+    auto exit_code = QCoreApplication::exec();
+    // QConcurrent::run() invocations are dispatched through the global
+    // thread pool. Wait until all threads in the pool are properly cleaned up.
+    QThreadPool::globalInstance()->waitForDone();
```
georgeliao, line 112: Do we still need this while we have all the `QFuture::waitForFinished()` calls in `~daemon()`?

src/daemon/daemon_main.cpp
```diff
@@ -92,14 +102,42 @@ int main_impl(int argc, char* argv[])
...
-    return ret;
+    return exit_code;
 }
 } // namespace
 
 int main(int argc, char* argv[])
 {
-    return mp::top_catch_all("daemon", /* fallback_return = */ EXIT_FAILURE, main_impl, argc, argv);
+    mp::Signal app_ready_signal{};
+    //
```
georgeliao, line 121: Can we keep the constructions of `app_ready_signal` and `handler` in the `main_impl`? That makes the code more encapsulated.

include/multipass/top_catch_all.h
```diff
@@ -99,7 +99,14 @@ inline auto multipass::top_catch_all(std::string_view log_category,
         detail::error(log_category);
     }
 
-    return std::forward<decltype(fallback_return)>(fallback_return);
```
georgeliao, line 102: This change seems to be reasonable. I only have a general question about this function. Its return type is `std::invoke_result_t<Fun, Args...>` which basically is the return type of `Fun`. In the first branch below, do we assume the `fallback_return` has the same return type as `Fun`; as to the else branch, it does not seem to make sense because it returns `fallback_return` itself with preserving the value category.

src/daemon/daemon_main.cpp
xmkg: Unfortunately yes. This is for making sure that the global thread pool finishes all of the queued work before exiting. Otherwise, the thread pool might exit too early, and the main thread might wait on some task that's not going to be finished, ever.

We desperately need a centralized asynchronous work coordinator to do all the async work in a well-defined, safe way.

src/daemon/daemon_main.cpp
xmkg: That would also mean moving the exception handling (i.e., top_catch_all) to the main_impl too, since we need to signal the `app_ready_signal` if something throws in the `main_impl`. I could add a function-level try-catch to the main_impl if you prefer that.

src/daemon/daemon_main.cpp
georgeliao: What I have read is that there might be other threaded tasks besides that ones submitted by `QtConcurrent::run(...)`. Anyway, if removing this cause some weird behavior, we probably should keep it.

georgeliao (dismissed):

ricab:
Aren't there actual [compilation issues](https://github.com/canonical/multipass/actions/runs/14703253415) here? e.g.:

> ‘Signal’ is not a member of ‘mpt’

xmkg:
> Aren't there actual [compilation issues](https://github.com/canonical/multipass/actions/runs/14703253415) here? e.g.:
> 
> > ‘Signal’ is not a member of ‘mpt’

Yes, there is. The pipeline was green before, IDK what happened. I'll push another commit to fix it.

georgeliao:
@ricab nice catch, I browsed the error log several times and did not find anything. Besides, the failure occurred on the debug build only. Therefore, I thought it was the disk space thing that should have been fixed on main. Now it appeared to be something else.

xmkg:
> @ricab nice catch, I browsed the error log several times and did not find anything. Besides, the failure occurred on the debug build only. Therefore, I thought it was the disk space thing that should have been fixed on main. Now it appeared to be something else.

Apparently, "main" had a mpt::Signal reference recently introduced in another MP, and since this branch was not rebased to main, and there were no merge conflicts, the PR's pipeline was green, whereas the merge queue's is not.

georgeliao (approved):

Sploder12 (approved):

PR #4017, opened 2025-03-28
[3rd-party/CMakeLists] Define CMAKE_POLICY_VERSION_MINIMUM=3.5

xmkg:
With CMake 4.0, support for compatibility with versions older than 3.5 has been removed. One of our dependencies (yaml-cpp) defines its policy version as 3.4, so the build with CMake 4 fails for that reason.

There's an option to override that behavior, called CMAKE_POLICY_VERSION_MINIMUM. This flag suppresses the error and allows CMake to try building the project anyway. yaml-cpp builds without any issues when the option is in place. The upstream also updated the cmake_minimum_required() to `3.4...3.14` without making any code changes, so it's safe to assume that the project is compatible with the CMake 3.5 policies.

https://github.com/jbeder/yaml-cpp/commit/c2680200486572baf8221ba052ef50b58ecd816e

MULTI-1919

### DIFF
--- a/3rd-party/CMakeLists.txt
+++ b/3rd-party/CMakeLists.txt
@@ -4,6 +4,20 @@ if (MSVC)
   add_compile_options(-wd5045) #Disable warning about Spectre mitigation
 endif()
 
+function(add_subdirectory_compat)
+    #
+    # This is for CMake 4.0 compatibility.
+    #
+    # CMake 4.0 drops support CMake policies below 3.5 and some of our dependencies
+    # declare their minimum required version as 3.4, and CMake 4.0 is not happy
+    # with that. The CMAKE_POLICY_VERSION_MINIMUM option allows us to override that
+    # without having to touch the dependency's source code.
+    #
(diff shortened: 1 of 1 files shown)


xmkg:
@ricab, this one is blocking the CI. Could you review it when you have the time?

ricab (approved):
Thanks again @xmkg! I am happy with this if CI is too, so approving.

A question for future PRs though: why the short comment width? Lines that short push more code out of the screen. I would expect it to stick to the usual formatting length, except to avoid breaking things that belong together or other such overriding concerns.

xmkg:
> Thanks again @xmkg! I am happy with this if CI is too, so approving.
> 
> A question for future PRs though: why the short comment width? Lines that short push more code out of the screen. I would expect it to stick to the usual formatting length, except to avoid breaking things that belong together or other such overriding concerns.

No particular reason, really. I agree that it looks odd. Fixed now.

xmkg:
@ricab sorry to ping again, this one is waiting for re-approval to merge.

ricab (approved):
Oh :man_facepalming:

PR #4018, opened 2025-03-28
Remove libvirt driver from GUI

Sploder12:
This PR removes libvirt from the list of drivers in the GUI. If a user has libvirt they will still be on libvirt, however, once they switch to a different driver libvirt will no-longer be selectable.

MULTI-1889

### DIFF
--- a/src/client/gui/lib/providers.dart
+++ b/src/client/gui/lib/providers.dart
@@ -7,6 +7,7 @@ import 'package:collection/collection.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:fpdart/fpdart.dart';
 import 'package:grpc/grpc.dart';
+import 'package:multipass_gui/platform/platform.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 import 'ffi.dart';
@@ -318,4 +319,13 @@ final networksProvider = Provider.autoDispose((ref) {
     }).ignore();
   }
   return BuiltSet<String>();
+});
+
+final driversProvider = Provider.autoDispose((ref) {
+  final driver = ref.watch(daemonSettingProvider(driverKey)).valueOrNull;
+  if (driver != null && !mpPlatform.drivers.containsKey(driver)) {
+    return Map<String, String>.from(mpPlatform.drivers)
+      ..addAll({driver: driver});
+  }
+  return mpPlatform.drivers;
 });
--- a/src/client/gui/lib/providers.dart
+++ b/src/client/gui/lib/providers.dart
@@ -7,7 +7,6 @@ import 'package:collection/collection.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:fpdart/fpdart.dart';
 import 'package:grpc/grpc.dart';
-import 'package:multipass_gui/platform/platform.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 import 'ffi.dart';
@@ -319,13 +318,4 @@ final networksProvider = Provider.autoDispose((ref) {
     }).ignore();
   }
   return BuiltSet<String>();
-});
-
-final driversProvider = Provider.autoDispose((ref) {
-  final driver = ref.watch(daemonSettingProvider(driverKey)).valueOrNull;
-  if (driver != null && !mpPlatform.drivers.containsKey(driver)) {
-    return Map<String, String>.from(mpPlatform.drivers)
-      ..addAll({driver: driver});
-  }
-  return mpPlatform.drivers;
 });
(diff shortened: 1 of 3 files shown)


src/client/gui/lib/providers.dart
```diff
@@ -7,6 +7,7 @@ import 'package:collection/collection.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:fpdart/fpdart.dart';
 import 'package:grpc/grpc.dart';
+import 'package:multipass_gui/platform/platform.dart';
```
andrei-toterman, line 10: I recommend a relative import rather than an absolute one, and then it should be placed separately from those imports, in the same group as `import 'ffi.dart';` below.

```suggestion
import 'platform/platform.dart';
```

src/client/gui/lib/providers.dart
```diff
@@ -319,3 +320,12 @@ final networksProvider = Provider.autoDispose((ref) {
...
 });
+
+final driversProvider = Provider.autoDispose((ref) {
+  final driver = ref.watch(daemonSettingProvider(driverKey)).valueOrNull;
+  if (driver != null && !mpPlatform.drivers.containsKey(driver)) {
+    return Map<String, String>.from(mpPlatform.drivers)
+      ..addAll({driver: driver});
+  }
+  return mpPlatform.drivers;
+});
```
andrei-toterman, line 331: Usually I place providers that are needed only for one widget closer to that widget, so this one could be moved in the same file as where its used, if there won't be any need for it to be accessed in multiple places.
Also, the code could be refactored a bit, like so
```suggestion
final driversProvider = Provider.autoDispose((ref) {
  final driver = ref.watch(daemonSettingProvider(driverKey)).valueOrNull;
  return {
    if (driver != null) driver: driver,
    ...mpPlatform.drivers,
  }.build();
});
```
Let me explain. First, the `.build()` at the end creates a `BuiltMap` instead of a regular `Map`, and those have equality comparison based on their actual values, as opposed to just object identity in the case of regular `Map`s, and that equality is used by providers to determine if they need to change. So in theory, with a regular `Map`, if I were to return a new `Map`, even if it had the same values as the previously returned one, the provider would still rebuild, since it would be a new map. With `BuiltMap`, as long as the actual values are the same, the provider won't rebuild. Of course, for this small case I don't think it would make any practical sense, it's just theoretically more correct, so the `.build()` is not mandatory.

Now for the change in logic. `...` is the spread operator, and `{...map}` creates a new map, and it takes populates it with all the values from `map`. And order matters. For example
```js
final map = {'a': 1};
print({...map, 'a': 2});
```
prints `{'a': 2}`, but
```js
final map = {'a': 1};
print({'a': 2, ...map});
```
prints `{'a': 1}`, as values specified later overwrite values specified earlier.
So in this case, if `driver` is null, the returned map is just equivalent to `mpPlatform.drivers`, and if it's not null, the returned map will contain that `driver`, and if `mpPlatform` also has the driver, the value from `mpPlatform` will end up in the map.

Sploder12: Interesting, I'll change to the spread operator! Thanks for the explanation!

levkropp (dismissed):
Just tested with
- `sudo snap install multipass` 
- setting the driver to libvirt in the gui 
- confirming it can launch instances after installing libvirt and connecting multipass to libvirt with `snap connect multipass:libvirt`

Then I installed this PR with `sudo snap refresh multipass --channel edge/pr4018` and was able to select a different backend in the GUI and could not return to the libvirt backend as expected

src/client/gui/lib/settings/virtualization_settings.dart
```diff
@@ -28,7 +38,7 @@ class VirtualizationSettings extends ConsumerWidget {
         label: 'Driver',
         width: 260,
         value: driver,
-        items: mpPlatform.drivers,
+        items: drivers.asMap(),
```
andrei-toterman, line 41: Hmm, since the new provider only relies on the current driver value to create the map, and we already watch the current driver provider in this widget, maybe we don't need to define this new provider at all, and we can just compute the value inline

```suggestion
        items: {
          if (driver != null) driver: driver,
          ...mpPlatform.drivers,
        },
```

Sploder12: ah, good point

levkropp (commented):
So this change works for not showing `libvirt` as one of the drivers in the settings menu after switching away from it, but it does break being able to launch instances with libvirt after `snap refresh` installs this version. I'm not sure if this was happening when it was a provider as well
image

ricab:
Hey @Sploder12, @levkropp, libvirt is very restrictive by default and a locally built Multipass daemon usually fails to launch VMs with "Permission denied" errors (that's regardless of the new permissions). I am not aware of that affecting the snap, so there is probably something else going on in here, but I wanted to make sure you guys don't confuse the two problems. 

The permissions error we've always had on local builds can be overcome with something like:

1. `sudoedit /etc/libvirt/qemu.conf`
2. uncomment `user = "root"` line
3. close the editor
4. `systemctl restart libvirtd`

ricab:
Glad to be of service :slightly_smiling_face:

levkropp (approved):
Can confirm that this fixed it on my end as well. Thank you @ricab !

andrei-toterman (approved):
Yup, all good :smile:

PR #4019, opened 2025-03-28
Remove lxd driver from GUI

Sploder12:
This PR removes lxd from the list of drivers in the GUI. If a user has lxd they will still be on lxd, however, once they switch to a different driver lxd will no-longer be selectable.

MULTI-1894


xmkg (approved):

andrei-toterman (dismissed):
:+1:

xmkg (approved):

PR #4022, opened 2025-04-03
[vcpkg] Update deps for compatibility with cmake 4.0

sharder996:
- Enforced pcre2 >= 10.43 for [cmake minimum version](https://github.com/PCRE2Project/pcre2/blob/3864abdb713f78831dd12d898ab31bbb0fa630b6/CMakeLists.txt#L108) compatibility with cmake 4.0

### DIFF
--- a/3rd-party/vcpkg-ports/poco/vcpkg.json
+++ b/3rd-party/vcpkg-ports/poco/vcpkg.json
@@ -8,7 +8,10 @@
   "supports": "!uwp",
   "dependencies": [
     "expat",
-    "pcre2",
+    {
+      "name": "pcre2",
+      "version>=": "10.43"
+    },
     {
       "name": "vcpkg-cmake",
       "host": true


xmkg (approved):
LGTM, thanks Scott!

ricab (approved):
Yep, thanks! 

@georgeliao I would just like your confirmation on this too. If you're good with it you can go ahead and merge.

georgeliao:
@sharder996  

> Updated vcpkg to 2025.02.14
Does it have to 2025.02.14? can it be 2025.01.13 which is more aligned with the grpc version?

xmkg:
> Does it have to 2025.02.14? can it be 2025.01.13 which is more aligned with the grpc version?

Can you explain why gRPC requires that specific version? I'm asking because I have a PR that also needs to update the vcpkg version.

sharder996:
@georgeliao @xmkg grpc doesn't require that specific version of vcpkg. The version of grpc we are using is still available in newer versions of vcpkg.

However, updating vcpkg here isn't strictly necessary since pcre2 10.43 is available in the version of vcpkg we currently use. Must have been an artifact of my testing. I can drop the commit if it's an issue.

xmkg:
> The version of grpc we are using is still available in newer versions of vcpkg.

Yeah, that's why I'm curious about @georgeliao's comment, it struck me a little bit odd. If gRPC somehow needs a particular version of vcpkg, that's what "baseline" is for, and we can pin it without having to block other work.

If something is preventing us from bumping the vcpkg version, we should fix it. Otherwise, it would cause trouble while migrating the rest of the dependencies

georgeliao:
@xmkg 
> I see. Then, this is a perfect opportunity to use a specific baseline different than the vcpkg version. 

We talked about this before. Discrepancy of these two should not be used unless we have to.  

> The change is not necessary here, but it is in one of my PRs. 

I am kind of surprised it (changing from 2025.01.13 to 2025.02.14) is a hard requirement on the fmt build. If it really is, as I mentioned above, we can allow the vcpkg version and baseline hash. Anyway, the discussion of that should be moved to that PR, not this one.

ricab:
OK, I think this deserves some live discussion going forward, to align on what to do when. This PR I'd rather merge ASAP, so I agree it makes sense to separate matters.

xmkg:
> We talked about this before. Discrepancy of these two should not be used unless we have to.

On what exact basis? To be completely honest, I don't understand the reason why you want these two to be in line so much. "builtin-baseline" is not a workaround; it is a core feature of the vcpkg, which is designed to deal with situations exactly like this. Are you proposing that we should try to fix and move the gRPC patch up every time we stumble upon a dependency that needs a newer vcpkg version? That's not feasible, and it would block other work. The fmt patch I mentioned is waiting more than 1 month just because we waited for gRPC to land first. My honest opinion is that this approach is not going to work well in the long term.

georgeliao (approved):
@sharder996 Thanks for the good work.

PR #4029, opened 2025-04-07
[gui] add 'Launch & Configure next' button to launch form

### LINKED ISSUE #4028: Add "Launch and Configure next" button
Add multiple instances in a row 

Add a Launch and Configure next button in the New instance window

**Additional context**

Image

https://github.com/user-attachments/assets/144a0d81-f00b-48a0-916b-0ca24a220542

[Figma link]([Figma link](https://www.figma.com/proto/kY5ZFvQSyqIGXpIz7KNNju/AZs-first-iteration?node-id=122-3121&p=f&t=FoybV0fR7EvB7oL1-1&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=122%3A3121&show-proto-sidebar=1))

levkropp:
resolves #4028 

image

This pull request introduces a new feature to the `LaunchForm` in the `src/client/gui/lib/catalogue/launch_form.dart` file, allowing users to launch and configure the next item without closing the form. The most important changes include adding a new button, modifying the launch method to support the new functionality, and adjusting the form reset behavior.

New feature for launching and configuring the next item:

* Added a `launchAndConfigureNextButton` to the form, which allows users to launch the current item and immediately configure the next one. [[1]](diffhunk://#diff-1859f216fc901caa87909945aef29585a3d98e3b5e265a7e804870ae8a3a652fR253-R257) [[2]](diffhunk://#diff-1859f216fc901caa87909945aef29585a3d98e3b5e265a7e804870ae8a3a652fR293-R294)

Modifications to the launch method:

* Updated the `launch` method to accept an optional `configureNext` parameter, which controls whether the form should reset after launching.

Form reset behavior:

* Adjusted the form reset logic to only reset the form and close the drawer if `configureNext` is `false`. [[1]](diffhunk://#diff-1859f216fc901caa87909945aef29585a3d98e3b5e265a7e804870ae8a3a652fR326-R336) [[2]](diffhunk://#diff-1859f216fc901caa87909945aef29585a3d98e3b5e265a7e804870ae8a3a652fL349)

### DIFF
--- a/src/client/gui/lib/catalogue/launch_form.dart
+++ b/src/client/gui/lib/catalogue/launch_form.dart
@@ -251,9 +251,9 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       child: const Text('Launch'),
     );
 
-    final launchAndConfigureNextButton = TextButton(
+    final launchAndConfigureNextButton = OutlinedButton(
       onPressed: () => launch(imageInfo, configureNext: true),
-      child: const Text('Launch & Configure Next'),
+      child: const Text('Launch & Configure next'),
     );
 
     final cancelButton = OutlinedButton(
--- a/src/client/gui/lib/catalogue/launch_form.dart
+++ b/src/client/gui/lib/catalogue/launch_form.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:basics/basics.dart';
 import 'package:flutter/material.dart' hide Switch, ImageInfo;
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:protobuf/protobuf.dart';
 import 'package:rxdart/rxdart.dart';
 
 import '../ffi.dart';
--- a/src/client/gui/lib/catalogue/launch_form.dart
+++ b/src/client/gui/lib/catalogue/launch_form.dart
@@ -250,6 +250,11 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       child: const Text('Launch'),
     );
 
+    final launchAndConfigureNextButton = TextButton(
+      onPressed: () => launch(imageInfo, configureNext: true),
+      child: const Text('Launch & Configure Next'),
+    );
+
     final cancelButton = OutlinedButton(
       onPressed: () => Scaffold.of(context).closeEndDrawer(),
       child: const Text('Cancel'),
@@ -285,6 +290,8 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
             Row(children: [
               launchButton,
               const SizedBox(width: 16),
+              launchAndConfigureNextButton,
+              const SizedBox(width: 16),
               cancelButton,
             ]),
           ]),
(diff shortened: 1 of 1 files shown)


src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -317,7 +322,19 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
...
     initiateLaunchFlow(ref, launchRequest, mountRequests);
-    Scaffold.of(context).closeEndDrawer();
+
+    if (!configureNext) {
+      formState.reset();
+      mountFormState?.reset();
+      setState(() {
+        mountRequests.clear();
+        addingMount = false;
+      });
```
andrei-toterman, line 332: Do you really need to clear those? Won't they just go away with the `Scaffold.of(context).closeEndDrawer();`?

andrei-toterman (changes requested):
Nope, does not LGTM! I approved a bit too early apparently :sweat_smile: I found a bug.
Sometimes I forget that Dart is not like C++ and every object is passed by reference all the time. And since we're reusing the launch request and mount list objects, reassigning their values messes with places that already use them. So we need to copy them before passing them to the launching code. You'll need to replace this
```dart
initiateLaunchFlow(ref, launchRequest, mountRequests);
```
with this
```dart
initiateLaunchFlow(
  ref,
  launchRequest.deepCopy(),
  mountRequests.map((r) => r.deepCopy()).toList(),
);
```

andrei-toterman (approved):

src/client/gui/lib/catalogue/launch_form.dart
```diff
@@ -316,8 +321,18 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       mountRequest.targetPaths.first.instanceName = launchRequest.instanceName;
     }
 
-    initiateLaunchFlow(ref, launchRequest, mountRequests);
-    Scaffold.of(context).closeEndDrawer();
+    initiateLaunchFlow(
+      ref,
+      launchRequest.deepCopy(),
```
Sploder12, line 326: Seems this deepCopy is causing issues. https://pub.dev/documentation/protobuf/latest/protobuf/GeneratedMessageGenericExtensions.html might help.

Sploder12 (changes requested):
Code looks good but it doesn't build for me or CI

src/client/gui/lib/catalogue/launch_form.dart
levkropp: We need to use clone() to do deep copies
https://protobuf.dev/reference/dart/dart-generated/ (Ctrl+F for clone() )
Making this change now

Sploder12: It seems `clone` is deprecated, I'm not sure we want to be using it in new code.

levkropp: Right, we needed to import the protobuf extensions for this because it's an extension while clone() works because clone was part of the base protobuf.

https://github.com/google/protobuf.dart/blob/7ac17d0f0805a04f18ac54210d2152c61a7e5098/protobuf/lib/src/protobuf/generated_message.dart#L49
  /// Creates a deep copy of the fields in this message.
  /// (The generated code uses [mergeFromMessage].)
  @Deprecated('Using this can add significant size overhead to your binary. '
      'Use [GeneratedMessageGenericExtensions.deepCopy] instead. '
      'Will be removed in next major version')
  GeneratedMessage clone();

Sploder12 (approved):
LGTM!

PR #4030, opened 2025-04-07
[gh/wf/linux.yml] allow workflow cancellation

xmkg:
Allow in-progress workflows to be canceled when an external event requests them (e.g., the workflow run has been superseded with a new commit).

MULTI-1939

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,8 +23,8 @@ on:
     types: [checks_requested]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true #
 
 jobs:
   Lint:
@@ -64,6 +64,9 @@ jobs:
         echo "${MATRIX}" | jq -cs '{"include": . }' | awk '{ print "matrix=" $0 }' >> $GITHUB_OUTPUT
 
   BuildAndTest:
+    concurrency:
+      group: ${{ github.workflow }}-buildtest-${{ github.head_ref || github.run_id }}-${{ matrix.build-type }}
+      cancel-in-progress: true
     needs: [Lint, GetMatrix]
 
     # Allow skipped Lint
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,10 +68,16 @@ jobs:
 
     # Allow skipped Lint
     # Let Lint fail on pull requests
-    if: ${{ !cancelled() && (success() ||
-           needs.Lint.result == 'skipped' ||
-           (needs.Lint.result == 'failure' && github.event_name == 'pull_request')
-          )}}
+    if: ${{ !cancelled() &&
+            (
+              success() ||
+              needs.Lint.result == 'skipped' ||
+              (
+                needs.Lint.result == 'failure' &&
+                github.event_name == 'pull_request'
+              )
+            )
+        }}
 
     outputs:
       label: ${{ steps.build-params.outputs.label }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,7 +24,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'merge_group' }}
+  cancel-in-progress: true
 
 jobs:
   Lint:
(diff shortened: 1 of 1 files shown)


ricab:
Good, this addresses #4048 now too.

xmkg:
> Good, this addresses #4048 now too.

Github's workflow cancellation is a real headache. The matrix part of the job seems resistant to every kind of cancellation request I've tried, except for force cancel. I'm still investigating why that is. I'll add some trace steps to jobs to see whether the cancellation signals (SIGINT, SIGTERM, etc.) are reaching to the steps at all.

xmkg:
> > Good, this addresses #4048 now too.
> 
> Github's workflow cancellation is a real headache. The matrix part of the job seems resistant to every kind of cancellation request I've tried, except for force cancel. I'm still investigating why that is. I'll add some trace steps to jobs to see whether the cancellation signals (SIGINT, SIGTERM, etc.) are reaching to the steps at all.

In case anyone needs I'm using the following bash functions to kill workflows forcefully:

```shell
github-kill-workflow(){
gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/$1/actions/runs/$2/force-cancel
}


github-kill-workflow-by-url() {
url="$1"

repo=$(echo "$url" | awk -F'[/?]' '{print $4 "/" $5}')
run_id=$(echo "$url" | awk -F'[/?]' '{print $8}')

echo "repo: $repo"
echo "run_id: $run_id"

github-kill-workflow $repo $run_id

}
```

Include them in your .bashrc, then use it like 

```bash
github-kill-workflow-by-url https://github.com/canonical/multipass/actions/runs/14502559305?pr=4030
```

xmkg:
> The matrix part of the job seems resistant to every kind of cancellation request I've tried, except for force cancel

Finally figured out why that is.

```yaml
    if: |
      ${{
        success()
        || needs.Lint.result == 'skipped'
        || ( needs.Lint.result == 'failure'
             && github.event_name == 'pull_request' )
      }}
```

The BuildAndTest is cancellable after commenting out the precondition.

> To cancel the workflow run, the server re-evaluates if conditions for all currently running jobs. If the condition evaluates to true, the job will not get canceled.

So, this expression somehow evaluates to true on re-check, even with `!cancelled() && (...)` addition. That's enough head-banging for today: I'll wrap this up tomorrow.

xmkg:
@ricab finally, everything works as it should. A new push cancels the previous runs, and the "cancel workflow" button works.

.github/workflows/linux.yml
```diff
@@ -23,14 +23,14 @@ on:
     types: [checks_requested]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'merge_group' }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
ricab, line 26: Just curious, did this affect anything?

.github/workflows/linux.yml
```diff
@@ -68,13 +68,10 @@ jobs:
...
-      ${{
-        success()
-        || needs.Lint.result == 'skipped'
-        || ( needs.Lint.result == 'failure'
-             && github.event_name == 'pull_request' )
-      }}
+    if: ${{ !cancelled() && (success() ||
+           needs.Lint.result == 'skipped' ||
+           (needs.Lint.result == 'failure' && github.event_name == 'pull_request')
+          )}}
```
ricab, line 74: Tiny nitpick, but this indentation looks unaligned.

ricab (dismissed):
Good job Mustafa, thanks! 2 tiny things inline, but pre-approving.

.github/workflows/linux.yml
xmkg: `cancel-in-progress` part? Yes. 

`group`? Not necessarily. I updated it just to be aligned with the [official docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-a-fallback-value). I can revert it if you would prefer that.

.github/workflows/linux.yml
xmkg: Done.

.github/workflows/linux.yml
ricab: Yeah I meant `group`. Makes sense.

.github/workflows/linux.yml
```diff
@@ -68,10 +68,16 @@ jobs:
...
+    if: ${{ !cancelled() &&
+            (
+              success() ||
+              needs.Lint.result == 'skipped' ||
+              (
+                needs.Lint.result == 'failure' &&
+                github.event_name == 'pull_request'
+              )
+            )
+        }}
```
ricab, line 80: That's a lot of line splitting now, but ok...

ricab (approved):
Alright, LGTM, thanks again @xmkg!

ricab:
I would just like to get @sharder996's review on this one if possible.

sharder996 (approved):
Yup, I think it looks good. Biggest thing is that Github will group CI runs by head-ref now.

PR #3908, opened 2025-01-30
Lexically normal mount paths (fix #3733)

ntorresalberto:
### Fixes for #3733 and Additional Issues Found During Inspection

This PR addresses the problems outlined in #3733, along with some related issues discovered during review. Details below.

#### Relative Paths Incorrectly Resolving to `/home/ubuntu`

The issue occurs when relative paths unintentionally resolve to `/home/ubuntu`, allowing unexpected mounts. Here’s how to reproduce it (same steps as in the issue, repeated for clarity):

```bash
$ mkdir foodir
$ multipass delete -p foo
$ multipass launch --name foo

# Expected to fail, as it does:
$ multipass mount foodir foo:/home/ubuntu  
mount failed: The following errors occurred:
unable to mount to "/home/ubuntu"

# Expected not to work, but it does (matches /home/ubuntu):
$ multipass mount foodir foo:.              
$ multipass mount foodir foo:../ubuntu      
```

#### Relative Paths Allowing Mounts to Restricted Directories

Similar path resolution issues allowed mounting to restricted locations such as `/dev/`, `/`:

```bash
# Expected not to work, but it does (matches /dev):
$ multipass mount foodir foo:../../dev  

# Expected not to work, but it does (matches /):
$ multipass mount foodir foo:../..          
```

#### Issues with Non-Normalized Paths Causing Duplicate Entries

Paths should be normalized before being stored in [vm_mounts](https://github.com/canonical/multipass/blob/724933cbe35e7ec622da63ca4ea19ede668c82c1/src/daemon/daemon.cpp#L1958) to prevent different representations of the same path from being treated as separate entries. Without normalization, paths like `/home/foo` and `../foo` are considered distinct, leading to inconsistencies.

```bash
# Expected to work, and it does:
$ multipass mount foodir foo:/home/foo                     

# Expected NOT to work, and correctly fails (absolute paths are handled properly):
$ multipass mount foodir foo:/home/../home/foo  
mount failed: The following errors occurred:
"/home/foo" is already mounted in 'foo'

# Expected not to work, but it does:
$ multipass mount foodir foo:../foo                                
```

---

#### Fix Implementation

A new function, make_abspath, has been added to src/utils/utils.cpp.

- Converts relative paths to absolute paths, defaulting from /home/ubuntu.
- Guarantees unique, normalized paths for each mount, preventing duplicate entries.

#### Validation & Tests After Fixes

```bash
$ multipass delete -p foo
$ multipass launch --name foo
[...]

PR #4031, opened 2025-04-07
Normalize mount targets

### LINKED ISSUE #3733: No warning displayed when mounting path over instance's home using "foo:." notation
**Describe the bug**
If I try to mount a path to an instance's home folder (e.g. `multipass mount /some/path foo:/home/ubuntu` a warning message is displayed and the action does not go through. But if I use a different syntax, e.g. `multipass mount /some/path foo:.` the system does not display a warning and allows me to complete the action.

**To Reproduce**
1. Try `multipass mount /some/path foo:/home/ubuntu` , see the warning message displayed.
1. Now try `multipass mount /some/path foo:.`

**Expected behavior**
I'd expect a consistent behavior for any actions that will overlay the user's home folder.

ricab:
Fixes #3733. 

Co-authored by @ntorresalberto (thanks!)

This takes #3908 and applies some changes on top. The main thing was to update path normalization to rely on Qt for mounts. This resolves a few problems:

1. Target paths were not correctly processed on Windows: the previous implementation used `std::filesystem` which would treat target paths as Windows paths on that platform, whereas they are meant for Linux VMs. In particular, that meant that a) paths were not correctly identified as relative or absolute; b) they were not normalized correctly. Qt, on the other hand, seems to understand Linux paths even on Windows. See https://gist.github.com/ricab/312893d71ad09c6271107eb902f48fdd and
https://godbolt.org/z/j5f7657v4.

2. Target paths weren't fully normalized. In particular, trailing slashes were not removed or added. The underlying call to `std::filesystem::path::lexically_normal` kept `/foo/bar/` and `/foo/bar` unchanged ¯\_(ツ)_/¯

It also streamlines the code around this a little and extracts a constant for `/home/ubuntu`. We should actually have one for the `ubuntu` user instead, but I didn't feel like replacing that all over in here.

See commit messages for further details.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -233,11 +233,12 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                                      "You can also use a shortcut of \"<name>\" to mean \"name=<name>\".",
                                      "spec");
     QCommandLineOption bridgedOption("bridged", "Adds one `--network bridged` network.");
-    QCommandLineOption mountOption("mount",
-                                   "Mount a local directory inside the instance. If <target> is omitted, the "
-                                   "mount point will be under /home/ubuntu/<source-dir>, where <source-dir> is "
-                                   "the name of the <source> directory.",
-                                   "source>:<target");
+    QCommandLineOption mountOption(
+        "mount",
+        QStringLiteral("Mount a local directory inside the instance. If <target> is omitted, the mount point will be "
+                       "under %1/<source-dir>, where <source-dir> is the name of the <source> directory.")
+            .arg(home_in_instance),
+        "source>:<target");
 
     parser->addOptions({cpusOption, diskOption, memOption, memOptionDeprecated, nameOption, cloudInitOption,
                         networkOption, bridgedOption, mountOption});
--- a/src/client/cli/cmd/mount.cpp
+++ b/src/client/cli/cmd/mount.cpp
@@ -110,12 +110,13 @@ QString cmd::Mount::description() const
 mp::ParseCode cmd::Mount::parse_args(mp::ArgParser* parser)
 {
     parser->addPositionalArgument("source", "Path of the local directory to mount", "<source>");
-    parser->addPositionalArgument("target",
-                                  "Target mount points, in <name>[:<path>] format, where <name> "
-                                  "is an instance name, and optional <path> is the mount point. "
-                                  "If omitted, the mount point will be under /home/ubuntu/<source-dir>, "
-                                  "where <source-dir> is the name of the <source> directory.",
-                                  "<target> [<target> ...]");
+    parser->addPositionalArgument(
+        "target",
+        QStringLiteral("Target mount points, in <name>[:<path>] format, where <name> is an instance name, and optional "
+                       "<path> is the mount point. If omitted, the mount point will be under %1/<source-dir>, where "
+                       "<source-dir> is the name of the <source> directory.")
+            .arg(home_in_instance),
+        "<target> [<target> ...]");
 
     QCommandLineOption gid_mappings({"g", "gid-map"},
                                     "A mapping of group IDs for use in the mount. File and folder ownership will be "
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1928,7 +1928,7 @@ try // clang-format on
         const auto q_target_path = path_entry.target_path().empty()
                                        ? MP_UTILS.default_mount_target(QString::fromStdString(request->source_path()))
                                        : QDir::cleanPath(QString::fromStdString(path_entry.target_path()));
-        const auto target_path = q_target_path.toStdString();
+        const auto target_path = mp::utils::make_abspath(q_target_path);
 
         auto it = operative_instances.find(name);
         if (it == operative_instances.end())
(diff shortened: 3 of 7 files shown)


ricab:
In all honesty, I tested on Linux but I haven't tested this on Windows yet. @Sploder12, would you be able to give me a hand with that part?

ricab:
Hey @ntorresalberto, I think the underpinnings of the CLA check were going through rework around the time you made your contribution and it looks like they lost your record. Would you be able to sign again? Many thanks in advance!

ntorresalberto:
@ricab I just signed it again and got the email, so I think it’s done this time.

tests/test_utils.cpp
Sploder12, line 1: Should we add some tests for `normalize_mount_target`?

ricab:
Yep, it's working now! Thanks again.

Sploder12 (commented):
Code looks good overall and testing on Windows went well! Just a question w.r.t. testing `normalize_mount_target`, coverage hits it but I don't think it's explicitly tested.

tests/test_utils.cpp
ricab: Yeah I should... I'll have to queue that for now, but will try to circle back later.

ricab: Done!

src/client/cli/cmd/launch.cpp
```diff
@@ -251,9 +251,11 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
     QCommandLineOption mountOption(
         "mount",
-        "Mount a local directory inside the instance. If <target> is omitted, the "
-        "mount point will be under /home/ubuntu/<source-dir>, where <source-dir> is "
-        "the name of the <source> directory.",
+        QStringLiteral(
+            "Mount a local directory inside the instance. If <target> is omitted, the mount point "
+            "will be "
+            "under %1/<source-dir>, where <source-dir> is the name of the <source> directory.")
+            .arg(home_in_instance),
```
Sploder12: formatting is a bit odd here

src/client/cli/cmd/mount.cpp
```diff
@@ -119,10 +119,12 @@ mp::ParseCode cmd::Mount::parse_args(mp::ArgParser* parser)
...
-        "Target mount points, in <name>[:<path>] format, where <name> "
-        "is an instance name, and optional <path> is the mount point. "
-        "If omitted, the mount point will be under /home/ubuntu/<source-dir>, "
-        "where <source-dir> is the name of the <source> directory.",
+        QStringLiteral("Target mount points, in <name>[:<path>] format, where <name> is an "
+                       "instance name, and optional "
+                       "<path> is the mount point. If omitted, the mount point will be under "
+                       "%1/<source-dir>, where "
+                       "<source-dir> is the name of the <source> directory.")
+            .arg(home_in_instance),
```
Sploder12, line 126: idem

tests/test_utils.cpp
```diff
@@ -671,3 +641,34 @@ TEST(Utils, checkFilesystemBytesAvailableReturnsNonNegative)
...
+           std::make_tuple("../..//folder", "/folder"),
+           std::make_tuple("/usr/local/bin//.././/.//bin/././", "/usr/local/bin"),
+           std::make_tuple("folder//subfolder", "/home/ubuntu/folder/subfolder"),
+           std::make_tuple("./Documents/../Downloads", "/home/ubuntu/Downloads"),
+           std::make_tuple("", "/home/ubuntu"),
+           std::make_tuple(".", "/home/ubuntu"),
+           std::make_tuple("..", "/home"),
+           std::make_tuple("folder/./subfolder", "/home/ubuntu/folder/subfolder"),
+           std::make_tuple("folder/../other-folder", "/home/ubuntu/other-folder"),
+           std::make_tuple("//", "/")));
```
Sploder12, line 674: Great use of parameterized tests!

Sploder12 (dismissed):
LGTM! Some string formatting is bit strange but I won't block on it

src/client/cli/cmd/launch.cpp
ricab: You're right, I'll fix that!

src/client/cli/cmd/mount.cpp
ricab: Yeah, I'll fix it.

ricab:
Done @Sploder12. I did a drive-by fix of a similar poor format.

Sploder12 (approved):

src/daemon/daemon.cpp
```diff
@@ -2168,8 +2169,11 @@ try
         }
         auto& vm = it->second;
 
-        if (mp::utils::invalid_target_path(q_target_path))
+        if (MP_UTILS.invalid_target_path(q_target_path))
         {
+            mpl::log(mpl::Level::error,
```
sharder996: Does this need to be an error? It makes the message in the CLI a little pedantic:
```
% multipass mount . foo:/home/ubuntu
[2025-09-25T19:26:02.748] [error] [daemon] Invalid target path: /home/ubuntu
unable to mount to "/home/ubuntu"
```

ricab: Good point. I'll decrease to warning.

ricab: The more I think about this, the more I get convinced that, by automatically sending logs to users, we're conflating purposes/audiences...

ricab: Done.

ricab:
OK, I believe this is ready for another look.

sharder996 (approved):
Thanks to everyone involved!

Sploder12 (approved):
LGTM!

PR #4034, opened 2025-04-08
Add cloud init yaml files into multipass repo

georgeliao:


Sploder12:
So if I'm understanding correctly we plan to no longer use the `multipass-blueprints` repo and have everything in the `multipass` repo?

georgeliao:
> So if I'm understanding correctly we plan to no longer use the multipass-blueprints repo and have everything in the multipass repo?

A bit of context, we did have two options
1. Keeping cloud-init files in a separate repo, that would be similar to the multipass blueprint repo setup. Just that the content of the repo will be converted to cloud-init files. 
2. Move the cloud-init files into multipass repo. 
 
Because the url of the file is the only thing we care about (see details in the documentation mentioned above), it does not matter which one we choose. So @ricab and I eventually went with the option 2.

sharder996 (approved):
LGTM

PR #4035, opened 2025-04-09
[cmake/configure-linker] fix ldd typo

xmkg:
fat-fingered `ldd` instead of `lld`, and it worked because ldd is a valid linker, too :)


ricab (approved):
Oh, good catch.

PR #4045, opened 2025-04-15
[cloud-init/netplan] explicitly name interfaces with `set-name`

xmkg:
until now, the interface names were assigned automatically. the names were given in a "first-come-first-served" fashion, i.e. whichever interface initializes first gets the "eth0", regardless of the declaration order. this makes the interface names inconsistent, which makes it harder to identify the "extra" interfaces.

this patch explicitly names the default interface as "primary" and the extra ones as "extra<N>" where the <N> is the zero-based index for the extra interface.

also tidied up the yaml_node_utils a bit.


xmkg:
Example "ip a" output from a VM launched with: `multipass launch --network enp6s0`:

```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: primary: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 52:54:00:b2:3a:83 brd ff:ff:ff:ff:ff:ff
    inet 10.211.127.96/24 metric 100 brd 10.211.127.255 scope global dynamic primary
       valid_lft 3555sec preferred_lft 3555sec
    inet6 fd42:7573:ad1b:2da4:5054:ff:feb2:3a83/64 scope global mngtmpaddr noprefixroute 
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:feb2:3a83/64 scope link 
       valid_lft forever preferred_lft forever
3: extra0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 52:54:00:a9:af:9c brd ff:ff:ff:ff:ff:ff
    inet 192.168.35.216/24 metric 200 brd 192.168.35.255 scope global dynamic extra0
       valid_lft 172756sec preferred_lft 172756sec
    inet6 fe80::5054:ff:fea9:af9c/64 scope link 
       valid_lft forever preferred_lft forever
```

Emitted netplan file:

```yaml
network:
  version: 2
  ethernets:
    extra0:
      match:
        macaddress: "52:54:00:a9:af:9c"
      optional: true
      dhcp-identifier: "mac"
      dhcp4: true
      dhcp4-overrides:
        route-metric: 200
      set-name: "extra0"
    primary:
      match:
        macaddress: "52:54:00:b2:3a:83"
      dhcp-identifier: "mac"
      dhcp4: true
      set-name: "primary"
```

src/utils/yaml_node_utils.cpp
```diff
@@ -147,30 +179,27 @@ YAML::Node mp::utils::add_extra_interface_to_network_config(const std::string& d
...
-                create_extra_interface_node(extra_interface_name, extra_interface.mac_address);
-
+            network_data["ethernets"][extra.name] = extra;
             return network_data;
         }
-
-        ++i;
     }
+
+    throw multipass::UnreachableCodeException();
```
ricab, line 204: Couldn't we just throw `std::logic_error`?

ricab: I don't know, my take is that is what `throw logic_error{}` is good for: a built-in way to signal a line of code that should not be reached. It's pretty unique in our code base too (only 1 occurrence today IIANM).

I would be happy with `std::unreachable` or even just an `assert`; not for the optimization, but for what it signals to the reader. The throw is reasonable too if we're a little less sure the logic will always hold and want to be alerted by users.

xmkg: I agree that an unreachable part of the code fits into the definition of a "logic_error". However, logic_error is a bit generic and intended to be used as a base type for others. I recall seeing a couple of places where I thought "unreachable would fit here", so there might be instances where it's not handled properly w.r.t. defensive programming principles.

I’m hesitant to use `std::unreachable()` as defined by the standard library because, by definition, it invokes undefined behavior. The `std::unreachable()` is like saying:

> This part of the code is not reachable, so don't bother structuring the intermediate/compiled code as if it's reachable.

I argue that definition is orthogonal to what we need here, which is:

> The code execution should not reach this point. If it does, the program is in an illegal state, and we should terminate the execution.

The "assert()" would fit here, but it's useless in release builds. Contracts might've helped here, but we're *faaar* away from it as it's planned for C++26.

IMO, the soft approach would be to throw an exception and let the caller handle it. But if we're going full pedantic here,  the correct thing to do would be to abort the program altogether. If the unreachable part is somehow reached, we can no longer make assumptions about the program state.

To conclude, I'll update the code to use `std::logic_error{}` for now. If we find more occurrences of this, we might roll a dedicated exception type then.

xmkg: A good blog post about the potential side-effects of `std::unreachable()`: https://izaron.github.io/posts/std-unreachable/

ricab: Hi Mustafa. I tend to agree with your points, but perhaps with a little nuance regarding defensive programming. Until we have contracts, deciding to add an assert or a throw I guess depends on likelihood and danger (and to a lesser extent, performance). 

It seems to me that we should limit how much we code for unverified preconditions and invariants. We exercise debug builds a lot so, when we do code for those cases, an assert that gets compiled out is often acceptable in practice. In cases that are more dangerous or error prone, an exception or even terminate can be a better option.

However, I am not convinced that we should ever add another "dedicated exception type". It seems to me that `std::logic_error` fits the bill perfectly, for such a generic thing. However, I would be OK with dedicated types for specific unverified conditions that caused "unreachable" code to be reached (like an unexpected null ptr or something).

xmkg: Done!

georgeliao:
The idea sounds good, will do a deep review soon.

georgeliao (commented):
Hi @xmkg 
Thanks for the good work. I particularly like the `set-name` field which is virtual machine OS visible name and the struct to YAML node encoding. Besides that, I only have some minor comments and questions.

src/utils/yaml_node_utils.cpp
```diff
@@ -23,35 +23,62 @@
...
 {
-    YAML::Node default_interface_data{};
-
-    default_interface_data["match"]["macaddress"] = default_interface_mac_address;
-    default_interface_data["dhcp4"] = true;
-    default_interface_data["dhcp-identifier"] = "mac";
+template <>
+struct convert<interface_details>
+{
+    static Node encode(const interface_details& iface)
```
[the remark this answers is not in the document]
xmkg: That's a good idea.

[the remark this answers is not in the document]
georgeliao: Indeed, we do not have a good way to cover that in test. I think we can leave it as it is. Thanks for looking into that.

src/utils/yaml_node_utils.cpp
```diff
@@ -147,30 +178,27 @@ YAML::Node mp::utils::add_extra_interface_to_network_config(const std::string& d
         YAML::Node network_data{};
         network_data["version"] = "2";
 
-        network_data["ethernets"]["default"] = create_default_interface_node(default_mac_addr);
+        const auto default_ = make_default_interface(default_mac_addr);
```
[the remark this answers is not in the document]
xmkg: Because `default` is a keyword. I can rename the variable to sth like `def` if you would prefer that.

src/utils/yaml_node_utils.cpp
```diff
@@ -147,30 +178,27 @@ YAML::Node mp::utils::add_extra_interface_to_network_config(const std::string& d
...
     YAML::Node network_data = YAML::Load(network_config_file_content);
 
-    int i = 0;
-    while (true)
+    // Iterate over possible extra interface names and find a vacant one.
+    for (int current_index = 0;; current_index++)
     {
-        const std::string extra_interface_name = "extra" + std::to_string(i);
-        if (!network_data["ethernets"][extra_interface_name].IsDefined())
+        const auto extra = make_extra_interface(extra_interface.mac_address, current_index);
```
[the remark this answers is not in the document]
xmkg: Replacing index with name would mean we have to repeat the `fmt::format(extra_name_pattern, index)` call for each `make_extra_interface` -- in turn making the code more error-prone. Since this is rather a cold path in code, I'd be inclined to keep it as-is. WDYT?

src/utils/yaml_node_utils.cpp
```diff
@@ -23,35 +23,62 @@
...
+
+static auto make_extra_interface(const std::string& mac_addr, std::uint32_t extra_interface_idx = 0)
+{
+    interface_details iface{};
+    iface.name = fmt::format(kExtraInterfaceNamePattern, extra_interface_idx);
+    iface.mac_addr = mac_addr;
+    iface.optional = true;
+    iface.route_metric = 200;
+    return iface;
+}
```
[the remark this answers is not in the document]
xmkg: Done

src/utils/yaml_node_utils.cpp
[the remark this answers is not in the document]
xmkg: Done

tests/test_cloud_init_iso.cpp
```diff
@@ -50,6 +51,7 @@ version: 2
     dhcp4-overrides:
       route-metric: 200
     optional: true
+    set-name: {3}
```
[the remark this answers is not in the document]
xmkg: Done

tests/test_cloud_init_iso.cpp
```diff
@@ -37,11 +37,12 @@ const std::string default_meta_data_content = fmt::format(meta_data_content_temp
 constexpr auto* network_config_data_content_template = R"(#cloud-config
 version: 2
 ethernets:
-  default:
+  primary:
     match:
       macaddress: "{0}"
     dhcp4: true
     dhcp-identifier: mac
+    set-name: {2}
```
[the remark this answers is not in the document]
xmkg: Done

georgeliao (approved):

georgeliao (approved):

xmkg:
@ricab, we discussed this PR out-of-band a while ago, and agreed on unifying the names like eth0, eth1, etc. I'm revisiting this PR, and I'm a bit concerned about the fact that "eth0" does not tell anything about the importance of the interface. We rely on the main/primary/default (or whatever you would call it) interface for Multipass's basic operation, and I think the name should indicate that. What do you think?

ricab:
Hey @xmkg,

> We rely on the main/primary/default (or whatever you would call it) interface for Multipass's basic operation

We won't for long though, right? With the vsock/agent we could give people a choice not to have it at all. Some people have requested that for use with Kubernetes and things that assume the first interface is the one they want. An in general, the interface that gets selected for something varies with different factors (e.g. route weights, DNS), so main/primary/default could end up being misleading.

I think this PR is valuable as a means of obtaining *predictable* interface names that are *consistent* across hypervisors and distros. Other than that, there is a tradeoff between *expressive* and *flexible*. I would be inclined to something generic enough that we can adapt as necessary. And also something habitual: non-generic scripts/globs/regexes might not account for arbitrary names (e.g. `e*`). Something like "ethN" or "enoN" would give us all that while keeping order and, therefore, and implicit sense of which interface is the "first".

ricab:
All that said, I don't feel super strong about this. Perhaps there are other aspects that I could be missing?

xmkg:
Rebased.

xmkg:
@ricab, this one's ready for another round.

src/utils/yaml_node_utils.cpp
```diff
@@ -25,34 +25,68 @@ namespace mp = multipass;
...
+    {
+        constexpr std::string_view kInterfaceNamePattern = "eth{}";
+        this->name = fmt::format(kInterfaceNamePattern, index);
+        this->mac_addr = mac_addr;
+        if (optional)
+        {
+            this->optional = true;
+            this->route_metric = 200;
+        }
+    }
```
Sploder12: couldn't most of this be in the member initializer list?

xmkg: yep, just wanted to be explicit until we have the designated initializers.

src/utils/yaml_node_utils.cpp
```diff
@@ -121,18 +155,25 @@ YAML::Node mp::utils::make_cloud_init_network_config(
...
-        if (extra_interfaces[i].auto_mode)
-        {
-            const std::string name = "extra" + std::to_string(i);
-            network_data["ethernets"][name] =
-                create_extra_interface_node(name, extra_interfaces[i].mac_address);
-        }
+        if (!extra.auto_mode)
+            continue;
+
+        const auto extra_interface = interface_details{/*mac_addr=*/extra.mac_address,
```
ricab, line 213: Might want to add the C++20 TODO here as well.

src/utils/yaml_node_utils.cpp
```diff
@@ -121,18 +155,25 @@ YAML::Node mp::utils::make_cloud_init_network_config(
...
 
-    for (size_t i = 0; i < extra_interfaces.size(); ++i)
+    const auto default_interface = interface_details{default_mac_addr};
+    network_data["ethernets"][default_interface.name] = default_interface;
+
+    auto extra_idx = kExtraInterfaceIndexStart;
+    // FIXME: C++20 range-for init statement:
+    // https://en.cppreference.com/w/cpp/language/range-for.html
+    // C++23: https://en.cppreference.com/w/cpp/ranges/enumerate_view.html
+    for (const auto& extra : extra_interfaces)
```
ricab, line 208: Until we can enumerate, I think the traditional for loop remains more adequate for this. Anyway, I am not going to split hairs on this one now.

ricab (approved):
Looks good Mustafa, thanks. A couple of small comments inline, but I am happy to approve.

src/utils/yaml_node_utils.cpp
xmkg: Ah, never mind -- too much coffee. I'll move them to the initializer list.

src/utils/yaml_node_utils.cpp
xmkg: Will do.

src/utils/yaml_node_utils.cpp
xmkg: Done

ricab (approved):
:+1:

Sploder12 (approved):
LGTM!

Issue #4042, opened 2025-04-14
Protech main branch


ricab:
**What are you trying to do?**
Require docs and/or build workflows before branches can merge, as applicable.

Until recently, Linux CI would run on all PRs and the `main` branch was protected to require them to succeed. That is no longer the case (some changes don't affect code). But we should still prevent merges until all relevant jobs run and checks pass.

**What's your proposed solution?**
We could add and gate on a separate workflow to decide what jobs need to run and succeed, depending on modification.

ricab:
Done now!

PR #4047, opened 2025-04-15
Revert "[ci] Skip Linux CI on unrelated changes"

ricab:
Reverts canonical/multipass#3993


xmkg (approved):
LGTM. Do we know why the change causes a problem?

ricab:
Yes, we have branch protections requiring those builds to succeed before we can merge, and apparently that applies to builds triggered by `pull_request` and not just `merge_group`. Furthermore, I think the filters are somehow applied to `merge_group` triggers as well. 

We'll need https://github.com/canonical/multipass/issues/4042 instead, but I'd like the CI merger to land first.

sharder996 (approved):
Adding my review as well! LGTM

PR #4050, opened 2025-04-16
Enabling qemu driver on linux arm cpu 

georgeliao:
https://warthogs.atlassian.net/browse/MULTI-1933

In the development environment, `qemu-system-arm` and `qemu-efi-aarch64` need to be installed and `sudo cp /usr/share/qemu-efi-aarch64/QEMU_EFI.fd /usr/share/qemu/QEMU_EFI.fd` need to be run in order to make `qemu-system-arm -bios QEMU_EFI.fd ...` find the `QEMU_EFI.fd` file and access it with the right permission. On the other side, the snapcraft [copies](https://github.com/canonical/multipass/blob/3d329cb5a94056953785580149c5536baeadc1b0/snap/snapcraft.yaml#L338) the file into a snap confined folder, so there above-mentioned issues do not exist in deployment. 

The snap file can be downloaded [here](https://www.dropbox.com/scl/fi/z6rgdmnqi0ksu4odormao/multipass_1.16.0-dev.3744-gaad30415_arm64.snap?rlkey=w0d88ywq1ujbv7c3xri69szss&st=vio2wmpw&dl=0). 

Regarding to how to get the an arm + ubuntu machine to test, [testflinger](https://certification.canonical.com/search/v2/physicalmachines/?term=in_testflinger%3Atrue+arch%3Aarm64) machine is a good option. You will need to install testfinger cmd via snap and [reserve](https://certification.canonical.com/docs/ops/tel-labs-docs/how-to/reserve_system_on_testflinger/) a machine for an interactive session.

### DIFF
--- a/src/platform/backends/qemu/qemu_vmstate_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vmstate_process_spec.cpp
@@ -29,17 +29,13 @@ QStringList mp::QemuVmStateProcessSpec::arguments() const
 {
     QStringList args;
 
-    args << platform_args;
-
-#if defined Q_PROCESSOR_X86
-    args << "-nographic"
-         << "-dump-vmstate" << file_name;
-#elif defined Q_PROCESSOR_ARM
-    args << "-machine"
+    args << platform_args
+#if defined Q_PROCESSOR_ARM
+         << "-machine"
          << "virt"
+#endif
          << "-nographic"
          << "-dump-vmstate" << file_name;
-#endif
 
     return args;
 }
--- a/src/platform/backends/qemu/qemu_vmstate_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vmstate_process_spec.cpp
@@ -31,8 +31,15 @@ QStringList mp::QemuVmStateProcessSpec::arguments() const
 
     args << platform_args;
 
+#if defined Q_PROCESSOR_X86
     args << "-nographic"
          << "-dump-vmstate" << file_name;
+#elif defined Q_PROCESSOR_ARM
+    args << "-machine"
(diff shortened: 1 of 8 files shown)


xmkg (commented):
LGTM code-wise (only a minor comment). I haven't tested the functionality yet.

src/platform/backends/qemu/qemu_vmstate_process_spec.cpp
```diff
@@ -29,9 +29,12 @@ QStringList mp::QemuVmStateProcessSpec::arguments() const
 {
     QStringList args;
 
-    args << platform_args;
-
-    args << "-nographic"
+    args << platform_args
+#if defined Q_PROCESSOR_ARM
+         << "-machine"
```
georgeliao, line 34: This `Q_PROCESSOR_ARM` branch is needed for linux arm but it affects both linux arm and macos arm. However, in the case of macos arms, the default argument value is also virt, so the behavior on macos arm remains the same.

georgeliao:
@ricab In development environment, that needs to be done to make it run, not in snap environment. For reviewer's testing, I would assume only the snap environment is involved. right?

ricab:
Ah right, thanks. Perhaps this PR could add a line to say so in the Linux build instructions?

Sploder12 (dismissed):
LGTM! The default driver was qemu when I installed the snap on an arm64 testflinger machine. Launching an instance worked fine as well. 

Had to install the stable snap first, which was using lxd, before the manually installed snap worked (was giving me authentication and dnsmasq issues), but I don't believe that is related to this.

xmkg (dismissed):

xmkg:
Looks good to me (secondary review).

PR #4058, opened 2025-04-24
One Repo To Rule Them All

ricab:


ricab:
Alright, this is ready for review. A few points:

- The expected macOS and Windows checks are not relevant here: they correspond to the old triggers onto the private repo. New macOS/Windows jobs are run directly here.
- The CLA check is complaining on commits that were already on full. It doesn't make sense to attempt editing commits from years ago which were already accepted IMO. I have fixed the authorship of any new commits (unless something escaped me).
- Coverage in full wasn't tracked and it's no surprise that it was lower. A total -0.06% change isn't bad overall, I think.

src/platform/platform_proprietary.cpp
```diff
@@ -15,19 +15,12 @@
...
-#ifndef MULTIPASS_UNSUPPORTED_REMOTE_EXCEPTION_H
-#define MULTIPASS_UNSUPPORTED_REMOTE_EXCEPTION_H
+#include <multipass/format.h>
+#include <multipass/platform.h>
 
-#include <stdexcept>
+namespace mp = multipass;
 
-namespace multipass
+std::string mp::platform::host_version()
```
sharder996, line 23: This file isn't really named appropriately anymore. And, this function is really applicable to the `platform` directory since there is no platform specific code. WDYT

ricab: Hey Scott, I retraced my steps concerning this and I referred to this in this [commit](https://github.com/canonical/multipass/commit/d6cadea974a7240e2aba4c80a7137269dff6cde7)'s message. I guess it's still correct to refer to macOS and Windows as proprietary platforms and include common implementations here, no? Otherwise I'd have to duplicate the code in both.

ricab: But it does change based on the platform: on Linux, there is a different implementation (to deal with snaps).

ricab: I've added some file-level docs, let me know if that looks better.

sharder996: Ah, I see. Makes more sense now.

georgeliao:
@ricab 
I have browsed through all the diff between `public/merger` and `private/full`, things look good to me. I also did smoke test on all three platforms, everything works fine.

georgeliao (dismissed):

ricab:
Thanks @georgeliao.

How about you @sharder996? Does this have your stamp of approval?

sharder996:
@ricab Yup, +1 from me!

sharder996 (dismissed):

ricab:
Alright, this isn't going to merge because it keeps waiting for the old checks. I have dropped the commit to skip doc checks, since that is now in through #4076. @sharder996 or @georgeliao, could one of you please give this a green checkmark? Then I'll merge manually.

georgeliao (approved):

ricab:
dance

townsend2010:
Just noticed this was merged.  What a miraculous day!!!!

PR #4059, opened 2025-04-24
Merge ci

ricab:


ricab:
Hitting a [32 char limit](https://snapcraft.io/docs/snapcraft-yaml-schema#p-21225-version) on the version...

Doh

PR #4060, opened 2025-04-24
[gui] vm_delete button uses normal buildButton styling

levkropp:
image

This pull request simplifies the `VmActionButton` widget in the `src/client/gui/lib/vm_action.dart` file by removing special styling related to the `delete` action. The Delete button previously was styled with a red background color, which was too strong of a call-to-action compared to the other Vm action buttons.


Sploder12 (approved):
LGTM!

PR #4061, opened 2025-04-24
Az implementation

Sploder12:
This PR adds implementations for making VMs unavailable and available. Also has unavailable VMs ignore commands that change anything related to that VM. There are a few known race conditions that are outside the scope of AZs to fix.


src/daemon/daemon.cpp
```diff
@@ -3517,6 +3563,13 @@ mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::string& name, con
     {
         auto it = operative_instances.find(name);
         auto vm = it->second;
+
+        // unchecked iterator access :)
```
xmkg, line 3567: ;)

src/daemon/daemon.cpp
```diff
@@ -2028,6 +2030,12 @@ try // clang-format on
             continue;
         }
 
+        if (vm->current_state() == VirtualMachine::State::unavailable)
+        {
+            mpl::log(mpl::Level::info, name, "Ignoring mount since instance unavailable.");
```
xmkg, line 2035: Hmm... I think we need a better way to represent "unavailability". My main concern is that this approach is error-prone and repetitive. The problem is that "unavailable" instances are still a part of the operative instances, which is wrong. An unavailable instance is not operative since it can't be put into an operative state unless it's made available, so an unavailable instance shouldn't be a part of the "operative" set of VMs in the first place. 

Right now, we have two sets of VMs: "operative" and "deleted". I think what we need here is a third category, "unavailable". This way, we can just update the error messages as `does not exist or unavailable` and call it a day, until we overhaul the VM management logic.  @ricab, what do you think?

ricab: Hmm good point. I haven't thought a lot about it, but I think I agree with you @xmkg. We need to define also what happens at the intersection of deleted/unavailable i.e., deleted instances that belong to an unavailable AZ. I guess we just declare "deleted" takes precedence over "unavailable"? We need to answer a few questions:

- does deleting and instance unsubscribe it from the AZ?
- does enabling/disabling an AZ affect deleted instances? (I hope not)
- what happens on recover of an instance whose AZ is unavailable?

xmkg: Hi @ricab, good questions. Regarding the deleted + unavailable at the same time, I think it's reasonable to make deleted take precedence over unavailable. Another avenue might be to consolidate all instances to a single container and represent the "availability" state in a bit flag. This way, an instance can be deleted & unavailable at the same time. 

Consolidating all instances into a single container carries a cost: lookup and iteration speed. This can be troublesome for users with many instances. Otherwise, std::ranges filtering would work just fine. If iteration speed matters,  another option might be using a multi-index container, which allows for different indices over a set of values instead of just one.

Regarding the last three questions, those should be answered in the AZ spec itself, if not already. @Sploder12 :wink:

Sploder12: @xmkg I fully agree we need a better way to represent "unavailability", it is known that my current approach is terrible. But, my issue with using operative instances and a new unavailable table is that we'd be moving even more responsibility into the daemon. Also operative instances is the most unsafe thing in our entire code base at the moment. It guarantees race conditions.

xmkg: @Sploder12 understood. I reviewed the patch in my local environment as well, and it seems like there is no common place that we can place the "unavailable" check. The commands are inconsistent in how they retrieve the instances. Some of them use the `operative_instances` directly, whereas some use the `select_instances` function, and so on. We desperately need to refactor this because it can be implemented in a much simpler and safer way.

@ricab, shall we proceed as-is and refactor this later on?

src/daemon/daemon.cpp
```diff
@@ -3589,6 +3635,12 @@ mp::Daemon::async_wait_for_ssh_and_start_mounts_for(const std::string& name, con
             return fmt::to_string(errors);
         }
         const auto vm = it->second;
+
+        if (vm->current_state() == VirtualMachine::State::unavailable)
+        {
+            return fmt::to_string(errors);
```
xmkg: Maybe push a message into "errors", stating that a VM is unavailable here?

include/multipass/virtual_machine.h
```diff
@@ -137,6 +137,7 @@ class VirtualMachine : private DisabledCopyMove
 
 protected:
     const QDir instance_dir;
+    bool was_running{false};
```
ricab: Hmm, why place this here? And why protected? Can't this be a private field in `BaseVirtualMachine`?

src/daemon/daemon.cpp
ricab: I've been thinking quite a bit about this and wrote a bunch (sorry). We need to discuss, but I think the effect of what follows will be small for this PR. But better have our ideas in order.

#### Deleted instances vs AZs

I see two possible approaches:

1. instances remain subscribed on deletion
    * deleted instances' state is updated on AZ status change. The following changes are possible on a deleted instance
        + arbitrary state -> unavailable (available -> unavailable is a particular case)
        + unavailable -> available
    * recovering a deleted instance does not touch the state

2. instances unsubscribe on deletion
    * states don't change while an instance is deleted
    * instances would need to keep track of their AZ in a different way
    * recovering a deleted instance does two new things
        + resubscribe it to its AZ
        + update its state according to AZ state

If this was being implemented from scratch, I would probably prefer 2 and drop the AZ manager entity entirely, as discussed with Andrei in early AZ days. I still don't see the benefit, but it's not the time to U-turn now. 

So, that leaves option 1. IIUC, that corresponds to what is implemented, correct @Sploder12?

#### Which container

`operative_instances` is just a name. The idea was to contrast it with `deleted_instances` and indicate that those are disjoint collections. Previously it was called `instances` (IIRC), which sounded like a superset. If necessary we could rename again (e.g. `retained_instances`).

I would love to have a single collection, as @xmkg suggested, with ranges and "deleted" as just another state in the FSM (same representation). It would simplify a lot of code.

> Consolidating all instances into a single container carries a cost: lookup and iteration speed.

In the sizes we're we're talking about, I am pretty sure it would be OK, if we *did it right*. If the deleted state was directly available in a plain field of the VM class, I would be willing to bet a user with 1000 deleted instances and 1 other instance would notice no degradation of performance with one vs two hash tables. 1000 iterations of "if X continue" are really fast, especially if X is already cached. It might even work the other way around due to cache locality. And I'd bet very few people ever reached 100 instances anyway.

We would have to *do it right* though, in the sense that we would need to make sure we shortcut the state update when it was deleted, preferably minimizing virtual calls in that case (another case for template method). 1000 iterations of going to the backend to update the state *would be* very noticeable.

I suspect that perf did not play much of a role in the decision to have 2 collections in the first place. My guess is that it was just the immediate representation of a trash can in code. Idk...

All that said, this would be a very significant refactor. I wonder if AZs are the right place for it... Especially given that we don't have ranges yet.

> @ricab, shall we proceed as-is and refactor this later on?

Meh, probably... :/

src/daemon/daemon.cpp
```diff
@@ -2202,6 +2204,12 @@ try
             continue;
         }
 
+        if (vm->current_state() == VirtualMachine::State::unavailable)
```
ricab, line 2088: Shouldn't we `add_fmt_to(errors,...)` instead, as in line 2196? If the instance is unavailable, trying to mount something on it is an error, right?

src/daemon/daemon.cpp
```diff
@@ -2451,6 +2462,16 @@ try
             operation = [this, delay_minutes = std::chrono::minutes(request->time_minutes())](
                             VirtualMachine& vm) { return this->shutdown_vm(vm, delay_minutes); };
 
+        operation = [op = std::move(operation)](VirtualMachine& vm) {
+            if (vm.current_state() == VirtualMachine::State::unavailable)
+            {
+                mpl::log(mpl::Level::info, vm.vm_name, "Ignoring stop since instance is unavailable.");
```
ricab: OK, in this case it is basically a no-op, so this makes sense.

src/daemon/daemon.cpp
```diff
@@ -2527,6 +2554,12 @@ try
 
     const auto& instance_targets = instance_selection.operative_selection;
     status = cmd_vms(instance_targets, [this](auto& vm) {
+        if (vm.current_state() == VirtualMachine::State::unavailable)
```
ricab: Hmm, restart used to fail when the instance wasn't running. Why would we introduce this new case?

States are already handled in `reboot_vm` (though the error is now wrong https://github.com/canonical/multipass/issues/4314). We could argue that we should still restart other instances (do the state filtering first), but that would be a separate matter.

src/daemon/daemon.cpp
```diff
@@ -2607,6 +2640,12 @@ try
             {
                 const auto& instance_name = vm_it->first;
 
+                if (vm_it->second->current_state() == VirtualMachine::State::unavailable)
+                {
+                    mpl::log(mpl::Level::info, instance_name, "Ignoring delete since instance is unavailable.");
```
ricab: Ouch, this looks a little radical. Suppose that are 5 running instances in a zone before you disable it. You then decide to delete some of them. Do we require people to restart all of them first?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -59,7 +59,7 @@ constexpr auto yes_overwrite = true;
 
 void assert_vm_stopped(St state)
 {
-    assert(state == St::off || state == St::stopped);
+    assert(state == St::off || state == St::stopped || state == St::unavailable);
```
ricab: Why add this here? I don't think we should allow taking or restoring snapshots on unavailable instances. Those are the only two operations using this.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -238,6 +238,29 @@ void mp::BaseVirtualMachine::check_state_for_shutdown(ShutdownPolicy shutdown_po
...
+{
+    if (available)
+    {
+        state = State::off;
+        update_state();
+        if (was_running)
+        {
+            start();
+
+            // normally the daemon sets the state to running...
```
ricab, line 258: 🤦‍♂️

src/daemon/daemon.cpp
```diff
@@ -3942,6 +3988,12 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
             return fmt::to_string(errors);
         }
         const auto vm = it->second;
+
+        if (vm->current_state() == VirtualMachine::State::unavailable)
```
ricab: Could you explain the rationale for this check here, for my benefit please? Is it just trying to reduce the chances of AZ and instance state going out of sync due to races?

ricab (commented):
Thanks @Sploder12. Overall this looks close to what we need, but I have a few comments for discussion.

src/daemon/daemon.cpp
ricab: Discussed and alternatives are considered worse at this point.

src/daemon/daemon.cpp
```diff
@@ -2383,6 +2391,11 @@ try
                            "Cannot start the instance '{}' while suspending.",
                            name);
             continue;
+        case VirtualMachine::State::unavailable:
+            fmt::format_to(std::back_inserter(start_errors),
+                           "Cannot start the instance '{}' while unavailable.",
+                           name);
+            continue;
```
ricab, line 2277: Can we merge this with the above case?

src/daemon/daemon.cpp
```diff
@@ -2452,6 +2465,18 @@ try
             operation = [this, delay_minutes = std::chrono::minutes(request->time_minutes())](
                             VirtualMachine& vm) { return this->shutdown_vm(vm, delay_minutes); };
 
+        operation = [op = std::move(operation)](VirtualMachine& vm) {
+            if (vm.current_state() == VirtualMachine::State::unavailable)
+            {
+                mpl::log(mpl::Level::info,
+                         vm.vm_name,
+                         "Ignoring stop since instance is unavailable.");
+                return grpc::Status::OK;
```
ricab: Shouldn't this be handled by the VM instead? If you try to stop a VM that is already stopped, it knows to ignore it. Shouldn't that be the case with this one too? Perhaps with the log, which could be present in both cases IMO.

src/daemon/daemon.cpp
```diff
@@ -2487,6 +2512,14 @@ try
     {
         config->factory->require_suspend_support();
         status = cmd_vms(instance_selection.operative_selection, [this](auto& vm) {
+            if (vm.current_state() == VirtualMachine::State::unavailable)
+            {
+                mpl::log(mpl::Level::info,
+                         vm.vm_name,
+                         "Ignoring suspend since instance is unavailable.");
+                return grpc::Status::OK;
+            }
```
ricab, line 2385: Idem: shouldn't this be handled by the VM? It would go in the direction of relieving the daemon from such concerns.

src/platform/backends/shared/base_availability_zone.cpp
```diff
@@ -118,8 +118,32 @@ void BaseAvailabilityZone::set_available(const bool new_available)
...
+    try
+    {
+        for (auto& vm : m.vms)
+            vm.get().set_available(new_available);
+    }
+    catch (...)
+    {
+        // if an error occurs fallback to available.
+        m.available = true;
+        serialize();
```
ricab: Couldn't we just do this at the end?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -238,6 +238,29 @@ void mp::BaseVirtualMachine::check_state_for_shutdown(ShutdownPolicy shutdown_po
...
+            state = State::running;
+            update_state();
+        }
+        return;
+    }
+
+    was_running = state == State::running || state == State::starting;
```
ricab: What if the state was restarting?

src/platform/backends/shared/base_availability_zone.cpp
Sploder12: since the catch is rethrowing, control flow would never reach the end of the function

src/daemon/daemon.cpp
```diff
@@ -2259,9 +2267,12 @@ try
             continue;
         }
         case VirtualMachine::State::suspending:
+        case VirtualMachine::State::unavailable:
             fmt::format_to(std::back_inserter(start_errors),
-                           "Cannot start the instance '{}' while suspending.",
-                           name);
+                           "Cannot start the instance '{}' while {}.",
+                           name,
+                           vm.current_state() == VirtualMachine::State::suspending ? "suspending"
```
ricab, line 2275: This will be easier once the Hyper-V branch is in: https://github.com/canonical/multipass/pull/4080/files#diff-d847b18fbc46d755306da3eb6cac52de71043c1cc87ff345cdf8b034ad141ffaR158

You could add a TODO here, or convince @xmkg to split that off 😉

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -204,6 +204,10 @@ void mp::BaseVirtualMachine::check_state_for_shutdown(ShutdownPolicy shutdown_po
     {
         throw VMStateIdempotentException{"Ignoring shutdown since instance is already stopped."};
     }
+    if (state == State::unavailable)
+    {
+        throw VMStateIdempotentException{"Ignoring shutdown since instance is unavailable."};
+    }
```
ricab: How about merging with the above branch here as well. Again, the formatter would help. If you want to simplify, I wouldn't be shocked if the message read "Ignoring shutdown since instance is already unavailable".

src/platform/backends/shared/base_availability_zone.cpp
ricab: Right. You could use a scope guard, but no need really.

src/platform/backends/hyperv/hyperv_virtual_machine.cpp
```diff
@@ -412,6 +412,10 @@ void mp::HyperVVirtualMachine::suspend()
     {
         mpl::info(vm_name, "Ignoring suspend issued while stopped");
     }
+    else if (present_state == State::unavailable)
+    {
+        mpl::log(mpl::Level::info, vm_name, "Ignoring suspend since instance is unavailable");
```
ricab: idem: merge with the above?

Otherwise, use `mpl::info`?

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -426,9 +426,11 @@ void mp::QemuVirtualMachine::suspend()
 
         vm_process.reset(nullptr);
     }
-    else if (state == State::off || state == State::suspended)
+    else if (state == State::off || state == State::suspended || state == State::unavailable)
     {
-        mpl::info(vm_name, "Ignoring suspend issued while stopped/suspended");
+        mpl::log(mpl::Level::info,
+                 vm_name,
+                 "Ignoring suspend issued while stopped/suspended/unavailable");
```
ricab: idem: state formatting would help

src/platform/backends/shared/base_availability_zone.cpp
```diff
@@ -117,8 +117,32 @@ void BaseAvailabilityZone::set_available(const bool new_available)
...
+        // make sure nothing is still unavailable.
+        for (auto& vm : m.vms)
+        {
+            // setting the state here breaks encapsulation, but it's already broken.
+            std::unique_lock vm_lock{vm.get().state_mutex};
+            if (vm.get().current_state() == VirtualMachine::State::unavailable)
+            {
+                vm.get().state = VirtualMachine::State::off;
+                vm.get().update_state();
+            }
```
ricab, line 150: Have you considered force-stop instead? That would avoid breaking encapsulation yet again and make sure there were no leftover processes.

ricab (approved):
I still have a few suggestions, but I will leave them to your consideration and pre-approve.

src/platform/backends/shared/base_availability_zone.cpp
Sploder12: shutdown is ignored when an instance is unavailable. There aren't any leftover processes since an unavailable instance is always off. Using `set_available(true)` could make the VM start which is undesirable.

src/daemon/daemon.cpp
```diff
@@ -3730,6 +3765,7 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
             return fmt::to_string(errors);
         }
         const auto vm = it->second;
+
```
tobe2098: Is this newline added for clarity?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -238,6 +241,29 @@ void mp::BaseVirtualMachine::check_state_for_shutdown(ShutdownPolicy shutdown_po
...
+            start();
+
+            // normally the daemon sets the state to running...
+            state = State::running;
+            update_state();
+        }
+        return;
+    }
+
+    was_running = state == State::running || state == State::starting || state == State::restarting;
```
tobe2098, line 265: I do not know how relevant this would be, but running set_available(false) twice will always lose the was_running state. The simplest fix would be that if the state of availability is the same as the one being set we should return early at the beggining, but there could be a better approach.

tobe2098: An alternative would be to have a separate availability variable from state, but I see the general implementation is already mature around the concept of unavailability being "just another state".

tests/test_base_virtual_machine.cpp
```diff
@@ -1441,4 +1445,34 @@ TEST_F(BaseVM, sshExecRethrowsSSHExceptionsWhenConnected)
...
+TEST_F(BaseVM, setAvailableRestartsRunning)
+{
+    StubBaseVirtualMachine base_vm(zone, St::running);
+
+    base_vm.set_available(false);
+    ASSERT_EQ(base_vm.current_state(), St::unavailable);
+
+    base_vm.set_available(true);
+    EXPECT_EQ(base_vm.current_state(), St::running);
+}
```
tobe2098, line 1468: Related to what I said before about losing the was_running state, in this test or in another one we could make clear what behavior is intended (if state is intended to be lost, two calls with false lose the state or the opposite).

tobe2098 (changes requested):
Looks very good! Just a nitpick and a question about how set_availability works.

src/platform/backends/shared/base_virtual_machine.cpp
tobe2098: This is just a nitpick since the logic works now, but maybe 
```
if (available == state) return;
else if (available){
...
```
Would simplify the checks/logic? Could there be some code that we want to execute on repeated .set_available(available)?

Sploder12: there is no code we would want to execute on repeated calls

tobe2098 (approved):
LGTM!

PR #4064, opened 2025-04-25
[exceptions] add LoggedExceptionBase type

xmkg:
LoggedExceptionBase is an extension to FormattedExceptionBase where the produced exception message would be logged too. This streamlines the scenario where the exception message is also desired to be logged.


xmkg:
Ditching this idea.

PR #4065, opened 2025-04-29
[ci] Free disk space on Linux builds

ricab:
Jobs started running out of disk space.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -85,6 +85,14 @@ jobs:
 
     timeout-minutes: 120
     steps:
+
+    # Free some disk space to avoid the "No space left on device" error.
+    - name: Free Disk Space
+      uses: jlumbroso/free-disk-space@v1.3.1
+      with:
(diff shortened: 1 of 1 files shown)


georgeliao (approved):
Thanks for the good work. It looks sane to me. Moreover, the linux debug build passed.

levkropp (approved):
LGTM! The only thing that I see that we might want to keep in mind is according to the [github for free-disk-space](https://github.com/jlumbroso/free-disk-space), it can take up to 3 minutes to run with the most amount of time being uninstalling packages with `apt`, but in the context of the total `BuildAndTest` job that's not a very long time at all

ricab:
Thanks Lev! It took 9 minutes in my first attempts, but I pruned it down to only some quick stuff and I think it's taking around 1min in our case. You can see the time logged next to the step in GHA, e.g. [here](https://github.com/canonical/multipass/actions/runs/14730559374/job/41348470802?pr=4065)

Issue #3919, opened 2025-02-05
Multipass GUI does not honour question close setting when quitting from the tray


ricab:
**Describe the bug**

This setting:

![Image](https://github.com/user-attachments/assets/2604ad9e-f620-4cdf-9e44-df3908e88efb)

Doesn't appear to honoured when I click quit on the tray. Running instances carry on even if I have "Stop running instances" and the question is not shown if I have "Ask...".

**To Reproduce**
1. Open the GUI
2. Start an instance
3. Configure the setting above
4. Click the tray icon and select "Quit"

**Expected behavior**
The setting would be honoured, by stopping instances or focusing the GUI and showing a question, as appropriate.

Artem-OSSRevival:
Hi there. I am Artem from [OSS Revival](https://github.com/Project-OSS-Revival). We specialize in revitalizing and maintaining open-source projects for long-term stability. I'm already working on a solution to this issue.

PR #4068, opened 2025-05-01
Honour the close settings when quitting from the tray

Artem-OSSRevival:
Closes: #3919
Previously, the tray menu action **'Quit'** caused a function to immediately destroy the window. Now the usual process of closing window is taking place. This allows to consider the settings and perform the necessary tasks before closing.

Additionally, the restoration functionality of a hidden/minimized window has been added if there is a setting of the need to show a dialog before closing (**'Ask about running instances'**).

**To reproduce:**
Set setting _'Stop running instances'_ or _'Ask about running instances'_
settings

Launch instance:
run1_gui

**Don't close the app Window, but Quit from the tray.**

Before:
multipass_#3919_before

After:
multipass_#3919_after

### DIFF
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -166,7 +166,18 @@ class _AppState extends ConsumerState<App> with WindowListener {
     final daemonAvailable = ref.read(daemonAvailableProvider);
     final vmsRunning =
         ref.read(vmStatusesProvider).values.contains(Status.RUNNING);
-    if (!daemonAvailable || !vmsRunning) windowManager.destroy();
+    final closeJob = ref.read(guiSettingProvider(onAppCloseKey));
+
+    // nothing to do
+    if (!daemonAvailable || !vmsRunning || closeJob == 'nothing') {
+      windowManager.destroy();
+      return;
+    }
+
+    // checking the need to restore the window
+    if (!await windowManager.isVisible() || await windowManager.isMinimized()) {
+      windowManager.showAndRestore();
+    }
 
     stopAllInstances() {
       final notificationsNotifier = ref.read(notificationsProvider.notifier);
@@ -181,31 +192,28 @@ class _AppState extends ConsumerState<App> with WindowListener {
       );
     }
 
-    switch (ref.read(guiSettingProvider(onAppCloseKey))) {
-      case 'nothing':
-        windowManager.destroy();
-      case 'stop':
-        stopAllInstances();
-      default:
-        showDialog(
-          context: context,
-          barrierDismissible: false,
-          builder: (context) => BeforeQuitDialog(
-            onStop: (remember) {
-              ref
-                  .read(guiSettingProvider(onAppCloseKey).notifier)
-                  .set(remember ? 'stop' : 'ask');
-              stopAllInstances();
-              Navigator.pop(context);
-            },
-            onKeep: (remember) {
-              ref
-                  .read(guiSettingProvider(onAppCloseKey).notifier)
-                  .set(remember ? 'nothing' : 'ask');
-              windowManager.destroy();
-            },
-          ),
-        );
+    if (closeJob == 'ask') {
+      showDialog(
+        context: context,
+        barrierDismissible: false,
+        builder: (context) => BeforeQuitDialog(
+          onStop: (remember) {
+            ref
+                .read(guiSettingProvider(onAppCloseKey).notifier)
+                .set(remember ? 'stop' : 'ask');
+            stopAllInstances();
+            Navigator.pop(context);
+          },
+          onKeep: (remember) {
+            ref
+                .read(guiSettingProvider(onAppCloseKey).notifier)
+                .set(remember ? 'nothing' : 'ask');
+            windowManager.destroy();
+          },
+        ),
+      );
+    } else {
+      stopAllInstances();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ricab:
Nice, thank you @Artem-OSSRevival!

It looks like the Lint job is failing in CI. Would you mind formatting the code as [recommended in the log](https://github.com/canonical/multipass/actions/runs/14772418522/job/41531374448?pr=4068#step:3:142)?

src/client/gui/lib/main.dart
```diff
@@ -187,6 +187,10 @@ class _AppState extends ConsumerState<App> with WindowListener {
       case 'stop':
         stopAllInstances();
       default:
+        if (!await windowManager.isVisible() ||
+            await windowManager.isMinimized()) {
+          windowManager.showAndRestore();
+        }
```
ricab, line 193: I am debating whether the window should be shown also when stopping all instances (case stop). It can take a while to stop all instances. If the window is visible, there's a spinner informing the user of what is going on, but not if it's hidden. So I think I am inclined that way. On the other hand, if there are no instances to stop, there is no point bringing the window back up...

ricab (commented):
Hi @Artem-OSSRevival, thank you for the PR! 

This is certainly an improvement. There is only a case that I am thinking we could handle better (comment inline). I would like to know your thoughts.

Also, it appears you [haven't signed](https://github.com/canonical/multipass/actions/runs/14797651287/job/41643294957?pr=4068) the [CLA](https://ubuntu.com/legal/contributors) yet? Would you mind doing so, please?

src/client/gui/lib/main.dart
Artem-OSSRevival: The window will be shown only if there is a pop-up dialogue case.

ricab: I understand what you mean, that pop-up is not ideal, particularly if there are no instances to stop. However, depending on how many VMs there are and what they are doing, stopping them can take a lot longer than a couple of seconds (we even had to increase the timeout to 5 minutes [in the past](https://github.com/canonical/multipass/commit/e29d848e7f406ef9213f50aa8ecc77b3a46e4789)). Until they stop, there would be no feedback and it would just feel like the tray isn't reacting.

Artem-OSSRevival: Yes, with a very long delay, it would be better to show the presence of a stopping process.
May be implemented something like:

```
closeJob = ref.read(guiSettingProvider(onAppCloseKey);
if (closeJob != 'nothing' &&
     (!await windowManager.isVisible() ||
      await windowManager.isMinimized())
{
          windowManager.showAndRestore();
}
```
or even restore only invisible:
```
case 'stop':
  if (!await windowManager.isVisible()) windowManager.showAndRestore();
  stopAllInstances();
default:
```

ricab: Something like that, although I think we'd better drop the switch at that point. Perhaps something like, in line 169, get the setting and add `|| closeJob == 'nothing'` to the `if` (and return). Then add an else to take care of stop and default cases, where the window should be restored if invisible/minimized. WDYT?

Artem-OSSRevival: Yes, I thought about it initially, but was afraid that you would not want to get rid of switch...case. Without the switch, I can do something like this (already tried locally):

```
void onWindowClose() async {
  if (!await windowManager.isPreventClose()) return;
  final daemonAvailable = ref.read(daemonAvailableProvider);
  final vmsRunning =
      ref.read(vmStatusesProvider).values.contains(Status.RUNNING);
  final closeJob = ref.read(guiSettingProvider(onAppCloseKey));

  // nothing to do
  if (!daemonAvailable ||
      !vmsRunning ||
      closeJob == 'nothing') {
    windowManager.destroy();
    return;
  }

  // checking the need to restore the window
  if (!await windowManager.isVisible() ||
      await windowManager.isMinimized()) {
    windowManager.showAndRestore();
  }

  stopAllInstances() {
    final notificationsNotifier = ref.read(notificationsProvider.notifier);
    notificationsNotifier.addOperation(
      ref.read(grpcClientProvider).stop([]),
      loading: 'Stopping all instances',
      onError: (error) => 'Failed to stop all instances: $error',
      onSuccess: (_) {
        windowManager.destroy();
        return 'Stopped all instances';
      },
    );
  }

  if (closeJob == 'ask') {
    showDialog(
      context: context,
      barrierDismissible: false,
      builder: (context) => BeforeQuitDialog(
        onStop: (remember) {
          ref
              .read(guiSettingProvider(onAppCloseKey).notifier)
              .set(remember ? 'stop' : 'ask');
          stopAllInstances();
          Navigator.pop(context);
        },
        onKeep: (remember) {
          ref
              .read(guiSettingProvider(onAppCloseKey).notifier)
              .set(remember ? 'nothing' : 'ask');
          windowManager.destroy();
        },
      ),
    );
  } else {
    stopAllInstances();
  }
}
```

ricab: Yep, that makes the most sense to me.

ricab:
Hey @Artem-OSSRevival, would you mind rebasing this and doing a formatting pass to get it through CI? Thanks!

ricab (approved):
Great stuff, thanks!

ricab:
Hi @Artem-OSSRevival, our CI is unfortunately unable to deal with external PRs properly ATM. I have opened https://github.com/canonical/multipass/issues/4134 for that, but for the time being we'll need to push your branch to our repo and make a clone of this PR. You will still retain authorship of the commits, of course, and we'll be sure to mention you in the PR that gets merged! I hope you don't mind :slightly_smiling_face: 

Thanks again for your valuable contribution!

ghost:
Hi @ricab. No problems. The main thing is the prosperity of the project :)

Issue #3682, opened 2024-09-18
Bash-complete more instances in`multipass stop --force`


ricab:
When the --force flag is used with `multipass stop`, suspended/ing and re/starting instances should be completed too. For example:

```sh
$ multipass ls
Name                    State             IPv4             Image
natty-nilgai            Suspended         --               Ubuntu 24.04 LTS
$ multipass stop --force n[TAB]
```

`natty-nilgai` should be completed.

roy9495:
Hi, can I work on this issue?

ricab:
Hi @roy9495, certainly! Thank you for your interest!

Artem-OSSRevival:
Hi there. If no one is mind, I will work on this Issue.

ricab:
Hi @Artem-OSSRevival, we haven't had any news on this, so go for it!

PR #4072, opened 2025-05-06
Added more Bash-complete instances in multipass stop --force

Artem-OSSRevival:
Closes: #3682
Previously, when the TAB pressed, only ‘Running’ instances were completed.
Now, if the `--force` flag is provided, additionally taken into account 'Starting', 'Restarting', 'Suspending' and 'Suspended' ones.

**To reproduce:**
Run service:
`sudo /usr/local/bin/multipassd`

Run GUI:
`/usr/local/bin/multipass.gui`

start and then **suspend the instance**

**OR** start/stop/suspend tasks can be done from the CLI:
`multipass start --all`
`multipass suspend --all`

**Before:**
Open Bash and add the script:
`source <multipass>/completions/bash/multipass`

Check the instance is suspended:
`multipass ls`

Try the bash-completion (**'n'** is the first letter of the instance’s name):
`multipass stop --force n[TAB]`

_Nothing happens._

**After:**
Reopen Bash and add the script:
`source <multipass>/completions/bash/multipass`
`multipass ls`
`multipass stop --force n[TAB]`

`natty-nilgai` (real instance name) will be completed.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -348,7 +348,21 @@ _multipass_complete()
 
     if [[ "$prev_opts" = false ]]; then
         case "${cmd}" in
-            "exec"|"stop"|"suspend"|"restart")
+            "exec"|"suspend"|"restart")
+                _multipass_instances "Running"
+            ;;
+            "stop")
+                _get_comp_words_by_ref -n := -w WORDS -i CWORD cur prev
+                for ((i=2; i<CWORD; i++)); do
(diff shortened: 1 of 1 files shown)


xmkg (approved):
LGTM, thanks for the contribution! A small improvement would be accepting an array of states to _multipass_instances to avoid repetition, but I'm okay with the current state as well.

ricab:
?

Artem-OSSRevival:
Sorry, my fault :(

PR #4074, opened 2025-05-06
[tests/mock_logger] Add substr as Description() for distinction

xmkg:
Right now, it's hard to tell which expect_log call is failed to be satisfied because the output is ambiguous:

```
 error: Actual function call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr)))...
         Expected: to be called once
```
The patch fixes that by utilizing the matcher description:
````
 error: Actual function "my expected log string" call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr)))...
         Expected: to be called once
```

### DIFF
--- a/tests/mock_logger.cpp
+++ b/tests/mock_logger.cpp
@@ -17,6 +17,8 @@
 
 #include "mock_logger.h"
 
+#include <fmt/format.h>
+
 #include <type_traits>
 
 namespace mp = multipass;
@@ -50,7 +52,9 @@ mpt::MockLogger::Scope::~Scope()
 
 void mpt::MockLogger::expect_log(mpl::Level lvl, const std::string& substr, const Cardinality& times)
 {
-    EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).Times(times).Description(substr);
+    EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr)))
+        .Times(times)
+        .Description(fmt::format("log(level: {}, substr: '{}')", logging::as_string(lvl), substr));
 }
 
 void mpt::MockLogger::screen_logs(mpl::Level lvl)
--- a/tests/mock_logger.cpp
+++ b/tests/mock_logger.cpp
@@ -50,7 +50,7 @@ mpt::MockLogger::Scope::~Scope()
 
 void mpt::MockLogger::expect_log(mpl::Level lvl, const std::string& substr, const Cardinality& times)
 {
-    EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).Times(times);
+    EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).Times(times).Description(substr);
 }
 
 void mpt::MockLogger::screen_logs(mpl::Level lvl)


ricab (commented):
Good idea, but the sentence looks a little strange. Could do `Description(fmt::format("log with {}", substr))` or maybe just use operator << ?

xmkg:
> Good idea, but the sentence looks a little strange. Could do `Description(fmt::format("log with {}", substr))` or maybe just use operator << ?

Hmm, EXPECT_CALL does not seem to implement "operator <<". I agree that we could use a nicer message, though. How about sth like:

```
Actual function "expect_log(level: debug, substr: "message", times: 1)" call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).
```

ricab:
Hmm, it's `log` whose call count doesn't match, not `expect_log`, right?

xmkg:
> Hmm, it's `log` whose call count doesn't match, not `expect_log`, right?

The main problem comes from the assumed placement of the description. It's right between the [actual] and [expected], and whatever we write into the description will be associated with the actual part as a cognitive default. Given that we don't have an easy way to manipulate this ordering, let's ditch the function name altogether then. How about:

```
Actual function "level: debug, substr: `message`" call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))
 ```

ricab:
There is maybe something I am missing, but why don't you say this?

> Actual function "log(level: debug, substr: "message")" call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).

IOW, the actual call count for function `log` with those arguments didn't match the expectation. That would be accurate, no?

xmkg:
> There is maybe something I am missing, but why don't you say this?
> 
> > Actual function "log(level: debug, substr: "message")" call count doesn't match EXPECT_CALL(*this, log(lvl, _, HasSubstr(substr))).
> 
> IOW, the actual call count for function `log` with those arguments didn't match the expectation. That would be accurate, no?

I suppose so. There's a slight disconnect between what we actually write in unit tests (expect_log) vs. what's actually being logged (log), but ok. I could live with that.

xmkg:
> Regarding the disconnect you mention, perhaps I haven't been sleeping enough and I may be short-circuiting, but `expect_log` is what the unit tests call to set a call-count expectation on `log`... That is what the message is saying doesn't match. If anything, it's `EXPECT_CALL` that could be replaced with `expect_log` (not necessary).

Yeah, log() is hidden behind the expect_log() call, so what I'm saying is it might be *slightly* harder to correlate the output with the actual unit test case, but that's not that important. I'd take any single clue that would lead to the correct expect_log() line when something goes wrong, so I don't have any strong preference on that.

Regarding the EXPECT_CALL part: unfortunately we can't easily touch the EXPECT_CALL part of the description. I agree that it would look much cleaner, but IMO it's not worth the effort.

ricab:
> Yeah, log() is hidden behind the expect_log() call [...]

`EXPECT_CALL` is — `log()` itself is called in the code being tested. But I understand what you mean better now. It would be good to have an indication of where the failed `expect_log()` was called in the tests. I agree what you have is good enough though :+1:

xmkg:
> Hey @xmkg to keep the web a little simpler, we normally avoid merging the target branch back into the topic branch and rebase instead (except if there are many commits with conflicts). Would you mind doing that here, please?

Done

ricab:
Thank you!

xmkg:
@georgeliao, can I get a secondary review when you're available? The changes are trivial.

sharder996 (approved):
+1 from me

PR #4087, opened 2025-05-15
Fix flakey SFTP server test

### LINKED ISSUE #4086: `SftpServer.handles_readdir_attributes_preserved` Test Fails Due to Timestamp Mismatch
**Describe the bug**

The test `SftpServer.handles_readdir_attributes_preserved` has failed in a CI run for an unrelated PR due to a 1-second timestamp mismatch. This occurs because the atime (access time) in the test file attributes differs by 1 second from the lastModified() timestamp retrieved during the test.

**To Reproduce**

Run the test a lot of times (thousands?)

**Expected behavior**
The test should pass.

**Logs**
```
/Users/cibot/actions-runner/_work/multipass/multipass/tests/test_sftpserver.cpp:1675: Failure
Expected equality of these values:
  test_file_attrs.atime
    Which is: 1747265942
  (uint32_t)test_file_info.lastModified().toSecsSinceEpoch()
    Which is: 1747265943
[  FAILED  ] SftpServer.handles_readdir_attributes_preserved (2 ms)
```

**Additional info**

[Link to the CI run](https://github.com/canonical/multipass/actions/runs/15021195835/job/42249021518?pr=3926)

Sploder12:
The test was expecting returned lastRead to be the same as the expected lastModified, which has a really small chance of not being true. Changed it to check the returned lastModified to be the same as the expected lastModified.

Resolves #4086

### DIFF
--- a/tests/test_sftpserver.cpp
+++ b/tests/test_sftpserver.cpp
@@ -1671,7 +1671,7 @@ TEST_F(SftpServer, handles_readdir_attributes_preserved)
     EXPECT_EQ(test_file_attrs.size, (uint64_t)test_file_info.size());
     EXPECT_EQ(test_file_attrs.gid, test_file_info.groupId());
     EXPECT_EQ(test_file_attrs.uid, test_file_info.ownerId());
-    EXPECT_EQ(test_file_attrs.atime,
+    EXPECT_EQ(test_file_attrs.mtime,
               (uint32_t)test_file_info.lastModified().toSecsSinceEpoch()); // atime64 is zero, expected?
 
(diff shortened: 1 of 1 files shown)


xmkg (approved):
The test seems to use the wrong timestamp in the first place. The fix looks good to me.

xmkg:
https://app.codecov.io/gh/canonical/multipass/pull/4087/indirect-changes

Interesting. We have a test that covers the line that has been reported in the "indirect changes", but Codecov still claims that it's not being covered. We should investigate, but that's not in the scope of this PR.

PR #4089, opened 2025-05-16
[gui] Fix incorrect VM created timestamp while downloading image

levkropp:
image
This pull request introduces a change to the `GeneralDetails` widget in the `src/client/gui/lib/vm_details/vm_details_general.dart` file that ensures a dash ('-') is displayed instead of a timestamp when the VM's image is still downloading and the creation timestamp is zero.

### DIFF
--- a/src/client/gui/lib/vm_details/vm_details_general.dart
+++ b/src/client/gui/lib/vm_details/vm_details_general.dart
@@ -13,6 +13,14 @@ import 'vm_action_buttons.dart';
 import 'vm_details.dart';
 import 'vm_status_icon.dart';
 
+extension InstanceDetailsExtensions on InstanceDetails {
+  String formattedCreationTime({required bool isLaunching}) {
+    final ts = creationTimestamp;
+    if (isLaunching && ts.seconds == 0 && ts.nanos == 0) return '-';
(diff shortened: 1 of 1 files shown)


src/client/gui/lib/vm_details/vm_details_general.dart
```diff
@@ -172,8 +172,13 @@ class GeneralDetails extends ConsumerWidget {
       height: baseVmStatHeight,
       label: 'CREATED',
       child: CopyableText(
-        DateFormat('yyyy-MM-dd HH:mm:ss')
-            .format(info.instanceInfo.creationTimestamp.toDateTime()),
+        // Show a dash when VM is still launching and timestamp is therefore 0
```
xmkg, line 175: The blob below is barely readable. How about using something like:

```dart
extension InstanceDetailsExtensions on InstanceDetails {
  String formattedCreationTime({required bool isLaunching}) {
    final ts = creationTimestamp;
    if (isLaunching && ts.seconds == 0 && ts.nanos == 0) return '-';
    return DateFormat('yyyy-MM-dd HH:mm:ss').format(ts.toDateTime());
  }
}
```

... and then:

```dart
info.instanceInfo.formattedCreationTime(isLaunching: isLaunching)
```

Sploder12 (approved):
LGTM!

ricab:
Geez!

PR #4090, opened 2025-05-16
[cpp headers] replace header guards with #pragma once

xmkg:
This PR replaces all instances of header guards with much simpler and widely supported `#pragma once`.

MULTI-1995


ricab (commented):
LGTM, but it looks like a bunch of those files have space errors that are tripping up linting. Want to fix them too?

xmkg:
> LGTM, but it looks like a bunch of those files have space errors that are tripping up linting. Want to fix them too?

Done!

xmkg:
> Hey @xmkg, I am afraid this will need another go 💦

No worries. I anticipated friction between the line-length branch and this one. Rebased to main, let's see if everything's still fine.

xmkg:
> But, you missed MULTIPASS_BLUEPRINTS_TEST_LAMBDAS_H ;)

Nah, that doesn't count... That's a source guard -- protecting against double inclusion of a source file. This PR is header guards only...

Just joking :stuck_out_tongue: Probably somebody copied this from a header and forgot to remove the guard. I'll remove it :) Good catch!

xmkg:
> Great work! Neat to see this fixes some cases of bad header guards and missing copyrights too. But, you missed `MULTIPASS_BLUEPRINTS_TEST_LAMBDAS_H` ;)

Done!

Sploder12:
> Nah, that doesn't count... That's a source guard -- protecting against double inclusion of a source file. This PR is header guards only...

I completely missed that was from a `.cpp`, interesting...

epicseven-cup:
I like this pr

PR #4104, opened 2025-05-21
Convert all the snake case name of test cases into camel cases

georgeliao:


ricab (commented):
Hey Jia, we've thus far been starting individual test cases with lower case. Apart from that, I'd be good with this (provided CI is too).

xmkg:
> _ No description provided. _

Hey @georgeliao, thanks for the great work!

As my memory serves, you were writing a script to do this. Maybe we can add that script to do the linting as well -- if that's not hard to do, of course. What do you think?

georgeliao:
@xmkg 
I will share the script soon, do you mean add the python script to the multipass repo or just shared it via https://paste.ubuntu.com/?

georgeliao:
@xmkg 
I can add that as well, but maybe in a different PR, because the regex is different. The current python script is about matching the test case name whose format is snake case and replace that with transformed name. Meanwhile, the linter script is about checking whether the captured test case name is compliant with camel case format.

tests/linux/test_syslog_logger.cpp
```diff
@@ -46,7 +46,7 @@ struct syslog_logger_test : ::testing::Test
     MockSyslogWrapper& mock_syslog = *mock_syslog_guardedmock.first;
 };
 
-TEST_F(syslog_logger_test, call_log)
+TEST_F(syslog_logger_test, callLog)
```
xmkg, line 49: Idem

tests/test_sftp_utils.cpp
```diff
@@ -50,7 +50,7 @@ struct SFTPUtils : testing::Test
     fs::path target_path = "target/path";
 };
 
-TEST_F(SFTPUtils, get_full_local_file_target__target_is_dir_child_is_not)
+TEST_F(SFTPUtils, get_full_local_file_target_target_is_dir_child_is_not)
```
xmkg, line 53: This file is still snake case :snake:

xmkg: ... probably it has sth to do with the double underscore @georgeliao

georgeliao: hmm, an oversight of mine, I have noticed the double underscore thing. And what I tried to do was to replace the __ to _ and run the script again. However, apparently, I forgot to run the script.

ricab:
Hey @georgeliao, do you want to fix the test names in this PR as well, or is it ready for another review?

Also, could you please attach the script here or somewhere else that's permanent? The pastebin link is set to expire in a month.

georgeliao:
> do you want to fix the test names in this PR as well, or is it ready for another review?
Which exact name? test suite name or test case name? I think both were fixed.

georgeliao:
> Also, could you please attach the script here or somewhere else that's permanent? The pastebin link is set to expire in a month.

The PR intro is updated, it is pasted there which should stay around much longer.

ricab:
@georgeliao, no urgency, but just so you know this is blocked getting into the merge queue.

xmkg (approved):

PR #4105, opened 2025-05-21
[tics] fix tics action name

sharder996:
Noticed that this didn't run over the weekend and saw that this might be the issue.

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -1,9 +1,9 @@
 name: TICS
 
 on:
-    schedule:
-    # 0:00 UTC every Sunday
-    - cron: 11 4 * * 0
+  schedule:
+  # 0:00 UTC every Sunday
+  - cron: 11 4 * * 0
 
 jobs:
   build:
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -1,4 +1,4 @@
-name: Linux
+name: TICS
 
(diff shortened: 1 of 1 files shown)


ricab (approved):
Ah, right. LGTM, thanks!

sharder996:
@ricab The run failed on [`build-params`](https://github.com/canonical/multipass/actions/runs/15092271045/job/42422443949) because it's a cron job and not from a PR or release. I removed that step from the job, but I'm not sure if the build still needs the build label.

ricab:
Makes sense @sharder996, but does the version get recorded anywhere? I suppose it would look something like "1.16.0-dev.3683+g75321d1a"? It would be good to have a way to link back to the commit or build.

sharder996:
@ricab, one should be able to see the relevant commit in the CI run, no? For example, the previously mentioned failed [CI run](https://github.com/canonical/multipass/actions/runs/15092271045) used commit [585bdd3](https://github.com/canonical/multipass/commit/585bdd39a3e7c94cce66116d391d1876413708f0).

ricab:
Sure, I was rather wondering how to recover the trail back from TICS results? It's probably the last successful TICS run, but I wouldn't mind having a way to verify that in the results. Anyway, probably off-topic here...

sharder996:
Merging in order to hopefully get a successfully TICS run in tomorrow.

PR #4106, opened 2025-05-21
Drop QCOW2 version conversion on existing instances

ricab:
As previously planned.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -199,19 +199,6 @@ auto generate_metadata(const QStringList& platform_args, const QStringList& proc
     return metadata;
 }
 
-void convert_to_qcow2_v3_if_necessary(const mp::Path& image_path, const std::string& vm_name)
-{
-    try
-    {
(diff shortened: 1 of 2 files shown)


tests/qemu/test_qemu_img_utils.cpp
```diff
@@ -148,6 +165,29 @@ void test_image_conversion(const char* img_path, const char* expected_img_path,
     EXPECT_EQ(process_count, expected_final_process_count);
 }
 
+template <class Matcher>
+void test_image_amendment(const char* img_path,
```
ricab, line 169: This is not how I would normally have done these tests, if from scratch. However, I didn't feel like modifying this whole file for this, so I just followed the existing approach.

Sploder12 (commented):
Looking good! Has some conflicts and a new test isn't camelCase

georgeliao:
It looks good to me. Just one clarifying question to make sure I got the goal of the PR right, `mp::backend::amend_to_qcow2_v3` should still be called in `mp::QemuVirtualMachineFactory::prepare_source_image` function as a way of converting the initial image, right?

ricab:
Hey @georgeliao, that's right. We still want the new version. This just removes conversion for existing instances, which should all have been converted by now.

Issue #3967, opened 2025-03-05
SSH sessions getting dropped on macOS


ricab:
> Multipass also became unusable for me after upgrading to 1.15.0 on Sequoia M1. 
> 
> First any SSH session to VM will be lost after some time, ranging from seconds to 1-2 mins. 
> 
> The logs is filled infinitely with 
> 
> `[2025-03-05T22:06:34.233] [info] [enduring-coonhound] SSH session disconnected
> [2025-03-05T22:06:34.233] [debug] [enduring-coonhound] Caching new SSH session
> [2025-03-05T22:06:36.214] [info] [enduring-coonhound] SSH session disconnected
> [2025-03-05T22:06:36.214] [debug] [enduring-coonhound] Caching new SSH session
> [2025-03-05T22:06:38.221] [info] [enduring-coonhound] SSH session disconnected
> [2025-03-05T22:06:38.222] [debug] [enduring-coonhound] Caching new SSH session
> [2025-03-05T22:06:40.224] [info] [enduring-coonhound] SSH session disconnected
> [2025-03-05T22:06:40.224] [debug] [enduring-coonhound] Caching new SSH session
> [2025-03-05T22:06:42.232] [info] [enduring-coonhound] SSH session disconnected
> [2025-03-05T22:06:42.233] [debug] [enduring-coonhound] Caching new SSH session
> [2025-03-05T22:06:44.239] [info] [enduring-coonhound] SSH session disconnected
> `
> 
> and the UI keeps showing "waiting for demon ..."
> 
> I'm thinking go back to try VirtualBox now ... 

 _Originally posted by @dohoangkhiem in [#3721](https://github.com/canonical/multipass/issues/3721#issuecomment-2701218364)_

ricab:
Hi @dohoangkhiem, your case appears to be different from the other one. Would you mind providing a few more details? 

- logs (Please provide logs from the daemon, see [accessing logs](https://canonical.com/multipass/docs/accessing-logs) on where to find them on your platform.)
- any steps you to reproduce the issue?
- macOS version
- `multipass get local.driver`
- have you tried https://github.com/canonical/multipass/releases/tag/v1.15.1 ?

kedmison:
Sorry for jumping in on this issue, but I'm experiencing it too, on MacOS Sequoia 15.3.1.
```
 % sudo tail -10  /Library/Logs/Multipass/multipassd.log
[2025-03-06T15:28:31.674] [info] [primary] SSH session disconnected
[...]

PR #4107, opened 2025-05-21
[vm] Drop transition safety net for runtime errors

ricab:
Remove safety-net catch for generic runtime errors when waiting for SSH.


xmkg:
> Remove safety-net catch for generic runtime errors when waiting for SSH.

Hi @ricab,

Is this related to https://github.com/canonical/multipass/issues/3967?

ricab:
Hey @xmkg, I don't see an obvious connection. If we entered the generic catch, we would log a warning. I don't remember ever seeing such a warning, so I thought we were ready to let this go.

xmkg (approved):

levkropp (approved):
LGTM, we're not throwing any std::runtime_errors in anything related to SSH (we are for SFTP, but that is irrelevant here) so removing this catch should not have any issues from what I can see!

Issue #3054, opened 2023-04-25
[cli] When everything else goes well, `multipass shell` should return the exit code of the inner shell 


ricab:
1. `multipass shell`
2. `exit 1` (inside the instance)
3. `echo $?` says 0.

FWIW, `multipass exec` DTRT.

Artem-OSSRevival:
In the commit above was implemented the bug fix, which allows to return the 'exit_code' value from the shell in the 'ReturnCode' range.

Artem-OSSRevival:
To avoid misunderstandings about the commits above. The PR was also created by me, just from a personal account...

jimporter:
This is partially fixed now due to #4109, but when the shell exits with a non-zero status, multipass always returns 2. As part of #4428, we'll look at this some more to ensure that we handle exit statuses the same across multipass.

PR #4109, opened 2025-05-23
Added the return of the inner shell exit code

artemvlas:
Closes: #3054
Previously, with a successful exit from the instance’s shell, the exit code was always 0. Now the shell returns integer 'exit_code'. If this value goes beyond the 'ReturnCode' range (0-3), will be returned the default 0 (ReturnCode::Ok).

**To reproduce:**
Start service:
`sudo /usr/local/bin/multipassd`

Open another terminal/tab. List of existing instances:
`multipass ls`

Start some instance:
`multipass start 'instance_name'`

Switch to instance’s shell:
`multipass shell 'instance_name'`

Exit the instance's shell with some exit code:
`exit 1`

Try to get this code from the main shell:
`echo $?`

**Before:**
get result: `0`

**After:**
get result: `1` (custom exit code value in 0-3 range)

### DIFF
--- a/include/multipass/ssh/ssh_client.h
+++ b/include/multipass/ssh/ssh_client.h
@@ -47,7 +47,7 @@ public:
 
     int exec(const std::vector<std::string>& args);
     int exec(const std::vector<std::vector<std::string>>& args_list);
-    void connect();
+    int connect();
 
 private:
     void handle_ssh_events();
--- a/src/client/cli/cmd/shell.cpp
+++ b/src/client/cli/cmd/shell.cpp
@@ -76,7 +76,10 @@ mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)
         {
             auto console_creator = [this](auto channel) { return term->make_console(channel); };
             mp::SSHClient ssh_client{host, port, username, priv_key_blob, console_creator};
-            ssh_client.connect();
+            const int exit_code = ssh_client.connect();
+
+            if (exit_code != 0)
+                return ReturnCode::CommandFail;
         }
         catch (const std::exception& e)
         {
--- a/src/ssh/ssh_client.cpp
+++ b/src/ssh/ssh_client.cpp
@@ -63,9 +63,9 @@ mp::SSHClient::SSHClient(SSHSessionUPtr ssh_session, ConsoleCreator console_crea
 {
 }
 
-void mp::SSHClient::connect()
+int mp::SSHClient::connect()
 {
-    exec(std::vector<std::string>{});
+    return exec(std::vector<std::string>{});
 }
 
 int mp::SSHClient::exec(const std::vector<std::string>& args)


sharder996 (changes requested):
This is a good change, but I believe you are referenced the wrong error codes. `mp::ReturnCode` won't be returned here, but a code from the libssh library. I believe [this](https://github.com/canonical/libssh/blob/f23d1454e50d0dbb314edd9bf4227ab72303484b/src/channels.c#L3484) is the function that will be returning the code you are interested in.

artemvlas:
@sharder996, the `on_success` lambda should return the _ReturnCode_ value (see the **while loop** below) as well as `cmd::Shell::run` function itself.

sharder996:
Ok, I understand a bit better now. It doesn't make sense to me that we are comparing ssh return codes to our own internal enum, but that issue can be addressed elsewhere. However, this the issue that I see:

```
% multipass exec foo -- exit 127
% echo $?                         
127
% multipass sh foo
ubuntu@foo:~$ exit 127
logout
% echo $?                       
0
```

Looking at `exec.cpp`, you got the right idea in returning an ssh code from `SSHClient::connect()`, but I think this condition; `if (exit_code >= ReturnCode::Ok && exit_code < ReturnCode::Retry)` in `shell.cpp` should be a bit less restrictive. Because we're just trying to avoid the infinite loop, `ReturnCode::Retry` is the only code we care about. Hence why all other codes are being ignored. You just need to be careful with comparing an signed int to an unsigned int.

BTW, `% multipass exec foo -- exit 4` goes into an infinite loop, as well as exiting with any signed int that converts to unsigned int{4}; i.e. `-252`.

To truly fix this, we should be able to set any code from ssh without having to worry about our own internal `ReturnCode`.

artemvlas:
And what about the type returned by `cmd::Shell::run`? The function should return the `ReturnCode`  value. If we put the arbitrary int value as the enum, this can lead to other errors.
As far as I understand, you propose to remove the condition line
`if (exit_code >= ReturnCode::Ok && exit_code < ReturnCode::Retry)`
In this case, it will be possible to cast any number, regardless of existing enum values. But this is quite dangerous in terms of possible errors... Really! Forced casting of an arbitrary number (with no check of the existing enum values) to the enum is usually not good itself.

But if these possible errors (now or in the future) due to the enum values range do not worry you, we can, of course, remove this condition line. Of course, the return of any number, regardless of the enum values, would be much more convenient for the user...

artemvlas:
If the `cmd::Shell::run`  return type was **int**, then there would be no problems at all.

sharder996:
Ok, the issue I see is that a single return variable from `mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)` is being used to return an exit code from more than one source. The first is the `mp::ReturnCode` exit code which is being used to convey information about how Multipass handled the command (`multipass start`, `multipass launch`, etc). The second is from `int mp::SSHClient::connect()` which is being used to convey information about what went on inside the ssh session. The issue arises when we apply logic meant for `mp::ReturnCode` against the return from `int mp::SSHClient::connect()` [here](https://github.com/canonical/multipass/pull/4109/files#diff-8f88aafd4921814658b6948787e53ccabc58fe34aa00ee3356988ef78a35ba68R83-R84).

Now, to truly fix this properly would greatly increases the scope as we would have to find a clean way conveying both of these pieces of information separately and appropriately set the return type of the Multipass CLI command. At this point, I am willing to make a compromise that has `int mp::SSHClient::connect()` being returned through `mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)` with the exception of `ReturnCode::Retry` (since that causes an infinite retry loop) and creating a more generalized solution at a later point in time.

artemvlas:
@ricab, did I understand correctly?

```
try
{
    ...
    const int exit_code = ssh_client.connect();

    if (exit_code != 0)
        return ReturnCode::CommandFail;
}
catch (...
```

ricab:
Hi @artemvlas, I think that'a about the best we can do without changing the return type.

ricab:
This is looking good to me now @artemvlas, thanks! I am waiting for CI before actually approving. 

One suggestion would be to log a warning when exec returns something other than 0, to avoid loosing the information entirely. But it doesn't need to be in this PR.

ricab:
Converting to a draft and back, to see if that kicks CI.

ricab (approved):
Approving on secondary. Thanks again @artemvlas.

Note that CI passed everything that did not require a secret.

sharder996 (approved):
Yup, looks good.

@ricab Will you happen the process of merging?

sharder996:
@artemvlas, since your PR came from an external repository, it cannot fully run our CI for security reasons. I have taken your commits and put them on a branch in this repository, while of course maintaining the authorship in your name. It can be found at #4266.

Thank you again for your contribution! It is highly appreciated!

PR #4115, opened 2025-05-27
Update daemon permissions

Sploder12:
With our recent changes to permissions, it doesn't make sense for the daemon binary to not have similar restrictions applied to it. This PR applies those restrictions by making the daemon owned by root similar to how the storage directory is now owned by root.

MULTI-1998


sharder996 (approved):
+1 from me. Tested and verified.

PR #4119, opened 2025-05-28
[tests/simple_streams] add test case for no image-downloads

xmkg:
It is causing spurious coverage-related CI failures, such as: https://app.codecov.io/gh/canonical/multipass/pull/4090/indirect-changes

MULTI-2005


ricab (approved):
LGTM

PR #4122, opened 2025-05-28
[format] remove some redundant clang-format off/on blocks

xmkg:


ricab (approved):
Yeah clang-format doesn't need these anymore. Well spotted.

PR #4125, opened 2025-05-30
[smb_mount_handler] normalize share names

xmkg:
The share name right now uses UUID of the target, then vm_name:path which might contain characters that are not allowed in a samba share name. This patch fixes that by forming the share name from two UUID's.

### DIFF
--- a/src/platform/backends/shared/windows/smb_mount_handler.cpp
+++ b/src/platform/backends/shared/windows/smb_mount_handler.cpp
@@ -193,11 +193,9 @@ SmbMountHandler::SmbMountHandler(VirtualMachine* vm,
     : MountHandler{vm, ssh_key_provider, std::move(mount_spec), target},
       source{QString::fromStdString(get_mount_spec().get_source_path())},
       // share name must be unique and 80 chars max
-      share_name{QString("%1_%2:%3")
-                     .arg(MP_UTILS.make_uuid(target),
-                          QString::fromStdString(vm->vm_name),
-                          QString::fromStdString(target))
-                     .left(80)},
+      // UUIDS are 36 chars each, and +1 for dash: 73 characters.
+      share_name{QString::fromStdString(
+          fmt::format("{}-{}", MP_UTILS.make_uuid(vm->vm_name), MP_UTILS.make_uuid(target)))},
       cred_dir{cred_dir},
       smb_manager{&smb_manager}
 {


ricab (approved):
Yep, makes sense. Thanks! (secondary only on this one.)

xmkg:
@Sploder12 @sharder996 could I interest any of you to be the primary reviewer :)

Sploder12:
>could I interest any of you to be the primary reviewer :)

Code looks good to me, but I can't get SMB mounts working at all even without these changes

dmesg:
```
CIFS: Attempting to mount //trevor-shoe/ac90fd5e-81d8-3916-bb7a-55565a60004a-30e9403d-018e-39ad-91ba-2ff46d202c11
CIFS: Status code returned 0xc000006d STATUS_LOGON_FAILURE
CIFS: VFS: \\trevor-shoe Send error in SessSetup = -13
CIFS: VFS: cifs_mount failed w/return code = -13`
```

CLI:
```
error mounting "/home/ubuntu/mnt": Error: mount error(13): Permission denied
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs) and kernel log messages (dmesg)
```

xmkg:
> > could I interest any of you to be the primary reviewer :)
> 
> Code looks good to me, but I can't get SMB mounts working at all even without these changes
> 
> dmesg:
> 
> ```
> CIFS: Attempting to mount //trevor-shoe/ac90fd5e-81d8-3916-bb7a-55565a60004a-30e9403d-018e-39ad-91ba-2ff46d202c11
> CIFS: Status code returned 0xc000006d STATUS_LOGON_FAILURE
> CIFS: VFS: \\trevor-shoe Send error in SessSetup = -13
> CIFS: VFS: cifs_mount failed w/return code = -13`
> ```
> 
> CLI:
> 
> ```
> mount failed: The following errors occurred:
> error mounting "/home/ubuntu/mnt": Error: mount error(13): Permission denied
> Refer to the mount.cifs(8) manual page (e.g. man mount.cifs) and kernel log messages (dmesg)
> ```

Hmmm, interesting. Which Windows version?

Sploder12:
> Hmmm, interesting. Which Windows version?

Windows 11 Pro with Hyper-V backend

xmkg:
> > Hmmm, interesting. Which Windows version?
> 
> Windows 11 Pro with Hyper-V backend

Circling back on this, it didn't work for me out of the box as well. I'll debug it to see what's really going on.

xmkg:
> > Hmmm, interesting. Which Windows version?
> 
> Windows 11 Pro with Hyper-V backend

It turns out that I was mistakenly entering my PIN instead of my account password while trying to mount :man_facepalming: When I used my actual password, it worked.

Sploder12:
> It turns out that I was mistakenly entering my PIN instead of my account password while trying to mount 🤦‍♂️ When I used my actual password, it worked.

I'm getting the same issue even with my account password

xmkg:
> > It turns out that I was mistakenly entering my PIN instead of my account password while trying to mount 🤦‍♂️ When I used my actual password, it worked.
> 
> I'm getting the same issue even with my account password

Hmmm.. that's weird. We need some logs from Windows SMB to see what's really going on.

xmkg:
> > > It turns out that I was mistakenly entering my PIN instead of my account password while trying to mount 🤦‍♂️ When I used my actual password, it worked.
> > 
> > 
> > I'm getting the same issue even with my account password
> 
> Hmmm.. that's weird. We need some logs from Windows SMB to see what's really going on.

@Sploder12, could you capture the packets between the SMB server and the client? Perhaps the packets would provide a clue about the problem.

sharder996:
I tested native mounts both here and on `main` and everything works as expected. No "permission denied" issues. On Windows 10 Pro, btw.

Sploder12:
> @Sploder12, could you capture the packets between the SMB server and the client? Perhaps the packets would provide a clue about the problem.

From the packets it is definitely failing on logon. I made a VM to avoid sharing my Microsoft password, and after setting a password (Multipass forces some password even if the user has no password), it worked fine. My guess is that there are some additional restrictions when using a Microsoft account as your Windows account that are interfering.

sharder996 (approved):
Considering the past few comments, I'm giving this an approval.

xmkg:
> > @Sploder12, could you capture the packets between the SMB server and the client? Perhaps the packets would provide a clue about the problem.
> 
> From the packets it is definitely failing on logon. I made a VM to avoid sharing my Microsoft password, and after setting a password (Multipass forces some password even if the user has no password), it worked fine. My guess is that there are some additional restrictions when using a Microsoft account as your Windows account that are interfering.

@Sploder12 that's weird, at least it points to an authentication problem rather than a bug in our implementation. Did you try `main` in the new VM? Does it work? (it shouldn't).

> I tested native mounts both here and on `main` and everything works as expected. No "permission denied" issues. On Windows 10 Pro, btw.

@sharder996, then the problem should've started happening in one of the patches that Windows 11 had for its SMB implementation. One curious update is this one: https://techcommunity.microsoft.com/blog/filecab/smb-security-hardening-in-windows-server-2025--windows-11/4226591

On the other hand, I dug this up from Microsoft's spec on [File System Control Codes] (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/efbfe127-73ad-4140-9967-ec6500e66d5e):

```
2.1.6 Share name
A share name has the following characteristics:
- A share name MUST be no more than 80 characters in length.
- The following characters are illegal in a share name:
" \ / [ ] : | < > + = ; , * ?
- Control characters in range 0x00 through 0x1F, inclusive, are illegal in a share name.
- All other Unicode characters are legal.
```

So, appending a UNIX pathname to a UUID forms an invalid share name according to these criteria. They probably have started enforcing this in Windows 11 and Windows Server 2025, and that could be the reason why the `main` branch is still working on Windows 10.

Sploder12:
> @Sploder12 that's weird, at least it points to an authentication problem rather than a bug in our implementation. Did you try `main` in the new VM? Does it work? (it shouldn't).

`main` gives me a file doesn't exist error but packets confirm it's the invalid share name issue

xmkg:
> > @Sploder12 that's weird, at least it points to an authentication problem rather than a bug in our implementation. Did you try `main` in the new VM? Does it work? (it shouldn't).
> 
> `main` gives me a file doesn't exist error but packets confirm it's the invalid share name issue

@Sploder12 The permission issue looks unrelated to the changes introduced in this patch. Can we create a separate issue for it to unblock and land this?

PR #4126, opened 2025-05-30
Harden bash completion security a little

ricab:
Improve the security of our bash completion in some respects, most notably by avoiding evaluating function input. See commit messages for further details.

MULTI-2010


levkropp (approved):
Yep, completions are working and listing instances for `stop`, `snapshot`, keys, etc!

PR #4128, opened 2025-05-30
[utils] remove unused function

sharder996:


ricab (dismissed):
Looks good, thanks Scott! CI agrees we don't need it.

georgeliao:
Thanks for the initiative. Can we also remove other unused ones in that group like `remove_directories` and `filename_for`?

PR #4135, opened 2025-06-02
[ci] dynamic ci

sharder996:
This PR revamps how our GitHub workflows are invoked so that only the necessary workflows are run and we avoid situations where a change to the `README.md` gets blocked on an Apple runner failing to compile Multipass due to a spurious seg fault.

This setup works by having a single orchestration workflow which filters the changed files with [dorny/paths-filter](https://github.com/dorny/paths-filter), defines relevant outputs, and then spawns the relevant workflows. A "Finalize" job in the orchestration workflow then checks the return statuses of the spawned workflows and sets it own return status accordingly.

Regarding branch protection rules, the idea is that we would protect on `Dynamic-CI \ Filter changed files` and `Dynamic-CI \ Finalize` to make sure that the PR is properly analyzed and and errors from spawned workflows are propagated up, effectively gating the PR from being merged.

Below are some example draft PRs demonstrating various scenarios and how the new CI setup reacts to them.

1. Documentation changes (#4136)
- https://github.com/canonical/multipass/actions/runs/15478104252/job/43578947769?pr=4136 demonstrates a run that has been cancelled.
2. Documentation changes that fail to build (#4143)
3. Source code changes (#4137)
4. Changes that are neither documentation or source code (#4142)
5. Changing both documentation and code, but having one fail (#4144)

Merge queues aren't officially supported in the `dorny/paths-filter` action, but there are several user submitted examples of people using the action in this manner. Because of repository permissions this has yet to have been tested.

Another GitHub action; [tj-actions/changed-files](https://github.com/tj-actions/changed-files) was considered for added functionality of being able to spawn workflows. However, I chose not to go with it due to it's increased complexity and granularity of control that isn't strictly needed in our setup.

One other thing, `docs/readthedocs.com:canonical-multipass` will continue to be built for every PR regardless of changed files. The trigger for this workflow is managed by [ReadtheDocs](https://app.readthedocs.com/dashboard/canonical-multipass/pull-requests/) which does not offer the granularity or control of being able to selectively control when the workflow is run.

### Future development

One slight improvement that can be made to further optimize CI would be to bisect `windows-macos.yml` and further filter on files that only affect Windows or macOS builds. However, care would need to be taken in order to ensure that there is no unnoticed cross-platform dependencies in these filtered files.

Closes #4136 
Closes #4143 
Closes #4137 
Closes #4142
Closes #4144 

---

MULTI-1932

---

### DIFF
--- a/.github/workflows/dynamic-ci.yml
+++ b/.github/workflows/dynamic-ci.yml
@@ -74,12 +74,14 @@ jobs:
     needs: filter
     if: needs.filter.outputs.code == 'true'
     uses: "./.github/workflows/linux.yml"
+    secrets: inherit
 
   windows-macos:
     name: Windows/macOS
     needs: filter
     if: needs.filter.outputs.code == 'true'
     uses: "./.github/workflows/windows-macos.yml"
+    secrets: inherit
 
   finalize:
     name: Finalize


ricab:
Addresses #4042.

levkropp:
The names for these checks are quite long now, maybe instead of
#### Dynamic CI / Windows/macOS / BuildAndTest (multipass-macos-12-x64, x86_64, 12)
we consolidate it down to 
#### CI / Windows/macOS / BuildAndTest (macos-12, x86_64)
or even drop the first CI / entirely

sharder996:
@levkropp "(multipass-macos-12-x64, x86_64, 12)" comes from the name of our self-hosted runners. Only way to change this is to rename the runners. Since they're going away soon, I don't feel obliged to change it.

"Dynamic CI" can't be dropped entirely because GitHub will then use the name of the file, making it even longer. I can change it to "CI" to make it shorter if you like. Other suggestions are welcome as well.

levkropp:
> "Dynamic CI" can't be dropped entirely because GitHub will then use the name of the file, making it even longer. I can change it to "CI" to make it shorter if you like.

Yeah just "CI" would be nice, I think. I'm interested in hearing other people's opinions/suggestions as well!

Is it possible to remove (pull_request) being appended to every CI job's name? We used to not have that previously either

sharder996:
The appended "(pull request)" is because the CI setup is now using reusable workflows which are being labelled with the corresponding triggering context. In merge queues, they will be reflected differently. As far as I know, there is no way to change this.

.github/workflows/dynamic-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+        with:
+          # 'base' and 'ref' are ignored for pull requests
+          # For merge group events, compare against the target branch (main)
+          base: ${{ github.event.merge_group.base_ref }}
+          # For merge group events, use the merge group head ref
+          ref: ${{ github.event.merge_group.head_ref }}
+          filters: |
+            docs:
+              - '.github/workflows/automatic-doc-checks.yml'
+              - 'docs/**'
```
xmkg, line 40: Should we also include the markdown files here?

.github/workflows/dynamic-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+            code:
+              - ".github/workflows/linux*"
```
xmkg, line 42: Would it make sense to remove the code filter and use !doc in the non-doc workflow dispatch calls? My concern here is that we might forget updating this as time passes, especially on files that have a non-direct influence on the outcome.

.github/workflows/dynamic-ci.yml
sharder996: The markdown files (`README.md`, `BUILD.linux.md`, etc) don't get processed/linted with `.github/workflows/automatic-doc-checks.yml`.

.github/workflows/dynamic-ci.yml
sharder996: The disadvantage of doing that is that CI would build Multipass for changes to unrelated files such as `renovate.json5`. We could add another filter for these *meta*-files and call somethings like `!doc && !meta`, but then we would have to remember to maintain that list of files instead of a list of code paths. The `code` directories are all top-level directories so in theory they shouldn't be changing often.

.github/workflows/dynamic-ci.yml
xmkg: Right, and right now, a change in a .md file won't trigger any workflow. Makes sense.

.github/workflows/dynamic-ci.yml
xmkg: After thinking about it again, it wouldn't be hard to notice that the required workflow hasn't run for a PR where it should.

xmkg:
We'll probably need to merge this with admin rights, and then update the main's branch protection rules to require Dynamic CI instead.

sharder996:
Yup, I want to check with @ricab before merging to make sure that we don't need CI for the 1.16 release since CI might be down for a bit while this setup is verified and the branch protection rules are changed.

PR #4145, opened 2025-06-06
[feat] Add wait-ready command to check Multipass daemon readiness

### LINKED ISSUE #452: Command suggestion: waitready
In tests for the snapcraft CLI, we install multipass and then run snapcraft builds. There's a bit of a race condition here where if multipass isn't completely fired up by the time the build starts, snapcraft errors out because multipass can't reach its daemon. LXD solved this issue by introducing a `waitready` command that we have utilized in our tests. Can we get similar functionality in multipass?

### LINKED ISSUE #3723: Error "launch failed: Remote "release" is unknown or unreachable. If image mirror is enabled, please confirm it is valid." when launching image too quickly after installing Multipass
**Describe the bug**
I am encountering the following error when trying to launch an image too quickly after installing Multipass.
```
launch failed: Remote "release" is unknown or unreachable. If image mirror is enabled, please confirm it is valid.
```
This is occurring on a self-hosted Github runner that supports nested virtualisation.
 
**To Reproduce**
How, and what happened?
1. `sudo snap install multipass``multipass ...`
2. `multipass launch 16.04 --name test`

When adding a 10s delay between steps 1 and 2 the issue does not occur. 

**Expected behavior**
For the image to be able to launch after the snap install was completed.

rluna319:
**Feature: Add `multipass wait-ready` command**

**Description:**
This PR introduces a new CLI command, `multipass wait-ready`. This command allows users and scripts to pause execution until the Multipass daemon is fully initialized and ready to accept requests. This is particularly beneficial for automation workflows that need to ensure the daemon is operational before proceeding with other Multipass operations.

The `wait-ready` command works by:
1.  Initially polling the daemon's gRPC socket until it becomes available.
2.  Once the socket is available, it sends a `WaitReady` request to the daemon.
3.  The daemon, upon receiving this request, verifies its readiness by attempting to update image manifests (which implicitly checks connectivity to image servers).
4.  The command supports a `--timeout` option (in seconds). If the daemon isn't ready within the specified timeout, the command exits with an error. If no timeout is given, it waits indefinitely.
5.  An animated spinner and context-relevant message provide user feedback during the waiting period.

**Motivation:**
- Improves the reliability of automated scripts and CI/CD pipelines that interact with Multipass by providing a deterministic way to wait for daemon initialization.
- Addresses scenarios where subsequent Multipass commands might fail if the daemon is still starting up.

**Related Issues:**
Fixes #452
Fixes #3723

**Key Implementation Changes:**
- **Client-side (`wait_ready.cpp`):** New command logic, including retry for socket connection, timeout handling (reusing `mp::cmd::parse_timeout`), and spinner integration.
- **Protocol (`multipass.proto`):** Added `WaitReadyRequest` / `WaitReadyReply` messages and `WaitReady` RPC method.
- **Daemon-side (`daemon_rpc.cpp`, `daemon.cpp`):**
    - New gRPC service handler in `daemon_rpc.cpp`.
    - Core readiness check in `daemon.cpp` leverages the existing `wait_update_manifests_all_and_optionally_applied_force()` function.

**How to Test:**
1.  Build this branch and `$ cd build`.
2.  Ensure the Multipass daemon is stopped.
3.  To verify `--timeout`, simply run the `wait-ready` command with the `--timeout` option:
```Bash
$ ./bin/multipass wait-ready --timeout 5
```
> Be sure not to start up the daemon first or the command might exit before the timeout.
- Expected: Info message "Waiting for Multipass daemon to be ready...", spinner is shown, command times out after ~5 seconds with an appropriate error message "Timed out waiting for Multipass daemon to be ready."
4. (Optional) To replicate the issue that happens when the client tries to use the daemon before its ready (without `wait-ready`), start up the daemon and immediately use `multipass launch`:
```Bash
$ sudo -v
$ (sudo ./bin/multipassd &) && ./bin/multipass launch
```
- Expected: Error message "launch failed: cannot connect to the multipass socket", "launch failed: Remote "" is unknown or unreachable.", or something like issue #3723
5.  Stop the Multipass daemon
```Bash
$ sudo pkill multipassd
```
6. To verify `wait-ready` command solves the issue, fit the `wait-ready` command between daemon startup and `multipass launch`:
```Bash
$ sudo -v
$ (sudo ./bin/multipassd &) && ./bin/multipass wait-ready && ./bin/multipass launch
```
> Optionally, add `--timeout` if desired.
- Expected: Info message "Waiting for Multipass daemon to be ready...", spinner is shown, command exits successfully (ReturnCode::Ok) once the daemon is ready (i.e., can connect to image servers) and the client begins retrieving the default _release_ image for the `multipass launch` process. _You may Ctrl+C here to terminate the launch process._

**Points for Reviewers:**
- Confirmation that using `wait_update_manifests_all_and_optionally_applied_force(true)` is an appropriate and robust proxy for overall daemon readiness.
- Feedback on the client-side error messages and user experience (spinner, timeout message) is appreciated. The current spinner implementation addresses @ricab's suggestion in #452.
- **Unit Tests:** Per @ricab 's comments in issue #452, unit tests have not yet been implemented for this initial PR to facilitate earlier review. I am prepared to add them based on feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ cmake-build-*
 .DS_Store
 CMakeLists.txt.user
 build
+Testing
 
 # package build artifacts
 *.deb
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ cmake-build-*
 .DS_Store
 CMakeLists.txt.user
 build
-Testing
 
 # package build artifacts
 *.deb
--- a/src/client/cli/cmd/wait_ready.cpp
+++ b/src/client/cli/cmd/wait_ready.cpp
@@ -17,16 +17,19 @@
 
 #include "wait_ready.h"
 #include "common_cli.h"
-
 #include "animated_spinner.h"
+
+#include <multipass/timer.h>
+#include <multipass/constants.h>
 #include <multipass/exceptions/cmd_exceptions.h>
 #include <multipass/cli/argparser.h>
 #include <multipass/cli/formatter.h>
-#include <multipass/timer.h>
+
+#include <chrono>
+#include <thread>
 
 namespace mp = multipass;
 namespace cmd = multipass::cmd;
-namespace mpl = multipass::logging;
 
 mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
 {
@@ -35,29 +38,29 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
     {
         return parser->returnCodeFrom(ret);
     }
+    
+    mp::AnimatedSpinner spinner(cout);
+    spinner.start("Waiting for Multipass daemon to be ready");
 
-    if (!spinner){
-        spinner = std::make_unique<multipass::AnimatedSpinner>(cout);
-        spinner->start("Waiting for Multipass daemon to be ready...");
-    }
+    std::unique_ptr<mp::utils::Timer> timer;
 
     // If the user has specified a timeout, we will create a timer
     if (parser->isSet("timeout")){
         timer = cmd::make_timer(parser->value("timeout").toInt(),
-                                spinner.get(),
+                                &spinner,
                                 cerr,
                                 "Timed out waiting for Multipass daemon to be ready.");
         timer->start();
     }
     
-    auto on_success = [this](WaitReadyReply& reply) {
+    auto on_success = [this, &spinner, &timer](WaitReadyReply& reply) {
         if (timer)
             timer->stop();
-        spinner->stop();
+        spinner.stop();
         return ReturnCode::Ok;
     };
 
-    auto on_failure = [this](grpc::Status& status) {
+    auto on_failure = [this, &spinner, &timer](grpc::Status& status) {
 
         if (status.error_code() == grpc::StatusCode::NOT_FOUND && 
             status.error_message() == "cannot connect to the multipass socket")
@@ -70,7 +73,7 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
 
         if (timer)
             timer->stop();
-        spinner->stop();
+        spinner.stop();
         
         // For any other error, we will handle it as a standard failure
         return standard_failure_handler_for(name(), cerr, status);
@@ -100,12 +103,11 @@ QString cmd::WaitReady::short_help() const
 QString cmd::WaitReady::description() const
 {
     return QStringLiteral(
-        "Wait for the Multipass daemon to be ready.\n"
-        "This command will block until the daemon is ready to accept requests.\n"
-        "It can be used to ensure that the daemon is running before executing "
-        "other commands.\n"
-        "If a timeout is specified, it will wait for that duration before "
-        "exiting with an error if the daemon is not ready.");
+        "Wait for the Multipass daemon to be ready. This command will block until the\n"
+        "daemon has initialized, fetched up-to-date image information, and is ready to\n"
+        "accept requests. Its main use is to prevent failures caused by incomplete\n"
+        "initialization in batch operations. An optional timeout aborts the command if\n"
+        "reached.");
 }
 
 mp::ParseCode cmd::WaitReady::parse_args(mp::ArgParser* parser)
--- a/src/client/cli/cmd/wait_ready.cpp
+++ b/src/client/cli/cmd/wait_ready.cpp
@@ -1,0 +1,131 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "wait_ready.h"
+#include "common_cli.h"
+
+#include "animated_spinner.h"
+#include <multipass/exceptions/cmd_exceptions.h>
+#include <multipass/cli/argparser.h>
+#include <multipass/cli/formatter.h>
+#include <multipass/timer.h>
+
+namespace mp = multipass;
+namespace cmd = multipass::cmd;
+namespace mpl = multipass::logging;
+
+mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
+{
+    auto ret = parse_args(parser);
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+
+    if (!spinner){
+        spinner = std::make_unique<multipass::AnimatedSpinner>(cout);
+        spinner->start("Waiting for Multipass daemon to be ready...");
+    }
+
+    // If the user has specified a timeout, we will create a timer
+    if (parser->isSet("timeout")){
+        timer = cmd::make_timer(parser->value("timeout").toInt(),
+                                spinner.get(),
+                                cerr,
+                                "Timed out waiting for Multipass daemon to be ready.");
+        timer->start();
+    }
+    
+    auto on_success = [this](WaitReadyReply& reply) {
+        if (timer)
+            timer->stop();
+        spinner->stop();
+        return ReturnCode::Ok;
+    };
+
+    auto on_failure = [this](grpc::Status& status) {
+
+        if (status.error_code() == grpc::StatusCode::NOT_FOUND && 
+            status.error_message() == "cannot connect to the multipass socket")
+        {
+            // This is the expected state for when the daemon is not yet ready
+            // Sleep for a short duration and signal to retry
+            std::this_thread::sleep_for(std::chrono::milliseconds(500));
+            return ReturnCode::Retry;
+        }
+        // For any other error, we will handle it as a standard failure
+        cerr << "Error Code: " << status.error_code() << "\n"
+             << "Error Message: " << status.error_message() << "\n";
+        return standard_failure_handler_for(name(), cerr, status);
+    };
+
+    request.set_verbosity_level(parser->verbosityLevel());
+    
+    ReturnCode return_code;
+
+    while ((return_code = dispatch(&RpcMethod::wait_ready, request, on_success, on_failure)) == 
+            ReturnCode::Retry)
+        ;
+
+    return return_code;
+}
+
+std::string cmd::WaitReady::name() const
+{
+    return "wait-ready";
+}
+
+QString cmd::WaitReady::short_help() const
+{
+    return QStringLiteral("Wait for the Multipass daemon to be ready");
+}
+
+QString cmd::WaitReady::description() const
+{
+    return QStringLiteral(
+        "Wait for the Multipass daemon to be ready.\n"
+        "This command will block until the daemon is ready to accept requests.\n"
+        "It can be used to ensure that the daemon is running before executing "
+        "other commands.\n"
+        "If a timeout is specified, it will wait for that duration before "
+        "exiting with an error if the daemon is not ready.");
+}
+
+mp::ParseCode cmd::WaitReady::parse_args(mp::ArgParser* parser)
+{
+    mp::cmd::add_timeout(parser);
+
+    auto status = parser->commandParse(this);
+
+    // Check if the command was parsed successfully
+    if (status != ParseCode::Ok)
+    {
+        return status;
+    }
+
+    try
+    {
+        mp::cmd::parse_timeout(parser);
+    }
+    catch (const mp::ValidationException& e)
+    {
+        cerr << "error: " << e.what() << "\n";
+        return ParseCode::CommandLineError;
+    }
+
+    return status;
+}
--- a/src/client/cli/cmd/wait_ready.cpp
+++ b/src/client/cli/cmd/wait_ready.cpp
@@ -67,9 +67,12 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
             std::this_thread::sleep_for(std::chrono::milliseconds(500));
             return ReturnCode::Retry;
         }
+
+        if (timer)
+            timer->stop();
+        spinner->stop();
+        
         // For any other error, we will handle it as a standard failure
-        cerr << "Error Code: " << status.error_code() << "\n"
-             << "Error Message: " << status.error_message() << "\n";
         return standard_failure_handler_for(name(), cerr, status);
     };
 
--- a/src/client/cli/cmd/wait_ready.cpp
+++ b/src/client/cli/cmd/wait_ready.cpp
@@ -16,14 +16,12 @@
  */
 
 #include "wait_ready.h"
-#include "common_cli.h"
 #include "animated_spinner.h"
+#include "common_cli.h"
 
 #include <multipass/timer.h>
-#include <multipass/constants.h>
-#include <multipass/exceptions/cmd_exceptions.h>
 #include <multipass/cli/argparser.h>
-#include <multipass/cli/formatter.h>
+#include <multipass/exceptions/cmd_exceptions.h>
 
 #include <chrono>
 #include <thread>
@@ -38,22 +36,23 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
     {
         return parser->returnCodeFrom(ret);
     }
-    
-    mp::AnimatedSpinner spinner(cout);
-    spinner.start("Waiting for Multipass daemon to be ready");
+
+    mp::AnimatedSpinner spinner{cout};
+    spinner.start("Waiting for the Multipass daemon to be ready");
 
     std::unique_ptr<mp::utils::Timer> timer;
 
     // If the user has specified a timeout, we will create a timer
-    if (parser->isSet("timeout")){
+    if (parser->isSet("timeout"))
+    {
         timer = cmd::make_timer(parser->value("timeout").toInt(),
                                 &spinner,
                                 cerr,
-                                "Timed out waiting for Multipass daemon to be ready.");
+                                "Timed out waiting for the Multipass daemon to be ready.");
         timer->start();
     }
-    
-    auto on_success = [this, &spinner, &timer](WaitReadyReply& reply) {
+
+    auto on_success = [&spinner, &timer](WaitReadyReply& reply) {
         if (timer)
             timer->stop();
         spinner.stop();
@@ -61,9 +60,9 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
     };
 
     auto on_failure = [this, &spinner, &timer](grpc::Status& status) {
-
-        if (status.error_code() == grpc::StatusCode::NOT_FOUND && 
-            status.error_message() == "cannot connect to the multipass socket")
+        if (status.error_code() == grpc::StatusCode::NOT_FOUND &&
+            (status.error_message() == "cannot connect to the multipass socket" ||
+             status.error_message() == "cannot connect to the image servers"))
         {
             // This is the expected state for when the daemon is not yet ready
             // Sleep for a short duration and signal to retry
@@ -74,17 +73,17 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
         if (timer)
             timer->stop();
         spinner.stop();
-        
+
         // For any other error, we will handle it as a standard failure
         return standard_failure_handler_for(name(), cerr, status);
     };
 
     request.set_verbosity_level(parser->verbosityLevel());
-    
+
     ReturnCode return_code;
 
-    while ((return_code = dispatch(&RpcMethod::wait_ready, request, on_success, on_failure)) == 
-            ReturnCode::Retry)
+    while ((return_code = dispatch(&RpcMethod::wait_ready, request, on_success, on_failure)) ==
+           ReturnCode::Retry)
         ;
 
     return return_code;
--- a/src/client/cli/cmd/wait_ready.cpp
+++ b/src/client/cli/cmd/wait_ready.cpp
@@ -19,9 +19,9 @@
 #include "animated_spinner.h"
 #include "common_cli.h"
 
-#include <multipass/timer.h>
 #include <multipass/cli/argparser.h>
 #include <multipass/exceptions/cmd_exceptions.h>
+#include <multipass/timer.h>
 
 #include <chrono>
 #include <thread>
@@ -60,9 +60,7 @@ mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
     };
 
     auto on_failure = [this, &spinner, &timer](grpc::Status& status) {
-        if (status.error_code() == grpc::StatusCode::NOT_FOUND &&
-            (status.error_message() == "cannot connect to the multipass socket" ||
-             status.error_message() == "cannot connect to the image servers"))
+        if (status.error_code() == grpc::StatusCode::NOT_FOUND)
         {
             // This is the expected state for when the daemon is not yet ready
             // Sleep for a short duration and signal to retry
--- a/src/client/cli/cmd/wait_ready.h
+++ b/src/client/cli/cmd/wait_ready.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_WAIT_READY_H
+#define MULTIPASS_WAIT_READY_H
+
+#include <multipass/cli/command.h>
+#include "animated_spinner.h"
+#include <multipass/timer.h>
+
+
+#include <chrono>
+#include <thread>
+
+namespace multipass
+{
+class Formatter;
+
+namespace cmd
+{
+class WaitReady final : public Command
+{
+public:
+    using Command::Command;
+    ReturnCode run(ArgParser* parser) override;
+
+    std::string name() const override;
+    QString short_help() const override;
+    QString description() const override;
+
+private:
+    WaitReadyRequest request;
+
+    ParseCode parse_args(ArgParser* parser);
+    std::unique_ptr<multipass::AnimatedSpinner> spinner;
+    std::unique_ptr<multipass::utils::Timer> timer;
+};
+} // namespace cmd
+} // namespace multipass
+#endif // MULTIPASS_WAIT_READY_H
--- a/src/client/cli/cmd/wait_ready.h
+++ b/src/client/cli/cmd/wait_ready.h
@@ -19,12 +19,6 @@
 #define MULTIPASS_WAIT_READY_H
 
 #include <multipass/cli/command.h>
-#include "animated_spinner.h"
-#include <multipass/timer.h>
-
-
-#include <chrono>
-#include <thread>
 
 namespace multipass
 {
@@ -46,8 +40,6 @@ private:
     WaitReadyRequest request;
 
     ParseCode parse_args(ArgParser* parser);
-    std::unique_ptr<multipass::AnimatedSpinner> spinner;
-    std::unique_ptr<multipass::utils::Timer> timer;
 };
 } // namespace cmd
 } // namespace multipass
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -619,6 +619,7 @@ auto connect_rpc(mp::DaemonRpc& rpc, mp::Daemon& daemon)
     QObject::connect(&rpc, &mp::DaemonRpc::on_snapshot, &daemon, &mp::Daemon::snapshot);
     QObject::connect(&rpc, &mp::DaemonRpc::on_restore, &daemon, &mp::Daemon::restore);
     QObject::connect(&rpc, &mp::DaemonRpc::on_daemon_info, &daemon, &mp::Daemon::daemon_info);
+    QObject::connect(&rpc, &mp::DaemonRpc::on_wait_ready, &daemon, &mp::Daemon::wait_ready);
 }
 
 enum class InstanceGroup
@@ -3124,6 +3125,41 @@ try
 
     server->Write(response);
     status_promise->set_value(grpc::Status{});
+}
+catch (const std::exception& e)
+{
+    status_promise->set_value(grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), ""));
+}
+
+void mp::Daemon::wait_ready(const WaitReadyRequest* request,
+                       grpc::ServerReaderWriterInterface<WaitReadyReply, WaitReadyRequest>* server,
+                       std::promise<grpc::Status>* status_promise)
+try
+{
+    WaitReadyReply response;
+
+
+    mpl::ClientLogger<WaitReadyReply, WaitReadyRequest> logger{
+        mpl::level_from(request->verbosity_level()),
+        *config->logger,
+        server};
+
+    logger.log(mpl::Level::debug, "daemon", "Checking connection to image servers...");
+    
+    // We use wait_update_manifests_all_and_optionally_applied_force to check connectivity to image servers.
+    // If the force_manifest_network_download is true, it will download the manifests even if already cached.
+    try 
+    {
+        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/true);
+        logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
+        server->Write(response);
+        status_promise->set_value(grpc::Status::OK);
+    }
+    catch(const std::exception& e)
+    {
+        logger.log(mpl::Level::error, "daemon", fmt::format("Failed to connect to image servers: {}", e.what()));
+        status_promise->set_value(grpc::Status(grpc::StatusCode::UNAVAILABLE, e.what(), ""));
+    }
 }
 catch (const std::exception& e)
 {
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3150,7 +3150,7 @@ try
     // If the force_manifest_network_download is true, it will download the manifests even if already cached.
     try 
     {
-        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/true);
+        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/false);
         logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
         server->Write(response);
         status_promise->set_value(grpc::Status::OK);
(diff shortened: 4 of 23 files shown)


ricab:
Hey @rluna319, thank you for this! We'll review when we get a chance.

ricab:
Hey @levkropp, would you mind doing a secondary review of this one? Than I can do a very light "tertiary" review.

levkropp (dismissed):
Yep, tested and working with the examples provided. Great additions!

Needs a pass through `clang-format-16` to satisfy the linter I believe

src/daemon/daemon.cpp
```diff
@@ -3127,6 +3128,41 @@ catch (const std::exception& e)
...
+        *config->logger,
+        server};
+
+    logger.log(mpl::Level::debug, "daemon", "Checking connection to image servers...");
+    
+    // We use wait_update_manifests_all_and_optionally_applied_force to check connectivity to image servers.
+    // If the force_manifest_network_download is true, it will download the manifests even if already cached.
+    try 
+    {
+        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/true);
```
xmkg: I'm skeptical about forcing a manifest download here. The daemon's constructor already dispatches a manifest update by constructing the `update_manifests_all_task` object, so wouldn't `force_manifest_network_download=false` be enough to ensure daemon readiness?

~~I do believe we should have an option to force it, though. In that case, we can mimic what `find` is already doing, i.e., `--force-update`.~~  Let me know if I'm missing something.

xmkg (changes requested):
Great work, @rluna319! I appreciate the work you've put into this. I was on the verge of implementing a similar feature for the upcoming CLI tests I'm working on, and you can't imagine how happy I was to see this PR :)

Overall, the code looks good! I built and tested your changes locally, and everything works as expected. I have a small request in line, and also, having some unit tests would be nice.

src/daemon/daemon.cpp
xmkg: About adding `--force-update`, we discussed with @ricab out-of-band and agreed upon that it is redundant to add it here since it's possible to force a manifest update by calling `multipass find --force-update`.

rluna319: @xmkg Thank you for the review!

> I'm skeptical about forcing a manifest download here. The daemon's constructor already dispatches a manifest update by constructing the `update_manifests_all_task` object, so wouldn't `force_manifest_network_download=false` be enough to ensure daemon readiness?

This was a concern of mine when deciding to reuse this function out of its intended use case. The goal of forcing a manifest download here for the `wait-ready` is to ensure that the daemon can reach the image servers (solving #3723) not entirely to force the manifests to be downloaded/updated.

The `update_manifests_all_task` is started in the background but to my understanding its an asynchronous task dispatched to Qt and I don't think there is any way to check that it is completed before the client issues a command. 

If the daemon has not yet established its connection to the image servers, the image won't be able to be downloaded and we get an error when running `launch`:
- If the manifest cache is empty or invalid, you'll see "Remote '' is unknown or unreachable."
- If the manifest cache is valid but the image is not cached and the daemon can't reach the image server, you'll see something like "Remote 'release' is unknown or unreachable."

I guess it would really depend on our exact definition of "readiness." Checking that the daemon is simply reachable is one way to approach that. But to cover all bases and prevent errors like #3723, I think the wait-ready should include waiting for the daemon to be able to connect to the image servers as well. 

****

In the event that the user is working offline and has the image cached or is using custom images, always checking connectivity to image servers is NOT desirable. I had not accounted for this until right now. So this is a fix that should be implemented for sure. Perhaps implementing your suggestion:
> ~I do believe we should have an option to force it, though. In that case, we can mimic what `find` is already doing, i.e., `--force-update`.~ 

is a good way to approach this.

I haven't explored a more custom method of checking image server connectivity. There might be a cleaner way of doing this. 

And just as well, I suppose a quick and dirty workaround would be to just have `force_manifest_network_download=false` and after `wait-ready` use `find --force-update` (if the user requires an image from the image servers) before using `launch`.

xmkg: > The `update_manifests_all_task` is started in the background but to my understanding its an asynchronous task dispatched to Qt...

Yes, your understanding is correct. The asynchronous task is constructed and dispatched upon construction of the `Daemon` object: 

https://github.com/canonical/multipass/blob/526c3e5aaf0f743183305573505eb1af2dfafc24/src/daemon/daemon.h#L263-L269

The `async_periodic_download_task` immediately dispatches the `work`, being the manifest fetch task for our instance:

https://github.com/canonical/multipass/blob/526c3e5aaf0f743183305573505eb1af2dfafc24/include/multipass/async_periodic_download_task.h#L69

So the primary reason `why` https://github.com/canonical/multipass/issues/3723 happens is that the daemon dispatches the task, but it's still in-flight at the time that launch happens, hence the `launch` failure. 

> ... and I don't think there is any way to check that it is completed before the client issues a command.

The class has a `wait_ongoing_task_finish` function, which blocks the calling thread until the async job finishes:

https://github.com/canonical/multipass/blob/526c3e5aaf0f743183305573505eb1af2dfafc24/include/multipass/async_periodic_download_task.h#L118-L121

It's a no-op if there is no task in flight. The `wait_ongoing_task_finish` is called unconditionally when the `wait_update_manifests_all_and_optionally_applied_force` function is called:

https://github.com/canonical/multipass/blob/526c3e5aaf0f743183305573505eb1af2dfafc24/src/daemon/daemon.cpp#L4016-L4019

Calling `wait_update_manifests_all_and_optionally_applied_force` would make the caller wait for the task, and given that there will be at least 1 guaranteed async dispatch, it shouldn't be necessary to do a forced manifest fetch. 

Perhaps it might also be a good idea to call `wait_update_manifests_all_and_optionally_applied_force(false)` in functions that depend on the manifests, such as `launch`. @ricab opinions?

ricab: Actually, after thinking a little more about this, I think I would rather simplify and just have `wait-ready` complete successfully only after the manifests have been successfully updated. Yes, people can use the daemon offline with cached images, but nothing in here prevents that. They would just not be able to use the wait-ready functionality because, in a sense, the daemon can't be fully ready until it retrieves image info.

> I guess it would really depend on our exact definition of "readiness." 

Exactly, and I think the definition of ready that makes most sense is "fully ready for normal operation", which requires online connectivity and updated manifests.

> Perhaps it might also be a good idea to call `wait_update_manifests_all_and_optionally_applied_force(false)` in functions that depend on the manifests, such as `launch`.

It might be a good idea, as long as we updated the spinner message while doing so, because it can take a long time if there are connectivity issues (exponential back-off). And the launch command would then just hang. As it is, the launch just fails in such circumstances. That would be for a different PR though.

ricab: If the command was run after 15min (period for the manifest update), it would again hang for a little bit. It's no big deal but, to be very precise, you could add an accessor to `AsyncPeriodicDownloadTask` reporting whether the task had ever completed. Then replace the call to `mp::Daemon::wait_update_manifests_all_and_optionally_applied_force` with something like `update_manifests_all_task.ever_completed() || update_manifests_all_task.wait_ongoing_task_finish()`. Or perhaps even better, just add a method to `wait_for_first_completion()`. Names up for improvement of course.

rluna319: > The class has a wait_ongoing_task_finish function, which blocks the calling thread until the async job finishes

@xmkg thank you for pointing that out. A bit of an oversight there. `update_manifests_all_task.wait_ongoing_task_finish()` does allow us to wait till the task finishes. 

>So the primary reason why https://github.com/canonical/multipass/issues/3723 happens is that the daemon dispatches the task, but it's still in-flight at the time that launch happens, hence the launch failure.

Right and I thought forcing the manifest download was the only way to guarantee that the demon can connect to the image servers. However, upon further review, I think my understanding was incorrect. I thought that in the event that the manifest cache still exists and is valid, the daemon would NOT attempt to connect to the image servers (which is why I thought it needed to be forced). But reading into the Qt documentation more, i think `QNetworkRequest::CacheLoadControl::PreferNetwork` always chooses to re-validate the cache with the server (if its available). Meaning the connection to the image servers is verified whether it is forced or not. 

So with this, yes `force_manifest_network_download=false` is enough to determine the daemon's readiness.

My inexperience with Qt shows, I do apologize for the oversights.

>Exactly, and I think the definition of ready that makes most sense is "fully ready for normal operation", which requires online connectivity and updated manifests.

I see. So for clarification, this would mean that waiting for a manifest update should not be an optional task in `wait-ready` and will remain as is (aside from the change to not force the download)?

ricab: > I see. So for clarification, this would mean that waiting for a manifest update should not be an optional task in `wait-ready` and will remain as is (aside from the change to not force the download)?

That's my inclination. I imagine that's what most people would want. We could add a `--no-network`, but how many people would want to use that really? Probably not a lot. So I guess we make that choice and avoid the option, simplifying the command. @xmkg WDYT?

xmkg: > My inexperience with Qt shows, I do apologize for the oversights.

No worries -- I'm not an expert in QT myself, and it surprises me every day too (mostly bad ones).

> ... you could add an accessor to AsyncPeriodicDownloadTask reporting whether the task had ever completed

This would be cool. 

> That's my inclination. I imagine that's what most people would want. We could add a --no-network, but how many people would want to use that really? Probably not a lot. So I guess we make that choice and avoid the option, simplifying the command. @xmkg WDYT?

Sounds good to me :+1:

src/client/cli/cmd/wait_ready.h
```diff
@@ -0,0 +1,54 @@
...
+    std::string name() const override;
+    QString short_help() const override;
+    QString description() const override;
+
+private:
+    WaitReadyRequest request;
+
+    ParseCode parse_args(ArgParser* parser);
+    std::unique_ptr<multipass::AnimatedSpinner> spinner;
+    std::unique_ptr<multipass::utils::Timer> timer;
```
ricab: These are only used in a single method, right? If so, could they be variables in it, to restrict the scope to the minimum necessary?

src/client/cli/cmd/wait_ready.h
```diff
@@ -0,0 +1,54 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef MULTIPASS_WAIT_READY_H
+#define MULTIPASS_WAIT_READY_H
+
+#include <multipass/cli/command.h>
+#include "animated_spinner.h"
+#include <multipass/timer.h>
```
ricab: These would get moved to the cpp if you limited the scope of the fields as suggested below, but in any case, can the include order go from nearest dependency to the farthest one? i.e. in the cpp, you could include wait_ready.h first, then files in the same directory, subdirectories, sibling directories, cousin directories, etc., local 3rd party, system-wide 3rd-party (like Qt), and finally standard library.

This approach helps ensure that includes work by themselves, i.e., don't depend on something else having been included first, avoiding breaking builds when includes are removed or replaced.

src/client/cli/cmd/wait_ready.h
```diff
@@ -0,0 +1,54 @@
...
+
+
+#include <chrono>
+#include <thread>
```
ricab: You should be able to move these to the cpp.

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "wait_ready.h"
+#include "common_cli.h"
+
+#include "animated_spinner.h"
```
ricab: I am surprised clang-format allowed this.

```suggestion
#include "animated_spinner.h"
#include "common_cli.h"

```

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // Check if the command was parsed successfully
+    if (status != ParseCode::Ok)
+    {
+        return status;
+    }
+
+    try
+    {
+        mp::cmd::parse_timeout(parser);
```
ricab, line 124: Well done reusing this!

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+        "other commands.\n"
+        "If a timeout is specified, it will wait for that duration before "
+        "exiting with an error if the daemon is not ready.");
+}
+
+mp::ParseCode cmd::WaitReady::parse_args(mp::ArgParser* parser)
+{
+    mp::cmd::add_timeout(parser);
+
+    auto status = parser->commandParse(this);
```
ricab, line 114: And laudable here too. Unfortunately, the help text includes stuff that doesn't make sense in here, as it is specific to instance-starting operations. Perhaps you could adapt it a little though?

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+QString cmd::WaitReady::description() const
+{
+    return QStringLiteral(
+        "Wait for the Multipass daemon to be ready.\n"
+        "This command will block until the daemon is ready to accept requests.\n"
+        "It can be used to ensure that the daemon is running before executing "
+        "other commands.\n"
+        "If a timeout is specified, it will wait for that duration before "
+        "exiting with an error if the daemon is not ready.");
```
ricab: Good description, but I think we should detail what we consider "ready". And the formatting in other commands usually wrap the description at 80 chars. How about

```suggestion
        "Wait for the Multipass daemon to be ready. This command will block until the\n"
        "daemon has initialized, fetched up-to-date image information, and is ready to\n"
        "accept requests. Its main use is to prevent failures caused by incomplete\n"
        "initialization in batch operations. An optional timeout aborts the command if\n"
        "reached.");
```

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    auto ret = parse_args(parser);
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+
+    if (!spinner){
+        spinner = std::make_unique<multipass::AnimatedSpinner>(cout);
+        spinner->start("Waiting for Multipass daemon to be ready...");
```
ricab: I believe other spinner messages have no punctuation at the end. 

```suggestion
        spinner->start("Waiting for the Multipass daemon to be ready");
```

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,131 @@
...
+mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
+{
+    auto ret = parse_args(parser);
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+
+    if (!spinner){
+        spinner = std::make_unique<multipass::AnimatedSpinner>(cout);
```
ricab: Going back to the matter of field vs local variable, no need for a pointer (or heap allocation) right? I believe a regular var on the stack should be enough.

ricab:
@rluna319, could you also rebase on the latest main, to see if we can get some CI to run on this? Appreciate it.

rluna319:
@ricab I had not included your requested changes on my latest commit. I had not seen them before I made the push, my fault. I did rebase on the latest main however.

I will work on your requested changes ASAP.

Would you like me to amend your requested changes to my latest commit or have it in a separate commit?

ricab:
Hey @rluna319 no worries. We generally prefer additional commits, to make it easy to identify new changes. Thanks!

src/client/cli/cmd/wait_ready.cpp
rluna319: Sure, would simply removing the portion concerning instances suffice? More or less leaving it consistent with its description in the other commands with the `--timeout` option.

"Maximum time, in seconds, to wait for the command to complete. Note that some background operations may continue beyond that."

Or we could tailor it exclusively for the wait-ready with something like:

"Maximum time, in seconds, to wait for the Multipass daemon to be ready. Note that some background operations may continue beyond that."

rluna319: Also, in the original help text. There is mention of a `default_timeout`. This is strictly regarding the instances, not concerning `--timeout` arg itself, correct? We can safely disregard this in our adapted help text?

ricab: Right, I suggest one of two approaches:

- Easy: dedicated message here and rename `add_timeout` to `add_instance_timeout`
- Nice: split the instance-specific part of `add_timeout` into `add_instance_timeout` (calling the former and completing it); then call add_timeout in here

rluna319:
@ricab I've added all of your requested changes to the recent commit. I also rebased on the latest main as well.

.gitignore
```diff
@@ -5,6 +5,7 @@ cmake-build-*
 .DS_Store
 CMakeLists.txt.user
 build
+Testing
```
ricab: Hmm, what is the idea here?

src/daemon/daemon.cpp
```diff
@@ -3130,6 +3131,41 @@ catch (const std::exception& e)
...
+    {
+        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/false);
+        logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
+        server->Write(response);
+        status_promise->set_value(grpc::Status::OK);
+    }
+    catch(const std::exception& e)
+    {
+        logger.log(mpl::Level::error, "daemon", fmt::format("Failed to connect to image servers: {}", e.what()));
+        status_promise->set_value(grpc::Status(grpc::StatusCode::FAILED_PRECONDITION, e.what(), ""));
```
ricab: Might as well `throw;` and avoid some repetition.

src/daemon/daemon.cpp
```diff
@@ -3130,6 +3131,41 @@ catch (const std::exception& e)
...
+
+    logger.log(mpl::Level::debug, "daemon", "Checking connection to image servers...");
+    
+    // We use wait_update_manifests_all_and_optionally_applied_force to check connectivity to image servers.
+    // If the force_manifest_network_download is true, it will download the manifests even if already cached.
+    try 
+    {
+        wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/false);
+        logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
+        server->Write(response);
```
ricab: I don't think this is necessary, is it? Was this for the logs? I believe the client logger takes are of that.

tests/test_daemon_wait_ready.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    EXPECT_CALL(*mock_image_host, update_manifests(false)).Times(1).WillOnce([]() {
+        throw mp::DownloadException{"dummy_url", "dummy_cause"};
+    });
+
+    config_builder.image_hosts[0] = std::move(mock_image_host);
+    mp::Daemon daemon{config_builder.build()};
+    
+    send_command({"wait-ready"});
+}
```
ricab: Missing a newline.

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,136 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "wait_ready.h"
+#include "common_cli.h"
+#include "animated_spinner.h"
+
+#include <multipass/timer.h>
+#include <multipass/constants.h>
```
ricab: I believe you can remove this include.

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <multipass/exceptions/cmd_exceptions.h>
+#include <multipass/cli/argparser.h>
+#include <multipass/cli/formatter.h>
```
ricab: And this one.

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,136 @@
...
+mp::ReturnCode cmd::WaitReady::run(mp::ArgParser* parser)
+{
+    auto ret = parse_args(parser);
+    if (ret != ParseCode::Ok)
+    {
+        return parser->returnCodeFrom(ret);
+    }
+    
+    mp::AnimatedSpinner spinner(cout);
+    spinner.start("Waiting for Multipass daemon to be ready");
```
ricab:
```suggestion
    spinner.start("Waiting for the Multipass daemon to be ready");
```

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    std::unique_ptr<mp::utils::Timer> timer;
+
+    // If the user has specified a timeout, we will create a timer
+    if (parser->isSet("timeout")){
+        timer = cmd::make_timer(parser->value("timeout").toInt(),
+                                &spinner,
+                                cerr,
+                                "Timed out waiting for Multipass daemon to be ready.");
```
ricab:
```suggestion
                                "Timed out waiting for the Multipass daemon to be ready.");
```

src/daemon/daemon.cpp
ricab: However, I wonder if this is really what we want. If I unplug from the internet, start the daemon, then run `wait-ready`, it fails immediately:

```
$ ./bin/multipass wait-ready
Waiting for Multipass daemon to be ready /[2025-07-25T15:27:13.619] [error] [daemon] Failed to connect to image servers: failed to download from 'https://cdimage.ubuntu.com/ubuntu-core/16/stable/current/ubuntu-core-16-amd64.img.xzwait-ready failed: failed to download from 'https://cdimage.ubuntu.com/ubuntu-core/16/stable/current/ubuntu-core-16-amd64.img.xz': Host cdimage.ubuntu.com not found
```

If it kept waiting, the daemon might be able to connect eventually. So, in an environment that took a while to connect, people would have run wait-ready in a loop, which kind of defeats the purpose, at least for the disconnected use case. 

OTOH, this PR has been going for a while and I don't want to ask for much more. It's certainly valuable even so and I am thinking we could probably improve this aspect at a later time, if there was demand for it.

@xmkg, I would love to hear your opinion.

ricab (commented):
Getting real close @rluna319! I am going to let CI go through and I think the Lint workflow is going to request some formatting changes. Other than that, I have mostly minor requests. There's only one for real behavior, but I think we can discuss and split off as a separate task if we need it.

.gitignore
rluna319: Ah that is an error on my part that should not have been included. I was running the unit tests locally and it was creating some directories and test result files. I put that there at one point out of confusion and forgot it was there.

I will remove it.

src/daemon/daemon.cpp
rluna319: >If it kept waiting, the daemon might be able to connect eventually. So, in an environment that took a while to connect, people would have run wait-ready in a loop, which kind of defeats the purpose, at least for the disconnected use case.

This was something I meant to ask about and address on the second commit. It slipped my mind.

Setting the return code in the inner try catch block  from `FAILED_PRECONDITION` to `UNAVAILABLE` (what I had previously but unintentional at the time) would have it retry instead. And then we also just change the log level to debug instead of error. That might not be the nicest solution however, as it would go back to the client which would send another dispatch. If its more suitable, we would just implement some retry logic in the daemon wait-ready handler instead.

>OTOH, this PR has been going for a while and I don't want to ask for much more. It's certainly valuable even so and I am thinking we could probably improve this aspect at a later time, if there was demand for it.

Also sounds good to me. I don't mind doing a bit more either. I much appreciate the experience this contribution is providing me.

src/daemon/daemon.cpp
rluna319: I did think it was necessary for the logs. You are correct, the client logger does take care of that. My mistake. I will remove it.

src/daemon/daemon.cpp
rluna319: @ricab @xmkg we could do something like this:

```C++
// We use wait_update_manifests_all_and_optionally_applied_force to check connectivity to image servers.
    while (true)
    {
        try
        {
            wait_update_manifests_all_and_optionally_applied_force(/*force_manifest_network_download=*/false);
            logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
            status_promise->set_value(grpc::Status::OK);
            return; // Success, exit the function.
        }
        catch(const mp::DownloadException& e)
        {
            logger.log(mpl::Level::debug, "daemon", fmt::format("Failed to connect to image servers, retrying...: {}", e.what()));
            // The client is responsible for the timeout, so we just wait and retry.
            std::this_thread::sleep_for(std::chrono::seconds(1));
        }
    }
```

The outer catch block should catch any other exception. I believe the only exception we care about concerning the connection to the image servers is `DownloadException` which we would just retry on till we reach the `--timeout` if specified. 

An even nicer solution would mind the cause of the `DownloadException` as there might be some errors we don't want to retry on. In any case, if the user uses the `-vvvv` option to observe any error going on that is unexpected, they could just terminate the cmd themselves.

ricab: Hey @rluna319, I think the first option you proposed would work best. Letting the client drive retries would work well with the timeout approach. And it would let different clients choose different polling intervals (later on, we could even add `--interval` to the command if we wanted). Most importantly, it would prevent the main thread getting stuck in a loop. 

> Setting the return code in the inner try catch block from `FAILED_PRECONDITION` to `UNAVAILABLE`

+1

> And then we also just change the log level to debug instead of error.

Hmm, I think it is still an error, no? See comments for logging levels: https://github.com/canonical/multipass/blob/eb05c4cc4fb9f583e863b05fb859564c31d62c29/include/multipass/logging/level.h#L32-L42

Possibly a warning, if the status message is clear enough (note that the CLI shows only errors by default).

> An even nicer solution would mind the cause of the DownloadException as there might be some errors we don't want to retry on.

True, but that would require finer error handing at the throw site, after cataloging possible errors, and I'd rather abbreviate that here. As long as we let other exceptions through to the default handling, including `AbortedDownloadException` (which, despite the name, is unrelated to DownloadException).

rluna319: @ricab I like the idea of being able to set different polling intervals. ~~Currently, we have the client poll for the daemon till the socket becomes available. This relies on catching the return code `NOT_FOUND` (in the client) which is returned by the dispatcher if the status code promised to it is set to `UNAVAILABLE`. If we return the status code `UNAVAILABLE` from the daemon after we catch the `DownloadException`, the same happens, the dispatcher will set the status code to `NOT_FOUND` before it reaches the client. This means there wouldn't be any differentiation between polling for socket availability and polling for connectivity to the image server. So the same `--interval` set by the client would be used for both polling for the socket and connectivity to image servers . Moreover, the error message would say "cannot connect to the multipass socket" which is not entirely accurate when the error is from a `DownloadException`. I don't see either as a major issue. But I wanted to point this out before moving forward with the implementation.~~

>Hmm, I think it is still an error, no?

Very true, I was thinking about the default verbosity and didn't want it to constantly be printing to the terminal.

>Possibly a warning, if the status message is clear enough

I'd interpret an error being used for situation where the command exits completely and doesn't retry. e.g. If we had a `--retry` option, which would let the client choose whether to retry or not, if it's not set then we have it log as an error, otherwise a warning. However, if you look at it from the daemon's end it is an error, but the client is choosing to retry on this error.

>note that the CLI shows only errors by default

I suppose this would be the only thing turning me away from leaving it as error.

>True, but that would require finer error handing at the throw site, after cataloging possible errors

`DownloadException`s have a cause attached. `QNetworkReply` has an enumerated list `NetworkError` which I believe is used for the cause when throwing the `DownloadException` after checking the cache. So, we might be able to filter through the possible errors that way if and when this could be implemented in the future at some point.

rluna319: > (later on, we could even add --interval to the command if we wanted)

We could just implement `--interval` later on yes. So for this PR I would just change the status code set to `UNAVAILABLE` instead and leave it at that.

Actually, we could just return `NOT_FOUND` from the daemon instead with our own error message. This is much better I think.

ricab: That makes sense to me, sounds very reasonable. I'd leave custom intervals and finer DownloadException handling for later, but up to you :slightly_smiling_face:

ricab:
Hey @rluna319, it looks like you [haven't signed our CLA](https://github.com/canonical/multipass/actions/runs/16452896886/job/46735070225?pr=4145) yet. Would you mind doing so, please? https://ubuntu.com/legal/contributors

Also, there are a few [linting problems](https://github.com/canonical/multipass/actions/runs/16452897071/job/46735084487?pr=4145). Editors often have options to strip trailing whitespace and keep a single newline at the end. You can check whitespace sanity on a diff with `git diff --check ...`. For example: `git diff --check main...wait-ready`.

`git clang-format` might take care of whitespaces too, if you set it up. That might be a good idea anyway because it's going to run once the whitespace check succeeds.

src/daemon/daemon.cpp
rluna319: I'll leave that for later if it's desired.

rluna319:
@ricab I signed the CLA, went through linter errors (and other CI build/test errors) and fixed them to the best of my knowledge, ran the code through `git clang-format` and `git diff --check main`. I've also rebased off the latest main.

rluna319:
I removed some commented out includes that shouldn't have been in that commit.

3rd-party/jsoncpp
ricab, line 1: Hmm, I don't think this should be here. Did you add an old submodule after rebasing?

src/client/cli/cmd/wait_ready.cpp
```diff
@@ -0,0 +1,135 @@
...
+        if (timer)
+            timer->stop();
+        spinner.stop();
+        return ReturnCode::Ok;
+    };
+
+    auto on_failure = [this, &spinner, &timer](grpc::Status& status) {
+        if (status.error_code() == grpc::StatusCode::NOT_FOUND &&
+            (status.error_message() == "cannot connect to the multipass socket" ||
+             status.error_message() == "cannot connect to the image servers"))
```
ricab: Sorry if I misunderstood what you planned before, but why check the error messages? I don't think clients should depend on message strings for behavior (that's what the error code is for). Is there any other case that would fall under `NOT_FOUND`?

src/daemon/daemon.cpp
```diff
@@ -3130,6 +3131,46 @@ catch (const std::exception& e)
...
+    try
+    {
+        wait_update_manifests_all_and_optionally_applied_force(
+            /*force_manifest_network_download=*/false);
+        logger.log(mpl::Level::debug, "daemon", "Successfully connected to image servers.");
+        status_promise->set_value(grpc::Status::OK);
+    }
+    catch (const mp::DownloadException& e)
+    {
+        logger.log(mpl::Level::warning,
```
ricab, line 3160: :+1: on warning

3rd-party/jsoncpp
rluna319: https://github.com/canonical/multipass/blob/db41306c7f7570baa3ae9652f2f6a726fed90fd2/.gitmodules#L17

I think main currently tracks jsoncpp as a submodule. Am I rebasing wrong?

src/client/cli/cmd/wait_ready.cpp
rluna319: It goes back to what I was trying to say about using `UNAVAILABLE` before I decided that `NOT_FOUND` would be more suitable. If we were to implement the `--interval` option, the specified interval would be used for both waiting for the daemon socket to become available and waiting for the daemon to connect to the image servers.

If we set the status to `UNAVAILABLE`, the dispatcher will change this to `NOT_FOUND` with the "cannot connect to the multipass socket" string. Using `NOT_FOUND` lets us at least set the error string to a more accurate message about image server connectivity. After that the goal was to be able to have `--interval` only affect the polling for the image servers, not socket availability. So the only way to differentiate in wait_ready cpp would be to check the error string. 

 >I don't think clients should depend on message strings for behavior (that's what the error code is for).
 
 I agree. This is probably not a good approach thinking back on it. Especially now that we aren't implementing the `--interval` option yet.
 
 >Is there any other case that would fall under NOT_FOUND
 
 Regarding wait_ready no not that i'm aware of. 
 
 We should be able to safely get rid of checking the error messages.

rluna319:
I see that I still missed some linter errors. I thought I ran everything through `clang-format` correctly. I'll address those.

3rd-party/jsoncpp
xmkg: jsoncpp's head in this branch points to > `9059f5cad030ba11d37818847443a53918c327b1`(1.9.4) where it should be `89e2973c754a9c02a49974d839779b151e95afd6`(1.9.6).  `git -C 3rd-party/jsoncpp/ checkout 1.9.6` should fix it.

rluna319: Oh I see. I might have done something out of order accidentally. I'm not sure how I didn't receive any conflicts when rebasing. I don't watch the diffs carefully enough. I will address this. Thank you.

rluna319:
@ricab Ive addressed the issues. It says some of the workflows are awaiting approval to be run.

xmkg:
> @ricab Ive addressed the issues. It says some of the workflows are awaiting approval to be run.

Approved the workflow.

rluna319:
Thank you @xmkg.

I think that the remaining workflows that fail are due to some internal CI issues. Something about a missing token. Is this something I need to fix or can fix on my end?

xmkg:
> Thank you @xmkg.
> 
> I think that the remaining workflows that fail are due to some internal CI issues. Something about a missing token. Is this something I need to fix or can fix on my end?

Yeah, unfortunately, our CI is not happy with PRs coming from external forks. We're working on making this process smoother, but the necessary machinery is not yet in place. Would you mind if I create a PR for your PR, just for the sake of running the pipeline? We can track the progress and make changes here.

rluna319:
Oh ok. Sure thing.

xmkg:
Opened https://github.com/canonical/multipass/pull/4304.

xmkg:
@rluna319, we've landed a bunch of CI-related fixes to `main`. Could you rebase this PR?

rluna319:
@xmkg I have rebased the PR. Workflows awaiting approval.

xmkg:
> @xmkg I have rebased the PR. Workflows awaiting approval.

Great, thanks. Synced #4304.

rluna319:
@xmkg Great thank you. I see all the CI tests have passed in #4304. Should I mind the coverage report? I've looked at it already and I can cover the line in `wait_ready.cpp` that it says isn't covered. 

However, i'm having trouble covering the other one in `daemon.cpp` where we catch any other std::exception. Throwing a runtime error in update_manifests just causes the test to hang. I'm guessing its the way that exception is or isn't handled in the asynchronous task. I'm not sure what other exception would make it to that catch block. Any other exceptions that ive traced within the codepath are already handled locally. Perhaps i'm just missing something i'm not aware of.

xmkg:
> @xmkg Great thank you. I see all the CI tests have passed in #4304. Should I mind the coverage report? I've looked at it already and I can cover the line in `wait_ready.cpp` that it says isn't covered.

That would be nice.

> However, i'm having trouble covering the other one in `daemon.cpp` where we catch any other std::exception. Throwing a runtime error in update_manifests just causes the test to hang. I'm guessing its the way that exception is or isn't handled in the asynchronous task.

I'd say don't worry about it. There's a way to trigger that exception catcher... but it's a bit cheeky, though. We could use a logger that throws an exception on `.log()`. That would make the coverage tool happy, but I'm okay with it as-is, too.

rluna319:
@xmkg I added the coverage for `wait_ready.cpp`.

I tried to implement your suggestion, throwing and exception on `log()`, but wasn't able to do it. I kept getting the error that the Logger has no member named "gmock_log". I thought maybe I would have to add a mock method for `log()` in `mock_logger.cpp` but I felt that would be overkill if coverage for that line wasn't crucial. I tried doing a local mock method for the logger as well but I couldn't get that to work either. I may just lack the experience to implement this properly.

xmkg (approved):
Looks good to me, thanks @rluna319!

ricab (approved):
I just did a final pass and it's looking great. Thanks again!

rluna319:
Awesome! Thank you for allowing me to contribute @xmkg @ricab. A great experience, I learned a lot. Looking forward to the next!

ricab:
Our pleasure @rluna319! This was a great contribution, very well pondered. Feel free to send more in!

PR #4148, opened 2025-06-06
docs: fix OGP string

### LINKED ISSUE #4139: docs: Your slugs in the metadata are incorrect
*Please describe the question or issue you're facing with "snapshot - Multipass documentation".*

It is happening in all your slugs, but I am using this as an example. 
When you save the page in a program like Zotero, the slug that is saved is: 
https://documentation.ubuntu.com/explanation/snapshot/

But in reality the page is in:
https://documentation.ubuntu.com/multipass/en/latest/explanation/snapshot/

No idea why it is happening, but I'm having to manually correct all the links. 

Martin

---
*Reported from: https://documentation.ubuntu.com/multipass/en/latest/explanation/snapshot/*

Image

Image

edibotopic:
The OGP var defined in `conf.py` affects the meta tags in the rendered HTML of the documentation. 

The string needs to end with a forward slash and include the default `<lang>` and `<version>` for the documentation.

It is currently defined as `"https://documentation.ubuntu.com/multipass"`, which yields
incorrect meta tags, for example:

```html
<meta property="og:url" content="https://documentation.ubuntu.com/how-to-guides/install-multipass/" />
```

Changing to `"https://documentation.ubuntu.com/multipass/en/latest/"`, yields the correct URL in the content property:

```html
<meta property="og:url" content="https://documentation.ubuntu.com/multipass/en/latest/how-to-guides/install-multipass/" />
```

When saving a page to Zotero, for example, we then get the correct documentation URL:

ogp-after

Whereas before, the URL was wrong:

ogp-before

This should hopefully fix #4139 .

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,7 +70,7 @@ copyright = "%s CC-BY-SA, %s" % (datetime.date.today().year, author)
 # NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
 #       and is used by social media platforms; see https://ogp.me/
 
-ogp_site_url = "https://documentation.ubuntu.com/multipass"
+ogp_site_url = "https://documentation.ubuntu.com/multipass/en/latest/"
 
 
 # Preview name of the documentation website


edibotopic:
Closing in favour of: https://github.com/canonical/multipass/pull/4150

PR #4155, opened 2025-06-11
[ci] add vcpkg binary caching

sharder996:
This PR introduces [binary caching](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=linux-runner) for vcpkg. Since going open source, we can take advantage of [GitHub Packages](https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages) to host binary caches.

Binary caching is enabled by setting the `VCPKG_BINARY_SOURCES` environment variable. A tutorial for how to do this can be found [here](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages) and was generally followed for changes in this PR.

For MacOS and ARM, there is seemingly no official support or guide, and with the differences in bundled packages in GitHub hosted runners, a little bit of tailoring is required. On non-Windows machines, once a binary source has been registered, vcpkg seems to be hard coded to call `mono nuget` in order to restore or upload binaries. `mono` expects a .Net assembly compiled for Intel architecture, so when an ARM compatible package is bundled in the image from GitHub this fails. To get around this, NuGet is manually downloaded and symlinked so that vcpkg will call it.

On Linux, binary caching must be configured within the snapcraft image. Again the appropriate NuGet binary needs to be downloaded and configured so that it is callable from the command line with `mono nuget`.

Build times are considerably shorter once all dependencies are cached with the "Configure" step taking less than a few minutes. On average, this saves around 30 minutes which is approximately a 33-50% improvement in build times.

Because of the expectancy that self-hosted runners will be dropped soon, binary caching is disabled for them.

#### GitHub Packages

GitHub Packages does not offer any sort of retention policy meaning that as vcpkg binaries change versions, Packages will increasingly become more and more cluttered with old versions of binaries. A solution to this problem is to setup a scheduled GitHub Action with [actions/delete-package-versions](https://github.com/marketplace/actions/delete-package-versions) and `min-versions-to-keep: 1` to only keep the most recent version of a binary.

#### Visibility

There are two storage plans for GitHub Packages which depends on visibility. Publicly visible packages show up on the [front page](https://github.com/orgs/canonical/packages?repo_name=multipass) of the Multipass repo, but have unlimited storage and bandwidth quota. Having these packages there is a bit unsightly and to the uneducated visitors they are seemingly irrelevant to Multipass and might confuse someone trying to download the Multipass application. Setting the visibility to private would be preferential, but is subject to [billing](https://docs.github.com/en/billing/managing-billing-for-your-products/about-billing-for-github-packages).

An alternative solution for storage is to use our [AWS S3 bucket](https://learn.microsoft.com/en-us/vcpkg/reference/binarycaching#aws). At this point, it is unclear whether this removes binaries from GitHub Packages or is simply a different backend storage provider for GitHub Packages. Also, since I do not have access to Multipass' AWS account, available storage quotas and unknown and so whether this is a viable solution is also unknown.

---

P.S. You can even setup binary caching [locally](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-local?pivots=shell-bash) to speed up builds!

MULTI-2030

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -74,7 +74,10 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      FEED_URL: https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json
       SNAPCRAFT_BUILD_INFO: 1
+      USERNAME: ${{ github.repository_owner }}
+      VCPKG_BINARY_SOURCES: "clear;nuget,https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json,readwrite"
 
     timeout-minutes: 120
     steps:
@@ -162,7 +165,22 @@ jobs:
         fi
 
         if [ -n "${VCPKG_BINARY_SOURCES}" ]; then
-          sed -i "/build-environment:/a \    - VCPKG_BINARY_SOURCES: ${VCPKG_BINARY_SOURCES}" snap/snapcraft.yaml
+          sed -i '/build-environment:/ {
+          a\
+            - VCPKG_BINARY_SOURCES: '"${VCPKG_BINARY_SOURCES}"'\
+            - FEED_URL: '"${FEED_URL}"'\
+            - USERNAME: '"${USERNAME}"'\
+            - GITHUB_TOKEN: '"${{ secrets.GITHUB_TOKEN }}"'
+          }' snap/snapcraft.yaml
+          sed -i "/build-packages:/a \    - mono-complete" snap/snapcraft.yaml
+          sed -i '0,/override-build: |/ {
+            /override-build: |/ a\
+              curl -L -o /usr/local/bin/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe\
+              ln -sf /usr/local/bin/nuget.exe /usr/local/bin/nuget\
+              export PATH=$PATH:/usr/local/bin\
+              mono /usr/local/bin/nuget.exe sources add -name GitHubPackages -source "${FEED_URL}" -username "${USERNAME}" -password "${GITHUB_TOKEN}" -storePasswordInClearText\
+              mono /usr/local/bin/nuget.exe setapikey "${GITHUB_TOKEN}" -source "${FEED_URL}"
+          }' snap/snapcraft.yaml
         fi
 
         if [ -n "${GITHUB_ACTIONS}" ]; then
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -179,7 +179,7 @@ jobs:
         fi
 
     - name: Install nuget
-      if: ${{ runner.os == 'macOS' && matrix.arch == 'arm64' && env.VCPKG_BINARY_SOURCES }}
+      if: ${{ contains(matrix.runs-on, 'macos-15') }}
       run: |
         brew install nuget mono
 
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -170,6 +170,7 @@ jobs:
         vcpkgDirectory: '${{ github.workspace }}/3rd-party/vcpkg'
 
     - name: Configure NuGet Source
+      continue-on-error: true
       if: ${{ runner.os == 'Windows' }}
       shell: 'bash'
       run: |
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -20,7 +20,10 @@ env:
   HOMEBREW_NO_AUTO_UPDATE: 1
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  FEED_URL: https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json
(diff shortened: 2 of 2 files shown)


xmkg:
Hi @sharder996, before reviewing, I'd like to say that this is awesome! I was thinking of doing something like this, but glad to see that you took the initiative and beat me to it :)

.github/workflows/windows-macos.yml
```diff
@@ -166,6 +169,48 @@ jobs:
...
+      shell: bash
+      run: |
+        if [[ "${{ matrix['runs-on'] }}" == *multipass* ]]; then
+          echo "VCPKG_BINARY_SOURCES=" >> $GITHUB_ENV
+        fi
+
+    - name: Install nuget
+      if: ${{ runner.os == 'macOS' && matrix.arch == 'arm64' && env.VCPKG_BINARY_SOURCES }}
+      run: |
+        brew install nuget mono
```
xmkg, line 184: I think this step is only required for the `macos-15' runner since the GH macOS runner 13 and 14 images already include NuGet and Mono by default:

https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md 

... but that's not the case for `macos-15`:

https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md

Reason: (excerpt from https://github.com/actions/runner-images/issues/10686)

```
Mono	6.12.*	N/A	Removed from the macOS-15 images due to maintenance reasons
NuGet	6.3.*	N/A	Removed from the macOS-15 images due to maintenance reasons
```
I think we can just check if matrix.runs-on == `macos-15` for this one if I'm not missing anything else. WDYT?

.github/workflows/windows-macos.yml
```diff
@@ -166,6 +169,48 @@ jobs:
...
+          -source "${{ env.FEED_URL }}" \
+          -storepasswordincleartext \
+          -name "GitHubPackages" \
+          -username "${{ env.USERNAME }}" \
+          -password "${{ secrets.GITHUB_TOKEN }}"
+        ${{ matrix.mono }} "$($VCPKG_ROOT/vcpkg fetch nuget | tail -n 1)" \
+          setapikey "${{ secrets.GITHUB_TOKEN }}" \
+          -source "${{ env.FEED_URL }}"
+
+    - name: Overwrite Homebrew nuget shim with nuget DLL
```
xmkg, line 204: Hmm.. can you elaborate why this was necessary?

.github/workflows/windows-macos.yml
```diff
@@ -166,6 +169,48 @@ jobs:
...
+      run: |
+        brew install nuget mono
+
+    - name: Configure NuGet Source
+      if: ${{ env.VCPKG_BINARY_SOURCES }} # TODO: remove this when self-hosted runners are dropped
+      id: setup_nuget
+      continue-on-error: true
+      shell: 'bash'
+      run: |
+        ${{ matrix.mono }} "$($VCPKG_ROOT/vcpkg fetch nuget | tail -n 1)" \
```
xmkg, line 193: In order to not lose context we can put a link to the official guide as comment:

https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=linux-runner

.github/workflows/linux.yml
```diff
@@ -162,7 +165,22 @@ jobs:
         fi
 
         if [ -n "${VCPKG_BINARY_SOURCES}" ]; then
-          sed -i "/build-environment:/a \    - VCPKG_BINARY_SOURCES: ${VCPKG_BINARY_SOURCES}" snap/snapcraft.yaml
+          sed -i '/build-environment:/ {
```
xmkg, line 168: Not related to this PR, but this whole sed'ing and patch'ing of the YAML file makes me wonder if we should switch to using an inline Python script to modify the YAML instead.

xmkg (commented):
Looks good, a few minor comments.

.github/workflows/windows-macos.yml
sharder996: This is why:
> For MacOS and ARM, there is seemingly no official support or guide, and with the differences in bundled packages in GitHub hosted runners, a little bit of tailoring is required. On non-Windows machines, once a binary source has been registered, vcpkg seems to be hard coded to call mono nuget in order to restore or upload binaries. mono expects a .Net assembly compiled for Intel architecture, so when an ARM compatible package is bundled in the image from GitHub this fails. To get around this, NuGet is manually downloaded and symlinked so that vcpkg will call it.

.github/workflows/windows-macos.yml
sharder996: Done

.github/workflows/windows-macos.yml
sharder996: Makes sense to me. Will verify.

.github/workflows/linux.yml
sharder996: Yeah, the `sed` language isn't the most pleasant to work with. Let me know how strongly you feel about it and if you would like to see it changed here.

sharder996:
Rebased on top of `main` to include `vcpkg` update.

.github/workflows/linux.yml
xmkg: I think we can leave it for later.

Issue #4147, opened 2025-06-06
GUI should provide better feedback when unable to authenticate


tmerten:
**Describe the bug**

Cannot launch a multipass VM or start the GUI on OSX

**To Reproduce**

1. `brew install multipass`
2. `multipass launch`
launch failed: The client is not authenticated with the Multipass service.
Please use 'multipass authenticate' before proceeding.
3. `multipass authenticate`
Please enter passphrase:
authenticate failed: Passphrase is not set. Please `multipass set local.passphrase` with a trusted client.
4. `multipass set local.passphrase`
Please enter passphrase:
Please re-enter passphrase:
set failed: The client is not authenticated with the Multipass service.
Please use 'multipass authenticate' before proceeding.
5. `multipass authenticate`
Stuck in that loop

The UI shows: `Waiting for daemon...`

(Note that I had a very early version of multipass installed, removed it as some point and recently reinstalled 1.15.1)

**Expected behavior**

Some error message explaining what is wrong.

**Logs**

```
[2025-06-06T08:38:54.966] [warning] [Qt] QNetworkDiskCache::prepare() unable to open temporary file
[2025-06-06T08:38:54.966] [debug] [update] Latest Multipass release available is version 1.15.1
[2025-06-06T08:38:54.968] [debug] [blueprint provider] Loading "anbox-cloud-appliance" v1
[2025-06-06T08:38:54.969] [debug] [blueprint provider] Loading "charm-dev" v1
[2025-06-06T08:38:54.970] [debug] [blueprint provider] Loading "docker" v1
[2025-06-06T08:38:54.970] [debug] [blueprint provider] Loading "jellyfin" v1
[2025-06-06T08:38:54.971] [debug] [blueprint provider] Loading "minikube" v1
[2025-06-06T08:38:54.971] [debug] [blueprint provider] Loading "ros2-humble" v1
[2025-06-06T08:38:54.971] [debug] [blueprint provider] Loading "ros2-jazzy" v1
[2025-06-06T08:38:54.978] [info] [rpc] gRPC listening on unix:/var/run/multipass_socket
[2025-06-06T08:38:54.978] [debug] [async task] fetch manifest periodically
[2025-06-06T08:38:54.983] [info] [VMImageHost] Did not find any supported products in "appliance"
[...]

PR #4157, opened 2025-06-12
[daemon] Improve authentication-related error msgs

ricab:
Addresses #4147.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2854,8 +2854,8 @@ try
         return status_promise->set_value(grpc::Status(
             grpc::StatusCode::FAILED_PRECONDITION,
             "Incorrect passphrase. No passphrase is set.\n\n"
-            "To authenticate trusted users, first ask an authenticated user to set a passphrase "
-            "and share it with them. To set a passphrase, use the command "
+            "To authenticate, first ask an authenticated user to set a passphrase and share it "
+            "with you, provided they trust you. To set a passphrase, use the command "
             "'multipass set local.passphrase'. Note that only the user who installs Multipass is "
             "automatically authenticated."));
     }


ricab:
Hey @tmerten, just created this one to improve those messages. WDYT?

ricab:
Thanks for the review @tmerten!

I am afraid that the Multipass daemon doesn't know which users are authenticated to it. It only gets a certificate from the connecting client and verifies it against a trusted set of certificates, but it doesn't have a notion of who it belongs to. Providing that would be a new feature.

Multipass is meant mostly for developer and personal machines, which are typically not multi-user to begin with. So, when this happens, it is usually because people install as root but connect with another user. We have a more expansive [how-to doc](https://documentation.ubuntu.com/multipass/en/latest/how-to-guides/customise-multipass/authenticate-clients-with-the-multipass-service/) for authentication, which I now updated slightly to hint at that most common scenario.

docs/how-to-guides/customise-multipass/authenticate-clients-with-the-multipass-service.md
```diff
@@ -67,7 +68,10 @@ Here you can find solutions and workarounds for common issues that may arise.
 
 ### The client cannot be authorised and the passphrase cannot be set
 
-It is possible that another client that is privileged to connect to the Multipass socket will connect first and make it seemingly impossible to set the `local.passphrase` and also `authorize` the client with the service.
+It is possible that another client that is privileged to connect to the Multipass socket will
+connect first and make it seemingly impossible to set the `local.passphrase` and also `authorize`
+the client with the service. This usually when Multipass is installed as root/admin but the client
```
sharder996: Typo:
> This usually when Multipass

ricab: Done

src/daemon/daemon.cpp
```diff
@@ -2851,10 +2851,13 @@ try
...
     {
-        return status_promise->set_value(
-            grpc::Status(grpc::StatusCode::FAILED_PRECONDITION,
-                         "Passphrase is not set. Please `multipass set local.passphrase` with a "
-                         "trusted client."));
+        return status_promise->set_value(grpc::Status(
+            grpc::StatusCode::FAILED_PRECONDITION,
+            "Incorrect passphrase. No passphrase is set.\n\n"
+            "To authenticate trusted users, first ask an authenticated user to set a passphrase "
+            "and share it with them. To set a passphrase, use the command "
```
sharder996: I feel like this sentence is a bit confusing grammatically in regard to pronouns. Using "them" doesn't make sense when "you" are the one experiencing the error. Hence, I think something like:

> To authenticate, first ask an authenticated user to set a passphrase and share it with you."

ricab: Yeah I went back and forth on this one... the thing is that I wanted to make it clear that the password should only be shared with trusted users. "Them" referred to "trusted users", so I think the pronoun is in grammatical agreement, but I guess you meant that it is strange that "you" ends up being is a third party, semantically.

How about we take your suggestion, but append "(provided s/he trusts you)"?

sharder996: I would go with something like:
> To authenticate, first ask an authenticated user to set a passphrase and share it with you, provided they trust you.

ricab: done

PR #4159, opened 2025-06-13
[test] Test macOS 13 on ARM

ricab:
Just to be sure... It's not included in https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners but it is included in https://github.com/actions/runner-images?tab=readme-ov-file#available-images so I just want to check.


ricab:
:crying_cat_face: https://github.com/canonical/multipass/actions/runs/15644481287/job/44079257134?pr=4159#step:9:29

PR #4171, opened 2025-06-18
[snap/snapcraft.yaml] add `gettext` to the build-packages

xmkg:
Qemu build fails with the following output in an arm64 environment:

[4/12] Generating po/bg/LC_MESSAGES/qemu-bg.mo with a custom command FAILED: po/bg/LC_MESSAGES/qemu.mo
/snap/gnome-42-2204-sdk/current/usr/bin/msgfmt -o po/bg/LC_MESSAGES/qemu.mo ../po/bg.po /snap/gnome-42-2204-sdk/current/usr/bin/msgfmt: error while loading shared libraries: libgettextsrc-0.21.so: cannot open shared object file: No such file or directory ninja: build stopped: subcommand failed.

The gettext package provides libgettextsrc-0.21.so.

MULTI-2036

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -286,6 +286,7 @@ parts:
     - libseccomp-dev
     - meson
     - python3-venv
+    - gettext
     stage-packages:
     - libpixman-1-0
     stage:


snap/snapcraft.yaml
```diff
@@ -221,6 +221,8 @@ parts:
     - -DMULTIPASS_UPSTREAM=origin
     - -DMULTIPASS_ENABLE_FLUTTER_GUI=on
     override-build: |
+      # Bootstrap vcpkg explicitly so we can see the bootstrap log
+      ${CRAFT_PART_SRC}/3rd-party/vcpkg/bootstrap-vcpkg.sh
```
ricab: Isn't CMake a better place for this though? We can then run on all builds, not just snapcraft. Just need to call the right script depending on the platform.

What I used to debug should be close to what we need, so maybe I push that in another PR. WDYT?

xmkg: Yeah, we could do that, but it does not add much value to local builds since we can poke around freely. For GitHub, I'd rather have it as a dedicated step anyway, which leaves us with Snapcraft. 

I don't mind having it either way, so let me know how you'd prefer it.

ricab: For the record, we discussed and I ended up submitting https://github.com/canonical/multipass/pull/4173

PR #4177, opened 2025-06-23
[utils] enhance YAML emission with recursive handling

### LINKED ISSUE #4176: Cloud-init permissions field type incorrectly parsed as integer (e.g., 0755 → 493)
**Description**

When launching a Multipass instance with a valid cloud-init YAML config that includes a `write_files` section with:

```yaml
permissions: "0755"
```

Multipass appears to convert this value into an **integer** (`493`) when it hands off the config to the VM, resulting in a cloud-init schema validation warning or failure:

```
Cloud config schema errors: write_files.0.permissions: 493 is not of type 'string'
```

Even though the original YAML correctly specifies `"0755"` (as a string), the cloud instance receives:

```yaml
permissions: 493
```

This results in the launch command failing for a timeout.

---

**Steps to Reproduce**

1. Launch a multipass instance using the charm-dev cloud-init:
```bash
multipass launch --name cdo --cpus 4 --memory 8G --disk 50G --cloud-init https://raw.githubusercontent.com/canonical/multipass/refs/heads/main/data/cloud-init-yaml/cloud-init-charm-dev.yaml
```

2. Wait until command terminates or times out
```bash
```

3. Enter the instance and inspect the cloud init schemas:
```bash
multipass shell cdo
sudo cloud-init schema --system
  Found cloud-config data types: user-data, vendor-data, network-config

  1. user-data at /var/lib/cloud/instances/cdo/cloud-config.txt:
    Invalid user-data /var/lib/cloud/instances/cdo/cloud-config.txt
    Error: Cloud config schema errors: write_files.0.permissions: 493 is not of type 'string'
  ...
```
> Cloud-init logs also show a schema validation warning related to this field.

5. Inspecting the cloud config reveals the file does in fact contain an integer rather than the originally specificed string:
```bash
sudo cat /var/lib/cloud/instances/cdo/cloud-config.txt | grep permission
    permissions: 493
```

**Expected Behavior**

Multipass should preserve string types in YAML, especially for fields like `permissions`, which cloud-init strictly expects to be strings (e.g., `"0644"`, `"0755"`). There should be no implicit conversion to integers.

---

**Workaround**

Manually force the value to be treated as a string using YAML type tags:

```yaml
permissions: !!str "0755"
```

But this workaround shouldn't be necessary for already well-formed YAML.

---

**Environment**

* Multipass version: 1.15.1
* Host OS: 24.04 (Canonical corporate laptop)
* Guest OS: Ubuntu 24.04 (default)
* Cloud-init version in guest: (optional)

---

**Suggested Fix**

Ensure the YAML parser used in Multipass preserves string types for cloud-init passthrough and does not interpret values like `"0755"` into integers.

---

levkropp:
fixes #4176 

This pull request refactors the `emit_yaml` function in `src/utils/yaml_node_utils.cpp` to handle YAML nodes recursively, allowing for the addition of special handling, which is needed to properly handle scalar strings that resemble octal numbers.

Enhancements to YAML node handling:

* Refactored `std::string mp::utils::emit_yaml(const YAML::Node& node)` to introduce a recursive helper function (`emit_node`) for processing YAML nodes. This allows the function to handle `Map`, `Sequence`, and scalar nodes more robustly.
* Added special handling for scalar strings that look like octal numbers (e.g., "0755") by tagging them with `YAML::LocalTag("str")` to preserve their string representation in the emitted YAML.

These changes seem necessary due to how `yaml-cpp` handles re-emission of YAML input. If the YAML input uses quotes ("0755"), the YAML spec says this is a string and should remain a string, however it appears that `yaml-cpp` is seeing a scalar "0755", and then writing it out in a way that causes it to be emitted as a number. These changes should not affect anything with regards to YAML parsing other than this particular case where a cloud-init file has a `write_files` section with a `permissions` string defined.


src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
 {
     YAML::Emitter emitter;
     emitter.SetIndent(2);
-    emitter << node;
+
+    // Helper function to handle nodes recursively
+    std::function<void(const YAML::Node&)> emit_node;
+    emit_node = [&emitter, &emit_node](const YAML::Node& n) {
```
xmkg, line 64: Hi @levkropp, just out of curiosity -- Can we use this approach to let's say, enforce quotation of string values containing a specific character (e.g. colon `:`). I'm asking this because if it's so, we can drop the custom YAML-CPP patch we have and use the upstream as-is, one less custom repository and patch to worry about.

xmkg: The patch I'm talking about is this: https://github.com/canonical/yaml-cpp/commit/d96aa990060b147e4e2b296c92491d039a5a6bd7

levkropp:
I have confirmed that:

- Our personal fork's changes to `yaml-cpp` are not responsible for these issues with processing permissions octals
- The latest upstream of `yaml-cpp` still has these issues with octal scalars
- We can implement our existing necessary fixes to colons directly into `emit_yaml` and no longer need to use our own fork of `yaml_cpp`  (our fork's changes only is necessary for strings with spaces after colons (e.g. `key: value`) which we can indeed handle in emit_yaml)

HOWEVER: I am getting issues with timeouts, even though the cloud-init configuration completes successfully. For some reason on the latest `yaml-cpp` there are issues on the SSH or gRPC level regarding communication. The `multipass exec` command we run to check if a cloud-init config succeeded works when run manually, but it never gets called

levkropp:
Because of the issues I have been seeing with updating to the latest upstream of `yaml-cpp`, I have reverted the PR back to the single commit which implements the fix for #4176 , and have created a new branch to deal with investigations into what is necessary to make everything work without the fork's changes. I will create a separate PR to deal with updating `yaml-cpp` to upstream.

```
~/yaml-cloudinit-octal-parsing/build$ ./bin/multipass launch --name cdo --cpus 4 --memory 8G --disk 50G --cloud-init https://raw.githubusercontent.com/canonical/multipass/refs/heads/main/data/cloud-init-yaml/cloud-init-charm-dev.yaml
Launched: cdo
~/yaml-cloudinit-octal-parsing/build$ ./bin/multipass shell cdo
Welcome to Ubuntu 24.04.2 LTS (GNU/Linux 6.8.0-60-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/pro

 System information as of Mon Jul  7 11:31:25 EDT 2025

  System load:  0.22              Processes:             133
  Usage of /:   3.8% of 47.39GB   Users logged in:       0
  Memory usage: 3%                IPv4 address for ens3: 10.44.19.6
  Swap usage:   0%


Expanded Security Maintenance for Applications is not enabled.

85 updates can be applied immediately.
65 of these updates are standard security updates.
To see these additional updates run: apt list --upgradable

Enable ESM Apps to receive additional future security updates.
See https://ubuntu.com/esm or run: sudo pro status


To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

ubuntu@cdo:~$ sudo cloud-init schema --system
Found cloud-config data types: user-data, vendor-data, network-config

1. user-data at /var/lib/cloud/instances/cdo/cloud-config.txt:
  Valid schema user-data

2. vendor-data at /var/lib/cloud/instances/cdo/vendor-cloud-config.txt:
Cloud config schema deprecations: system_info:  Deprecated in version 24.2. System and/or distro specific settings. This is not intended to be overridden by user data or vendor data.
  Valid schema vendor-data

3. network-config at /var/lib/cloud/instances/cdo/network-config.json:
  Valid schema network-config
ubuntu@cdo:~$ 

```

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+        default:
+        {
+            // Special handling for strings that look like octal numbers (e.g. "0755")
+            if (n.IsScalar())
+            {
+                std::string value = n.Scalar();
+                if (value.length() >= 2 && value[0] == '0' &&
+                    std::all_of(value.begin() + 1, value.end(), ::isdigit))
+                {
+                    emitter << YAML::LocalTag("str") << value;
```
xmkg, line 101: Codecov is claiming that this line is not covered by the unit tests. Would you mind adding a case for it?

src/utils/yaml_node_utils.cpp
xmkg, line 64: Nice trick for handling self-call in a lambda 👍

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+            }
+            emitter << YAML::EndSeq;
+            break;
+        }
+        default:
+        {
+            // Special handling for strings that look like octal numbers (e.g. "0755")
+            if (n.IsScalar())
+            {
+                std::string value = n.Scalar();
```
xmkg:
```suggestion
                const std::string value = n.Scalar();
```

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+                if (value.length() >= 2 && value[0] == '0' &&
```
xmkg, line 98: This could use a comment.

src/utils/yaml_node_utils.cpp
xmkg, line 101:
```suggestion
                    emitter << YAML::LocalTag("str") << value;
                    return;
```

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+                    std::all_of(value.begin() + 1, value.end(), ::isdigit))
+                {
+                    emitter << YAML::LocalTag("str") << value;
+                }
+                else
+                {
+                    emitter << n;
+                }
```
xmkg:
```suggestion
             
```

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+            }
+            else
+            {
+                emitter << n;
+            }
```
xmkg:
```suggestion
                emitter << n;
```

xmkg (changes requested):

src/utils/yaml_node_utils.cpp
```diff
@@ -58,7 +58,62 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
...
+        case YAML::NodeType::Map:
+        {
+            emitter << YAML::BeginMap;
+            for (const auto& kv : n)
+            {
+                emitter << YAML::Key;
+                emit_node(kv.first);
+                emitter << YAML::Value;
+                emit_node(kv.second);
+            }
```
xmkg, line 76:
```suggestion
            for (const auto& [k,v] : n)
            {
                emitter << YAML::Key;
                emit_node(k);
                emitter << YAML::Value;
                emit_node(v);
            }
```

levkropp: This change does not compile

```
yaml-cloudinit-octal-parsing/src/utils/yaml_node_utils.cpp: In lambda function:
yaml-cloudinit-octal-parsing/src/utils/yaml_node_utils.cpp:70:30: error: cannot decompose inaccessible member ‘YAML::Node::m_isValid’ of ‘YAML::Node’
   70 |             for (const auto& [k,v] : n)
      |                              ^~~~~
      
```

It appears that structured bindings cannot be used with YAML::Node iterators because the underlying type does not support public tuple-like access.

xmkg (approved):
LGTM 👍

Sploder12 (approved):
LGTM!

PR #4181, opened 2025-06-26
[gui] update to flutter 3.32.4

sharder996:
Mostly changes due to new linting rules in `dart format`

### DIFF
--- a/src/client/gui/windows/CMakeLists.txt
+++ b/src/client/gui/windows/CMakeLists.txt
@@ -8,7 +8,7 @@ set(BINARY_NAME "multipass.gui")
 
 # Explicitly opt in to modern CMake behaviors to avoid warnings with recent
 # versions of CMake.
-cmake_policy(SET CMP0063 NEW)
+cmake_policy(VERSION 3.14...3.25)
 
 # Define build configuration option.
 get_property(IS_MULTICONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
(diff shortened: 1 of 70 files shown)


src/client/gui/windows/CMakeLists.txt
```diff
@@ -86,6 +87,14 @@ if(PLUGIN_BUNDLED_LIBRARIES)
...
 endif()
 
+# Copy the native assets provided by the build.dart from all packages.
+set(NATIVE_ASSETS_DIR "${PROJECT_BUILD_DIR}native_assets/windows/")
+if(EXISTS "${NATIVE_ASSETS_DIR}")
+  install(DIRECTORY "${NATIVE_ASSETS_DIR}"
+    DESTINATION "${INSTALL_BUNDLE_LIB_DIR}"
+    COMPONENT Runtime)
+endif()
+
```
ricab, line 97: Not a review, but is this to workaround that flutter build issue on Windows?

sharder996: I don't believe so. I went through the upgrade process as described in Andrei's handover doc while trying to preserve as much of Andrei's original work as possible, but I never encountered that bug as described in [#165200](https://github.com/flutter/flutter/issues/165200). I can replicate the behaviour as described there, but didn't encounter it while upgrading.

The code block you referenced originally came from [here](https://github.com/canonical/multipass/pull/3980/commits/00cc7d3a217264a70228d7328a56fc9a04473b17#diff-93fbf113286cc367a2a7c02a04e86f1ea855f61e2f02d1db6ff86cf926f0c842R90-R95). I'm not 100% sure why Andrei added it, but my best guess is that depending on what dependencies one has in their Flutter project, `dart` can generate `native_assets`. I did not notice any of these assets being generated from our project so I modified the `install` to only conditionally include the directory. Otherwise, the build process fails due to the directory not existing.

I am still doing some final verification, but so far I have not noticed any change in the operation of the GUI.

sharder996:
@xmkg I will rerun linting commands, fix merge conflicts, and rebase once you've reviewed.

PR #4183, opened 2025-06-26
Minimun resources per instance

### LINKED ISSUE #3777: Unify VM resource minima in Daemon/CLI/GUI
We currently have a minimum of 512M for memory in the GUI, but 128M in CLI. Similar situation with disk: current minima are 512M in the CLI but 1G in the GUI.

We should probably settle on a single value everywhere. Since the images themselves don't have documented minimum requirements, we should probably go with the safest boundary and so align CLI/Daemon with the GUI.

**Additional context**
This should prevent problems like https://github.com/canonical/multipass/issues/3612.

vicdeveloperr:
Fixes #3777.


xmkg:
Hi @vicdeveloperr, thanks for the contribution!

Could you elaborate a bit about how you established the values for the minimum resources?

vicdeveloperr:
> Hi @vicdeveloperr, thanks for the contribution!
> 
> Could you elaborate a bit about how you established the values for the minimum resources?

Yes. I already update my first comment to include the explanation.

sharder996:
We might want to align with [Ubuntu Server - System Requirements](https://documentation.ubuntu.com/server/reference/installation/system-requirements/) and [Ubuntu Core - System Requirements](https://documentation.ubuntu.com/core/reference/system-requirements/).

ricab:
Those are great finds @sharder996! We should refuse anything below what they point out there in both GUI and the daemon, IMO. And we should probably link to that from our documentation

PR #4184, opened 2025-06-26
[utils] remove multipass::Path

levkropp:
This pull request refactors the multipass codebase by removing the custom Path type (an alias for QString) and using QString directly throughout the code. This streamlines the codebase and eliminates the unnecessary Path abstraction, which affects e.g. Intellisense when dealing with strings, especially in areas like conversions to/from `std::string`. By standardizing on QString, the code becomes more consistent and easier to maintain.


xmkg:
@levkropp my 2c's here; I think we should keep the `multipass::Path` alias until we replace it with a proper path type, i.e., (std::filesystem::path). Right now, we at least know that a variable is a Path by looking at the type, because the alias is named in such a way. If we go ahead and remove it, we would lose this info, and it would be harder to refactor this later on.

Hence, I'm against removing it, unless replaced with `std::filesystem::path`.

levkropp:
> Right now, we at least know that a variable is a Path by looking at the type, because the alias is named in such a way. If we go ahead and remove it, we would lose this info

I disagree with this because every variable name that uses `Path` is suffixed with `_path` or `_dir` or in some cases even `_dir_path`

I'd be fine with keeping Path if we address the variable naming conventions (we know the variable is a Path by looking at the type), and remove the inconsistent suffixes

xmkg:
> > Right now, we at least know that a variable is a Path by looking at the type, because the alias is named in such a way. If we go ahead and remove it, we would lose this info
> 
> I disagree with this because every variable name that uses `Path` is suffixed with `_path` or `_dir` or in some cases even `_dir_path`

What I'm saying is it would be much easier to locate instances of `multipass::Path` than an arbitrary suffix. I'm not against removing the suffixes since they're mostly useful when the type itself is not descriptive enough. I'd prefer to see them gone first than the type alias.

PR #4194, opened 2025-06-30
Fix client key permissions

### LINKED ISSUE #4193: 'C:/Users/USERNAME/AppData/Local/multipass-client-certificate/multipass_cert_key.pem': Permission denied(13)
FYI this error occurred (every time) when installing Multipass 1.16.0 RC (v1.16.0-rc3) on a clean Windows 11 Pro:

```
C:\Users\USERNAME>multipass version
[error] [client] Caught an unhandled exception: failed to open file 'C:/Users/USERNAME/AppData/Local/multipass-client-certificate/multipass_cert_key.pem': Permission denied(13)
```

FYI the error did _not_ occur when installing Multipass 1.16.0 RC (v1.16.0-rc3) on top of (a prior install) of Multipass 1.15.1

Sploder12:
There was a change in logic in when permissions are set for ssl certificates which caused newly created client certificates to have overly restrictive permissions. This fixes that.

Also there was a bug where unix permissions weren't converted to Windows permissions correctly. This fixes that.

Setting the permissions of the file to read by everyone does not change security since the parent folder maintains the user only access. However, in the future it would be nice to give read permissions directly to the user instead of everyone.

closes #4193

### DIFF
--- a/src/platform/platform_win.cpp
+++ b/src/platform/platform_win.cpp
@@ -460,7 +460,7 @@ DWORD set_specific_perms(LPSTR path, WELL_KNOWN_SID_TYPE sid_type, DWORD access_
 
 DWORD convert_permissions(int unix_perms)
 {
-    if (unix_perms & 07)
+    if (unix_perms == 07)
         return GENERIC_ALL;
 
     DWORD access_mask = 0;
--- a/src/cert/ssl_cert_provider.cpp
+++ b/src/cert/ssl_cert_provider.cpp
@@ -388,6 +388,11 @@ mp::SSLCertProvider::KeyCertificatePair make_cert_key_pair(const QDir& cert_dir,
         client_cert_key.write(priv_key_path);
         client_cert.write(cert_path);
 
+        MP_PLATFORM.set_permissions(priv_key_path.toStdU16String(),
+                                    std::filesystem::perms::owner_all |
+                                        std::filesystem::perms::group_read |
+                                        std::filesystem::perms::others_read);
+
         return {client_cert.as_pem(), client_cert_key.as_pem()};
     }
 }


ricab:
How come this affected only Windows? Or was that not so?

Sploder12:
> How come this affected only Windows? Or was that not so?

I never checked if it affected the other platforms, I wouldn't be surprised if it did

ricab (changes requested):
I tried this and it worked fine on a pristine install on Windows, but read-by-all private key?

Sploder12:
> I tried this and it worked fine on a pristine install on Windows, but read-by-all private key?

Yep, unfortunately that's how it was and as far as I know how it needs to be without a big change in the clients.

holta:
> I tried this and it worked fine on a pristine install on Windows, but read-by-all private key?

The failure was repeated for me: hypothetically (might) also relate to residue left over from 1.15.1 and/or 1.16 RC?

Sploder12:
> The failure was repeated for me: hypothetically (might) also relate to residue left over from 1.15.1 and/or 1.16 RC?

This fix doesn't affect existing files so residue from 1.16 RC would still cause this error

xmkg:
> There was a change in logic in when permissions are set for ssl certificates which caused newly created client certificates to have overly restrictive permissions. This fixes that.
> 
> Setting the permissions of the file to read by everyone does not change security since the parent folder maintains the user only access. However, in the future it would be nice to give read permissions directly to the user instead of everyone.
> 
> closes #4193

@Sploder12 thanks for the fix!

If I understand it correctly, we're basically falling back to the parent folder's permissions, right? The user who installed the multipass would naturally have the proper permissions to the parent folder;  therefore, they can access the file, but not the others.

Sploder12:
> If I understand it correctly, we're basically falling back to the parent folder's permissions, right? The user who installed the multipass would naturally have the proper permissions to the parent folder; therefore, they can access the file, but not the others.

Yep!

src/platform/platform_win.cpp
```diff
@@ -460,7 +460,7 @@ DWORD set_specific_perms(LPSTR path, WELL_KNOWN_SID_TYPE sid_type, DWORD access_
 
 DWORD convert_permissions(int unix_perms)
 {
-    if (unix_perms & 07)
+    if (unix_perms == 07)
```
ricab, line 463: Why does this need to receive an int, rather than the original perms? All of `set_permissions` should be able to use constants therein, rather than magic numbers, no? Anyway, something for another time.

ricab (approved):
Alright, I've verified this on Linux and Windows and all looks good now. Permissions on the root certificate, private keys, and client certificates all look correct now. 

image

image

ricab:
@xmkg I am going ahead and merging this, to speed things up, but if you could still review it after the fact (either here or in a new RC), I'd appreciate it. Thanks!

Issue #4195, opened 2025-06-30
[sshfs] Multipass logs leaking filepaths from CI environments


sharder996:
In several compilation units, we are using the `__FILE__` macro in logging messages. This is causing Multipass to leak file paths from our CI environment. i.e. `[trace] [ssh process] /Users/cibot/actions-runner-2/_work/multipass-private/multipass-private/src/ssh/ssh_process.cpp:167 read_stream(type = 0, timeout = -1)`.

We should not be printing out the full path for the file. This part of the log message should either be removed or truncated to only include the top level `src/` directory.

prachimaskar:
Hi @sharder996 👋

I’ve created a PR that replaces `__FILE__` with `__FILENAME__` in `SSHProcess::read_stream` log statements to avoid leaking full file paths from CI environments.

PR: #4196

✅ Only the filename appears now in logs  
✅ No functional change — only improves log privacy & readability

Please let me know if more files need this change or if further updates are needed. Thank you!

mmustafasenoglu:
Hi @sharder996 👋

I looked into this issue and it appears to already be resolved in the current codebase.

The `extract_filename` function in `src/logging/log_location.cpp` uses `std::string_view::find_last_of` to strip the full path down to just the basename:

```cpp
std::string_view extract_filename(std::string_view path)
{
    assert(path.empty() || (*path.rbegin() != '/' && *path.rbegin() != '\\'));
    auto pos = path.find_last_of("/\\\\");
    return pos == std::string_view::npos ? path : path.substr(pos + 1);
}
```

And `include/multipass/logging/log_location.h` uses `std::source_location` with `extract_filename` to ensure only the filename (not the full CI path) appears in log output:

```cpp
detail::extract_filename(location.file_name()),
```

I also confirmed there are no remaining uses of the `__FILE__` macro anywhere in the codebase. The `trace_location`, `debug_location`, and `log_location` functions all go through `extract_filename`.

It seems like this was addressed as part of the C++20 migration. This issue can probably be closed. :slightly_smiling_face:

sharder996:
Thanks for catching this, @mmustafasenoglu!

Yes, I see it got fixed awhile ago, the tagging in GitHub just got lost between opening and closing subsequent PRs.

mmustafasenoglu:
You're welcome! Happy to hear it was already resolved. Should we close this PR then?

PR #4196, opened 2025-07-01
Fix: Handle SSHProcess exit code and timeout safely

prachimaskar:
This PR fixes the log file path output in SSHProcess::read_stream by replacing __FILE__ with __FILENAME__, so only the file name (not full path) appears in log messages.

✅ Improves readability of logs
✅ Helps developers debug faster
✅ No functional behavior change — log formatting only

Please let me know if anything else needs to be updated. Thanks for reviewing!

### DIFF
--- a/src/ssh/ssh_process.cpp
+++ b/src/ssh/ssh_process.cpp
@@ -25,6 +25,8 @@
 #include <libssh/callbacks.h>
 
 #include <array>
+#define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
+
 #include <cerrno>
 #include <cstring>
 #include <sstream>
@@ -177,7 +179,7 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
     mpl::log(mpl::Level::trace,
              category,
              fmt::format("{}:{} {}(type = {}, timeout = {}): ",
-                         __FILE__,
+                         __FILENAME__,
                          __LINE__,
                          __FUNCTION__,
                          static_cast<int>(type),
@@ -187,7 +189,7 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
     {
         mpl::log(mpl::Level::trace,
                  category,
-                 fmt::format("{}:{} {}(): channel closed", __FILE__, __LINE__, __FUNCTION__));
+                 fmt::format("{}:{} {}(): channel closed", __FILENAME__, __LINE__, __FUNCTION__));
         return std::string();
     }
 
@@ -205,7 +207,7 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
         mpl::log(
             mpl::Level::trace,
             category,
-            fmt::format("{}:{} {}(): num_bytes = {}", __FILE__, __LINE__, __FUNCTION__, num_bytes));
+            fmt::format("{}:{} {}(): num_bytes = {}", __FILENAME__, __LINE__, __FUNCTION__, num_bytes));
         if (num_bytes < 0)
         {
             // Latest libssh now returns an error if the channel has been closed instead of
@@ -215,7 +217,7 @@ std::string mp::SSHProcess::read_stream(StreamType type, int timeout)
                 mpl::log(
                     mpl::Level::trace,
                     category,
-                    fmt::format("{}:{} {}(): channel closed", __FILE__, __LINE__, __FUNCTION__));
+                    fmt::format("{}:{} {}(): channel closed", __FILENAME__, __LINE__, __FUNCTION__));
                 return output.str();
             }
 


prachimaskar:
Hi @ricab 👋

This PR is ready for review.  
Could you please approve the workflows so the required checks can run?

Let me know if you'd like me to update or improve anything further.

Thanks a lot!

src/ssh/ssh_process.cpp
```diff
@@ -25,6 +25,8 @@
 #include <libssh/callbacks.h>
 
 #include <array>
+#define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
```
ricab: I am afraid that this is specific to GCC, we'd need a generic approach instead.

We're planning to move to C++20 soon. There you could use https://en.cppreference.com/w/cpp/utility/source_location.html instead.

I just created https://github.com/canonical/multipass/pull/4197. Only a draft for now, but you could base your PR on top of that one and we could merge when ready.

ricab (changes requested):
Hi @prachimaskar, thanks for the PR! I am afraid the current implementation isn't generic enough, but see my proposal below :slightly_smiling_face:

src/ssh/ssh_process.cpp
prachimaskar: Hi @ricab, thank you for the feedback! 🙏

You're absolutely right — the macro I used is GCC-specific and not portable. Using `std::source_location` from C++20 is a much better and cleaner approach.

I’ve checked PR #4197 and will update this PR accordingly based on that design.

Thanks again for the guidance — I’ll push the changes soon!

ricab: Well, that still gives the full path (sorry I missed that at first)... But you could use `string_view::find_last_of` in a consteval function to extract the component.

prachimaskar: Hi @ricab 👋

I’ve updated the PR to use `std::source_location` with a `consteval` function to extract only the filename from the path, as suggested. 

Thanks for your detailed feedback — please let me know if you’d like any more changes 🙏

src/ssh/ssh_process.cpp
```diff
@@ -14,7 +14,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
+#include <source_location>
```
xmkg: Not a thorough review, but `<source_location>` is a C++20 feature, where we're still on C++17. We're planning to switch to C++20 though: https://github.com/canonical/multipass/pull/4197

xmkg: Nevermind, this has already been discussed above.

ricab:
For future reference, this is blocked by the move to C++20.

prachimaskar:
Hi @ricab 👋

I updated this PR to use std::source_location + consteval for filename stripping as you suggested. It still shows “changes requested” but I guess that’s only because project isn’t on C++20 yet.

Since it’s blocked on the migration (#4197), I’ll wait for that to land and then rebase. Thanks a lot for your guidance 🙏 Please let me know if I should keep this updated in the meantime.

ricab (commented):
Hi @prachimaskar! It is looking good, but  I'll need to do another pass on this once we actually have C++20, to confirm it all works nicely. We're all pretty busy and couldn't do that yet, but I don't see anything else at this point. 

If you want you can rebase on the `cpp20` branch, but you don't have to. Don't worry if you need to forget about this for now, we can always pick up ourselves.

Thanks again :slightly_smiling_face:

ricab:
PS: I am not sure why it keeps showing me as "requested changes" but, to be clear, I don't have any further requests at this time.

prachimaskar:
Hi @ricab 👋
Got it, thanks for confirming 🙂 I’ll wait for cpp20 migration and rebase then. Appreciate your help 🙏

sharder996:
@prachimaskar There is a [PR#4378](https://github.com/canonical/multipass/pull/4378) up for migrating to C++20. If you like, you can rebase on top of it and target it in this PR.

sharder996:
Fixes #4195

ricab:
Hey @xmkg, I just noticed that this got closed. I guess it happened automatically when https://github.com/canonical/multipass/pull/4378 was merged, right? Should we reopen?

xmkg:
Hey @ricab, I have no clue how that happened, that was not intentional. I'll reopen this.

xmkg:
@ricab I see what's going on. This branch is targeted to the C++20 branch, which is now deleted, orphaning this branch, and apparently, GitHub's approach is just to close the orphaned PR. The web UI doesn't allow reopening or retargeting the PR so I guess we would have to open a new PR for this. 🤷

prachimaskar:
Thanks @ricab and @xmkg 🙂
I saw that this PR got closed automatically after the c++20 branch was deleted.
I’ll rebase my fix-log-filepath-leak branch on main and open a new PR with the same changes (replacing #4196).
I’ll tag you both there once it’s up. 🙏

Issue #3354, opened 2024-01-04
[hyperv] `multipass ls` shows stale IP address after instance restart


holta:
After an instance is rebooted on Windows, `multipass list` and `multipass info` both display the instance's original/stale (erroneous!) IP address on top:

```
C:\Users\XXXX>multipass version
multipass   1.13.0-rc.1313+g66f773628.win
multipassd  1.13.0-rc.1313+g66f773628.win

C:\Users\XXXX>multipass get local.driver
hyperv

C:\Users\XXXX>multipass launch -n primary
Launched: primary
Skipping mount due to disabled mounts feature                                  # ⚠️ SHOULD WE IGNORE THIS WARNING? ⚠️

C:\Users\XXXX>multipass exec primary -- hostname -I
192.168.176.167

C:\Users\XXXX>multipass list
Name                    State             IPv4             Image
primary                 Running           192.168.176.167  Ubuntu 22.04 LTS    # ✅ ACCURATE IP ADDRESS ✅

C:\Users\XXXX>multipass restart

C:\Users\XXXX>multipass exec primary -- hostname -I
192.168.178.108

C:\Users\XXXX>multipass list
Name                    State             IPv4             Image
primary                 Running           192.168.176.167  Ubuntu 22.04 LTS    # ❌ STALE/ERRONEOUS IP ADDRESS ❌
                                          192.168.178.108                      # ✅ ACCURATE IP ADDRESS ✅

C:\Users\XXXX>multipass restart

C:\Users\XXXX>multipass exec primary -- hostname -I
192.168.181.12

C:\Users\XXXX>multipass list
Name                    State             IPv4             Image
primary                 Running           192.168.176.167  Ubuntu 22.04 LTS    # ❌ STALE/ERRONEOUS IP ADDRESS ❌
                                          192.168.181.12                       # ✅ ACCURATE IP ADDRESS ✅
```

WHY THIS HURTS: Newcomers and students get very confused, as we ask them to use the Multipass instance's IP Address and they naturally try the (top-most) STALE/ERRONEOUS IP Address.

UGLY WORKAROUND: **Start > Run > services.msc > Multipass Service > Restart** works!  (To stop the display of the stale/erroneous IP address.)

QUESTION: We would rather not ask everyday users and newcomers to Restart "Multipass Service" !  Is the underlying bug possibly solvable in coming months?

[...]

PR #4207, opened 2025-07-02
[hyper-v] Invalidate cached mgmt IP address on update_state()

xmkg:
The cached management IP address becomes stale on some operations when it happens out-of-band, like "multipass restart", where it's a SSH-induced guest reboot.

This patch fixes that by invalidating the cached IP address on every update_state() invocation.

Fixes: #3354

MULTI-2068

### DIFF
--- a/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
+++ b/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
@@ -440,6 +440,21 @@ void mp::HyperVVirtualMachine::ensure_vm_is_running()
 
 void mp::HyperVVirtualMachine::update_state()
 {
+    // Invalidate the management IP address on state update.
+    {
+        // The reason "why" is as follows:
+        // - The daemon starts and constructs the VM
+        // - The mgmt IP address gets cached on the first invocation of management_ipv4()
+        // - `multipass restart` initiates a guest reboot with "sudo reboot"
+        // -  It's an "out-of-band" reboot in perspective of Hyper-V backend since
+        //    it does not trigger the "on_restart" function afterwards.
+        // - The cached "vm->ip address is not reset
+        // - management_ipv4() continues to report the stale IP address from the previous boot
+        mpl::log(mpl::Level::debug,
+                 vm_name,
+                 "Invalidating cached mgmt IP address upon state update");
+        ip = std::nullopt;
+    }
     monitor->persist_state_for(vm_name, state);
 }
 
--- a/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
+++ b/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
@@ -442,14 +442,8 @@ void mp::HyperVVirtualMachine::update_state()
     // Invalidate the management IP address on state update.
     if (current_state() == VirtualMachine::State::running)
     {
-        // The reason "why" is as follows:
-        // - The daemon starts and constructs the VM
-        // - The mgmt IP address gets cached on the first invocation of management_ipv4()
-        // - `multipass restart` initiates a guest reboot with "sudo reboot"
-        // -  It's an "out-of-band" reboot in perspective of Hyper-V backend since
-        //    it does not trigger the "on_restart" function afterwards.
-        // - The cached "vm->ip address is not reset
-        // - management_ipv4() continues to report the stale IP address from the previous boot
+        // Cached IPs become stale when the guest is restarted from within. By resetting them here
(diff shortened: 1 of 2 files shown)


src/platform/backends/hyperv/hyperv_virtual_machine.cpp
```diff
@@ -453,6 +453,21 @@ void mp::HyperVVirtualMachine::ensure_vm_is_running()
...
+        // - The mgmt IP address gets cached on the first invocation of management_ipv4()
+        // - `multipass restart` initiates a guest reboot with "sudo reboot"
+        // -  It's an "out-of-band" reboot in perspective of Hyper-V backend since
+        //    it does not trigger the "on_restart" function afterwards.
+        // - The cached "vm->ip address is not reset
+        // - management_ipv4() continues to report the stale IP address from the previous boot
+        mpl::log(mpl::Level::debug,
+                 vm_name,
+                 "Invalidating cached mgmt IP address upon state update");
+        ip = std::nullopt;
```
ricab: There is also `VirtualMachine::management_ip`, unfortunately. Should we reset that one too? And should we do so in the BaseVM?

I know we discussed disregarding deprecated backends, but I am wary of letting those two things going out of sync. `management_ip` should be updated once a new IP is obtained, but I am thinking we should reduce the slant as much as possible. However, I didn't try it, so I don't know if it would have any other undesired effects. 

Of course we shouldn't have the two fields to begin with, but that's going to get fixed in your new implementation or my refactor.

ricab: Also, are we sure it is OK to do this on *any* state update? To address `multipass restart` we would need this when we get to `running`, but probably not when we're `restarting`, `suspending` or in `delayed_shutdown`, right? Should we limit this to the running state?

src/platform/backends/hyperv/hyperv_virtual_machine.cpp
```diff
@@ -453,6 +453,21 @@ void mp::HyperVVirtualMachine::ensure_vm_is_running()
...
+    // Invalidate the management IP address on state update.
+    {
+        // The reason "why" is as follows:
+        // - The daemon starts and constructs the VM
+        // - The mgmt IP address gets cached on the first invocation of management_ipv4()
+        // - `multipass restart` initiates a guest reboot with "sudo reboot"
+        // -  It's an "out-of-band" reboot in perspective of Hyper-V backend since
+        //    it does not trigger the "on_restart" function afterwards.
+        // - The cached "vm->ip address is not reset
+        // - management_ipv4() continues to report the stale IP address from the previous boot
```
ricab: I find this is a bit long, and I believe the top comment is pretty clear in the code. How about this instead?

```suggestion
    {
        // Cached IPs become stale when the guest is restarted from within. By resetting them here
        // we at least cover multipass's restart initiatives, which include state updates.
```

This makes it clear what cases this code helps with.

ricab (commented):
Many thanks for (re-)spotting and fixing this @xmkg! 

The general idea makes sense to me, although in the long run I hope we can find a way to be alerted to ACPI events in the guest. I suspect it would be possible to subscribe over a VSOCK/SSH connection, otherwise we might use a heartbeat based approach.

I have a few requests inline, but I think we should let this one simmer in main for a while either way. It's been like this forever and the fix could have effects that we're not anticipating, so I wouldn't rush to release it now.

src/platform/backends/hyperv/hyperv_virtual_machine.cpp
xmkg: hyperv_vm->ip is functionally equivalent to base->management_ip. I don't know why there's a separate variable for it in the first place. Removed the Hyper-V's IP in favor of management_ip.

Regarding the invalidation upon state change, yeah, it might make sense to limit it to running.

src/platform/backends/hyperv/hyperv_virtual_machine.cpp
xmkg: I don't mind either way; the backend is going to get nuked anyway. I'll update it.

Sploder12:
This is important for hostname resolution, is it done simmering?

xmkg:
> This is important for hostname resolution, is it done simmering?

@ricab ^^ WDYT? Should we include it in 1.17?

xmkg:
Great! @ricab, we already discussed this, but still want to confirm that you're on board with the idea of ignoring commit linter for older PR's and commits. If so, I'll go ahead and merge this.

ricab:
Yes, it's fine. It's a transition period.

Issue #4192, opened 2025-06-30
Feature Request: GUI Toolbar for Copy/Paste/Other Operations


ghost:
Reference: #4180 

As an alternative or complement to the right-click context menu, introduce a small graphical toolbar at the top of the Multipass shell window with buttons for common clipboard actions:

- Copy

- Paste

- Select All

- ...

This would be especially helpful for less technical users or those unfamiliar with terminal keyboard shortcuts. A simple, visible UI element can significantly improve usability and discoverability of clipboard features:

- Offers a more accessible interface for new or non-technical users

- Complements existing terminal interactions (keyboard, context menu)

- Makes clipboard operations obvious and one-click away

vikorama:
Thank you @alexio-dev, good point to surface these essential actions more prominently for beginner users.  You [mentioned that Mac uses tabs for this](https://github.com/canonical/multipass/pull/4211#issuecomment-3056346895). Could you please add screenshots of such tabs on Mac?

ghost:
#4234

alexio-dev:
> Thank you [@alexio-dev](https://github.com/alexio-dev), good point to surface these essential actions more prominently for beginner users. You [mentioned that Mac uses tabs for this](https://github.com/canonical/multipass/pull/4211#issuecomment-3056346895). Could you please add screenshots of such tabs on Mac?

Unfortunately, I can't do it.

PR #4211, opened 2025-07-03
[gui] add `Select All` option to terminal right click menu

levkropp:
image
This pull request introduces a new option to the right click menu in the terminal tab in the GUI, enabling users to select all text within the terminal buffer. The changes include the addition of a new `Intent` class, modifications to the context menu, and implementation of the logic for selecting all terminal text.

### New Feature: Select All Terminal Text

* **Added `SelectAllTerminalTextIntent` class**: Introduced a new `Intent` class to handle the "Select All" functionality. 
* **Updated context menu**: Added a new "Select All" option in the terminal's context menu, which triggers the `SelectAllTerminalTextIntent`. 
* **Implemented selection logic**: Defined the logic for selecting all meaningful text in the terminal buffer, ensuring that only non-empty lines are included in the selection. This involves creating anchors for the start and end of the selection and applying it via the `terminalController`.

### DIFF
--- a/src/client/gui/lib/vm_details/terminal.dart
+++ b/src/client/gui/lib/vm_details/terminal.dart
@@ -176,6 +176,10 @@ class ResetTerminalFontIntent extends Intent {
   const ResetTerminalFontIntent();
 }
 
+class SelectAllTerminalTextIntent extends Intent {
+  const SelectAllTerminalTextIntent();
+}
+
 class _VmTerminalState extends ConsumerState<VmTerminal> {
   static const defaultFontSize = 13.0;
   static const minFontSize = 2.5;
@@ -291,6 +295,16 @@ class _VmTerminalState extends ConsumerState<VmTerminal> {
           );
         },
       ),
+      ContextMenuButtonItem(
+        label: 'Select All',
+        onPressed: () {
+          ContextMenuController.removeAny();
+          Actions.maybeInvoke(
+            context,
+            const SelectAllTerminalTextIntent(),
+          );
+        },
+      ),
     ];
 
     final style = Theme.of(context).textButtonTheme.style?.copyWith(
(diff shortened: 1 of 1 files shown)


ricab:
This is a nice initiative Lev, but it would be the first right-click menu. @vikorama's, I'd love your opinion in that regard.

vikorama:
@ricab seems that this right-click menu is already there, at least on Mac:

image

So, if it's only about adding the 'Select all' option, I can't think of arguments for or against it. 

@levkropp How did you come up with this idea? Would it improve any of your own workflows? As I user I normally just drag a cursor to select what I want.

ricab:
Ohh, my bad, sorry. I thought it was a new thing.

levkropp:
> @levkropp How did you come up with this idea? Would it improve any of your own workflows? As I user I normally just drag a cursor to select what I want.

This idea came from #4192 which suggests a GUI toolbar at the top of the screen as an accessibility enhancement for Multipass. It suggests options for Copy, Paste, and Select All among others

image

Select All is a common option in right click menus for terminals. Above is the right click menu of the built in Terminal in Visual Studio Code as an example.  Use cases would be for people who want to quickly select all information on their terminal screen with just 2 clicks, or those who have accessibility issues regarding the ability to click-and-drag

So regardless of future implementation plans for #4192's toolbar itself, I think Select All being implemented is a valuable addition to our GUI terminal's right click menu.

vikorama:
Thank you for sharing the original request Lev! Something to look into, as I heard about similar struggles from other users.

This PR looks like a good step in the direction!

levkropp:
Thanks @vikorama ! Could I bother you for an approval on this?

ghost:
It would also be interesting to add the `+` and `-` buttons for zoom in one line, such as the GNOME terminal.

ghost:
> @ricab seems that this right-click menu is already there, at least on Mac:
> 
> image
> 
> So, if it's only about adding the 'Select all' option, I can't think of arguments for or against it. 
> 
> @levkropp How did you come up with this idea? Would it improve any of your own workflows? As I user I normally just drag a cursor to select what I want.  
> 
> 

macOS offers tabs everywhere, if it is feasible it would be useful to block those of the system and provide a universal one.

vikorama:
> Thanks @vikorama ! Could I bother you for an approval on this? Here's an Intel Mac package for this PR: https://multipass-ci.s3.amazonaws.com/pr4211/multipass-1.17.0-dev.3884.pr4211%2Bg94a20f20.mac.x86_64-Darwin.pkg

@levkropp it's an honor. The Select all works like a charm, so I can approve this PR. One thing tho is that I realised I could not triple-click within a paragraph of text so that the whole paragraph gets selected. This is the expected behaviour in Mac's and Windows' terminals (Ubuntu too, right?). Is this easy to fix maybe in this PR?

levkropp:
> One thing tho is that I realised I could not triple-click within a paragraph of text so that the whole paragraph gets selected. This is the expected behaviour in Mac's and Windows' terminals (Ubuntu too, right?). Is this easy to fix maybe in this PR?

Hey @vikorama , I wasn't able to reproduce this triple-click behavior on a KDE Konsole on Kubuntu 24.04 or on a Powershell Terminal on Windows 11 Pro. This may be a Mac-specific (Terminal.app) feature?

It also seems that this is not a very straightforward feature to implement in our Flutter GUI as it stands (I attempted a quick solution by wrapping the Terminal in a GestureDetector and calling the function to select all text when a triple click is detected, but this does not work) and may even require changes to our fork of terminal plugin we are using.

src/client/gui/lib/vm_details/terminal.dart
```diff
@@ -405,6 +419,45 @@ class _VmTerminalState extends ConsumerState<VmTerminal> {
...
+          if (buffer.height == 0) return null;
+
+          // Find the last line with actual content
+          int lastContentLine = buffer.height - 1;
+          for (int i = buffer.height - 1; i >= 0; i--) {
+            final line = buffer.lines[i];
+            // Check if the line has any non-empty content
+            bool hasContent = false;
+            for (int j = 0; j < line.length; j++) {
+              if (line.getCodePoint(j) != 0 && line.getCodePoint(j) != 32) {
```
Sploder12, line 437: I wonder if there is a nicer way to do this, but it seems `LineBuffer` is not trivially iterable and `' '.runes.first` isn't much better than `32`

PR #4212, opened 2025-07-04
[client/gui] symlink libdart_ffi to bundle/lib post build

xmkg:
The GUI binary, multipass.gui's RUNPATH is $ORIGIN/lib, which resolves to `$REPO_ROOT/build/bin/bundle/lib/`, whereas the libdart_ffi.so is in `$REPO_ROOT/build/lib` folder. Hence, the multipass.gui fails to launch on local builds if libdart_ffi.so is not somehow placed to the library search path manually (i.e. by copying, or by LD_LIBRARY_PATH).

This patch adds a libdart_ffi.so symlink to the bundle/lib so the GUI can find the dart_ffi library on local build without relying on a manual intervention.

### DIFF
--- a/src/client/gui/CMakeLists.txt
+++ b/src/client/gui/CMakeLists.txt
@@ -42,4 +42,11 @@ if(MULTIPASS_ENABLE_FLUTTER_GUI)
   else() # Windows
     include("CMakeLists.txt.windows")
   endif()
+
+  add_custom_command(TARGET dart_ffi POST_BUILD
+      COMMAND ${CMAKE_COMMAND} -E create_symlink
+      $<TARGET_FILE:dart_ffi>
+      "${CMAKE_BINARY_DIR}/bin/bundle/lib/$<TARGET_FILE_NAME:dart_ffi>"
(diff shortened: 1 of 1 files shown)


xmkg:
@ricab @sharder996 @Sploder12, could I interest any of you in doing a secondary review? :)

ricab (approved):
LGTM, thanks Mustafa!

Just a question out of curiosity: how would the symlink get more complicated than the copy? Either command needs updating on dir-structure changes, no?

xmkg:
> LGTM, thanks Mustafa!
> 
> Just a question out of curiosity: how would the symlink get more complicated than the copy? Either command needs updating on dir-structure changes, no?

Unfortunately for us, the `bundle` directory itself is a symlink too, so to put a proper relative symlink for libdart_ffi in `bundle`, we'd need to resolve all the symlink components in the path, which is a bit more complex (and uglier) than just copying the file.

xmkg:
> Tested on MacOS. LGTM!
> 
> Trade you for a rubber stamp on #4181

Fair enough :) Re-approved :+1:

PR #4215, opened 2025-07-07
[cmake] Fix version derivation on win/mac

ricab:
On Windows and macOS, our version includes a "win" or "mac" suffix in
the <build> component. In actual releases, that is all that the <build>
component consists of. This updates that part of the version derivation
logic such that the semantic version grammar is respected and WiX
accepts our version. We no longer have "+full" in our release tags and
we need our suffixes to be properly separated from the rest of the
version. In particular, they need to come after a single plus ('+') and
must be separated from other parts of the <build> component with dots
('.').

See
https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions.


CMakeLists.txt
```diff
@@ -239,15 +239,23 @@ function(determine_version OUTPUT_VARIABLE)
...
-    string(REGEX MATCH "\\+full$" FULL_SUFFIX_MATCH ${NEW_VERSION})
-    if(NOT FULL_SUFFIX_MATCH)
-      set(NEW_VERSION ${NEW_VERSION}.full)
+  # assert that the output of `git describe` does not include "full" any longer (after merger)
+  string(FIND "${NEW_VERSION}" "full" FULL_POSITION)
+  if(FULL_POSITION GREATER_EQUAL 0)
+      message(FATAL_ERROR "Version string should not contain 'full': ${NEW_VERSION}")
+  endif()
+
+  if(APPLE OR WIN32)
```
levkropp, line 248: Not blocking on this because everything still works, but I'm curious as to why have we switched from if(APPLE OR MSVC) to if(APPLE OR WIN32) here? Is this a more correct way to detect Windows platforms in CMake?

levkropp (approved):
LTGM! Getting rid of the adding and editing "full" string code makes sense, and I don't see any issues or regressions on Windows

sharder996 (approved):
LGTM!

CMakeLists.txt
ricab: Yes, we could conceivably use another compiler on Windows, but we'd still want to enter the branch for the platform. I believe [WIN32](https://cmake.org/cmake/help/latest/variable/WIN32.html) is the correct way to detect it. I suppose "32" is historic.

PR #4221, opened 2025-07-09
[ci] delete old vcpkg cached binaries

sharder996:
Binaries from vcpkg caching will slowly accumulate as versions of dependencies change. This workflow will periodically go through GitHub packages and delete old versions.

For example, [fmt](https://github.com/canonical/multipass/pkgs/nuget/fmt_x64-linux-release) has an old version that isn't used anymore.

---

MULTI-2084

### DIFF
--- a/.github/workflows/prune-packages.yml
+++ b/.github/workflows/prune-packages.yml
@@ -1,0 +1,46 @@
+name: Prune packages
+
+on:
+  workflow_dispatch:
+  schedule:
+  # 04:11 UTC on the first day of each month
+  - cron: '11 4 1 * *'
+
+jobs:
+  list-packages:
+    runs-on: ubuntu-latest
+
+    outputs:
+      matrix: ${{ steps.make-matrix.outputs.matrix }}
+
+    steps:
+      - name: Build package-name matrix
+        id: make-matrix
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-packages.yml
```diff
@@ -0,0 +1,46 @@
...
+            // Build matrix: one entry per package-name
+            const matrix = { include: names.map(n => ({ "package-name": n })) };
+            core.setOutput("matrix", JSON.stringify(matrix));
+
+  prune:
+    needs: list-packages
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix: ${{ fromJSON(needs.list-packages.outputs.matrix) }}
```
xmkg, line 38: Wouldn't making this a "matrix" spawn a job per package? If so, it might be a good idea to throttle it by using `max-parallel`.

sharder996: Sure, that might be a prudent thing to add.

From a technical standpoint it shouldn't be an issue, but if there's a large number of jobs spawned there might be issues with something like DDOS detection on GitHub's side.

xmkg (approved):

Sploder12 (approved):
LGTM!

PR #4226, opened 2025-07-10
[ssl-cert-provider] regenerate gRPC server certs on cert mismatch

### LINKED ISSUE #4218: [multipassd] regenerate gRPC server certificate on root ca mismatch
**Describe the bug**
When ` /usr/local/share/ca-certificates/multipass_root_cert.pem` (the root CA) is updated, the daemon's gRPC server certificate is not regenerated automatically if it already exists. This causes client connection failures due to certificate verification failure:

```bash
\E0000 00:00:1752045756.797423  129964 ssl_transport_security.cc:1659] Handshake failed with fatal error SSL_ERROR_SSL: error:0A000086:SSL routines::certificate verify failed
```

**Expected behavior**
The daemon should verify that the current root CA is indeed the issuer of the server certificate. If not, the daemon should regenerate the server certificate.

**Additional info**
Workaround: Remove multipassd gRPC server certificate to force a regeneration: `sudo rm /root/.local/share/multipassd/certificates/localhost.pem`

xmkg:
The previously generated gRPC server certificate's issuer could be different than the current one in certain circumstances, like a manual installation after a snap installation of Multipass.

The patch fixes that by loading both the root and the subordinate cert for the gRPC server and verifying that the root certificate is the issuer.

Also added some logs to the certificate regeneration logic.

Fixes #4218

MULTI-2100

### DIFF
--- a/tests/test_ssl_cert_provider.cpp
+++ b/tests/test_ssl_cert_provider.cpp
@@ -243,6 +243,98 @@ WizZyQgYv6Z/AosKpE6DcyIYXGuGn2U/Icpxsn/ZycRel2shM4dP5OBg
     EXPECT_THAT(actual_key, StrNe(subordinate_key_contents));
 }
 
+TEST_F(SSLCertProviderFixture, regeneratesCertificatesWhenRootCertIsMissing)
+{
+    const auto [mock_platform, _] = mpt::MockPlatform::inject<NiceMock>();
+    // move the multipass_root_cert.pem into the temporary directory so it will be deleted
+    // automatically later
+    EXPECT_CALL(*mock_platform, get_root_cert_path())
+        .WillRepeatedly(
+            Return(std::filesystem::path{cert_dir.toStdU16String()} / "multipass_root_cert.pem"));
+
+    constexpr auto subordinate_cert_contents =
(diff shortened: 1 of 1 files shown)


src/cert/ssl_cert_provider.cpp
```diff
@@ -351,13 +379,52 @@ mp::SSLCertProvider::KeyCertificatePair make_cert_key_pair(const QDir& cert_dir,
...
+                mpl::warn(kLogCategory,
+                          "Existing root certificate (`{}`) is not the signer of the gRPC "
+                          "server certificate (`{}`)",
+                          root_cert_path,
+                          cert_path.toStdString());
+            }
+            else
+            {
+                mpl::warn(kLogCategory,
+                          "Could not load either of the root (`{}`) or subordinate (`{}`) "
```
ricab, line 421: "Could not load either..." is a little ambiguous — it could mean that neither could be loaded. 

How about "Failed to load either..." instead? I think removing the negative fixes it.

ricab (approved):
Nicely done, thanks! I have a small suggestion below, but considering it optional, so approving on secondary.

Sploder12 (approved):

ricab:
(keeping the respective bugs in the milestone)

PR #4232, opened 2025-07-10
[ci] Restrict GITHUB_TOKEN permissions

ricab:


ricab:
Strange, I was expecting things to break a little. I wonder if this is one of those things that are determined by the base branch... I guess that would make sense security wise.

ricab:
Here's a successful TICS run: https://github.com/canonical/multipass/actions/runs/16224997135/job/45814741325

ricab:
OK, I think this is ready for review. The only workflow I couldn't test was "Snaps", due to limiting triggers. Should be easier to test after #4251. But it's not a big deal if we go a few days without it.

ricab:
Here's a successful "Monitor USNs" run on this branch: https://github.com/canonical/multipass/actions/runs/16416338031

.github/workflows/windows-macos.yml
```diff
@@ -27,6 +30,10 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ !cancelled() }}
 
+    permissions:
```
sharder996, line 33: Did you happen to verify `ccache` uploads?

sharder996: `ccache` uploads normally only happen on `main` or in merge queues IIUC

ricab: Good point, I had not verified them. I did now:

https://github.com/canonical/multipass/actions/runs/16416323921/job/46383138924#step:43:5
https://github.com/canonical/multipass/actions/runs/16416323921/job/46383138751#step:67:7

So it looks like caches are uploaded everywhere, which is a little wasteful because they'd never get picked up.

sharder996: Thanks for pointing that out for me!

.github/workflows/dynamic-ci.yml
```diff
@@ -10,6 +10,11 @@ on:
   merge_group:
     types: [checks_requested]
 
+permissions:
+  contents: read
+  checks: read
+  packages: write
```
sharder996, line 16: Why does this need `packages: write`?

ricab: I am puzzled too, but GH wants that for [some reason](https://github.com/canonical/multipass/actions/runs/16423631542):

> The nested job 'BuildAndTest' is requesting 'packages: write', but is only allowed 'packages: none'.

This occurs despite the job having those permissions assigned in the workflow:

https://github.com/canonical/multipass/pull/4232/files#diff-8607409182b6fe31b140078316b843d9fbfe6eb2f9344d5dcc537ea527c98b66R82

sharder996: Think I found the answer:
> Nested reusable workflows require all workflows in the chain to be accessible to the caller, and permissions can only be maintained or reduced—not elevated—throughout the chain

https://docs.github.com/en/actions/how-tos/sharing-automations/reuse-workflows?utm_source=chatgpt.com#nesting-reusable-workflows

sharder996 (approved):
Looks good now!

Thanks for answering my questions, @ricab

ricab:
Of course, thanks for the reviews!

PR #4236, opened 2025-07-11
[dnsmasq_server] wait for a reasonable amount of time for exit

xmkg:
The existing timeout values (term: 1000ms, kill:100ms) are not good enough to allow dnsmasq to terminate itself. it causes daemon to get stuck on shutdown:

> [info] [dnsmasq] failed to terminate nicely, killing
> [warning] [dnsmasq] failed to kill
> [warning] [Qt] QProcess: Destroyed while process ("dnsmasq") is still running.

Bump terminate timeout to 10s, and kill timeout to 5s.

### DIFF
--- a/src/platform/backends/qemu/linux/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/linux/dnsmasq_server.cpp
@@ -26,6 +26,7 @@
 
 #include <QDir>
 
+#include <chrono>
 #include <fstream>
 
 namespace mp = multipass;
@@ -69,6 +70,9 @@ mp::DNSMasqServer::DNSMasqServer(const Path& data_dir,
 
 mp::DNSMasqServer::~DNSMasqServer()
 {
+    constexpr static auto terminate_timeout = std::chrono::milliseconds(10000).count();
+    constexpr static auto kill_timeout = std::chrono::milliseconds(5000).count();
+
     if (dnsmasq_cmd && dnsmasq_cmd->running())
     {
         QObject::disconnect(finish_connection);
@@ -76,13 +80,15 @@ mp::DNSMasqServer::~DNSMasqServer()
         mpl::log(mpl::Level::debug, "dnsmasq", "terminating");
         dnsmasq_cmd->terminate();
 
-        if (!dnsmasq_cmd->wait_for_finished(1000))
+        if (!dnsmasq_cmd->wait_for_finished(terminate_timeout))
         {
             mpl::log(mpl::Level::info, "dnsmasq", "failed to terminate nicely, killing");
-
             dnsmasq_cmd->kill();
-            if (!dnsmasq_cmd->wait_for_finished(100))
-                mpl::log(mpl::Level::warning, "dnsmasq", "failed to kill");
+
+            if (!dnsmasq_cmd->wait_for_finished(kill_timeout))
+            {
+                mpl::log(mpl::Level::warning, "dnsmasq", "failed to kill (timed out)");
+            }
         }
     }
 }


src/platform/backends/qemu/linux/dnsmasq_server.cpp
```diff
@@ -69,20 +70,25 @@ mp::DNSMasqServer::DNSMasqServer(const Path& data_dir,
 
 mp::DNSMasqServer::~DNSMasqServer()
 {
+    constexpr static auto terminate_timeout = std::chrono::milliseconds(10000).count();
```
ricab: Does it ever terminate successfully? I seem to remember that it doesn't react to SIGTERM, although that may be because I don't wait long enough or I only try when something already went wrong. 

I am concerned that this would always add 10s to quit.

tests/qemu/linux/test_dnsmasq_server.cpp
```diff
@@ -345,4 +354,23 @@ TEST_F(DNSMasqServerMockedProcess, dnsmasqLogsErrorWhenItDies)
     mp::ProcessState state{-1, mp::ProcessState::Error{QProcess::Crashed, msg}};
     emit dnsmasq_proc->finished(state);
 }
+
+TEST_F(DNSMasqServerMockedProcess, dnsmasqCallsKillWhenTerminateTimesOut)
```
ricab, line 349: :+1:

src/platform/backends/qemu/linux/dnsmasq_server.cpp
xmkg: I thought `dnsmasq` would be a good citizen and wrap things up properly when signaled with `TERM`. I tested it under snap confinement, and it does not respond to `TERM`, but `KILL` works. Initially, I thought it might have something to do with the confinement itself, but I'll test it in a non-snap build as well.

xmkg:
> That said, do you know what causes the daemon to hang? Is the QProcess destructor blocking?

I haven't dug deep too much -- the daemon gets blocked on exit when "[dnsmasq] failed to kill" appears in logs. The daemon proceeds with exit if I kill the `dnsmasq` instance, so it appears like it's somehow pending on the `dnsmasq`. I assumed that the daemon is not waiting long enough for the kill (100ms), so I bumped it up to a more reasonable amount. I need to reproduce this bug under a debugger.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
xmkg: > I thought `dnsmasq` would be a good citizen and wrap things up properly when signaled with `TERM`. I tested it under snap confinement, and it does not respond to `TERM`, but `KILL` works. Initially, I thought it might have something to do with the confinement itself, but I'll test it in a non-snap build as well.

Hmm, for snap, `dnsmasq` is not getting SIGTERM at all:

```bash
❯ sudo strace -e trace=signal -p 554642
strace: Process 554642 attached
--- SIGCONT {si_signo=SIGCONT, si_code=SI_USER, si_pid=1, si_uid=0} ---
+++ killed by SIGKILL +++

```

Curious..

xmkg: > > I thought `dnsmasq` would be a good citizen and wrap things up properly when signaled with `TERM`. I tested it under snap confinement, and it does not respond to `TERM`, but `KILL` works. Initially, I thought it might have something to do with the confinement itself, but I'll test it in a non-snap build as well.
> 
> Hmm, for snap, `dnsmasq` is not getting SIGTERM at all:
> 
> ```shell
> ❯ sudo strace -e trace=signal -p 554642
> strace: Process 554642 attached
> --- SIGCONT {si_signo=SIGCONT, si_code=SI_USER, si_pid=1, si_uid=0} ---
> +++ killed by SIGKILL +++
> ```
> 
> Curious..

The signal is simply not getting there:
```
<gdb attached to multipassd, `557100` is the dnsmasq pid>
(gdb) call (int)kill(557100, 15)
$1 = 0 # < succeeds
```
```
sudo strace -e trace=signal -p 557100
strace: Process 557100 attached
# ... cricket sound ...
```

ricab: I guess the AppArmor profile needs tweaking?

xmkg: > I guess the AppArmor profile needs tweaking?

That'd be my guess as well.

xmkg: > I guess the AppArmor profile needs tweaking?

After a couple of hours debugging AppArmor profiles, I've found out that there's nothing wrong with them, and started looking for other culprits, and I think I've found it.

 `dnsmasq` is a child process, meaning it's essentially a fork of `multipassd`, which means at T+0 whatever signals are blocked in `multipassd` would be blocked in the child too. `multipassd` blocks the following signals because of the `quit_watchdog` logic:

```
2: INT
10: USR1
15: TERM
```

So all children would inherit this by default. This also explains why I was able to send a SIGHUP, but not SIGTERM, SIGUSR1, or SIGINT. 

I'll submit a separate PR to fix it.

xmkg: @ricab https://github.com/canonical/multipass/pull/4237 is up.

xmkg:
> Also, I wonder if we could live with something a little shorter (say 5s and 3s?), especially because I have the impression that dnsmasq does not react well to terminate?

This will no longer be the case after https://github.com/canonical/multipass/pull/4237 lands :tada: So with that out of the way, would you think a 10-second timeout on graceful exit is OK?

src/platform/backends/qemu/linux/dnsmasq_server.cpp
ricab: Impressive sleuthing :clap:

jimporter: I debated a bit about this syntax to get the number `10000` but annotate that it's in milliseconds. The best would probably be to make `BasicProcess` take `std::chrono` durations, but that's a more-invasive change that should probably happen separately.

So I think this is ok for now, but I'll take a look at how hard it would be to change the `BasicProcess` interface. If it's not too bad, I'll file a PR so we can simplify this code later.

xmkg: Done. Also implemented `saturate_cast` to clamp the input ms values to the `int` range.

include/multipass/process/process.h
```diff
@@ -94,6 +97,21 @@ class Process : public QObject
...
+
+    bool wait_for_finished(std::chrono::milliseconds ms)
+    {
+        return wait_for_finished(saturate_cast<int>(ms.count()));
+    }
+
+    bool wait_for_ready_read(std::chrono::milliseconds ms)
+    {
+        return wait_for_ready_read(saturate_cast<int>(ms.count()));
+    }
```
ricab: If it is not too much trouble, could we keep implementations separate from the interface listing? This file was still following that. Here's a diff for your convenience:

```
diff --git a/include/multipass/process/process.h b/include/multipass/process/process.h
index 455bb57..2d1e033 100644
--- a/include/multipass/process/process.h
+++ b/include/multipass/process/process.h
@@ -96,21 +96,9 @@ public:
     virtual bool wait_for_started(int msecs = 30000) = 0;
     virtual bool wait_for_finished(int msecs = 30000) = 0;
     virtual bool wait_for_ready_read(int msecs = 30000) = 0;
-
-    bool wait_for_started(std::chrono::milliseconds ms)
-    {
-        return wait_for_started(saturate_cast<int>(ms.count()));
-    }
-
-    bool wait_for_finished(std::chrono::milliseconds ms)
-    {
-        return wait_for_finished(saturate_cast<int>(ms.count()));
-    }
-
-    bool wait_for_ready_read(std::chrono::milliseconds ms)
-    {
-        return wait_for_ready_read(saturate_cast<int>(ms.count()));
-    }
+    bool wait_for_started(std::chrono::milliseconds ms);
+    bool wait_for_finished(std::chrono::milliseconds ms);
+    bool wait_for_ready_read(std::chrono::milliseconds ms);
 
     virtual bool running() const = 0;
     virtual ProcessState process_state() const = 0;
@@ -141,4 +129,19 @@ protected:
 };
 } // namespace multipass
 
+inline bool multipass::Process::wait_for_started(std::chrono::milliseconds ms)
+{
+    return wait_for_started(saturate_cast<int>(ms.count()));
+}
+
+inline bool multipass::Process::wait_for_finished(std::chrono::milliseconds ms)
+{
+    return wait_for_finished(saturate_cast<int>(ms.count()));
+}
+
+inline bool multipass::Process::wait_for_ready_read(std::chrono::milliseconds ms)
+{
+    return wait_for_ready_read(saturate_cast<int>(ms.count()));
+}
+
 Q_DECLARE_METATYPE(multipass::ProcessState)
```

ricab: One other thing: why not keep the default? These aren't virtual, so that wouldn't be an issue.

include/multipass/utils/saturate_cast.h
```diff
@@ -0,0 +1,50 @@
...
+
+    if (std::cmp_less(from_value, min_to))
+        return min_to;
+
+    if (std::cmp_greater(from_value, max_to))
+        return max_to;
+
+    return static_cast<To>(from_value);
+}
+} // namespace multipass
```
ricab, line 52: Did you take this implementation from somewhere? Could we cover this in unit tests?

include/multipass/utils/saturate_cast.h
```diff
@@ -0,0 +1,50 @@
...
+namespace multipass
+{
+
+/**
+ * Converts the value from_value to a value of type To, clamping x between the minimum and maximum
+ * values of type To.
+ *
+ * @tparam To Type to be converted
+ * @tparam From From type (auto-deduced)
+ */
```
ricab, line 35: Can we have a TODO to remove in C++26, please?

include/multipass/utils/saturate_cast.h
xmkg: Initially, I re-implemented libstdc++'s approach, but noticed that it would be a much simpler implementation with std::cmp_less and std::cmp_greater. I don't know if there are any stdlib impls using this approach, though.

xmkg: It looks like libc++ is using this exact approach, too:

https://github.com/llvm/llvm-project/blob/e9e9ba4a694b778038bbda0120fca075e9b6bd1e/libcxx/include/__numeric/saturation_arithmetic.h#L97-L109

I guess we won't need unit tests in that case, no?

include/multipass/utils/saturate_cast.h
xmkg: done

include/multipass/process/process.h
xmkg: Separated definitions from declarations. I didn't quite catch the part about the "default" though. Could you elaborate, please?

xmkg: Ahh nvm, you're talking about the default value for the arg. I'll add it.

xmkg: @ricab adding a default for the non-virtual overload makes the non-arg calls ambiguous -- did you mean something else? I can remove the default arg from the virtual overloads if that sounds good to you.

EDIT: Made the change, lemme know how you feel about it.

ricab: Ahh of course, I missed that. Yes, removing from the virtual overloads makes sense to me.

Thanks also for the decl/impl split!

include/multipass/utils/saturate_cast.h
ricab: Hmm OK, I guess I'll overlook it in this case. It's not the same implementation (that would need attribution), but it's pretty close. And it is something to remove in the future, so fine.

include/multipass/process/process.h
```diff
@@ -120,7 +127,23 @@ class Process : public QObject
 
 protected:
     virtual void setup_child_process() = 0;
+    constexpr static inline auto kDefaultWaitAmount = std::chrono::milliseconds{30000};
```
ricab: Hmm, got your cases mixed up from other projects? 😉 We use snake_case for variables:

https://github.com/canonical/multipass/blob/472d158d1e189665d21c153a56ebf94c06b6ad11/CONTRIBUTING.md?plain=1#L262

Also, I understand the motivation for the "k" prefix: it makes it clear that it is a constant. But it has its downsides too: it is a duplicate source of truth, so if we ever wanted to make it a variable, we'd need to change the name everywhere.

It's similar as with `m` prefixes for member variables. Personally, I find it more trouble than worth, especially when IDE's can quickly show types and definitions. I'd rather have fewer knots to untie. That said, I understand this is a matter of preference. We could debate this in the team if you want (though I seem to remember we did so already?). For now, we'd better be consistent throughout the codebase, don't you think?

include/multipass/utils/saturate_cast.h
```diff
@@ -0,0 +1,52 @@
...
+ * values of type To.
+ *
+ * @tparam To Type to be converted
+ * @tparam From From type (auto-deduced)
+ *
+ * TODO: Remove when C++26 arrives.
+ */
+template <typename To, typename From>
+requires(std::is_integral<To>::value&& std::is_integral<From>::value) constexpr To
+    saturate_cast(From from_value) noexcept
```
ricab: Suggestion for readability:

```cpp
// clang-format off
template <typename To, typename From>
requires(std::is_integral_v<To>&& std::is_integral_v<From>)
constexpr auto saturate_cast(From from_value) noexcept -> To
// clang-format on
```

Adding a space before the "and", switching to `_v` shortcuts, switching to trailing return, and splitting the lines where it makes sense (had to disable clang-format). It's just a suggestion though, feel free to pick only some of those, as you prefer.

include/multipass/process/process.h
xmkg: > Hmm, got your cases mixed up from other projects?

No, it is intentional, and we do have existing instances of this in the codebase:

image

> Personally, I find it more trouble than worth, especially when IDEs can quickly show types and definitions.

I'd argue IDEs are also good at renaming variables, too, especially with a good language server like clangd.

> though I seem to remember we did so already?

I don't recall debating this before, and there are more than one PRs that already landed that you've reviewed featuring this naming convention, so I didn't know we had a preference on naming constants. I don't have a strong preference on this, so I'm fine either way. I guess it's worth adding a linter check for it? (e.g.., .clang-tidy)

include/multipass/utils/saturate_cast.h
xmkg: Sounds good. I would say if we're concerned about this, we should look into configuring .clang-format to our liking instead of disabling the formatter. I'll look into how we can configure clang-format to produce a similar output to your example.

xmkg: @ricab it seems like RequiresClausePosition is the config flag we're looking for. It's defaulted to `OwnLine` when Standard is set to c++20, which produces sth like this:

```c++
template <typename To, typename From>
  requires(std::is_integral_v<To> && std::is_integral_v<From>)
constexpr To saturate_cast(From from_value) noexcept {}
```

The `SingleLine`, though, produces output similar to your example:

```c++
template <typename To, typename From>
requires(std::is_integral_v<To> && std::is_integral_v<From>)
constexpr To saturate_cast(From from_value) noexcept {}
```

I've set the C++ standard to 20, and set the said config flag to SingleLine. Let me know what you think.

include/multipass/process/process.h
xmkg: So, a .clang-tidy rule for the example above would look like below:

```.clang-tidy
Checks: 'readability-identifier-naming'
CheckOptions:
  - { key: readability-identifier-naming.MacroDefinitionCase, value: UPPER_CASE }
  - { key: readability-identifier-naming.ConstexprVariableCase, value: lower_case }
```

The clang-tidy would produce the following warning when being executed on the process.h:

```bash
❯ clang-tidy-18 include/multipass/process/process.h -p build/ --config-file=/workspace/multipass/.clang-tidy
3506 warnings generated.
/workspace/multipass/include/multipass/process/process.h:130:34: warning: invalid case style for constexpr variable 'kDefaultWaitAmount' [readability-identifier-naming]
  100 |     constexpr static inline auto kDefaultWaitAmount = std::chrono::milliseconds{30000};
      |                                  ^~~~~~~~~~~~~~~~~~
      |                                  k_default_wait_amount
Suppressed 3505 warnings (3505 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

What do you think? Should we introduce these rules and start enforcing them in the linting step?

ricab: OK, I thought I remembered seeing a few cases, but I wasn't sure if it had been merged. I'm surprised it's those many cases. Did you author all of them?

> I'd argue IDEs are also good at renaming variables, too, especially with a good language server like clangd.

That's a good point. It still doesn't beat single source of truth in my view.

> I don't recall debating this before

I am not sure either, I just had some reminiscence, but maybe not. Anyway, I guess the main idea here is consistency. Another couple of rules that apply here: https://github.com/canonical/multipass/blob/400453090349ef316e083af3468d01959cbe9ec4/CONTRIBUTING.md?plain=1#L239-L241

> What do you think? Should we introduce these rules and start enforcing them in the linting step?

Sure, that would be better. Sounds good to me, thanks!

ricab: Feel free to separate into another PR though, to avoid dragging this on any longer.

include/multipass/utils/saturate_cast.h
ricab: Good find. I will let you choose. Now that I look at it, I think I actually like the first form better. It's more habitual, isn't it?

What I didn't like was that it split after "constexpr To" when I tried, even though it had the perfect seam right before the `constexpr`. But anyway, let's just choose something and stick with it. Feel free to bring it to the team of course.

include/multipass/process/process.h
xmkg: I think it's better to introduce this in a separate PR. I'll rename the constant and let the renaming of the others be in the upcoming PR.

xmkg: > Did you author all of them?

Yes, I think so.

xmkg: Done

include/multipass/utils/saturate_cast.h
xmkg: I think the OwnLine looks more natural, too. Reverted to the default setting of "RequiresClausePosition" (OwnLine).

.clang-format
```diff
@@ -3,7 +3,7 @@
 BasedOnStyle: LLVM
 ---
 Language: Cpp
-Standard: c++17
+Standard: c++20
```
jimporter: Good catch. I was going to file a PR for this, but you beat me to it.

PR #4237, opened 2025-07-11
[linux/process_factory] unblock all signals after a child fork()

xmkg:
multipassd blocks SIGINT, SIGTERM and SIGUSR1 as a part of the watchdog logic. As spawning a child process involves fork()ing multipassd, the blocked signal set would be inherited by the child too. We rely on sending signals to child process for various purposes, graceful exit is being the most prominent one.

This patch clears the blocked signals after a fork so the child would get an unblocked signal set by default.

### DIFF
--- a/src/platform/backends/shared/linux/process_factory.cpp
+++ b/src/platform/backends/shared/linux/process_factory.cpp
@@ -25,6 +25,8 @@
 #include <multipass/process/simple_process_spec.h>
 #include <multipass/snap_utils.h>
 
+#include <signal.h>
+
 namespace mp = multipass;
 namespace mpl = multipass::logging;
 
@@ -52,6 +54,15 @@ public:
     void setup_child_process() final
     {
         mp::BasicProcess::setup_child_process();
+
+        // This function runs after fork, but before exec, which is a perfect
+        // place to reset the signal masks.
+        // Unblock all signals for the child process
+        {
+            sigset_t set;
+            sigemptyset(&set);
+            sigprocmask(SIG_SETMASK, &set, nullptr);
+        }
 
         apparmor.next_exec_under_policy(process_spec->apparmor_profile_name().toLatin1());
     }


xmkg:
Testing this one is easy. Before patch (daemon log):

```
[debug] [dnsmasq] terminating
[info] [dnsmasq] failed to terminate nicely, killing
```

After patch:
```
[debug] [dnsmasq] terminating
```

The daemon shouldn't have to resort to killing dnsmasq since `dnsmasq` now can receive the SIGTERM :tada: (also valid for the other child processes)

Also, for further verification, run this to verify which signals are blocked by multipassd. It should be SIGINT, USR1, and TERM:
```
sigblk=$(cat /proc/$(pidof multipassd)/status | awk '/SigBlk/ { print $2 }')
for i in {1..64}; do
  if (( (0x$sigblk) & (1 << ($i - 1)) )); then
    echo "$i: $(kill -l $i)"
  fi
done
```

For `dnsmasq`, the following command block should return nothing:
```
sigblk=$(cat /proc/$(pgrep -f "dnsmasq.*multipassd")/status | awk '/SigBlk/ { print $2 }')
for i in {1..64}; do
  if (( (0x$sigblk) & (1 << ($i - 1)) )); then
    echo "$i: $(kill -l $i)"
  fi
done
```

xmkg:
@jimporter I would be happy if you could review this one :)

src/platform/backends/shared/linux/apparmor.cpp
```diff
@@ -143,3 +144,50 @@ void mp::AppArmor::next_exec_under_policy(const QByteArray& aa_policy_name) cons
...
+        int saved = errno;
+        ::close(fd);
+        errno = saved;
+
+        if (n == (ssize_t)len)
+            return 0;     // success
+        last_err = errno; // remember why it failed
+    }
+
+    errno = last_err;
```
jimporter: Is `last_err` necessary? It looks like the last statement executed by the for loop is always `last_err = errno;` (lines 177 and 188), so when we reach line 191, `last_err` and `errno` should always be equal. I think we could just remove all references to `last_err` and the behavior would still be the same.

Alternately, maybe we want to handle errors slightly differently here? If `::open` fails for just one of the paths, does that mean this function failed? Or would we consider it a success?

(None of this *really* matters that much since I don't think we consult the return value or `errno` after calling this function, but we may as well think about what the expected status is while the code is fresh in our minds.)

src/platform/backends/shared/linux/process_factory.cpp
```diff
@@ -51,9 +55,36 @@ class AppArmoredProcess : public mp::BasicProcess
 
     void setup_child_process() final
     {
+        // This function runs after fork, but before exec, which is a perfect
+        // place to reset the signal masks.
+        //
+        // BEWARE: This function is called right in between fork and exec, which is a *very* special
+        // state. This is a critical section where only a handful of functions can be called:
```
jimporter: Just to clarify this section, I believe the real problem we have is that we're between `fork(2)` and `exec(3)` in a *multithreaded* application. It's not so bad to be between those two calls in a single-threaded application. (In theory, we could call [`pthread_atfork(3)`](https://www.man7.org/linux/man-pages/man3/pthread_atfork.3.html) to resolve this, but I'm guessing from their docs that Qt doesn't do that, and so we'd have a hard time cleaning up mutexes and such on Qt's behalf.)

How about this wording to make sure we mention the multithreaded part of the problem?

```suggestion
        // BEWARE: This function is called right in between fork and exec. Since this application is
        // multithreaded, we need to be *very* careful; fork replicates the state of mutexes and
        // other pthreads constructs, so we can only call async-signal-safe functions:
```

src/platform/backends/shared/linux/apparmor.cpp
xmkg: > Is last_err necessary? It looks like the last statement executed by the for loop is always last_err = errno; (lines 177 and 188), so when we reach line 191, last_err and errno should always be equal. I think we could just remove all references to last_err and the behavior would still be the same.

Well spotted, I think `last_err` is redundant at this point.

> Alternately, maybe we want to handle errors slightly differently here? If ::open fails for just one of the paths, does that mean this function failed? Or would we consider it a success?

It's a failure if all of the `::open` + `::write` attempts fail.

xmkg:
@tobe2098, could I interest you in reviewing and testing this?

@jimporter, it's ready for another round of review.

tobe2098 (approved):
LGTM. I experienced errors where after a daemon crash killing dnamasq had to be stopped via sudo kill -9 and starting the daemon was not possible if not done, so I hope this maybe solves that. So far tests seem satisfactory.

xmkg:
> PD: Funnily enough, the fix works well from a regular terminal, but the dnsmasq termination still fails if the daemon was running from the terminal inside vscode, and the logs show:

It looks like the AppArmor profile we have is not happy with vscode sending a signal. Is this a snap installation of vscode? If so, the AppArmor profile would need additional permissions to allow signaling.

tobe2098:
No, my VSCode is snap-less. I do not know to what extent this can be covered, from my testing it seems like it is more related to how the process started. The VSCode terminal is a PTY, not a TTY, and if the process starts in that terminal the kill signal never reaches dnsmasq, and curiously enough that terminal cannot sudo kill -9 dnsmasq either. The processes share the pgid, so it is not that, but the key is in the start of the process, not sending the signal. The VSCode terminal can properly kill the parent and child processes if they started in the default terminal.

src/platform/backends/shared/linux/apparmor.cpp
```diff
@@ -143,3 +144,46 @@ void mp::AppArmor::next_exec_under_policy(const QByteArray& aa_policy_name) cons
...
+        ssize_t n = ::write(fd, exec, (ssize_t)len); // must be one atomic write
+        int saved = errno;
+        ::close(fd);
+        errno = saved;
+
+        if (n == (ssize_t)len)
+            return 0; // success
+    }
+
+    return errno;
```
jimporter, line 188: Above, we `return -1` on error, which is the usual convention, so we should probably do it here too:

```suggestion
    return -1;
```

Issue #4228, opened 2025-07-10
Minimum requirements when configuring new instance


vikorama:
**What are you trying to do?**
Launch a VM with disk space less than the image needs

**What's your proposed solution?**
There was a [ticket](https://warthogs.atlassian.net/browse/MULTI-974) a couple of cycles ago to adjust the minimum and maximum ranges of the sliders according to what an image needs and a host has.

So, one of the acceptance criteria would be: 
**Given** I configure a new instance for Ubuntu Server
**Then** I should see the disk slider set to the minimum of 5GB

**Additional context**

<img width="918" height="683" alt="Image" src="https://github.com/user-attachments/assets/25de28e9-5eff-45be-8685-dc681352dffe" />

PR #4239, opened 2025-07-11
[gui] Set minimum RAM and disk size based on image type in launch form

levkropp:
implements #4228 

This pull request introduces a feature to dynamically determine minimum RAM and disk space requirements based on the image type in the `LaunchForm` widget. These changes enhance the configurability of sliders for memory and disk space.

Key changes:

### Dynamic minimums for sliders:
* [`src/client/gui/lib/catalogue/launch_form.dart`](diffhunk://#diff-1859f216fc901caa87909945aef29585a3d98e3b5e265a7e804870ae8a3a652fR98-R111): Added logic to calculate `minRam` and `minDisk` based on whether the image type is identified as "core" or not. Core images have lower minimum requirements (512 MiB RAM, 1 GiB disk), while non-core images have higher minimums (1024 MiB RAM, 5 GiB disk). These values are set as the minimums for the `RamSlider` and `DiskSlider`.

### DIFF
--- a/src/client/gui/lib/catalogue/launch_form.dart
+++ b/src/client/gui/lib/catalogue/launch_form.dart
@@ -95,13 +95,20 @@ class _LaunchFormState extends ConsumerState<LaunchForm> {
       onSaved: (value) => launchRequest.numCores = value!,
     );
 
+    // Determine minimums based on image type
+    final isCore = imageInfo.aliasesInfo.any((a) => a.alias.contains('core'));
+    final minRam = isCore ? 512.mebi : 1024.mebi;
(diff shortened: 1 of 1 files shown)


ricab:
Ideally the daemon would provide this information for the images it reports, but I guess we can have this as a stop gap for now. However, could we possibly use a stricter regex, rather than just a substring, to avoid problems like Trevor reports. Perhaps something like `.*\bcore\d{0,2}\b.*`? (to be tweaked)

Sploder12 (approved):
this more restrictive regex should prevent most future issues, LGTM!

sharder996 (approved):

PR #4240, opened 2025-07-11
[ci/lint] add `flutter pub get` before `dart format`

xmkg:
Fetch Flutter deps before running the formatter, as it can influence the outcome.

MULTI-2102


sharder996 (approved):
Yup, got no more additional files linted locally.

PR #4249, opened 2025-07-14
[ci] Install nuget on macOS 14 with brew too

ricab:
Fixes vcpkg caching on macOS 14.

Without this, the incantation to overwrite nuget fails: https://github.com/canonical/multipass/actions/runs/16224912911/job/45814502234#step:16:28

We could instead tweak the recipe to find nuget parts from the baked install, but this is a quicker fix that costs 12s per CI run and keeps our CI setup a little simpler (no need to branch based on macOS version).

### DIFF
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -170,7 +170,7 @@ jobs:
         fi
 
     - name: Install nuget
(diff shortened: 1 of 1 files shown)


sharder996 (approved):
Looks good in CI. Lets get this in and keep monitoring stability.

PR #4251, opened 2025-07-15
Tweak USNs workflow

ricab:
Rename "Snaps" workflow and allow triggering it manually.

### DIFF
--- a/.github/workflows/monitor-usns.yml
+++ b/.github/workflows/monitor-usns.yml
@@ -1,4 +1,4 @@
-name: Monitor USNs
+name: Monitor USNs  # Ubuntu Security Notices
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/monitor-usns.yml
```diff
@@ -1,8 +1,9 @@
-name: Snaps
+name: Monitor USNs
```
xmkg: Forgive my ignorance, but what does `USN` mean here?

ricab: Sorry, [Ubuntu Security Notices](https://ubuntu.com/security/notices). I'll add it in a comment.

ricab:
Thanks both

PR #4257, opened 2025-07-18
[ci] use dedicated self-hosted tiobe runner

sharder996:
Using single label to prevent confusion with multiple labels.

https://canonical-self-hosted-github-runner-docs.readthedocs-hosted.com/en/latest/usage/available_runners/

MULTI-2105

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


xmkg (dismissed):

ricab (commented):
Thanks @sharder996. Could we have it run at both large and 2xlarge to begin with, to compare the run times and decide if the benefit justifies the larger machine here as well?

PR #4261, opened 2025-07-21
[CI] CI tweaking

sharder996:
- Edited the schema for naming ccache dumps; `matrix.runs-on` is already a unique identifier so cleaned up some obtuse naming
- Stopped throttling builds on ARM Mac
- Removed cache size limit
- Removed some outdated directories in `.gitignore` that are related to building with Travis CI
- Removed wait for lint check; the output doesn't make a difference when gating a PR, so why gate compilation on it
- Removed the `cla-check` workflow. A `cla-check` is required by ruleset on the Canonical org level, so this workflow was doing double duty

### DIFF
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -81,7 +81,7 @@ jobs:
     #
     # https://github.com/protocolbuffers/protobuf/issues/21447
     - uses: maxim-lobanov/setup-xcode@v1
-      if: ${{ contains(matrix.runs-on, 'macos-15') }}
+      if: ${{ contains(matrix.runs-on, 'macos-latest') }}
       with:
         xcode-version: '16.2'
 


.github/workflows/windows-macos.yml
```diff
@@ -60,13 +41,12 @@ jobs:
 
         MATRIX='{"include": ['
         MATRIX+='{"runs-on": "windows-latest", "mono": ""},'
-        MATRIX+='{"runs-on": "macos-13", "arch": "x86_64", "target": "13", "mono": "mono"},'
-        MATRIX+='{"runs-on": "macos-14", "arch": "arm64", "target": "14", "mono": ""},'
```
xmkg, line 64: Why are we removing `macos-14` and `macos-15`?

.github/workflows/windows-macos.yml
```diff
@@ -138,6 +113,7 @@ jobs:
       if: ${{ matrix.arch }}
       run: echo "ARCH_WRAPPER=arch -arch ${{ matrix.arch }}" >> ${GITHUB_ENV}
 
+      # TODO: find a friendlier approach for installing pip packages on GitHub runners
```
xmkg, line 112: This would require creating a virtual environment (venv), which is a bit overkill for a throwaway environment, and might complicate our CI code a bit more. I agree that it would be a nice addition, though.

.github/workflows/windows-macos.yml
```diff
@@ -75,37 +55,32 @@ jobs:
...
       matrix: ${{ fromJSON(needs.GetMatrix.outputs.matrix) }}
       fail-fast: ${{ github.event_name == 'merge_group' }}
 
     outputs:
       build-label: ${{ steps.build-params.outputs.label }}
-      # Use artifacts from macos-13 and macos-14 in official package
+      # Use artifacts from x86 and ARM Macs in official package
       mac-x86-package-artifact-name: ${{ steps.publish-data.outputs.macos-13-x86_64-package-artifact-name }}
-      mac-arm-package-artifact-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-artifact-name }}
+      mac-arm-package-artifact-name: ${{ steps.publish-data.outputs.macos-latest-arm64-package-artifact-name }}
```
xmkg: `latest` is a moving target -- what's the reason to unpin from a specific version?

.github/workflows/windows-macos.yml
sharder996: My thought process was that the version of the build platform isn't important, it's the build target that matters. So, in the same way we build on `windows-latest` or `ubuntu-latest` we can use `macos-latest` and target macOS-14 without any trouble. This also reduces the number of minutes we are using for each workflow dispatch.

.github/workflows/windows-macos.yml
sharder996: Just to stay up-to-date with macOS versions. Same as what we do with `windows-latest` and `ubuntu-latest`.

xmkg: I'm okay with it as long as the build artifacts produced by the `macos-latest` are compatible with the prior versions.

sharder996: I don't see any reason they wouldn't be. That's the purpose of `set(CMAKE_OSX_DEPLOYMENT_TARGET "14.0")`

ricab: Was that tested? 

I've received (from Chris) the information that, to be compatible with a macOS version, our packages had to be built on that version or our build would have to change to become backward compatible. It's not something I've personally verified, but we've been operating under that assumption for a long time. That is the reason behind the minimum version changes, custom runners, etc. 

I'd be wary to change that now without assurance that things run fine in the oldest supported version (that will be macOS-14 going forward).

.github/workflows/windows-macos.yml
```diff
@@ -60,13 +41,12 @@ jobs:
 
         MATRIX='{"include": ['
         MATRIX+='{"runs-on": "windows-latest", "mono": ""},'
-        MATRIX+='{"runs-on": "macos-13", "arch": "x86_64", "target": "13", "mono": "mono"},'
-        MATRIX+='{"runs-on": "macos-14", "arch": "arm64", "target": "14", "mono": ""},'
-        MATRIX+='{"runs-on": "macos-15", "arch": "arm64", "target": "15", "mono": ""},'
+        MATRIX+='{"runs-on": "macos-13", "arch": "x86_64", "mono": "mono"},'
+        MATRIX+='{"runs-on": "macos-latest", "arch": "arm64", "mono": ""},'
```
ricab: Until we verify that backward compatibility I mentioned in the other comment, there are 3 builds that make sense to me:

- 2 builds (1 per architecture) on the oldest macOS version that GH provides on both architectures (macOS 14 ATTOW)
- 1 build on macos-latest, to remain compatible with the latest stuff (only one arch supported now)

.github/workflows/windows-macos.yml
```diff
@@ -60,13 +41,12 @@ jobs:
...
 
         if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" && "${{ github.event.inputs.host-my-apple }}" == "true" ]]; then
-          MATRIX+='{"runs-on": "multipass-macos-12-x64", "arch": "x86_64", "target": "12"},'
-          MATRIX+='{"runs-on": "multipass-macos-12-arm", "arch": "arm64", "target": "12"},'
+          MATRIX+='{"runs-on": "multipass-macos-12-x64", "arch": "x86_64"},'
+          MATRIX+='{"runs-on": "multipass-macos-12-arm", "arch": "arm64"},'
         fi
```
ricab: I think we can drop this now.

.github/workflows/windows-macos.yml
```diff
@@ -75,37 +55,32 @@ jobs:
...
       mac-x86-package-artifact-name: ${{ steps.publish-data.outputs.macos-13-x86_64-package-artifact-name }}
-      mac-arm-package-artifact-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-artifact-name }}
+      mac-arm-package-artifact-name: ${{ steps.publish-data.outputs.macos-latest-arm64-package-artifact-name }}
       mac-x86-package-file-name: ${{ steps.publish-data.outputs.macos-13-x86_64-package-file-name }}
-      mac-arm-package-file-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-file-name }}
+      mac-arm-package-file-name: ${{ steps.publish-data.outputs.macos-latest-arm64-package-file-name }}
 
     runs-on: ${{ matrix.runs-on }}
-    continue-on-error: ${{ contains(matrix.runs-on, 'macos-13') || contains(matrix.runs-on, 'macos-14') || contains(matrix.runs-on, 'macos-15') }}
+    continue-on-error: ${{ contains(matrix.runs-on, 'macos') }}
```
ricab: Shouldn't we just drop this. IIUC this was here while experimenting with GH runners, to avoid blocking on those, as long as custom ones succeeded.

.github/workflows/windows-macos.yml
```diff
@@ -197,7 +173,7 @@ jobs:
         fi
 
     - name: Install nuget
-      if: ${{ contains(matrix.runs-on, 'macos-15') || contains(matrix.runs-on, 'macos-14') }}
+      if: ${{ contains(matrix.runs-on, 'macos-latest') }}
```
ricab: Can't we just remove the condition?

.github/workflows/windows-macos.yml
```diff
@@ -247,10 +222,9 @@ jobs:
       if: ${{ runner.os == 'macOS' }}
       uses: actions/cache@v4
       with:
-        key: "ccache-${{ runner.os }}-${{ env.MACOSX_DEPLOYMENT_TARGET }}-${{ matrix.arch }}-\
-          ${{ steps.setup-ccache.outputs.cache-key }}"
+        key: "ccache-${{ matrix.runs-on }}-${{ steps.setup-ccache.outputs.cache-key }}"
```
ricab, line 224: Shouldn't the arch remain part of it? Otherwise we'd be relying on different macOS versions for different architectures.

.github/workflows/windows-macos.yml
```diff
@@ -295,10 +269,6 @@ jobs:
         ${{ runner.os == 'Windows' && '-DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe' || '' }}
         ${{ github.workspace }}
 
-    - name: Throttle parallel build processes on arm64
-      if: ${{ runner.os == 'macOS' && matrix.arch == 'arm64' }}
-      run: echo CMAKE_BUILD_PARALLEL_LEVEL=2 >> ${GITHUB_ENV}
```
ricab, line 312: :+1:

.github/workflows/windows-macos.yml
sharder996: No, because the Intel Mac behaves differently.

.github/workflows/windows-macos.yml
sharder996: `matrix.runs-on` should already be a unique identifier since we are multiplexing from the [workflow label](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories), no? GitHub will already provide the architecture in the label if it is relevant (i.e. `ubuntu-24.04` vs `ubuntu-24.04-arm`).

Please correct me if I'm misunderstanding.

.github/workflows/windows-macos.yml
sharder996: Good catch! That was introduced in https://github.com/canonical/multipass/commit/06f159b7946a4d3918a04c0edf64af97eb2b70bc. I will remove.

.github/workflows/windows-macos.yml
sharder996: Ok, you might be right here. I guess `DEPLOYMENT_TARGET` doesn't do what it says in that it specifies a minimum, but not a build target. Did some testing in a MacOS 14 VM and I immediately get a `built for macOS 15.0 which is newer than running OS` error.

.github/workflows/windows-macos.yml
sharder996: Reverted the macOS 14 runner. Still switching to `macOS-latest` from `macOS-15` so that we can catch new build errors early as Apple updates macOS.

.github/workflows/windows-macos.yml
ricab: OK, so the difference is in the architecture, not the macOS version, right? If so, better check the architecture here instead, no? It would make it clearer and future proof. 

I mean, I hope they wouldn't, but suppose GH moved the intel runners to 14, relabeled them as `macOS-14`, and removed or relabeled arm 14 (e.g. macos-14-arm). And suppose we reacted by relabeling things in the matrix, but forgot to change this. It's not a big deal, but it feels safer.

.github/workflows/windows-macos.yml
ricab: Oh OK, I understand better now. We can assume that there is never going to be a label that covers both architectures. Sounds good.

.github/workflows/windows-macos.yml
```diff
@@ -59,33 +34,22 @@ jobs:
...
-        MATRIX+='{"runs-on": "macos-14", "arch": "arm64", "target": "14", "mono": ""},'
-        MATRIX+='{"runs-on": "macos-15", "arch": "arm64", "target": "15", "mono": ""},'
-
-        if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" && "${{ github.event.inputs.host-my-apple }}" == "true" ]]; then
-          MATRIX+='{"runs-on": "multipass-macos-12-x64", "arch": "x86_64", "target": "12"},'
-          MATRIX+='{"runs-on": "multipass-macos-12-arm", "arch": "arm64", "target": "12"},'
-        fi
-
-        MATRIX=${MATRIX%,}
+        MATRIX+='{"runs-on": "windows-latest"},'
```
ricab, line 37: We should probably introduce an older Windows target, especially when we move to the API-based Hyper-V. But that's for another chapter.

.github/workflows/windows-macos.yml
ricab: Then again, this might cause problems if there is a relabeling of the sort I mentioned in the other comment. Wouldn't it be safer to just keep the architecture until they throw intel away for good?

.github/workflows/windows-macos.yml
```diff
@@ -374,12 +331,19 @@ jobs:
 
     needs: [BuildAndTest]
 
-    runs-on: macos-13
+    runs-on: macos-latest
```
ricab: Why change this? The oldest supported macOS on Intel is a tried and true option. Did you confirm version compatibility when doing this step in latest? Are we sure our combined script does the right thing when running on ARM?

This will no longer be necessary when Intel goes out of the picture, so I don't see a reason to move to latest here.

.github/workflows/windows-macos.yml
sharder996: The relabelling would simply cause a mismatch in caches dumps. This would persist until a new cache had been uploaded.

What would you prefer?

ricab: OK, let's leave it like this then.

sharder996:
@ricab Yes, I have tested the package on ARM.

sharder996:
@ricab @xmkg 

The Canonical org ruleset for the CLA check seems to have disappeared so I've opted to not remove it here just quite yet.

ricab:
Oh OK. Makes sense.

ricab (approved):
LGTM, thanks again!

ricab (approved):

ricab:
@xmkg, you still OK with this? If so, feel free to merge.

PR #4266, opened 2025-07-21
Return failure on exec command failure

sharder996:
Work done by @artemvlas in #4109 

---

Addresses #3054 
Closes #4109


src/client/cli/cmd/shell.cpp
```diff
@@ -76,7 +76,10 @@ mp::ReturnCode cmd::Shell::run(mp::ArgParser* parser)
         {
             auto console_creator = [this](auto channel) { return term->make_console(channel); };
             mp::SSHClient ssh_client{host, port, username, priv_key_blob, console_creator};
-            ssh_client.connect();
+            const int exit_code = ssh_client.connect();
```
xmkg, line 79: It would be nice if we did have UT coverage for this part.

ricab: As discussed elsewhere, let's open an exception for this one.

ricab (approved):

PR #4269, opened 2025-07-22
[ci] Fix newlines in runner tags

ricab:


ricab:
@sharder996 so I think we can merge this, no?

PR #4274, opened 2025-07-24
[tics] Use large tiobe runner

sharder996:
Between 2xlarge and large tiobe runners, processing times are negligible


ricab (commented):
Makes sense, but need to remove the `smaller` input parameter too.

PR #4279, opened 2025-07-24
[ci] Remove redundant shell setup

ricab:
Drop leading `set` calls for options that are now default in bash steps.
See https://is.gd/4HQTkt.

### DIFF
--- a/.github/actions/whitespace/action.yml
+++ b/.github/actions/whitespace/action.yml
@@ -14,8 +14,6 @@ runs:
       shell: bash
       id: derive-exclusions
       run: |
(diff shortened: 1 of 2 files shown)


ricab:
Hey @sharder996, I tried that and my IDE complained about a "forbidden property" and missing entry in steps:

image

I am not sure if the IDE is right. Also, some jobs in other workflows don't specify a shell, so they get a slightly different invocation. To use this across the board we might have to tweak and test them.

sharder996:
@ricab Ok, we can we leave it for now.

PR #4284, opened 2025-07-29
[rpc/CMakeLists.txt] add `/bigobj` flag

xmkg:
RPC Compilation fails due to the produced object size in Windows; this PR fixes that by passing the "/bigobj" flag.


Sploder12 (approved):
LGTM, I do this a lot locally already, it'll be nice to have it in

PR #4288, opened 2025-08-01
[sshfs-mount] make chown return 0 on Windows

xmkg:
chown is a platform-specific function which depends uid and gid. these values are not available in Windows, and the chown impl is by default returns -1. This causes filesystem operation failures in Windows hosts when sshfs mounts are used in guests. One example is as follows:

ubuntu@foo:~/Downloads$ echo "test" > foo.txt
-bash: foo.txt: Operation not permitted

This operation fails because it first lands in handle_open, which eventually calls chown to change the file's owner, which is a failure by default in Windows.

The patch fixes that by introducing a shim function, chown_shim which returns success when either uid or gid is set to no_id_info_available.

MULTI-2155

### DIFF
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -253,6 +253,20 @@ int reverse_id_for(const mp::id_mappings& id_maps, const int id, const int defau
                ? (default_found == id_maps.cend() ? default_id : default_found->first)
                : found->first;
 }
+
+int chown_shim(const char* path, unsigned int uid, unsigned int gid)
+{
+    // Do not bother calling the chown if IDs are set to no_id_available
+    // (e.g. in Windows)
+    if (uid == static_cast<decltype(gid)>(mp::no_id_info_available) ||
+        gid == static_cast<decltype(gid)>(mp::no_id_info_available))
+    {
+        return 0;
+    }
+
+    return MP_PLATFORM.chown(path, uid, gid);
+}
+
 } // namespace
 
 mp::SftpServer::SftpServer(SSHSession&& session,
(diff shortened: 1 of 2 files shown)


src/sshfs_mount/sftp_server.cpp
```diff
@@ -253,6 +253,20 @@ int reverse_id_for(const mp::id_mappings& id_maps, const int id, const int defau
                ? (default_found == id_maps.cend() ? default_id : default_found->first)
                : found->first;
 }
+
+int chown_shim(const char* path, unsigned int uid, unsigned int gid)
+{
+    // Do not bother calling the chown if IDs are set to no_id_available
+    // (e.g. in Windows)
+    if (uid == static_cast<decltype(gid)>(mp::no_id_info_available) ||
```
xmkg: decltype(uid)

Sploder12:
Looks like some tests need modification for Windows, macOS failure looks like a flakey test

xmkg:
> Looks like some tests need modification for Windows, macOS failure looks like a flakey test

Should be good to go now

Sploder12:
> Let me try to dig that up

Looks like we discussed throwing an exception but that got thrown out at some point. I think the jist was we didn't want it to silently do nothing or pretend to do something. Logging a warning should be adequate imo, wdyt?

xmkg:
> > Let me try to dig that up
> 
> Looks like we discussed throwing an exception but that got thrown out at some point. I think the jist was we didn't want it to silently do nothing or pretend to do something. Logging a warning should be adequate imo, wdyt?

The right thing to do would be to translate the UNIX perms to the Windows equivalents and perform the action, which is non-trivial. Logging it at the warning level would be quite spammy since it's a filesystem API -- maybe TRACE would be more appropriate. What do you think?

Sploder12:
> Logging it at the warning level would be quite spammy since it's a filesystem API -- maybe TRACE would be more appropriate. What do you think?

Good point, I think TRACE would be appropriate

sharder996 (approved):

ricab:
Hey @xmkg, @Sploder12, could one of you please confirm this should go into the patch release? Thanks.

xmkg:
> Hey @xmkg, @Sploder12, could one of you please confirm this should go into the patch release? Thanks.

Hey, yes this is the one we were discussing today.

PR #4289, opened 2025-08-04
[tests/cloud_init_iso] fix `read` mock stack overrun

xmkg:
read_returns_five_bytes_string mock was using std::strcpy to copy five characters from the string "NonJo". In reality, strcpy always terminates the destination with a '\0', which leads to an overflow since the destination is a 5-byte stack-allocated buffer.

This patch fixes that by replacing std::strcpy with std::memcpy.

MULTI-2156

### DIFF
--- a/tests/test_cloud_init_iso.cpp
+++ b/tests/test_cloud_init_iso.cpp
@@ -218,10 +218,16 @@ TEST_F(CloudInitIso, readsIsoFileJolietVolumeDescriptorMalformed)
 
     auto read_returns_five_bytes_string =
         [](std::ifstream& file, char* buffer, std::streamsize size) -> std::ifstream& {
-        [&] {
-            ASSERT_EQ(5, size);
-            ASSERT_NE(nullptr, buffer);
-        }();
+        if (size != 5)
+        {
+            throw std::invalid_argument{"Size must be 5."};
+        }
+
+        if (nullptr == buffer)
+        {
+            throw std::invalid_argument{"Buffer must not be null!"};
+        }
+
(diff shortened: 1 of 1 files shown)


tests/test_cloud_init_iso.cpp
```diff
@@ -217,8 +217,11 @@ TEST_F(CloudInitIso, readsIsoFileJolietVolumeDescriptorMalformed)
         .WillOnce(original_implementation_of_read);
 
     auto read_returns_five_bytes_string =
-        [](std::ifstream& file, char* buffer, std::streamsize) -> std::ifstream& {
-        std::strcpy(buffer, "NonJo");
+        [](std::ifstream& file, char* buffer, std::streamsize size) -> std::ifstream& {
+        EXPECT_EQ(5, size);
+        EXPECT_NE(nullptr, buffer);
```
Sploder12: Should these be `ASSERT`s to prevent an overrun in the case of an error?

xmkg: ASSERT_* macros cannot be used in functions returning something other than void. I can wrap them with an IILE, though.

xmkg: done

Sploder12: It seems `ASSERT` doesn't immediately end the test, just returns from the current function. So an IILE makes `ASSERT` equivalent to an `EXPECT`...

xmkg: ahhh. then it might be better to throw an exception here.

PR #4293, opened 2025-08-06
[3rd-party images] Retrieve Debian and Fedora images

### LINKED ISSUE #3541: multipass find does not show image name core, core18, core20, core22, core24 and appliance serie, on m2 macos
**Describe the bug**
The `multipass find` command on m2 macos seems to show an incomplete image list. 

**To Reproduce**
How, and what happened?
1. Install the multipass stable version 1.13.1
2. Run `multipass find`
3. See the image list below which is incomplete
```
multipass find
Image                       Aliases           Version          Description
20.04                       focal             20240513         Ubuntu 20.04 LTS
22.04                       jammy             20240514         Ubuntu 22.04 LTS
23.10                       mantic            20240514         Ubuntu 23.10
24.04                       noble,lts         20240523.1       Ubuntu 24.04 LTS

Blueprint                   Aliases           Version          Description
anbox-cloud-appliance                         latest           Anbox Cloud Appliance
charm-dev                                     latest           A development and testing environment for charmers
docker                                        0.4              A Docker environment with Portainer and related tools
jellyfin                                      latest           Jellyfin is a Free Software Media System that puts you in control of managing and streaming your media.
minikube                                      latest           minikube is local Kubernetes
ros-noetic                                    0.1              A development and testing environment for ROS Noetic.
ros2-humble                                   0.1              A development and testing environment for ROS 2 Humble.
```

**Expected behavior**
The macos shows the complete image list via `multipass find` just like windows
```
C:\Users\georg\canonical_projects\multipass-private_build
λ multipass find
Image                       Aliases           Version          Description
core                        core16            20200818         Ubuntu Core 16
core18                                        20211124         Ubuntu Core 18
core20                                        20230119         Ubuntu Core 20
core22                                        20230717         Ubuntu Core 22
core24                                        20240603         Ubuntu Core 24
20.04                       focal             20240513         Ubuntu 20.04 LTS
22.04                       jammy             20240514         Ubuntu 22.04 LTS
24.04                       noble,lts         20240523.1       Ubuntu 24.04 LTS
appliance:adguard-home                        20200812         Ubuntu AdGuard Home Appliance
appliance:mosquitto                           20200812         Ubuntu Mosquitto Appliance
appliance:nextcloud                           20200812         Ubuntu Nextcloud Appliance
appliance:openhab                             20200812         Ubuntu openHAB Home Appliance
appliance:plexmediaserver                     20200812         Ubuntu Plex Media Server Appliance

Blueprint                   Aliases           Version          Description
anbox-cloud-appliance                         latest           Anbox Cloud Appliance
charm-dev                                     latest           A development and testing environment for charmers
docker                                        0.4              A Docker environment with Portainer and related tools
jellyfin                                      latest           Jellyfin is a Free Software Media System that puts you in control of managing and streaming your media.
minikube                                      latest           minikube is local Kubernetes
ros-noetic                                    0.1              A development and testing environment for ROS Noetic.
ros2-humble                                   0.1              A development and testing environment for ROS 2 Humble.
```
**Logs**
cat multipassd.log
```
[2024-06-04T12:25:30.730] [debug] [update] Latest Multipass release available is version 1.13.1
[2024-06-04T12:25:30.806] [debug] [blueprint provider] Loading "anbox-cloud-appliance" v1
[2024-06-04T12:25:30.807] [debug] [blueprint provider] Loading "charm-dev" v1
[2024-06-04T12:25:30.808] [debug] [blueprint provider] Loading "docker" v1
[2024-06-04T12:25:30.808] [debug] [blueprint provider] Loading "jellyfin" v1
[2024-06-04T12:25:30.809] [debug] [blueprint provider] Loading "minikube" v1
[2024-06-04T12:25:30.809] [debug] [blueprint provider] Loading "ros-noetic" v1
[2024-06-04T12:25:30.809] [debug] [blueprint provider] Loading "ros2-humble" v1
[2024-06-04T12:25:30.815] [info] [rpc] gRPC listening on unix:/var/run/multipass_socket
[2024-06-04T12:25:30.816] [debug] [async task] fetch manifest periodically
[2024-06-04T12:25:30.816] [info] [daemon] Starting Multipass 1.13.1+mac
[2024-06-04T12:25:30.816] [info] [daemon] Daemon arguments: /Library/Application Support/com.canonical.multipass/bin/multipassd --verbosity debug
[2024-06-04T12:25:30.995] [info] [VMImageHost] Did not find any supported products in "appliance"
[2024-06-04T12:25:37.616] [debug] [async task] fetch manifest from the internet
[2024-06-04T12:25:37.813] [info] [VMImageHost] Did not find any supported products in "appliance"
[2024-06-04T12:46:52.956] [debug] [async task] fetch manifest from the internet
[2024-06-04T12:46:53.144] [info] [VMImageHost] Did not find any supported products in "appliance"
[2024-06-04T12:46:54.077] [debug] [blueprint provider] Loading "anbox-cloud-appliance" v1
[2024-06-04T12:46:54.078] [debug] [blueprint provider] Loading "charm-dev" v1
[2024-06-04T12:46:54.079] [debug] [blueprint provider] Loading "docker" v1
[2024-06-04T12:46:54.079] [debug] [blueprint provider] Loading "jellyfin" v1
[2024-06-04T12:46:54.080] [debug] [blueprint provider] Loading "minikube" v1
[2024-06-04T12:46:54.080] [debug] [blueprint provider] Loading "ros-noetic" v1
[2024-06-04T12:46:54.080] [debug] [blueprint provider] Loading "ros2-humble" v1
```

**Additional info**
 - OS: macOS Sonoma 14.5, m2 CPU
- `multipass version`
multipass   1.13.1+mac
multipassd  1.13.1+mac
qemu
**Additional context**
1. firewall has always been off. 
2. The above log says "did not find any supported products in "appliance"", not sure why only this one failed. 
3. https://cdimage.ubuntu.com/ubuntu-core/24/stable/current/ can be accessed and can be viewed in the web-browser, why says the server is on the internet connection is fine.
4. Did many times uninstall via sudo `sh "/Library/Application Support/com.canonical.multipass/uninstall.sh"` and cleaned up the cache data, the problem still persists.

sharder996:
Several notables changes presented here:

1. Debian and Fedora images presented in `multipass find`:

Details for 3rd-party cloud images are located in `data/distributions/distribution-info.json` and roughly follow the format of `VMImageInfo` and `simplestreams`. Once this PR is merged in, the `CustomVMImageHost` will fetch this file from the Multipass GitHub repository. For local developer testing, the environment variable `MULTIPASS_DISTRIBUTIONS_URL` can be set to a local directory, i.e. `file:///Users/you/path/to/your/manifest.json`. Reported images are sorted in descending order by their `remote_name` placing 3rd-party images at the bottom of the list, since the remote name is empty.

2. Ubuntu Core images are now fetched via the `UbuntuVMImageHost` class:

Ubuntu Core images utilize [simplestreams](https://cdimage.ubuntu.com/ubuntu-core/streams/v1/com.ubuntu.cdimage:ubuntu-core.json) for distribution. Several tweaks were made to `simple_streams_manifest.cpp` to accommodate them. As a result, core images are also now reported on ARM64 CPU architectures.

3. Golden files extracted out of `test_output_formatter.cpp`:

Test files for the formatters are now located under `tests/test_data/formatters/<formatter>/`. This eliminates the need for escaping characters, line length linting, and translating text output from the command line into C++ strings.

4. Removed hardcoded "Ubuntu" strings within formatters, installers, and CLI

- remove hardcoded "Ubuntu" strings from `info` and `list` formatters
- retroactively populate the image vault with new `OS` field
  - fix bug where query type was not being loaded from the DB correctly
  - this function will need to be removed in a future release of Multipass, although the function is idempotent
  - all pre-existing instances will have the field populated with "Ubuntu"
- remove specific references to Ubuntu within the CLI and installer in favour of more generic "cloud VMs/instances"

5. Fedora compatible disk info fetching

- Include BTRFS filesystems when counting disk info
- Filter out filesystems that are mounted on multiple directories

6. Drive-by fixes:

- Determining the location of `test_data` is now done at compile time rather than runtime allowing the use of the location in objects that are constructed at compile time.
- Moved image host classes to their own subdirectory and renamed some classes to follow the established inheritance format of `virtual` -> `base` -> implementation.
- `test_data` persistence optimization:
  - When templated test files, i.e. `*\.in$`, are edited or new files are added to `test_data`, a reconfigure is triggered by cmake.
  - Non-templated test files are persisted into the build directory at compile time if they have been edited.
- New exception type; `ImageNotFoundException`, created for when an image is not found.
- Optimized sending of `remote_name` in the `FindReply` `protobuf` message type.

Fixes #3541

---

MULTI-2048
MULTI-2074
MULTI-2151
MULTI-2153

### DIFF
--- a/src/client/cli/formatter/table_formatter.cpp
+++ b/src/client/cli/formatter/table_formatter.cpp
@@ -36,33 +36,27 @@ void format_images(Dest&& dest,
 {
     fmt::format_to(dest, "{:<28}{:<18}{:<17}{:<}\n", "Image", "Aliases", "Version", "Description");
 
-    std::vector<const mp::FindReply_ImageInfo*> sorted_images;
-    sorted_images.reserve(static_cast<size_t>(images_info.size()));
-    for (const auto& img : images_info)
-    {
-        sorted_images.push_back(&img);
-    }
-
+    auto sorted_images = images_info;
     std::sort(sorted_images.begin(),
               sorted_images.end(),
-              [](const mp::FindReply_ImageInfo* a, const mp::FindReply_ImageInfo* b) {
-                  return a->remote_name() > b->remote_name() ||
-                         (a->aliases().size() && b->aliases().size() &&
-                          (a->aliases()[0] < b->aliases()[0]));
+              [](const mp::FindReply_ImageInfo& a, const mp::FindReply_ImageInfo& b) {
+                  return a.remote_name() > b.remote_name() ||
+                         (a.aliases().size() && b.aliases().size() &&
+                          (a.aliases()[0] < b.aliases()[0]));
               });
 
-    for (const auto* image : sorted_images)
+    for (const auto& image : sorted_images)
     {
-        auto aliases = image->aliases();
+        auto aliases = image.aliases();
         mp::format::filter_aliases(aliases);
 
         fmt::format_to(
             dest,
             "{:<28}{:<18}{:<17}{:<}\n",
-            mp::format::image_string_for(image->remote_name(), aliases[0]),
+            mp::format::image_string_for(image.remote_name(), aliases[0]),
             fmt::format("{}", fmt::join(aliases.cbegin() + 1, aliases.cend(), ",")),
-            image->version(),
-            fmt::format("{}{}", image->os().empty() ? "" : image->os() + " ", image->release()));
+            image.version(),
+            fmt::format("{}{}", image.os().empty() ? "" : image.os() + " ", image.release()));
     }
     fmt::format_to(dest, "\n");
 }


src/daemon/daemon_config.cpp
```diff
@@ -181,8 +159,12 @@ std::unique_ptr<const mp::DaemonConfig> mp::DaemonConfigBuilder::build()
...
                  UbuntuVMImageRemote{"https://cloud-images.ubuntu.com/",
                                      "buildd/daily/",
-                                     &admits_snapcraft_image,
-                                     std::make_optional<QString>(mp::mirror_key)}}},
+                                     mp::image_mutators::snapcraft_mutator,
+                                     std::make_optional<QString>(mp::mirror_key)}},
+                {mp::core_remote,
+                 UbuntuVMImageRemote{"https://cdimage.ubuntu.com/",
+                                     "ubuntu-core/",
+                                     mp::image_mutators::core_mutator}}},
```
ricab, line 164: Not meant as a review. I opened https://github.com/canonical/multipass/issues/4431 and started doing it, but then remembered you might have already done that, so I came looking 🙂 

Doesn't this need to come first if we are to maintain image order in `multipass find`?

ricab:
Fixes #4431

src/daemon/daemon_config.cpp
sharder996: Not sure what you're referring to by "*this need to come first*", but image order should remain unchanged.

packaging/macos/WELCOME.html
```diff
@@ -7,11 +7,11 @@
 </style>
 </head>
 <body>
-<h2>Multipass: Ubuntu at your fingertips.</h2>
+<h2>Multipass: Cloud VMs at your fingertips.</h2>
```
jimporter, line 10: I thought one of the use cases for Multipass was to be able to run a dev environment in a VM (e.g. in the [Rockcraft tutorial](https://documentation.ubuntu.com/rockcraft/latest/tutorial/hello-world/)). Would it be better to say "VMs" instead of "Cloud VMs" here (and elsewhere)?

I'm not sure how much discussion there has already been about this topic. One area I found a bit confusing when reading the docs initially was whether Multipass was for dev VMs or for testing out a cloud env. I *think* the idea is that it's both?

(Ditto this comment for the other places that mention "cloud VMs".)

src/daemon/daemon.cpp
```diff
@@ -159,23 +159,27 @@ auto make_cloud_init_vendor_config(const mp::SSHKeyProvider& key_provider,
...
-    auto pollinate_user_agent_string =
-        fmt::format("multipass/version/{} # written by Multipass\n", multipass::version_string);
-    pollinate_user_agent_string +=
-        fmt::format("multipass/driver/{} # written by Multipass\n", backend_version_string);
-    pollinate_user_agent_string += fmt::format("multipass/host/{} # written by Multipass\n",
-                                               multipass::platform::host_version());
-    pollinate_user_agent_string += fmt::format("multipass/alias/{}{} # written by Multipass\n",
-                                               !remote_name.empty() ? remote_name + ":" : "",
-                                               pollinate_alias);
+    if (request->image() != "fedora")
```
jimporter, line 165: Add a comment here about why, for future readers? Just looking at the code, I'm not sure what makes Fedora a special case, and whether I'd need to add another distro here in the future.

(In the long term, there's probably some abstract rule that makes sense here, but one step at a time...)

src/daemon/default_vm_image_vault.cpp
```diff
@@ -251,6 +252,12 @@ mp::DefaultVMImageVault::DefaultVMImageVault(std::vector<VMImageHost*> image_hos
       prepared_image_records{load_db(cache_dir.filePath(image_db_name))},
       instance_image_records{load_db(data_dir.filePath(instance_db_name))}
 {
+    // TODO: Remove after Multipass 1.17
+    // The OS field will be unpopulated for existing images in the vault. As of 1.16, the only
+    // images in the prepared image records are Ubuntu images. Therefore, we can safely assume that
+    // if the OS field is empty, it was a previously existing Ubuntu cloud image. The same can be
+    // said for instance image records with instances created with the Alias Query::Type.
+    amend_db();
```
jimporter, line 260: I haven't looked at the vault code yet... do we tag the vault data with a version identifier? From this comment, I'm guessing we don't. It might be a good idea to add eventually (i.e. not in this PR), since then it's easy to migrate the vault from one schema to another.

This case is easy enough since it's just adding a missing field, but there might be more complicated updates in the future.

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+std::vector<std::pair<std::string, mp::VMImageInfo>> mp::CustomVMImageHost::all_info_for(
+    const Query& query)
+{
+    std::vector<std::pair<std::string, mp::VMImageInfo>> images;
+
+    auto image = info_for(query);
+    if (image != std::nullopt)
+        images.push_back(std::make_pair(query.remote_name, *image));
```
jimporter: I think this should work, and be slightly more efficient?
```suggestion
    if (auto image = info_for(query))
        images.emplace_back(query.remote_name, std::move(*image));
```

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,225 @@
...
+std::vector<mp::VMImageInfo> mp::CustomVMImageHost::all_images_for(const std::string& remote_name,
+                                                                   const bool allow_unsupported)
+{
+    std::vector<mp::VMImageInfo> images;
+    auto custom_manifest = manifest_from(remote_name);
+    std::copy(custom_manifest->products.begin(),
+              custom_manifest->products.end(),
+              std::back_inserter(images));
+
+    return images;
```
jimporter: Can we just return `custom_manifest->products` here? Since we're returning by value, we'll get a copy automatically. We also probably want to check if the manifest is non-null, since it looks like that would be possible:

```suggestion
    if (auto custom_manifest = manifest_from(remote_name))
        return custom_manifest->products;
    return {};
```

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+std::vector<std::string> mp::CustomVMImageHost::supported_remotes()
+{
+    return std::vector<std::string>{remote};
```
jimporter: We could omit the `std::vector<std::string>`. Either just `return remote;` or `return {remote};` if you want to make it more obvious that it's a copy.

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    throw mp::ImageNotFoundException(full_hash);
+}
+
+void mp::CustomVMImageHost::fetch_manifests(const bool force_update)
+{
+    try
+    {
+        std::unique_ptr<mp::CustomManifest> custom_manifest = std::make_unique<mp::CustomManifest>(
```
jimporter:
```suggestion
        auto custom_manifest = std::make_unique<mp::CustomManifest>(
```

src/image_host/ubuntu_image_host.cpp
```diff
@@ -125,36 +107,24 @@ std::vector<std::pair<std::string, mp::VMImageInfo>> mp::UbuntuVMImageHost::all_
...
-                continue;
-
             if (!info->supported && !query.allow_unsupported)
                 throw mp::UnsupportedImageException(query.release);
 
-            images.push_back(std::make_pair(
-                remote_name,
-                with_location_fully_resolved(QString::fromStdString(remote_url_from(remote_name)),
-                                             *info)));
+            images.push_back(std::make_pair(remote_name, *info));
```
jimporter: Very minor change, but I usually find `emplace_back` easier to read:
```suggestion
            images.emplace_back(remote_name, *info);
```

src/image_host/ubuntu_image_host.cpp
```diff
@@ -125,36 +107,24 @@ std::vector<std::pair<std::string, mp::VMImageInfo>> mp::UbuntuVMImageHost::all_
...
-                    (entry.supported || query.allow_unsupported) &&
+                if (entry.id.startsWith(key) && (entry.supported || query.allow_unsupported) &&
                     found_hashes.find(entry.id.toStdString()) == found_hashes.end())
                 {
-                    images.push_back(
-                        std::make_pair(remote_name,
-                                       with_location_fully_resolved(
-                                           QString::fromStdString(remote_url_from(remote_name)),
-                                           entry)));
+                    images.push_back(std::make_pair(remote_name, entry));
```
jimporter: Likewise here:
```suggestion
                    images.emplace_back(remote_name, entry);
```

tests/apple/mock_apple_virtualization_wrapper.h
```diff
@@ -0,0 +1,28 @@
...
+
+// namespace multipass::test
+// {
+// struct MockAppleVirtualizationWrapper : public apple::AppleVirtualizationInterface
+// {
+//     MOCK_METHOD(hyperv::OperationResult, create_vm, (), (const, override));
+//     MOCK_METHOD(hyperv::OperationResult, start_vm, (), (const, override));
+//     MOCK_METHOD(hyperv::OperationResult, stop_vm, (), (const, override));
+// };
+} // namespace multipass::test
```
jimporter: This is all commented out. Is this just in-progress?

tests/apple/test_apple_backend.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <tuple>
+
+namespace mp = multipass;
+namespace mpl = multipass::logging;
+namespace mpt = multipass::test;
+using namespace testing;
+
+namespace multipass::test
+{
+}
```
jimporter: Likewise. Is this just in-progress?

packaging/macos/WELCOME.html
sharder996: I believe this would be a good question for @geoffreynyaga. Its my impression that Multipass is officially a product for launching Cloud VMs. Dev VMs are just a convenient parallel application that is easy to also include.

src/daemon/daemon.cpp
sharder996: The pollinate server, which is hosted by Canonical, is also scheduled for deprecation in the near future.

src/daemon/default_vm_image_vault.cpp
sharder996: Yeah, a version key might be beneficial in the future.

tests/apple/mock_apple_virtualization_wrapper.h
sharder996: Oops, looks like some unrelated work snuck in while I was rebasing

tests/apple/test_apple_backend.cpp
sharder996: idem

src/client/cli/formatter/format_utils.cpp
```diff
@@ -102,16 +103,15 @@ mp::Formatter* mp::format::formatter_for(const std::string& format)
     return nullptr;
 }
 
-void mp::format::filter_aliases(
-    google::protobuf::RepeatedPtrField<multipass::FindReply_AliasInfo>& aliases)
+void mp::format::filter_aliases(google::protobuf::RepeatedPtrField<std::string>& aliases)
 {
     for (auto i = aliases.size() - 1; i >= 0; i--)
     {
         if (aliases.size() == 1) // Keep at least the first entry
```
tobe2098, line 110: Why this check inside the for loop? Could an early return before the for loop be more readable?

src/client/cli/formatter/table_formatter.cpp
```diff
@@ -36,18 +36,32 @@ void format_images(Dest&& dest,
...
+    for (const auto& img : images_info)
+    {
+        sorted_images.push_back(&img);
+    }
+
+    std::stable_sort(sorted_images.begin(),
+                     sorted_images.end(),
+                     [](const mp::FindReply_ImageInfo* a, const mp::FindReply_ImageInfo* b) {
+                         return a->remote_name() > b->remote_name();
+                     });
```
tobe2098: I agree with the idea of remote_name sorting, but should we not have a more established order? If the implementation details that result in the starting order of images_info change, the order in versions would change at that point as well. 
We should probably decide for an order of versions (desc or asc) on the CLI redesign, but for now we should probably make sure that the order we are currently enforcing is enforced irrespective of future unrelated source changes. What do you think?

tests/test_data/formatters/yaml/empty_networks_reply.yaml
```diff
@@ -0,0 +1 @@
+
```
tobe2098, line 1: Are two empty .yaml necessary? It seems like a single empty .yaml would do the job. What do you think?

tests/test_data/formatters/json/empty_networks_reply.json
```diff
@@ -0,0 +1,4 @@
+{
+    "list": [
+    ]
+}
```
tobe2098, line 4: Idem for empty .jsons

src/image_host/ubuntu_image_host.cpp
```diff
@@ -237,32 +194,35 @@ std::vector<std::string> mp::UbuntuVMImageHost::supported_remotes()
...
 {
-    auto fetch_one_remote = [this, is_force_update_from_network](
-                                const std::pair<std::string, UbuntuVMImageRemote>& remote_pair)
+    auto fetch_one_remote =
+        [this, force_update](const std::pair<std::string, UbuntuVMImageRemote>& remote_pair)
         -> std::pair<std::string, std::unique_ptr<SimpleStreamsManifest>> {
-        const auto& [remote_name, remote_info] = remote_pair;
+        const auto& remote_name = remote_pair.first;
+        const auto& remote_info = remote_pair.second;
+
```
tobe2098: Is there a reason why this cannot be a structured binding?

src/daemon/ubuntu_image_host.cpp
```diff
@@ -315,69 +275,36 @@ const mp::VMImageInfo* mp::UbuntuVMImageHost::match_alias(
...
+      image_mutator{custom_image_mutator},
       mirror_key(std::move(mirror_key))
 {
 }
 
-const QString mp::UbuntuVMImageRemote::get_url() const
-{
-    return get_mirror_url().value_or(get_official_url());
-}
-
```
tobe2098, line 380: Very nice redesign to have the host incorporated in the simplestreams manifest and do away with the previous extra complexity

tests/test_custom_image_host.cpp
```diff
@@ -17,277 +17,195 @@
...
-
-        ASSERT_TRUE(info);
-        EXPECT_EQ(info->image_location, url.toDisplayString());
-        EXPECT_EQ(info->id, id);
-        EXPECT_EQ(info->release, release);
-        EXPECT_EQ(info->release_title, release_title);
-        EXPECT_TRUE(info->supported);
-        EXPECT_EQ(info->version, QLocale::c().toString(QDateTime::currentDateTime(), "yyyyMMdd"));
-    }
-}
```
tobe2098, line 109: Just double-checking, are the tests removed due to redundancy in covering the same code (i.e., this one tests code that is already tested in allInfoForNoRemoteReturnsOneAliasMatch).

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,221 @@
...
+constexpr auto no_remote{""};
+constexpr auto manifest_endpoint{"https://raw.githubusercontent.com/canonical/multipass/refs/heads/"
+                                 "main/data/distributions/distribution-info.json"};
+
+auto get_manifest_url()
+{
+    return qEnvironmentVariable(mp::distributions_url_env_var).isEmpty()
+               ? manifest_endpoint
+               : qEnvironmentVariable(mp::distributions_url_env_var);
+}
```
tobe2098, line 49: Apart from being useful to test the PR in particular, are there other circumstances in which the env var is going to be of use?

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,221 @@
...
+    for (auto it = root_obj.begin(); it != root_obj.end(); ++it)
+    {
+        const QString distro_name = it.key();
+        const QJsonObject distro_obj = it.value().toObject();
+
+        QStringList aliases = distro_obj.value("aliases").toString().split(",", Qt::SkipEmptyParts);
+        for (QString& alias : aliases)
+            alias = alias.trimmed();
+
+        images.push_back(mp::VMImageInfo{aliases,
```
tobe2098: Could consider using emplace here, WDYT?

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,221 @@
...
+    {
+        action(manifest.first, info);
+    }
+}
+
+mp::VMImageInfo mp::CustomVMImageHost::info_for_full_hash_impl(const std::string& full_hash)
+{
+    for (const auto& product : manifest.second->products)
+    {
+        if (product.id.toStdString() == full_hash)
```
tobe2098: Should this comparison be case insensitive?

src/image_host/custom_image_host.cpp
```diff
@@ -0,0 +1,221 @@
...
+    throw mp::ImageNotFoundException(full_hash);
+}
+
+void mp::CustomVMImageHost::fetch_manifests(const bool force_update)
+{
+    try
+    {
+        auto custom_manifest = std::make_unique<mp::CustomManifest>(
+            fetch_image_info(arch, url_downloader, force_update));
+        manifest = std::make_pair(no_remote, std::move(custom_manifest));
```
tobe2098, line 202: I understand that ImageHosts are found remote-wise in a map in the defaultVMImageVault, but since manifest.first is still used in for_each_entry_do_impl and manifest_from, is it safe/intended to have empty remotes in manifest.first for CustomVMImageHosts?

src/client/cli/formatter/format_utils.cpp
sharder996: I think moving it outside the loop would change the behaviour of the loop. We want to keep at least one alias. Considering the situation where the filter applies to all aliases in the passed container, we want to avoid deleting the last aliases; indexed at `aliases[0]` or when `aliases.size() == 1`.

src/client/cli/formatter/table_formatter.cpp
sharder996: You're referring to sorting within a `remote_name`?

tests/test_data/formatters/yaml/empty_networks_reply.yaml
sharder996: I'm not sure what I would name the merged file.... It's not exactly an empty reply since an empty reply can vary depending on the Multipass command. I feel like not being able to succinctly name the output file suggests that they should remain separate.

This might also be an artifact of mismatching CLI commands. Might be something to consider for the CLI evaluation. I'll make a command in the JIRA card.

tests/test_data/formatters/json/empty_networks_reply.json
sharder996: idem

src/image_host/ubuntu_image_host.cpp
sharder996: I know there was a reason, but I believe C++20 means we can now!

src/image_host/custom_image_host.cpp
sharder996: It's meant for developer use in case we want to test adding different distributions or versions. I took inspiration from the [late blueprints feature](https://github.com/canonical/multipass/blob/0e82c4bb73440675e7518b95796e4d2702b3f3f9/include/multipass/constants.h#L47)

src/image_host/custom_image_host.cpp
sharder996: Its an empty string, not a null string so it should be no different than a string with a length `n > 0`. All string operations are still valid.

tests/test_custom_image_host.cpp
sharder996: Its a bit confusing because I used the same filename, but implemented an image host in a different way. We're not hardcoding values in the `custom_image_host.cpp` anymore, so the `ExpectedDataSuite` isn't really applicable anymore.

src/client/cli/formatter/format_utils.cpp
tobe2098: Oh I see, you are correct.

src/client/cli/formatter/table_formatter.cpp
tobe2098: Yes, sorry if it was unclear. If some implementation detail changes the within remote_name order, the behavior will be changing in an "out of our control" manner. If we incorporate CLI tests eventually, `multipass find` could cause failures in these order changes.

src/image_host/ubuntu_image_host.cpp
tobe2098: Oh, I thought you moved away from them, did not think that the rebase made it look that way.

src/client/cli/formatter/table_formatter.cpp
tobe2098: Looks good, but then a stable_sort is not needed anymore, we can use `std::sort` instead. WDYT?

src/image_host/custom_image_host.cpp
tobe2098: I meant in terms of identifying the `CustomVMImageHost` from other custom hosts. I see that the current implementation only has one CustomVMImageHost, but functions like `info_for` calling `manifest_from` use the remote name from the query to identify the remote by comparing it to the one in the object. Are queries to the customVMImageHost always empty strings?
From the code in launch.cpp and find.cpp it seems like the custom remote is considered to be an empty string as well, which would not allow the user to `multipass find custom:`. Maybe this is more related to the CLI revisit in UX. WDTY?

tobe2098 (commented):
Looking very good! Just a couple of questions and I think the PR is there.

PD: Curiously enough, the default user in Fedora and also Debian (I think) is still `ubuntu`, which is probably related to the default cloud-init running in intialization. I mention it because I saw the conversation about default vs ubuntu user.

src/image_host/custom_image_host.cpp
sharder996: > From the code in launch.cpp and find.cpp it seems like the custom remote is considered to be an empty string as well, which would not allow the user to multipass find custom:. Maybe this is more related to the CLI revisit in UX. WDTY?

This is correct.

There is also other baked in logic in Multipass where all remotes except the `release` remote get printed out in the output of `multipass find`. I didn't want to have `custom:debian` in `find`, so I just continued to use an empty string as the remote. This is unchanged behaviour from the Ubuntu Core images that used to be fetched with the `CustomImageHost`.

src/image_host/ubuntu_image_host.cpp
sharder996: Previously, there was an issue with passing an element from a structured binding into a lambda, but I believe that has been changed now in C++20.

sharder996:
> PD: Curiously enough, the default user in Fedora and also Debian (I think) is still `ubuntu`, which is probably related to the default cloud-init running in intialization. I mention it because I saw the conversation about default vs ubuntu user.

This was discussed and the consensus was that we would retain `ubuntu` as the default user and revisit at a later time or if users complained.

packaging/macos/WELCOME.html
geoffreynyaga: @jimporter 

This is a great question. 
When I joined the team recently, I was advised that we are strongly pushing Multipass to become a 'pocket cloud' solution and strongly leaning on launching cloud VMs. I know its confusing since these two descriptions are in the landing page of our docs
```text
Multipass is a tool to generate cloud-style Ubuntu VMs quickly on Linux, macOS and Windows. 

It provides a simple but powerful CLI that enables you to quickly access an Ubuntu command line or create your own local mini-cloud.
```
and

``` text
... developers can use Multipass to prototype cloud deployments and to create fresh, customised Linux dev environments on any machine
```
There's a great deal of marketing landing pages and blogs that are upcoming and a clearer description will be issued.

sharder996: @geoffreynyaga Should I change the wording or can we merge this PR and revisit this at a later date?

PR #4294, opened 2025-08-08
[json-utils] implement file lock for write_json function

xmkg:
The write_json might be called from different threads and even from different processes concurrently. In some platforms, this has been causing intermittent failures like:

> multipass delete foo-bar
Could not commit transactional file; filename: C:/Users/musti/AppData/Local/Temp/multipass-test/data/multipassd-vm-instances.json, error: Access is denied

This patch fixes that by introducing an interprocess lock file to synchronize writes on per-file basis. Also, the patch introduces a retry logic with backoff that includes a jitter to decrease the chances of a .commit() collision even further.

This is mostly beneficial for Windows, but makes the whole process safer for the other platforms as well.

The lockfile guarantees that no concurrent write_json can happen to the same file at the same time. The retry logic ensures that the commit would be retried if the file has been locked by an external factor (e.g., Windows indexing, antiviruses, etc). 

Added some logging to help with troubleshooting.

MULTI-2159

### DIFF
--- a/tests/test_daemon.cpp
+++ b/tests/test_daemon.cpp
@@ -1109,6 +1109,10 @@ TEST_F(DaemonCreateLaunchAliasTestSuite, blueprintFoundDoesNotMountUnwrittableWo
         .WillOnce(Return(temp_dir.path()));
 
     auto [mock_file_ops, guard] = mpt::MockFileOps::inject();
+    EXPECT_CALL(*mock_file_ops, setStaleLockTime(_, _)).WillRepeatedly(Return());
+    EXPECT_CALL(*mock_file_ops, tryLock(_, _)).WillRepeatedly(Return(true));
+    EXPECT_CALL(*mock_file_ops, write(_, _)).WillRepeatedly(Return(1234));
+    EXPECT_CALL(*mock_file_ops, commit(_)).WillRepeatedly(Return(true));
     EXPECT_CALL(*mock_file_ops, open(_, _)).WillRepeatedly(Return(true));
     EXPECT_CALL(*mock_file_ops, write(_, _)).WillRepeatedly(Return(1234));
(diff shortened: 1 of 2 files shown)


xmkg:
This patch fixes the issue in the environment where I've discovered it (Windows w/ cli-tests). The lockfile completely eliminates the problem. I've commented out the lockfile part to see if retry logic also fixes it, and it does:

```
[2025-08-08T15:20:59.776] [debug] [instance5d51eda4-02a0-49ca-bf30-783a83c1e091] Dropping cached SSH session        
[2025-08-08T15:20:59.776] [debug] [daemon] Instance deleted: instance5d51eda4-02a0-49ca-bf30-783a83c1e091
[2025-08-08T15:20:59.784] [debug] [JsonUtils] Saved file `C:/Users/musti/AppData/Local/Temp/multipass-test/data/multipassd-vm-instances.json` successfully in attempt #1
^^ this is probably from another thread
[2025-08-08T15:20:59.784] [warning] [JsonUtils] Failed to write `C:/Users/musti/AppData/Local/Temp/multipass-test/data/multipassd-vm-instances.json` in attempt #1 (reason: Access is denied.), will retry after 22ms ms delay.
^^ first attempt fails since the thread above is writing to the file
[2025-08-08T15:20:59.826] [debug] [JsonUtils] Saved file `C:/Users/musti/AppData/Local/Temp/multipass-test/data/multipassd-vm-instances.json` successfully in attempt #2
^^ second attempt succeeds after giving a 22ms breathing room to the first thread
```

For reference, the command that I run the `cli-tests` is as follows: `python -m pytest -o log_cli_level=100 -s --tb=long -vv tools/cli_tests --build-root ../multipass-private-2/build/bin/ --data-root=$env:TEMP/multipass-test --remove-all-instances --print-cli-output --print-daemon-output --driver hyperv_api -m launch`

sharder996 (approved):
LGTM

PR #4286, opened 2025-07-30
[3rd-party] Upgrade QEMU to v10.0.3

sharder996:
Fixes #3851 
Fixes #4111 

---

MULTI-1716

PR #4297, opened 2025-08-11
[flutter] Introduce official .gitignore file to src/client/gui

xmkg:
Some of the build artifacts like generated_plugin_registrant are
mistakenly version-controlled because of the incomplete .gitignore
rules we have for Flutter. This patch adds the .gitignore file from
the official Flutter repository to avoid that.

Also cherry-picked @sharder996's QEMU Python dep fix to make this PR a complete CI fix.

MULTI-2160


xmkg:
@sharder996 @Sploder12, could you review this to unblock the CI at your earliest convenience?

Sploder12 (approved):
It is ignoring the proper files! Including the entire flutter `.gitignore` seems a bit overkill but I'd rather cover too much than not enough

xmkg:
> It is ignoring the proper files! Including the entire flutter `.gitignore` seems a bit overkill but I'd rather cover too much than not enough

The intent is to be in sync with the upstream Flutter .gitignore, so we don't have to worry about it anymore. This way, the .gitignore content upgrade can go hand-in-hand with the Flutter upgrade itself. Maybe we can even automatically copy it from the submodule in the future.

.github/workflows/windows-macos.yml
```diff
@@ -164,6 +164,12 @@ jobs:
           python \
           wget
 
+    - name: Combine the macOS package
```
sharder996: Wait! This isn't appropriately named. My bad. I'll fix in #4286

sharder996 (changes requested):

sharder996 (approved):

PR #4299, opened 2025-08-11
[rust] init multipass crate & oxidize petname

levkropp:
This pull request introduces Rust to Multipass by migrating the petname (name generation e.g. `hearty-kangaroo`) functionality from C++ to Rust and integrates it into the build and runtime environments. This transition involves replacing the old C++ name generator with a new Rust implementation, updating the build system to compile and link the Rust static library, and adapting relevant CI and packaging scripts to support Rust. The changes also update C++ code to use the Rust FFI for name generation and remove legacy C++ petname code.

**Petname/name generation migration to Rust:**

* Added new Rust crate for petname functionality (`Cargo.toml`, `build.rs`, `src/lib.rs`), exposing FFI functions for name generation and tests. 
* Updated C++ code (`src/client/gui/ffi/dart_ffi.cpp`, `src/daemon/daemon.cpp`) to use Rust FFI for petname generation instead of the old C++ `NameGenerator`.
* Removed legacy C++ petname and name generator code (`include/multipass/name_generator.h`, `src/petname/make_name_generator.cpp`).

**Build system and CI integration:**

* Updated CMake configuration to build the Rust static library and link it as an interface library for petname, removing previous code-generation steps. (`src/petname/CMakeLists.txt`)
* Added Rust-related files to CI triggers and installed Rust toolchain in CI workflows for Windows/macOS (`.github/workflows/dynamic-ci.yml`, `.github/workflows/windows-macos.yml`). 

**Packaging and environment updates:**

* Added `rustup` as a build dependency for Debian and Snap packaging, and ensured the Rust toolchain is initialized during Snap builds (`debian/control`, `snap/snapcraft.yaml`)

**Daemon configuration cleanup:**

* Removed all references to the old C++ `NameGenerator` from daemon configuration code (`src/daemon/daemon_config.cpp`, `src/daemon/daemon_config.h`).


src/daemon/daemon.cpp
```diff
@@ -19,8 +19,10 @@
 #include "base_cloud_init_config.h"
 #include "instance_settings_handler.h"
 #include "runtime_instance_info_helper.h"
+#include "rust/cxx.h"
```
ricab: As we discussed elsewhere, we can probably do away with this.

src/daemon/daemon.cpp
```diff
@@ -19,8 +19,10 @@
...
 #include "snapshot_settings_handler.h"
 
+#include <multipass-petname/src/lib.rs.h>
```
ricab: IIUC, this is generated, so I would say it is more distant than the usual headers in multipass/, right? 

There could be cases where dependencies go back and forth between rust and C++, but by default I would say that regular C++ header are closer kin than generated ones from the cxx crate. In that sense, I suggest moving this to a separate section, just after all the `<multipass/...>` includes, but before any third party.

crates/petname/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    petname.make_name()
+}
+
+// CXX FFI Bridge for this module
+#[cxx::bridge(namespace = "multipass::petname")]
+mod ffi {
+    extern "Rust" {
+        type Petname;
+        fn new_petname(num_words: i32, separator: &str) -> Box<Petname>;
+        fn make_name(petname: &mut Petname) -> String;
```
matthew-hagemann: I'd recommend looking at returning results here, which are [supported binding](https://cxx.rs/binding/result.html) in `cxx`. I don't think the errors you'll get returned from panics will be very ergonomic, which will make them harder to reason with in more complex crates. 

It will also make tests much easier to write (see next comment)

crates/petname/tests/petname_tests.rs
```diff
@@ -0,0 +1,69 @@
+use multipass_petname::*;
+use std::collections::HashSet;
+
+fn split_name<'a>(name: &'a str, separator: &str) -> Vec<&'a str> {
+    name.split(separator).collect()
+}
+
+#[test]
```
matthew-hagemann: Another advantage of having your functions return results is that you can add test cases for error path's and easily assert that you get the Result you expect without panicking the test. Currently you are only covering the happy paths.

crates/petname/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+}
+
+pub fn new_petname(num_words: i32, separator: &str) -> Box<Petname> {
+    let num_words = match num_words {
+        1 => NumWords::One,
+        2 => NumWords::Two,
+        3 => NumWords::Three,
+        _ => panic!("Invalid number of words"),
```
matthew-hagemann: I'd avoid explicitly panicking from function input. Looking at what is supported in `cxx`'s [Result bindings](https://cxx.rs/binding/result.html):

> The Rust implementation (outside of the bridge module) may pick any error type as long as it has a std::fmt::Display impl.

I would define a crate error you can use in your functions and return that in error cases.

Something like this:
```rust
  #[derive(Debug)]
  pub enum PetnameError {
      InvalidWordCount(i32),
      EmptyWordList,
  }

  impl fmt::Display for PetnameError {
      fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
          match self {
              PetnameError::InvalidWordCount(n) => write!(f, "Invalid word count: {}", n),
              PetnameError::EmptyWordList => write!(f, "Word list is empty"),
          }
      }
  }

  pub fn new_petname(num_words: i32, separator: &str) -> Result<Box<Petname>, PetnameError> {
      let num_words = match num_words {
          1 => NumWords::One,
          2 => NumWords::Two,
          3 => NumWords::Three,
          _ => return Err(PetnameError::InvalidWordCount(num_words)),
      };
      Ok(Box::new(Petname::new(num_words, separator)))
  }
```

And then you can have unhappy path tests that look like this:
```rust
  #[test]
  fn new_petname_rejects_invalid_word_counts() {
      assert!(new_petname(0, "-").is_err());
      assert!(new_petname(4, "-").is_err());
      assert!(new_petname(-1, "-").is_err());
      assert!(new_petname(100, "-").is_err());
  }
```

crates/petname/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    pub fn make_name(&self) -> String {
+        let mut rng = thread_rng();
+        let adjectives: Vec<&str> = ADJECTIVES.lines().collect();
+        let adverbs: Vec<&str> = ADVERBS.lines().collect();
+        let names: Vec<&str> = NAMES.lines().collect();
+
+        let adjective = adjectives.choose(&mut rng).unwrap();
+        let adverb = adverbs.choose(&mut rng).unwrap();
+        let name = names.choose(&mut rng).unwrap();
```
matthew-hagemann: These unwraps could also lead to panics, I'd have error handling for that returned a result here as well indicating what went wrong with which word list and return that in a result.

crates/petname/tests/petname_tests.rs
```diff
@@ -0,0 +1,69 @@
...
+    for _ in 0..(10 * expected_num_unique_names) {
+        let name = make_name(&mut name_generator);
+        name_set.insert(name);
+    }
+
+    assert!(name_set.len() >= expected_num_unique_names);
+}
+
+#[test]
+fn ffi_integration_test() {
```
matthew-hagemann: Not sure this is strictly an integration test? Might be nice to have a test on the cpp side that invoked some success and error cases via the FFI call to assert everything works as expected.

crates/petname/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+use rand::seq::SliceRandom;
+use rand::thread_rng;
+
+const ADJECTIVES: &str = include_str!("../../../src/petname/adjectives.txt");
+const ADVERBS: &str = include_str!("../../../src/petname/adverbs.txt");
+const NAMES: &str = include_str!("../../../src/petname/names.txt");
```
matthew-hagemann: If the petname's are now fully owned by the rust crate, I'd consider moving these into the crate so you are not reliant on relative pathing out of the crate file structure.

matthew-hagemann (changes requested):
This looks very exciting 🙂 

I've left some comments regarding error handling, specifically around how we can avoid using panic's for propagating errors to the cpp side. While I think the error cases here are fairly simple and the current approach works, getting in a foundation now that uses Rust's Result returns will set us up for more complex Rust libraries in the future.

matthew-hagemann:
For Rust CI, you can take a look at the [Rust code sanity](https://github.com/canonical/desktop-engineering/blob/main/gh-actions/rust/code-sanity/action.yaml) desktop action, and the [app-center-ratings ci](https://github.com/ubuntu/app-center-ratings/blob/main/.github/workflows/ci.yaml) to see the linter, formatter, testing and sec checks other projects are using.

src/daemon/daemon.cpp
```diff
@@ -217,12 +217,17 @@ auto name_from(const std::string& requested_name,
     }
     else
     {
-        auto name = name_gen.make_name();
+        // Create a Rust petname generator with 2 words and "-" separator
+        auto petname_generator = multipass::petname::new_petname(2, "-");
+        std::string name = std::string(multipass::petname::make_name(*petname_generator));
```
xmkg: Why wrap it with std::string here? I'd assume the return value of the make_name implements copy and move assignment operators, so string wrap shouldn't be necessary.

src/daemon/daemon.cpp
```diff
@@ -217,12 +217,17 @@ auto name_from(const std::string& requested_name,
...
         constexpr int num_retries = 100;
         for (int i = 0; i < num_retries; i++)
         {
             if (currently_used_names.find(name) != currently_used_names.end())
+            {
+                name = std::string(multipass::petname::make_name(*petname_generator));
```
xmkg: idem

src/petname/CMakeLists.txt
```diff
@@ -14,25 +14,54 @@
...
+  set(CARGO_PATH_ENV "PATH=$ENV{HOME}/.cargo/bin:$ENV{PATH}")
+endif()
 
-add_executable(text_to_string_array
-  text_to_string_array.cpp)
+# Set the target directory for this specific crate
+set(PETNAME_TARGET_DIR ${CMAKE_CURRENT_BINARY_DIR}/petname-target)
+set(PETNAME_CRATE_DIR ${CMAKE_SOURCE_DIR}/crates/petname)
 
 add_custom_command(
```
xmkg: Would something like https://github.com/corrosion-rs/corrosion simplify this?

levkropp: https://corrosion-rs.github.io/corrosion/ffi_bindings.html#cxx-integration Corrosion seems to have experimental support for cxx. Looking at it, I think it's fine if we adapt it at this stage because the "Limitations" section does not apply to us (we were always building as a staticlib anyways)

I've integrated it into the project using [FetchContent](https://corrosion-rs.github.io/corrosion/setup_corrosion.html#fetchcontent) because the docs appear to recommend that approach over a submodule

src/daemon/daemon.cpp
levkropp: `multipass::petname::make_name()` returns a `rust::cxxbridge1::String`, not a `std::string`, which needs to be explicitly converted. AFAICT this is a deliberate design choice of `cxx` to avoid hidden performance costs of string conversion

tests/test_daemon.cpp
```diff
@@ -690,13 +680,17 @@ TEST_P(DaemonCreateLaunchTestSuite, generatesNameOnCreationWhenClientDoesNotProv
 {
     const std::string expected_name{"pied-piper-valley"};
```
sharder996: This variable is no longer being used then.

tests/test_daemon.cpp
```diff
@@ -690,13 +680,17 @@ TEST_P(DaemonCreateLaunchTestSuite, generatesNameOnCreationWhenClientDoesNotProv
...
+    use_a_mock_vm_factory();
     mp::Daemon daemon{config_builder.build()};
 
     std::stringstream stream;
     send_command({GetParam()}, stream);
 
-    EXPECT_THAT(stream.str(), HasSubstr(expected_name));
+    // The stream should contain some generated name output (not the exact name since we can't
+    // control Rust petname generation)
+    EXPECT_THAT(stream.str(), Not(IsEmpty()));
```
sharder996: Maybe we could be a little more intelligent with matching, i.e., `\w+-\w+`

src/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ add_subdirectory(daemon)
 add_subdirectory(iso)
 add_subdirectory(logging)
 add_subdirectory(network)
-add_subdirectory(petname)
+add_subdirectory(rustipass)
```
sharder996: Maintain alphabetical ordering? 😄

snap/snapcraft.yaml
sharder996, line 1: Running unit tests?

.github/workflows/windows-macos.yml
sharder996, line 1: Unit tests?

sharder996 (changes requested):
Alright, I'll take the plunge and do my best here!

First off, I'd be curious to here about other developers setups for debugging code. There is `rust-lldb` and `rust-gdb` which are simple wrappers for their respective debuggers, but there are a few inconsistencies nonetheless.

Next, I think there are some improvements (or clarifications) that could be made about the build process. I'll list my findings below:

1. I don't know if its possible, but it would be nice if C++ and Rust unit tests could be integrated so that they can be executed from one command. Maybe a `cmake` command?

2. `ninja clean` doesn't seem to have any effect on Rust code. No rebuilding of Rust code happens after cleaning.

3. I noticed this warning while building. Are there any compatibility concerns with older versions of macOS?
```
ld: warning: object file (/Users/scott/Documents/dev/multipass/build/src/rustipass/librustipass.a[52](c16f17691ff6f04b-cxx.o)) was built for newer 'macOS' version (26.0) than being linked (15.6)
```

4. It seems like the Rust static library is continuing to be built with debug symbols:
```
cmake -Bbuild -H. -GNinja -DCMAKE_PREFIX_PATH=/Users/scott/Qt/6.2.4/macos/bin -DMULTIPASS_ENABLE_FLUTTER_GUI=OFF
...
[123/497] cd /Users/scott/Documents/dev/multipass...d/./cargo/build -- -Cdefault-linker-libraries=ye
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
```

5. I don't suppose Rust integrates with Codecov? Is there any sort of coverage data that can be generated?

Thanks for taking the initiative with integrating Rust! It's a big undertaking!

.github/workflows/windows-macos.yml
```diff
@@ -255,6 +256,29 @@ jobs:
...
+    - name: Initialize Rust toolchain
+      shell: bash
+      run: |
+        if [[ "${{ runner.os }}" == "Windows" ]]; then
+          # On Windows, rustup.install from chocolatey should be available
+          if command -v rustup >/dev/null 2>&1; then
+            rustup default stable
+            rustup update
+          else
+            echo "Warning: rustup not found, but continuing..."
```
xmkg: What's the reason that we're not raising a failure here? Considering that we were unable to bootstrap the Rust toolchain, wouldn't the Rust steps fail anyway?

tests/test_daemon.cpp
```diff
@@ -688,15 +679,19 @@ TEST_P(DaemonCreateLaunchTestSuite, onCreationHandlesInstanceImagePreparationFai
 
 TEST_P(DaemonCreateLaunchTestSuite, generatesNameOnCreationWhenClientDoesNotProvideOne)
 {
-    const std::string expected_name{"pied-piper-valley"};
-
-    config_builder.name_generator = std::make_unique<StubNameGenerator>(expected_name);
+    // Test now uses the actual Rust petname generation directly
```
xmkg: Hmm.. That reveals a problem in the current approach, which is that we cannot "mock" Rust modules. It's not a big deal for the name generation, but it will be for others since we can't unit test the stuff that depends on Rust modules anymore.

For the name generation, I think it would be best to hide the Rust module behind a mockable singleton, so we can still mock it.

xmkg: @ricab @jimporter I'd like to get your opinions on how to approach this ^^

ricab: I agree, very well spotted. We need to be able to mock things. If rust meant no mocking, it would not be feasible for unit tests.

From what I see, the daemon used to point to the name generator, but this makes it global. That's the problem. We should be able to call into rust to obtain *a* name generator to set in daemon_config, and we should be able to replace it. 

Thinking aloud here, 'cause I haven't looked into how the translation layer works, but can it expose a class? If so, are this class's methods virtual on the C++ side? If so, then we should be able to mock. 

I am not sure if we could have a pure interface, since there is no inheritance in rust. Not sure how the translation layer exposes traits. Could we declare and export a trait from rust, then derive from it in C++?

If none of that is possible, we'd need to wrap things in C++. That would be a pity (lots of boiler plate). It wouldn't have to be a `MockableSingleton` wrapper, but some wrapper. `MockableSingleton` could be an option, but in this case it seems to me that there wouldn't be a big benefit: we don't need to inject the dependency deeply through the stack. It's just the daemon, and the daemon config already wraps it, so we don't pollute functions with artificial parameters all over (we could have another couple of levels if we refactored to have a VM manager, but not a big deal). Whatever is easier and requires the least boiler plate, I guess?

snap/snapcraft.yaml
levkropp: We define `-DMULTIPASS_ENABLE_TESTS=off` in the snapcraft.yaml's `cmake-parameters`. I've added a command to run `cargo test` inside the snap but it doesn't seem like we have historically been interested in running tests during snapcraft

.github/workflows/dynamic-ci.yml
```diff
@@ -59,6 +59,9 @@ jobs:
               - "CMakeLists.txt"
               - "vcpkg-configuration.json"
               - "vcpkg.json"
+              - "Cargo.toml"
+              - "Cargo.lock"
+              - "build.rs"
```
Sploder12: Does this include the rust source?

.github/workflows/linux.yml
```diff
@@ -198,12 +198,13 @@ jobs:
...
-        # Enable coredumps by setting the core dump size to "unlimited", and run the tests.
+        # Enable coredumps by setting the core dump size to "unlimited", and run the unified test target.
         /snap/bin/lxc --project snapcraft exec $instance_name -- bash -c "\
-           ulimit -c unlimited &&        \
+           source /root/.cargo/env && \
+           ulimit -c unlimited && \
            env CTEST_OUTPUT_ON_FAILURE=1 \
            LD_LIBRARY_PATH=/root/stage/usr/lib/x86_64-linux-gnu/:/root/stage/lib/:/root/parts/multipass/build/lib/ \
-           /root/parts/multipass/build/bin/multipass_tests"
+           cmake --build /root/parts/multipass/build --target all_tests"
```
xmkg: Why not use something like `ctest --build-dir /root/parts/multipass/build --verbose` here? (assuming rust's test target is also defined via [add_test(...)](https://cmake.org/cmake/help/latest/command/add_test.html))

tests/CMakeLists.txt
```diff
@@ -240,3 +239,29 @@ endforeach()
...
+# Platform-specific cargo path handling
+if(WIN32)
+  # Windows: cargo should be in PATH, but ensure we use CMD to avoid bash conflicts
+  set(RUST_TEST_COMMAND cmd /c "cargo test --manifest-path ${CMAKE_SOURCE_DIR}/Cargo.toml")
+else()
+  # Unix-like: cargo should be available in PATH
+  set(RUST_TEST_COMMAND cargo test --manifest-path ${CMAKE_SOURCE_DIR}/Cargo.toml)
+endif()
+
+# Custom command to run both test suites with proper error handling
```
xmkg: How about using CTest for this?

snap/snapcraft.yaml
```diff
@@ -221,6 +221,17 @@ parts:
...
+      # Install Rust stable toolchain
+      if [ ! -f "$HOME/.cargo/env" ]; then
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+        source $HOME/.cargo/env
+      fi
+      # Ensure Rust is in PATH for this build
+      export PATH="$HOME/.cargo/bin:$PATH"
+
+      # Run the rust tests
+      (cd ${CRAFT_PART_SRC}/crates && cargo test --all)
```
xmkg: We're already running the tests in a dedicated CI step, right? So this being here would mean running the Rust tests twice.

.github/workflows/windows-macos.yml
```diff
@@ -276,9 +302,8 @@ jobs:
       run: ccache --show-stats --zero-stats
 
     - name: Test
-      working-directory: ${{ env.BUILD_DIR }}
       run: |
-        ${{ env.ARCH_WRAPPER }} bin/multipass_tests
+        ${{ env.ARCH_WRAPPER }} cmake --build ${{ env.BUILD_DIR }} --target all_tests
```
xmkg: idem (ctest)

xmkg (changes requested):

crates/rustipass/src/petname/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+fn uses_custom_separator() {
+    let mut name_generator = new_petname(3, "_").expect("Should create valid petname");
+    let name = make_name(&mut name_generator).expect("Should generate name");
+    let tokens = split_name(&name, "_");
+    assert_eq!(tokens.len(), 3);
+    assert!(name.contains("_"));
+}
+
+#[test]
+fn generates_two_tokens_by_default() {
```
Sploder12: the name of this test feels a bit off, 2 is being explicitly set

crates/rustipass/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    cxx_build::bridge("src/lib.rs")
+        .flag_if_supported("-std=c++17")
```
Sploder12: This will need to be changed soon

src/rustipass/rust_petname_generator.cpp
```diff
@@ -0,0 +1,52 @@
...
+                                       e.what());
+          }
+      }())
+{
+}
+
+std::string RustPetnameGenerator::make_name()
+{
+    try
+    {
```
Sploder12: nitpick: could use a function try block here

src/rustipass/rust_petname_generator.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace multipass
+{
+
+RustPetnameGenerator::RustPetnameGenerator(int num_words, const std::string& separator)
+    : petname_generator([&]() {
+          try
+          {
+              return multipass::petname::new_petname(num_words, separator.c_str());
+          }
```
Sploder12: nitpick: function try block would remove the need for a lambda

tests/mock_name_generator.h
```diff
@@ -13,16 +13,24 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
- *
  */
 
-#include "petname.h"
+#ifndef MULTIPASS_MOCK_NAME_GENERATOR_H
+#define MULTIPASS_MOCK_NAME_GENERATOR_H
```
Sploder12: should `#pragma once`

tests/test_daemon.cpp
```diff
@@ -69,6 +69,7 @@
 
 #include <memory>
 #include <ostream>
+#include <regex>
```
Sploder12: unused?

crates/rustipass/src/petname/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut gen2 = new_petname(2, "-").expect("Should create valid petname");
+    let mut gen3 = new_petname(3, "-").expect("Should create valid petname");
+
+    let one_word_name = make_name(&mut gen1).expect("Should generate name");
+    let tokens = split_name(&one_word_name, "-");
+    assert_eq!(tokens.len(), 1);
+
+    let two_word_name = make_name(&mut gen2).expect("Should generate name");
+    let tokens = split_name(&two_word_name, "-");
+    assert_eq!(tokens.len(), 2);
```
Sploder12, line 21: it might be worth checking the tokens aren't empty  since "-foobar" `["", "foobar"]` or "foobar-" `["foobar", ""]` would pass this test

Sploder12 (commented):
Looking good! Just some nitpicks and question on dynamic-ci

crates/rustipass/build.rs
levkropp: Switching to `.flag_if_supported("-std=c++20")` doesn't seem to cause any problems, and since we will be switching to C++ 20 very shortly after this PR is merged (perhaps even before this PR is merged), it's probably okay to go forward with `.flag_if_supported("-std=c++20")`  immediately

ricab:
I believe we can close this now.

PR #4306, opened 2025-08-14
[.github/dependabot.yml] enable dependabot for vcpkg

xmkg:
https://github.blog/changelog/2025-08-12-dependabot-version-updates-now-support-vcpkg/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for vcpkg
+  - package-ecosystem: "vcpkg"
+    # Look for `vcpkg.json` file in the root directory
+    directory: "/"
+    # Check the vcpkg registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,10 @@ updates:
     # Check the vcpkg registry for updates every day (weekdays)
     schedule:
       interval: "daily"
+  # Enable version updates for selected git submodules
+  - package-ecosystem: "gitsubmodule"
+    directory: "/"
+    allow:
+      - dependency-name: "3rd-party/vcpkg"
+    schedule:
+      interval: "daily"


sharder996:
Can dependabot and renovatebot be consolidated into one?

xmkg:
> Can dependabot and renovatebot be consolidated into one?

Unfortunately, renovatebot [does not support vcpkg](https://docs.renovatebot.com/modules/manager/#supported-managers). So, that leaves us with either migrating the action updates to dependabot or maintaining the hybrid approach until renovatebot begins supporting vcpkg. Using conan is still an option, so I guess the final decision depends on the outcome of the "streamlining dependencies" task that I have in my backlog.

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    # Look for `vcpkg.json` file in the root directory
+    directory: "/"
+    # Check the vcpkg registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"
+  # Enable version updates for selected git submodules
+  - package-ecosystem: "gitsubmodule"
+    directory: "/"
+    allow:
+      - dependency-name: "3rd-party/vcpkg"
```
ricab, line 14: This would update on every new commit, right? Is that what we want? And is that why you selected only vcpkg?

ricab: For reference: https://github.com/dependabot/dependabot-core/issues/1639

For repos that have something like a "stable" branch, it might be possible to use [`target-branch`](https://www.gaige.net/renovating-git-tags.html).

I looked into renovate and it doesn't appear to support this yet either, except via [custom regex-based managers](https://www.gaige.net/renovating-git-tags.html).

ricab (approved):
LGTM, thanks for this @xmkg! Only submodules left to deal with after this :smiley:

.github/dependabot.yml
xmkg: You're right! I haven't noticed this -- thanks for pointing that out. So it seems like we can't use the dependabot vcpkg job directly, because we prefer tagged vcpkg releases, and the vcpkg version and the baseline should ideally go hand-in-hand.

So the ideal approach would be to implement a custom regex-based manager in Renovate and implement an additional workflow for the baseline bump. I'll look into implementing a reusable custom manager for submodules so we can use it for the other submodules too (e.g., Flutter).

xmkg:
Scrapping this PR in favor of the Renovate custom regex-based manager (to be implemented in the future).

PR #4311, opened 2025-08-18
[macos] remove unused gRPC socket definition

### LINKED ISSUE #4287: Documentation implies multipassd does not listen on a TCP socket on MacOS
**Description of documentation issue or suggestion**
On this page: https://documentation.ubuntu.com/multipass/en/latest/explanation/authentication/ it is implied that multipassd uses a unix domain socket for communication on MacOS and Linux. That may well be so, but it is also listening on TCP/IP port 50051 on all interfaces and anyone looking to secure their system should be aware of this.

(checked on 1.16.0 on MacOS)

**Proposal to fix issue or enact suggestion**
The documentation should make it clear that multipassd is also listening on a TCP/IP socket (specifically 50051) and to what degree that is authenticated.

It would also be helpful, generally to make it clear that if a passphrase is not set, users cannot authenticate.

levkropp:
resolves #4287 

Old versions of Multipass on MacOS had launchd listen to a hardcoded tcp socket opened at port 50051. This was updated in https://github.com/canonical/multipass/commit/0f77459b9aa3ebba13100bd63898b8b25aede952 but the definition in the plist file was never updated.  This PR removes the unused TCP socket from being created on port 50051

### DIFF
--- a/packaging/macos/com.canonical.multipassd.plist.in
+++ b/packaging/macos/com.canonical.multipassd.plist.in
@@ -21,17 +21,6 @@
     <key>KeepAlive</key>
     <true/>
 
-    <key>Sockets</key>
-    <dict>
-       <key>Listeners</key>
-       <dict>
-            <key>SockServiceName</key>
-            <string>50051</string>
-            <key>SockType</key>
-            <string>stream</string>
-        </dict>
-    </dict>
-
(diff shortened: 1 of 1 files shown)


Sploder12 (approved):
LGTM, confirmed port 50051 is not being used anymore!

PR #4316, opened 2025-08-21
[az] remove round robin from GUI and CLI

levkropp:
This pull request removes the "round robin" functionality for assigning newly created instances to availability zones in the GUI and CLI. Instead of remembering the zone that was last used to launch a vm and then rotating from there, we now always try to place newly created instances without a specified zone parameter in zone1, failing over to zone2 if zone1 is unavailable, and failing over to zone3 if zones 1 and 2 are not available.
#### GUI
The "Auto (automatically selected)" option in the configure instance page (launch form) is no longer available

image

When "quick launching' an instance from the Catalogue page, the created instance will always be placed in zone1. If zone1 is not available, the instance will launch in zone2, etc.

#### CLI

The `multipass launch` command without a specified `--zone` argument will always launch an instance in zone1. If zone1 is not available, the instance will launch in zone2, etc.

MULTI-2205

### DIFF
--- a/src/platform/backends/shared/base_availability_zone_manager.cpp
+++ b/src/platform/backends/shared/base_availability_zone_manager.cpp
@@ -159,14 +159,16 @@ std::string BaseAvailabilityZoneManager::ZoneCollection::next_available()
 {
     std::unique_lock lock{mutex};
 
-    // Iterate through all zones starting with zone1 to find an available one
-    for (auto zone_it = zones.begin(); zone_it != zones.end(); ++zone_it)
+    // Locate the first available zone
+    auto zone_it = std::find_if(zones.begin(), zones.end(), [](const auto& zone) {
+        return zone->is_available();
+    });
+
+    // Check if an available zone was found
+    if (zone_it != zones.end())
     {
-        if ((*zone_it)->is_available())
-        {
-            automatic_zone = zone_it;
-            return (*zone_it)->get_name();
-        }
+        automatic_zone = zone_it;
+        return (*zone_it)->get_name();
     }
 
     // If none are available, throw an exception


src/client/cli/cmd/launch.cpp
```diff
@@ -487,6 +487,11 @@ mp::ReturnCode cmd::Launch::request_launch(const ArgParser* parser)
         }
         request.set_zone(zone.toStdString());
     }
+    else
+    {
+        // Use default zone1 if no explicit zone provided
+        request.set_zone("zone1");
+    }
```
Sploder12: this would be handled by the backend when passing an empty zone name, no?

Sploder12 (commented):
Looks pretty good, just a question on the CLI. We'll have to remove the round-robin implementation on the daemon as well.

src/platform/backends/shared/base_availability_zone_manager.cpp
```diff
@@ -82,18 +58,72 @@ BaseAvailabilityZoneManager::data BaseAvailabilityZoneManager::read_from_file(
     const std::filesystem::path& zones_directory)
 {
     mpl::debug(category, "reading AZ manager from file '{}'", file_path);
+
+    // Try to read the manager file
+    QJsonObject manager_data;
```
Sploder12: bit out of place for this PR but oh well

src/platform/backends/shared/base_availability_zone_manager.cpp
```diff
@@ -82,18 +58,72 @@ BaseAvailabilityZoneManager::data BaseAvailabilityZoneManager::read_from_file(
...
+            zone_data["type"] = "default";
+
+            try
+            {
+                MP_JSONUTILS.write_json(zone_data, QString::fromUtf8(zone_file.u8string()));
+            }
+            catch (const std::exception& e)
+            {
+                mpl::debug(category, "failed to write zone file '{}': {}", zone_file, e.what());
+            }
```
Sploder12: this should probably be a separate function

src/platform/backends/shared/base_availability_zone_manager.cpp
```diff
@@ -82,18 +58,72 @@ BaseAvailabilityZoneManager::data BaseAvailabilityZoneManager::read_from_file(
...
+    manager_data["default_zone"] = QString::fromStdString(m.zone_collection.zones[0]->get_name());
+
+    try
+    {
+        MP_JSONUTILS.write_json(manager_data, QString::fromUtf8((data_dir / az_file).u8string()));
+    }
+    catch (const std::exception& e)
+    {
+        mpl::debug(category, "failed to write manager file '{}': {}", az_file, e.what());
+    }
```
Sploder12: Do we really need this?

src/platform/backends/shared/base_availability_zone_manager.cpp
```diff
@@ -158,21 +158,18 @@ BaseAvailabilityZoneManager::ZoneCollection::ZoneCollection(
...
+    {
+        if ((*zone_it)->is_available())
         {
-            automatic_zone = current;
-            return az.get_name();
+            automatic_zone = zone_it;
+            return (*zone_it)->get_name();
         }
-    } while (current != start);
+    }
```
Sploder12: this looks like the perfect place for a [std::find_if](https://en.cppreference.com/w/cpp/algorithm/find.html)

Sploder12 (changes requested):
Looking pretty good, but there is a regression in behavior. Selecting an unavailable zone in the dropdown will force the selection back to the default zone, before you could select the unavailable zone and enable it.

sharder996 (approved):

PR #4318, opened 2025-08-22
[macos] Increase the timeout for launchd

### LINKED ISSUE #4319: [macOS] Multipass daemon gets killed before VMs have a chance to suspend
**Describe the bug**
When the daemon needs to be stopped (e.g. to upgrade), launchd will send SIGTERM, then wait for 5s, and then send SIGKILL. When running QEMU VMs take longer than that to suspend, they get orphaned but linger around. When the daemon is started anew, it restarts the previously running VM. At that point there are two processes for the same VM and problems cascade (e.g. due to shared files, MAC addresses, mounts).

**To Reproduce**
How, and what happened?
1. `multipass launch`
2. wait for the VM to get to running state
3. `pgrep qemu` - note the process ID
4. upgrade to a newer multipass version
5. `pgrep QEMU` - verify that the old process ID is still around (along with a new one for the same VM)

**Expected behavior**
Launchd waits for longer than 5 seconds, so that VMs have a good chance to suspend.

**Additional info**
- OS: macOS 15.5
- `multipass version`: observed as far back as 1.14.1, but appears more likely in the 1.16 track

ricab:
Increase the timeout after which launchd tires to wait for the daemon to
quit on SIGTERM and sends it a SIGKILL. The daemon can take a while to
quit before suspending VMs finish writing to disk. Fixes #4319.

### DIFF
--- a/packaging/macos/com.canonical.multipassd.plist.in
+++ b/packaging/macos/com.canonical.multipassd.plist.in
@@ -25,6 +25,9 @@
     <string>/Library/Logs/Multipass/multipassd.log</string>
     <key>StandardErrorPath</key>
     <string>/Library/Logs/Multipass/multipassd.log</string>
+
+    <key>ExitTimeOut</key>
+    <integer>45</integer>
 </dict>
 
 </plist>


ricab:
Just tested the package and it behaves as expected:

```
$ log show --last 3m | grep multipassd.*cleanup
2025-08-26 19:22:29.733676+0100 0x30a62    Default     0x0                  1      0    launchd: [system/com.canonical.multipassd [5872]:] scheduling cleanup in 45 sec after sending Terminated: 15
```

It used to say "5 sec".

sharder996 (approved):
+1

PR #4327, opened 2025-08-27
Release/1.16

ricab:
We need to do this via a PR now, to get the CLA check.


ricab:
Remember to restore the branch on merge! It's necessary for version derivation in the candidate channel.

PR #4332, opened 2025-09-03
[commit-msg-hook] Strip comments before processing

xmkg:
The commit message linter enforces the rules for Git comments in the COMMIT_MSG, which causes trouble when a long path appears in the change summary.

The patch addresses this issue by removing all comments from the input before processing.

MULTI-2221


ricab (approved):
Right, thanks! Tested manually and it works properly.

PR #4333, opened 2025-09-03
[git] Strip trailing whitespace in msg validation

ricab:


ricab:
Note, git does not strip it from the message that the hook verifies, but it does strip it from the final message that gets recorded for the commit, by default.

xmkg (approved):

PR #4334, opened 2025-09-03
Fix MacOS CI failing to install cmake

Sploder12:
Fixes the failure in MacOS CI as described by https://github.com/actions/runner-images/issues/12912
The fix checks if cmake is already installed before installing via brew.


.github/workflows/windows-macos.yml
```diff
@@ -147,9 +147,12 @@ jobs:
     - name: Install dependencies from brew
       if: ${{ runner.os == 'macOS' }}
       run: |
+        # Avoid reinstalling cmake
+        ${{ env.ARCH_WRAPPER }} \
+        brew list cmake > /dev/null || ${{ env.ARCH_WRAPPER }} brew install cmake
```
ricab: This line split is a little difficult to spot. I would prefer splitting after the or and indent the second line:

```suggestion
        ${{ env.ARCH_WRAPPER }} brew list cmake > /dev/null || \
          ${{ env.ARCH_WRAPPER }} brew install cmake
```

WDYT?

ricab (commented):
Thanks @Sploder12. Looks good overall, but I have a comment inline. Also shouldn't we `brew update` first?

Sploder12:
> Also shouldn't we `brew update` first?

We didn't before but I don't see why not

levkropp:
changes LGTM, but note that a [fix is in progress](https://github.com/actions/runner-images/issues/12912#issuecomment-3242862309) so this change may be obsoleted soon? Also seeing consistent segfaults on MacOS 15's CI, not sure if that is related to these changes or not

Sploder12:
> changes LGTM, but note that a [fix is in progress](https://github.com/actions/runner-images/issues/12912#issuecomment-3242862309) so this change may be obsoleted soon? Also seeing consistent segfaults on MacOS 15's CI, not sure if that is related to these changes or not

Yes, this would be obsoleted soon, but everything code related is blocked on this until then. The segfaults are certainly interesting... my guess is we have some terrible issue in a test and a slight change in tool versions caused it to surface, assuming it's at all related to this

.github/workflows/windows-macos.yml
```diff
@@ -147,9 +147,15 @@ jobs:
     - name: Install dependencies from brew
       if: ${{ runner.os == 'macOS' }}
       run: |
+        ${{ env.ARCH_WRAPPER }} \
+        brew update
+
```
ricab: Hey @Sploder12, sorry I think I misled you here. I really meant it only as a question. After searching a little, it's probably better not to update, for a few reasons: 

- GitHub images should have a recent and exercised cache
- Less risk of bleeding edge stuff breaking our CI and more predictable builds
- Updates take some time (perhaps 1 minute)

Could we drop that commit after all, please?

ricab: Ah and: https://github.com/canonical/multipass/blob/e9e36da6435902793ab615ffd17bf73a39d2afe0/.github/workflows/windows-macos.yml#L20

ricab:
Bah this `Client.transferCmdInstanceSourceLocalTarget` test is now crashing consistently. @xmkg is this one of those you mentioned? Do you have a fix?

xmkg:
> Bah this `Client.transferCmdInstanceSourceLocalTarget` test is now crashing consistently. @xmkg is this one of those you mentioned? Do you have a fix?

Unfortunately, it's not one of those. I tried running the "Client" tests with a UBsan build but it worked fine. I'll try a different angle.

xmkg:
> > Bah this `Client.transferCmdInstanceSourceLocalTarget` test is now crashing consistently. @xmkg is this one of those you mentioned? Do you have a fix?
> 
> Unfortunately, it's not one of those. I tried running the "Client" tests with a UBsan build but it worked fine. I'll try a different angle.

Provided a workaround via https://github.com/canonical/multipass/pull/4350. The real fix will require a gRPC version bump (or a vcpkg baseline upgrade).

PR #4335, opened 2025-09-03
[git] Support autosquash keywords in commit msg

jimporter:


git-hooks/commit-msg.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, msg):
         self.subject = self.lines[0].rstrip() if self.lines else ""
         self.body = self.lines[1:]
 
+        # Strip special keywords used by `git rebase --autosquash`
+        self.subject = re.sub(r"^(fixup|squash)! ", "", self.subject)
```
xmkg: Hi @jimporter, thanks for the contribution! The change looks good to me -- I just have a small request. Would you mind adding a test case for it?

jimporter:
> Your PR uncovers a distinction in the two ways the script is used today: locally and in CI. I think your changes make sense when verifying local commits, but we should probably continue to refuse "fixup" messages from merging in PRs. A simple approach to achieve that would be to add a flag (e.g. `--strict`) and pass it in CI. WDYT?

That makes sense to me. One possible future enhancement would be to add another local Git hook for `pre-push` that would perform strict validation of all the commit messages to be pushed. That way, we could catch any autostash keywords that snuck into the branch even before we run Github's CI.

> A note on workflows and guidelines. I know there is debate on few big commits vs many atomic commits. It is a tradeoff. Big commits avoid repeated conflicts when rebasing and make the graph look simpler. Atomic commits encourage splitting tasks and more methodical planning.

I fully agree. I always try to author atomic commits, but because of that, I'm also a heavy user of `git rebase`. That, way I can iterate on a series of commits and fix problems in earlier WIP commits without extra fix-it commits in the history that would probably be uninteresting to someone looking through the commit logs in future years.

git-hooks/commit-msg.py
jimporter: Thanks for the catch. I didn't notice the tests farther down in the file. I've now added a test case.

ricab (approved):
Alright, this looks great, thanks @jimporter!

(secondary review)

xmkg (approved):

PR #4336, opened 2025-09-04
[lxd-image-vault] Fix chrono time overflow

xmkg:
The milliseconds value obtained from the QDateTime can go high negative and that causes a silent chrono integer overflow bug where a ratio conversion silently overflows the destination:

```
runtime error: signed integer overflow: -62135596800000 * 1000000
cannot be represented in type 'long'
```
This patch fixes that by adding safety nets around the time calcuation. Added an unit test for the case where "last_used_at" is absent.


src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
+    const auto images = retrieve_image_list();
 
     for (const auto image : images)
     {
         auto image_info = image.toObject();
         auto properties = image_info["properties"].toObject();
 
-        auto last_used = std::chrono::system_clock::time_point(std::chrono::milliseconds(
-            QDateTime::fromString(image_info["last_used_at"].toString(), Qt::ISODateWithMs)
-                .toMSecsSinceEpoch()));
```
jimporter, line 410: Looking at the test, was this old code buggy because `"last_used_at"` was sometimes missing from both `image_info` and `properties`? In that case, I think `image_info["last_used_at"].toString()` would return a null string (likewise for `properties`), and then `QDateTime::fromString` constructs a junk date value. Am I getting this right?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
+                QDateTime::fromString(json["last_used_at"].toString(), Qt::ISODateWithMs);
+
+            // This can happen if the date is before 01-01-1970
+            if (qtime.isValid() && qtime.toMSecsSinceEpoch() >= 0)
+            {
+                const std::chrono::milliseconds msecs_since_epoch{qtime.toMSecsSinceEpoch()};
+                last_used = std::chrono::system_clock::time_point{msecs_since_epoch};
+                break;
+            }
+        }
```
jimporter: What about moving this to a small utility function? Then we could write a few unit tests for our new JSON-to-`time_point` conversion to make sure it does what we expect in all cases. Something along the lines of this function:

```c++
std::optional<std::chrono::system_clock::time_point>
json_to_time_point(const QJsonValue &value)
{
    using namespace std::chrono;
    if (value.isString())
    {
        const auto qtime = QDateTime::fromString(value.toString(), Qt::ISODateWithMs);
        if(qtime.isValid())
            return system_clock::time_point(milliseconds(qtime.toMSecsSinceEpoch()));
    }
    return std::nullopt;
}
```

Then you could call it something like:

```c++
auto last_used = json_to_time_point(image_info["last_used_at"]);
if (!last_used)
    last_used = json_to_time_point(properties["last_used_at"]);
```

tests/lxd/mock_lxd_server_responses.h
```diff
@@ -1148,6 +1148,47 @@ const QByteArray image_info_unused_updated_image{
...
+    "    \"cached\": false,"
+    "    \"filename\": \"ubuntu-18.04-server-cloudimg-amd64-lxd.tar.xz\","
+    "    \"fingerprint\": \"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\","
+    "    \"size\": 345572108,"
+    "    \"type\": \"virtual-machine\","
+    "    \"created_at\": \"2020-05-18T20:00:00-04:00\","
+    "    \"uploaded_at\": \"2020-06-04T14:25:41.622958476-04:00\""
+    "    }"
+    "  ]"
+    "}"};
```
jimporter, line 1190: Would a raw string literal be easier to read here? Something like:

```c++
const QByteArray image_info_without_last_used_at{R"json(
  {
    "type": "sync",
    "status": "Success",
    ...
  }
)json"};

File: src/platform/backends/lxd/lxd_vm_image_vault.cpp (hunk 1 above)
Comment (xmkg) [reply src/platform/backends/lxd/lxd_vm_image_vault.cpp:410]: No, the problem is that the test data contains a bogus date value (`01-01-0001`), and that causes an integer overflow in the std::chrono code. The chrono code tries to convert the milliseconds to nanoseconds at some point.  The date can be `0001-01-01` at minimum, which means `-62135596800000ms * 1'000'000 = -62,135,596,800,000,000,000 ns (-6.2e19)`, and `-6.2e19` cannot fit into `int64_t` (`-9.22e18` < -`6.2e19`), hence the problem.

**Comment by xmkg** [comment]
Hi @jimporter, thanks for the thorough review!

I agree with all of your comments, but this backend code is deprecated and will be removed quite soon. The only reason I've submitted a patch is to make `ubsan` happy in sanitizer builds, so there would be less noise. For that reason, I believe it's not worth doing any more refactoring. What do you think?

File: src/platform/backends/lxd/lxd_vm_image_vault.cpp [hunk 4]
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
+            if (!json.contains("last_used_at"))
+                continue;
+
+            const auto qtime =
+                QDateTime::fromString(json["last_used_at"].toString(), Qt::ISODateWithMs);
+
+            // This can happen if the date is before 01-01-1970
+            if (qtime.isValid() && qtime.toMSecsSinceEpoch() >= 0)
+            {
+                const std::chrono::milliseconds msecs_since_epoch{qtime.toMSecsSinceEpoch()};
```
tobe2098: Since you are calling qtime.toMSecsSinceEpoch() twice in the if statement block, it could be better to do the declaration in the if statement itself:
```
if(const std::chrono::milliseconds msecs_since_epoch{qtime.toMSecsSinceEpoch()}; qtime.isValid()&&msecs_since_epoch>=0){
last_used=....;
```
What do you think?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
-        // property during update
-        if (last_used < std::chrono::system_clock::time_point(0ms) &&
-            properties.contains("last_used_at"))
+        if (!last_used.has_value())
         {
-            last_used = std::chrono::system_clock::time_point(std::chrono::milliseconds(
-                QDateTime::fromString(properties["last_used_at"].toString(), Qt::ISODateWithMs)
-                    .toMSecsSinceEpoch()));
+            mpl::warn(category,
+                      "Source image `{}` does not have `last_used_at` property, skipping it!");
```
tobe2098, line 432: More of a question than anything else, should we skip images that do not have the `last_used_at` property? Is it possible to download an image without using it and have it not have that property inside multipass?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
tobe2098: I would definitely agree if this check had to be performed in other parts of the code, which is probably true. 

If not, it would be a lot of extra code (accounting for tests) for something that could already be tested within the current function's tests via the raw JSON you suggested, no? What do you think?

src/platform/backends/lxd/lxd_vm_image_vault.cpp
xmkg: The reason we're calling the `toMsecsSinceEpoch()` twice here is intentional. The ~~std::chrono::milliseconds~~ `std::chrono::system_clock::time_point` constructor tries to convert the incoming value to `nanoseconds` for normalization, and if the `toMSecsSinceEpoch() * 1'000'000` produces a value that can't fit into a `int64_t`, an overflow happens. So, we have to guard against the values that can overflow. That has been the case for one of our unit tests, which used `01-01-0001` as a date, which is too big of a negative nanoseconds value to fit into an `int64_t`. 

The code ensures that we have a positive milliseconds value first. The bug can still happen if the date value is a positive value that can overflow. I suppose we could use `std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds>` to prevent the nanoseconds normalization altogether, but it would require a few tweaks down the line since the system_clock::now() is `implementation-defined clock based` (usually nanoseconds, or 100 nanoseconds). 

I think this is enough bikeshedding for a code that's going to be removed soon :)

src/platform/backends/lxd/lxd_vm_image_vault.cpp
tobe2098: I have looked at the documentation of Qt and std::chrono, and so far the only reason for an overflow that I could find is that the specifications on cplusplus of std::chrono::milliseconds is that the representation be a signed integer type of at least 45 bits. If for some reason it is not an int64_t and uses less bits, it could happen without nanoseconds. 
I have found nothing that could end up with a conversion to nano or microseconds in that snippet.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
tobe2098: I agree :)

src/platform/backends/lxd/lxd_vm_image_vault.cpp
tobe2098: Looking at the GLIBCXX implementation I also do not see any conversion/normalization to nanoseconds, if given a qint64 from .toMSecSinceEpoch() it should construct milliseconds directly. Unless the implementation is truly a under-64-bit representation, I'm at a loss as to why that is happening.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
xmkg: The expiration logic depends on the entity having a valid timestamp. If it doesn't, we can't determine whether it's expired because we lack that essential information. This assumes that the absence of `last_used_at` doesn't have meaningful significance as "the image has never been used." Even then, a `cached but never used` image could either mean it was just downloaded or has sat there indefinitely -- so we still can't decide if we should remove it or not. So, the code simply refuses to take any action, and notifies the user via a warning log.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
xmkg: The line that's triggering the bug is this:

```c++
        auto last_used = std::chrono::system_clock::time_point(std::chrono::milliseconds(
            QDateTime::fromString(image_info["last_used_at"].toString(), Qt::ISODateWithMs)
                .toMSecsSinceEpoch()));
```

Where the code is converted to milliseconds, then std::chrono::system_clock::time_point, where the `time_point` is in `nanoseconds` (at least in the libc++). 

The flow is as follows:

- std::chrono::milliseconds constructor (no duration_cast happens)
- std::chrono::system_clock::time_point constructor (duration_cast, from milliseconds -> nanoseconds)

```
--- Intentional SIGILL triggered by ubsan signaling an integer overflow ---
std::chrono::__duration_cast_impl<std::chrono::duration<long, std::ratio<1l, 1000000000l> >, std::ratio<1000000l, 1l>, long, false, true>::__cast<long, std::ratio<1l, 1000l> >(const std::chrono::duration<long, std::ratio<1l, 1000l> > & __d) (/usr/include/c++/13/bits/chrono.h:225)
std::chrono::duration_cast<std::chrono::duration<long, std::ratio<1l, 1000000000l> >, long, std::ratio<1l, 1000l> >(const std::chrono::duration<long, std::ratio<1l, 1000l> > & __d) (/usr/include/c++/13/bits/chrono.h:287)
std::chrono::duration<long, std::ratio<1l, 1000000000l> >::duration<long, std::ratio<1l, 1000l>, void>(std::chrono::duration<long, std::ratio<1l, 1000000000l> > * this, const std::chrono::duration<long, std::ratio<1l, 1000l> > & __d) (/usr/include/c++/13/bits/chrono.h:582)
multipass::LXDVMImageVault::prune_expired_images(multipass::LXDVMImageVault * this) (/workspace/multipass/src/platform/backends/lxd/lxd_vm_image_vault.cpp:408)
```

image

xmkg: I've realized that I said `std::chrono::milliseconds` in the previous discussion, which probably led to this confusion, sorry about that :)

xmkg: I appreciate your due diligence and attention to detail, @tobe2098!

jimporter: Got it, that makes sense. Then I think we might want to be careful about overflow in the opposite direction too, so as to avoid the [year 2262 problem](https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs#Year_2262).

src/platform/backends/lxd/lxd_vm_image_vault.cpp
jimporter: Even without separate unit tests, one thing I like about having a utility function is that the date retrieval code in `prune_expired_images` is just 3 lines (my second block above). That makes it easier for me to follow what's happening.

Having some unit tests for that function does have the benefit that it would let us feed several dates to it to make sure we catch all the possible overflow conditions though.

That said, since this code is going away soon, I don't mind doing whatever fixes the bug with the least amount of additional code.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
-        auto last_used = std::chrono::system_clock::time_point(std::chrono::milliseconds(
-            QDateTime::fromString(image_info["last_used_at"].toString(), Qt::ISODateWithMs)
-                .toMSecsSinceEpoch()));
+        std::optional<std::chrono::system_clock::time_point> last_used{std::nullopt};
+
+        std::array<std::reference_wrapper<const QJsonObject>, 2> qjson_targets{image_info,
+                                                                               properties};
+
+        for (const auto& target : qjson_targets)
+        {
```
jimporter: If we stick with this for loop, I think we could simplify it like so:

```suggestion
        for (const auto& target : {std::cref(image_info), std::cref(properties)})
        {
```

src/platform/backends/lxd/lxd_vm_image_vault.cpp
```diff
@@ -398,28 +398,44 @@ bool mp::LXDVMImageVault::has_record_for(const std::string& name)
...
+            const QJsonObject& json = target.get();
+            if (!json.contains("last_used_at"))
+                continue;
+
+            const auto qtime =
+                QDateTime::fromString(json["last_used_at"].toString(), Qt::ISODateWithMs);
+
+            // This can happen if the date is before 01-01-1970
+            if (qtime.isValid() && qtime.toMSecsSinceEpoch() >= 0)
```
jimporter: The discussion in this PR was really useful for explaining what the bug is. Could you add a short summary to this comment about the milli -> nano conversion/overflow? Probably just a sentence or two would be plenty.

We could also check for overflow in either direction here. Since this code is going away sooner or later, maybe we could keep it simple, like:

```suggestion
            constexpr std::int64_t max_msecs = LLONG_MAX / 1'000'000;
            if (qtime.isValid() && std::abs(qtime.toMSecsSinceEpoch()) < max_msecs)
```

src/platform/backends/lxd/lxd_vm_image_vault.cpp
xmkg: That looks better to my eye, too. I'll update it.

src/platform/backends/lxd/lxd_vm_image_vault.cpp
xmkg: I've decided to use a milliseconds-based time_point type to avoid the problem altogether. It should be safe now since no conversion would happen (besides ns --> ms, which is fine).

PR #4337, opened 2025-09-04
[commit-msg-hook] Allow verbatim text with `>`

xmkg:
Sometimes we might need to copy output verbatim from some other tool or quote additional context. The commit message hook shouldn't enforce the rules to those parts.

MULTI-2224

### DIFF
--- a/git-hooks/commit-msg.py
+++ b/git-hooks/commit-msg.py
@@ -30,7 +30,7 @@ class CommitMsgRulesChecker:
         RULE6 =   "MSG6.  Do not end the subject line with a period."
         RULE8 =   "MSG8.  If adding a body, separate it from the subject with a blank line."
         RULE10 =  "MSG10. Do not include more than 1 consecutive blank line, except in quoted text."
-        RULE12 =  "MSG12. Wrap the body at 72 characters, except on quoted lines and references."
+        RULE12 =  "MSG12. Wrap the body at 72 characters, except on lines consisting only of blockquotes, references, or sign-offs."
         # fmt: on
 
     def __init__(self, msg):
@@ -57,7 +57,7 @@ class CommitMsgRulesChecker:
 
         self.errors = self.validate_all()
 
-    def should_ignore_line(self, line):
+    def is_inflexible_line(self, line):
         return bool(
             # Ignore quoted lines
             re.match("(>).*\n?", line)
@@ -103,7 +103,7 @@ class CommitMsgRulesChecker:
         return not any(both_blank(l1, l2) for l1, l2 in zip(self.body, self.body[1:]))
 
     def validate_rule12(self):
-        return all(len(line) <= 72 for line in self.body if not self.should_ignore_line(line))
+        return all(len(line) <= 72 for line in self.body if not self.is_inflexible_line(line))
 
 
 def validate(msg):
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -178,7 +178,7 @@ try to find a generic unifying category, or choose the most relevant.<br>
 **MSG9.** Use multiple paragraphs in the body if needed. Separate them with a blank line.<br>
 **MSG10.** Do not include more than 1 consecutive blank line, except in quoted text.<br>
 **MSG11.** Use punctuation normally in the body.<br>
-**MSG12.** Wrap the body at 72 characters, except on quoted lines and references.<br>
+**MSG12.** Wrap the body at 72 characters, except on lines consisting only of blockquotes, references, or sign-offs.<br>
 **MSG13.** Use the body to explain *what* and *why*, rather than *how*.<br>
 **MSG14.** Be descriptive but succinct and avoid filler text.<br>
 **MSG15.** Omit the body if the subject is self-explanatory.<br>


git-hooks/commit-msg.py
```diff
@@ -46,8 +46,11 @@ def __init__(self, msg):
 
         self.enough = False
 
+        self.msg = msg
         # Strip comments and trailing whitespace
-        self.msg = re.sub(r"^#.*\n?", "", msg, flags=re.MULTILINE)
+        self.msg = re.sub(r"^#.*\n?", "", self.msg, flags=re.MULTILINE)
+        # Strip verbatim text (rules should not apply)
+        self.msg = re.sub(r"^>.*\n?", "", self.msg, flags=re.MULTILINE)
```
jimporter: Could this run afoul of `MSG10`? ("Do not include more than 1 consecutive blank line")

For example, I think the following commit message would fail:

```
[tag] Do something useful

The standard says the following:

> 1.1: Lorem ipsum dolor sit amet, consectetur adipiscing elit...

> 1.3: Sed ut perspiciatis unde omnis iste natus error sit voluptatem...

Therefore, we need to etc, etc
```

xmkg: Good point, and well spotted! I wonder how we should approach this. Should the regex optionally swallow the empty new line before and after? @ricab opinions?

ricab: Oh, good catch! Hmm, tricky. I guess we should replace sequences of "line-quote-line" with just "line", but "quote" would need to match multi-line quotes. Something like `"\n?(>.*\n)+\n?"`? Although there will be a bunch of corner cases... For example:

```
[tag] Do something useful

The standard says the following:
> 1.1: Lorem ipsum dolor sit amet, consectetur adipiscing elit...


Therefore...
```

We want to catch the double lines there, so do we replace "quote-line" with just "line" too? But then Jim's example could end up with two blank lines again.

I wonder if we should just replace `>` with `#` and then find a way to invoke git's stripping functionality. Or keep the message and modify rules to account for the special cases.

BTW, we should also keep comments with the strict flag.

ricab: Perhaps we should let consecutive lines go when not strict. That would match what git does by default on commit. When verifying the final message (on push or CI), we would refuse any that remained.

jimporter: I can think of two possible ways to do it:

1. Replace `^>.*` with some text like `placeholder`. This is nice and simple, but might bite us later if we ever tried to print `self.msg` back out to the user.
2. Replace `^>.*\n{0,2}` with the empty string. That *should* work, since it'll consume the newline at the end of the `>` line, plus up to one additional newline (for a blank line afterwards). This logic is a bit trickier, but with some tests, we should be safe.

xmkg: @ricab @jimporter thanks for the ideas! I think we should stop removing content from the message since it would complicate things with rule interactions, as @jimporter pointed out. I've moved the logic to a function and stopped stripping the comments/verbatim text altogether in the latest revision, so the exception can be applied in the line-based rules on demand. Let me know how you feel about this approach.

git-hooks/commit-msg.py
```diff
@@ -46,8 +46,8 @@ def __init__(self, msg):
 
         self.enough = False
 
-        # Strip comments and trailing whitespace
-        self.msg = re.sub(r"^#.*\n?", "", msg, flags=re.MULTILINE)
```
jimporter, line 56: I'm of two minds about stripping comments here. Stripping them out here means that `self.msg` would hold the same value as the message Git will use for the commit, which means our individual rule implementations don't need to account for comments. For example, this commit message is ok in the end:

```
[tag] Some commit
# This is a comment

My description here.
```

I think that runs afoul of our rule 8 implementation now though. We could just match the [default Git commit message cleanup behavior exactly](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupmode):

> Strip leading and trailing empty lines, trailing whitespace, commentary and collapse consecutive empty lines.

That said, I don't have a strong feeling on this, especially since there's no reliable way that I can find that would authoritatively tell us what cleanup mode Git will *actually* use. If a user passes `--cleanup=scissors` for example, we have no way of telling. The best we could do is guess (ok, I guess we could do something like `ps -o args=` but that's a bit too complex for this).

xmkg: Good point. It would be helpful if we could obtain a preprocessed version of the message that reflects the exact flags the user has specified, but I couldn't find a clean way to do that either. 

The closest thing is [git stripspace](https://git-scm.com/docs/git-stripspace), and it does exactly what you've quoted. I'd say it's good enough for our use case if we don't be pedantic about it. WDYT?

jimporter: Interesting, I didn't know about `git stripspace`. It seems like it would work here (aside from the issues already mentioned). I also wouldn't mind if we just reimplement the comment/space-stripping rules in Python, since they're not that complicated.

git-hooks/commit-msg.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, msg):
 
         self.errors = self.validate_all()
 
+    def should_ignore_line(self, line):
+        return bool(re.match("(#|>).*\n?", line))
```
ricab: I think having quotes as an exception that is part of the individual rules is the right option. Comments, on the other hand, are not part of the rules as they are stated here. Nor should they be in my opinion. I see two main situations:

- when running in CI, comment stripping is supposed to have taken place, assuming `cleanup=strip` as indicated below. If they are still present at that point, I think the action should fail
- when running as a commit hook, comments won't have been stripped and should be skipped.

Jim introduced `--strict` in the meantime, which seems to fit this perfectly. So, how about matching only quotes here, but calling `git stripspace` at the top when not running in strict mode? That should rid us of comments and take care of newlines and anything else pending cleanup. WDYT?

git-hooks/commit-msg.py
```diff
@@ -28,8 +28,8 @@ class Rules(Enum):
         RULE5 =   "MSG5.  Limit the subject line to 50 characters (category included)."
         RULE6 =   "MSG6.  Do not end the subject line with a period."
         RULE8 =   "MSG8.  If adding a body, separate it from the subject with a blank line."
-        RULE10 =  "MSG10. Do not include more than 1 consecutive blank line."
-        RULE12 =  "MSG12. Wrap the body at 72 characters."
+        RULE10 =  "MSG10. Do not include more than 1 consecutive blank line, except in quoted text."
```
ricab, line 32: This would need changing in CONTRIBUTING.md too. I know, I know, duplicate sources of truth... I thought it would be convenient here 😅 I wish GH supported includes in markdown. But feel free to implement some parsing instead heheh 😜

git-hooks/commit-msg.py
```diff
@@ -28,8 +28,8 @@ class Rules(Enum):
...
+        RULE12 =  "MSG12. Wrap the body at 72 characters, except on quoted lines and URLs."
```
ricab: I am thinking we should minimize exceptions and use quotes for anything that needs to break the limit, including URLs. That would save us from having to implement more parsing and require the URL to be on its own line, clearly marked as something we did not directly author. WDYT?

git-hooks/commit-msg.py
ricab: TIL `git stripspace` too. Why don't you call that?

Note that, the way this is done already "Assumes that commits are made with `--cleanup=strip` (when used as a hook)." https://github.com/canonical/multipass/blob/07ac2a6e0b0f4ef8fd6ee11c8156e98619c1b1d5/git-hooks/commit-msg.py#L415

git-hooks/commit-msg.py
xmkg: Sounds good! I'll make the changes.

git-hooks/commit-msg.py
xmkg: I think it might look a bit weird and break the natural flow of a sentence, and also penalize the lines that already have no trouble fitting the line length limit with a URL. Moreover, all the commit authors will have to remember to do a line break and quote whenever they're typing a URL, which adds more cognitive burden.

```
blah blah something that is going to
> https://exceed.the.limit.with.url/at-the-end
```

```
see more at
> https://example.com
and
> https://example.com/b
and
> https://example.com/c
```

A middle-ground might be to use footnote references for the instances that would exceed the line length limit:

```
blah blah something that is going to [1].

[1]: https://exceed.the.limit.with.url/at-the-end
```
```
See more at [1], [2], and [3].

[1]: https://example.com
[2]: https://example.com/b
[3]: https://example.com/c
```

... and just allow in-line URLs where they do not break the existing rules:

```
An example can be seen here: https://example.com/

Blah blah.
```

The rule to implement this is relatively straightforward, and it's a widely accepted notation, so we would have an easier time teaching contributors how to use it. What do you think?

ricab: Oh, I didn't mean rejecting URLs if they didn't break any rules. If they fit in the line, no special treatment should be necessary, right? I mentioned quotes only as a way to work around problematic cases. But I am fine with your proposal too. Would you make it mandatory, or allow inline links (within boundaries)?

xmkg: > Oh, I didn't mean rejecting URLs if they didn't break any rules. If they fit in the line, no special treatment should be necessary, right? 

Yeah, I think the same way too.

> Would you make it mandatory, or allow inline links (within boundaries)?

I think it's fine to have inline links as long as it doesn't break the rules. Cool, I'll add an exception for `^\[\d+\]+:.*$` then.

jimporter: For long URLs, what if the MSG12 rule allowed single-word lines to exceed 72 characters? That would mainly apply to URLs, but it could also cover long identifier names or anything else where a line break in the middle would be confusing.

I don't have a strong opinion here though, so if people prefer another solution, that's ok too.

ricab: That makes sense to me too. I guess it could be added here or in a separate PR. Fine by me either way. I'll leave it up to @xmkg

git-hooks/commit-msg.py
xmkg: Done

git-hooks/commit-msg.py
xmkg: That sounds good to me as well. I'd prefer to have that discussion in a different PR though, since there might be some nuances to clarify.

git-hooks/commit-msg.py
```diff
@@ -49,23 +45,30 @@ def __init__(self, msg, *, strict=False):
...
         ]
 
-        self.strict = strict
         self.enough = False
 
-        # Strip comments and trailing whitespace
-        self.msg = re.sub(r"^#.*\n?", "", msg, flags=re.MULTILINE)
-        self.msg = self.msg.rstrip()
+        self.msg = subprocess.check_output(
+            ["git", "stripspace", "-s"], input=msg, text=True
```
ricab, line 51: 👍

git-hooks/commit-msg.py
```diff
@@ -28,16 +29,11 @@ class Rules(Enum):
         RULE5 =   "MSG5.  Limit the subject line to 50 characters (category included)."
         RULE6 =   "MSG6.  Do not end the subject line with a period."
         RULE8 =   "MSG8.  If adding a body, separate it from the subject with a blank line."
-        RULE10 =  "MSG10. Do not include more than 1 consecutive blank line."
-        RULE12 =  "MSG12. Wrap the body at 72 characters."
+        RULE10 =  "MSG10. Do not include more than 1 consecutive blank line, except in quoted text."
+        RULE12 =  "MSG12. Wrap the body at 72 characters, except on quoted lines and references."
```
ricab: Could we say?

```suggestion
        RULE12 =  "MSG12. Wrap the body at 72 characters, except on lines consisting only of blockquotes, references, or sign-offs"
```

To try to align with the code and be as precise as restrictive as possible in the intended exceptions.

git-hooks/commit-msg.py
```diff
@@ -49,23 +45,30 @@ def __init__(self, msg, *, strict=False):
...
         self.subject = self.lines[0].rstrip() if self.lines else ""
         self.body = self.lines[1:]
 
-        if not self.strict:
-            # In non-strict mode, allow special keywords used by `git rebase --autosquash`
-            self.subject = re.sub(r"^(fixup|squash)! ", "", self.subject)
-
         self.errors = self.validate_all()
 
+    def should_ignore_line(self, line):
```
ricab: Initially I was going to propose moving this inside `validate_rule12` or renaming to `should_ignore_length`. That would make the current scope clearer. But then I figured that we may need this in other rules as well, for the same reasons, and would naturally reuse the function. So how about renaming it as something like:

- `is_external_line`
- `is_constrained_line`
- `is_unavoidable_line`
- `is_modifiable_line`
- `is_uncontrollable_line`
- `is_untouchable_line`
- `is_inflexible_line` 

It took me a little to land on it, but I think `is_inflexible_line` is the closest to what we want to express. It leaves it open to reuse, without implying that it should be ignored in general. WDYT?

git-hooks/commit-msg.py
```diff
@@ -359,15 +349,74 @@ def test_rule12_body_line_comment_exceeds_72_chars(self):
...
+        for msg in invalid_messages:
+            self._test_rule("MSG12", msg, expect_failure=False)
+
+    def test_rule12_body_line_contains_a_long_signed_off_by(self):
+        invalid_messages = [
+            "[msg] Fellowship of the Ring\n\n"
+            "> This body line contains footnote references; [1] and [2]",
+            "Signed-Off-By: Aragorn II, King Elessar Telcontar, High King of the Reunited Kingdom of Gondor and Arnor. <aragorn@middleearth.localhost>",
+            "Signed-Off-By: Legolas Greenleaf Thranduilion, Prince of the Woodland Realm. <legolas@middleearth.localhost>",
+            "Signed-Off-By: Gimli, son of Glóin, of the House of Durin. <gimli@middleearth.localhost>",
```
ricab, line 405: 🤩

ricab (commented):
Hey @xmkg, It looks good and it makes it easy to add Jim's suggestion for single-word lines in the future. I have only a couple of suggestions and a couple of compliments.

git-hooks/commit-msg.py
xmkg: `inflexible` sounds good to me, too. I'll update the function name.

git-hooks/commit-msg.py
xmkg: Done

git-hooks/commit-msg.py
xmkg: Done

ricab (approved):
LGTM, thanks again!

PR #4339, opened 2025-09-04
[test-platform-linux] Override libvirt.so properly

xmkg:
The with_minimally_mocked_libvirt function was constructing a new
libvirt wrapper before the test body, but it's essentially a no-op
since the libvirt factory uses its own instance of the wrapper.

The patch fixes that by properly overriding the default libvirt.so
path so the mock implementation would be loaded.


sharder996:
Is this needed for CLI integration tests? We still submitting bug fixes for deprecated features 😉

xmkg:
> Is this needed for CLI integration tests? We still submitting bug fixes for deprecated features 😉 

It is just to make the UBsan happy during the test execution so we don't have to filter out the libvirt backend tests.

xmkg:
Dropping this in favor of https://github.com/canonical/multipass/pull/4365.

PR #4340, opened 2025-09-04
[format-utils] Validate timestamps before compare

xmkg:
The TimestampToNanoseconds function is only defined for valid timestamp values. The multipass::format::sorted function is converting the ts values without checking if they're valid first, leading to UB.

The patch fixes that by ensuring timestamp validity first.

### DIFF
--- a/include/multipass/cli/format_utils.h
+++ b/include/multipass/cli/format_utils.h
@@ -104,16 +104,28 @@ Container multipass::format::sorted(const Container& items)
             if constexpr (std::is_same_v<T, multipass::DetailedInfoItem>)
             {
                 if (a.has_snapshot_info() && a.name() == b.name())
-                    return TimeUtil::TimestampToNanoseconds(
-                               a.snapshot_info().fundamentals().creation_timestamp()) <
-                           TimeUtil::TimestampToNanoseconds(
-                               b.snapshot_info().fundamentals().creation_timestamp());
+                {
+                    const auto lhs = a.snapshot_info().fundamentals().creation_timestamp();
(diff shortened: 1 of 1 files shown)


sharder996 (approved):
Just needs a rebase to get through CI.

include/multipass/cli/format_utils.h
```diff
@@ -104,16 +104,28 @@ Container multipass::format::sorted(const Container& items)
...
             {
                 if (a.has_snapshot_info() && a.name() == b.name())
-                    return TimeUtil::TimestampToNanoseconds(
-                               a.snapshot_info().fundamentals().creation_timestamp()) <
-                           TimeUtil::TimestampToNanoseconds(
-                               b.snapshot_info().fundamentals().creation_timestamp());
+                {
+                    const auto lhs = a.snapshot_info().fundamentals().creation_timestamp();
+                    const auto rhs = b.snapshot_info().fundamentals().creation_timestamp();
+                    if (TimeUtil::IsTimestampValid(lhs) && TimeUtil::IsTimestampValid(rhs))
```
sharder996, line 110: This would already be getting checked in the [`protobuf`](https://github.com/protocolbuffers/protobuf/blob/2f270c49a2886945fc405dbdaa5fe9467d3d31fa/src/google/protobuf/util/time_util.cc#L359) library, but I believe only for debug builds.

Since this timestamp is originally coming from a json file that could potentially be modified or corrupted, I would say this is just good defensive programming.

ricab (approved):
LGTM, thanks!

Issue #4198, opened 2025-07-01
[windows] Launch fails with custom image from different drive


Sploder12:
**Describe the bug**
When using `file://` to launch a custom image, trying to use a file from a drive that isn't `C:` will result in a file not found error. 

**To Reproduce**
1. `multipass launch file://D:\<PATH_TO_IMAGE>` 
2. See "Custom image `/<PATH_TO_IMAGE>` does not exist."

**Expected behavior**
The custom image is launched.

**Additional info**
- OS: Windows 11
- `multipass version` 1.16.0 RC

**Additional context**
This is related to how we use `QUrl` in `default_vm_image_vault.cpp`. `.path()` does not contain the drive, the drive is stored in `.host()` (if present).

jimporter:
Taking a look at this.

From the [Wikipedia summary](https://en.wikipedia.org/wiki/File_URI_scheme#Windows_2) of `file:` URLs, I think `file://D:\...` is incorrect syntax, and there should be an extra slash near the beginning:

> To access a file "the file.txt", the following might be used.
>
> For a network location:
>
>     file://hostname/path/to/the%20file.txt
>
> Or for a local file, the hostname is omitted, but the slash is not (note the third slash):
>
>     file:///c:/path/to/the%20file.txt

Still, this seems like a common enough error to make, so if nothing else, a better error message would be nice. It might also be worth checking the value of `.host()` for other issues. For example, it would be surprising if a user passed `file://remote_host/path/to/file`, but multipass opened `/path/to/file` on the *local* host.

Sploder12:
Oops, good catch. Unfortunately the issue still remains

``` Powershell
PS C:\Users\trevor> ls d:/tmpiso
    Directory: D:\tmpiso

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         9/18/2025   7:42 AM     6345887744 ubuntu24.04.iso

PS C:\Users\trevor> multipass launch file:///d:/tmpiso/ubuntu24.04.iso
launch failed: Custom image `/d:/tmpiso/ubuntu24.04.iso` does not exist.

PS C:\Users\trevor> multipass launch file://d:/tmpiso/ubuntu24.04.iso
launch failed: Custom image `/tmpiso/ubuntu24.04.iso` does not exist.
```

jimporter:
> PS C:\Users\trevor> multipass launch file:///d:/tmpiso/ubuntu24.04.iso
[...]

PR #4341, opened 2025-09-04
[daemon] Improve file URL error message (#4198)

jimporter:
Local file URLs should have three slashes after "file:".

This just adds an error message alerting the user to the fact that non-local file URLs aren't supported (at least, not yet), and suggesting a fix. The wording feels a little clumsy to me, but I'm not sure of a better way to phrase it without becoming confusing.


sharder996:
Hi @jimporter, there are parts of our GitHub workflows that use tokens which aren't authorized for PRs from external contributors. Since you're in the Canonical org now, simply mirroring this branch into the Multipass repo and closing this PR would probably be the easiest.

jimporter:
@sharder996 Thanks. Moved this PR over to https://github.com/canonical/multipass/pull/4352.

PR #4348, opened 2025-09-08
[docs] Fix linux build instructions

tobe2098:
Fixes autostart file path in the linux build instructions. 
Also adds an additional recommendation to said instructions.

### DIFF
--- a/BUILD.linux.md
+++ b/BUILD.linux.md
@@ -93,6 +93,12 @@ Optionally, enable auto-complete in Bash:
 source <multipass>/completions/bash/multipass
 ```
 
+To be able to use the binaries without specifying their path:
+
+```
+export PATH=<build-dir>/bin
+```
+
 Now you can use the `multipass` command from your terminal (for example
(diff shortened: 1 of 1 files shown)


BUILD.linux.md
```diff
@@ -93,6 +93,12 @@ Optionally, enable auto-complete in Bash:
 source <multipass>/completions/bash/multipass
 ```
 
+To be able to use the binaries without specifying their path:
+
+```
+export PATH=<build-dir>/bin
```
ricab: The instructions create an in-source build and so use `<multipass>/build` for the build directory in other places. I am thinking this should follow that tune here too. WDYT?

ricab (commented):
Thanks @tobe2098, looks good overall. Just a tiny thing inline.

BUILD.linux.md
tobe2098: I agree, that way it is more understandable

ricab (approved):
Looks good @tobe2098, thank you!

PR #4352, opened 2025-09-09
[daemon] Improve file URL error message (#4198)

jimporter:
Local file URLs should have three slashes after "file:".

This just adds an error message alerting the user to the fact that non-local file URLs aren't supported (at least, not yet), and suggesting a fix. The wording feels a little clumsy to me, but I'm not sure of a better way to phrase it without becoming confusing.

(Recreated this from #4341 in the original repo so CI passes.)

### DIFF
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -281,6 +281,10 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type,
         QUrl image_url(QString::fromStdString(query.release));
         VMImage source_image, vm_image;
 
+        if (image_url.host().size() != 0)
+            throw std::runtime_error(
+                fmt::format("Invalid file URL `{}`; did you forget a slash?", query.release));
+
         if (!QFile::exists(image_url.path()))
             throw std::runtime_error(
                 fmt::format("Custom image `{}` does not exist.", image_url.path()));
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -285,11 +285,11 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type,
             throw std::runtime_error(
                 fmt::format("Invalid file URL `{}`; did you forget a slash?", query.release));
 
-        if (!QFile::exists(image_url.path()))
-            throw std::runtime_error(
-                fmt::format("Custom image `{}` does not exist.", image_url.path()));
+        source_image.image_path = image_url.toLocalFile();
 
-        source_image.image_path = image_url.path();
+        if (!QFile::exists(source_image.image_path))
+            throw std::runtime_error(
+                fmt::format("Custom image `{}` does not exist.", source_image.image_path));
 
         if (source_image.image_path.endsWith(".xz"))
         {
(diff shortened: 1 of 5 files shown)


src/daemon/default_vm_image_vault.cpp
```diff
@@ -281,6 +281,12 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type,
         QUrl image_url(QString::fromStdString(query.release));
         VMImage source_image, vm_image;
 
+        if (image_url.host().size() != 0)
+            throw std::runtime_error(fmt::format("Unexpected hostname in file URL; did you mean "
```
xmkg: I don't think the "unexpected hostname" part would be valuable to the user. Could we say something like, "Malformed file URL, did you miss a slash?" without going into details?

xmkg (changes requested):
Hi @jimporter, I think it's a good UX improvement! I have a small comment regarding the wording.

src/daemon/default_vm_image_vault.cpp
jimporter: Sounds good to me. I replaced "malformed" with "invalid" though, since a file URL with two slashes is *technically* well-formed according to the spec; it's just not valid within Multipass.

jimporter:
> Can I get a unit test for the added conditional?

Now done (assuming CI passes, anyway...)

jimporter:
Ok, I found that there's similar code in the LXD backend (which I think is going away, but we might as well fix this since it's easy). Added the relevant code/tests for that too.

sharder996 (commented):
I wanted to try testing this, but I'm unable to trigger the error:
```
% multipass launch file://path/to/image   
launch failed: Custom image `/Users/scott/Documents/dev/multipass/build-gui/path/to/image` does not exist.
```

jimporter:
> I wanted to try testing this, but I'm unable to trigger the error:
> 
> ```
> % multipass launch file://path/to/image   
> launch failed: Custom image `/Users/scott/Documents/dev/multipass/build-gui/path/to/image` does not exist.
> ```

I think you need to run on Windows to reproduce this error from the command line, since the following code in the client transforms the `file` URI in case the user passed in a relative path: https://github.com/canonical/multipass/blob/6906c24d3267030fade72e003a48f6c24f510266/src/client/cli/cmd/launch.cpp#L289-L294

This ends up "fixing" things on non-Windows, since the leading character of an absolute file name is always `/` there. However, I'm not sure it's the right thing to do. A user passing in a "relative URI" is a very strange thing to do, and it'd probably be safer to require users to spell that like `file://$PWD/path/to/image`, or something like that.

I'd be ok with removing the above code, but it's a compatibility break, and maybe there are people out there relying on it. We could also add a warning in that code too, indicating that this behavior will go away sooner or later.

xmkg (dismissed):
LGTM (secondary review)

tests/disabling_macros.h
```diff
@@ -27,17 +27,20 @@
 #define DISABLE_ON_WINDOWS(testName) DISABLED_##testName
 #define DISABLE_ON_WINDOWS_AND_MACOS(testName) DISABLED_##testName
 #define DISABLE_ON_MACOS(testName) testName
+#define DISABLE_ON_LINUX_AND_MACOS(testName) testName
```
sharder996: Depending on how technical you want to be, this could be renamed `DISABLE_ON_UNIX`. The naming scheme kinda makes sense since we already have a `platform_unix.cpp` file.

sharder996 (approved):

tests/disabling_macros.h
sharder996: I'll approve anyways and let you decided whether you want to merge or discuss a different name.

jimporter: That makes sense to me.

sharder996 (approved):

Issue #4134, opened 2025-06-02
Fix CI for external contributions


ricab:
**What are you trying to do?**
PRs from forks do not have the necessary secrets to upload packages to AWS or coverage reports Codecov. This causes builds to fail, which prevents PRs from being merged.

**What's your proposed solution?**
Somehow allow PRs to be merged regardless. Possible approaches
  - For Codecov, investigate what is the recommended way to deal with external PRs.
  - Let workflows succeed even when upload steps fail
  - Split out Upload steps such that the existing branch protections succeed.
  - Extend solution for https://github.com/canonical/multipass/issues/4042

Here's an example of how external PRs fail: https://github.com/canonical/multipass/pull/4072

PR #4353, opened 2025-09-10
[actions] WIP: Use codecov's GitHub Action

jimporter:
Note: this is currently a draft, filed as a PR from my fork to see if I can get Codecov to work properly from external forks.

This should improve the situation for #4134.


.github/workflows/linux.yml
```diff
@@ -158,7 +158,7 @@ jobs:
               curl -L -o /usr/local/bin/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe\
               ln -sf /usr/local/bin/nuget.exe /usr/local/bin/nuget\
               export PATH=$PATH:/usr/local/bin\
-              mono /usr/local/bin/nuget.exe sources add -name GitHubPackages -source "${FEED_URL}" -username "${USERNAME}" -password "${GITHUB_TOKEN}" -storePasswordInClearText\
+              mono /usr/local/bin/nuget.exe sources add -name GitHubPackages -source "${FEED_URL}" -username "${USERNAME}" -password "${GITHUB_TOKEN}" -storePasswordInClearText || :\
```
ricab: Could we maybe get warned if this fails? Or perhaps only run it when expected to work?

jimporter: I'm not actually sure why it fails, since it doesn't seem to be consistent from looking at previous CI jobs. The only thing I can think of right now is that the CI configuration seems to preserve some state between runs as a cache, and this is one of the things that carries over across runs. If that's the case, then I think we could change this to something like, "If the source isn't already added, then add it." That way, other errors still cause CI to fail.

ricab: Yes, we save vcpkg caches for reuse. When I saw this change, I assumed that you allowed the error because you had noticed this failing in forks. And I assumed that the failure would be caused by either a mismatch in the repo/account associated with the cache or unavailable permissions to access canonical caches. Indeed, all those would be good reasons to ignore vcpkg caches and let the build go through.

However, just ignoring the error in general would risk missing relevant problems when the command should succeed. I thought we'd better try to avoid going unaware of a broken vcpkg caching mechanism, if it occurred.

> If that's the case, then I think we could change this to something like, "If the source isn't already added, then add it." That way, other errors still cause CI to fail.

Adding the source only if it isn't there yet makes sense to me, but would it help with external forks somehow? Sorry, I am a little confused by this.

BTW, we should make sure and avoid uploading any caches from external forks (not sure if that would require any further changes off the top of my head).

@sharder996, you added those caches, so I'd love to hear your input on this.

jimporter: Digging through the logs, I see [this](https://github.com/canonical/multipass/actions/runs/17600387041/job/50009414564#step:22:102) in the failing runs in this PR:

```
Updating build for multipass ('PULL' step changed)
...
:: + mono /usr/local/bin/nuget.exe sources add -name GitHubPackages -source https://nuget.pkg.github.com/canonical/index.json -username canonical*** -storePasswordInClearText
:: The name specified has already been added to the list of available package sources. Provide a unique name.
'override-build' in part 'multipass' failed with code 1.
```

Every other CI job I've looked at says `Skipping build for multipass (already ran)`, so the `nuget sources add ...` part doesn't run in those jobs. I don't think this is solely caused by external forks though, since [this CI run](https://github.com/canonical/multipass/actions/runs/17562083277/job/49931500960) is from my fork, and it passed the Linux Release job.

jimporter: I updated this so that it first checks whether the `GitHubPackages` source is already present. That should be pretty safe (though I'm still not 100% clear on what triggers this code path).

.github/workflows/linux.yml
```diff
@@ -237,7 +237,13 @@ jobs:
           ulimit -c unlimited && \
           env CTEST_OUTPUT_ON_FAILURE=1 \
             cmake --build /root/parts/multipass/build --target covreport"
-        bash <(curl -s https://codecov.io/bash) -Z -s ${{ steps.coverage-setup.outputs.build }}
+
+    - name: Upload coverage
+      uses: codecov/codecov-action@v5
```
sharder996, line 243: How does pathing work here? My initial thought is that you might have to pull the codecov report out of the snapcraft container.

.github/workflows/linux.yml
```diff
@@ -158,7 +158,9 @@ jobs:
               curl -L -o /usr/local/bin/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe\
               ln -sf /usr/local/bin/nuget.exe /usr/local/bin/nuget\
               export PATH=$PATH:/usr/local/bin\
-              mono /usr/local/bin/nuget.exe sources add -name GitHubPackages -source "${FEED_URL}" -username "${USERNAME}" -password "${GITHUB_TOKEN}" -storePasswordInClearText\
+              if ! mono /usr/local/bin/nuget.exe sources list -format short | grep ${FEED_URL} &> /dev/null; then\
```
sharder996, line 161: Is it the name, `-name GitHubPackages`, that should be looked for?

.github/workflows/linux.yml
sharder996: Ah, I'm not sure what triggers `Updating build for multipass ('PULL' step changed)`, but its probably complaining because that nuget source will have already been added during the [Build](https://github.com/canonical/multipass/actions/runs/17600387041/job/50009414564#step:13:27) stage.

.github/workflows/linux.yml
jimporter: `nuget sources list -format short` [produces output like this](https://github.com/NuGet/Home/issues/13011#issuecomment-1811593599):

```
E https://example.com/index.json
```

So as long as the feed URL associated with the `GitHubPackages` source name is in the output, then it's already been added.

.github/workflows/linux.yml
jimporter: It should be the same as before. Previously, we used the [`-s` argument](https://github.com/codecov/codecov-bash/blob/1817ab23d6b35ff6596e48d15db366b71f17ada3/codecov#L112-L113) from the Bash uploader (which has been deprecated for quite some time now), and now we use the [`directory` argument](https://github.com/codecov/codecov-action/blob/5a1091511ad55cbe89839c7260b706298ca349f7/action.yml#L19-L21) from the action. These *should* have the same meaning.

However, I did notice that I forgot to include `if: ${{ matrix.build-type == 'Coverage' }}` in the coverage uploading step. Now fixed.

.github/workflows/linux.yml
jimporter: @sharder996 That makes sense. Then we probably want one of the following solutions:
1. This one (check if the source is added before trying)
2. Move the `nuget sources add` call to some other spot that only runs once
3. Ensure Snap doesn't have to update the build in this step

I don't have a terribly strong opinion about how we do it, especially since I'm not sure what the right way to do things in Snap is (at least, not yet).

.github/workflows/linux.yml
sharder996: Right. I see a directory being mounted into the lxc container beforehand (outside the diff window).

.github/workflows/linux.yml
sharder996: I would say 2 or 3 would be preferable, but I do not have any ideas on how to implement them. For the time being, I would say that 1 is fine.

ricab (approved):
LGTM, thanks a lot for this @jimporter!

Issue #4113, opened 2025-05-26
Revise build instructions


ricab:
**Description of documentation issue or suggestion**
Our build instructions need to be revised, improved, and updated. There are multiple aspects where they are obsolete, inconsistent, or just wrong. Comments on #4110 point out a few examples.

dmora4:
Hi @ricab 
I would like to help out with this issue. I have looked over the discussion you cited ([#4110](https://github.com/canonical/multipass/pull/4110)) and I have gathered the examples pointed out in the comments:
* Consider the build architecture when specifying the PATH environment variable
* Consider the build architecture when specifying the path for the DCMAKE_PREFIX_PATH cmake argument
* Remove mentions of installing QT via brew
* Either keep "latest stable version" and remove the specific Qt6 version, or keep the version and remove  remove reference to "latest stable version"
* Provide more context in the Enable Hyper-V section since it is not strictly required

Let me know if this issue is meant to capture only the examples cited above or if the scope of this issue involves a larger overhaul of the build instructions. Either way I am happy to help!

ricab:
Hi @dmora4, thank you for your interest! This was meant as an umbrella to cover all problems with the build instructions, but you don't need to fix everything in one go. You could submit PRs dealing with a subset of issues and mention that it "addresses" this issue (rather than "fixes"). See https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

Of course you can also fix it all 😄 Either way, your contributions would be very welcome!

ayaankhan98:
@dmora4 let me know if you working on this, otherwise I can raise a PR addressing the suggested changes.

dmora4:
@ayaankhan98 No, I haven't gotten around to it so go right ahead!

prachimaskar:
Hi 👋 I'd love to contribute to this issue.

I’ve reviewed the discussion and would be happy to help by:
- Removing the outdated Qt via Homebrew reference
- Clarifying the Qt version requirement (Qt 6.2.4)
- Adding architecture-specific notes for `PATH` and `CMAKE_PREFIX_PATH`
- Improving the Hyper-V section for clarity

If no one is currently working on this, I'd be grateful for the opportunity to take it forward and will open a PR soon. Please let me know if that sounds good — thank you!

geoffreynyaga:
Closing this as it was fixed in PR  https://github.com/canonical/multipass/pull/4203

PR #4354, opened 2025-09-10
Docs/fix windows build instructions

tobe2098:
Following from #4113, the build instructions for Windows were reviewed and tested on Windows 11 Pro. The instructions were adapted to cover the gap to the newer OS to allow building in its environment. The installation of `git` and `qt6` were also unified to be done through chocolatey and aqt respectively.

Additionally, general formatting and organization of the file was changed minimally to improve visual clarity and to unify formatting in the file.

The Windows 10 Pro specific instructions were left as-is when they would conflict with the Windows 11 Pro instructions. In those cases, a separate section for Windows 11 Pro was made.

Untested sections so far:
Permissions/privileges for multipassd in Windows 11
ninja package in Windows 11
cmder and powershell-only console setups in Windows 11
Installation while disabling the Anti-Virus in Windows 11

### DIFF
--- a/BUILD.windows.md
+++ b/BUILD.windows.md
@@ -6,23 +6,30 @@
 
 Install chocolatey, a package manager, to download the rest of the dependencies: <https://chocolatey.org/>.
 
-Press Windows Key+X and Run Windows PowerShell(Admin) then follow the chocolatey instructions to "Install with
-Powershell.exe".
+Press Windows Key+X and Run Windows PowerShell(Admin) then follow the chocolatey instructions to
+"Install with Powershell.exe".
 
 ### Dependencies
-After chocolatey is installed you can now install the rest of the dependencies from the Powershell(Admin). To get the best results, in the following order:
-```
+
+After chocolatey is installed you can now install the rest of the dependencies from the
+Powershell(Admin). To get the best results, in the following order:
+
+```[pwsh]
 choco install cmake ninja qemu openssl git wget unzip -yfd
 ```
-```
+
+```[pwsh]
 choco install visualstudio2019buildtools visualstudio2019-workload-vctools -yfd
 ```
 
-You may have to disable Windows Defender Real-time protection if you want the packages to install quicker. Search for
-Windows Defender Security Center, go to Virus & threat protection, then Virus and thread protection settings, disable Real-time protection.
+You may have to disable Windows Defender Real-time protection if you want the packages to install
+quicker. Search for Windows Defender Security Center, go to Virus & threat protection, then Virus
+and thread protection settings, disable Real-time protection.
 
-NOTE: visualcpp-build-tools is only the installer package. For this reason, choco cannot detect any new compiler tool
-updates so choco upgrade will report no new updates available. To update the compiler and related tooling or fix a broken `visualstudio2019buildtools` installation do the following:
+NOTE: visualcpp-build-tools is only the installer package. For this reason, choco cannot detect any
+new compiler tool updates so choco upgrade will report no new updates available. To update the
+compiler and related tooling or fix a broken `visualstudio2019buildtools` installation do the following:
+
 1. Go to "Add or remove programs"
 2. Search for the Microsoft Visual Studio Installer
 3. Click Modify
@@ -30,24 +37,27 @@ updates so choco upgrade will report no new updates available. To update the com
 5. For Windows 11, add "C++/CLI support for v142 build tools" in the "Desktop development with C++" kit
 6. Complete the installation
 
-
 ### Git
 
 You need to enable symlinks in Windows Git, have a look at
 [the git-for-windows docs](https://github.com/git-for-windows/git/wiki/Symbolic-Links).
 
 For Windows 11:
+
 1. Go to "Developer Settings"
 2. Enable "Developer mode"
 
 ### Qt6
 
 To install Qt6, use `aqt`. First install it with chocolatey:
-```
+
+```[pwsh]
 choco install aqt -yfd
 ```
+
 Then specify the following options in the installation command:
-```
+
+```[pwsh]
 aqt install-qt windows desktop 6.2.4 win64_msvc2019_64 -O C:/Qt
 ```
 
@@ -67,7 +77,8 @@ Search for "Edit environment variables for your account" then edit your Path var
 Cmder is a sane terminal emulator for windows, which includes git and SSH support among other things.
 
 Install with chocolatey:
-```
+
+```[pwsh]
 choco install cmder -yfd
 ```
 
@@ -81,23 +92,30 @@ The following will setup a task that you can use to build things with the VS2019
     ``/icon "%CMDER_ROOT%\icons\cmder.ico"``
 6. In the commands box, add:
 
-    ``cmd /k ""%ConEmuDir%\..\init.bat" && "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64" -new_console:d:%USERPROFILE%``
+    ```[]
+    cmd /k ""%ConEmuDir%\..\init.bat" && "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64" -new_console:d:%USERPROFILE%
+    ```
+
 7. Give the task a name (first box), such as vs2019 and click Save Settings.
 8. Now run cmder, click on the green "+" and click on the vs2019 task
-This will open a new terminal tab and run the VS2019 setup. CMake can now find the VS compiler. Go to the [Building](./BUILD.windows.md#building) section.
+This will open a new terminal tab and run the VS2019 setup. CMake can now find the VS compiler.
+Go to the [Building](./BUILD.windows.md#building) section.
 
 #### x64 Native Tools Command Prompt
 
-x64 Native Tools Command Prompt is the native console from the MVSC installation. On startup it updates its environment variables to include all tools installed via MSVC for 64-bit. If you are using 32-bit, use the x86 Native Tools Command Prompt.
+x64 Native Tools Command Prompt is the native console from the MVSC installation. On startup it
+updates its environment variables to include all tools installed via MSVC for 64-bit. If you are
+using 32-bit, use the x86 Native Tools Command Prompt.
 
-If you want to use the environment variables in a different console, look at: https://learn.microsoft.com/nb-no/cpp/build/building-on-the-command-line?view=msvc-170.
+If you want to use the environment variables in a different console, look at:
+[Use the Microsoft C++ toolset from the command line](https://learn.microsoft.com/nb-no/cpp/build/building-on-the-command-line?view=msvc-170).
 
 #### Powershell only
 
 Using a Visual-Studio command-prompt-enabled PowerShell to build the project is also possible.
 Open a new PowerShell terminal, then invoke the following commands:
 
-```pwsh
+```[pwsh]
 # Try to determine the Visual Studio install path
 $VSPath = & "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -products Microsoft.VisualStudio.Product.BuildTools -latest -property installationPath
 
@@ -110,16 +128,18 @@ Enter-VsDevShell -VsInstallPath "$VSPath" -DevCmdArguments '-arch=x64'
 
 ## Building
 
-```
+```[batch]
 cd <multipass>
 git submodule update --init --recursive
 mkdir build
 cd build
 ```
-```
+
+```[batch]
 cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=..\3rd-party\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_PREFIX_PATH=C:\Qt\6.2.4\msvc2019_64\ ../
 ```
-```
+
+```[batch]
 cmake --build .
 ```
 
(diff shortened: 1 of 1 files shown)


ricab:
Hey @levkropp, would you mind reviewing this one please?

BUILD.windows.md
```diff
@@ -1,44 +1,55 @@
...
-After chocolatey is installed you can now install the rest of the dependencies:
-
-    choco install visualstudio2019buildtools visualstudio2019-workload-vctools cmake ninja cmder qemu-img openssl -yfd
+### Dependencies
+After chocolatey is installed you can now install the rest of the dependencies from the Powershell(Admin). To get the best results, in the following order:
+```
+choco install cmake ninja qemu openssl git wget unzip -yfd
+```
+```
+choco install visualstudio2019buildtools visualstudio2019-workload-vctools -yfd
```
sharder996, line 22: We should probably move to instructing to build with `visualstudio2022buildtools`, especially since CI is using that version, I believe. I've been using 2022 build tools now and its a simply swap. Just need to remember to update the cmder script below.

BUILD.windows.md
```diff
@@ -1,44 +1,55 @@
...
-[qtbase archive](https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_624/qt.qt6.624.win64_msvc2019_64/6.2.4-0-202203140926qtbase-Windows-Windows_10_21H2-MSVC2019-Windows-Windows_10_21H2-X86_64.7z)
-and extract it to `C:\Qt` (so it ends up in `C:\Qt\6.2.4`).
+To install Qt6, use `aqt`. First install it with chocolatey:
+```
+choco install aqt -yfd
+```
+Then specify the following options in the installation command:
+```
+aqt install-qt windows desktop 6.2.4 win64_msvc2019_64 -O C:/Qt
+```
```
sharder996, line 58: Personally, I would like to keep the link to the archive here. It has come in handy for myself a few times.

BUILD.windows.md
```diff
@@ -99,40 +108,55 @@ Import-Module "$VSPath\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"
...
-Running multipass
----------------------------------------
+```
+cd <multipass>
+git submodule update --init --recursive
+mkdir build
+cd build
+```
+```
+cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=..\3rd-party\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_PREFIX_PATH=C:\Qt\6.2.4\msvc2019_64\ ../
```
sharder996, line 138: The toolchain file and prefix path location shouldn't be necessary if you're using cmder.

BUILD.windows.md
```diff
@@ -99,40 +108,55 @@ Import-Module "$VSPath\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"
...
+```
+```
+cmake --build .
+```
+
+This builds `multipass` and `multipassd`.
+To create an installer, run `ninja package`.
```
levkropp: If we are replacing our `ninja` command with `cmake --build .` perhaps we should change the `ninja package` command to something like `cmake --build . --target package` as well

BUILD.windows.md
sharder996, line 1: I see some formatting changes that are removing some newlines. While it doesn't really matter to me, the reason the document is formatted the way it is, is because of [.clang-format](https://github.com/canonical/multipass/blob/802febe291cf8c3bb59eaa3d3eb0586ba5ed9f7d/.clang-format#L32). So, in the name of staying consistent, you might be inclined to continue following it.

sharder996 (changes requested):
Just a few comments from taking a glance at the diff.

Thanks for updating the instructions @tobe2098

levkropp (approved):
Looks like a great improvement! Just have one nitpick but not going to block on it. I agree with @sharder996 that the archive link as an alternative should be kept but using `aqt` like our Github CI is a perfect alternative to installing Qt from the website. Thank you @tobe2098 !

BUILD.windows.md
tobe2098: I have not tested the 2022 tools in a clean install, but I can certainly add it as an alternative. At least until it can be tested it in a clean install again.

BUILD.windows.md
tobe2098: Yes, I agree that will be better.

BUILD.windows.md
tobe2098: I did not test the installation with cmder, but for the x64 native console I found the installation failed unless those paths were specified in the cmake command. I will put that it is not necessary if using cmder.

BUILD.windows.md
tobe2098: I will review this, it is important to stay consistent in the format

BUILD.windows.md
tobe2098: I agree as well :)

BUILD.windows.md
```diff
@@ -49,46 +86,63 @@ Search for "Edit environment variables for your account" then edit your Path var
...
-``cmd /k ""%ConEmuDir%\..\init.bat" && "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64" -new_console:d:%USERPROFILE%``
+```[pwsh]
+choco install cmder -yfd
+```
+
+The following will setup a task that you can use to build things with the VS2019 compiler toolchain.
 
-Give the task a name (first box), such as vs2019 and click Save Settings.
+1. Run cmder which should be installed by default on C:\tools\cmder\Cmder.exe
+2. Click on the green "+" sign on the right lower corner of cmder
```
AhiyaHiya: In this PR, would it be possible to update the following sentence?:
```
Click on the green "+" sign on the right lower corner of cmder
```

It would be clearer to the user, that they should select the dropdown arrow and not the "Create new console" button(which is the green "+" button):
```
Click the dropdown arrow next to the green "+" sign in the lower right corner of cmder
```

BUILD.windows.md
```diff
@@ -1,43 +1,80 @@
...
-Alternatively, download the
-[qtbase archive](https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_624/qt.qt6.624.win64_msvc2019_64/6.2.4-0-202203140926qtbase-Windows-Windows_10_21H2-MSVC2019-Windows-Windows_10_21H2-X86_64.7z)
+```[pwsh]
+aqt install-qt windows desktop 6.2.4 win64_msvc2019_64 -O C:/Qt
+```
+
+If you installed the VS2022 toolchain, use the following options:
+
+```[pwsh]
+aqt install-qt windows desktop 6.2.4 win64_msvc2022_64 -O C:/Qt
```
AhiyaHiya: On Windows 11, I attempted the above command in `Administrator: Developer PowerShell for VS 2022` and encountered the following error:

```
PS C:\Program Files\Microsoft Visual Studio\2022\Community> aqt install-qt windows desktop 6.2.4 win64_msvc2022_64 -O C:/Qt
INFO    : aqtinstall(aqt) v3.3.0 on Python 3.13.3 [CPython MSC v.1943 64 bit (AMD64)]
ERROR   : The packages ['qt_base'] were not found while parsing XML of package information!
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt windows desktop --arch 6.2.4' to show architectures available.
*
```

When querying for which version of Qt has `win64_msvc2022_64`, I found that the earliest version of Qt to have `win64_msvc2022_64` was `6.8.0`:

```
PS C:\Program Files\Microsoft Visual Studio\2022\Community> aqt list-qt windows desktop --arch 6.7.3
win64_llvm_mingw win64_mingw win64_msvc2019_64 win64_msvc2019_arm64

PS C:\Program Files\Microsoft Visual Studio\2022\Community> aqt list-qt windows desktop --arch 6.8.0
win64_llvm_mingw win64_mingw win64_msvc2022_64 win64_msvc2022_arm64_cross_compiled
```

I'm assuming that the doc may have to be updated to reflect the newer Qt version?

BUILD.windows.md
tobe2098: I apologize, since I only did formatting in that section I did not check for correctness. Good catch!

BUILD.windows.md
tobe2098: I see. I added the MSVC2022 due to demand, but did not test. Great catch! I will set the version properly with a warning for the fact that we do not build and test with it, since we build and test with 6.2.4 mainly.

BUILD.windows.md
tobe2098: > We should probably move to instructing to build with `visualstudio2022buildtools`, especially since CI is using that version, I believe. I've been using 2022 build tools now and its a simply swap. Just need to remember to update the cmder script below.

@sharder996, did you test this locally? @AhiyaHiya seems to have issues installing Qt 6.2.4 with MSVC2022. If we cannot support building with Qt >6.2.4 I think we should not support building with MSVC2022 either then. What do you think?

BUILD.windows.md
sharder996: I've been using it myself with the Cmder configuration on Windows 10, so there might be a configuration difference as to why I haven't experienced any behaviour like this.

@xmkg You've been using 2022, correct? Do you have any additional notes for using the newer compiler version?

BUILD.windows.md
sharder996: I replied in the other thread and summoned Mustafa, who I'm pretty sure has been using some updated build tools. We'll see what he has to say.

BUILD.windows.md
tobe2098: Is it possible that Qt 6.2.4 with arch MSVC2019 is compatible with MSVC2022? Seems like the only possible conclusion, because the problem is that Qt <6.8.0 does not offer MSVC2022 architecture either in aqt or in the archive.
Which installation options (the architecture) are you @sharder996 and @xmkg using?

sharder996: I went and checked and I still have Visual Studio Build Tools 2019 installed, so Qt might be using that in the background. Unless @xmkg has a different setup, it might then still be too early to switch over to 2022.

tobe2098: As far as I know, unless we update past Qt 6.8 we cannot switch over to 2022. The curious part is that in the CI pipeline we are running windows-latest, and according to [docs](https://github.com/ilammy/msvc-dev-cmd) , if msvc-setup is run on windows-latest it installs by default the MSVC2022, so somehow that works. 

I think we can conclude for now that we have to wait to update the MSVC version. Thank you both!

BUILD.windows.md
```diff
@@ -1,43 +1,67 @@
...
+choco install cmake ninja qemu openssl git wget unzip -yfd
+```
+
+```[pwsh]
+choco install visualstudio2019buildtools visualstudio2019-workload-vctools -yfd
+```
+
+You may have to disable Windows Defender Real-time protection if you want the packages to install
+quicker. Search for Windows Defender Security Center, go to Virus & threat protection, then Virus
+and thread protection settings, disable Real-time protection.
```
sharder996: Being an open-source repo with community influence, I think we should avoid this altogether and never recommend turning off security features. I would reserve this info only for developer documentation or on a case-by-case basis if a contributor is being bothered by the setup process.

sharder996 (changes requested):

BUILD.windows.md
levkropp: > I think we should avoid this altogether and never recommend turning off security features.

We are already instructing to enable Developer Mode in Windows 11 as part of developing for Windows, which in Microsoft's own words "could expose your device and personal data to security risks or harm your device". I would also consider Windows build instructions for Multipass in the category of "developer documentation".

That being said, **I still agree that we should not have this section** because we don't do it in our CI for Windows Server 2022 (current `windows-latest`), so we are not practicing what we preach here. Personally, I have never had an issue with the speed of installation of packages that warranted me investigating turning off Real-Time Protection.

If turning off Real-Time Protection is only relevant for Windows 10, then it would be important to note that [Windows 10 Pro support ends on October 14th 2025](https://learn.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro) and perhaps a larger discussion should be had about continuing to support building on Windows 10 in our documentation at all.

xmkg: I agree with what @sharder996 said here. I think we shouldn't nudge users towards disabling a security feature in favor of performance. Developers are often power users, and the performance impact of Windows Defender is well-known in the Windows-y circles, so it's relatively easy to obtain this knowledge when needed.

BUILD.windows.md
xmkg: > Unless @xmkg has a different setup, it might then still be too early to switch over to 2022.

I've had success with both MSVC2019 and MSVC2022, but with the QT upgrade caveat that @tobe2098 pointed out. I've used the MSVC2022 built artifacts during the Hyper-V feature development process and encountered no problems so far.

It seems like we're already going to have to upgrade the QT anyway (as a part of the C++20 upgrade task @jimporter is working on), so bumping MSVC to 2022 would be a nice upgrade, too. The only part that makes me *slightly* concerned is whether the MSVC2022-built artifacts would work in Windows 10 or not, but I'd expect them to work as long as the MSVC2022 runtime is installed in the system; but still, that's something to test.

BUILD.windows.md
tobe2098: I agree. I did not touch it because I did not test it, but, as said, it is not a necessary step and building is perfectly fine without it. I also agree with the security-related comments.

BUILD.windows.md
xmkg: I think it's a good move to replace the build system commands with the build system generator ones. We're relying on CMake for build system abstraction, and the "generator" is a configure-time parameter of the BSG. Therefore, we should not assume anything about the generated build system.

BUILD.windows.md
tobe2098: I think, ideally, these instructions need a re-write once we move away from Qt as a whole, and will not have to deal with this anymore. 
Apart from that, I think that once the upgrade of Qt is done we can test and see if it will work in Windows 10 and test the build instructions again, although there should be no issues.

BUILD.windows.md
```diff
@@ -99,40 +125,63 @@ Import-Module "$VSPath\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"
...
+mkdir build
+cd build
+```
+
+```[batch]
+cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=..\3rd-party\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_PREFIX_PATH=C:\Qt\6.2.4\msvc2019_64\ ../
+```
+
+```[batch]
+cmake --build .
```
sharder996, line 142: The previous command you used Ninja as a generator. We should use ninja here then too.

levkropp: @sharder996 , I believe @tobe2098 has made the decision to switch to `cmake --build` for all commands that were previously ninja, see discussion on https://github.com/canonical/multipass/pull/4354#discussion_r2338037033

sharder996: Ok, then the above command shouldn't be using `GNinja`.

tobe2098: I see what you mean @sharder996, but I found that in W11 not specifying the build system in the cmake configuring command did not work, cmake opts by default to use Make. Apart from the errors (I do not have them written down) we are/were recommending to use `ninja` in our guide, and as such I consider the flag to be necessary. Unless we want to change completely to `make` for the building process.

sharder996: Ok, thanks for the explanation!

sharder996 (approved):
lGTM!

BUILD.windows.md
```diff
@@ -1,43 +1,63 @@
...
+## Environment Setup
 
 ### Chocolatey
 
-Install chocolatey, a package manager, to download the rest of the dependencies: <https://chocolatey.org/>
+Install chocolatey, a package manager, to download the rest of the dependencies: <https://chocolatey.org/>.
 
-Press Windows Key+X and Run Windows PowerShell(Admin) then follow the chocolatey instructions to "Install with
-Powershell.exe"
+Press Windows Key+X and Run Windows PowerShell(Admin) then follow the chocolatey instructions to
```
AhiyaHiya: On Windows 10, after pressing `Windows Key + X`, the Quick Link Menu does show `Windows PowerShell (Admin)` but on Windows 11, pressing `Windows Key + X` to bring up the Quick Link Menu will show `Terminal (Admin)` instead.

tobe2098: You are correct, even though it opens a Powershell. I will issue a change.

sharder996 (approved):

PR #4356, opened 2025-09-10
[3rd-party images] Images info web scraper

sharder996:
A simple web scraper that can be run from CI on a scheduled GitHub workflow.

The scraper will overwrite `data/distributions/distribution-info.json` and if any changes are detected, a Pull Request will be created updating the file.

MULTI-2164

### DIFF
--- a/tools/distro-scraper/scraper/models.py
+++ b/tools/distro-scraper/scraper/models.py
@@ -1,4 +1,8 @@
 from pydantic import BaseModel, Field, field_validator
+from typing import Literal
+
+
+ImageArch = Literal["x86_64", "arm64"]
 
 
 class ImageItem(BaseModel, extra="forbid"):
@@ -22,23 +26,6 @@ class ScraperResult(BaseModel, extra="forbid"):
     release: str = Field(..., description="Release identifier")
     release_codename: str = Field(..., description="Human-readable release codename")
     release_title: str = Field(..., description="Release title or version")
-    items: dict[str, ImageItem] = Field(
+    items: dict[ImageArch, ImageItem] = Field(
         ..., description="Map of architecture labels to image metadata"
     )
-
-    @field_validator("items")
-    @classmethod
-    def validate_architecture_keys(
-        cls, v: dict[str, ImageItem]
-    ) -> dict[str, ImageItem]:
-        """
-        Ensure items dict only contains valid architecture keys.
-        """
-        allowed_keys = {"x86_64", "arm64"}
-        invalid_keys = set(v.keys()) - allowed_keys
-        if invalid_keys:
-            raise ValueError(
-                f"Invalid architecture keys: {invalid_keys}. "
-                f"Only {allowed_keys} are allowed."
-            )
-        return v


tools/distributions-scraper/scraper/debian.py
```diff
@@ -0,0 +1,88 @@
+import base64
+import requests
+from email.parser import Parser
+from scraper.base import BaseScraper
+
+
+def _fetch_items(codename, version):
```
ricab: This function is pretty long, and has 5 levels of nesting. I know this is only adjacent to Multipass, but it will need to be read and maintained. Would you mind splitting the code and organizing into smaller, adequately named units, please? It should then be easier to quickly understand the different parts of the procedure and how they relate to each other.

tools/distributions-scraper/scraper/fedora.py
```diff
@@ -0,0 +1,72 @@
+import requests
+from scraper.base import BaseScraper
+from dateutil import parser as dateparser
+
+
+class FedoraScraper(BaseScraper):
```
ricab: idem

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,81 @@
...
+    loop = asyncio.get_event_loop()
+    name = scraper_instance.name
+    try:
+        result = await loop.run_in_executor(executor, scraper_instance.fetch)
+        return name, result, None
+    except Exception as e:
+        return name, None, str(e)
+
+
+async def main():
```
ricab: And here, if you can. Places where section comments about what it's doing are good candidates for extraction.

ricab (commented):
Only a very superficial review, but this is all great stuff and you made very quick work of it, so well done! Just a little request for code organization in a few places.

[Secondary review]

.github/workflows/update-images.yaml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
```
jimporter: Might as well use the latest Python here... (and the fact that it's Python version π is fun :))
```suggestion
          python-version: '3.14'
```

tools/distributions-scraper/requirements.txt
jimporter, line 1: Since this scraper has multiple files, how about making a full Python project? Then the requirements just go into `pyproject.toml`, and we can use [plugin entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) (that's with setuptools, but we could use another build system) to manage the various scrapers.

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+from scraper.base import BaseScraper
+import scraper
+
+
+DEFAULT_OUTPUT_FILE = (
+    pathlib.Path(__file__).resolve().parent.parent.parent
+    / "data"
+    / "distributions"
+    / "distribution-info.json"
+)
```
jimporter: It might be simpler to pass this in as a command-line arg; then we don't need to worry if we move this script somewhere else in the repo.

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+    for _, module_name, _ in pkgutil.iter_modules(scraper.__path__):
+        if module_name == "base":
+            continue
+
+        module = importlib.import_module(f"scraper.{module_name}")
+        for attr_name in dir(module):
+            obj = getattr(module, attr_name)
+            if isinstance(obj, type) and issubclass(obj, BaseScraper) and obj is not BaseScraper:
+                scrapers.append(obj())
+    return scrapers
```
jimporter: If we use plugin entry points, this would be a lot simpler. Something like the following would work (I think):

```python
from importlib import metadata

def load_scrapers():
    return [i.load() for i in metadata.entry_points(group="dist_scraper.scrapers")]
```

... or whatever name we want for that entry point.

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def write_output_file(output, path = DEFAULT_OUTPUT_FILE):
+    """Write JSON output to the given path, creating parent directories as needed."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    json_str = json.dumps(output, indent=2, sort_keys=True) + "\n"
+    path.write_text(json_str)
```
jimporter: Not sure how large the JSON data is, but we could dump to the file directly, which might save some memory and feels more idiomatic to me:
```suggestion
    with path.open('w') as f:
        json.dump(output, f, indent=2, sort_keys=True)
        f.write('\n')
```

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+    logger.info("Output written to %s", path)
+
```
jimporter: It might be worth adding a linter for all our Python code; that'd probably error on this line (there's one blank line between functions, but Python recommends 2).

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+        if error:
+            errors[name] = error
+        elif data:
+            output[name] = data
+
+    # Only include errors if there are any
+    if errors:
+        output["errors"] = errors
+        for name, err in errors.items():
+            print(f"- {name}: {err}")
```
jimporter: Should we just use the logging facilities for this?

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+        # Write partial output for debugging
+        print(f"Output:\n{json.dumps(output, indent=2, sort_keys=True)}")
```
jimporter: Ditto here.

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+        sys.exit(1)
```
jimporter: `exit` is a built-in, so the `sys` isn't necessary:
```suggestion
        exit(1)
```

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+    write_output_file(output)
+    logger.info("All scrapers succeeded.")
+
+
+if __name__ == "__main__":
+    configure_logging()
+    try:
+        asyncio.run(main())
+    except KeyboardInterrupt:
+        logging.getLogger(__name__).info("Interrupted by user")
```
jimporter:
```suggestion
        logger.info("Interrupted by user")
```

tools/distributions-scraper/scraper/debian.py
```diff
@@ -0,0 +1,180 @@
...
+    resp.raise_for_status()
+    return resp.text
+
+
+def _fetch_json(url: str, timeout: int = DEFAULT_TIMEOUT):
+    """GET a URL and return JSON-decoded content."""
+    logger.info("Fetching Debian manifest from %s", url)
+    resp = requests.get(url, timeout=timeout)
+    resp.raise_for_status()
+    return resp.json()
```
jimporter: These could probably be async; I've usually seen `aiohttp` used for async HTTP requests.

tools/distributions-scraper/scraper/debian.py
```diff
@@ -0,0 +1,180 @@
...
+    resp = requests.head(url, allow_redirects=True, timeout=timeout)
+    resp.raise_for_status()
+    length = resp.headers.get("Content-Length")
+    if length is None:
+        return None
+    try:
+        return int(length)
+    except (TypeError, ValueError):
+        logger.warning("Invalid Content-Length header: %s", length)
+        return None
```
jimporter: Ditto here (async).

tools/distributions-scraper/scraper/debian.py
```diff
@@ -0,0 +1,180 @@
...
+
+        upload_item = _find_qcow2_upload(manifest)
+        if not upload_item:
+            logger.info("No qcow2 upload found for %s %s", codename, arch)
+            continue
+
+        metadata = upload_item.get("metadata") or {}
+        labels = metadata.get("labels") or {}
+        annotations = metadata.get("annotations") or {}
+        data = upload_item.get("data") or {}
```
jimporter:
```suggestion
        metadata = upload_item.get("metadata", {})
        labels = metadata.get("labels", {})
        annotations = metadata.get("annotations", {})
        data = upload_item.get("data", {})
```

tools/distributions-scraper/scraper/debian.py
```diff
@@ -0,0 +1,180 @@
...
+    """Fetch image manifests for known arches and build the items mapping.
+
+    Preserves original mapping and output structure.
+    """
+    arch_map = {
+        "amd64": "x86_64",
+        "arm64": "arm64",
+    }
+
+    items: Dict[str, Dict] = {}
```
jimporter: I think you can use the actual `dict` type here instead of relying on the special type-annotations `Dict` (and elsewhere for `List`, `Tuple`, etc):
```suggestion
    items: dict[str, dict] = {}
```

Or maybe adding our own type annotation for JSON. Something like [this](https://stackoverflow.com/questions/68154122/how-do-i-type-annotate-json-data-in-python) would probably work?

```python
JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None
```

tools/distributions-scraper/main.py
```diff
@@ -0,0 +1,106 @@
...
+
+async def main():
+    scrapers = load_scrapers()
+    output = {}
+    errors = {}
+
+    # Run scrapers concurrently using a thread pool
+    with ThreadPoolExecutor(max_workers=len(scrapers)) as executor:
+        tasks = [run_scraper(s, executor) for s in scrapers]
+        completed = await asyncio.gather(*tasks)
```
jimporter: If we handle requests asynchronously, I don't think we'd need to use a thread pool here. As far as I can tell, the only thing we need/want concurrency for is network requests.

jimporter (changes requested):
Most of this looks good. The main things I noticed that would help here would be making this into a Python project via `pyproject.toml` and using an async requests library like `aiohttp`. We might also want to add some Python linters/type-checkers to CI so we can catch any small style problems.

tools/distributions-scraper/main.py
sharder996: What would you suggest? Linting all python code in the Multipass project or only the diff?

tools/distro-scraper/README.md
```diff
@@ -0,0 +1,94 @@
...
+### Running as a Python module:
+```bash
+python -m scraper <output_file>
+```
+
+The scraper will:
+1. Load all registered scraper plugins
+2. Fetch image metadata from each distribution concurrently
+3. Validate the output against the expected schema
+4. Write the combined results to the specified JSON file
```
jimporter: Maybe put this near the top, since it's a good summary of how the script works?

.github/workflows/update-images.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.14'
+
+      - name: Create virtual environment
+        run: |
+          python -m venv .venv
+          source .venv/bin/activate
+          echo PATH=$PATH >> $GITHUB_ENV
```
jimporter, line 30: Is this necessary? I don't usually bother setting up a venv in GH actions. I imagine this works, but I'd have to think more about whether there are other env vars we need to forward along beyond `$PATH`...

tools/distro-scraper/scraper/models.py
```diff
@@ -0,0 +1,48 @@
...
+        if invalid_keys:
+            raise ValueError(
+                f"Invalid architecture keys: {invalid_keys}. "
+                f"Only {allowed_keys} are allowed."
+            )
+        return v
+
+    class Config:
+        # Reject extra fields to catch errors
+        extra = "forbid"
```
jimporter: This syntax is [deprecated](https://docs.pydantic.dev/2.12/concepts/config/):
> In Pydantic V1, the Config class was used. This is still supported, but deprecated.

Instead...

tools/distro-scraper/scraper/models.py
```diff
@@ -0,0 +1,48 @@
...
+    Schema for an individual architecture's image metadata.
+    """
+
+    image_location: str = Field(..., description="URL to the image file")
+    id: str = Field(..., description="Hash/checksum of the image")
+    version: str = Field(..., description="Version string of the image")
+    size: int = Field(..., description="Size of the image in bytes")
+
+
+class ScraperResult(BaseModel):
```
jimporter: ... how about this?

```suggestion
class ScraperResult(BaseModel, extra="forbid"):
```

tools/distro-scraper/scraper/models.py
```diff
@@ -0,0 +1,48 @@
+from pydantic import BaseModel, Field, field_validator
+
+
+class ImageItem(BaseModel):
```
jimporter: Should this forbid extra values too?

```suggestion
class ImageItem(BaseModel, extra="forbid"):
```

tools/distro-scraper/scraper/models.py
```diff
@@ -0,0 +1,48 @@
...
+        Ensure items dict only contains valid architecture keys.
+        """
+        allowed_keys = {"x86_64", "arm64"}
+        invalid_keys = set(v.keys()) - allowed_keys
+        if invalid_keys:
+            raise ValueError(
+                f"Invalid architecture keys: {invalid_keys}. "
+                f"Only {allowed_keys} are allowed."
+            )
+        return v
```
jimporter: Instead of this, you could define an enum for the key and then Pydantic does all the validation work for you. Given this:

```python
from enum import Enum


class ImageArch(Enum):
    x86_64 = "x86_64"
    arm64 = "arm64"
```

Then you can do:

```suggestion
    items: dict[ImageArch, ImageItem] = Field(
        ..., description="Map of architecture labels to image metadata"
    )
```

tools/distro-scraper/scraper/scrapers/debian.py
```diff
@@ -0,0 +1,218 @@
...
+        Returns a dict with keys: Version, Codename
+        """
+        parser = Parser()
+        parsed = parser.parsestr(content)
+        version = parsed.get("Version")
+        codename = parsed.get("Codename")
+        self.logger.info(
+            "Parsed Release file: Version=%s, Codename=%s", version, codename
+        )
+        return {"Version": version, "Codename": codename}
```
jimporter: Since the parsed `Message` already has a dict-like interface (and we're returning a dict), we could return the `parsed` message directly. Or maybe we should return the version and codename as a tuple so that the caller can get at those more directly (which would also help make this easier to statically type-check).

tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Map Fedora architecture names to the labels used in our items output.
+
+        Returns None for unsupported arches.
+        """
+        if arch == "aarch64":
+            return "arm64"
+        if arch == "x86_64":
+            return "x86_64"
+        return None
```
jimporter:
```suggestion
        raise ValueError(f"unsupported arch {arch}")
```

tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -0,0 +1,165 @@
...
+
+            # Filter out unsupported architectures first
+            supported_images = []
+            for img in latest_images:
+                arch = img.get("arch")
+                label = self._map_arch_label(arch)
+                if label:
+                    supported_images.append((img, label))
+                else:
+                    self.logger.info("Skipping unsupported architecture: %s", arch)
```
jimporter:
```suggestion
                try:
                    supported_images.append((img, self._map_arch_label(arch)))
                except ValueError:
                    self.logger.info("Skipping unsupported architecture: %s", arch)
```

tobe2098 (approved):
LGTM, good work! Although I agree with @jimporter 's comments

.github/workflows/update-images.yaml
sharder996: It just avoids warnings (an maybe errors) about installing system-wide packages and having to use `--break-system-packages`. An attempt to just make the setup cleaner.

tools/distro-scraper/scraper/models.py
```diff
@@ -1,7 +1,12 @@
 from pydantic import BaseModel, Field, field_validator
+from enum import Enum
```
jimporter: Nit: remove this.

sharder996: Whoops 😆

jimporter (approved):

PR #4358, opened 2025-09-11
Solve error from suspended instances on GUI close

### LINKED ISSUE #4200: Suspended instances cause GUI to fail to stop instances and quit
**Describe the bug**
When I try to close the GUI and choose to "Stop instances", it will fail to stop any suspended instance and consequently fail to close.

Image

**To Reproduce**

1. `multipass launch -n a`
1. `multipass launch -n b`
1. `multipass suspend b`
1. Open GUI
2. Configure GUI to Stop instances in settings (or ask)
3. Try to close the GUI (choose to stop instances if set to ask)

**Expected behavior**
I expected the GUI not to even attempt to stop suspended instances.

**Additional info**
- OS: Linux
- CPU: x86
- `multipass version`: 1.15.1

**Additional context**
The issue does not occur if there are no running instances, even if there are suspended ones. This offers a workaround: manually stopping running instances first.

tobe2098:
Raised in #4200, the GUI had a bug relating to stopping instances. Similar to `multipass stop --all`, if the GUI is configured to close all running instances or the user selects that option in the closing prompt, the GUI will attempt to close all instances before closing the GUI itself. This will cause an error and prevent the GUI from closing if there is at least one running instance and at least one suspended instances. If at least one suspended instance is lexicographically sorted behind the running instance the GUI cannot be closed with the option of "closing all running instances".

To solve that, the GUI now only requests to close the instances that are actually running.

Then `multipass_tests` was run and all tests passed.

Fixes #4200 

MULTI-2234

### DIFF
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -184,9 +184,15 @@ class _AppState extends ConsumerState<App> with WindowListener {
     }
 
     stopAllInstances() {
+      final runningVMs = ref
+          .read(vmStatusesProvider)
+          .entries
+          .where((entry) => entry.value == Status.RUNNING)
+          .map((entry) => entry.key)
+          .toList();
       final notificationsNotifier = ref.read(notificationsProvider.notifier);
       notificationsNotifier.addOperation(
-        ref.read(grpcClientProvider).stop([]),
+        ref.read(grpcClientProvider).stop(runningVMs),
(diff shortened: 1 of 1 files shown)


sharder996:
Hi @tobe2098, quick tip.

If you add the text "Fixes #{issue number}", then GitHub will automatically link this PR with the issue and close the issue when this gets merged in.

Issue #4209, opened 2025-07-02
Multipass Upgrade (e.g. to 1.16.0 RC3, and likewise RC4) keeps failing on Windows — "Multipass Service ... could not be stopped"


holta:
This happens almost every single time — e.g. when trying to upgrade to 1.16.0 RC3 or [1.16.0 RC4](https://github.com/canonical/multipass/releases/tag/v1.16.0-rc4) on Windows 11 Pro — despite Multipass having no VMs (no instances) at all:

![Image](https://github.com/user-attachments/assets/e6c0fed1-d3eb-4d77-b3d2-7c78a30475ee)

![Image](https://github.com/user-attachments/assets/b6abcab2-b5c7-4d20-aab0-7786d7ec6252)

Would you have any suggestions or recommendations to get past this?  _Thank you!_

Sploder12:
Hi @holta, sorry for all the issues. Could you please send your logs? [here](https://documentation.ubuntu.com/multipass/en/latest/how-to-guides/troubleshoot/access-logs/) is a guide on how to access them.

holta:
@Sploder12 

Windows logs for today attached — please remove the .TXT suffix: 

[multipass-event-log-2025-07-02.evtx.TXT](https://github.com/user-attachments/files/21022785/multipass-event-log-2025-07-02.evtx.TXT)

> Logs from the installation and uninstall process can be found under %APPDATA%\Local\Temp. Sort the contents of the directory by “Date Modified” to bring the newest files to the top. The name of the file containing the logs follows the pattern MSI[0-9a-z].LOG.

PART TWO: Just FYI when trying to get Multipass install and uninstall logs, the instructions ( pasted in above, from the 2nd paragraph of https://documentation.ubuntu.com/multipass/en/latest/how-to-guides/troubleshoot/access-logs/ ) do not work.

1. When pasting `%APPDATA%\Local\Temp` into Windows' "File Explorer", it responds:

```
Windows can't find 'C:\Users\USERNAME\AppData\Roaming\Local\Temp'. Check the spelling and try again.
```

2. When pasting `%APPDATA%\Local` into Windows' "File Explorer", it responds:

```
Windows can't find 'C:\Users\USERNAME\AppData\Roaming\Local'. Check the spelling and try again.
```

3. Whereas %APPDATA% == `C:\Users\USERNAME\AppData\Roaming` does in fact exist.

holta:
[...]

PR #4359, opened 2025-09-11
Fix Windows installer logs path

Sploder12:
Previously the documentation had the path of the install logs as `%APPDATA%\Local\Temp`, this by default resolves to `C:\Users\<user>\AppData\Roaming\Local\Temp` which is not the correct path. This PR uses `%TEMP%` which by default resolves to `C:\Users\<user>\AppData\Local\Temp` which is where the logs are stored*.

Thanks to @holta for pointing out this issue in #4209

*as pointed out by @xmkg, logs may not always be in the `%TEMP%` folder

### DIFF
--- a/docs/how-to-guides/troubleshoot/access-logs.md
+++ b/docs/how-to-guides/troubleshoot/access-logs.md
@@ -37,7 +37,7 @@ The Multipass GUI produces its own logs, that can be found under `~/Library/Appl
 
 On Windows, the Event system is used and Event Viewer lets you access them. Our logs are currently under "Windows Logs/Application", where you can filter by "Multipass" Event source. You can then export the selected events to a file.
 
-Logs from the installation and uninstall process can be found under `%APPDATA%\Local\Temp`. Sort the contents of the directory by "Date Modified" to bring the newest files to the top. The name of the file containing the logs follows the pattern `MSI[0-9a-z].LOG`.
+Logs from the installation and uninstall process can be found under `%TEMP%`. Sort the contents of the directory by "Date Modified" to bring the newest files to the top. The name of the file containing the logs follows the pattern `MSI[0-9a-z].LOG`.
 
 The Multipass GUI produces its own logs, that can be found under `%APPDATA%\com.canonical\Multipass GUI\multipass_gui.log`
 


docs/how-to-guides/troubleshoot/access-logs.md
```diff
@@ -37,7 +37,7 @@ The Multipass GUI produces its own logs, that can be found under `~/Library/Appl
 
 On Windows, the Event system is used and Event Viewer lets you access them. Our logs are currently under "Windows Logs/Application", where you can filter by "Multipass" Event source. You can then export the selected events to a file.
 
-Logs from the installation and uninstall process can be found under `%APPDATA%\Local\Temp`. Sort the contents of the directory by "Date Modified" to bring the newest files to the top. The name of the file containing the logs follows the pattern `MSI[0-9a-z].LOG`.
+Logs from the installation and uninstall process can be found under `%LOCALAPPDATA%\Temp`. Sort the contents of the directory by "Date Modified" to bring the newest files to the top. The name of the file containing the logs follows the pattern `MSI[0-9a-z].LOG`.
```
xmkg: So... This one is tricky because the log path actually depends on the system configuration, the installation mode (system vs. user), and the explicit log path override.  Our Windows installer explicitly enables the MSI logging by setting the [MsiLogging](https://learn.microsoft.com/en-us/windows/win32/msi/msilogging) property:

```
 <Property Id="MsiLogging" Value="vx!" />
```
`vx!` denotes `verbose`, `extra debugging info`, and `force flush per line`, respectively.

[MsiLogFileLocation](https://learn.microsoft.com/en-us/windows/win32/msi/msilogfilelocation) is the read-only property where the actual logs are stored. The best way to solve this would be to display the log path when installation has failed or been cancelled. A lesser `good` way to do it would be replacing the `%LOCALAPPDATA%\Temp` with simply `%TEMP%`.  I think we can use `%TEMP%` for now. @Sploder12 opinions?

Sploder12: Well that makes things complicated... `%TEMP%` should work for now. Displaying a message would be nice but that's harder to represent in the docs I feel, it'd be nice to have it in a consistent location. Maybe move/copy it into `%TEMP%` if it isn't already there? Honestly not sure what the best solution is for this.

sharder996:
Something up with the Documentation Check. Clearly not related to changes here so merging manually.

PR #4024, opened 2025-04-03
add deprecation warnings for lxd and libvirt related commands

georgeliao:
[MULTI-1887](https://warthogs.atlassian.net/browse/MULTI-1887)
[MULTI-1892](https://warthogs.atlassian.net/browse/MULTI-1892)

First, about the warning message, I reused the message from hyperkit. It alludes the migration we want to provide. Are we certain that we will offer a migration plan? Otherwise we need to tune warning message a bit. 

For functional testing, **only** the following operations should trigger the warning message
`multipass set local.driver=lxd/libvirt`
`multipass list`, `multipass launch`, `multipass info` when the current driver is lxd or libvirt

[MULTI-1887]: https://warthogs.atlassian.net/browse/MULTI-1887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MULTI-1888]: https://warthogs.atlassian.net/browse/MULTI-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MULTI-1892]: https://warthogs.atlassian.net/browse/MULTI-1892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MULTI-1893]: https://warthogs.atlassian.net/browse/MULTI-1893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4365, opened 2025-09-17
Remove support for libvirt as multipass backend

### LINKED ISSUE #4364: Remove libvirt backend
This issue is here to track the necessary changes to remove libvirt as a Multipass backend.

## Code to edit
- multipass/src/platform/backends/libvirt/*
	- src/platform/backends/libvirt_virtual_machine_factory.cpp delete
	- src/platform/backends/libvirt_virtual_machine_factory.h delete
	- src/platform/backends/libvirt_virtual_machine.cpp delete
	- src/platform/backends/libvirt_virtual_machine.h delete
	- src/platform/backends/libvirt_wrapper.h delete
	- src/platform/backends/libvirt_wrapper.cpp delete
- multipass/tests/linux/test_platform_linux.cpp
	- l.84-88
	- l.139-143
	- l.150-155 Purpose of the test is not clear
	- l.157-161 Purpose of the test is not clear
		- Edit LIBVIRT for another backend driver?
	- Add libvirt to Unsupported Drivers tests?
- multipass/tests/libvirt/\*
	- tests/libvirt/broken_libvirt_library.cpp delete
	- tests/libvirt/fake_libvirt_library.cpp delete
	- tests/libvirt/test_libvirt_backend.cpp delete
- multipass/src/platform/platform_linux.cpp
	- l.272
	- l.33
	- l.414-415
- CMAKE:
	- src/platform/backends/libvirt/CMakeLists.txt delete
	- tests/libvirt/CMakeLists.txt delete
	- CMakeLists.txt
		- l.337
- .github/workflows/linux.patch
	- l.29-49
- debian/control
	- l.9
- include/multipass/utils.h
	- l.345 Deprecation messages
- src/utils/utils.cpp
	- l.638-658 Deprecation messages
- snap/snapcraft.yaml
	- l.62
	- l.109-122
	- l.147
- src/client/cli/cmd/set.cpp
	- l.103 (deprecation warnings)
- src/daemon/daemon.cpp
	- l.1381 (deprecation warnings)
- tests/test_cli_client.cpp
	- l.3591-3604 
- tests/test_daemon_launch.cpp
	- l.45 Relation to libvirt and LXD not clear
- tests/test_daemon.cpp
	- l.145 Relation to libvirt and LXD not clear

## Docs changes
- BUILD.linux
	- l.58
	- l.67
- docs/explanation/driver.md
	- l.5
	- l.20
	- l.46
	- l.60
- docs/how-to-guides/customise-multipass/set-up-the-driver.md
	- l.38-67
	- l.111
	- l.374
- docs/reference/settings/local-driver.md
	- l.16

tobe2098:
The pull request removes from the multipass repo all components related to libvirt. This is part of the effort to restrict multipass to a single backend per platform.

There may still be driver specific logic that has been missed in the initial scan over the code.

Fixes #4364.

MULTI-1891


src/platform/backends/qemu/CMakeLists.txt
tobe2098, line 1: This should already be there, since daemon should link with libqemu.a. This absence caused problems when the CMakeLists.txt from libvirt, which contained daemon, was removed.

tests/linux/test_platform_linux.cpp
```diff
@@ -136,24 +128,11 @@ TEST_F(PlatformLinux, testDefaultDriverProducesCorrectFactory)
...
 }
 
-TEST_F(PlatformLinux, testQemuInEnvVarIsIgnored)
-{
-    mpt::SetEnvScope env(mp::driver_env_var, "QEMU");
-    auto test = [this] { aux_test_driver_factory<mp::LibVirtVirtualMachineFactory>("libvirt"); };
-    with_minimally_mocked_libvirt(test);
-}
-
 TEST_F(PlatformLinux, testLibvirtInEnvVarIsIgnored)
```
tobe2098, line 136: It is not clear to me the purpose of this test beyond the removal of a previously existing qputenv() framework, since qgetenv() never gets called with mp::driver_env_var in the codebase apart from here.
I left the other test in case it is important for us to check that the env var is ignored.

tests/test_cli_client.cpp
```diff
@@ -3588,21 +3588,6 @@ TEST_F(Client, setCmdRejectsBadValues)
...
-    EXPECT_CALL(mock_settings, set(Eq(mp::petenv_key), Eq("")));
-    EXPECT_THAT(send_command({"set", keyval_arg(mp::petenv_key, "")}), Eq(mp::ReturnCode::Ok));
-
-    EXPECT_CALL(mock_settings, set(Eq(mp::petenv_key), Eq("some primary")));
-    EXPECT_THAT(send_command({"set", keyval_arg(mp::petenv_key, "some primary")}),
-                Eq(mp::ReturnCode::Ok));
-}
-
-#endif // MULTIPASS_PLATFORM_LINUX
-
```
tobe2098, line 3607: It is not clear why this concerns specific libvirt to qemu switching

tests/test_daemon.cpp
```diff
@@ -142,7 +142,7 @@ a few more tests for `false`, since there are different portions of code dependi
         EXPECT_CALL(mock_settings, get(Eq(mp::bridged_interface_key)))
             .WillRepeatedly(Return("eth8"));
         EXPECT_CALL(mock_settings, get(Eq(mp::driver_key)))
-            .WillRepeatedly(Return("qemu")); // TODO lxd and libvirt migration, remove
+            .WillRepeatedly(Return("qemu")); // TODO lxd migration, remove
```
tobe2098, line 147: Not clear why this should be removed beyond the fact that once qemu is the only driver driver_key will have limited utility

tests/test_cli_client.cpp
ricab: Hey Antoni, I got curious so I had a look at git history and this was spurious. There were other tests within this ifdef before this one was added, presumably because it looked similar. Those other tests were later removed and only this one remained. 

In short, the test is unrelated to `libvirt`, it is still valid, and it should not have been within this `ifdef`. Could you please remove only the `ifdef` instead?

tobe2098: Makes sense, I expected something along those lines -> comments being outdated/wrong. Will do the suggested changes

tobe2098 (commented):
Tested locally both the build and the snap, both seem to work fine. All google tests also passed.

tobe2098:
Additionally to all the changes directly related to libvirt I have found the following backend-dependant functions that could warrant a change/adaptation when LXD is removed:
- In BaseVirtualMachineFactory:
  -  require_clone_support throws by default in virtual function, only not implemented in lxd and libvirt, otherwise does nothing. Can be removed entirely once both are gone.
  - require_suspend_support does nothing by default in virtual function and only implemented in lxd to throw. Can be removed entirely once lxd is gone.
  - require_snapshot_support throws by default in virtual function, only not implemented in lxd and libvirt, otherwise does nothing. Can be removed entirely once both are gone.
  - clone_vm_impl is the same but the derived virtual functions do not have empty body. Cannot be removed.
- In BaseVirtualMachine:
  - require_snapshots_support, same as above.
  - add_network_interface (not impl by libvirt) and make_native_mount (not impl by vbox and libvirt) throw by default but have implementations in the other backends, so cannot be removed.
  
Should I add a comment in the removable functions post LXD removal?

ricab:
Yes please @tobe2098. Well spotted!

ricab:
Hey Antoni, I noticed `src/platform/backends/shared/qemu_img_utils/` while browsing code and started wondering: is anything in there used anywhere other than QEMU and libvirt? Possibly LXD? Would you mind having a look, to see if we could consolidate into just QEMU? Thanks!

tobe2098:
On it

> Hey Antoni, I noticed `src/platform/backends/shared/qemu_img_utils/` while browsing code and started wondering: is anything in there used anywhere other than QEMU and libvirt? Possibly LXD? Would you mind having a look, to see if we could consolidate into just QEMU? Thanks!

Both LXD and libvirt (ofc also QEMU) use the functions on that folder. 
There is one particular function that is used by libvirt and qemu only, which could be moved in this PR to the qemu backend, but I think that we should move all of them once we remove LXD, since the rest will have to be moved then.

ricab:
OK got it, that makes sense. I wouldn't object doing them together when removing LXD if you prefer. Up to you. You can also add TODO tags in the code if you think that's helpful.

docs/explanation/driver.md
```diff
@@ -43,7 +43,7 @@ Nonetheless, instances are preserved across drivers. After switching back to a p
 
 There are two exceptions to the above:
 
-  - On Linux, QEMU and libvirt (deprecated) share the same driver scope.
+  <!-- - On Linux, QEMU and libvirt (deprecated) share the same driver scope. -->
```
sharder996: I think this should be removed, no?

src/client/cli/cmd/set.cpp
```diff
@@ -100,8 +100,8 @@ mp::ParseCode cmd::Set::parse_args(mp::ArgParser* parser)
                 val = keyval.at(1);
 
 #ifdef MULTIPASS_PLATFORM_LINUX
-                // TODO lxd and libvirt migration, remove
-                if (key == mp::driver_key && (val == "lxd" || val == "libvirt"))
+                // TODO lxd migration, remove
+                if (key == mp::driver_key && (val == "lxd"))
```
sharder996:
```suggestion
                if (key == mp::driver_key && val == "lxd")
```

We could lose the extra parentheses.

tests/test_daemon.cpp
sharder996: https://github.com/canonical/multipass/pull/4024#discussion_r2053691494

src/platform/backends/qemu/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_library(qemu_backend STATIC
   qemu_virtual_machine.cpp)
 
 target_link_libraries(qemu_backend
+  daemon
```
sharder996, line 29: Why was this addition needed?

sharder996 (changes requested):
Looks good! Just a few nitpicks

src/platform/backends/qemu/CMakeLists.txt
tobe2098: Yes, if not added qemu does not compile. The link library is also present in the CMakeLists of LXD (or should be) and was present in the CMakeLists of libvirt.

tobe2098: Sorry, misread without the 'why'. It is needed  because the daemon lib has to be built before the qemu backend goes into linking, or the compilation fails. It was previously not failing because that library was present in the CMakeLists of libvirt, and as such was compiled before qemu started compilation, managing to fulfill the requirements for linking accidentally.

tests/test_daemon.cpp
tobe2098: Thanks!

src/platform/backends/qemu/CMakeLists.txt
sharder996: Thanks for the explanation!

sharder996 (approved):

PR #4366, opened 2025-09-17
[tics] Notify Multipass team on workflow failure

sharder996:
This PR adds a step to the TICS workflow that sends a notification to the [Multipass](https://chat.canonical.com/canonical/channels/multipass) Mattermost channel when the workflow fails. Below is an example message that would be sent to Mattermost:

```
🔴 @mp
Scheduled [TICS](https://github.com/canonical/multipass/actions/runs/17801613177) workflow failed.
```

MULTI-2240

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -88,7 +88,7 @@ jobs:
         installTics: true
 
     - name: Report workflow failure
-      if: ${{ failure() || cancelled() }}
+      if: ${{ !success() }}
       uses: mattermost/action-mattermost-notify@master
       with:
         MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL }}


.github/workflows/tics.yml
```diff
@@ -86,3 +86,15 @@ jobs:
         viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
         ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
         installTics: true
+
+    - name: Report workflow failure
+      if: ${{ failure() }}
```
ricab: Should this include `cancelled()`? That would at least help with confirming this works, after merging.

ricab (commented):
Looks good Scott 👍 just a question inline

.github/workflows/tics.yml
sharder996: Should the message be reworded then? Something like `Scheduled TICS workflow exited before completing`?

ricab: Yep, probably a good idea. We can go back to only failure if we get tired.

.github/workflows/tics.yml
```diff
@@ -86,3 +86,15 @@ jobs:
...
+      uses: mattermost/action-mattermost-notify@master
+      with:
+        MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
+        MATTERMOST_CHANNEL: multipass
+        TEXT: |
+          :red_circle: @mp
+          Scheduled [TICS](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) workflow failed.
+        MATTERMOST_USERNAME: ${{ github.triggering_actor }}
```
xmkg, line 99: So in the actual Mattermost notification, I see a real person's handle instead of a bot. Is that expected?

sharder996: The message that is in Mattermost was generated from a workflow that I, myself triggered. I'm not sure if the handle will change for a scheduled workflow that was triggered by GitHub. I will ask around, though.

sharder996: I did some asking around in ~IS, but nobody was able to tell me why `MATTERMOST_USERNAME` is not being respected. Why the real person's handle is showing up is because that is the person who created the webhook for me.

xmkg: The [Mattermost documentation](https://developers.mattermost.com/integrate/webhooks/incoming/#parameters) about the webhooks says the following:

```
username	Overrides the username the message posts as.
Defaults to the username set during webhook creation; if no username was set during creation, webhook is used.
The [Enable integrations to override usernames](https://docs.mattermost.com/configure/configuration-settings.html#enable-integrations-to-override-usernames) configuration setting must be enabled for the username override to take effect.
```

Could it be that the "Enable integrations to override usernames" thingy is not enabled for this webhook?

.github/workflows/tics.yml
```diff
@@ -86,3 +86,15 @@ jobs:
         viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
         ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
         installTics: true
+
+    - name: Report workflow failure
+      if: ${{ failure() || cancelled() }}
```
ricab: And then why not just `!success()`?

sharder996: Yes, that would simplify it.

.github/workflows/tics.yml
sharder996: Perhaps... I will pass it on to ~IS and see what they say.

sharder996:
Current [status](https://portal.admin.canonical.com/C181065/) from IS on creating a bot account for webhooks:

> We need to dig up the documentation for this and migrate to RTD alongside doing this ticket

PR #4368, opened 2025-09-18
Mock sleep in wait-ready tests

ricab:
Use Utils singleton to sleep wait-ready and mock the call in tests. This allows verifying the call and shaving a second off our tests.


Sploder12 (approved):
LGTM!

I wonder if something similar could be done for the `TimeoutSuite` tests 🤔

ricab:
Perhaps. I have had those in my local exclude list for a long time.

Thanks for the reviews!

PR #4372, opened 2025-09-19
[macos] Copy libdart_ffi dylib to Frameworks dir

levkropp:
This is the MacOS equivalent of #4212 that copies the libdart_ffi.dylib file to the app bundle's Frameworks directory after the Flutter build is complete in order to avoid GUI errors on startup because the GUI cannot find the libdart_ffi library.

With this change, the shortcut to the Multipass.app Application in the build/bin folder is now usable right after a successful GUI build.

The reason for the two copy operations is because the MacOS build creates two separate Multipass.app bundles in different directories during the build process. One at `build/bin/macos/Build/Products/Release/Multipass.app` and one at `build/build/bin/macos/Build/Products/Release/Multipass.app`

### DIFF
--- a/src/client/gui/CMakeLists.txt.macos
+++ b/src/client/gui/CMakeLists.txt.macos
@@ -42,7 +42,18 @@ add_custom_command(OUTPUT ${MULTIPASS_GUI_EXECUTABLE_FULL_PATH}
 )
 
 add_custom_target(${MULTIPASS_GUI_EXECUTABLE} ALL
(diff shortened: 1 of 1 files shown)


levkropp:
Closing this for now as the copying of libdart_ffi.dylib does not provide a use in MacOS (launching the daemon locally instead of installing the package does not allow for a connection between the GUI and the daemon on MacOS)

This could perhaps be revisited for pure GUI development/iteration improvements if it is possible for the built GUI bundle to see the installed multipassd daemon without having to repackage and reinstall it, but this needs to be tested and still provides only a minimal value

PR #4377, opened 2025-09-19
[format] Add formatting rules to `.editorconfig`

jimporter:
This is just a first pass at copying some of our style rules into a [`.editorconfig`](https://editorconfig.org/) file so that (hopefully) everyone's favorite editors will see these rules and set the various options appropriately. There are probably more rules we could set, but this should cover the basics.

If you're an Emacser like me, then you can add this to your `.dir-locals.el` to enable editorconfig support in the Multipass repo, if it's not already enabled globally in your config: `((nil . ((mode . editorconfig))))`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,44 @@
+root = true
+
+[CMakeLists.txt]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{cpp,h}]
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 100
+trim_trailing_whitespace = true
+
+[*.dart]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.{json,json5}]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+indent_style = space
+insert_final_newline = true
+
+[*.py]
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 79
+trim_trailing_whitespace = true
+
+[*.{yaml,yml}]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_size = 4
+max_line_length = 100
+
+[*.dart]
+indent_style = space
+indent_size = 2
+
+[*.{yaml,yml}]
+indent_style = space
+indent_size = 2
```
sharder996, line 41: A couple other relevant file types are `json`, `xml`, and `py`

sharder996: You should also be able to other rules that align with our style as well, i.e., `trim_trailing_whitespace` and `insert_final_newline`

jimporter: I added `trim_trailing_whitespace` and `insert_final_newline` as global settings for all files since they should be pretty universal, and then also added rules for `json` and `py`. I skipped `xml`, since `find . -name "*.xml" -not -path "./3rd-party/*"` returned no results (though I could add it in anyway if you want).

sharder996: We don't have any files with the `xml` file extension, but all the Windows installer code (`.wxs`) under `packaging/windows` is `xml` formatted code. There are a couple other extensions there too that are `xml` formatted: `wxl`, `wixproj`, `xsl`, `vcxproj`.

xmkg (dismissed):

.editorconfig
```diff
@@ -0,0 +1,28 @@
...
+
+[*.py]
+indent_style = space
+indent_size = 4
+max_line_length = 79
+
+[*.{yaml,yml}]
+indent_style = space
+indent_size = 2
```
sharder996, line 41: > We don't have any files with the `xml` file extension, but all the Windows installer code (`.wxs`) under `packaging/windows` is `xml` formatted code. There are a couple other extensions there too that are `xml` formatted: `wxl`, `wixproj`, `xsl`, `vcxproj`.

Adding this here again just in case it got missed in the previous resolved comment.

jimporter: I looked into these and they have differing indent widths (the Windows installer stuff is 4 spaces and the `vcxproj` file is 2). Since 4 seems to be the most common, I set that as the default for all files, and then just override the width as needed for languages like Dart and YAML.

sharder996 (changes requested):
CMake and Markdown files are 2 space indent.

jimporter:
> CMake and Markdown files are 2 space indent.

Ok, I added a rule for `CMakeLists.txt`.

Are you sure about the Markdown indent width though? My understanding is that it needs to use 4-space indents for various things to work, like [indented code blocks](https://daringfireball.net/projects/markdown/syntax#precode) and subsequent paragraphs in lists.

sharder996:
> Are you sure about the Markdown indent width though? My understanding is that it needs to use 4-space indents for various things to work, like [indented code blocks](https://daringfireball.net/projects/markdown/syntax#precode) and subsequent paragraphs in lists.

Not 100% sure, but I noticed in [README.md](https://github.com/canonical/multipass/blob/cec7758ed6998633a56bdcc538d71ad119d97d0a/README.md?plain=1#L28), 2 spaces being used to align with a list of bullet points.

jimporter:
> > Are you sure about the Markdown indent width though? My understanding is that it needs to use 4-space indents for various things to work, like [indented code blocks](https://daringfireball.net/projects/markdown/syntax#precode) and subsequent paragraphs in lists.
> 
> Not 100% sure, but I noticed in [README.md](https://github.com/canonical/multipass/blob/cec7758ed6998633a56bdcc538d71ad119d97d0a/README.md?plain=1#L28), 2 spaces being used to align with a list of bullet points.

It looks like most of the indenting we use in Markdown is "visual" indenting to line things up under some other line. Numbered lists thus get 3 spaces of indent: https://github.com/canonical/multipass/blob/cec7758ed6998633a56bdcc538d71ad119d97d0a/docs/how-to-guides/install-multipass.md?plain=1#L75-L85

sharder996 (approved):
Ok, I'll give this a tentative approval. I'm a little concerned about having a globally defined indent size and that it might mess with people default configs in their text editors. WDYT @xmkg?

jimporter:
> Ok, I'll give this a tentative approval. I'm a little concerned about having a globally defined indent size and that it might mess with people default configs in their text editors. WDYT @xmkg?

I'd also be ok with dropping that and going back to only specifying configuration for specific formats. A common default for all files in our repo means we all have consistent configs in our editors regardless of the file though. It depends on whether we'd rather err on the side of everyone having the same settings, or on the side of leaving the rules for "unknown" file types up to each user.

(In Emacs at least, the `*` wildcard would override my global default configuration (I use 2 spaces for pretty much everything), but I could in turn override the project's `.editorconfig` rules by adding whatever setting I want to my `.dir-locals.el`.)

sharder996:
Yeah, I would rather go with explicitly specifying rules for each file type and inevitably miss a few types than ignorantly override people's global settings. People can be particular about their editor customizations (i.e. pressing `<tab><tab>` to get 4 spaces)and I'd rather not interrupt their workflow and mess with that.

xmkg (approved):

sharder996 (approved):

PR #4381, opened 2025-09-22
[macos] IP address validation

sharder996:
This PR adds some validation to the process that resolves IP addresses from MAC addresses on macOS. Due to past changes made by Apple, we now use ARP to resolve IP addresses. The ARP table is, however, a cache that can potentially contain old and stale IP addresses before they are purged.

For launching Ubuntu Core images, this creates several problems. The initialization of an Ubuntu Core image is completed in two phases which is separated by the guest rebooting itself. Networking is partially initialized during the first phase, and macOS assigns it an IP address. During the second phase after the guest reboots, guest networking finishes initializing causing macOS to assign it another IP address. Now, ARP contains two IP addresses for the same MAC address.

To solve this, I have added a couple countermeasures:
1. Before returning an IP address, ping the address looking for a response signalling the machine is ready for communication.
2. If there are multiple IP addresses for the same MAC address, iterate over all IP's pinging each one and return the address who delivers a ping response.

Addresses #3541 

---

MULTI-2242

### DIFF
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -20,6 +20,7 @@
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
 #include <multipass/process/simple_process_spec.h>
+#include <shared/shared_backend_utils.h>
 
 namespace mp = multipass;
 
@@ -63,16 +64,6 @@ QString get_arp_output()
     return QString{arp_process->read_all_standard_output()};
 }
 
-bool ping_ip(const mp::IPAddress& ip_addr)
-{
-    constexpr auto timeout = 500;
-    const auto arp_process = mp::platform::make_process(
-        mp::simple_process_spec("ping", {"-c", "1", QString::fromStdString(ip_addr.as_string())}));
-    const auto arp_exit_state = arp_process->execute(timeout);
-
-    return arp_exit_state.completed_successfully();
-}
-
 } // namespace
 
 std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& mac_address)
@@ -108,7 +99,8 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
                        current_ip.as_string(),
                        mac_address);
 
-            if ((!best_match.has_value() || current_ip > best_match.value()) && ping_ip(current_ip))
+            if ((!best_match.has_value() || current_ip > best_match.value()) &&
+                mp::backend::can_reach_gateway(current_ip.as_string()))
             {
                 best_match = current_ip;
             }
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -17,6 +17,7 @@
 
 #include "backend_utils.h"
 
+#include <multipass/logging/log.h>
 #include <multipass/platform.h>
 #include <multipass/process/simple_process_spec.h>
 
@@ -24,6 +25,8 @@ namespace mp = multipass;
 
 namespace
 {
+constexpr auto category = "backend-utils";
+
 QString simplify_mac_address(const QString& input_mac_address)
 {
     // Trim the (first) leading 0 of each segment of the a MAC address.
@@ -99,6 +102,11 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
         if (match.captured(2) == arp_format_mac_address)
         {
             mp::IPAddress current_ip{match.captured(1).toStdString()};
+
+            mpl::trace(category,
+                       "Found ip match \'{}\': for mac address: \'{}\'",
+                       current_ip.as_string(),
+                       mac_address);
 
             if ((!best_match.has_value() || current_ip > best_match.value()) && ping_ip(current_ip))
             {
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -73,22 +73,29 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     // ? (192.168.64.255) at ff:ff:ff:ff:ff:ff on bridge100 ifscope [bridge]
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
 
-    const QString arp_ouput_stream = get_arp_output();
+    const QString arp_output = get_arp_output();
     const QRegularExpression ip_and_mac_address_pair_regex(R"(\(([^)\s]+)\) at ([^\s]+))");
-    QRegularExpressionMatchIterator iter =
-        ip_and_mac_address_pair_regex.globalMatch(arp_ouput_stream);
+    QRegularExpressionMatchIterator iter = ip_and_mac_address_pair_regex.globalMatch(arp_output);
 
     const QString arp_format_mac_address =
         simplify_mac_address(QString::fromStdString(mac_address));
+
+    std::optional<mp::IPAddress> best_match;
+
     while (iter.hasNext())
     {
         QRegularExpressionMatch match = iter.next();
         // group 0 is the full match, 1 and 2 are the inner groups
         if (match.captured(2) == arp_format_mac_address)
         {
-            return {mp::IPAddress{match.captured(1).toStdString()}};
+            mp::IPAddress current_ip{match.captured(1).toStdString()};
+
+            if (!best_match.has_value() || current_ip > best_match.value())
+            {
+                best_match = current_ip;
+            }
         }
     }
 
-    return std::nullopt;
+    return best_match;
 }
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -74,6 +74,7 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     // ? (192.168.64.2) at 52:54:0:2a:12:b6 on bridge100 ifscope [bridge]
     // ? (192.168.64.3) at 52:54:0:85:72:55 on bridge100 ifscope [bridge]
     // ? (192.168.64.4) at 52:54:0:e1:cd:ab on bridge100 ifscope [bridge]
+    // ? (192.168.64.5) at 52:54:0:e1:cd:ab on bridge100 ifscope [bridge]
     // ? (192.168.64.255) at ff:ff:ff:ff:ff:ff on bridge100 ifscope [bridge]
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
 
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -84,8 +84,6 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     const QString arp_format_mac_address =
         simplify_mac_address(QString::fromStdString(mac_address));
 
-    std::optional<mp::IPAddress> best_match;
-
     while (iter.hasNext())
     {
         QRegularExpressionMatch match = iter.next();
@@ -99,13 +97,12 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
                        current_ip.as_string(),
                        mac_address);
 
-            if ((!best_match.has_value() || current_ip > best_match.value()) &&
-                mp::backend::can_reach_gateway(current_ip.as_string()))
+            if (mp::backend::can_reach_gateway(current_ip.as_string()))
             {
-                best_match = current_ip;
+                return {mp::IPAddress{match.captured(1).toStdString()}};
             }
         }
     }
 
-    return best_match;
+    return std::nullopt;
 }
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -60,6 +60,16 @@ QString get_arp_output()
     return QString{arp_process->read_all_standard_output()};
 }
 
+bool ping_ip(const mp::IPAddress& ip_addr)
+{
+    constexpr auto timeout = 500;
+    const auto arp_process = mp::platform::make_process(
+        mp::simple_process_spec("ping", {"-c", "1", QString::fromStdString(ip_addr.as_string())}));
+    const auto arp_exit_state = arp_process->execute(timeout);
+
+    return arp_exit_state.completed_successfully();
+}
+
 } // namespace
 
 std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& mac_address)
@@ -90,7 +100,7 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
         {
             mp::IPAddress current_ip{match.captured(1).toStdString()};
 
-            if (!best_match.has_value() || current_ip > best_match.value())
+            if ((!best_match.has_value() || current_ip > best_match.value()) && ping_ip(current_ip))
             {
                 best_match = current_ip;
             }
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -20,7 +20,7 @@
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
 #include <multipass/process/simple_process_spec.h>
-#include <shared/shared_backend_utils.h>
+#include <multipass/utils.h>
 
 namespace mp = multipass;
 
@@ -64,6 +64,11 @@ QString get_arp_output()
     return QString{arp_process->read_all_standard_output()};
 }
 
+bool can_reach_gateway(const std::string& ip)
+{
+    return MP_UTILS.run_cmd_for_status("ping", {"-n", "-q", ip.c_str(), "-c", "1", "-t", "1"});
+}
+
 } // namespace
 
 std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& mac_address)
@@ -98,7 +103,7 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
                        current_ip.as_string(),
                        mac_address);
 
-            if (mp::backend::can_reach_gateway(current_ip.as_string()))
+            if (can_reach_gateway(current_ip.as_string()))
             {
                 return current_ip;
             }
--- a/src/platform/backends/shared/macos/backend_utils.cpp
+++ b/src/platform/backends/shared/macos/backend_utils.cpp
@@ -79,8 +79,8 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
 
     const QString arp_output = get_arp_output();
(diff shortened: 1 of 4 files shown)


src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -73,22 +86,34 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     // ? (192.168.64.255) at ff:ff:ff:ff:ff:ff on bridge100 ifscope [bridge]
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
```
ricab, line 85: Could you update the example output with a case of duplicate IPs?

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -60,6 +63,16 @@ QString get_arp_output()
...
 }
 
+bool ping_ip(const mp::IPAddress& ip_addr)
+{
+    constexpr auto timeout = 500;
+    const auto arp_process = mp::platform::make_process(
+        mp::simple_process_spec("ping", {"-c", "1", QString::fromStdString(ip_addr.as_string())}));
+    const auto arp_exit_state = arp_process->execute(timeout);
+
+    return arp_exit_state.completed_successfully();
```
ricab: I wish there was a simple and reliable way to do this without forking, but I didn't find any, without introducing new dependencies, objecting-C wrapping, etc. This is waiting for ARP in a loop anyway, so performance isn't a concern. All things considered, I agree this is the right choice 👍

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -60,6 +63,16 @@ QString get_arp_output()
     return QString{arp_process->read_all_standard_output()};
 }
 
+bool ping_ip(const mp::IPAddress& ip_addr)
+{
+    constexpr auto timeout = 500;
```
ricab: In your experience, is this usually enough? I am concerned that it might not be if the machine is bogged down for some reason. OTOH, if we went much higher, we would risk not being able to try all IPs (for example, in a situation of two consecutive reboots and 3 IPs in cache), within the overall 2s timeout. Is that what you were thinking? It might be worth a comment explaining the delicate balance.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -73,22 +86,34 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
     // ? (192.168.64.255) at ff:ff:ff:ff:ff:ff on bridge100 ifscope [bridge]
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
 
-    const QString arp_ouput_stream = get_arp_output();
+    const QString arp_output = get_arp_output();
```
ricab, line 86: 👍

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -73,22 +86,34 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
         {
-            return {mp::IPAddress{match.captured(1).toStdString()}};
+            mp::IPAddress current_ip{match.captured(1).toStdString()};
+
+            mpl::trace(category,
+                       "Found ip match \'{}\': for mac address: \'{}\'",
+                       current_ip.as_string(),
+                       mac_address);
+
+            if ((!best_match.has_value() || current_ip > best_match.value()) && ping_ip(current_ip))
```
ricab: Hmm, if an IP is found that we can ping, that's the one, right? So why continue looping. I would expect this to return as soon as it finds something pingable.

Also, I don't quite get the logic of the largest IP. I would understand if you decided the order in which you visit IPs by which one is larger, to improve chances that we ping the right one first. But that is not what's happening here.

ricab (changes requested):
Hey @sharder996, this is a great initiative, we have been needing such an improvement for a long time!

I have two main concerns:

1. I seem to remember ARP not always being populated with the newer IP upon reboot. For example, when doing `multipass restart`. macOS has been evolving in virtualization and related, so maybe there is a better way to get the IP now?
2. the logic of the largest IPs (see below). Also, did you consider exploring `arp` to see if there are any guarantees regarding the order of the output? Failing that, did you observe any reliable pattern in practice? For example, newer IPs always showing up before or after older ones for the same MAC address? The criterion of larger IP would fail when they cycle. 

Regarding 2, whatever we choose, it should be only used to decide visiting order, no? We could probably settle for poor visiting order (e.g. when IPs cycle), but I would rather avoid skipping the right IP entirely.

[secondary review]

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: Same... 😞

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: For reference, when I found a responsive IP address the return was pretty much instantaneous. Irrelevant now since I'm using the shared utils version of `ping`. I found this solution to work, but a `waittime` of 1 millisecond as was previously used might be a bit aggressive.

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: Yes, opting to return early now as soon as a responsive IP address is found. A slight optimization might be to iterate through the match in the reverse order, but `QRegularExpressionMatchIterator` doesn't support reverse iterators and storing all matches in a list would mean iterating over all matches anyways. Not much of a saving in my eyes.

ricab: It's not the iteration itself that you save, it's the `ping`. And you wouldn't need to store the matches in a list, you could just use [QRegularExpressionMatch](https://doc.qt.io/qt-6/qregularexpressionmatch.html) and iterate from n to 0.

src/platform/backends/shared/shared_backend_utils.h
```diff
@@ -76,5 +76,10 @@ void ensure_vm_is_running_for(VirtualMachine* virtual_machine,
         throw StartException(virtual_machine->vm_name, msg);
     }
 }
+
+inline bool can_reach_gateway(const std::string& ip)
+{
+    return MP_UTILS.run_cmd_for_status("ping", {"-n", "-q", ip.c_str(), "-c", "1", "-W", "1"});
```
ricab: Shouldn't we have a timeout still? What happens if we take longer than 1s with pings? And why did you drop the 2s change?

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: QRegularExpressionMatch only provides a single match from the `arp` output. We would need all the matches and then iterate over them in reverse order.

ricab: OK. I think it would be a worthwhile addition (one way or another), but we can open an issue for the future if you prefer to limit this one. That might give us a better notion of how important it would be.

src/platform/backends/shared/shared_backend_utils.h
sharder996: There is something of a timeout built into the command which was in place previously:

```
-W waittime
    Time in milliseconds to wait for a reply for each packet sent.  If a reply arrives 
    later, the packet is not printed as replied, but considered as replied when
    calculating statistics.
```

However, I wonder if using this would be better:

```
-t timeout
    Specify a timeout, in seconds, before ping exits regardless of how many packets have
    been received.
```

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: Yes, I agree saving the `ping` would be desirable. However, the problem that this PR is trying to solve is an edge case in the first case where the only situation this was coming up was with launching Ubuntu Core images. Secondly, I only ever saw there being two IPs for the same MAC address so we aren't exactly executing a lot of pings.

src/platform/backends/shared/shared_backend_utils.h
sharder996: I dropped the 2s change because I understood it to be a polling rate, but it didn't have that effect in practise. I decided to just keep it as is for the time being.

ricab: > I dropped the 2s change because I understood it to be a polling rate, but it didn't have that effect in practise.

Could you elaborate, please? I thought it would have a double effect: limiting the time to get the IP, thus establishing a period for polling again. I think this is relevant because we might run out of time once we start pinging multiple IPs per iteration.

ricab: As for the `ping` timeouts, it looks like it has different meaning depending on platform. On Linux, I have:

```
-t ttl
           ping only. Set the IP Time to Live.

[...]

 -w deadline
           Specify a timeout, in seconds, before ping exits regardless of how
           many packets have been sent or received. In this case ping does not
           stop after count packet are sent, it waits either for deadline
           expire or until count probes are answered or for some error
           notification from network.

 -W timeout
           Time to wait for a response, in seconds. The option affects only
           timeout in absence of any responses, otherwise ping waits for two
           RTTs. Real number allowed with dot as a decimal separator
           (regardless locale setup). 0 means infinite timeout.
```

Also, you missed the last sentence from `-W` on macOS: `n.b. Many other Unix variants specify this option in Seconds.`

It looks like you're going to need a separate ping after all?

sharder996: > Also, you missed the last sentence from `-W` on macOS: `n.b. Many other Unix variants specify this option in Seconds.`

No mention of this on my Mac's manual page... However, it looks like the usage of `ping` actually varies quite a bit platform to platform. So, platform specific implementation it is.

sharder996: > > I dropped the 2s change because I understood it to be a polling rate, but it didn't have that effect in practise.
> 
> Could you elaborate, please? I thought it would have a double effect: limiting the time to get the IP, thus establishing a period for polling again. I think this is relevant because we might run out of time once we start pinging multiple IPs per iteration.

Ok, I think this variable was inappropriately named and causing some confusion. This timeout is specifically for the creation of an ssh session; resolving an IP address, pinging the address, creating the ssh session. This series of actions will loop until the general launch timeout is reached. There is a 1 second delay between the end of each action which is defined in https://github.com/canonical/multipass/blob/b78866f4947fa0f1c732cf095c5acb9fa77161ac/include/multipass/utils.h#L326

In practise, I found each loop of an attempted ssh session creation to take 2-3s, which includes first trying to ping an unresponsive IP address. So, rather than prematurely terminate the ssh session I went on the generous side and specified a timeout of 5s.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -70,23 +74,34 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+            mp::IPAddress current_ip{match.captured(1).toStdString()};
+
+            mpl::trace(category,
+                       "Found ip match \'{}\': for mac address: \'{}\'",
+                       current_ip.as_string(),
+                       mac_address);
+
+            if (MP_PLATFORM.can_reach_gateway(current_ip.as_string()))
+            {
+                return {mp::IPAddress{match.captured(1).toStdString()}};
```
ricab: Why repeat this?

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -70,23 +74,34 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+    // ? (192.168.64.5) at 52:54:0:e1:cd:ab on bridge100 ifscope [bridge]
     // ? (192.168.64.255) at ff:ff:ff:ff:ff:ff on bridge100 ifscope [bridge]
     // ? (224.0.0.251) at 1:0:5e:0:0:fb on en0 ifscope permanent [ethernet]
 
-    const QString arp_ouput_stream = get_arp_output();
+    const QString arp_output = get_arp_output();
     const QRegularExpression ip_and_mac_address_pair_regex(R"(\(([^)\s]+)\) at ([^\s]+))");
-    QRegularExpressionMatchIterator iter =
-        ip_and_mac_address_pair_regex.globalMatch(arp_ouput_stream);
+    QRegularExpressionMatchIterator iter = ip_and_mac_address_pair_regex.globalMatch(arp_output);
```
ricab: Might be a good chance to simplify this long name.

src/platform/backends/shared/shared_backend_utils.h
ricab: > This timeout is specifically for the creation of an ssh session; resolving an IP address, pinging the address, creating the ssh session.

I don't think that's right. `wait_until_ssh_up_helper` receives an overall timeout and declares an inner timeout (wait_step) for only the IP part (passed to `ssh_hostname()`). With the qemu backend, that inner timeout is passed to `mp::backend::ip_address_for` which in turn uses `try_action_for`, looping on `QemuPlatform::get_ip_for` until the inner timeout is reached. But it doesn't preempt any individual attempt at getting an IP, it just stops retrying after that much time has passed. (this scheme is all a big mess if you ask me, but that is out of scope). So, once it's called, `get_neighbour_ip` has all the time in the world to ping as many IPs as it wants. It might just not be called again in that *inner iteration*, but then `wait_until_ssh_up_helper` would do another *outer iteration* (until the outer timeout was reached), resulting in the IP resolution all over again anyway.

In short, I don't think that inner timeout makes a big difference. Increasing it would let us save a few CPU cycles going in and out of loops, but it would also decrease the resolution with which we respect the overall timeout. With your 5s, a start operation would often timeout only after say 63s or 64s when the user specified 60s.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -92,15 +92,15 @@ std::optional<mp::SSHSession> wait_until_ssh_up_helper(mp::VirtualMachine* virtu
                                                        std::chrono::milliseconds timeout,
                                                        const mp::SSHKeyProvider& key_provider)
 {
-    static constexpr auto wait_step = 1s;
+    static constexpr auto ssh_timeout = 5s;
```
ricab: Having had a closer look, I don't think we should have this increase, to avoid bigger deviations from the overall timeout. Also, I don't think `ssh_timeout` is accurate. `wait_step` made more sense to me TBH. Perhaps `hostname_time_step` or just `ip_step`?

src/platform/platform_win.cpp
```diff
@@ -1082,3 +1082,8 @@ std::filesystem::path mp::platform::Platform::get_root_cert_dir() const
     // Windows doesn't use `daemon_name` for the data directory (see `program_data_multipass_path`)
     return base_dir / "Multipass" / "data";
 }
+
+bool mp::platform::Platform::can_reach_gateway(const std::string& ip) const
```
ricab: This is never used, is it? I would prefer not to have code that doesn't get exercised lying around. 

IIANM, this is useful only for QEMU and libvirt. But libvirt is going away, which leaves only QEMU. There is a facility for QEMU things that depend on linux/macOS: `QemuPlatformDetail`. Before this PR, `can_reach_gateway` was only called from something that `QemuPlatformDetail` called anyway. With this PR, it is still always called from either linux or apple specific code. 

So why not base this PR on top of Antoni's PR and try to make `can_reach_gateway` part of the QemuPlatformDetail? It should be possible to move things using it in backend_utils there as well. That would untangle a good deal of spaghetti. Brave enough to give it a shot? 😉

tests/macos/test_backend_utils.cpp
```diff
@@ -63,6 +67,17 @@ struct GetNeighbourIPValidInputsTests
...
+
+    EXPECT_CALL(*mock_utils, run_cmd_for_status(QString("ping"), _, _))
+        .WillRepeatedly([](const QString&, const QStringList& args, auto&&) {
+            if (args.contains("192.168.64.5"))
+            {
+                return false;
+            }
+            return true;
+        });
+
```
ricab, line 76: 👍

ricab (changes requested):

src/platform/backends/shared/shared_backend_utils.h
sharder996: Ah, ok. Thanks for your explanation! What would you suggest for an inner timeout then? The most I've seen for multiple IPs is two IPs for a single MAC address, so operating with that assumption, and with the ping timeout being 1s we would have a maximum of 2s to ping every IP address for a given MAC address.

src/platform/backends/shared/macos/backend_utils.cpp
sharder996: No reason 😝

src/platform/backends/shared/base_virtual_machine.cpp
sharder996: I just didn't like the word "step", because it implies something that happens on a regular consistent basis. Hence, my initial confusion. When in reality, it is actually a timeout.

src/platform/platform_win.cpp
sharder996: I did some reorganizing...

Personally, I don't like the idea of putting `can_reach_gateway` in QemuPlatform or any Qemu related utils class since the action of pinging isn't really related to Qemu. Its more a system OS/platform dependent action. The `can_reach_gateway` isn't used anywhere else other than in each platform specific backend_utils so it doesn't make sense to have it available through a public interface either. Instead I just added it the unnamed namespace in the two places its needed.

src/platform/backends/shared/shared_backend_utils.h
ricab: Hmm, sorry I am not following. I hadn't inspected the code as closely before, so my earlier comments were probably misleading. 

From what I saw today, keeping that `wait_step` as 1s would help with overall timeout resolution, without preventing all IPs from being tested with ping. The operation to get IPs from ARP isn't really subject to any timeout AFAICT. That's what I meant by "`get_neighbour_ip` has all the time in the world to ping as many IPs as it wants". But perhaps I missed something?

src/platform/backends/shared/base_virtual_machine.cpp
ricab: Oh OK. I think step refers to one individual iteration of the loop here, whose duration will be dominated by the`ssh_hostname` call until an IP is found. But I am fine with changing step, it was "ssh" that I found confusing myself.

src/platform/platform_win.cpp
ricab: You're right, good call!

ricab (approved):
Alright, I am good with this once CI goes green. Good job!

src/platform/backends/shared/shared_backend_utils.h
sharder996: `wait_step` get applied as a timeout to `ssh_hostname`. `ssh_hostname` uses that timeout for `ip_address_for` which then uses it for an action with uses the passes `Callable` which contains `get_ip_for` and then ultimately `get_neighbour_ip`. So, yes, AFAIK `wait_step` is a timeout for pinging IPs.

src/platform/backends/shared/base_virtual_machine.cpp
sharder996: I guess it more of a hostname resolution timeout. But, that's getting a bit pedantic. I'm open to suggestions as well 😛 , but I'm also fine with leaving it as is.

src/platform/backends/shared/shared_backend_utils.h
ricab: That was my impression initially too, but look at `try_action_for`, which is where the timeout and the action get passed:

https://github.com/canonical/multipass/blob/90af9ee8ca8dc756b64901e48d0ac8aae3aa53ee/include/multipass/utils.h#L308-L330

All the timeout does is prevent further executions of the action, once it has elapsed. It does not, however, interrupt ongoing executions, right?

sharder996: Hmmm, I think you're right. Thanks again for the clarification!

PR #4384, opened 2025-09-22
[ci] Prune vcpkg binaries kaizen

sharder996:
Couple fixes and improvements here
- Only query NuGet package types
- Packages are fetched at the `org` level, so filter out non-Multipass packges
- Create Mattermost alert on workflow failure
- Entire workflow won't fail if one package fails to be deleted

Example workflow can be found [here](https://github.com/canonical/multipass/actions/runs/17925720184). Old packages were pruned in a previous test run [here](https://github.com/canonical/multipass/actions/runs/17924005630/job/50965570706).

### DIFF
--- a/.github/workflows/prune-packages.yml
+++ b/.github/workflows/prune-packages.yml
@@ -6,6 +6,9 @@ on:
   # 04:11 UTC on the first day of each month
   - cron: '11 4 1 * *'
 
+permissions:
+  packages: write
+
 jobs:
   list-packages:
     runs-on: ubuntu-latest
@@ -43,6 +46,9 @@ jobs:
   prune:
     needs: list-packages
     runs-on: ubuntu-latest
+
+    permissions:
+      packages: write
 
     strategy:
       matrix: ${{ fromJSON(needs.list-packages.outputs.matrix) }}
--- a/.github/workflows/prune-packages.yml
+++ b/.github/workflows/prune-packages.yml
@@ -19,15 +19,25 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            core.info(`Listing packages for org: ${owner}`);
+
             // Fetch all container packages in this org/user
             const response = await github.paginate(
               "GET /orgs/{org}/packages",
               { org: context.repo.owner, package_type: "nuget", per_page: 100 }
             );
-            // Extract unique names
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-packages.yml
```diff
@@ -45,3 +55,20 @@ jobs:
...
           package-type: nuget
           min-versions-to-keep: 1
+
+  notify:
+    runs-on: ubuntu-latest
+    needs: [list-packages, prune]
+    if: ${{ always() }}
+    steps:
+      - name: Report workflow failure
+        if: ${{ needs.list-packages.result != 'success' || needs.prune.result != 'success' }}
```
ricab: With the difference in logic with respect to #4366 ?

.github/workflows/prune-packages.yml
```diff
@@ -19,15 +19,25 @@ jobs:
...
           script: |
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            core.info(`Listing packages for org: ${owner}`);
+
             // Fetch all container packages in this org/user
             const response = await github.paginate(
               "GET /orgs/{org}/packages",
-              { org: context.repo.owner, package_type: "container", per_page: 100 }
+              { org: context.repo.owner, package_type: "nuget", per_page: 100 }
```
ricab, line 32: If we ever start uploading more than 100 packages in between runs of this, it would not be able to counter the growth. Should this be wrapped in a loop? See https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28

Unless you are certain that the number of packages that gets uploaded before this runs is way lower than 100, and will likely always remain that way?

.github/workflows/prune-packages.yml
sharder996: The logic here is different because the notification is in a different job rather than a subsequent step.

.github/workflows/prune-packages.yml
sharder996: This doesn't seem to be the case. Did a test workflow with 10 packages/page and it still picked up all the packages (75 in total).
https://github.com/canonical/multipass/actions/runs/17949413604/workflow

.github/workflows/prune-packages.yml
ricab: OK, but wouldn't `if: ${{ !success() }}` in line 62 achieve the same thing?

tobe2098 (commented):
LGTM

PR #4389, opened 2025-09-24
[feature] Remove blueprints

### LINKED ISSUE #4292: Remove deprecated blueprints feature
It will be removed sometime in the future. It might stay around for another release in case any regressions come up, but it hasn't been discussed by the team yet exactly when we want to remove the feature.

_Originally posted by @sharder996 in https://github.com/canonical/multipass/pull/4291#issuecomment-3157186319_

tobe2098:
Fixes #4292

MULTI-2239

This PR removes the blueprints feature, which was previously deprecated in favor of standalone cloud-inits for customized instances in Multipass (see [here](https://documentation.ubuntu.com/multipass/latest/how-to-guides/manage-instances/launch-customized-instances-with-multipass-and-cloud-init/)). 
This PR also removes Poco::ZIP since blueprints was the only place where this package was being used.
After this PR is merged https://github.com/canonical/multipass-blueprints will no longer be needed.

### DIFF
--- a/3rd-party/submodule_info.md
+++ b/3rd-party/submodule_info.md
@@ -46,11 +46,6 @@ Version: 1.9.4 |
 <https://github.com/open-source-parsers/jsoncpp.git> |
 <https://github.com/open-source-parsers/jsoncpp/releases>
 
-### POCO
-Version: 1.13.3 |
-<https://github.com/pocoproject/poco.git> |
-<https://github.com/pocoproject/poco/releases>
-
 ### vcpkg
 Version: 2025.04.09 |
 <https://github.com/microsoft/vcpkg.git> |
--- a/packaging/macos/LICENCE.html
+++ b/packaging/macos/LICENCE.html
@@ -29,7 +29,6 @@
   </ul>
 </li>
 <li>libssh: <a href="https://github.com/albaguirre/libssh">https://github.com/albaguirre/libssh</a></li>
-<li>poco: <a href="https://github.com/pocoproject/poco">https://github.com/pocoproject/poco</a></li>
 <li>qt: <a href="https://www.qt.io/">https://www.qt.io/</a></li>
 <li>semver: <a href="https://github.com/zmarko/semver">https://github.com/zmarko/semver</a></li>
 <li>xz-embedded: <a href="https://git.tukaani.org/?p=xz-embedded.git">https://git.tukaani.org/?p=xz-embedded.git</a></li>
@@ -72,7 +71,6 @@
   <li>OpenSSL Exception</li>
   </ul>
 </li>
-<li>poco: Boost Software License - Version 1.0 - August 17th, 2003</li>
 <li>qt: GNU LGPL Version 3; Copyright (c) 2016 The Qt Company</li>
 <li>semver: MIT License; Copyright (c) 2015 Marko &#x17D;ivanovi&#x107;</li>
 <li>xz-embedded: zx-embedded License;</li>
--- a/packaging/windows/LICENCE.rtf
+++ b/packaging/windows/LICENCE.rtf
@@ -15,7 +15,6 @@
 {\pard \ql \f0 \sa0 \li720 \fi-360 \endash \tx360\tab protobuf: https://github.com/google/protobuf\par}
 {\pard \ql \f0 \sa0 \li720 \fi-360 \endash \tx360\tab zlib: https://github.com/madler/zlib\sa180\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab libssh: https://git.libssh.org/projects/libssh.git/\par}
-{\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab poco: https://github.com/pocoproject/poco\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab qt: https://www.qt.io/\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab semver: https://github.com/zmarko/semver\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab xz-embedded: https://git.tukaani.org/?p=xz-embedded.git\par}
@@ -35,7 +34,6 @@
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab libssh:\par}
 {\pard \ql \f0 \sa0 \li720 \fi-360 \endash \tx360\tab GNU LGPL Version 2.1 License; See AUTHORS file for Copyright information\par}
 {\pard \ql \f0 \sa0 \li720 \fi-360 \endash \tx360\tab OpenSSL Exception\sa180\par}
-{\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab poco: Boost Software License - Version 1.0 - August 17th, 2003\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab qt: GNU LGPL Version 3; Copyright (c) 2016 The Qt Company\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab semver: MIT License; Copyright (c) 2015 Marko \u381 ?ivanovi\u263 ?\par}
 {\pard \ql \f0 \sa0 \li360 \fi-360 \bullet \tx360\tab xz-embedded: zx-embedded License;\par}
--- a/packaging/windows/LICENCE.txt
+++ b/packaging/windows/LICENCE.txt
@@ -13,7 +13,6 @@ https://github.com/canonical/multipass/
   - protobuf: https://github.com/google/protobuf
   - zlib: https://github.com/madler/zlib
 - libssh: https://git.libssh.org/projects/libssh.git/
-- poco: https://github.com/pocoproject/poco
 - qt: https://www.qt.io/
 - semver: https://github.com/zmarko/semver
 - xz-embedded: https://git.tukaani.org/?p=xz-embedded.git
@@ -36,7 +35,6 @@ Below is a summary of the licensing used in the respective projects. See their s
 - libssh:
   - GNU LGPL Version 2.1 License; See AUTHORS file for Copyright information
   - OpenSSL Exception
-- poco: Boost Software License - Version 1.0 - August 17th, 2003
 - qt: GNU LGPL Version 3; Copyright (c) 2016 The Qt Company
 - semver: MIT License; Copyright (c) 2015 Marko Živanović
 - xz-embedded: zx-embedded License;


src/daemon/daemon.cpp
```diff
@@ -503,36 +497,20 @@ std::vector<mp::NetworkInterface> validate_extra_interfaces(
...
-                vault.all_info_for(image_query).empty())
-                throw mp::ImageNotFoundException(request->image(), request->remote_name());
-        }
-    }
-    catch (const mp::IncompatibleBlueprintException&)
-    {
-        throw std::runtime_error(
-            fmt::format("The \"{}\" Blueprint is not compatible with this host.",
-                        request->image()));
-    }
```
tobe2098, line 524: Interestingly, the caught exception is never thrown in the codebase, so this piece of code should never happen.

tobe2098: This particular section of code is not very well written for tests. Apparently the non-blueprint tests that reach this point depend on the blueprint_provider to run properly by skipping the if block code.

tobe2098: Note: This function as previously written had improper mock behavior in tests, offloading image validation to the blueprint_provider even when no blueprints were involved. 186 tests are currently involved, although they could be linked to other issues in the code.

tests/test_daemon.cpp
tobe2098, line 1: Many of the tests in this file have had to have been readjusted because previous blueprint stubs were taking over the mocks in an improper way. Once the blueprint mocks were removed, unrelated tests failed, and the solution was to properly mock the VMImageVault and adjust the mocking in another couple of tests.

tobe2098 (commented):
All removals and additions seem correct, but I would appreciate a double check since the changes are very immersed in the rest of the code.

tobe2098:
I would also appreciate if someone would comment on the necessary steps to remove Poco as a dependency if some were missed.

tobe2098:
In case anyone is concerned, I have checked the codecov report and the two lines it mentions from the patch as not covered are unchanged lines that were not covered previously as well.

tests/test_daemon.cpp
```diff
@@ -655,24 +615,36 @@ TEST_P(DaemonCreateLaunchTestSuite, createsVirtualMachines)
...
+    EXPECT_CALL(*mock_factory, prepare_source_image(_));
+    EXPECT_CALL(*mock_image_vault_ptr, fetch_image(_, _, _, _, _, _))
+        .WillOnce([](const mp::FetchType& f,
+                     const mp::Query& q,
+                     const mp::VMImageVault::PrepareAction& action,
+                     const mp::ProgressMonitor& pm,
+                     const std::optional<std::string>& opt,
+                     const mp::Path& path) -> mp::VMImage {
+            return action({"", {}, {}, {}, {}, {}});
+        });
```
tobe2098: Here the added mocked function is necessary as a fix to the fix. Previously, the blueprint_provider stub would provide its functionality in the validate_image call in daemon.cpp, and then the image_vault stub would do its intended function in the fetch_image() call. By fixing the first problem, we run into another, which is that the mock class needs to replicate the stub behavior, which is what the mock does.

src/daemon/daemon.cpp
```diff
@@ -1753,90 +1726,55 @@ try
...
-                else
-                    info.aliases = QStringList({info.id.left(12)});
-
-                auto remote_name = (!request->remote_name().empty() ||
-                                    (request->remote_name().empty() && vm_images_info.size() > 1 &&
-                                     remote != default_remote))
-                                       ? remote
-                                       : "";
+        wait_update_manifests_all_and_optionally_applied_force(
+            request->force_manifest_network_download());
```
sharder996, line 1706: This invocation can be extracted out to avoid repeating it in every conditional block.

sharder996: And, while your at it can you give it a shorter and better name?

src/rpc/multipass.proto
```diff
@@ -140,8 +140,8 @@ message FindRequest {
     string remote_name = 2;
     int32 verbosity_level = 3;
     bool allow_unsupported = 4;
-    bool show_images = 5;
-    bool show_blueprints = 6;
+    reserved 5; //show_images bool
+    reserved 6; //show_blueprints bool
```
sharder996: I think the only reason to keep these around is for backwards compatibility. Since we own both the client and server, the only scenario where that would come up is if a user somehow used an older version of the Multipass client with a newer version of the daemon.

There is also the possibility of someone writing their own custom client, but we don't advertise grpc as being a supported API, so I don't think we need to keep these placeholders around.

tests/blueprint_test_lambdas.cpp
sharder996, line 1: Why can't we remove this whole file?

tests/test_daemon.cpp
```diff
@@ -646,6 +605,7 @@ struct DaemonLaunchTimeoutValueTestSuite : public Daemon,
 TEST_P(DaemonCreateLaunchTestSuite, createsVirtualMachines)
 {
     auto mock_factory = use_a_mock_vm_factory();
+    (void)mock_image_vault();
```
sharder996: This is rather unsightly and an unexpected requirement for writing a test. Could it possibly be abstracted away into `use_a_mock_vm_factory` or perhaps into the `Daemon` test struct `SetUp`?

tests/daemon_test_fixture.cpp
```diff
@@ -363,7 +377,7 @@ mpt::MockVirtualMachineFactory* mpt::DaemonTestFixture::use_a_mock_vm_factory()
                                                      {"wlan0", "wi-fi", "wireless adapter"}}));
 
     config_builder.factory = std::move(mock_factory);
-    return mock_factory_ptr;
+    return reinterpret_cast<MockVirtualMachineFactory*>(config_builder.factory.get());
```
sharder996: What is the reason for this change?

sharder996 (changes requested):
Looking good! Just a few comments and questions.

tests/blueprint_test_lambdas.cpp
tobe2098: My bad, before removing the blueprint tests I thought the image functions were used in other tests. Good catch!

tests/test_daemon.cpp
tobe2098: Good points, including the posterior one. I have fixed it to follow more the general theme of the tests, although I based myself of use_a_mock_vm_factory(), which seems to be a similar patch.

I am resolving once I push the changes.

src/daemon/daemon.cpp
tobe2098: I'm happy to change the name, but it seems that in one of the three blocks in which it is called, it is called after checking validity of remote name, so I do not think that it can be extracted without cause without extracting more code (the check itself). What do you think?

tests/daemon_test_fixture.h
```diff
@@ -53,6 +53,8 @@ struct fake_vm_properties
     VirtualMachine::State state = VirtualMachine::State::starting;
 };
 
+class MockVMImageVault;
```
sharder996: Is this needed to building on Linux? Isn't required when building on my Macbook.

tests/mock_vm_image_vault.h
```diff
@@ -39,6 +40,13 @@ class MockVMImageVault : public VMImageVault
         });
         ON_CALL(*this, has_record_for(_)).WillByDefault(Return(true));
         ON_CALL(*this, minimum_image_size_for(_)).WillByDefault(Return(MemorySize{"1048576"}));
+
+        auto mock_image_host = MockImageHost();
+
+        ON_CALL(*this, all_info_for(_))
+            .WillByDefault(Return(std::vector<std::pair<std::string, mp::VMImageInfo>>{
+                std::pair<std::string, mp::VMImageInfo>{"default",
+                                                        mock_image_host.mock_bionic_image_info}}));
```
sharder996: Is the `MockImageHost` only needed for this string? It would be nice if we could make `mock_bionic_image_info` accessible in another way so that we don't have to create a mock image host just for it.

tests/test_daemon.cpp
```diff
@@ -1957,6 +1437,8 @@ TEST_P(RefuseBridging, oldImage)
     std::string full_image_name = remote.empty() ? image : remote + ":" + image;
 
     auto mock_factory = use_a_mock_vm_factory();
+    auto mock_vm_image_vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
+    config_builder.vault = std::move(mock_vm_image_vault);
```
sharder996: I tried playing around with this a bit and without this mock I see a bunch of the failures are from an `ImageNotFoundException` being thrown. The `config_builder` is already using a `StubVMImageVault`. Would it be simpler to instead return a dummy `VMImage` when `info_for(...)` is being called and avoid having to define all these mocks?

tests/test_daemon.cpp
```diff
@@ -2154,8 +1615,8 @@ TEST_F(Daemon, ctorDropsRemovedInstances)
 TEST_P(ListIP, listsWithIp)
 {
     auto mock_factory = use_a_mock_vm_factory();
-    config_builder.vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
-
+    auto mock_vm_image_vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
+    config_builder.vault = std::move(mock_vm_image_vault);
```
sharder996: What is the motivation behind these changes? I don't believe this is doing anything different.

tests/daemon_test_fixture.h
tobe2098: Good catch! Forgot to remove it from the previous test fix framework

tests/mock_vm_image_vault.h
tobe2098: It was the fastest way to get a mock VMImageInfo without duplicating, but I guess we could put it in the MockVMImageVault namespace itself or construct it inplace.

tests/test_daemon.cpp
tobe2098: I understand what you mean. I am somewhat hesitant, because the way I understand it when you use the stub the function calls always return empty values, and it is supposed to be that way. In a sense it does tell us which are the minimum requirements for the code path being run in the test. I may be wrong in my understanding though, I'm open to discuss to what extent should the minimum behavior from the Stub be sufficient.

Another possible solution I did think of was replacing the Stub in the config_builder by the mock, but it does not sound good to me (at least this solution) for the same reason. 

In the end I think that since it is a hard-coded value I am inclined to agree with you on this. It does get rid of a lot of lines.

tests/test_daemon.cpp
tobe2098: Good catch, after editing so many tests I must have misread the existing code. Will be removed if the change to Stub is satisfactory.

tests/test_daemon.cpp
```diff
@@ -2198,8 +1613,8 @@ INSTANTIATE_TEST_SUITE_P(Daemon,
 
 TEST_F(Daemon, preventsRepetitionOfLoadedMacAddresses)
 {
-    config_builder.vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
-
+    auto mock_vm_image_vault = std::make_unique<NiceMock<mpt::MockVMImageVault>>();
+    config_builder.vault = std::move(mock_vm_image_vault);
```
sharder996: This diff is idempotent.

src/daemon/daemon.cpp
sharder996: Ok, we can leave it for now

tests/test_daemon.cpp
tobe2098: I thought I had undone it, sorry. During testing which tests required the mock I must have reintroduced it unwittingly.

tobe2098: I will also take a look at the newline and constructor movements. The constructor was moved earlier for the first testing fix, but now it is not necessary anymore.

tests/daemon_test_fixture.cpp
tobe2098: Now that I noticed, this change was originally because I thought the previous behavior was undefined. The returned pointer has the .get() value from the unique_ptr before it is moved. After some double-checking it is clear that the value of that .get() pointer never changes during the unique_ptr lifetime, so the change was unnecessary.

tests/test_daemon.cpp
sharder996: np 😄 Thanks for cleaning it up!

tobe2098:
Good catch!

PR #4390, opened 2025-09-24
Update Qt

jimporter:
Spinning this off from the update to C++20 to nail down some strange behavior on macOS.

### DIFF
--- a/BUILD.windows.md
+++ b/BUILD.windows.md
@@ -54,11 +54,11 @@ choco install aqt -yfd
 Then specify the following options in the installation command:
 
 ```[pwsh]
-aqt install-qt windows desktop 6.2.4 win64_msvc2019_64 -O C:/Qt
+aqt install-qt windows desktop 6.10.0 win64_msvc2022_64 -O C:/Qt
 ```
 
-Alternatively, download the [qtbase archive](https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_624/qt.qt6.624.win64_msvc2019_64/6.2.4-0-202203140926qtbase-Windows-Windows_10_21H2-MSVC2019-Windows-Windows_10_21H2-X86_64.7z)
-and extract it to `C:\Qt` (so it ends up in `C:\Qt\6.2.4`).
+Alternatively, download the [qtbase archive](https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_6100/qt6_6100/qt.qt6.6100.win64_msvc2022_64/6.10.0-0-202510021201qtbase-Windows-Windows_11_24H2-MSVC2022-Windows-Windows_11_24H2-X86_64.7z)
+and extract it to `C:\Qt\6.10.0`.
 
 ### Path setup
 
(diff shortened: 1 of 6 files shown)


jimporter:
* [Passing job](https://github.com/canonical/multipass/actions/runs/17981792379/job/51149546819) with just the first commit (which improves failing test output)
* [Failing job](https://github.com/canonical/multipass/actions/runs/17982780844/job/51153151059) with Qt upgrade

jimporter:
> ❌ Patch coverage is `96.15385%` with `1 line` in your changes missing coverage. Please review.

This is a strange coverage report...

image

sharder996:
How do you feel about moving to Qt 6.10 right away? Or, was there a specific reason for using 6.9?

jimporter:
> How do you feel about moving to Qt 6.10 right away? Or, was there a specific reason for using 6.9?

I'm not strongly opposed, though Ubuntu 25.10 uses Qt 6.9.2, so sticking with the 6.9 series on Mac and Windows would mean less version skew on our official builds (~once we start using 25.10 in CI anyway~ e: GitHub uses LTS images). If we don't think that matters though, we can switch to 6.10.

xmkg:
> > How do you feel about moving to Qt 6.10 right away? Or, was there a specific reason for using 6.9?
> 
> 
> 
> I'm not strongly opposed, though Ubuntu 25.10 uses Qt 6.9.2, so sticking with the 6.9 series on Mac and Windows would mean less version skew on our official builds (~once we start using 25.10 in CI anyway~ e: GitHub uses LTS images). If we don't think that matters though, we can switch to 6.10.

FTR I plan to migrate the QT to vcpkg too after this one lands, so feel free to use the latest and the greatest.

BUILD.windows.md
```diff
@@ -67,7 +67,7 @@ You'll have to manually add CMake and Qt to your account's PATH variable.
 Search for "Edit environment variables for your account" then edit your Path variable. Add the following:
 
 - `C:\Program Files\CMake\bin`
-- `C:\Qt\6.2.4\msvc2019_64\bin`
+- `C:\Qt\6.10.0\bin`
```
jimporter, line 70: Note: The directory structure for the qtbase binary archive is different now.

src/utils/utils.cpp
```diff
@@ -161,6 +162,25 @@ QDir mp::utils::base_dir(const QString& path)
     return info.absoluteDir();
 }
 
+std::filesystem::path mp::utils::normalize_path(const std::filesystem::path& path)
```
xmkg, line 165: Since we're replacing Qt's [QDir::cleanPath](https://github.com/qt/qtbase/blob/26c45a1b534752df4ebd89206079cd7cfe55df84/src/corelib/io/qdir.cpp#L2517), we need to ensure that this function behaves as intended. Can we have unit tests to ensure that? QT's own test case for testing the cleanPath might be a good starting point:

https://github.com/qt/qtbase/blob/26c45a1b534752df4ebd89206079cd7cfe55df84/tests/auto/corelib/io/qdir/tst_qdir.cpp#L1328

xmkg (changes requested):
Looks good overall -- I have a request regarding the new normalize_path util function.

sharder996 (commented):
Looking good here as well. +1 on what @xmkg said and then I think it can go in!

src/utils/utils.cpp
jimporter: Ok, I added a bunch of tests for `normalize_path`, inspired by Qt's tests (both the version that takes a path and the version that takes a string).

sharder996 (approved):
LGTM! Thanks @jimporter

Issue #3147, opened 2023-06-30
Subnet used for default bridge interface for qemu is "random"


jocado:
**Describe the bug**
Describe what your problem is.
When you install multipass for the first time, and use the qemu local.driver, there is some kind of logic that picks a seemingly random subnet to use on the default bridge interface.  It seems like it can be any `/24` from `10/8`

This makes it not very usable for various different scenarios.  For instance, it could clash with another network you use, either locally or via some kind of VPN.

The subnet used needs to be deterministic, or at least configurable, otherwise it does't really work for some users.

It seems like I could probably stop multipass, remove the interface, and hack in the IP ranges and info I need here: `/var/snap/multipass/common/data/multipassd/network`

But, that isn't documents, and presumably not really supported.

**To Reproduce**
How, and what happened?
1. Install multipass, using the qemu driver
1. You might get lucky with the network assignment, you might not

**Expected behavior**
What did you expect to happen?

It should be possible to configure the network range used.

**Logs**
Please provide logs from the daemon, see [accessing logs](https://multipass.run/docs/accessing-logs) on where to find them on your platform.

**Additional info**

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.2 LTS"
```

```
$ multipass version
multipass   1.12.0
multipassd  1.12.0
```

```
$ multipass get local.driver
qemu
```

**Additional context**

townsend2010:
Hi @jocado!

It is a bit "random", but we do check if the subnet is already in use before using the determined subnet.

Have you run into a problem with this or are you being preemptive?

jocado:
Hi @townsend2010 

Thanks for taking a look.

Yes, I have run into an issue personally, and I think it has potential to be an issue for others too.

My scenario is:
- Install Multipass [ bridge network is setup ]
- Then connect to a different network, and find there is a clash that impacts functionality.

Even further, in my particular case, I'm using a VPN client to connect to that different network, which advertises routes for a range of different networks in a "greedy" fashion, one of which clashes with the MP bridge, and the traffic on my own host doesn't reach the bridge interface [ due to some specifics of the VPN client implementation ].

You could easily move around offices, and eventually find you clash with a locally L2 connected network.

[...]

PR #4398, opened 2025-09-30
Subnet refactor

Sploder12:
This PR removes all occurrences where we use strings to represent subnets and replaces them with a dedicated and robust Subnet class, properly tested and made platform independent. It also combines and mocks all usage of `uniform_int` random number generation to aid in testing. Making these changes removes some dangerous assumptions we are making such as every subnet (including ones we do not own) having the mask `255.255.255.0`, ensures subnets are valid subnets, and gives us significantly more flexibility for future subnetting.

Also modifies Linux Qemu to use AZ subnets instead of its own.

Windows and MacOS do not implement the platform dependent parts yet, this will come in later PRs that include their AZ networking implementation.

The Subnet class does not handle prefix lengths `/31` and `/32` since I don't believe we plan on using them and they require special logic.


include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+#include <multipass/path.h>
+#include <multipass/singleton.h>
+
+#include "ip_address.h"
+
+#define MP_SUBNET_UTILS multipass::SubnetUtils::instance()
+
+namespace multipass
+{
+class Subnet
```
xmkg, line 28: This could use a fmt specialization

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <fmt/format.h>
+
+#include <algorithm>
+#include <limits>
+
+namespace mp = multipass;
+
+namespace
+{
+constexpr auto large_CIDR_err_fmt = "CIDR value must be non-negative and less than 31: {}";
```
xmkg: Is there a reason to move this to a dedicated variable?

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+mp::Subnet parse(const std::string& cidr_string)
+try
+{
+    if (auto i = cidr_string.find('/'); i != std::string::npos)
+    {
+        mp::IPAddress addr{cidr_string.substr(0, i)};
+
+        auto cidr = std::stoul(cidr_string.substr(i + 1));
+        if (cidr >= 31)
+            throw std::invalid_argument(fmt::format(large_CIDR_err_fmt, cidr));
```
xmkg: I think this deserves a dedicated exception type, and we can use https://github.com/canonical/multipass/blob/main/include/multipass/exceptions/formatted_exception_base.h as the base class.

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+    for (size_t i = 0; i < ip.octets.size(); ++i)
+    {
+        ip.octets[i] &= mask.octets[i];
+    }
+    return ip;
+}
+
+mp::Subnet generate_random_subnet(uint8_t cidr, mp::Subnet range)
+{
+    if (cidr >= 31)
```
xmkg: I feel like CIDR (prefix_len) could be a strong type

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+    throw std::invalid_argument(
+        fmt::format("CIDR address {} does not contain '/' seperator", cidr_string));
+}
+catch (const std::out_of_range& e)
+{
+    throw std::invalid_argument(fmt::format(large_CIDR_err_fmt, e.what()));
+}
+
+[[nodiscard]] mp::IPAddress get_subnet_mask(uint8_t cidr)
```
xmkg: CIDR is the whole notation's name `<ip>/<prefix_length>`. It would be better to name cidr as prefix_length

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    using Octects = decltype(mp::IPAddress::octets);
```
xmkg: Can we simplify this like:

```C++
std::uint32_t mask = (prefix_length == 0) ? 0 : (~std::uint32_t(0) << (32 - prefix_length));
```

Then, unpack it into the array?

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+        assert(start_octet < octets.size()); // sanity
+
+        // remain = 5, 1 << (8 - 5) = 00001000 -> 00000111 -> 11111000
+        octets[start_octet] = ~((1u << (value_size - remain)) - 1u);
+    }
+
+    return mp::IPAddress{octets};
+}
+
+[[nodiscard]] mp::IPAddress apply_mask(mp::IPAddress ip, uint8_t cidr)
```
xmkg: idem (cidr -> prefix_length)

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    const auto mask = get_subnet_mask(cidr);
+    for (size_t i = 0; i < ip.octets.size(); ++i)
+    {
+        ip.octets[i] &= mask.octets[i];
```
xmkg, line 69: How about overloading operator & (the binary one) for IPAddress?

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+    const auto subnet = static_cast<size_t>(MP_UTILS.random_int(0, possibleSubnets - 1));
+
+    // ex. 192.168.0.0 + (4 * 2^(32 - 24)) = 192.168.0.0 + 1024 = 192.168.4.0
+    mp::IPAddress id = range.identifier() + (subnet * (std::size_t{1} << (32 - cidr)));
+
+    return mp::Subnet{id, cidr};
+}
+} // namespace
+
+mp::Subnet::Subnet(IPAddress ip, uint8_t cidr) : id(apply_mask(ip, cidr)), cidr(cidr)
```
xmkg: What does `id` mean in this context? Network address? If so (the code looks that way), it's better to name it as such.

src/network/subnet.cpp
xmkg: Also, we could subclass the IPAddress type for a Mask type for extra safety, so no one can accidentally do "ip & ip".

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+{
+public:
+    Subnet(IPAddress ip, uint8_t cidr);
```
xmkg: cidr -> prefix_length

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+    Subnet(const std::string& cidr_string);
```
xmkg, line 62: this one is actually correct, e.g., "192.168.1.0/24"

include/multipass/subnet.h
xmkg: ip -> network address

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+
+    [[nodiscard]] IPAddress min_address() const;
+    [[nodiscard]] IPAddress max_address() const;
+    [[nodiscard]] uint32_t address_count() const;
+
+    [[nodiscard]] IPAddress identifier() const;
+    [[nodiscard]] uint8_t CIDR() const;
```
xmkg: prefix_length

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+{
+public:
+    Subnet(IPAddress ip, uint8_t cidr);
+    Subnet(const std::string& cidr_string);
+
+    [[nodiscard]] IPAddress min_address() const;
+    [[nodiscard]] IPAddress max_address() const;
+    [[nodiscard]] uint32_t address_count() const;
+
+    [[nodiscard]] IPAddress identifier() const;
```
xmkg: network_address

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+    [[nodiscard]] uint8_t CIDR() const;
+    [[nodiscard]] IPAddress subnet_mask() const;
+
+    // uses CIDR notation
+    [[nodiscard]] std::string as_string() const;
```
xmkg: Maybe name it as to_cidr()?

include/multipass/subnet.h
xmkg: ...or, you can define a strong type for CIDR if you want to be even more pedantic

include/multipass/subnet.h
```diff
@@ -0,0 +1,67 @@
...
+private:
+    IPAddress id;
+    uint8_t cidr;
+};
+
+struct SubnetUtils : Singleton<SubnetUtils>
+{
+    using Singleton<SubnetUtils>::Singleton;
+
+    [[nodiscard]] virtual Subnet generate_random_subnet(uint8_t cidr = 24,
```
xmkg: prefix_length. Also the name could be more descriptive: e.g., `random_subnet_from_range`

include/multipass/utils.h
```diff
@@ -254,6 +254,8 @@ class Utils : public Singleton<Utils>
     virtual bool is_ipv4_valid(const std::string& ipv4) const;
 
     virtual Path default_mount_target(const Path& source) const;
+
+    virtual intmax_t random_int(intmax_t a, intmax_t b) const;
```
xmkg: Any reason for going with intmax_t here?

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+                                           cidr,
+                                           range.as_string()));
+
+    // ex. 2^(24 - 16) = 256, [192.168.0.0/24, 192.168.255.0/24]
+    const size_t possibleSubnets = std::size_t{1} << (cidr - range.CIDR());
+
+    // narrowing conversion, possibleSubnets is guaranteed to be < 2^31 (4 bytes is safe)
+    static_assert(sizeof(decltype(MP_UTILS.random_int(0, possibleSubnets))) >= 4);
+
+    const auto subnet = static_cast<size_t>(MP_UTILS.random_int(0, possibleSubnets - 1));
```
xmkg: `subnet` is a bit confusing here. How about naming it as what it is? (subnet_block_idx)

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    if (cidr >= 31)
+        throw std::invalid_argument(fmt::format(large_CIDR_err_fmt, cidr));
+
+    if (cidr < range.CIDR())
+        throw std::logic_error(fmt::format("A subnet with cidr {} cannot be contained by {}",
+                                           cidr,
+                                           range.as_string()));
+
+    // ex. 2^(24 - 16) = 256, [192.168.0.0/24, 192.168.255.0/24]
```
xmkg: 2^16 is 65536, which is the number of /32 networks that can fit into the 192.168.0.0/16 private IPv4 block. The number of /24 networks is (2^16) / (2^8) = 256.

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+mp::IPAddress mp::Subnet::min_address() const
+{
+    return id + 1;
+}
+
+mp::IPAddress mp::Subnet::max_address() const
+{
+    // identifier + 2^(32 - cidr) - 1 - 1
+    return id + ((1ull << (32ull - cidr)) - 2ull);
```
xmkg: It would be nice to explain why we're subtracting 1 (network address) and another 1 (the broadcast address) here. (https://www.rfc-editor.org/rfc/rfc950.html):

```
            "In certain contexts, it is useful to have fixed addresses
            with functional significance rather than as identifiers of
            specific hosts.  When such usage is called for, the address
            zero is to be interpreted as meaning "this", as in "this
            network".  The address of all ones are to be interpreted as
            meaning "all", as in "all hosts".  For example, the address
            128.9.255.255 could be interpreted as meaning all hosts on
            the network 128.9.  Or, the address 0.0.0.37 could be
            interpreted as meaning host 37 on this network."
```

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+uint32_t mp::Subnet::address_count() const
```
xmkg: This could be named as usable_address_count since it does not include the network address and broadcast address, which are valid addresses in the subnet.

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+mp::Subnet::Subnet(IPAddress ip, uint8_t cidr) : id(apply_mask(ip, cidr)), cidr(cidr)
+{
+    if (cidr >= 31)
+    {
+        throw std::invalid_argument(fmt::format(large_CIDR_err_fmt, cidr));
+    }
+}
+
+mp::Subnet::Subnet(const std::string& cidr_string) : Subnet(parse(cidr_string))
+{
```
xmkg, line 81: As insurance, it'd also be nice to validate that the given subnet.network_addr() is _canonical_, (i.e., the last N bits are 0, depending on the prefix_length).

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+mp::Subnet mp::SubnetUtils::generate_random_subnet(uint8_t cidr, Subnet range) const
+{
+    // @TODO don't rely on pure randomness
+    for (auto i = 0; i < 100; ++i)
+    {
+        const auto subnet = ::generate_random_subnet(cidr, range);
+        if (MP_PLATFORM.subnet_used_locally(subnet))
+            continue;
+
+        if (MP_PLATFORM.can_reach_gateway(subnet.min_address()))
```
xmkg: What if the gateway is temporarily unreachable?

src/network/subnet.cpp
```diff
@@ -0,0 +1,219 @@
...
+    return (ip_res == 0) ? other.cidr <=> cidr : ip_res;
+}
+*/
+
+mp::Subnet mp::SubnetUtils::generate_random_subnet(uint8_t cidr, Subnet range) const
+{
+    // @TODO don't rely on pure randomness
+    for (auto i = 0; i < 100; ++i)
+    {
+        const auto subnet = ::generate_random_subnet(cidr, range);
```
xmkg: I _think_ instead of doing a *spray and pray*, we need to do some bookkeeping about the subnets we're lending, meaning:

- Using a *configurable* subnet range, and asking the user to specify a subnet range during the installation while providing a *sensible* default
- Determining minimums for both network count and host count so we can ask the user to give a subnet range that has plenty of space
- Handing out the subnets deterministically, and persisting which address spaces are used by whom

This will eliminate the need for heuristic probing and make it a hard contract that the user is responsible for providing a viable, unused address range if the default range already collides with the user's existing networks.

xmkg (changes requested):
First pass of review -- the overall shape is looking good! 

I have comments about the terminology, naming, and potential issues of lending out random subnets relying on heuristics.

include/multipass/utils.h
xmkg: https://thephd.dev/intmax_t-hell-c++-c

src/network/subnet.cpp
Sploder12: Ah, seems I misunderstood the naming, whoops. Good catch, thanks!

src/network/subnet.cpp
Sploder12: It was to keep the thrown exceptions consistent but using a custom exception type makes it unnecessary

src/network/subnet.cpp
Sploder12: What would make sense to do in case of a non-canonical network address? Throw an error, log a message? `ifconfig` (I intended to use for MacOS) returns canonical + 1 instead of the canonical for its subnets but if we stop relying on heuristics that'd no longer matter. Window's `Get-NetIPAddress` returns an IP that is assigned to the network interface inside its subnet

xmkg: I see two possible avenues:

a) Throw an exception and make the canonicalization the caller's responsibility
b) Subnet class canonicalizes the inputs by default 

I think option `b` might be the most convenient one for the use case you described. I'll leave it up to you.

include/multipass/utils.h
Sploder12: Can't use templates because of the virtual so I wanted to use the largest signed integer type. I didn't know `intmax_t` had so many issues. Also seems `uniform_int_distribution` is pretty strict with its allowed integral types so `long long` would probably be best here i think.

src/network/subnet.cpp
Sploder12: I like the idea but with how the current masking is only done in one place that is not part of an exposed interface I don't think it'd provide much benefit for the slight increase of complexity

src/network/subnet.cpp
Sploder12: Then a conflicting subnet is returned

src/network/subnet.cpp
Sploder12: Definitely need a better way of tracking and generating as you say, AZs track their own subnet but that isn't checked directly (possibly indirectly) in the generation of new subnets. I personally like your idea but anything user interaction probably needs eyes from design on it and would probably be out of scope for this PR.

xmkg: A relevant issue: https://github.com/canonical/multipass/issues/3147

src/network/subnet.cpp
```diff
@@ -0,0 +1,199 @@
...
+            throw mp::Subnet::PrefixLengthOutOfRange(prefix_length);
+
+        return mp::Subnet(addr, prefix_length);
+    }
+    throw std::invalid_argument(
+        fmt::format("CIDR {:?} does not contain '/' seperator", cidr_string));
+}
+catch (const mp::Subnet::PrefixLengthOutOfRange&)
+{
+    throw;
```
xmkg, line 51: Why are we catching it if we're just going to throw it again?

Sploder12: `mp::Subnet::PrefixLengthOutOfRange` is a subclass of `std::out_of_range` so we don't want to double wrap it. `std::out_of_range` is needed because of `std::stoull`

include/multipass/utils.h
```diff
@@ -254,6 +254,8 @@ class Utils : public Singleton<Utils>
     virtual bool is_ipv4_valid(const std::string& ipv4) const;
 
     virtual Path default_mount_target(const Path& source) const;
+
+    virtual long long random_int(long long a, long long b) const;
```
xmkg: maybe [[nodiscard]] ?

src/network/subnet.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <fmt/format.h>
+
+#include <algorithm>
+#include <limits>
+
+namespace mp = multipass;
+
+namespace
+{
+mp::Subnet parse(const std::string& cidr_string)
```
xmkg: [[nodiscard]]

src/network/subnet.cpp
xmkg: Well, what we could do is provide a *sensible* default now, and introduce the customization later. We can try to be smart about it by checking which blocks popular Container/VM/VPN technologies prefer by default (e.g., Docker, LXD, VirtualBox, OpenVPN). User issues could be a reference on which blocks to avoid.

src/network/subnet.cpp
xmkg: ahh right, that makes sense!

src/network/subnet.cpp
Sploder12: Sounds good, I'll look into it!

Sploder12: It looks like 192.168.0.0/16 is out of the question since it's too small and populated, Docker uses 172.16.0.0/12 by default. That leaves the 10.0.0.0/8 range which we already use. The low [10.0.0.0/16, 10.10.0.0/16] range seems pretty heavily used. The approach differs between applications, for example, Docker does not check for conflicts, LXD uses almost the exact same code we already do...

With the current AZs we only need 3 subnets, but I imagine that will increase in the future, my initial thought is 32 is plenty. For VMs, 62 seems reasonable per subnet. that would give us a minimum range of /21 allocating /26 subnets.

The random number generator decided 10.97.0.0/16 as a sensible default, we can hand out 256 /24 subnets from this range. Although, I'd want to get other opinions before removing the existing heuristics.

include/multipass/subnet.h
```diff
@@ -0,0 +1,119 @@
...
+    [[nodiscard]] IPAddress subnet_mask() const;
+
+    [[nodiscard]] std::string to_cidr() const;
+
+    // Subnets are either disjoint or the smaller is a subset of the larger
+    [[nodiscard]] bool contains(Subnet other) const;
+    [[nodiscard]] bool contains(IPAddress ip) const;
+
+    [[nodiscard]] std::strong_ordering operator<=>(const Subnet& other) const;
+    [[nodiscard]] bool operator==(const Subnet& other) const = default;
```
xmkg, line 82: Are there any reasons to explicitly define the equality operator?

Sploder12: `operator<=>` does not automatically generate the equality operators (unless defaulted which is not the case here)

xmkg: You're right, and the reasons for not defaulting == when it's absent are beyond me.  I do understand implementing the equality in terms of <=> might be expensive, but at least use it as a default when the user does not care... The user would've overridden the ==  operator already if that was a concern. ... sigh ...

xmkg (approved):
LGTM :+1: Thanks @Sploder12!

jimporter (approved):

PR #4399, opened 2025-10-01
[docs] Fix cross-references in how-to guides

geoffreynyaga:
The current implementation of internal linking to other pages/content across the documentation set has been largely relative links. 

Whilst this works and the link check CI passes, it does not scale well over time. Usong the recommend Myst referencing standard is positive because of the following reasons:

1. The links will still work even in situations where the linked markdown page/file gets renamed or moved to a new folder
2. The user gets redirected to the exact heading in the destination page
3. Standardized naming of heading references.

MULTI-2157

### DIFF
--- a/docs/how-to-guides/index.md
+++ b/docs/how-to-guides/index.md
@@ -23,7 +23,7 @@ Multipass allows you to create Ubuntu instances with a single command. As your n
 - [Add a network to an existing instance](manage-instances/add-a-network-to-an-existing-instance)
 - [Configure static IPs](manage-instances/configure-static-ips)
 - [Use a blueprint](manage-instances/use-a-blueprint)
-- [Use the Docker blueprint](manage-instances/use-the-docker-blueprint)
+- {ref}`how-to-guides-manage-instances-use-the-docker-blueprint`
 - [Run a Docker container in Multipass](manage-instances/run-a-docker-container-in-multipass)
 
 ## Customise Multipass
--- a/docs/how-to-guides/manage-instances/share-data-with-an-instance.md
+++ b/docs/how-to-guides/manage-instances/share-data-with-an-instance.md
@@ -1,7 +1,9 @@
 (how-to-guides-manage-instances-share-data-with-an-instance)=
 # Share data with an instance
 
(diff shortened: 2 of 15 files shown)


docs/how-to-guides/index.md
```diff
@@ -23,7 +23,7 @@ Multipass allows you to create Ubuntu instances with a single command. As your n
 - [Add a network to an existing instance](manage-instances/add-a-network-to-an-existing-instance)
 - [Configure static IPs](manage-instances/configure-static-ips)
 - [Use a blueprint](manage-instances/use-a-blueprint)
-- [Use the Docker blueprint](manage-instances/use-the-docker-blueprint)
+- {ref}`how-to-guides-manage-instances-use-the-docker-blueprint`
```
sharder996, line 26: Just curious why the different format?

docs/how-to-guides/manage-instances/share-data-with-an-instance.md
```diff
@@ -1,15 +1,17 @@
 (how-to-guides-manage-instances-share-data-with-an-instance)=
 # Share data with an instance
 
-> See also: [Instance](/explanation/instance), [Mount](/explanation/mount), [ID mapping](/explanation/id-mapping), [`launch`](/reference/command-line-interface/launch), [`mount`](/reference/command-line-interface/mount), [`umount`](/reference/command-line-interface/umount), [`transfer`](/reference/command-line-interface/transfer)
+``` {seealso}
```
sharder996, line 4: There is an inconsistent approach to defining these "See also" blocks. Any plans on unifying them?

docs/how-to-guides/index.md
geoffreynyaga: The page has deprecated content "**Use the Docker blueprint (deprecated)**".  We can either hardcode the string '(deprecated)' into the title or use the second `{ref}` option that directly reads the file's header title and hence captures the deprecation message. We also use the {ref} syntax on pages that we know will eventually change or has constant edits.

docs/how-to-guides/manage-instances/share-data-with-an-instance.md
geoffreynyaga: I have started to slowly implement the ``` {seealso} ... ``` syntax. Its a cleaner implementation, very aesthetic and occupies less space.

sharder996: Ok, as long as you're aware and slowly unifying them.

docs/how-to-guides/index.md
sharder996: Thanks for the explanation!

Sploder12 (approved):
LGTM!

PR #4408, opened 2025-10-06
Remove usage of deprecated libssh functions

### LINKED ISSUE #4227: Warning Report: Multiple deprecated function warnings in libssh usage
During compilation of the project, a significant number of warnings are generated due to the usage of deprecated functions from the libssh library. These deprecated APIs include functions such as `publickey_from_file`, `privatekey_from_file`, `privatekey_free`, `ssh_channel_get_exit_status`, `ssh_channel_select`, `ssh_scp_free`, `ssh_scp_close`, `ssh_scp_push_file64`, `ssh_scp_accept_request`, `sftp_server_init`, among others.

**The warnings appear mostly in the files:**

* `3rd-party/libssh/libssh/src/legacy.c`
* `3rd-party/libssh/libssh/src/scp.c`
* `src/sshfs_mount/sftp_server.cpp`
* `src/ssh/ssh_client.cpp`

**Sample warning messages:**

```
warning: ‘publickey_from_file’ is deprecated [-Wdeprecated-declarations]
warning: ‘privatekey_from_file’ is deprecated [-Wdeprecated-declarations]
warning: ‘privatekey_free’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_channel_get_exit_status’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_channel_select’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_scp_free’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_scp_close’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_scp_push_file64’ is deprecated [-Wdeprecated-declarations]
warning: ‘ssh_scp_accept_request’ is deprecated [-Wdeprecated-declarations]
warning: ‘sftp_server_init’ is deprecated [-Wdeprecated-declarations]
```

tobe2098:
The main purpose of this PR is to replace the deprecated functions in libssh, which are unmantained, by our own code, since those functions do not have replacements. This gives us control over those sections of code and allows us to maintain them. 

In terms of tests, the changes were as minimal as possible without changing outside the PREMOCK framework that is used for all libssh functions.

Some of the deprecated warnings came from the building of libssh itself, so those warnings are silenced at the submodule level.

Fixes #4227

MULTI-2243


tests/test_ssh_client.cpp
```diff
@@ -63,17 +67,23 @@ TEST_F(SSHClient, execMultipleCommandsReturnsOKNoFailure)
...
 
-    REPLACE(ssh_channel_get_exit_status, [&failure_code](auto) { return failure_code; });
+    REPLACE(ssh_channel_get_exit_state, [](ssh_channel_struct*, unsigned int* val, char**, int*) {
+        *val = 127;
+        return SSH_OK;
+    });
 
-    EXPECT_EQ(client.exec({"foo"}), failure_code);
+    EXPECT_EQ(client.exec({"foo"}), 127);
 }
```
tobe2098, line 89: Unfortunately LLVM and clang diverge on the treatment of lambda captured constants. See https://stackoverflow.com/questions/52416362/unused-lambda-capture-warning-when-capture-is-actually-used

3rd-party/libssh/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ set(BIN_INSTALL_DIR bin)
 if(LINUX)
     set(BUILD_SHARED_LIBS OFF)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+elseif(MSVC)
+    set(BUILD_SHARED_LIBS OFF)
```
tobe2098: To be able to build with libssh internal C functions that are not exposed via LIBSSH_API, it is necessary to not build shared libs in windows.

tobe2098 (commented):
Built and tested in Windows and Linux.

3rd-party/libssh/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ set(BIN_INSTALL_DIR bin)
...
 else()
     # This applies to other platforms (e.g., Windows, macOS)
```
Sploder12, line 41: This comment is out of date

src/ssh/ssh_client.cpp
```diff
@@ -154,5 +154,18 @@ int mp::SSHClient::exec_string(const std::string& cmd_line)
...
+
+int mp::SSHClient::ssh_channel_get_exit_status(ssh_channel channel)
+{
+    uint32_t exit_status = (uint32_t)-1;
+
+    int return_code{ssh_channel_get_exit_state(channel, &exit_status, NULL, NULL)};
+    if (return_code != SSH_OK)
+    {
+        return SSH_ERROR;
+    }
```
Sploder12: Slight change in behavior from original, if `return_code == SSH_AGAIN` (no exit status has been returned yet) then it used to `return -1`

src/sshfs_mount/sftp_server.cpp
```diff
@@ -55,11 +59,30 @@ enum Permissions
 auto make_sftp_session(ssh_session session, ssh_channel channel)
 {
     mp::SftpServer::SftpSessionUptr sftp_server_session{sftp_server_new(session, channel),
-                                                        sftp_free};
-    mp::SSH::throw_on_error(sftp_server_session,
-                            session,
-                            "[sftp] server init failed",
-                            sftp_server_init);
```
Sploder12, line 62: From libssh [source](https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/sftp.c) `/* @deprecated in favor of sftp_server_new() and callbacks based sftp server */`, maybe we should move to a callbacks based sftp server? Although, that'd require a ton more work with little benefit.

I'm not really a fan of including the private header, could make updates painful in the future. Also reimplementing the deprecated functions doesn't seem much better than just defining `SSH_SUPPRESS_DEPRECATED` IMO. I don't know, I'd like to get other's thoughts on this

src/sshfs_mount/sftp_server.h
```diff
@@ -53,7 +53,7 @@ class SftpServer
     void stop();
 
     using SSHSessionUptr = std::unique_ptr<ssh_session_struct, decltype(ssh_free)*>;
-    using SftpSessionUptr = std::unique_ptr<sftp_session_struct, decltype(sftp_free)*>;
+    using SftpSessionUptr = std::unique_ptr<sftp_session_struct, decltype(sftp_server_free)*>;
```
Sploder12, line 56: good catch!

tests/test_sftpserver.cpp
```diff
@@ -346,17 +365,16 @@ TEST_F(SftpServer, sshfsRestartsOnTimeout)
...
-
-    auto get_client_msg = [this, &num_calls](auto...) {
+    auto message{make_msg(SSH_FXP_INIT)};
+    auto get_client_msg = [this, &num_calls, &message, calls = 0](auto...) mutable {
         exit_status_mock.set_ssh_rc(SSH_OK);
         exit_status_mock.set_exit_status(num_calls == 1 ? exit_status_mock.failure_status
                                                         : exit_status_mock.success_status);
 
-        return nullptr;
+        return (calls++ & 1) ? nullptr : message.get();
```
Sploder12: this check seems odd (pun intended)

tests/test_ssh_client.cpp
Sploder12: could do `static constexpr int failure_code{127};` which would remove the need to capture and thus solve the issue

tests/test_sshfsmount.cpp
```diff
@@ -203,6 +203,25 @@ struct SshfsMount : public mp::test::SftpServerTest
...
+
+    auto mock_sftp_get_cli_msg(sftp_client_message message)
+    {
+        return [message, calls = 0](sftp_session) mutable {
+            if (calls <= 0)
+            {
+                calls++;
+                return message;
+            }
+            return static_cast<sftp_client_message>(nullptr);
```
Sploder12: would `return std::exchange(message, nullptr);` work here?

src/ssh/ssh_client.cpp
tobe2098: I see. The function is identical to the libssh one by implementation, what you are mentioning I think is already covered: If return_code != 0 (which includes SSH_AGAIN), it returns SSH_ERROR (defined as -1). So in theory this should still follow the original behavior. Is this what you meant?

src/sshfs_mount/sftp_server.cpp
tobe2098: I agree, but the comment plus the commit message leave us with little information. It is not ideal to reimplement, but at least now those functions are "maintained" now. I would be happy to work on moving to the callbacks based sftp server, and that was what I initally researched, but I found no existing documentation on that option. If anyone finds anything I will be happy to take a look at what changes are necessary.

In the case of get_exit_state, there is no other option really. Until we deal with SSH return codes with more consistency, reimplementing seemed like the only viable option.

tests/test_sftpserver.cpp
tobe2098: I know, but I reached it through trial and error. In the test it seems like there are 2 sftp server inits with a message_get in between, that is why the lambda looks so odd :D

tests/test_sshfsmount.cpp
tobe2098: Good catch! Looks like it should, I will test.

src/ssh/ssh_client.cpp
Sploder12: Ah yes it is, my bad. The [documentation](https://api.libssh.org/master/group__libssh__channel.html#ga9eea019dd0bbaa8a817fff2c762d1a2d) says "Return The exit status, -1 if no exit status has been returned (yet), or SSH_ERROR on error." which made me think there was a difference between `SSH_ERROR` and `-1`

src/sshfs_mount/sftp_server.cpp
Sploder12: Yeah the documentation is pretty much nonexistent for sftp server functionality from what I can tell. I could only find [this](https://git.libssh.org/projects/libssh.git/tree/src/sftpserver.c) example implementation but that itself isn't well commented or documented

tobe2098: Yeah, apart from the default subsystem request and default data callback... besides, it does say that those two are not supported in windows (ifdef + comment), so I do not think they are an option in this case. Sadly, until we do more research or someone finds some documentation we will have to go with the reimplementations. Is it possible that existing sftp servers all rely on the deprecated functions or old versions?

Sploder12: Our implementation is a pretty popular reference implementation so it's possible

tobe2098:
After some research, I have found in Gitlab the following libssh mirror that does implement an sample_sftpserver.c: https://gitlab.com/libssh/libssh-mirror/-/blob/master/examples/sample_sftpserver.c And in the PR they did they comment more on it. From a first (inexperienced) look, moving to callbacks may require a major refactor. Now at least we have a working (?) example.

I am unsure if this is a fork of our libssh or viceversa, but I think it would be nice to have this example.
Original PR that created example:
https://gitlab.com/libssh/libssh-mirror/-/merge_requests/263/diffs#cadf2ffbd25228d28ef40122660aa3c8ffb7d0d4
Merged PR:
https://gitlab.com/libssh/libssh-mirror/-/merge_requests/340/diffs

src/sshfs_mount/sftp_server.cpp
jimporter: Perhaps just add a comment here explaining the rationale and linking to an issue for moving to a callback-based version?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -55,11 +59,30 @@ enum Permissions
...
+                                                        sftp_server_free};
+
+    /* handles setting the sftp->client_version */
+    sftp_client_message msg{sftp_get_client_message(sftp_server_session.get())};
+    if (msg == NULL)
```
jimporter: Tiny nitpick:
```suggestion
    if (msg == nullptr)
```

tests/test_sftpserver.cpp
jimporter: How is `calls` different from `num_calls`? Could we use the latter value here? It looks like that would match what's happening with `exit_status_mock.set_exit_status` just above.

src/sshfs_mount/sftp_server.cpp
tobe2098: Good catch!

tobe2098:
On the test, the solution is not very clean. The execution order of the functions is exec->init->msg->exec->exec(not sshfs)->init->msg, (where after the PR init becomes a different msg) so a single index cannot account for behavior of both functions very cleanly

jimporter:
> On the test, the solution is not very clean. The execution order of the functions is exec->init->msg->exec->exec(not sshfs)->init->msg, (where after the PR init becomes a different msg) so a single index cannot account for behavior of both functions very cleanly

Perhaps pass in a queue and pop the front element off each time? That would be similar to how [`side_effect`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect) works in Python mocks.

tobe2098:
The queue does not seem cleaner than just checking `num_calls` against 2 and 5 like in the solution I pushed yesterday. I found a shortcut to make it cleaner than just checking against every number, since 2 and 5 share that `(num_calls + 1)%3==0`. Does this solution follow what you had in mind about `calls` vs`num_calls`?

PD: I also added a new test to cover the termination signal logging.

tobe2098:
ssh_init had to be added to the windows daemon and to the client due to static linking. It could be excluded from linux and macos where it is not necessary via macros, but it has no negative effects (as tested in linux).

3rd-party/libssh/CMakeLists.txt
xmkg: Instead of disabling the shared lib altogether, would something like [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) work here?

CMakeLists.txt
```diff
@@ -309,8 +309,10 @@ if(MSVC)
   add_definitions(-DWIN32_LEAN_AND_MEAN)
   set(MULTIPASS_BACKENDS hyperv virtualbox)
   set(MULTIPASS_PLATFORM windows)
+  target_compile_options(ssh PRIVATE /wd4996)
```
xmkg: Would it be possible to move this closer to the target's origin (i.e., "ssh")?

CMakeLists.txt
```diff
@@ -309,8 +309,10 @@ if(MSVC)
...
 else()
   add_compile_options(-Werror -Wall -pedantic -fPIC -Wno-error=deprecated-declarations)
+  target_compile_options(ssh PRIVATE -Wno-deprecated-declarations)
```
xmkg: idem

src/client/cli/client.cpp
```diff
@@ -83,6 +83,8 @@ mp::Client::Client(ClientConfig& config)
       term{config.term},
       aliases{config.term}
 {
+    ssh_init();
```
xmkg: Is there a reason for re-initializing the library again here? Wouldn't doing it in the `main` suffice?

src/daemon/daemon_main_win.cpp
```diff
@@ -127,6 +127,8 @@ void create_client_cert_if_necessary()
 
 int daemon_main(int argc, char* argv[], RegisterConsoleHandler register_console)
 {
+    // Due to static linking libssh must be initalized
+    ssh_init();
```
xmkg: Are there any disadvantages or "gotchas" for calling this for the dynamically linked libssh?

xmkg: Also, would it make sense to move this to the platform-agnostic main?

src/client/cli/client.cpp
tobe2098: Initialization in main for daemon, but client also requires initialization as it is a different process (tested, probably due to static linking behavior). Client requires init, but since the client can exit due to exceptions, putting it in the constructor and destructor seemed the best option. Will be removed if the [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) works.

src/daemon/daemon_main_win.cpp
tobe2098: I do not know, since I cannot test on macos, but my guess is that nothing happens (ssh_init is idempotent, and everything should have been initalized in dynamic libraries). In linux it is different because there is an implicit initialization of statically linked libraries (.init_array). As far as I know the daemon does not have a platform-agnostic main, there is windows and there is the rest. On linux (as said above) and macos (dynamically linked), the init is not necessary. Is this what you meant?

tobe2098:
In the end, the CMake flag you suggested works, so I am closing the related threads. Thanks!

ricab:
Hey @Sploder12, did you still want to review this one?

Sploder12 (approved):
LGTM!

PR #4417, opened 2025-10-07
[bash] Avoid using sed in bash major version check

### LINKED ISSUE #3267: BASH_VERSION=5.1.16(1)-release causes syntax error: invalid arithmetic operator (bash_complete / homebrew / macOS)
Homebrew on my macOS (12.7 Monterey) has upgraded bash to `5.1.16(1)-release`, which sets `BASH_VERSION=5.1.16(1)-release` in the environment.

```
$ bash --version
GNU bash, version 5.1.16(1)-release (x86_64-apple-darwin20.6.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

$ echo "$BASH_VERSION"
5.1.16(1)-release
```

Under said bash version the bash_complete script included in the Multipass Homebrew installation emits an error when sourced that looks like

```
-bash: [[: 5.1.16(1)-release: syntax error: invalid arithmetic operator (error token is ".1.16(1)-release")
```

I traced it down to the following line:

https://github.com/canonical/multipass/blob/ef40bb593eb7db4e76969be303f54733598d2aa5/completions/bash/multipass#L370

Quoting the use of `$BASH_VERSION` resolves the error on my system, like

```bash
if [[ $(echo "${BASH_VERSION}" | sed s/\\..*//g) -gt 3 ]]; then
    OLD_BASH=false
else
    OLD_BASH=true
fi
```

I see that Homebrew has bash `5.2.15` available, so upgrading bash may resolve the issue, but I wanted to pass on the information in case this is a general case of "should be quoted" (I'm no bash expert), and for anyone else fruitlessly trying to search for said error.

levkropp:
This pull request updates the Bash version check logic in the `_multipass_complete()` function to use a more robust method for extracting the major version number. This change replaces the previous `sed`-based extraction of the major version from `BASH_VERSION` with Bash parameter expansion, storing the result in `bash_major_version` for improved clarity and robustness. (`completions/bash/multipass`)

resolves #3267

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -463,7 +463,8 @@ _multipass_complete()
 
 # Check the Bash version. Bash < 4 does not have `compopt`, which we use for completing `launch --network`. We
 # disable some completions on that command based on this test.
-if [[ $(echo ${BASH_VERSION} | sed s/\\..*//g) -gt 3 ]]; then
+bash_major_version=${BASH_VERSION%%.*}
+if [[ $bash_major_version -gt 3 ]]; then
     OLD_BASH=false
 else
     OLD_BASH=true


completions/bash/multipass
```diff
@@ -463,7 +463,8 @@ _multipass_complete()
 
 # Check the Bash version. Bash < 4 does not have `compopt`, which we use for completing `launch --network`. We
 # disable some completions on that command based on this test.
-if [[ $(echo ${BASH_VERSION} | sed s/\\..*//g) -gt 3 ]]; then
+bash_major_version=${BASH_VERSION%%.*}
+if [[ $bash_major_version -gt 3 ]]; then
```
jimporter: I think we can use [`BASH_VERSINFO`](https://www.gnu.org/software/bash/manual/bash.html#index-BASH_005fVERSINFO) here:

```suggestion
if [[ ${BASH_VERSINFO[0]} -gt 3 ]]; then
```

From checking the [Bash 3.2 manual](https://bash.cyberciti.biz/bash-reference-manual/index.html), that variable was present back then too, so we should be ok. (And even if it was missing from some still older version, it'd expand to `""`, so I think we should be ok no matter what.)

PR #4419, opened 2025-10-07
[yaml] Use DoubleQuoted for octals instead of LocalTag

### LINKED ISSUE #4418: Cloud-init permissions field usage leads to invalid yaml blob
**Describe the bug**
Using a permission field in write_files results in malformed tag in the final YAML. The files do not get written because of the error. I faced this on the latest dev-build on main, and the related issue #4176 is already fixed over there.

VM's `/var/log/cloud-init-output.log`contains these warnings between the network information and ssh-key generation:

```
2025-10-07 15:37:02,788 - util.py[WARNING]: Failed loading yaml blob. Invalid format at line 31 column 18: "could not determine a constructor for the tag '!str'
  in "<unicode string>", line 31, column 18:
        permissions: !str 0755
                     ^"
2025-10-07 15:37:02,797 - util.py[WARNING]: Failed loading yaml blob. Invalid format at line 31 column 18: "could not determine a constructor for the tag '!str'
  in "<unicode string>", line 31, column 18:
        permissions: !str 0755
                     ^"
2025-10-07 15:37:02,797 - cloud_config.py[WARNING]: Failed at merging in cloud config part from part-001: empty cloud config
```

Seems like it is missing an exclamation mark for the [explicit tag](https://yaml.org/spec/1.2.2/#tags), or something else. Anyways, I'm not sure if it has to be an explicit tag. The original permission field is untagged.

**To Reproduce**
1. Follow "Steps to Reproduce" steps 1-3 from #4176. Step 3 should not return an error (fixed in #4176).
2. Run
    ```
    sudo cat /var/lib/cloud/instances/cdo/cloud-config.txt
    ```
    It should output:
    ```
    #cloud-config

    # Cloud-config part ignored SCHEMA_ERROR: part-001
    --- {}
    ...
    ```

**Expected behavior**
No errors. Permissions should get parsed correctly and the files should be written with the according permissions.

**Logs**
Unable to, sorry. It's company's computer.

**Additional info**
- OS: Windows 10 Pro, Version 10.0.19045 Build 19045
- `multipass version`
  ```
  multipass   1.17.0-dev.498.ci1400+gc0f5fce6.win
  multipassd  1.17.0-dev.498.ci1400+gc0f5fce6.win
  ``` 
- `multipass info`
  ```
  Name:           cdo
  State:          Running
  Snapshots:      0
  IPv4:           172.23.142.149
  Release:        Ubuntu 24.04.3 LTS
  Image hash:     02dc186dc491 (Ubuntu 24.04 LTS)
  CPU(s):         4
  Load:           0.00 0.00 0.00
  Disk usage:     1.9GiB out of 48.4GiB
  Memory usage:   419.2MiB out of 7.8GiB
  Mounts:         --
  ```
  `hyperv`

**Additional context**

levkropp:
This pull request updates the YAML emission logic to use double quotes for strings that look like octal numbers, rather than attempting to them explicitly as strings. Corresponding unit tests have also been updated to reflect this new behavior.

YAML::DoubleQuoted was originally avoided in favor of explicit tags due to perceived issues with timeouts using DoubleQuoted, but it appears that this is not a result of the use of DoubleQuoted (loading some cloud-init YAMLs simply takes a long time!)

fixes #4418

### DIFF
--- a/src/utils/yaml_node_utils.cpp
+++ b/src/utils/yaml_node_utils.cpp
@@ -131,7 +131,7 @@ std::string mp::utils::emit_yaml(const YAML::Node& node)
                 if (value.length() >= 2 && value[0] == '0' &&
                     std::all_of(value.begin() + 1, value.end(), ::isdigit))
                 {
-                    emitter << YAML::LocalTag("str") << value;
+                    emitter << YAML::DoubleQuoted << value;
                     break;
                 }
             }
--- a/tests/test_yaml_node_utils.cpp
+++ b/tests/test_yaml_node_utils.cpp
@@ -178,12 +178,10 @@ TEST(UtilsTests, emitYamlWithOctalString)
 
     const std::string result = mpu::emit_yaml(node);
 
-    // The octal strings should be tagged as strings to preserve their format
-    EXPECT_TRUE(result.find("permissions: !str \"0755\"") != std::string::npos ||
-                result.find("permissions: !str 0755") != std::string::npos);
-    EXPECT_TRUE(result.find("another_permission: !str \"0644\"") != std::string::npos ||
-                result.find("another_permission: !str 0644") != std::string::npos);
-    // Non-octal strings should not be tagged
+    // The octal strings should be double-quoted to preserve their format as strings
+    EXPECT_TRUE(result.find("permissions: \"0755\"") != std::string::npos);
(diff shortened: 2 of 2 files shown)


jimporter (approved):
This makes sense to me. I don't know if I fully understand what the timeout bug was though, so I'm not sure how to test this out locally.

PR #4420, opened 2025-10-08
[gui] Adding an about section in the Settings tab

### LINKED ISSUE #4201: Show "About" in the GUI window
The GUI tray icon has an "about" section, which shows the version. It would be nice for this to be present in the window too.

Perhaps in the help section? That page is pretty empty and I think it's common to include an about entry in help menus.

**Additional context**

Image

tobe2098:
The PR adds an About section, similar to the one found in the tray menu, that shows the multipass and multipassd versions and the Canonical copyright. Tested in Ubuntu and Windows 11.

Fixes #4201
MULTI-2294


levkropp:
Hi @tobe2098 , is this PR ready for a first round of review? Just some quick comments from looking at this quickly

image

It looks like you followed the tray menu's style exactly, but I think it would look better if `multipass version` and `multipassd version` followed the style of the other settings on the setting page (same font and color as Driver, Bridged network, written as `Multipass version` and `Multipass daemon version` instead, etc)

I think the version text would look nicer if it was either right-justified or centered. I'm slightly preferring centered since that aligns it with the Click to Copy tooltip and make it more obvious that this is not a setting you can change

Since we're not limited to the ASCII of a tray menu, instead of (C) we should probably just use the actual copyright symbol ©

---

Food for thought:

Why do we show both the multipass and multipassd versions to end users in the GUI on the tray menu and now here? In 99% of normal use cases they will be the same. It might be an interesting idea to always show just `Multipass version` and if there's a mismatch between client and daemon, THEN show the daemon version too, or even have a popup on GUI startup if a version mismatch is detected

tobe2098:
It is ready, thank you for the review. I think you are correct, homogeneous style is better. 

The different versions part is only really useful for us, since we can have the daemon running, rebuild and think that we are already in the new build fully, even though the daemon is still in the previous version. I agree that only showing it when it is different makes more sense, maybe with some warning color. The additional pop-up also makes sense, but I would keep the About section part. We can also apply the mismatch policy to the TrayMenu as well, since it has the same implications/uses. What do you think?

levkropp:
This is looking a lot better, thanks @tobe2098 !

image

My only nitpick is to remove the colon from `Multipass version` since we dont use colons for the rest of the headings. What do you think @amylily1011 ?

amylily1011:
LGTM! +1 with @levkropp on the colon.

tobe2098:
Done!

tobe2098:
That is somewhat strange, my tray does work properly in Ubuntu. 
image
In which environment are you testing this?

levkropp:
> That is somewhat strange, my tray does work properly in Ubuntu. In which environment are you testing this?

I was testing this on KDE with Kubuntu 24.04, so it is entirely possible that this issue is with KDE and how Flutter handles tray icons on KDE. I will test these changes on GNOME now as well

levkropp:
Alright so I tested it on GNOME as well, and the issue actually is that if you turn off the daemon and the GUI spins with "waiting for daemon", and THEN you run a new daemon with a different version than the client, the Settings tab updates to show both multipass and daemon version but the tray menu doesn't. This is probably because we're re-checking the provider for the multipass version for the settings page but not doing something similar for tray menu

tobe2098:
I'll see if I can fix it. I was testing the other way around, that is why I did not see it.

tobe2098:
This should keep the position consistent while having the same behavior as in the GUI

tobe2098:
I am somewhat interested: In what sense does it not work? Does it still show up always or is it disordered?

tobe2098:
What do you think @amylily1011, do you find the new changes to the version reporting in the GUI agreeable?

tobe2098:
No worries there. It can wait, it is nothing critical

amylily1011:
The UI LGTM! Thanks for the update 😃

PR #4422, opened 2025-10-09
[network] Bridges with long names use hash suffix

### LINKED ISSUE #2158: Mapping of interface name to bridge name not injective
**Describe the bug**
Multipass can't create a bridge to two interfaces named `eth123456789abc` and `eth123456789xyz`, because it would derive the same bridge name for both: `br-eth123456789`. 

**Additional info**
 - OS: Linux

levkropp:
This pull request introduces a new strategy for generating bridge names in the Linux backend to ensure uniqueness when interface names are long by truncating long interface names and appending a short hash. This ensures that bridge names are both readable and unique.
* Updated the bridge creation logic in `create_bridge_with` to use the new `generate_bridge_name` function, replacing the previous simple truncation method. 
* Added `QCryptographicHash` includes to both `backend_utils.cpp` and `test_backend_utils.cpp` to support the new hashing logic for bridge name generation.

Originally:

- `eth123456789abc` → `br-eth123456789` (truncated to 15 chars)
- `eth123456789xyz` → `br-eth123456789` (same name - __collision__)

With the fix:
- `eth123456789abc` → `br-eth12345-8f9` (unique hash suffix)
- `eth123456789xyz` → `br-eth12345-a2c` (different hash - __no collision__)

Short names (unchanged, fully readable):

- `eth0` → `br-eth0`
- `wlan0` → `br-wlan0`
- `enp0s3` → `br-enp0s3`

resolves #2158

### DIFF
--- a/tests/linux/test_backend_utils.cpp
+++ b/tests/linux/test_backend_utils.cpp
@@ -31,6 +31,7 @@
 #include <multipass/logging/log.h>
 #include <multipass/memory_size.h>
 
+#include <QCryptographicHash>
 #include <QMap>
 #include <QVariant>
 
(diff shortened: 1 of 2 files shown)


tests/linux/test_backend_utils.cpp
```diff
@@ -186,7 +187,33 @@ struct CreateBridgeTest : public Test
...
+        // Generate a short hash of the full interface name to ensure uniqueness
+        QCryptographicHash hash(QCryptographicHash::Md5);
+        hash.addData(child, strlen(child));
+        auto hash_result = hash.result().toHex();
+        auto hash_suffix = hash_result.left(hash_suffix_len);
+
+        // Use as much of the interface name as possible for readability
+        auto iface_portion = QString(child).left(max_iface_portion);
+
+        return QString("%1%2-%3").arg(base_name).arg(iface_portion).arg(hash_suffix);
```
ricab, line 216: This is basically a repetition of the same code. I think test expectations should be adapted instead.

ricab (changes requested):
The implementation looks good, but I don't think we should copy paste the code into tests. It shouldn't be hard to pick up and adapt.

xmkg:
Closing because we're no longer using QCryptographicHash for this.

PR #4425, opened 2025-10-10
[submodules] Update yaml-cpp to latest upstream master

levkropp:
The .gitmodules was updated to point to upstream, but the submodule commit reference was never updated. The submodule was still pointing to the old Canonical fork commit `d96aa99`. This PR updates yaml-cpp to point to the upstream master branch.

 This change is necessary for building on 25.10 because Ubuntu 25.10 ships with GCC 14 which requires an include for `cstdint` in the emitterutils.cpp file that is present in the latest yaml-cpp upstream but not in our fork


xmkg:
Closing this in favor of https://github.com/canonical/multipass/pull/4459.

PR #4434, opened 2025-10-14
[format] Stick with 100 characters in Python too

ricab:
Use 100 characters for the maximum line length in python files (same as in C++). Conforms to existing code.


ricab:
Aborting this one.

ricab:
OK, makes sense. Let's revisit together with linting in CI.

Issue #4369, opened 2025-09-18
[gui][snap] Continuous significant flickering when there are any redrawing events


tobe2098:
**Describe the bug**
The GUI flickers when there are any graphical changes in the interface. The bug occurs only in the 1.17-dev version (edge channel in snapstore). In 1.16 this is not found.

**To Reproduce**
How, and what happened?
1. `sudo snap install --channel edge multipass`
2. `multipass.gui`
3. Move the mouse around the screen or click some buttons
4. Flickering occurs

**Expected behavior**
No flickering in the GUI.

**Logs**
From the edge GUI
```
touch: cannot touch '/home/antoni/snap/multipass/common/.cache/desktop-runtime-date': No such file or directory
ln: failed to create symbolic link '/home/antoni/snap/multipass/15720/.local/share/glib-2.0/schemas/org.gtk.Settings.ColorChooser.gschema.xml': File exists
ln: failed to create symbolic link '/home/antoni/snap/multipass/15720/.local/share/glib-2.0/schemas/org.gtk.Settings.Debug.gschema.xml': File exists
ln: failed to create symbolic link '/home/antoni/snap/multipass/15720/.local/share/glib-2.0/schemas/org.gtk.Settings.EmojiChooser.gschema.xml': File exists
ln: failed to create symbolic link '/home/antoni/snap/multipass/15720/.local/share/glib-2.0/schemas/org.gtk.Settings.FileChooser.gschema.xml': File exists
/snap/multipass/15720/gnome-platform/usr/lib/x86_64-linux-gnu/glib-2.0/glib-compile-schemas: symbol lookup error: /snap/multipass/15720/gnome-platform/usr/lib/x86_64-linux-gnu/glib-2.0/glib-compile-schemas: undefined symbol: g_task_set_static_name
Cannot load module /snap/multipass/15720/gnome-platform/usr/lib/x86_64-linux-gnu/gtk-3.0/3.0.0/immodules/im-wayland.so: /snap/multipass/15720/gnome-platform/usr/lib/x86_64-linux-gnu/gtk-3.0/3.0.0/immodules/im-wayland.so: undefined symbol: g_string_free_and_steal
[...]

PR #4441, opened 2025-10-15
[submodules] Update flutter to 3.35.7

levkropp:
This PR updates our ref for our flutter submodule to upgrade to the latest stable version of Flutter which is currently 3.35.6. Switching to 3.35 should resolve https://github.com/canonical/multipass/issues/4369 according to https://github.com/flutter/flutter/issues/169508


tobe2098:
Testing the snap GUI from this PR shows the same behavior. Unless the build process somehow is still using the past version, the problem is still there. The flickering got better (not so aggressive), but it still flickers quite a bit.
[Screencast from 2025-10-16 15-01-21.webm](https://github.com/user-attachments/assets/92d4bca6-398e-40e5-a3c4-f4139e7c1680)

levkropp:
Thanks for checking @tobe2098 , are you still seeing the mesa warnings you originally reported in #4369 in the console log with this snap? can you try some of the environment variables mentioned in https://github.com/flutter/flutter/issues/169508 such as `LIBGL_ALWAYS_SOFTWARE=1 FLUTTER_LINUX_RENDERER=software` ? 

A user in https://github.com/ubuntu/app-center/issues/1947 mentioned that switching to software acceleration fixed this flutter issue for them. Maybe flutter 3.35 + forced software acceleration for mesa would resolve this

tobe2098:
Mesa en libGL errors are still present. `export FLUTTER_LINUX_RENDERER=software` fixes the issue cleanly, and the errors are not present in the logs anymore after that env change. I guess this is the best solution we can hope for for now.

The other issue I do not think applies, as in this case the hardware is too new rather than too old, and there is a dedicated GPU.

levkropp:
> Mesa en libGL errors are still present. `export FLUTTER_LINUX_RENDERER=software` fixes the issue cleanly, and the errors are not present in the logs anymore after that env change. I guess this is the best solution we can hope for for now.

Yeah from the other issues I have read on this, I believe that flutter 3.35 + software rendering is the only solution right now.

levkropp:
I can also confirm that `export FLUTTER_LINUX_RENDERER=software` fixes my flickering issues on my integrated graphics on my Linux Desktop when I am using a version of the Linux kernel that results in my GPU drivers not loading

This is not related to using mesa graphics specifically (I don't see the mesa errors with or without the env var), but it is more confirmation that

- Flutter has issues right now with hardware rendering on old integrated graphics and Mesa
- `export FLUTTER_LINUX_RENDERER=software` resolves these issues on Flutter 3.35

tobe2098 (approved):
LGTM

PR #4444, opened 2025-10-16
[cli] Remove dangling spinner

ricab:
It was no longer used. The daemon settings we have are very fast and a split second spinner looked poorly anyway.

MULTI-2359


Sploder12 (approved):
LGTM!

PR #4447, opened 2025-10-18
Added more description in README 

akashdeep-24:
I have updated the README.md . This will help new users understand the project more clearly

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ fresh Ubuntu environment with a single command. It uses KVM on Linux, Hyper-V on
 virtual machines with minimal overhead. It can also use VirtualBox on Windows and macOS. Multipass will fetch Ubuntu
 images for you and keep them up to date.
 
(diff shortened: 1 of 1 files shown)


geoffreynyaga:
@akashdeep-24 ,
Thank you for taking time to contribute to Multipass docs.
Looking at your addition, there's slight aspects of redundancy and some information that can already be found in the docs. I agree with @ricab that we need to add the aspects of "full Ubuntu VMS" and other typical use cases to the first paragraph. I would suggest we update the first paragraph of the README to this instead:


Multipass is a lightweight VM manager that lets developers quickly spin up **full Ubuntu virtual machines** on Linux, Windows, or macOS with a single command. It's ideal for **testing, development, and learning Linux environments**. It uses KVM on Linux, Hyper-V on Windows, and QEMU on macOS to run VMs with minimal overhead (it can also use VirtualBox on Windows and macOS). Multipass fetches the necessary Ubuntu images and keeps them up to date.

Issue #646, opened 2019-02-28
Handle yaml-cpp exceptions from a cloud-init config better


townsend2010:
Currently, we just bubble up any `yaml-cpp` exceptions back to the client, but the error message is out of context of what is really wrong.  For example, if a user creates a cloud-init config with incorrect yaml, an error could look like:
`launch failed: operator[] call on a scalar`
As can be seen, it's not clear where the error is.

We should catch the `yaml-cpp` exceptions and make it a more meaningful message to pass to the client and user.

chezanthony:
Hello, can i fix this?
Already looked into this and I'd like to contribute.

What happens when providing an invalid cloud-init file is that it gets parsed successfully by yaml-cpp, and the node gets added to the launch request.

Once the daemon receives the request, it tries to retrieve the values of the following keys: `users`, `ssh_authorized_keys`, and `packages`.

What happened in [#1237](https://github.com/canonical/multipass/issues/1237) is that when the bash script got passed as the cloud-init file via stdin, the entire script got parsed as a scalar node.
Attempting to retrieve a key from a scalar node is what causes the `BadSubscript` exception.

The main issue is that the request reached the daemon despite the invalid yaml.


I'm proposing to fix this by trapping this in `cmd::Launch::parse_args`:
```cpp
...
try
{
    if (cloud_init_file == "-")
    {
        ...
        node = YAML::Load(term->read_all_cin());
        (void)node["users"];
        (void)node["ssh_authorized_keys"];
        (void)node["packages"];
    }
}
...
catch (const YAML::BadSubscript& e)
{
    // handle here by throwing and showing relevant error message. command fails before launching request to daemon. 
}
```

chezanthony:
It should also be noted that while the issue could also be trapped by doing:
```cpp
...
try
{
    if (cloud_init_file == "-")
    {
        ...
        node = YAML::Load(term->read_all_cin());
        if (node.IsScalar && 0 == node.size())
        {
            // handle here
        }
    }
}
```
it only handles the specific case in [#1237](https://github.com/canonical/multipass/issues/1237).
If there are other cases that causes the above exception to arise, it will not be handled by this.
The first approach will fix any exception that arises from retrieving those keys in the daemon.

Ignacio-Trottaa:
Hi!  I'm interested in contributing to Multipass and I'd like to work on this issue. 
Is this still being worked on, or would it be available for me to take a look at it? 

[...]

Issue #1237, opened 2019-12-16
Bad error message when failed to parse cloud-init YAML: "launch failed: operator[] call on a scalar…"


alexellis:
I'm getting an issue when trying to use the --cloud-init flag on MacOS.

```
$ multipass version
multipass  0.10.0+mac
multipassd 0.10.0+mac
```

Works:

```sh
multipass launch --name openfaas
```

Fails:

```sh
multipass launch --name k3s-1 --cloud-init ./k3s.sh

cat k3s.sh | multipass launch --name k3s-1 --cloud-init -
```

k3s.sh

```sh
#!/bin/sh

curl -SLsf https://get.k3sup.dev | sh

k3sup install --local
```

Saviq:
Hi @alexellis, whatever you feed into `--cloud-init` has to be valid YAML, follow the format from [the `cloud-init` docs](http://cloudinit.readthedocs.io/en/latest/topics/examples.html#yaml-examples).

In your case something like this:
```yaml
#cloud-config
runcmd:
- curl -SLsf https://get.k3sup.dev | sh
- k3sub install --local
```

Refer to: [run commands on first boot](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot) in their docs.

Hope this helps! Will close this - please reopen if you think there's a problem still.

alexellis:
I think this needs to re-opened re: the error message is not at all helpful?

Do you have plans to allow user-data or a start-up script like most IaaS platforms would offer?

townsend2010:
Hi @alexellis,

The error you see comes from a third party submodule called yaml-cpp.  We have an open bug to try to make those exceptions more user friendly.  See #646.

As far as the other request to support user-data or a start-up script, I can't seem to find an enhancement request for this.  If you'd like, you could file one so we can track it :grin:

Thanks!

Saviq:
We do have #1040 for supplying files to user-data.

I will reopen this for improving the error message.

townsend2010:
@Saviq,

As I mentioned previously, #646 is opened to better handle the rather plain exceptions thrown from yamp-cpp, which is what the problem is here.

Saviq:
Oh missed that.

Saviq:
Duplicate of #646

PR #4448, opened 2025-10-19
[cli] Validate cloud-init node

chezanthony:
**What this PR does**

This change validates the cloud-init node before sending the launch request.
It ensures that required keys can be retrieved successfully and catches exceptions caused by invalid YAML files.

**Why it’s needed**

Previously, invalid cloud-init yaml files would not raise exceptions after parsing the command args.
This means that launch requests would be sent to the daemon with an invalid cloud-init node.
A `yaml-cpp` exception gets thrown at the daemon side with an unclear message.

Fixes: #646

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -426,8 +426,10 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
                 node = YAML::LoadFile(cloud_init_file_stdstr);
             }
 
-            for (const auto& key : {"users", "ssh_authorized_keys", "packages"}) [[maybe_unused]]
-                auto x = node[key];
+            for (const auto& key : {"users", "ssh_authorized_keys", "packages"})
+            {
+                [[maybe_unused]] auto x = node[key];
+            }
 
             request.set_cloud_init_user_data(YAML::Dump(node));
         }


src/client/cli/cmd/launch.cpp
```diff
@@ -425,8 +425,20 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
 
                 node = YAML::LoadFile(cloud_init_file_stdstr);
             }
+
+            (void)node["users"];
+            (void)node["ssh_authorized_keys"];
+            (void)node["packages"];
```
xmkg:
```suggestion
            for(const auto & key: {"users", "ssh_authorized_keys", "packages"})
                [[maybe_unused]] x = node[key];
```

src/client/cli/cmd/launch.cpp
```diff
@@ -425,8 +425,20 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
...
+
             request.set_cloud_init_user_data(YAML::Dump(node));
         }
+        catch (const YAML::BadSubscript& e)
+        {
+            auto err_detail =
+                fmt::format("{}\n{}", e.what(), "Please ensure key is included in cloud init file.");
```
xmkg:
```suggestion
            auto err_detail =
                fmt::format("{}\n{}", e.what(), "Please ensure the key is included in the cloud-init file.");
```

xmkg (changes requested):
Hey @chezanthony, thanks for the contribution! I have a few nitpicks for you. Also, could you add a unit test for the added code?

Thanks!

chezanthony:
Fixed as per comment. Added unit test as well.
Thanks @xmkg

chezanthony:
Apologies for the delay on this.
Just had a commit message that did not follow **Commit Message** standards :sweat_smile: 

That has been fixed.
Should be good now. Thanks.

src/client/cli/cmd/launch.cpp
```diff
@@ -425,8 +425,21 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
 
                 node = YAML::LoadFile(cloud_init_file_stdstr);
             }
+
+            for (const auto& key : {"users", "ssh_authorized_keys", "packages"}) [[maybe_unused]]
+                auto x = node[key];
```
sharder996:
```suggestion
            for (const auto& key : {"users", "ssh_authorized_keys", "packages"})
                [[maybe_unused]] auto x = node[key];
```

sharder996 (changes requested):

src/client/cli/cmd/launch.cpp
chezanthony: Fixed.
Added curly brackets to avoid clang-format findings (it forces the attribute to the previous line without the brackets).

Thanks @sharder996 .

sharder996 (approved):
LGTM!

Thanks for the contribution, @chezanthony!

src/client/cli/cmd/launch.cpp
```diff
@@ -425,8 +425,23 @@ mp::ParseCode cmd::Launch::parse_args(mp::ArgParser* parser)
 
                 node = YAML::LoadFile(cloud_init_file_stdstr);
             }
+
+            for (const auto& key : {"users", "ssh_authorized_keys", "packages"})
+            {
+                [[maybe_unused]] auto x = node[key];
+            }
```
xmkg, line 432: Hey @chezanthony, I finally had time to look into this in more detail. So, these fields are not "required" by default because multipass would merge those with the multipass defaults, meaning the resulting cloud-init.yaml file would have users/ssh_authorized_keys/packages regardless of their presence in the user-provided cloud-init file. Could you explain why this is needed a bit more? Thanks!

chezanthony: Hi @xmkg 

The purpose of accessing the fields is to trap the exception that is raised from passing a bad YAML file (e.g. malformed, not a YAML file, etc.) and show a more informative error message.

This is in relation to #646 and the main issue #1237.

If the YAML file is valid but the fields do not exist, the code would proceed normally.
If the YAML file is invalid the exception will be caught below with a relevant error message and the command exits with error.

xmkg: I see. Then I'd say we can move the exception catching and reporting to `prepare_user_data` function in daemon.cpp.

PR #4452, opened 2025-10-20
[ci] Automatically clean up old issues

sharder996:
GitHub issues with the `question` label will be marked as stale after 30 days and then closed as "Not planned" after an additional 14 days.

This will only affect GitHub issues create from this day on.

Issues that have an assignee or are part of a milestone are exempt.

See [GitHub docs](https://docs.github.com/en/actions/tutorials/manage-your-work/close-inactive-issues) for the inspiration and [actions/stale](https://github.com/marketplace/actions/close-stale-issues) for reference documentation.

MULTI-2349

### DIFF
--- a/.github/workflows/close-inactive-issues.yml
+++ b/.github/workflows/close-inactive-issues.yml
@@ -22,7 +22,7 @@ jobs:
           days-before-pr-stale: -1
           days-before-pr-close: -1
 
-          only-issue-labels: "question"
+          only-issue-labels: "needs-info"
           start-date: 2025-10-20
           exempt-all-milestones: true
           exempt-all-assignees: true


.github/workflows/close-inactive-issues.yml
```diff
@@ -0,0 +1,33 @@
...
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+
+          # Leave PRs alone
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+
+          only-issue-labels: "question"
+          start-date: 2025-10-20
+          exempt-all-milestones: true
+          exempt-assignees: true
```
jimporter:
```suggestion
          exempt-all-assignees: true
```

sharder996:
> Also, do you know what this would do in a fork? Should we add a check to the action that ensures it only runs in the main repo, like if: github.repository == "canonical/multipass"?

@jimporter Issues are inherently part of a given repository with only PRs that can come from forks. I've effectively disabled the action from doing anything to PRs with:
```
          # Leave PRs alone
          days-before-pr-stale: -1
          days-before-pr-close: -1
```
Are you suggesting that we should add an additional check to disable manipulating PRs in case this changes in the future?

.github/workflows/close-inactive-issues.yml
sharder996: Good catch. Thanks!

jimporter:
> > Also, do you know what this would do in a fork? Should we add a check to the action that ensures it only runs in the main repo, like if: github.repository == "canonical/multipass"?
> 
> @jimporter Issues are inherently part of a given repository with only PRs that can come from forks.

That makes sense. I was worried that the action would error out due to using `secrets.GITHUB_TOKEN`, but reading the docs more thoroughly, that's automatically set up by GitHub itself. Since it's the token for the fork + job, the action should have the correct permissions (I think).

> Are you suggesting that we should add an additional check to disable manipulating PRs in case this changes in the future?

Maybe being extra careful like this would make sense, but I don't think it's really necessary. I was mainly concerned with the secrets, but I've convinced myself that shouldn't be a problem.

.github/workflows/close-inactive-issues.yml
```diff
@@ -0,0 +1,33 @@
...
+permissions:
+  actions: write
+  issues: write
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
```
ricab: Does it need this even if it is not supposed to close any PRs?

.github/workflows/close-inactive-issues.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/stale@v10
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
```
ricab: Should we mention that this applies only to questions?

ricab (commented):
Thanks, makes sense to me. Just a couple of questions.

.github/workflows/close-inactive-issues.yml
sharder996: Sorry, copy-paste error from the guide.

.github/workflows/close-inactive-issues.yml
sharder996: How does this look?

.github/workflows/close-inactive-issues.yml
```diff
@@ -0,0 +1,29 @@
...
+  close-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v10
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue was marked stale because it has been labelled with \"question\" for 30 days."
+          close-issue-message: "This issue was closed because it has been 14 days since being marked as stale."
```
ricab, line 19: To be sure, is the issue marked as stale or closed if it is updated? Or does this apply only when there are no updates for longer the specified period?

sharder996: - 30 days, the issue will have the `stale` label added to it along with a comment with this message.
- 14 days after that, if the `stale` label is still present, the issue will be closed. One must manually remove the `stale` label to prevent this from happening.

ricab: Oh I see. OK fine with me, but we should probably make sure the whole team is aware.

jimporter (approved):

PR #4455, opened 2025-10-21
[windows-macos] Add macos-15-intel to matrix

xmkg:
According to the GH docs, macOS 15 will be the last version that will support Intel. The macos-15 runners will be available until Aug '27.

This patch introduces the macos-15-intel runner to test-drive it.

MULTI-2350

### DIFF
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -44,30 +44,16 @@ jobs:
           // macOS targets
           matrix.include.push({
             "name": "macOS 13 Ventura (Intel Silicon)",
-            "runs-on": "macos-13",
-            "arch": "x86_64"
+            "runs-on": "macos-13"
           },{
             "name": "macOS 14 Sonoma (Apple Silicon)",
-            "runs-on": "macos-14",
-            "arch": "arm64"
+            "runs-on": "macos-14"
           },{
             "name": "macOS Latest (Apple Silicon)",
-            "runs-on": "macos-latest",
-            "arch": "arm64",
-            // ##############################################################
-            // "nuget" and "mono" are removed from the macos-15 images, see:
-            // https://github.com/actions/runner-images/issues/10686
-            // ##############################################################
-            "brew_extra_pkgs": ["nuget", "mono"]
+            "runs-on": "macos-latest"
           },{
             "name": "macOS 15 Sequoia (Intel Silicon)",
             "runs-on": "macos-15-intel",
-            "arch": "x86_64",
-            // ##############################################################
-            // "nuget" and "mono" are removed from the macos-15 images, see:
-            // https://github.com/actions/runner-images/issues/10686
-            // ##############################################################
-            "brew_extra_pkgs": ["nuget", "mono"],
             "optional": true
           })
 
@@ -147,8 +133,6 @@ jobs:
 
     - name: Install dependencies from brew
       if: ${{ runner.os == 'macOS' }}
-      env:
-        BREW_EXTRA: ${{ matrix.brew_extra_pkgs && join(matrix.brew_extra_pkgs, ' ') || '' }}
       run: |
         # Avoid reinstalling cmake
         brew list cmake > /dev/null || brew install cmake


sharder996:
Is there a reason why we need to be running Intel macOS-15 runners?

xmkg:
> Is there a reason why we need to be running Intel macOS-15 runners?

GitHub's ARM runners do not support nested virtualization, so it's mainly needed for the cli-tests (or any kind of testing that needs to run Multipass VMs on macOS). Also, the macOS-13 Intel runners are [deprecated and will be removed this December](https://github.com/actions/runner-images/issues/13046).

.github/workflows/windows-macos.yml
```diff
@@ -30,17 +30,59 @@ jobs:
...
+          // Auto populate fields
+          for (const job of matrix.include) {
+            if (job["runs-on"]?.startsWith("macos"))
+              job.mono = "mono"
+
+            switch (true) {
+              case job["runs-on"]?.endsWith("-intel"):
+                job.arch = "x86_64";
+                break;
+              case job["runs-on"]?.endsWith("-arm"):
```
sharder996: Which runner does this apply to?

.github/workflows/windows-macos.yml
```diff
@@ -30,17 +30,59 @@ jobs:
...
+            "optional": true
+          })
+
+          // Auto populate fields
+          for (const job of matrix.include) {
+            if (job["runs-on"]?.startsWith("macos"))
+              job.mono = "mono"
+
+            switch (true) {
+              case job["runs-on"]?.endsWith("-intel"):
```
sharder996: `macos-13` should also receive this.

.github/workflows/windows-macos.yml
```diff
@@ -30,17 +30,59 @@ jobs:
...
+                job.arch = "x86_64";
+                break;
+              case job["runs-on"]?.endsWith("-arm"):
+                job.arch = "arm64";
+                break;
+            }
```
sharder996: I think I would rather just manually add this into the above dictionary literal.

sharder996 (changes requested):

sharder996:
> Also, the macOS-13 Intel runners are deprecated and will be removed this December.

Good to know. Guess we will need to prioritize backwards compatible macOS builds.

.github/workflows/windows-macos.yml
xmkg: Yes, I initially used the name field for this, but the name ended up being changed to a more appealing one. This is a bit overkill and redundant at this point, too.

xmkg: The weird thing is that the CI passes :shrug: Could that mean the `arch` is redundant at this point? I'll take a look.

xmkg: Found the original commit:

https://github.com/canonical/multipass/commit/bbe964b893df660ca52f2a7acf68914405068c2f

So, it seems like GH ARM runners were not a thing back then, so the arch wrapper is basically needed for cross-compiling macOS packages for the different target arch. That's a thing we're no longer doing since the arch-specific packages are being built in their native host environment (i.e., Intel and ARM silicon, respectively).

I'll move the arch vars to each respective JSON node, and submit another PR to remove the arch wrapper altogether.

ricab:
> > Also, the macOS-13 Intel runners are deprecated and will be removed this December.
> 
> Good to know. Guess we will need to prioritize backwards compatible macOS builds.

We discussed this before and decided then to move to 14 as a minimum. We announced so on [1.16.0](https://github.com/canonical/multipass/releases/tag/v1.16.0):

> This is also likely to be the last release that will support macOS 13.

If we get backward compatible builds before the next release, great. But we don't need to hold on it.

.github/workflows/windows-macos.yml
```diff
@@ -30,17 +30,54 @@ jobs:
...
+            "runs-on": "macos-latest",
+            "arch": "arm64",
+            // ##############################################################
+            // "nuget" and "mono" are removed from the macos-15 images, see:
+            // https://github.com/actions/runner-images/issues/10686
+            // ##############################################################
+            "brew_extra_pkgs": ["nuget", "mono"]
+          },{
+            "name": "macOS 15 Sequoia (Intel Silicon)",
+            "runs-on": "macos-15-intel",
```
ricab, line 56: Why use 15 here but latest on the previous one?

.github/workflows/windows-macos.yml
```diff
@@ -30,17 +30,54 @@ jobs:
...
+            // https://github.com/actions/runner-images/issues/10686
+            // ##############################################################
+            "brew_extra_pkgs": ["nuget", "mono"],
+            "optional": true
+          })
+
+          // Auto populate fields
+          for (const job of matrix.include) {
+            if (job["runs-on"]?.startsWith("macos"))
+              job.mono = "mono"
```
ricab: We had this on only one entry before. Why did you prefer to add to all now?

.github/workflows/windows-macos.yml
```diff
@@ -63,7 +100,9 @@ jobs:
       mac-x86-package-file-name: ${{ steps.publish-data.outputs.macos-13-x86_64-package-file-name }}
       mac-arm-package-file-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-file-name }}
 
+    name: Build and Test (${{ matrix.name }})
```
ricab, line 83: 👍

.github/workflows/windows-macos.yml
xmkg: Right now, `macos-latest` points to the `macos-15`, which is the ARM variant of the runner. `macos-15-intel` is the name for the Intel one. There's no such equivalent tag for Intel, e.g., `macos-latest-intel` is not a thing.

.github/workflows/windows-macos.yml
xmkg: `nuget` is a Windows executable (.exe) file, so it needs to run through `mono` regardless of the Mac version. I think some package managers ship a shim to NuGet, which does exactly this, but AFAICT it's not consistent across the runners.

sharder996:
> We discussed this before and decided then to move to 14 as a minimum. We announced so on [1.16.0](https://github.com/canonical/multipass/releases/tag/v1.16.0):

@ricab There are no macOS 14 Intel runners offered by GitHub.

.github/workflows/windows-macos.yml
sharder996: You can't apply this in a blanket approach. Its failing for some of the runners in CI:
https://github.com/canonical/multipass/actions/runs/18710495978/job/53357608697#step:16:32

ricab:
I see, you're right @sharder996, thanks for pointing that out. It's a little surprising that they would skip 14... There is [an image](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md) that I think is intel, but no hosted runners.

.github/workflows/windows-macos.yml
xmkg: > You can't apply this in a blanket approach. Its failing for some of the runners in CI:
> [canonical/multipass/actions/runs/18710495978/job/53357608697#step:16:32](https://github.com/canonical/multipass/actions/runs/18710495978/job/53357608697#step:16:32)

Right -- ditched the mono matrix var altogether. The `Configure NuGet Source` now checks if the NuGet is actually a PE32 binary before dispatching it through "mono".

xmkg: @sharder996 should be fixed now. Let me know how it looks.

PR #4457, opened 2025-10-21
[cmake] Get rid of the generate_grpc_cpp function

xmkg:
Use the protobuf_generate function provided by the FindProtobuf.cmake module.

MULTI-2351


src/rpc/CMakeLists.txt
```diff
@@ -22,10 +22,24 @@ else()
...
 
-target_link_libraries(rpc
-  gRPC)
+ # 2) Generate the .grpc.pb.cc/.grpc.pb.h
+protobuf_generate(
+    TARGET rpc
+    LANGUAGE grpc
+    PROTOS ${MULTIPASS_PROTOCOL_SPEC_ABS_PATH}
+    PROTOC_OUT_DIR ${GRPC_GENERATED_SOURCE_DIR}
+    PLUGIN protoc-gen-grpc=$<TARGET_FILE:gRPC::grpc_cpp_plugin>
```
sharder996, line 41: You'll need to bump the [min cmake version](https://github.com/canonical/multipass/blob/d1709fa293ddfc19d7dc9178bc7f7cf5b37d0ba0/CMakeLists.txt#L15) to guarantee the presence of this variable.

xmkg: Done

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required(VERSION 3.20)
+cmake_minimum_required(VERSION 3.25)
```
sharder996, line 15: Should only need `3.21`, no?

> PLUGIN <plugin>
>
>    Added in version 3.21.
>
>    An optional plugin executable. This could be, for example, the path to grpc_cpp_plugin.

xmkg: We'll need a couple of features from 3.25 pretty soon, namely cmake-presets enhancements and `add_subdirectory(... SYSTEM)`, so I'm bumping it in advance.

jimporter (approved):
This makes sense to me, and matches the recommended usage in the protobuf docs.

PR #4458, opened 2025-10-21
[cmake] Add CMake presets for snap builds

xmkg:
This PR removes CI-specific snapcraft.yaml patches in favor of CMake presets and YAML merging.

MULTI-2381

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -10,8 +10,24 @@
             "name": "common",
             "description": "Base preset for all.",
             "binaryDir": "${sourceDir}/build",
+            "hidden": true,
             "cacheVariables": {
                 "CMAKE_EXPORT_COMPILE_COMMANDS": {
+                    "type": "BOOL",
+                    "value": "ON"
+                }
+            }
+        },
+        {
+            "name": "debug",
+            "description": "Additive preset for debug builds.",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": {
+                    "type": "STRING",
+                    "value": "Debug"
+                },
+                "MULTIPASS_ENABLE_TESTS": {
                     "type": "BOOL",
                     "value": "ON"
                 }
@@ -48,6 +64,14 @@
                     "value": "ON"
                 }
             }
+        },
+        {
+            "name": "snap-debug",
+            "description": "Preset for snap builds in CI for test runs (debug).",
+            "inherits": [
+                "snap",
+                "debug"
+            ]
         },
         {
             "name": "ci-snap",
@@ -113,7 +137,10 @@
         {
             "name": "ci-snap-debug",
             "description": "Preset for snap builds in CI for test runs.",
-            "inherits": "ci-snap",
+            "inherits": [
+                "ci-snap",
+                "debug"
+            ],
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": {
                     "type": "STRING",


snap/snapcraft.yaml
```diff
@@ -195,14 +199,11 @@ parts:
...
     source: .
-    cmake-parameters:
-    - -DCMAKE_BUILD_TYPE=Release
-    - -DCMAKE_INSTALL_PREFIX=/
-    - -DMULTIPASS_ENABLE_TESTS=off
-    - -DMULTIPASS_UPSTREAM=origin
-    - -DMULTIPASS_ENABLE_FLUTTER_GUI=on
     override-build: |
-      craftctl default
+      cmake --preset "$CMAKE_PRESET" -S ${SNAPCRAFT_PART_SRC_WORK} -B ${SNAPCRAFT_PART_BUILD_WORK}
```
xmkg, line 206: Unfortunately, Snapcraft's CMake plugin does not support CMake presets yet, hence no call to `craftctl default`.

xmkg:
@jimporter @tobe2098 this one's ready to go, let me know what you folks think.

.github/workflows/linux.yml
```diff
@@ -85,14 +85,21 @@ jobs:
...
-          patch -p1 --no-backup-if-mismatch < ${PATCH_PREFIX}-${{ matrix.build-type }}.patch
+      run: |
+        go install github.com/mikefarah/yq/v4@latest
+
+        # Substitute the environment variables in the snapcraft-ci-override.yaml
+        envsubst < snap/snapcraft-ci-override.yaml.in | yq eval 'del(.. | select(. == null or . == "")) | del(.. | select((tag == "!!map" or tag == "!!seq") and length == 0))' -P > snap/snapcraft-ci-override.yaml
+
+        # Merge / override the original snapcraft.yaml
+        yq eval-all '. as $item ireduce ({}; . *+ $item)' snap/snapcraft.yaml snap/snapcraft-ci-override.yaml > snap/snapcraft.yaml.tmp
+        mv snap/snapcraft.yaml.tmp snap/snapcraft.yaml
```
jimporter, line 102: Just a question: would there be any value in putting this code in some script file (or something like that) so that it's easier to do this locally as well? That would make it easier to test out the Snap CI builds locally if we were trying to debug something.

Maybe that's something we should do in a followup though. The CMake preset work in this PR already gets us closer to being able to reproduce CI builds locally.

src/image_host/ubuntu_image_host.cpp
```diff
@@ -219,7 +219,7 @@ void mp::UbuntuVMImageHost::fetch_manifests(const bool force_update)
                 manifest_bytes_from_official,
                 manifest_bytes_from_mirror,
                 mirror_site.value_or(official_site),
-                [&remote_info](VMImageInfo& info) {
+                [&remote_info = remote_info](VMImageInfo& info) {
```
jimporter, line 223: Was the error here because `remote_info` is a structured binding? From doing a bit of research, that's [legal C++20](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1091r3.html#lambda-captures), and the error is [fixed in clang-16](https://github.com/llvm/llvm-project/commit/127bf44385424891eb04cff8e52d3f157fc2cb7c).

If we need to build with clang-15, then I suppose this repetition is necessary, but could we add a TODO comment here to remove this once we require a newer clang? (Or we could require clang-16 now if we feel up to it.)

xmkg: Yes, it's due to the clang bug you mentioned. Unfortunately, we're stuck with clang-15 in core22, so I'll add a TODO.

.github/workflows/linux.yml
xmkg: Right now, it wouldn't make much sense on its own unless we also define the environment variables that are being substituted. Perhaps we can capture all of that in a script for another PR later on -- how does that sound to you?

src/image_host/ubuntu_image_host.cpp
xmkg: Done

tobe2098 (dismissed):

CMakePresets.json
```diff
@@ -0,0 +1,136 @@
...
+                    "type": "STRING",
+                    "value": "Debug"
+                },
+                "MULTIPASS_ENABLE_TESTS": {
+                    "type": "BOOL",
+                    "value": "ON"
+                }
+            }
+        }
+    ],
```
jimporter, line 155: I could go either way on this, but what about adding a `snap-debug` preset like this for local development? (If you don't think that would be valuable, then don't bother; I just noticed it when trying the `snap` preset and realizing it doesn't build the tests.)

jimporter (approved):
Looks good to me. Just one small thought below, but approved either way.

xmkg:
> Not exactly a review but, if this impacts our local builds, could instructions be amended accordingly, please?

This only affects the snap build, so nothing changes in surface. It should work out-of-the-box as before.

CMakePresets.json
xmkg: Makes sense to me, will do!

xmkg: Done

tobe2098 (approved):
LGTM, nice work Mustafa!

PR #4459, opened 2025-10-21
[submodules] Replace yaml-cpp submodule with vcpkg

xmkg:
MULTI-2343

### DIFF
--- a/3rd-party/CMakeLists.txt
+++ b/3rd-party/CMakeLists.txt
@@ -69,23 +69,6 @@ if (NOT MSVC)
   target_compile_options(gRPC INTERFACE "-Wno-unused-parameter" "-Wno-non-virtual-dtor" "-Wno-pedantic")
 endif ()
 
-# YAML C++
-# Disable tests here to avoid double-including gtest
-option(YAML_CPP_BUILD_TOOLS OFF)
-option(YAML_CPP_BUILD_CONTRIB OFF)
-option(YAML_CPP_BUILD_TESTS OFF)
-
-add_subdirectory_compat(yaml-cpp EXCLUDE_FROM_ALL)
-set_target_properties(yaml-cpp PROPERTIES POSITION_INDEPENDENT_CODE ON)
-
-add_library(yaml INTERFACE)
-
-target_include_directories(yaml INTERFACE
-  ${CMAKE_CURRENT_SOURCE_DIR}/yaml-cpp/include)
-
-target_link_libraries(yaml INTERFACE
-  yaml-cpp)
-
 include(ExternalProject)
 
 # QEMU itself
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
-[submodule "3rd-party/yaml-cpp"]
-	path = 3rd-party/yaml-cpp
-	url = https://github.com/jbeder/yaml-cpp.git
-	ignore = dirty
 [submodule "3rd-party/libssh/libssh"]
 	path = 3rd-party/libssh/libssh
 	url = https://github.com/CanonicalLtd/libssh.git
--- a/3rd-party/submodule_info.md
+++ b/3rd-party/submodule_info.md
@@ -31,11 +31,6 @@ Version: v2024-12-30 |
 <https://github.com/tukaani-project/xz-embedded.git> |
 <https://tukaani.org/xz/embedded.html>
 
-### yaml-cpp
-Version: 0.8.0 (+[our patches](https://github.com/canonical/yaml-cpp/compare/0.8.0..multipass)) |
-<https://github.com/canonical/yaml-cpp.git> |
-<https://github.com/jbeder/yaml-cpp/releases>
-
 ### JsonCpp
 Version: 1.9.4 |
 <https://github.com/open-source-parsers/jsoncpp.git> |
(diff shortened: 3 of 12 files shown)


jimporter (approved):
LGTM! CI is green, and I've verified locally that there aren't any extra places we need to update our build scripts.

jimporter:
@xmkg Actually... one question: do you have any thoughts about whether we should pin our `yaml-cpp` dependency to a particular version? I don't have a strong feeling either way, though pinning would make our builds more reproducible. (We'd probably want to have the Renovate bot suggest PRs for upgrading though, so that we don't forget.)

xmkg:
> @xmkg Actually... one question: do you have any thoughts about whether we should pin our `yaml-cpp` dependency to a particular version? I don't have a strong feeling either way, though pinning would make our builds more reproducible. (We'd probably want to have the Renovate bot suggest PRs for upgrading though, so that we don't forget.)

Right now, the vcpkg's `builtin-baseline` acts as a repo-wide pin for all vcpkg deps, meaning that all the deps would be fetched from a particular state of vcpkg registry in time, i.e., `builtin-baseline` being pinned to commit "x" guarantees that we're always going to get whatever yaml-cpp version is present in "x":

https://github.com/canonical/multipass/blob/2eb4538731549013b7755ef88e1439922eea4644/vcpkg.json#L2

This also applies to the dependency's whole ancestry, too.

That's except for explicit version overrides, see: https://github.com/canonical/multipass/blob/2eb4538731549013b7755ef88e1439922eea4644/vcpkg.json#L8-L11

I'm okay with explicit pins too, but if our concerns are about reproducibility, "builtin-baseline" should suffice. Let me know what you think.

3rd-party/CMakeLists.txt
```diff
@@ -69,23 +69,6 @@ if (NOT MSVC)
   target_compile_options(gRPC INTERFACE "-Wno-unused-parameter" "-Wno-non-virtual-dtor" "-Wno-pedantic")
 endif ()
 
-# YAML C++
-# Disable tests here to avoid double-including gtest
-option(YAML_CPP_BUILD_TOOLS OFF)
-option(YAML_CPP_BUILD_CONTRIB OFF)
-option(YAML_CPP_BUILD_TESTS OFF)
```
ricab, line 76: Are these all off in vcpkg too?

3rd-party/CMakeLists.txt
```diff
@@ -69,23 +69,6 @@ if (NOT MSVC)
...
-
-add_subdirectory_compat(yaml-cpp EXCLUDE_FROM_ALL)
-set_target_properties(yaml-cpp PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
ricab, line 79: Does this still hold? Is it still relevant?

ricab (commented):
Thanks Mustafa! A couple of questions.

3rd-party/CMakeLists.txt
xmkg: Tools and tests are off, but contrib is not.

https://github.com/microsoft/vcpkg/blob/4334d8b4c8916018600212ab4dd4bbdc343065d1/ports/yaml-cpp/portfile.cmake#L17-L20

3rd-party/CMakeLists.txt
xmkg: It seems that's no longer relevant -- otherwise we would get a build failure.

jimporter:
> Right now, the vcpkg's `builtin-baseline` acts as a repo-wide pin for all vcpkg deps, meaning that all the deps would be fetched from a particular state of vcpkg registry in time, i.e., `builtin-baseline` being pinned to commit "x" guarantees that we're always going to get whatever yaml-cpp version is present in "x":

Thanks, that makes sense. I think that should be enough for now at least. If the list of packages we pull in via `vcpkg` gets larger, it might make sense to pin every package explicitly so that we can tell at a glance what versions we use, but I don't think that blocks this PR.

3rd-party/CMakeLists.txt
ricab: Could we turn contrib off too? It's [not much](https://github.com/jbeder/yaml-cpp/tree/master/src/contrib), but it could grown and anything we can save in build time would help.

3rd-party/CMakeLists.txt
ricab: Right.

3rd-party/CMakeLists.txt
xmkg: It's not trivial since the original vcpkg portfile does not expose that. We'd need to copy the yaml-cpp port and patch the `portfile.cmake`, which would be a maintenance burden. I'm working on figuring out a way to do that without having to patch. We could use that here, too, if it works out. How does that sound?

ricab: The super-build thing you mentioned? That would be fine with me. If it's too much trouble, leave it.

xmkg: Yeah, the superbuild thing. It's still a work-in-progress, though. I can follow up on this once it's ready. In the meantime, do you think we should pause on this or just move forward as-is?

ricab: Let's move forward.

PR #4461, opened 2025-10-21
 [semver] Replace semver with neargye-semver +vcpkg

xmkg:
neargye-semver is a better-maintained semver library than what we have and it's available in the vcpkg.
    
Hid the semver behind an opaque type + spaceship operator to abstract away the implementation.

MULTI-2341


Sploder12 (approved):
Changes make sense to me and looks good! Looks like it's failing Lint though

xmkg:
> Changes make sense to me and looks good! Looks like it's failing Lint though

Fixed now.

Sploder12 (dismissed):

PR #4463, opened 2025-10-21
[fetchcontent] Migrate googletest to vcpkg

xmkg:
MULTI-2345


.clang-format
```diff
@@ -41,4 +41,7 @@ PointerAlignment: Left
 ---
 Language: Proto
 DisableFormat: true
+---
+Language: Json
+IndentWidth: 4
```
jimporter, line 46: I have no opinion on what indent width to use (this seems fine though, and is more consistent with our C++ code). However, if we do merge this, we should update `.editorconfig` to match.

jimporter (changes requested):
Looks good, aside from one small comment.

.clang-format
xmkg: done

PR #4466, opened 2025-10-22
[ci] Match commit headers (like `Signed-off-by:`) case-insensitively

### LINKED ISSUE #4465: [ci] Match commit headers (like `Signed-off-by:`) case-insensitively
We're currently matching commit message headers case-sensitively in our commit hook, though they're case *in*sensitive (and our tests were checking that). I think we missed this since the test cases covering these had a sneaky error where we weren't concatenating the strings correctly, and instead we were running one subtest per line of the message.

jimporter:
This resolves #4465.

### DIFF
--- a/git-hooks/commit-msg.py
+++ b/git-hooks/commit-msg.py
@@ -77,10 +77,10 @@ class CommitMsgRulesChecker:
             re.match(r"\[\d+\]:.*\n?", line)
             or
             # Ignore Signed-off-by:
-            re.match(r"Signed-off-by:.*\n?", line)
+            re.match(r"Signed-off-by:.*\n?", line, re.IGNORECASE)
             or
             # Ignore Co-authored-by:
-            re.match(r"Co-authored-by:.*\n?", line)
+            re.match(r"Co-authored-by:.*\n?", line, re.IGNORECASE)
         )
 
     def validate_all(self):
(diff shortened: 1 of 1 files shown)


jimporter:
> I would say the case is for us to define, no?
> 
> To me, upper case looks better to start a sentence/paragraph. That's the case with the rest of the text and what I am most used to seeing on Git/GH.

That would be ok too, if that's what we decide. ([Git describes these](https://git-scm.com/docs/git-interpret-trailers) as being like RFC822 headers, so from Git's point of view, they should be case-insensitive.)

> I am not opposed to case insensitive, but I would like to understand the motivation. If we allow that, we should update the corresponding contribution rules.

I noticed this when looking over the change again and saw that our code was checking for "Sentence-case", but the tests used "Title-Case", and so they should have been failing. I don't have an opinion about which style(s) we accept though. I just modified the code to match the now-fixed test.

> What was the sneaky error exactly?

The test cases mixed string concatenation and commas incorrectly, so we weren't testing what we expected. For example:

```python
            "[msg] Subject\n\n"
            "> This body line (verbatim text) is clearly over 72 characters long[...]",
            "",
            "> This body line (verbatim text) is clearly over 72 characters long[...]",
            "",
            "Regular text"
```

From reading the surrounding context, that should be a single commit message to test, but it's actually five:
1. `""[msg] Subject\n\n> This body line (verbatim text) is clearly over 72 characters long[...]"`
2. `""`
3. `"> This body line (verbatim text) is clearly over 72 characters long[...]"`
4. `""`
5. `"Regular text"`

ricab:
Got it, thank for clarifying @jimporter! I have a slight preference for Sentence case, but I don't mind changing it if you like, provided it's all in sync with CONTRIBUTING.md.

jimporter:
> Got it, thank for clarifying @jimporter! I have a slight preference for Sentence case, but I don't mind changing it if you like, provided it's all in sync with CONTRIBUTING.md.

We could also require Sentence case, but I think if that's our decision, we should make a separate rule for it so that we can report a more-specific error message. (Currently, we'd only report an error if the commit message contained a header like this that broke some other rule, like our line-length rule.)

ricab:
You're right, for some reason I was under the impression that the rules implied case today.

xmkg (approved):
LGTM!

PR #4467, opened 2025-10-22
[windows-macos] Macos: remove the arch wrapper

xmkg:
The arch wrapper is needed when cross-compiling, i.e. intel->arm and vice versa. The reason why the arch wrapper was needed in the beginning was there were no ARM based runners as described in this commit:

> https://github.com/canonical/multipass/commit/bbe964b893df660ca52f2a7acf68914405068c2f

Since then, the arch wrapper has basically became redundant. We now use the respective arch for producing intel and arm packages, hence making the wrapper redundant.

Made several changes on the steps that depend on the "arch" variable:

- Replaced the arch check in nuget shim replace step with a file type check.
- Install nuget step now checks if nuget or mono is present or not. If not, then proceeds with installing them.

MULTI-2355

### DIFF
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -131,9 +131,13 @@ jobs:
           pixman \
           pkg-config \
           python \
-          wget \
-          mono \
-          nuget
+          wget
+
+        # Force brew nuget & mono
+        brew install nuget || true
+        brew install mono || true
+        brew link --overwrite mono
+        brew link --overwrite nuget
 
     - name: Install dependencies from pip
       if: ${{ runner.os == 'macOS' }}


.github/workflows/windows-macos.yml
```diff
@@ -332,8 +333,8 @@ jobs:
       shell: bash
       run: |
         NAME="${{ steps.cmake-package.outputs.name }}"
-        ARTIFACT_NAME="${{ matrix.runs-on }}-${{ matrix.arch }}-package-artifact-name=${{ matrix.runs-on }}-${NAME}"
-        FILE_NAME="${{ matrix.runs-on }}-${{ matrix.arch }}-package-file-name=${NAME}"
+        ARTIFACT_NAME="${{ matrix.runs-on }}-package-artifact-name=${{ matrix.runs-on }}-${NAME}"
+        FILE_NAME="${{ matrix.runs-on }}-package-file-name=${NAME}"
```
ricab, line 334: I believe we need to keep the arch here. Packages for the two architectures need to be combined later, but this prevents that: https://github.com/canonical/multipass/actions/runs/18712914069/job/53371065345?pr=4467

xmkg: Ahh, right, but we shouldn't have to keep the arch since `matrix.runs-on` is already unique. I just forgot to remove the arch tag from the following:

```
      mac-x86-package-artifact-name: ${{ steps.publish-data.outputs.macos-13-x86_64-package-artifact-name }}
      mac-arm-package-artifact-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-artifact-name }}
      mac-arm-package-file-name: ${{ steps.publish-data.outputs.macos-14-arm64-package-file-name }}
```

It would also be nice if the step had failed when it couldn't find anything to combine. I'll look into it.

xmkg:
> I think there might be some behaviour change since `nuget` and `mono` are being installed on all macOS runners. Looking at the checks, `macOS-14` and `macos-latest` aren't pulling cached vcpkg binaries and recompiling them.

Good catch. I think the `Overwrite Homebrew nuget shim with nuget DLL` step wasn't doing what it supposed to do due to a path mixup which I fixed a couple of mins ago. Let's see how that goes.

xmkg:
> > I think there might be some behaviour change since `nuget` and `mono` are being installed on all macOS runners. Looking at the checks, `macOS-14` and `macos-latest` aren't pulling cached vcpkg binaries and recompiling them.
> 
> Good catch. I think the `Overwrite Homebrew nuget shim with nuget DLL` step wasn't doing what it supposed to do due to a path mixup which I fixed a couple of mins ago. Let's see how that goes.

@sharder996 should be fixed now. Let me know if there any other issues.

sharder996 (approved):
LGTM

PR #4469, opened 2025-10-22
[docs] Alternative RDP procedure working with Wayland

msuchane:
The [Set up a graphical interface](https://documentation.ubuntu.com/multipass/latest/how-to-guides/customise-multipass/set-up-a-graphical-interface/) guide no longer works with Ubuntu 25.10 instances because Ubuntu 25.10 removes X11 support from the GNOME desktop.

This alternative procedure still uses RDP to connect but the server setup ensures that it's compatible with Wayland. As a result, you can establish a graphical connection to Ubuntu 25.10 and later, presumably.

### DIFF
--- a/docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
+++ b/docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
@@ -97,7 +97,6 @@ This RDP server only works with the default GNOME desktop. You can use it with U
     - gnome-remote-desktop
     - winpr3-utils
     runcmd:
-    - whoami > /etc/whoami
     - sudo -u gnome-remote-desktop winpr-makecert3 -silent -rdp -path ~gnome-remote-desktop rdp-tls
     - grdctl --system rdp set-tls-key ~gnome-remote-desktop/rdp-tls.key
     - grdctl --system rdp set-tls-cert ~gnome-remote-desktop/rdp-tls.crt
--- a/data/cloud-init-yaml/cloud-init-gnome-rdp.yaml
+++ b/data/cloud-init-yaml/cloud-init-gnome-rdp.yaml
@@ -7,7 +7,6 @@ packages:
   - gnome-remote-desktop
   - winpr3-utils
 runcmd:
-  - whoami > /etc/whoami
   - sudo -u gnome-remote-desktop winpr-makecert3 -silent -rdp -path ~gnome-remote-desktop rdp-tls
   - grdctl --system rdp set-tls-key ~gnome-remote-desktop/rdp-tls.key
   - grdctl --system rdp set-tls-cert ~gnome-remote-desktop/rdp-tls.crt


docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
```diff
@@ -50,9 +52,78 @@ sudo passwd ubuntu
...
 You will be asked to enter and re-enter a password.
 
-You are done on the server side!
+You are done on the server side! Quit the Ubuntu shell on the running instance with the `exit` command. Proceed to {ref}`connect-to-the-instance-with-an-rdp-client`.
+
+### Set up the GNOME Remote Desktop server
+
+1. Create the following `cloud-init.yaml` configuration file:
+
+   ```{code-block} yaml
```
xmkg: Do you think it's worth placing this in https://github.com/canonical/multipass/tree/main/data/cloud-init-yaml?

msuchane:
There's a couple of things to figure out:

- @sergio-costas, this procedure uses `cloud-init.yaml` while the XRDP procedure uses plain shell commands. Are those interchangeable or does the new procedure require `cloud-init.yaml`?

- @sergio-costas, most of the commands run with sudo except for this one:

    ```
    echo /usr/sbin/nologin >> /etc/shells
    ```

    Is this missing a `sudo tee`?

- @geoffreynyaga, how do you feel about the placement and organization? I haven't explained yet that on Ubuntu 25.10, you have to use the second procedure because X11 has been removed in Ubuntu 25.10. Any preferences how it should be framed?

Live preview: https://canonical-ubuntu-documentation-library--4469.com.readthedocs.build/multipass/4469/how-to-guides/customise-multipass/set-up-a-graphical-interface/

sergio-costas:
>     * @sergio-costas, this procedure uses `cloud-init.yaml` while the XRDP procedure uses plain shell commands. Are those interchangeable or does the new procedure require `cloud-init.yaml`?

It can be done manually too, but using a `cloud-init.yaml` file simplifies the job. A file like that can be done for XRDP too. It's just a way of automatize the process.

>     * @sergio-costas, most of the commands run with sudo except for this one:
>       ```
>       echo /usr/sbin/nologin >> /etc/shells
>       ```
>       Is this missing a `sudo tee`?

No. These commands are run as `root` by default, but the `sudos` in the other lines have `-u gnome-remote-desktop` as the first parameters, to run the commands as the `gnome-remote-desktop` user. That command, instead, is to ensure that the `nologin` program is in the list of valid shells and, thus, ensure that the `grdctl --headless rdp enable` command doesn't fail (the gnome-remote-desktop user has that shell), so it has to be run as `root`

msuchane:
Ah, thanks @sergio-costas, that makes absolute sense. Didn't notice the `-u gnome-remote-desktop`.

If shell commands and `cloud-init.yaml` are interchangeable, I'll leave it up to @geoffreynyaga whether to keep the procedures the way they are or whether to standardize on one of the formats. I'm okay with either.

docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
```diff
@@ -50,9 +52,78 @@ sudo passwd ubuntu
...
+   - loginctl enable-linger ubuntu
+   ```
+
+   Replace `<your-password>` with a secure password on the highlighted lines.
+
+1. Launch the configured Multipass instance:
+
+   ```text
+   multipass launch \
+      -vvvv
```
sergio-costas: This is not required.

docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
```diff
@@ -50,9 +52,78 @@ sudo passwd ubuntu
...
+      --memory 12G \
+      --cpus=2 \
+      --disk 30G \
+      --timeout 600 \
+      --cloud-init ./cloud-init.yaml 25.10
+   ```
+
+   The command sets the timeout to 10 minutes because the default timeout of 5 minutes isn't enough to initialise the instance.
```
sergio-costas, line 128: I would also comment that the other values can be adjusted (RAM, disk, cores...)

docs/how-to-guides/customise-multipass/set-up-a-graphical-interface.md
msuchane: Could be, if it's fine that the RDP configuration contains a replaceable password string.

sergio-costas:
> If shell commands and `cloud-init.yaml` are interchangeable, I'll leave it up to @geoffreynyaga whether to keep the procedures the way they are or whether to standardize on one of the formats. I'm okay with either.

The big advantage of the `cloud-init.yaml` file is that it greatly reduces the risk of typos or errors while doing the process, and also simplifies the process itself, since you don't have to keep a list of commands to do manually.

geoffreynyaga:
> There's a couple of things to figure out:
> 
> * @sergio-costas, this procedure uses `cloud-init.yaml` while the XRDP procedure uses plain shell commands. Are those interchangeable or does the new procedure require `cloud-init.yaml`?
> * @sergio-costas, most of the commands run with sudo except for this one:
>   ```
>   echo /usr/sbin/nologin >> /etc/shells
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   Is this missing a `sudo tee`?
> * @geoffreynyaga, how do you feel about the placement and organization? I haven't explained yet that on Ubuntu 25.10, you have to use the second procedure because X11 has been removed in Ubuntu 25.10. Any preferences how it should be framed?
> 
> Live preview: https://canonical-ubuntu-documentation-library--4469.com.readthedocs.build/multipass/4469/how-to-guides/customise-multipass/set-up-a-graphical-interface/

@msuchane I am super happy with the additions. Yes, at the moment we may need to have a disclaimer for people using 25.10 and above to follow the new cloud-init guide. I also support @xmkg idea of placing the yaml file in this location https://github.com/canonical/multipass/tree/main/data/cloud-init-yaml since we reference the custom setups a lot in this [guide](https://documentation.ubuntu.com/multipass/stable/how-to-guides/manage-instances/launch-customized-instances-with-multipass-and-cloud-init/)

sergio-costas:
> @msuchane I am super happy with the additions. Yes, at the moment we may need to have a disclaimer for people using 25.10 and above to follow the new cloud-init guide. I also support @xmkg idea of placing the yaml file in this location https://github.com/canonical/multipass/tree/main/data/cloud-init-yaml since we reference the custom setups a lot in this [guide](https://documentation.ubuntu.com/multipass/stable/how-to-guides/manage-instances/launch-customized-instances-with-multipass-and-cloud-init/)

Important detail: this is only for Gnome Shell and Ubuntu Desktop, because Gnome Shell removed X11 session support in version 49. Other desktops with support for X11 should continue using the XRDP method (thus, it is still valid for other Ubuntu flavors). Also, this method can be used in older versions of Gnome Shell, although in old ones can result in some characteristics still not implemented.

sergio-costas:
Also, I can prepare a cloud-init.yaml version for XRDP if you want...

sergio-costas:
@msuchane Also, in the preview, the text seems to indicate that you have to do both procedures, instead of explaining that the XRDP one is for desktops with support for X11 sessions, and the other one is for Gnome Shell sessions from version 49 onward...

sergio-costas:
Maybe the best course of action is to have both the `cloud-init.yaml` files for both, for the case where a new VM is being created, and the list of commands, to add desktop support to already created VMs. What do you say?

msuchane:
After some experimentation, I think that the RDP server choice is best conveyed in a table:

https://canonical-ubuntu-documentation-library--4469.com.readthedocs.build/multipass/4469/how-to-guides/customise-multipass/set-up-a-graphical-interface/

Regarding the shell commands vs. `cloud-init.yaml`, I'll just leave it up to the Multipass maintainers :-)

msuchane:
Hi @geoffreynyaga, I think you can merge this documentation now.

The GNOME+Wayland configuration has its cloud init file, and as for the XRDP configuration, you can either extend it later or leave it as a legacy reference.

PR #4472, opened 2025-10-23
Consolidate VM fields, part 1

ricab:
And other related refactors.

### DIFF
--- a/src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
+++ b/src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
@@ -504,12 +504,12 @@ std::string mp::VirtualBoxVirtualMachine::ssh_username()
     return desc.ssh_username;
 }
 
-auto mp::VirtualBoxVirtualMachine::management_ipv4() -> std::optional<IPAddress>
+std::optional<mp::IPAddress> mp::VirtualBoxVirtualMachine::management_ipv4()
 {
     return std::nullopt;
 }
 
-auto mp::VirtualBoxVirtualMachine::get_all_ipv4() -> std::vector<IPAddress>
+std::vector<mp::IPAddress> mp::VirtualBoxVirtualMachine::get_all_ipv4()
 {
     using namespace std;
 
--- a/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
+++ b/src/platform/backends/hyperv/hyperv_virtual_machine.cpp
@@ -452,7 +452,7 @@ std::string mp::HyperVVirtualMachine::ssh_username()
     return desc.ssh_username;
 }
 
-auto mp::HyperVVirtualMachine::management_ipv4() -> std::optional<IPAddress>
+std::optional<mp::IPAddress> mp::HyperVVirtualMachine::management_ipv4()
 {
     // Not using cached SSH session for this because a) the underlying functions do not
     // guarantee constness; b) we endure the penalty of creating a new session only when we
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -509,7 +509,7 @@ std::string mp::QemuVirtualMachine::ssh_username()
     return desc.ssh_username;
 }
 
-auto mp::QemuVirtualMachine::management_ipv4() -> std::optional<IPAddress>
+std::optional<mp::IPAddress> mp::QemuVirtualMachine::management_ipv4()
 {
     if (!management_ip)
         management_ip = qemu_platform->get_ip_for(desc.default_mac_address);


tests/mock_virtual_machine.h
```diff
@@ -37,14 +37,28 @@ template <typename T = VirtualMachine,
...
     template <typename... Args>
-    MockVirtualMachineT(Args&&... args)
+    explicit MockVirtualMachineT(Args&&... args)
         : MockVirtualMachineT{std::make_unique<TempDir>(), std::forward<Args>(args)...}
     {
     }
 
     template <typename... Args>
-    MockVirtualMachineT(std::unique_ptr<TempDir>&& tmp_dir, Args&&... args)
+        requires(std::is_same_v<VirtualMachine, T>)
```
jimporter: This is totally up to you, but I thought I'd mention it anyway: you can also use `std::same_as` here, which is the concept form of `std::is_same_v`.

tests/mock_virtual_machine.h
```diff
@@ -37,14 +37,28 @@ template <typename T = VirtualMachine,
...
+    explicit MockVirtualMachineT(std::unique_ptr<TempDir>&& tmp_dir, Args&&... args)
+        : T{std::forward<Args>(args)...}, tmp_dir{std::move(tmp_dir)}
+    {
+        setup_default_actions();
+    }
+
+    template <typename... Args>
+        requires(!std::is_same_v<VirtualMachine, T>)
```
jimporter: Ditto here.

src/network/ip_address.cpp
```diff
@@ -17,7 +17,11 @@
 
 #include <multipass/ip_address.h>
 
+#include <array>
+#include <cstdint>
 #include <sstream>
+#include <stdexcept>
+#include <string>
```
jimporter: Except for `stdexcept` (and `sstream`), these are all included in `ip_address.h`.
```suggestion
#include <sstream>
#include <stdexcept>
```

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
```diff
@@ -513,26 +504,22 @@ std::string mp::VirtualBoxVirtualMachine::ssh_username()
     return desc.ssh_username;
 }
 
-std::string mp::VirtualBoxVirtualMachine::management_ipv4()
+auto mp::VirtualBoxVirtualMachine::management_ipv4() -> std::optional<IPAddress>
```
jimporter: I don't have a strong opinion on trailing-return style, though I notice that we don't use it consistently across the codebase. Is there a reason for doing this here? Since there's no dependent-type stuff happening here, I don't think it's actually *necessary*...

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
```diff
@@ -513,26 +504,22 @@ std::string mp::VirtualBoxVirtualMachine::ssh_username()
...
 {
-    return "N/A";
+    return std::nullopt;
 }
 
-std::vector<std::string> mp::VirtualBoxVirtualMachine::get_all_ipv4()
+auto mp::VirtualBoxVirtualMachine::get_all_ipv4() -> std::vector<IPAddress>
```
jimporter: Ditto here.

tests/qemu/test_qemu_backend.cpp
```diff
@@ -345,7 +345,7 @@ TEST_F(QemuBackend, throwsWhenShutdownWhileStarting)
     while (machine->state != mp::VirtualMachine::State::off)
         std::this_thread::sleep_for(1ms);
 
-    MP_EXPECT_THROW_THAT(machine->ensure_vm_is_running(),
+    MP_EXPECT_THROW_THAT(dynamic_cast<mp::BaseVirtualMachine&>(*machine).ensure_vm_is_running(),
```
jimporter: What's the `dynamic_cast` for? I'm not sure I understand this bit.

jimporter (changes requested):
This looks good to me overall. I just have a few small comments about some of the details.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -277,58 +228,82 @@ void mp::BaseVirtualMachine::renew_ssh_session()
...
 }
+bool multipass::BaseVirtualMachine::unplugged()
+{
+    auto st = current_state();
+    return st == State::off || st == State::stopped;
+}
+
+void mp::BaseVirtualMachine::detect_aborted_start()
+{
+    const std::lock_guard lock{state_mutex};
```
xmkg: `const` is redundant here

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -277,58 +228,82 @@ void mp::BaseVirtualMachine::renew_ssh_session()
...
+    if (unplugged())
+    {
+        shutdown_while_starting = true;
+        state_wait.notify_all();
+
+        auto msg = std::string{"Instance shutdown during start"};
```
xmkg:
```suggestion
        std::string msg{"Instance shutdown during start"};
```

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -277,58 +228,82 @@ void mp::BaseVirtualMachine::renew_ssh_session()
...
-                current_state())) // spend time updating state only if we need a new session
+        if (!MP_UTILS.is_running(current_state())) // avoid wasting time
             throw SSHVMNotRunning{"SSH unavailable on instance {}: not running", vm_name};
     }
 
     mpl::debug(vm_name, "{} SSH session", ssh_session ? "Renewing cached" : "Caching new");
-
     ssh_session.emplace(ssh_hostname(), ssh_port(), ssh_username(), key_provider);
 }
+bool multipass::BaseVirtualMachine::unplugged()
```
xmkg, line 238: I wonder if we need a distinction between the `off` and `stopped` states, or even what the significance of that distinction would be..  I checked all the if conditions, and they're almost always checked together. I'd love to see either of these states gone if there's no valid excuse for it to exist.

src/platform/backends/shared/base_virtual_machine.h
```diff
@@ -87,15 +83,24 @@ class BaseVirtualMachine : public VirtualMachine
...
     virtual std::shared_ptr<Snapshot> make_specific_snapshot(const QString& filename);
     virtual std::shared_ptr<Snapshot> make_specific_snapshot(const std::string& snapshot_name,
                                                              const std::string& comment,
                                                              const std::string& instance_id,
                                                              const VMSpecs& specs,
                                                              std::shared_ptr<Snapshot> parent);
+
     virtual void drop_ssh_session(); // virtual to allow mocking
+    virtual bool unplugged();
+    virtual void refresh_start();
```
xmkg, line 106: I'd love to see what "refresh start" means here.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -847,3 +822,44 @@ void mp::BaseVirtualMachine::drop_ssh_session()
...
+        return log_and_retry(e, this);
+    }
+}
+
+void mp::BaseVirtualMachine::ssh_and_cross_to_running()
+{
+    static constexpr auto wait_step = 1s;
+    ssh_session.emplace(ssh_hostname(wait_step), ssh_port(), ssh_username(), key_provider);
+
+    std::lock_guard lock{state_mutex};
```
xmkg, line 854: I wonder if we should extract `lock-set state-update` to its own function for convenience.

src/rpc/multipass.proto
```diff
@@ -211,7 +211,6 @@ message InstanceDetails {
     string memory_usage = 5;
     string disk_usage = 6;
     repeated string ipv4 = 7;
-    repeated string ipv6 = 8;
```
xmkg, line 214: I think it would be better to reserve the field name & index instead, to prevent misuse and reserve it in case we need to actually implement the IPv6 support.

tests/mock_virtual_machine.h
ricab: Right, done. I replaced type traits with concepts for this file.

tests/mock_virtual_machine.h
ricab: Done

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
ricab: This is the rule that I try to follow *consistently*: if trailing return saves a namespace qualification on the return type, use it. Otherwise use the traditional format.

In this case, the alternative would be

```
std::optional<mp::IPAddress> mp::VirtualBoxVirtualMachine::management_ipv4()
              ~~~~
```

Now, because we already have namespace aliases (`mp`, `mpl`, etc.), it often ends up being more typing with trailing returns (as in here). But it saves the conceptual repetition: it implies the return in the same namespace, which would only need to be changed in one place in the signature if we wanted to modify it (e.g. rename).

It's a detail, not a big deal to me. We could decide something else. WDYT?

src/platform/backends/shared/base_virtual_machine.cpp
ricab: You think so? Why? This tells the reader right away that there will be no manual unlocking or funny stuff. And it prevents the writer from doing that by mistake.

src/platform/backends/shared/base_virtual_machine.cpp
ricab: OK, that's the same thing, but a little less typing. Fine by me.

src/platform/backends/shared/base_virtual_machine.cpp
xmkg: Forgive my ignorance, but AFAIK lock_guard provides no API functions to release the resource manually. Am I missing something here?

image

ricab: OK, I see what you mean. I still think it doesn't hurt and makes it clear to anyone who doesn't know the API by heart (like me...). FWIW, that page uses "const" in the example at the bottom too.

Anyway, I am fine with removing it also, so done.

src/platform/backends/shared/base_virtual_machine.cpp
ricab: Yeah I think they are redundant. I believe they were meant to have different meanings but never did. I considered removing one of them, but thought it was a little out of scope for this.

How about including that in the sequel? Would that be OK with you?

src/platform/backends/shared/base_virtual_machine.cpp
xmkg: Yeah, I just saw that too:

```c++
const std::lock_guard<std::mutex> lock(g_i_mutex);
```
I mean, it wouldn't hurt, but it practically changes no behavior.  I'm fine with it both ways.

src/platform/backends/shared/base_virtual_machine.cpp
xmkg: Of course! I'm OK with however you wish.

src/platform/backends/shared/base_virtual_machine.h
ricab: Right. I added some documentation. Let me know if you have ideas on how to make it better.

src/platform/backends/shared/base_virtual_machine.cpp
ricab: Yes, but that is something I was reserving for part 2.

src/rpc/multipass.proto
ricab: OK, done.

xmkg: I think my comment is a bit ambiguous. What I meant by "reserving" is actually to use the "reserve" keyword to reserve both index 8 and the "ipv6" name:

https://protobuf.dev/programming-guides/proto3/#reserved-field-numbers

This way, the protoc compiler won't emit any code (e.g., getter/setter), but the wire format would be aware of the reserved space.

https://protobuf.dev/programming-guides/proto3/#consequences

ricab: Oh, I see. It's a pitty that

> you can’t mix field names and field numbers in the same `reserved` statement

Anyway, done now.

ricab:
Hey @jimporter, I think this is ready for another pass, when you have a chance. Let me know your thoughts on the trailing returns!

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
jimporter: If I were making the call here, I'd go with the return type on the left, for two reasons:

1. The narrow reason: in practice, we don't save any typing here (and if we refactored/renamed, we'd probably do a global find-replace of `mp::` so we'd be ok; or else we'd use some syntax-aware renaming via `clangd` or something).
2. The broad reason: since most of our code has return types on the left, seeing a trailing return type makes me initially think, "There must be something unusual happening here. I should read this code more closely to understand it." In this case, that's a false alarm, so it's just a (very small) additional cognitive burden when reading this code.

    My default position on things like this is that similar constructs in our code should be written in a consistent style to emphasize their similarity to human readers.

However, I really don't have strong feelings here, and if you (or anyone) wanted trailing return types here, I wouldn't mind.

jimporter (approved):
This looks good to me, though I have a very weak preference for using leading return types (see my comment). Approved whichever way you go with that change though.

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
ricab: OK, I will do as you prefer. I believe the only relevant difference is the dependent lookup, dependent types, etc. Apart from that it doesn't affect the outcome. But I understand that it might require some getting used to and I am fine with the traditional form as well.

ricab:
Hey @jimporter, this should be good for another look. Thanks!

PR #4021, opened 2025-04-01
new catalogue page 

levkropp:
catalogue
This pull request includes significant updates to the `catalogue.dart` and `image_card.dart` files to enhance the user interface and functionality of the image catalogue in the GUI. The main changes involve adding support for non-Ubuntu distributions, refactoring the `ImageCard` widget, and improving the layout and styling of the catalogue screen.

### Enhancements to the Image Catalogue:

* Added a list of placeholder images for non-Ubuntu distributions, including Fedora, Debian, Arch, RHEL, SLES, and NixOS. (`src/client/gui/lib/catalogue/catalogue.dart`)
* Modified the `sortImages` function to include non-Ubuntu distributions in the returned list of images. (`src/client/gui/lib/catalogue/catalogue.dart`)
* Introduced the `_groupAndCreateCards` function to group images by type (Ubuntu, Ubuntu Core, Appliances, and other distributions) and create corresponding `ImageCard` widgets with descriptions. (`src/client/gui/lib/catalogue/catalogue.dart`)

### Refactoring and UI Improvements:

* Updated the `ImageCard` widget to be a `ConsumerStatefulWidget`, allowing dynamic updates to the selected image version. Added methods to determine the appropriate logo, distributor, and display title based on the operating system. (`src/client/gui/lib/catalogue/image_card.dart`)
* Enhanced the `ImageCard` widget to include a description, a version dropdown for Ubuntu images, and improved styling. (`src/client/gui/lib/catalogue/image_card.dart`)
* Improved the layout of the `CatalogueScreen` by replacing the divider with a `SizedBox` for spacing and adjusting the card layout to support a maximum of three columns with reduced spacing. (`src/client/gui/lib/catalogue/catalogue.dart`) [[1]](diffhunk://#diff-384559be6b7e94095025283d5c36a158d3b4a25978d0b5f258355f2a930cfb43L124-R262) [[2]](diffhunk://#diff-384559be6b7e94095025283d5c36a158d3b4a25978d0b5f258355f2a930cfb43L135-R291)

### Styling Adjustments:

* Updated the welcome text and description on the `CatalogueScreen` to use a lighter font weight for better readability. (`src/client/gui/lib/catalogue/catalogue.dart`)

PR #4474, opened 2025-10-23
[feature] Updated catalogue page

sharder996:
This PR implemented the design done by @vikorama which can is demonstrated in [Figma](https://www.figma.com/proto/PVfFgoRyB8XiaWfKh4MeiS/New-catalogue?node-id=29-7079&p=f&t=n0tvM2KZRTY3cCgG-1&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=29%3A7079&show-proto-sidebar=1).

The notable changes are the grouping of releases under a given OS as well as graphics for Debian and Fedora images.

See [MULTI-1960](https://warthogs.atlassian.net/browse/MULTI-1960) for info on the design.

Screenshot from 2025-11-26 14-27-20

Closes #4021

---

MULTI-2356

[MULTI-1960]: https://warthogs.atlassian.net/browse/MULTI-1960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/gui/lib/catalogue/image_card.dart
+++ b/src/client/gui/lib/catalogue/image_card.dart
@@ -111,6 +111,9 @@ class ImageCard extends ConsumerWidget {
               width: double.infinity,
               decoration: BoxDecoration(
                 color: const Color(0xfff5f5f5),
+                border: Border(
+                  bottom: BorderSide(color: Colors.black, width: 1),
+                ),
               ),
               padding: const EdgeInsets.symmetric(horizontal: 8),
               child: DropdownButton<String>(
@@ -118,6 +121,7 @@ class ImageCard extends ConsumerWidget {
                 icon: const Icon(Icons.keyboard_arrow_down),
                 focusColor: Colors.transparent,
                 isExpanded: true,
+                underline: const SizedBox(),
                 items: versions
                     .map((v) => DropdownMenuItem(
                           value: v.release,


vikorama:
@sharder996 Looking great! A couple of adjustments I'd do to the cards:


1. The cards in one row are always the same size, but the rows can have different height if there's not enough content (hopefully with time we ensure all images are "advertised" in the same way). There shouldn't be so much white space, unless necessary for row alignment.
image

2. The dropdown should be at the same position in the row

3. The dropdown should match the dropdown style from the settings (chevron not filled, light-grey background **F5F5F5** hex code).

4. The outline of the cards should have higher contrast with the background. In our design system it's called high-contrast with the **707070** Hex code.

5. [Bonus - not related to this issue specifically] Overall, in our Design system the buttons don't have border radius (the corners are not rounded). I don't know if it's an easy fix in Flutter, but would significantly increase consistency with the Design system.

sharder996:
@vikorama I've addressed all your points and hopefully this is closer to the design. LMK

xmkg:
So I've tested the GUI and tried to resize the window. The "core images" box's combobox state becomes inconsistent when you make the window smaller horizontally:
Screenshot from 2025-11-07 10-19-26
Screenshot from 2025-11-07 10-19-41

In this instance, the "ubuntu core" combobox gets unselected. When you try to select something, it remembers that for a while for the "collapsed" state:

Screenshot from 2025-11-07 10-19-49
Screenshot from 2025-11-07 10-19-59

Tried launching instances from both states -- it launches what the combobox says.

sharder996:
🤦 I'm remembering why I avoided web development like the plague

amylily1011:
When the images are on the same row they seem to be the same size 👍 
Screenshot 2025-11-25 at 17 02 59

When they are in two different rows, ones without dropdown seems smaller, making it look disproportion to the upper one. ⚠️ 
Screenshot 2025-11-25 at 17 03 20


When they aligned in columns, the size are also quite different. ⚠️ 
Screenshot 2025-11-25 at 17 04 28


I think the box should be the same sizes regardless of whether or not there isa dropdown. Do you think we can make the box height the same on all of them?

**Dropdown 🤔** 
Screenshot 2025-11-25 at 17 10 19
If we want to strictly make it look like drop downs in the vanilla design system (like in @vikorama 's comment), the gray line should be longer covering the side padding (and the color is black) . The bottom padding under the gray line should be zero.

vikorama:
> When they are in two different rows, ones without dropdown seems smaller, making it look disproportion to the upper one. ⚠️ 

@amylily1011 I’m afraid it’s on me (see my comment above in this thread). The cards appeared too empty with a lot of white space when they all were the same size, so I suggested adjusting the size. 

Maybe we can add the release dropdown to all images in the catalogue (Debian, Fedora)? But it’d be greyed out if there’s no choice. Do we have that info on non-Ubuntu images?

sharder996:
@vikorama No, at the moment there is only one release offered for Debian and Fedora images.

vikorama:
> @vikorama No, at the moment there is only one release offered for Debian and Fedora images.

If it's just one release, can we still display it? I meant [smth like this](https://canonical.github.io/react-components/?path=/story/components-multiselect--disabled):
image

amylily1011:
> > @vikorama No, at the moment there is only one release offered for Debian and Fedora images.
> 
> If it's just one release, can we still display it? I meant [smth like this](https://canonical.github.io/react-components/?path=/story/components-multiselect--disabled): image

This LGTM

sharder996:
@amylily1011 @vikorama 
Made a few tweaks to the image cards. PR description is updated with a new screenshot.

amylily1011 (approved):
LGTM

xmkg:
Looks good to me, and works as expected (Linux).

PR #4476, opened 2025-10-24
Bugfix/cloud init restart qemu

### LINKED ISSUE #4199: Support Cloud-Init Reboots in multipass launch
I’m trying to use `multipass launch` with `cloud-init` to provision a VM that installs the `ubuntu-advantage-tools` package, attaches an Ubuntu Pro token, and then reboots. This is to automate enabling Ubuntu Pro in a VM as part of CI workflows.

Support rebooting a VM during cloud-init when using `multipass launch --cloud-init`. Currently, if a `reboot` command is issued via `runcmd`, `multipass` times out with the error:

```
```

However, despite the failure, the VM does start successfully and is usable. The issue appears to be that `multipass` does not handle reboots during the cloud-init phase gracefully, causing it to misinterpret the VM state.

A potential solution would be for `multipass` to detect reboot cycles during cloud-init and wait until the instance stabilizes before determining success or failure.

**Additional context**
Here’s an example of the `cloud-init` config that causes the issue:

```yaml
package_update: true
package_upgrade: true
packages:
  - ubuntu-advantage-tools
runcmd: |
  - pro attach <redacted> --no-auto-enable
  - reboot
```

Command used:

```bash
echo "<cloud-init-config-above>" | sudo multipass launch --name my-multipass-vm --cloud-init - --cpus 4 --disk 20GB --memory 8GB 22.04
```

After the timeout, the VM is running and responds to commands via `multipass exec`, and `pro status` shows the token was successfully attached.

This feature would be helpful for fully automated provisioning flows that involve system reboots as part of setup.

tobe2098:
This PR allows reboots during cloud-init configurations on `multipass launch`. The current solution is to re-establish an ssh session when the state is not running during a cloud-init's `ssh_exec("[ -e /var/lib/cloud/instance/boot-finished ]")`, and if established it sets the state to Running. This solution has a problem, which is that there is a nested try_action_for. This means that the timeout for cloud-init can potentially duplicate (case: cloud-init waits until end of timeout, then machine is not running, wait_ssh is run within with the same timeout). 

An alternative solution would require that qemu machines on restart go through the wait_for_ssh_up function twice on a restart (with the exception of cloud-init reboots, in which they only run that function once per restart). Which could be considered its own drawback. (This solution is not feasible, it would require backend-wise code in daemon.cpp)

The last solution, which is to actively query the qmp process, will return the VM machine to a running state as soon as the restart happens. From QEMU's perspective (and possibly from the other hypervisors' as well), if the virtual CPU is running, the VM is running. Booting detection would only be a requirement for the qemu guest agent, which we currently have disabled in snapcraft.yaml.

Fixes #4199

MULTI-2293

### DIFF
--- a/src/platform/backends/shared/base_virtual_machine.cpp
+++ b/src/platform/backends/shared/base_virtual_machine.cpp
@@ -38,7 +38,6 @@
 #include <QDir>
 
 #include <chrono>
-#include <functional>
 #include <mutex>
 #include <stdexcept>
 


tobe2098:
In the end, to avoid nested try_action_for (by nesting a wait_for_ssh_up inside the _wait_for_cloud_init), the chosen solution is to actively query the VM state when calling current_state(). The query is async, which means that a single call will not be enough to get the immediate state. The only states that QMP recognizes are found in https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#vm-run-state. Out of those, the only I can say with confidence overlaps with our usage of QMP is the running state. 

This solution does not work well because there seems to be conflicts with the ssh connections and the unknown state. The qemu VMs are either in the running state (from QMP's viewpoint) or the qemu process is not running and the daemon keeps track of which state the qemu machine is supposed to be. Thus the QMP command "query-state" will always return running if the process is running or not be sent if the process is not running. At which point it conflicts with the unknown state (we do not want to try to ssh that VM), as the state is updated to running when the state is checked. A fix to this would be to not run the command in that case. But then it becomes the boolean (state != UNKNOWN && vm_process && vm_process->running()). Since the daemon framework does not follow the QMP "query-state" very neatly, I think this solution is not very good.

In the end, what seems to be the proper solution is that if cloud-init finds itself trying to ssh a non-running VM, that it attempts to establish a ssh_session. To do so I encapsulated the lambda from wait_until_ssh_up_helper in another function to be able to use it in both places.

Additionally, I added logging QMP errors as errors in the daemon. I am open to discussion about whether the log level should be otherwise, or if it is necessary at all taking into account that the json will be logged as debug.

include/multipass/exceptions/ssh_exception.h
```diff
@@ -46,4 +46,12 @@ class SSHExecFailure : public SSHException
...
     int ec;
 };
+
+class SSHVMNotRunning : public SSHException
+{
+public:
+    explicit SSHVMNotRunning(const std::string& what_arg) : SSHException(what_arg)
+    {
+    }
+};
```
xmkg:
```suggestion
struct SSHVMNotRunning : SSHException
{
    using SSHException::SSHException;
};
```

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -616,6 +616,11 @@ void mp::QemuVirtualMachine::initialize_vm_process()
                 }
             }
         }
+        else if (qmp_object.contains("error"))
+        {
+            auto error = qmp_object["error"].toObject();
```
xmkg: nitpick: can be `const`

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -273,7 +280,8 @@ void mp::BaseVirtualMachine::renew_ssh_session()
         const std::unique_lock lock{state_mutex};
         if (!MP_UTILS.is_running(
                 current_state())) // spend time updating state only if we need a new session
-            throw SSHException{fmt::format("SSH unavailable on instance {}: not running", vm_name)};
+            throw SSHVMNotRunning{
```
xmkg: SSHVMNotRunning could inherit from https://github.com/canonical/multipass/blob/main/include/multipass/exceptions/formatted_exception_base.h instead.

PR #4478, opened 2025-10-24
[cmake] Add option to force enabling VirtualBox

ricab:
It used to be possible to build and use the `virtualbox` driver on Linux, but that got lost somewhere along the way. 

So add FORCE_ENABLE_VIRTUALBOX, an option to force enabling the VirtualBox backend. It applies to Linux and macOS only, as VirtualBox is always enabled on Windows. This is useful to test VirtualBox changes from Linux.

Also fix variable overwrite.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,9 @@ endif()
 
 cmake_host_system_information(RESULT HOST_OS_NAME QUERY OS_NAME)
 
+include(CMakeDependentOption)
+cmake_dependent_option(FORCE_ENABLE_VIRTUALBOX "Enable VirtualBox" ON UNIX OFF)
+
 if("${HOST_OS_NAME}" STREQUAL "macOS")
   # needs to be set before "project"
   set(VCPKG_HOST_OS "osx")
@@ -167,7 +170,7 @@ if(UNIX)
     set(HOST_ARCH "aarch64")
   endif ()
 
-  if (APPLE AND ${HOST_ARCH} MATCHES "x86_64")
+  if (FORCE_ENABLE_VIRTUALBOX OR (APPLE AND ${HOST_ARCH} MATCHES "x86_64"))
       list(APPEND MULTIPASS_BACKENDS virtualbox)
   endif()
 


tobe2098:
Only meaningful info I got is that it is related to ssh and networking:
```
[2025-11-21T18:25:17.154] [debug] [a] Waiting for SSH to be up
[2025-11-21T18:25:17.154] [debug] [a] Running: VBoxManage, controlvm, a, natpf1, delete, ssh
[2025-11-21T18:25:17.210] [debug] [a] Running: VBoxManage, controlvm, a, natpf1, ssh,tcp,,39525,,22
[2025-11-21T18:25:17.263] [trace] [a] ssh connection failed: 'Connection refused'
```
After that it just keeps failing every second or so.

Also, there is no `dnsmasq` process running with the daemon, which I do not know if it is critical or it gets the IPs in some other way.

Switching back to qemu (after loading kvm modules) works fine.

ricab:
Thanks @tobe2098. 

> LGTM, tested in linux only since macos intel has virtualbox enabled by default. There is a small issue when using virtualbox, after disabling the kvm modules to allow virtualbox to run, I could not launch a virtualbox vm. Looking from the GUI the VM is running but it does not get an IP (at least none that multipassd sees) and there is only the login tty1, where i cannot get past.

Right, this wasn't meant to make VirtualBox fully functional on Linux. The goal is just letting us enable it for development purposes.

> Also, there is no `dnsmasq` process running with the daemon, which I do not know if it is critical or it gets the IPs in some other way.

Dnsmasq is used only for QEMU on Linux. But in Virtualbox, there is no meaningful IP that we could access from the host. Instead, we use port forwarding. There's probably something wrong with it on Linux or we are not ssh-ing correctly or something.

The backend will be deprecated and removed, but this is useful to us developers because it let's us deal with it from Linux, at least to an extent.

tobe2098:
Oh, I see. Do you mean to test compilation and tests? In that case, since it is only for dev, it LGTM. I cannot approve manually right now since I am away, but go ahead and feel free to queue in the merge queue.

tobe2098 (approved):

ricab:
Essentially, yes. And whatever us developers are willing and able to coerce it to do :slightly_smiling_face:

PR #4480, opened 2025-10-29
[cmake] Flex vcpkg setup a little

ricab:
Support overriding the location of vcpkg. This provides a workaround to build on Windows with the source in a network location (samba mount in my case). 

Windows (wisely) refuses to execute vcpkg scripts from a network location (even with the Unrestricted ExecutionPolicy). This allows using a local vcpkg while still building the project from a remote source.

Also fix configuring from PowerShell by explicitly bootstrapping in cmd.

### DIFF
--- a/BUILD.macOS.md
+++ b/BUILD.macOS.md
@@ -123,7 +123,7 @@ The second command runs Xcode's first launch setup, which installs additional co
 
 CMake will fetch all necessary content, build vcpkg dependencies, and initialize the build system.
 You can also specify the `-DCMAKE_BUILD_TYPE` option to set the build type (e.g., `Debug`,
-`Release`, `Coverage`, etc.).
+`Release`, etc.).
 
 To use a different vcpkg, pass `-DMULTIPASS_VCPKG_LOCATION="path/to/vcpkg"` to CMake.
 It should point to the root vcpkg location, where the top bootstrap scripts are located.
--- a/BUILD.windows.md
+++ b/BUILD.windows.md
@@ -142,7 +142,7 @@ specify the vcpkg toolchain manually with
 `-DCMAKE_TOOLCHAIN_FILE=..\3rd-party\vcpkg\scripts\buildsystems\vcpkg.cmake `.
 Use `CMAKE_PREFIX_PATH` to select a specific Qt version.
 To specify the build type, use `-DCMAKE_BUILD_TYPE` option to set the build type (e.g.,
-`Debug`, `Release`, `Coverage`, etc.).
+`Debug`, `Release`, etc.).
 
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -69,24 +69,26 @@ endif()
 
 message(STATUS "Bootstrapping vcpkg...")
 
-if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
-  set(VCPKG_BOOTSTRAP_EXTENSION "bat")
-else()
-  set(VCPKG_BOOTSTRAP_EXTENSION "sh")
-endif()
+set(VCPKG_LOCATION "${CMAKE_CURRENT_SOURCE_DIR}/3rd-party/vcpkg" CACHE PATH
```
xmkg: Can we briefly mention this in our build docs?

xmkg: also nitpick: vcpkg itself uses the `VCPKG_` prefix, so it might be a good idea to prefix this with `MULTIPASS_`

ricab: Sure, will do!

ricab: Done. Let me know how it looks.

BUILD.macOS.md
```diff
@@ -47,24 +47,32 @@ Building
 
     cd <multipass>
     git submodule update --init --recursive
+    mkdir build
+    cd build
+
+CMake will fetch all necessary content, build vcpkg dependencies, and initialize the build system.
+You can also specify the `-DCMAKE_BUILD_TYPE` option to set the build type (e.g., `Debug`,
```
sharder996, line 125: `clang` doesn't support lcov so it maybe we shouldn't mention `Coverage` here?

BUILD.windows.md
```diff
@@ -135,15 +135,26 @@ mkdir build
 cd build
 ```
 
+CMake will automatically fetch all necessary content, build vcpkg dependencies, and initialize the
+build system. You can also specify the `-DCMAKE_BUILD_TYPE` option to set the build type (e.g.,
+`Debug`, `Release`, `Coverage`, etc.).
```
sharder996: Again, not sure if `Coverage` is supported with msvc.

sharder996 (changes requested):
Remove `no-merge` tag?

BUILD.macOS.md
ricab: OK, removed.

BUILD.windows.md
ricab: Removed.

ricab:
I have rebased this and adapted build instructions a little, as they had changed in the meantime. I also addressed review requests and removed the no-merge tag, so this should be ready for another pass.

xmkg (approved):

PR #4491, opened 2025-11-04
[utils] Remove old Qt JSON utils

jimporter:
It's finally happening! There's no longer any Qt JSON code in the tree, so we can remove the old utility functions from `json_utils.h`.

### DIFF
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -38,6 +38,7 @@
 
 #include <QFile>
 #include <QProcess>
+#include <QRegularExpression>
 #include <QString>
 #include <QTemporaryFile>
 
--- a/include/multipass/cli/alias_dict.h
+++ b/include/multipass/cli/alias_dict.h
@@ -28,8 +28,6 @@
 
 #include <boost/json.hpp>
 
-class QJsonObject;
-
 namespace multipass
 {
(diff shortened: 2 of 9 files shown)


src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -38,6 +38,7 @@
 
 #include <QFile>
 #include <QProcess>
+#include <QRegularExpression>
```
tobe2098, line 41: Why this addition?

tobe2098 (changes requested):
Very nice Jim! Just one small nitpick, otherwise it looks good to me

src/platform/backends/qemu/qemu_virtual_machine.cpp
jimporter: We use `QRegularExpression` in `extract_snapshot_tags`, and I think the QJson headers transitively included the `QRegularExpression` header. So now we have to include it explicitly.

jimporter:
This is currently failing because it needs to be synchronized with the merger of the AZs branch. There's one last QJson bit that slipped through the cracks from that branch...

jimporter:
> This is currently failing because it needs to be synchronized with the merger of the AZs branch. There's one last QJson bit that slipped through the cracks from that branch...

Now fixed.

tobe2098 (approved):
LGTM!

PR #4493, opened 2025-11-04
[tests] Fix mock delegation on base classes

ricab:
Fix delegation of mock calls on base classes when the method returns a
reference. To that end, make sure the underlying lambda keeps the right
return type. This avoids returning dangling references to temporaries.

### DIFF
--- a/tests/common.h
+++ b/tests/common.h
@@ -76,9 +76,10 @@
 //     `MP_DELEGATE_MOCK_CALLS_ON_BASE_WITH_MATCHERS(mock_widget, Widget, render, (A<Canvas>()))`
 // This will redirect the version of `MockWidget::render` that takes one argument of type `Canvas`.
 #define MP_DELEGATE_MOCK_CALLS_ON_BASE_WITH_MATCHERS(mock, method, BaseT, ...)                     \
-    ON_CALL(mock, method __VA_ARGS__).WillByDefault([m = &mock](auto&&... args) {                  \
-        return m->BaseT::method(std::forward<decltype(args)>(args)...);                            \
-    })
+    ON_CALL(mock, method __VA_ARGS__)                                                              \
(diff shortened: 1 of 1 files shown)


ricab:
An example method that would lead to UB without this PR, when used with `MP_DELEGATE_MOCK_CALLS_ON_BASE_WITH_MATCHERS`: https://github.com/canonical/multipass/blob/36d3d74287716bea2cefd437926dfc643941662d/tests/mock_virtual_machine.h#L130

ricab:
A demonstration of the `decltype(auto)` return: https://godbolt.org/z/M53nbKqPj

ricab:
Thanks for the review @tobe2098! I believe const-ref returns were also problematic. I had this with one such case (the `get_name` method I linked above).

The default action would end up returning a reference to a local copy that got lost. Lifetime extension wouldn't apply. See: https://godbolt.org/z/xEsTheacx

From https://en.cppreference.com/w/cpp/language/reference_initialization.html#Lifetime_of_a_temporary :

> a temporary bound to a return value of a function in a return statement is not extended: it is destroyed immediately at the end of the return expression. Such return statement always returns a dangling reference. 

The action has an implicit return somewhere. I think in the past we even had to use `.WillByDefault(Return(...))` but gmock now lets us omit the `Return` part.

tobe2098:
Yes, I think I understand now: the base method returns the reference, the lambda deduces a non reference, which it returns, and then that non-reference is converted to a reference because the mock expects a reference.
Again, really good catch!

jimporter (approved):
Looks good to me, and I agree with everything @tobe2098 said. This is the Right Thing to do here. (Funnily enough, I was actually just looking at this macro since I had to start using it in the Boost.JSON migration.)

ricab:
Thank you both for your reviews!

PR #4496, opened 2025-11-04
[macOS] Backwards compatible builds

sharder996:
This PR adds the ability to build on the latest macOS runners while still targeting an older version of macOS.

The `CMAKE_OSX_DEPLOYMENT_TARGET` is propagated to vcpkg so that all dependencies are also built for the correct version of macOS.

Because OpenSSL was originally fetched via brew installation, its binary would inherently target the version of macOS that it was being installed on. To remedy this, OpenSSL is now fetched from vcpkg.

`windows-macos.yml` is subsequently pruned to no longer spawn unneeded macOS runners.

The deployment target can be queried from a Multipass binary with the command `otool -l multipassd | grep -A 3 LC_BUILD_VERSION`.

Updated licence information in the installer with the appropriate licence as found in the [openssl repository](https://github.com/openssl/openssl). Also, updated the licence information for semver as its source recently changed in #4461.

Verified on a macOS 14 virtual machine. Nested virtualization is only supported on M3 and newer, so you will need to remove the CPU checks in `preinstall-multipassd.sh` and `CPack.distribution.dist.in`.

Closes https://github.com/canonical/multipass/pull/4023
Closes https://github.com/canonical/multipass-private/pull/739

---

MULTI-1920
MULTI-2272

### DIFF
--- a/BUILD.macOS.md
+++ b/BUILD.macOS.md
@@ -38,7 +38,7 @@ Install Qt6:
 ### Cmake
 
 Building a Multipass package requires cmake 3.9 or greater. The most convenient
-means to obtain these dependencies is with Homebrew <https://brew.sh/>.
+means to obtain cmake is with Homebrew <https://brew.sh/>.
 
     brew install cmake
 


CMakeLists.txt
```diff
@@ -61,6 +62,16 @@ if(NOT VCPKG_BUILD_DEFAULT)
     set(VCPKG_HOST_ARCH "arm64")
   endif()
 
+  # Propagate macOS deployment target to vcpkg triplet
+  if(APPLE AND DEFINED CMAKE_OSX_DEPLOYMENT_TARGET)
+    configure_file(
```
xmkg, line 68: Can we configure these files to the bin dir so it won't pollute the source tree?

sharder996: Good point! I'll do that.

packaging/windows/wix/multipass.wixproj
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project Sdk="WixToolset.Sdk/5.0.0">
+<Project Sdk="WixToolset.Sdk/5.0.2">
```
ricab, line 2: Any special reason to update this, or just moving to the latest? It's fine, just curious.

packaging/windows/wix/multipass.wixproj
```diff
@@ -36,6 +36,9 @@
     </ItemGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\custom-actions\custom-actions.vcxproj" />
+        <ProjectReference Include="..\custom-actions\custom-actions.vcxproj">
+            <Project>{D4E3F6C1-1A2B-4C5D-8E9F-0A1B2C3D4E5F}</Project>
+            <SetPlatform>Platform=x64</SetPlatform>
+        </ProjectReference>
```
ricab, line 42: And what is this about?

src/platform/platform_win.cpp
```diff
@@ -60,12 +60,6 @@
 
 #include <libssh/sftp.h>
 
-// Needed for OpenSSL dll compatibility
-extern "C"
-{
-#include <openssl/applink.c>
-}
-
```
ricab, line 68: How did you come to the conclusion that this could now be removed?

CMakeLists.txt
```diff
@@ -30,11 +30,12 @@ if(NOT ${IS_RUNNING_IN_CI})
 endif()
 
 cmake_host_system_information(RESULT HOST_OS_NAME QUERY OS_NAME)
+cmake_host_system_information(RESULT HOST_ARCH QUERY OS_PLATFORM)
 
 if("${HOST_OS_NAME}" STREQUAL "macOS")
   # needs to be set before "project"
   set(VCPKG_HOST_OS "osx")
-  if (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64")
```
ricab, line 37: Was this not working?

CMakeLists.txt
```diff
@@ -30,11 +30,12 @@ if(NOT ${IS_RUNNING_IN_CI})
 endif()
 
 cmake_host_system_information(RESULT HOST_OS_NAME QUERY OS_NAME)
+cmake_host_system_information(RESULT HOST_ARCH QUERY OS_PLATFORM)
```
ricab, line 33: There is already a `HOST_ARCH` variable below and some tinkering between arm/aarch. Could that perhaps be unified here?

packaging/windows/wix/multipass.wixproj
sharder996: Just moving to the latest.

packaging/windows/wix/multipass.wixproj
sharder996: Without specifying this, my dev environment was having trouble determining what architecture to build for. I didn't find that it affected anything unless we want to build for Windows ARM.

CMakeLists.txt
sharder996: `CMAKE_HOST_SYSTEM_PROCESSOR` is only populated after `project` is called causing `CMAKE_OSX_DEPLOYMENT_TARGET` to be empty.

src/platform/platform_win.cpp
sharder996: It's an old way to including `applink` which isn't compatible with how vcpkg fetches OpenSSL. It's now replaced with `OpenSSL::applink` in `CMakeLists.txt`.

packaging/windows/wix/multipass.wixproj
ricab: OK, we probably will, at some point, but I guess we can cross that bridge when we come to it.

src/platform/platform_win.cpp
ricab: OK, thanks.

CMakeLists.txt
ricab: Oh so CMAKE_OSX_DEPLOYMENT_TARGET used to be empty in arm builds?

CMakeLists.txt
ricab: Thanks!

CMakeLists.txt
sharder996: Yes, we were getting lucky because the default deployment target was the OS version of the runner Multipass was being built on.

ricab (approved):
LGTM. Thanks again, great to have this restraint lifted :+1:

ricab:
BTW, tested on Windows and macOS (Intel). Both worked fine.

PR #4501, opened 2025-11-11
[docs] Correct Qt paths in BUILD.windows

tobe2098:
This PR is to correct a couple of inaccuracies in the BUILD.windows.md file. 
1. The PATH variable to Qt still follows the convention, at least if installed following the guide (aqt with arch specified). 
2. The path given to cmake in Qt should be updated to the new version and arch.


tobe2098:
~~Moving openssl and grpc to vcpkg has also broken the installation process as it existed before, so we can also fix it in this PR if we find a solution.~~
Although not tried in a new installation, making sure vcpkg is not installed via MSVC fixes this issue.

PR #4502, opened 2025-11-12
[cxx20] Use defaulted comparison operators

jimporter:
This replaces a few manually-implemented comparison operators with the defaults. I guess these got missed in the big C++20 PR.

### DIFF
--- a/include/multipass/cloud_init_iso.h
+++ b/include/multipass/cloud_init_iso.h
@@ -42,18 +42,12 @@ public:
     void write_to(const std::filesystem::path& path);
     void read_from(const std::filesystem::path& path);
 
-    friend bool operator==(const CloudInitIso& lhs, const CloudInitIso& rhs)
-    {
-        return lhs.files == rhs.files;
-    }
+    friend bool operator==(const CloudInitIso& lhs, const CloudInitIso& rhs) = default;
 
 private:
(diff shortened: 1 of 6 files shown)


Sploder12:
Doesn't build on MacOS, see discussion of IPAddress comparisons on my subnet refactor PR for context

jimporter:
> Doesn't build on MacOS, see discussion of IPAddress comparisons on my subnet refactor PR for context

Ah ha. Since you're already working on that, I'll just stay out of your way and remove the `IPAddress` change from this PR.

tobe2098 (changes requested):
Ubuntu builds and runs the tests without problems, but Windows 11 Pro fails the test: `Utils/NormalizeMountTargetTest.mountTargetsNormalizeCorrectly/3, where GetParam() = ("../..//folder","/folder")`. I am unsure if this is due to this PR, since the CI pipeline has been changed to use Windows Server 2025.

jimporter:
> After some research I found that it may be to a missing rebase to origin/main, since in main it is fixed, and the normalize target mounts utils function is different due to your commit, so I will be approving this.

Thanks. I've rebased so hopefully everything passes CI now.

tobe2098:
It does pass now. Great job!

PR #4505, opened 2025-11-13
[docs] Add instructions for windows build to run the GUI

tobe2098:
Added missing instructions on how to run the GUI, which is fairly simple. I also added the addition of putting the `<multipass>/build/bin` in the path for convenience, and corrected a possible mistake in the installation of qemu/qemu-img.

### DIFF
--- a/BUILD.windows.md
+++ b/BUILD.windows.md
@@ -170,10 +170,10 @@ Enable-WindowsOptionalFeature -Online -FeatureName:Microsoft-Hyper-V -All
 ### Run `multipass`
 
 With the `multipassd` daemon now running on another shell (or as a windows service) you can now run `multipass`.
-You can also add `<multipass>\build\bin` to the Path environment variable to be able to run the commands from anywhere.
 
 1. Press Windows Key + X, Select Windows PowerShell, or Terminal.
-2. Then, try `multipass help`.
+2. [Extend the Path variable](./BUILD.windows.md#Extend the Path variable).
+3. Then, try `multipass help`.
 
 ### Permissions/privileges for `multipassd`
 
@@ -190,9 +190,13 @@ needs to be part of the Hyper-V Administrators group:
 
 ### Run `multipass.gui`
 
-Add to the Path environment variable the following paths:
+1. [Extend the Path variable](./BUILD.windows.md#Extend the Path variable).
+2. Now you can run `multipass.gui`.
 
-- `<multipass>\build\bin`
-- `<multipass>\build\bin\windows\x64\runner\Release`
+### Extend the Path variable
 
-Now you can run `multipass.gui`.
+To avoid conflict with a multipass installation, include the necessary paths to the Path variable in your current session only:
+```
+$env:Path += ";<multipass>\build\bin"
+$env:Path += ";<multipass>\build\bin\windows\x64\runner\Release"
+```


BUILD.windows.md
```diff
@@ -15,7 +15,7 @@ After chocolatey is installed you can now install the rest of the dependencies f
 Powershell(Admin). To get the best results, in the following order:
 
 ```[pwsh]
-choco install cmake ninja qemu openssl git wget unzip -yfd
+choco install cmake ninja qemu-img openssl git wget unzip -yfd
```
ricab: It looks like `qemu-img` is stuck at 2.3.0 in choco, whereas `qemu` comes with qemu-img 10.1.0. I see you replaced it in https://github.com/canonical/multipass/commit/56402e14a6790e49bac53cf96f41b035b8b19ff0 (inadvertently?) but I wonder if we should use such an outdated version.

BUILD.windows.md
```diff
@@ -170,6 +170,7 @@ Enable-WindowsOptionalFeature -Online -FeatureName:Microsoft-Hyper-V -All
 ### Run `multipass`
 
 With the `multipassd` daemon now running on another shell (or as a windows service) you can now run `multipass`.
+You can also add `<multipass>/build/bin` to the Path environment variable to be able to run the commands from anywhere.
```
ricab: Invert slashes?

But also, how about introducing a separate section with instructions for extending the Path, so that it could be referred to from `### Run multipass{,.gui}`?

BUILD.windows.md
```diff
@@ -186,3 +187,12 @@ needs to be part of the Hyper-V Administrators group:
 2. Under System Tools->Local Users and Groups->Groups
 3. Select on Hyper-V Administrators, add your account
 4. Sign out or reboot for changes to take effect
+
+### Run `multipass.gui`
+
+Add to the Path environment variable the following paths:
+
+- `<multipass>\build\bin`
+- `<multipass>\build\bin\windows\x64\runner\Release`
```
ricab: It would be nice to include a PowerShell command to add these. Preferably in the current shell (and children) only, to avoid conflicts with installed Multipass packages (e.g. new libraries or executables present here but not there, so cross loaded between installs)

BUILD.windows.md
tobe2098: The reason I changed it (or maybe we should add it instead?) Was that just with qemu multipass hangs on launch being unable to convert to the vhdx format without `qemu-img`. Once I installed it it worked fine, buy before it would not be able to launch an instance.

ricab: Do you mean the reason why you changed it back, or why you changed to `qemu` at first? And did you have `qemu-img` in the path?

Anyway, I think I remember having similar problems with newer qemu, so fine for now, I guess.

tobe2098: At first I think we noticed the version and kept qemu for being newer, but now I found this issue with missing qemu-img. I have added it back (now there are the two of them), until we have an opportunity to test with a clean install again. Both qemu and qemu-img are added to the Path by choco I think.

tobe2098: From uninstalling `qemu-img` and installing `qemu` it seems like choco does not add the second one to Path. Should we suggest the use of `qemu-img` or should we suggest a Path fix?

tobe2098: After further research, `qemu-img` is the utility that we need, not `qemu`, since multipass does no emulation in windows. It is a sub-utility of `qemu`, not part of the package necessarily, so we could drop installing `qemu` as a whole for the utility. Apart from the apparent Path problems with Chocolatey.

tobe2098: Apparently, choco installs qemu in a separate folder instead of inside chocolatey/bin/, which is added to the path, while the other is not. Uninstalling also has weird behavior, there are files left behind.

ricab: Hmm, so the decision is between a severily outdated qemu-img or installing a bunch of stuff we don't need and adding it to the path. Did launch work for you with the qemu folder in the path?

tobe2098: More or less. I have not tried, but qemu-img is in that folder, so it is a solid assumption that adding to the path will work. Since it is so outdated, maybe it is worth getting the whole QEMU package. If we are already changing the path for cmake and qt, it would just be one more.

tobe2098:
Since our own Windows VM setup script (authored by @xmkg) relies on the `qemu-img` choco package I think it is best to leave it as is.

Issue #4506, opened 2025-11-13
[GUI] Default behavior on close not reflected in settings


tobe2098:
**Describe the bug**
The first time you run `multipass.gui` in your system it behaves as if you set it to close all running instances on exiting the GUI. When looking at the settings this is not the set option. If the option is "re-set" to "ask", the GUI behaves as normal.  

**To Reproduce**
From a clean system:
1. Build multipass
2. Run multipass.gui
3. Close the GUI with a running instance
4. The GUI will not ask to close
5. The settings option is set to "ask"

**Expected behavior**
It should behave as the default settings option indicates it should: to ask to close or keep the instances running when the GUI is closed.

**Additional info**
- OS: Windows 11 Pro
- CPU architecture or model: amd64
- `multipass version` : `1.17.0-dev.1005+ga67aacdb.win`

ricab:
I could reproduce with `snap install --edge multipass` too.

Ignacio-Trottaa:
Hi! I'm interested in making my first contribution to Multipass.  
Can I take this issue to work on it and try to resolve it? 

Thanks!

tobe2098:
Hi @Ignacio-Trottaa, feel free to contribute! I can assign the issue to you if you cannot. And feel free to ask questions if you feel lost.

Ignacio-Trottaa:
@tobe2098 Thank you so much, I appreciate it! I'm excited to make my first contribution to Multipass. I'll start by setting up my environment to reproduce the issue. It's great to know I can reach out with questions, and I'll be sure to do so if I get lost.

Tanish-Tandon:
Hi, I’d like to work on this issue.
Could you please assign it to me?
Thanks!

tobe2098:
Hi @Tanish-Tandon, @Ignacio-Trottaa has started to work on this issue already. Since he is still working on it, would you mind searching for another issue? Thank you :)

Tanish-Tandon:
No worries, thanks for letting me know!
I’ll look for another good first issue.

Ignacio-Trottaa:
@tobe2098 Hi, i have a problem: When I read the README and follow the steps to compile and run the project, my session close. 
I am using Ubuntu 24.04 LTS.
idk what is the problem

tobe2098:
Hi @Ignacio-Trottaa, I am sorry to hear that. What do you mean by that? Are you getting logged out or is your computer crashing during the build? Would you be able to get console output or tell me during which step does it happen? 
[...]

PR #4514, opened 2025-11-18
Fix GUI default onAppClose behavior on first run

JohnGahagan:
### **Description:**
This change ensures that the GUI provides non-null default values for (onAppClose) on first run.

Previously, the GuiSettingNotifier could return null for these settings on the first run, causing the GUI to close running instances automatically, even though the UI displayed "Ask" as the default.

With this fix:

- GuiSettingNotifier now returns explicit defaults for first-run scenarios.
- The runtime behavior now matches the defaults displayed in the GUI, improving consistency and preventing unexpected 
closures.

Impact:

- Resolves a bug where the GUI would close running instances immediately on first launch.
- Ensures first-run experience is aligned with UI expectations.

Resolves: #4506

### DIFF
--- a/src/client/gui/lib/providers.dart
+++ b/src/client/gui/lib/providers.dart
@@ -360,7 +360,13 @@ class GuiSettingNotifier extends Notifier<String?> {
   @override
   String? build() {
     final sharedPreferences = ref.read(sharedPreferencesProvider);
-    return sharedPreferences.getString(arg);
+// Define defaults for specific keys
+    final defaultValues = {
+      onAppCloseKey: 'ask',
+    };
+
+    // Return the stored value, or the default value, or null
+    return sharedPreferences.getString(arg) ?? defaultValues[arg];
   }
 
   void set(String value) {


tobe2098:
Hi @JohnGahagan, I just noticed your PR. You should post a message in the issue so we can see you are working on this. For the CI checks you will need to sign the CLA and format the commits according to our CONTRIBUTING guide: https://github.com/canonical/multipass/blob/main/CONTRIBUTING.md

JohnGahagan:
Hi @tobe2098, I signed the Canonical CLA and updated the commit message to match the CONTRIBUTING linter requirements.

tobe2098:
Thank you @JohnGahagan for your contribution! It seems like you will still have to add trailing newlines to a couple of files for the linter. The CLA checker seems to not be working right now, but I will re-run the workflow on Monday. :+1:

JohnGahagan:
@tobe2098 I removed the trailing whitespaces. Seems like I am still being rejected by the CLA, however.

tobe2098:
@JohnGahagan , thank you. The CLA was a service failure. The message from the new commit should also follow the format, but the PR is otherwise complete.
Apparently the dart code also needs additional formatting as well.
Almost there!

JohnGahagan:
@tobe2098 Should be good to go now!

JohnGahagan:
@tobe2098 Looks like all the checks have passed! If anything, else is needed to get this merged, please let me know!

src/client/gui/lib/providers.dart
```diff
@@ -360,6 +360,10 @@ class GuiSettingNotifier extends Notifier<String?> {
   @override
   String? build() {
     final sharedPreferences = ref.read(sharedPreferencesProvider);
+    // Default values for the first run
+    if (arg == onAppCloseKey) {
+      return sharedPreferences.getString(arg) ?? 'ask';
+    }
     return sharedPreferences.getString(arg);
```
tobe2098:
```suggestion
// Define defaults for specific keys
  final defaultValues = {
    onAppCloseKey: 'ask',
  };

  // Return the stored value, or the default value, or null
  return sharedPreferences.getString(arg) ?? defaultValues[arg];
```

tobe2098 (changes requested):
Your solution is good and works well, but it repeats the return statement. Since we are introducing "default values', what do you think about this suggestion?

JohnGahagan:
@tobe2098 Your solution is much better, mine just checked for that specific case while yours generalize this issue so if they are more issues like this in other features it is much easier to apply. The way I did it, to scale it we would have to keep adding if conditions.

tobe2098:
Thank you @JohnGahagan , but the last commit is not properly formatted. Also, do you think you could reduce the number of commits? Our commit history is long, and your fix could be reduced to one commit without losing any information. I would recommend to `rebase -i HEAD~6` and change the commits below the first one to `squash`. What do you think?
Usually when correcting previous changes you can either `commit --ammend`, adding your current changes your last commit or `rebase` and then change a commit to `edit`.

JohnGahagan:
Hi @tobe2098 done (I think I did it correctly).

I rebased and squashed the branch down to a single commit, updated the commit message to match the CONTRIBUTING format, and force-pushed the cleaned history to the PR branch. The previous “combination/merge” commits should be gone now.

If you’re able, could you re-run the checks / take another look? Thanks again!

tobe2098:
Thank you @JohnGahagan , it is almost perfect now. The verify commits indicates you need to wrap the body at 72 chars. If you experience issues with commit messages I recommend you copy `git-hooks/commit-msg.py` to `.git/hooks/commit-msg` to have the checks run when you commit.
Apparently there is also some linting errors from dart. You can follow the indications from the workflow itself to resolve this.

JohnGahagan:
@tobe2098 One more run? 🤞

PR #4515, opened 2025-11-19
[utils] Add Boost.JSON utilities

jimporter:
These are some useful utilities for the upcoming Boost.JSON migration. I've pulled them out into a separate PR so that it's (hopefully?) easier to review. To see the full context and how all these utilities get used, you can look at the full [`boost-json` branch](https://github.com/canonical/multipass/compare/main...boost-json).

### DIFF
--- a/include/multipass/json_utils.h
+++ b/include/multipass/json_utils.h
@@ -29,6 +29,7 @@
 #include <boost/json.hpp>
 
 #include <optional>
+#include <ostream>
 #include <string>
 #include <vector>
 
@@ -123,6 +124,17 @@ T tag_invoke(const boost::json::value_to_tag<T>&,
     }
     return result;
 }
+
+struct PrettyPrintOptions
+{
+    int indent = 4;
+    bool trailing_newline = true;
+};
+
+void pretty_print(std::ostream& os,
+                  const boost::json::value& value,
+                  const PrettyPrintOptions& opts = {});
+std::string pretty_print(const boost::json::value& value, const PrettyPrintOptions& opts = {});
 
 // Temporary conversion functions to migrate between Qt and Boost JSON values.
 boost::json::value qjson_to_boost_json(const QJsonValue& value);
--- a/src/utils/json_utils.cpp
+++ b/src/utils/json_utils.cpp
@@ -25,10 +25,59 @@
 #include <QJsonArray>
 #include <QJsonDocument>
 
+#include <sstream>
+#include <stack>
 #include <stdexcept>
+#include <variant>
 
 namespace mp = multipass;
 namespace mpu = multipass::utils;
+
+namespace
+{
+using JsonObjectIter = std::counted_iterator<boost::json::object::const_iterator>;
+using JsonArrayIter = std::counted_iterator<boost::json::array::const_iterator>;
+using JsonIter = std::variant<JsonObjectIter, JsonArrayIter>;
+
+JsonObjectIter counted_iter(const boost::json::object& object)
+{
+    return {object.begin(), static_cast<std::ptrdiff_t>(object.size())};
+}
+
+JsonArrayIter counted_iter(const boost::json::array& array)
+{
+    return {array.begin(), static_cast<std::ptrdiff_t>(array.size())};
+}
+
+void indent(std::ostream& os, std::size_t width)
+{
+    for (std::size_t i = 0; i != width; i++)
+        os.put(' ');
+}
+
+template <typename T>
+void maybe_put_comma(std::ostream& os, const std::counted_iterator<T>& iter)
+{
+    // Add a comma if there are any more elements after the current one.
+    if (iter.count())
+        os.put(',');
+}
+
+void maybe_put_comma(std::ostream& os, const JsonIter& iter)
+{
+    std::visit([&os](auto&& arg) { maybe_put_comma(os, arg); }, iter);
+}
+
+void pretty_print_scalar(std::ostream& os, const boost::json::value& value)
+{
+    assert(!value.is_object() && !value.is_array());
+    // Boost.JSON always prints doubles in scientific notation, but that's not pretty!
+    if (value.kind() == boost::json::kind::double_)
+        fmt::print(os, "{}", value.as_double());
+    else
+        os << serialize(value);
+}
+} // namespace
 
 mp::JsonUtils::JsonUtils(const Singleton<JsonUtils>::PrivatePass& pass) noexcept
     : Singleton<JsonUtils>{pass}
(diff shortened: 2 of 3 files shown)


src/utils/json_utils.cpp
```diff
@@ -131,3 +134,219 @@ std::optional<std::vector<mp::NetworkInterface>> mp::JsonUtils::read_extra_inter
...
+    auto doc = QJsonDocument::fromJson(
+        QByteArray{json_data.data(), static_cast<qsizetype>(json_data.size())});
+    if (doc.isArray())
+        return doc.array();
+    else if (doc.isObject())
+        return doc.object();
+
+    assert(false && "unsupported type");
+    std::abort();
+}
```
jimporter, line 337: Since these are temporary and will go away once all the Qt JSON code is removed, I didn't write tests for them. Once the code starts using these functions, existing tests will cover them.

tests/test_json_utils.cpp
```diff
@@ -234,9 +82,192 @@ TEST_F(TestJsonUtils, throwsOnWrongMac)
...
+    std::string expected = GetParam();
+    boost::json::value json = boost::json::parse(expected);
+    EXPECT_EQ(mp::pretty_print(json, {.trailing_newline = false}), expected);
+}
+
+TEST_P(JsonPrettyPrintTest, prettyPrintsCustomIndentCorrectly)
+{
+    std::string input = GetParam();
+    boost::json::value json = boost::json::parse(input);
+    // Replace all runs of 4 spaces with 2 spaces. This isn't assumes there are no runs of 4 spaces
```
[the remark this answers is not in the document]
jimporter: Resolved. I also added a `NOTE` marker to the comment so that this stands out a bit more when reading it.

include/multipass/json_utils.h
```diff
@@ -53,4 +57,117 @@ class JsonUtils : public Singleton<JsonUtils>
...
+}
+
+template <typename T, typename Key, typename U = T>
+T lookup_or(const boost::json::value& json, Key&& key, U&& fallback)
+{
+    if (auto elem = detail::if_contains(json, std::forward<Key>(key)))
+        return value_to<T>(*elem);
+    else
+        return std::forward<U>(fallback);
+}
```
tobe2098: Two nitpicks: 
1. Do we need implicit conversion from U to T? If not we could have just typename T and enforce the type, WDYT?
2. Seeing as you are setting the default Context typename to void, these two templates could be a single one with an `if constexpr (std::is_same_v<Context, void>)` with the argument being `const Context* ctx`. It would require the context being a pointer though. Do you think it is better?

include/multipass/json_utils.h
```diff
@@ -53,4 +57,117 @@ class JsonUtils : public Singleton<JsonUtils>
...
+        elem.as_object().erase(cfg.key_field);
+        result.emplace(key, value_to<typename T::mapped_type>(elem));
+    }
+    return result;
+}
+
+struct PrettyPrintOptions
+{
+    int indent = 4;
+    bool trailing_newline = true;
```
tobe2098, line 131: Small nitpick/question: Is there a reason to prefer assignment initialization over bracket{} initialization?

include/multipass/json_utils.h
```diff
@@ -53,4 +57,117 @@ class JsonUtils : public Singleton<JsonUtils>
...
+void tag_invoke(const boost::json::value_from_tag&,
+                boost::json::value& json,
+                const T& mapping,
+                const MapAsJsonArray& cfg)
+{
+    auto& arr = json.emplace_array();
+    for (const auto& [key, value] : mapping)
+    {
+        auto elem = boost::json::value_from(value);
+        elem.as_object().emplace(cfg.key_field, boost::json::value_from(key));
```
tobe2098, line 106: Just a question: This assumes that elem is an object. This template would match a std::map<std::string,int> and throw (tested), which could happen recursively while calling a more nested type that worked properly. Are there some guarantees that the conversion will not be used in such types? If not I would recommend restricting the template with concepts further or alter the template to properly adapt if the conversion from elem does not produce a json object. WDYT?

include/multipass/json_utils.h
```diff
@@ -53,4 +57,117 @@ class JsonUtils : public Singleton<JsonUtils>
...
+    T result;
+    for (const auto& i : json.as_array())
+    {
+        boost::json::value elem = i;
+        auto key = value_to<typename T::key_type>(elem.at(cfg.key_field));
+        elem.as_object().erase(cfg.key_field);
+        result.emplace(key, value_to<typename T::mapped_type>(elem));
+    }
+    return result;
+}
```
tobe2098, line 126: This tag_invoke similarly assumes the inner structure of the json. In this case restriction does not work so we could make the code more resilient or add testing, but the generic mapping approach may end up being less convenient than doing it type-wise. Although the throw can be left to the caller. WDYT?

include/multipass/json_utils.h
tobe2098: The concept restriction would go along the lines of requiring T::mapped_type to produce an object when converted.

src/utils/json_utils.cpp
```diff
@@ -25,11 +25,60 @@
...
+#include <variant>
 
 namespace mp = multipass;
 namespace mpu = multipass::utils;
 
+namespace
+{
+using object_iter_t = std::counted_iterator<boost::json::object::const_iterator>;
+using array_iter_t = std::counted_iterator<boost::json::array::const_iterator>;
+using some_iter_t = std::variant<object_iter_t, array_iter_t>;
```
tobe2098: Just a nitpick, but maybe value_iter_t/bvalue_iter_t/boost_iter_t is more appropiate?

src/utils/json_utils.cpp
```diff
@@ -131,3 +180,176 @@ std::optional<std::vector<mp::NetworkInterface>> mp::JsonUtils::read_extra_inter
...
+// processing that new container.
+void mp::pretty_print(std::ostream& os,
+                      const boost::json::value& value,
+                      const PrettyPrintOptions& opts)
+{
+    std::stack<some_iter_t> stack;
+
+    // Populate our stack with the top-level sequence (or just return immediately if `value` is a
+    // scalar).
+    if (auto&& obj = value.if_object())
```
tobe2098, line 200: Nice touch with the universal reference :+1:

tobe2098 (changes requested):
It looks very good, I just have some questions relating to some aspects of the templates. I really like the pretty_printing implementation!

include/multipass/json_utils.h
jimporter: 1. The `typename U = T` bit is an attempt to do two things:
    1. Keep `T` out of the argument list so that the compiler can't infer the value of `T`. I wanted to make sure that anyone using this function explicitly names the return type for consistency.
    2. Allow passing an initializer list like `{}` as the value for `fallback`. To do that, the compiler needs to know the type, so it defaults to `T` in that case.

    (It would be nice if I could write some tests that verify that things like `lookup_or(json, "key", 123)` fail to compile, but I don't think we have a clean way of testing expected compilation failures in Multipass.)
2. We could use `boost::json::detail::no_context` for this, though it's kind of bad form to use something from another project's `detail` namespace. Or we could change `Context` to be a variadic template type, which is probably the most concise way to do this. (I thought about doing the latter, but the current way seemed easier for a reader to understand the API. I don't mind changing it if you think it'd be better to remove the duplication, though.)

include/multipass/json_utils.h
jimporter: I find assignment initialization a bit easier to read, but this was mainly just that I see assignment-init more often in the code base. I don't mind changing this if we have a preference one way or the other.

include/multipass/json_utils.h
jimporter: This is a tricky problem. Technically, the `boost::json::value_from` call can return absolutely any kind of value. Even for a single type, it could sometimes return an object, and other times return a string. I could add a more-detailed exception to catch this at runtime, but that's probably the best we can do.

As for this happening recursively, I don't think that's something we need to worry about. Since the `boost::json::value_from` call doesn't pass in the `MapAsJsonArray` context, they'll get serialized using the default method. (Maybe it would be useful to be able to pass a context in for the inner values to use, but I didn't want to add features we don't need yet.)

I added a unit test to verify that `MapAsJsonArray` doesn't recursively turn maps into arrays though.

include/multipass/json_utils.h
jimporter: I'm not sure what you mean by "type-wise" here. Do you just mean implementing behavior like this for specific classes in their `tag_invoke` functions?

If it helps, here are a couple examples of how we'll use this `MapAsJsonArray` context:
* https://github.com/canonical/multipass/commit/26096b5c79c3f3f56e9ce88d42200c7abc071988#diff-b90a3fdc7a07efca765a66d371e797787978d49202e82620444c98ae7bd4c750R44
* https://github.com/canonical/multipass/commit/5b5917bfcde589fc59bc1c59e4b2d66cc17566d1#diff-1e85b8a2cf572ddde7dfdf15f49e4110863ffd2b28088e1f82aacf7874c5f417R113

src/utils/json_utils.cpp
jimporter: Good call. How about `JsonObjectIter`, `JsonArrayIter` and `JsonIter`? (Using CamelCase now to comply with our style guidelines.)

include/multipass/json_utils.h
tobe2098: What about doing std::optional<Context> timeout = std::nullopt, and then if constexpr (ctx.has_value())? I think it captures the fact that it is an optional argument (no pun intended) better. WDYT?

include/multipass/json_utils.h
tobe2098: If assignment is widespread in the codebase we can leave it, but we should switch. Having no implicit casts is really nice.

include/multipass/json_utils.h
tobe2098: I missed the `Map` context. It's good that it stops, if we do not need it recursively it's best to leave it as is.

include/multipass/json_utils.h
tobe2098: I see, my concern was that it is assumed that elem contains an object and that it looks like {key_name: key, value: value}. Since it is non-recursive, I don't think more checks are necessary. It would be nice to be able to keep track of which jsons are mapped with MapAsJsonArray context, so we can have some railings to not allow the conversion if it is guaranteed that it will fail. How does that sound?

src/utils/json_utils.cpp
tobe2098: Sounds good :+1:

xmkg (dismissed):
LGTM, thanks Jim!

include/multipass/json_utils.h
jimporter: @tobe2098 I tried using `std::optional`, but it didn't work since we'd have to convert `SomeContext` to `std::optional<T>`, and C++ isn't smart enough to deduce `T` in this case.

Instead, I changed this to use a template parameter pack, and then added a requirement that the pack has 0 or 1 elements. (This last part isn't *really* necessary: we pass the pack to `boost::json::value_to`, and this only takes 1 or 2 arguments, including the JSON value. Still, this should hopefully make error messages nicer and also make the intent clearer to future programmers.)

What do you think about this implementation?

include/multipass/json_utils.h
jimporter: I'm not totally sure what you mean here. I can try to add some friendlier error messages here, though the calls to `as_array` and `as_object` will already throw exceptions if we get a JSON value of the wrong kind.

include/multipass/json_utils.h
tobe2098: Oh, I see. It looks perfect, nice idea. Thank you for the effort!

PD: I think it is possible to the "single optional argument" idea the following way:
```
struct DefaultCallMarker {};
template <typename T = DefaultCallMarker>
int identify(std::optional<T> value = {}) {
  if constexpr (std::is_same_v<T, DefaultCallMarker>) {
    return -1;
  } else {
    if (value.has_value())
      return int(value.value());
    else
      return -2;
  }
}

int main() {
  std::cout << identify() << ' ' << identify(std::optional(4)) << ' '
            << identify();
  return 0;
}
```
But your approach is way more ergonomic and elegant :)

include/multipass/json_utils.h
tobe2098: Sorry, I should have been clearer. It seems as mappings mapped from a JSON or vice-versa (although the JSON files are excluded from this because we do not keep track of their state) are intended to be mapped and unmapped by using the same context, as far as I understand:
https://github.com/canonical/multipass/pull/4515/files#diff-1574075cf759d2d73af33b9821c7e09c84426f574d42cc2c428c3f5c8a36cfe1R177-R183

The concern was that on usage of this particular conversion, the mapping context should be attached to the data structure we are using. So that mapping back and forth can be done easily without mistakes, since the mapping is accessible at the same places the data is accessible. The concern is not in the implementation but in the usage. In terms of implementation I think it is perfect. Good job!

tobe2098 (approved):
LGTM Jim, great work! Looking forward to the absence of QT :)

include/multipass/json_utils.h
jimporter: Ah, got it. That's actually a very interesting possibility. For custom types, I think I have an idea of how we'd do that: we could add a type trait that takes our type as the parameter, and gives back the key name that Boost.JSON should use. Then we just need an overload of `tag_invoke` that requires `T` to specialize that trait. The only concerns I have are that this gets into some relatively complex template programming, and that the `T` in this case is a `std::unordered_map` of some kind. Is it bad manners to override the behavior for a template instantiation of a standard class template? I'm not sure...

@tobe2098 What are your thoughts about trying to do this now vs as a followup? I'll probably do a prototype of it either way, if only as an educational exercise, so I really don't have any problems with waiting to merge this if you prefer. Otherwise, we can always merge improvements later.

tobe2098: We can keep discussing it over mattermost, but for now we should merge this. WDYT?

tobe2098: We can keep discussing it over mattermost, but for now we should merge this. WDYT?

tobe2098:
This may require a rebase to pass the CI if we already merged the fix for clang-tidy to main. Auto-merge does not work otherwise I think

PR #4516, opened 2025-11-19
[utils] Move write_json to write_transactionally

jimporter:
99% of write_json's implementation is about file I/O, so this makes more sense in FileOps. It will also make it easier for Boost.JSON to use this function eventually.

### DIFF
--- a/src/utils/file_ops.cpp
+++ b/src/utils/file_ops.cpp
@@ -33,7 +33,36 @@ namespace fs = mp::fs;
 
 namespace
 {
-constexpr static auto kLogCategory = "FileOps";
+constexpr static auto log_category = "FileOps";
+
+class BackoffTimer
+{
+public:
+    using Duration = std::chrono::milliseconds;
+    BackoffTimer(Duration factor, Duration max_duration, Duration max_jitter)
+        : jitter(0, max_jitter.count()), factor(factor), max_duration(max_duration)
+    {
+    }
+
+    // Delay with jitter + backoff. Typical series produced by this formula would be:
+    // [2, 14,23,60,90,168,216,213,218,218]
+    // [14,20,30,40,98,174,221,208,206,214]
+    Duration operator()(int attempt)
+    {
+        return std::min(max_duration, factor * (1 << (attempt - 1))) + Duration(jitter(rng));
+    }
+
+private:
+    // Share the random engine with all `BackoffTimer` instances on this thread.
+    thread_local static std::mt19937 rng;
+
+    std::uniform_int_distribution<Duration::rep> jitter;
+    Duration factor;
+    Duration max_duration;
+};
+
+thread_local std::mt19937 BackoffTimer::rng(std::random_device{}());
+
 } // namespace
 
 mp::NamedFd::NamedFd(const fs::path& path, int fd) : path{path}, fd{fd}
@@ -53,33 +82,33 @@ mp::FileOps::FileOps(const Singleton<FileOps>::PrivatePass& pass) noexcept
 
 void mp::FileOps::write_transactionally(const QString& file_name, const QByteArrayView& data) const
 {
-    constexpr static auto kStaleLockTime = std::chrono::seconds{10};
-    constexpr static auto kLockAcquireTimeout = std::chrono::seconds{10};
+    using namespace std::literals::chrono_literals;
+    constexpr static auto stale_lock_time = 10s;
+    constexpr static auto lock_acquire_timeout = 10s;
+    constexpr static auto max_attempts = 10;
+
+    BackoffTimer backoff{/*factor=*/10ms, /*max_duration=*/200ms, /*max_jitter=*/25ms};
     const QFileInfo fi{file_name};
 
-    const auto dir = fi.absoluteDir();
-    if (!mkpath(fi.absoluteDir(), "."))
+    if (const auto dir = fi.absoluteDir(); !mkpath(dir, "."))
         throw std::runtime_error(fmt::format("Could not create path '{}'", dir.absolutePath()));
 
     // Interprocess lock file to ensure that we can synchronize the request from
     // both the daemon and the client.
     QLockFile lock(fi.filePath() + ".lock");
 
-    // Make the lock file stale after a while to avoid deadlocking
-    // on process crashes, etc.
-    setStaleLockTime(lock, kStaleLockTime);
+    // Make the lock file stale after a while to avoid deadlocking on process crashes, etc.
+    setStaleLockTime(lock, stale_lock_time);
 
     // Acquire lock file before attempting to write.
-    if (!tryLock(lock, kLockAcquireTimeout))
-    { // wait up to 10s
+    if (!tryLock(lock, lock_acquire_timeout))
+    {
         throw std::runtime_error(fmt::format("Could not acquire lock for '{}'", file_name));
     }
 
-    constexpr static auto max_attempts = 10;
-
-    // The retry logic is here because the destination file might be locked for any reason
-    // (e.g. OS background indexing) so we will retry writing it until it's successful
-    // or the attempts are exhausted.
+    // The retry logic is here because the destination file might be locked for any reason (e.g. OS
+    // background indexing) so we will retry writing it until it's successful or the attempts are
+    // exhausted.
     for (auto attempt = 1; attempt <= max_attempts; attempt++)
     {
         QSaveFile db_file{file_name};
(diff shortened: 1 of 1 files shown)


tests/test_base_snapshot.cpp
ricab, line 1: This line should probably be simplified now: https://github.com/canonical/multipass/blob/9b7fd2ad110820f03579f0056c609ffb31dcf9db/tests/test_base_snapshot.cpp#L425

src/utils/file_ops.cpp
```diff
@@ -38,6 +50,89 @@ mp::FileOps::FileOps(const Singleton<FileOps>::PrivatePass& pass) noexcept
 {
 }
 
+void mp::FileOps::write_transactionally(const QString& file_name, const QByteArray& data) const
```
ricab: Since you have this fresh in your mind, it might be a good change to split this function into smaller chunks and sticking to [naming conventions](https://github.com/canonical/multipass/blob/e0019a82401bd17b9330cb678a9bdd9690ef5914/CONTRIBUTING.md?plain=1#L263) :wink:

src/client/common/alias_dict.cpp
```diff
@@ -407,7 +407,9 @@ void mp::AliasDict::load_dict()
 void mp::AliasDict::save_dict()
 {
     sanitize_contexts();
-    MP_JSONUTILS.write_json(to_json(), QString::fromStdString(aliases_file));
+
+    MP_FILEOPS.write_transactionally(QString::fromStdString(aliases_file),
+                                     QJsonDocument{to_json()}.toJson());
```
ricab, line 411: If we weren't replacing Qt for Json, I might suggest a util to go from `QJsonObject` to a `QByteArray`, especially given the poor method naming on Qt's part for this. Since it will get replaced very soon, I won't make that suggestion :grin:

ricab (commented):
Makes sense to me @jimporter. It clarifies the code a bunch, separates concerns better, and makes it more reusable. Only a small detail and an optional solicitation, if you're in feeling Christmas approaching 🎅 

[secondary]

tests/test_base_snapshot.cpp
jimporter: Good catch! (Though note that this will get simplified even further in an [upcoming commit](https://github.com/canonical/multipass/commit/5b5917bfcde589fc59bc1c59e4b2d66cc17566d1#diff-1dda53f0bcd7bfaeedddf0c9d5b4aadad0fbdafb303ee8d59dca420b88aaf9eaR393).)

src/client/common/alias_dict.cpp
jimporter: Yep. :) If we were sticking with Qt JSON, another way to simplify this would have been for the `to_json()` member function to return a `QJsonDocument` instead. Eventually, this will become:

```c++
    auto json = boost::json::value_from(*this);
    MP_FILEOPS.write_transactionally(QString::fromStdString(aliases_file),
                                     pretty_print(json));
```

That's longer still, but we could always add `write_json` back as a simple wrapper for Boost.JSON when we're done. (Since we only have to write files like this in a few places, I don't mind the extra verbosity. It makes it explicit what's actually happening.)

src/utils/file_ops.cpp
jimporter: Ok, done. I pulled the backoff-timing logic into a separate class, which simplifies that code a bit (and keeps us from repeatedly instantiating new `std::random_device`s every call). I also fixed a small bug in the implementation where we'd call `sleep_for` after all our attempts were exhausted. We can just throw an exception immediately in that case.

src/utils/file_ops.cpp
```diff
@@ -16,13 +16,55 @@
...
+
 #include <fcntl.h>
 
 namespace mp = multipass;
+namespace mpl = multipass::logging;
 namespace fs = mp::fs;
 
+namespace
+{
+constexpr static auto kLogCategory = "FileOps";
```
ricab: This one slipped through.

src/utils/file_ops.cpp
```diff
@@ -16,13 +16,55 @@
...
+
+class BackoffTimer
+{
+public:
+    using duration = std::chrono::milliseconds;
```
ricab: Here on the other hand, we should start with a capital: https://github.com/canonical/multipass/blob/89beb65c374083397bb3f118dc1f09342c5f0edb/CONTRIBUTING.md?plain=1#L263

src/utils/file_ops.cpp
```diff
@@ -16,13 +16,55 @@
...
+    BackoffTimer(duration factor, duration max_duration, duration max_jitter)
+        : rng(std::random_device{}()),
+          jitter(0, max_jitter.count()),
+          factor(factor),
+          max_duration(max_duration)
+    {
+    }
+
+    // Delay with jitter + backoff. A typical series produced by this formula would look like as
+    // follows:
```
ricab:
```suggestion
    // Delay with jitter + backoff. Typical series produced by this formula would be:
```

src/utils/file_ops.cpp
```diff
@@ -16,13 +16,55 @@
...
+    {
+        return std::min(max_duration, factor * (1 << attempt++) + duration(jitter(rng)));
+    }
+
+private:
+    std::mt19937 rng;
+    std::uniform_int_distribution<duration::rep> jitter;
+    duration factor;
+    duration max_duration;
+    int attempt = 0;
```
ricab: There are now two variables keeping track of iterations. Shouldn't this be received in `operator()` instead?

src/utils/file_ops.cpp
```diff
@@ -38,6 +80,79 @@ mp::FileOps::FileOps(const Singleton<FileOps>::PrivatePass& pass) noexcept
...
+{
+    using namespace std::literals::chrono_literals;
+    constexpr static auto stale_lock_time = 10s;
+    constexpr static auto lock_acquire_timeout = 10s;
+    constexpr static auto max_attempts = 10;
+
+    BackoffTimer backoff{/*factor=*/10ms, /*max_duration=*/200ms, /*max_jitter=*/25ms};
+    const QFileInfo fi{file_name};
+
+    if (const auto dir = fi.absoluteDir(); !mkpath(dir, "."))
```
ricab, line 93: Good catch 👍

src/utils/file_ops.cpp
```diff
@@ -38,6 +80,79 @@ mp::FileOps::FileOps(const Singleton<FileOps>::PrivatePass& pass) noexcept
...
+                  "Failed to write `{}` in attempt #{} (reason: {}), will retry after {} ms delay.",
+                  file_name,
+                  attempt,
+                  db_file.errorString(),
+                  delay);
+
+        std::this_thread::sleep_for(delay);
+    }
+
+    assert(false && "We should never get here");
```
ricab, line 153: 👍

ricab (commented):
Yep, well done. Only a few remaining details.

src/utils/file_ops.cpp
jimporter: Oops! Fixed.

src/utils/file_ops.cpp
jimporter: That makes sense, and then `operator()` mutates less internal state (though naturally it still mutates the `rng`).

src/utils/file_ops.cpp
jimporter: Fixed. (I'm sure I'll forget about that again at some point since I'm *very* used to camel case being used for template identifiers.)

src/utils/file_ops.cpp
jimporter: Done.

src/utils/file_ops.cpp
ricab: Yeah, we should teach clang-format to catch that...

ricab (dismissed):
Yep, I am happy with this if CI is.

jimporter:
@ricab I updated this a bit, and added another commit at the beginning. That one makes all our `FileOps` wrappers for `QFile` use the original type for the file argument, which makes it easier to use those functions as needed (specifically, this helps because I changed `write_transactionally` to take a `QByteArrayView` instead of a `QByteArray`, and so I used a different overload of `FileOps::write`).

src/utils/file_ops.cpp
```diff
@@ -16,14 +16,52 @@
...
+namespace
+{
+constexpr static auto log_category = "FileOps";
+
+class BackoffTimer
+{
+public:
+    using Duration = std::chrono::milliseconds;
+    BackoffTimer(Duration factor, Duration max_duration, Duration max_jitter)
+        : rng(std::random_device{}()),
```
tobe2098: Is there a reason we need a truly random number generator when available in hardware? Each generation is a syscall in those devices. Would a pseudo rng like std::mt19937_64 be sufficient for this?

jimporter: This is just to set the seed for our PRNG. How about making the PRNG `thread_local` so we don't have to recreate it as often?

tobe2098: Oh, sorry, I missed the second set of parenthesis. I think one extra syscall per transactional write is not a large overhead. Making it `thread_local` would be a nice economic approach, I will leave it up to you :+1:

PR #4521, opened 2025-11-20
gui: enforce default close behavior when unset (Fixes #4506)

### LINKED ISSUE #4506: [GUI] Default behavior on close not reflected in settings

vardhan30016:
This PR fixes Issue #4506, where the default close behavior was not reflected
in the GUI when the setting `onAppCloseKey` was not set.

### Root cause
`closeJob` was read from settings but not given a default when missing.

### Fix
- When `closeJob` is null or empty, enforce the default action `"ask"`.
- Immediately update the provider state so UI and behavior match.

### Result
Multipass GUI now behaves correctly when first installed or when the user has
no stored preference.

Fixes #4506

— @vardhan30016

### DIFF
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -168,7 +168,16 @@ class _AppState extends ConsumerState<App> with WindowListener {
     final daemonAvailable = ref.read(daemonAvailableProvider);
     final vmsRunning =
         ref.read(vmStatusesProvider).values.contains(Status.RUNNING);
-    final closeJob = ref.read(guiSettingProvider(onAppCloseKey));
+    var closeJob = ref.read(guiSettingProvider(onAppCloseKey));
+
+// Fix for Issue #4506: enforce default when setting is empty or not set
+if (closeJob == null || closeJob.toString().trim().isEmpty) {
+  const defaultCloseAction = 'ask';
+  // Update provider so future reads are correct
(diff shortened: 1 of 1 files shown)


vardhannaidux:
Thanks for the update! Yes, this PR is still in progress — currently waiting for a maintainer review and approval.
All checks have passed except the ones that require maintainer approval to run, which is expected for contributors without write access.

Once a maintainer reviews and approves the workflow, the remaining checks will run and the PR can move forward toward merging.

Let me know if anything else is needed from my side! 🙌

tobe2098:
Sorry @vardhan30016 , there was already an existing PR when you started this one and we gave preference to that one. I encourage you to find other issues to resolve in the repository, and we will give you any guidance when needed.
Thank you for your effort

PR #4524, opened 2025-11-21
[cleanup] Trim header includes (generic+linux)

tobe2098:
This PR removes some unused headers and namespace aliases as part of cleanup.

The added items in .gitignore is for clangd users.

MULTI-2367

### DIFF
--- a/tests/stub_vm_image_vault.h
+++ b/tests/stub_vm_image_vault.h
@@ -23,8 +23,6 @@
 #include <multipass/vm_image.h>
 #include <multipass/vm_image_vault.h>
 
-namespace mp = multipass;
-
 namespace multipass
 {
 namespace test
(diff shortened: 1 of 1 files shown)


sharder996 (dismissed):

tests/stub_vm_image_vault.h
```diff
@@ -20,8 +20,11 @@
 #include "temp_file.h"
 
 #include "mock_image_host.h"
+#include <multipass/vm_image.h>
 #include <multipass/vm_image_vault.h>
 
+namespace mp = multipass;
```
jimporter: I think we should put this in the corresponding `.cpp` file for consistency with the rest of our code, where we generally avoid namespace aliases in headers.

.gitignore
```diff
@@ -1,6 +1,8 @@
 .idea
 .vs
 .vscode
+.clangd
+.cache
```
jimporter, line 5: Since we're already including a lot of tool-specific ignores, I'm ok with this. Though I did want to mention that I usually just put things like this in a [global gitignore file](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration.html#_core_excludesfile). That way, it's not necessary to add these to each project.

jimporter (approved):
This all looks good to me, aside from one small nitpick below.

tests/stub_vm_image_vault.h
tobe2098: Good catch!

tobe2098: Actually, if you see lines 64-65, the namespace alias is required there (it must have been in another header that I removed), but I think it is best to keep to that and undo the alias there.

PR #4526, opened 2025-11-24
Add codecov configuration file

xmkg:
MULTI-2380

### DIFF
--- a/src/utils/file_ops.cpp
+++ b/src/utils/file_ops.cpp
@@ -23,6 +23,8 @@
 namespace mp = multipass;
 namespace fs = mp::fs;
 
+// LCOV_EXCL_START
+
 mp::NamedFd::NamedFd(const fs::path& path, int fd) : path{path}, fd{fd}
 {
 }
@@ -312,3 +314,5 @@ fs::perms mp::FileOps::get_permissions(const fs::path& file) const
 {
     return fs::status(file).permissions();
 }
+
+// LCOV_EXCL_STOP


.github/.codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  round: up
```
xmkg, line 2: This is to avoid superfluous coverage regressions like `0.007 pct down`

ricab (commented):
Thanks @xmkg, it will be nice to be able to ignore things. A couple of questions:

- should we try to keep TICs in sync with respect to ignored files?
- shouldn't changes to this file cause a rebuild of the source (in dynamic CI)? The coverage run would be affected.

xmkg:
> Thanks @xmkg, it will be nice to be able to ignore things. A couple of questions:
> 
> * should we try to keep TICs in sync with respect to ignored files?

I think it would be better to solve this right at the source. We can tell it directly to "lcov" so all downstream coverage utilities would exclude it. I'll check if "lcov" has a config file. 

> * shouldn't changes to this file cause a rebuild of the source (in dynamic CI)? The coverage run would be affected.

I mean, it does not change how the coverage run happens; it changes how the coverage results are interpreted, so in theory, a rebuild shouldn't be necessary. Implementing that, though, will definitely take longer than just removing an exclusion in dynamic CI filtering. I think we can run the CI from scratch for now and revisit later.

xmkg:
> I think it would be better to solve this right at the source. We can tell it directly to "lcov" so all downstream coverage utilities would exclude it. I'll check if "lcov" has a config file.

@ricab Unfortunately, there's no lcov config file that we can use for this purpose. There is https://linux.die.net/man/5/lcovrc, but the options are mostly cosmetic and HTML-related. I think we can use LCOV_EXCL_START/LCOV_EXCL_STOP though: https://community.codecov.com/t/lcov-excl-start-not-working-in-coverage-reports/2153/7

xmkg:
> ## [Codecov](https://app.codecov.io/gh/canonical/multipass/pull/4526?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) Report
> ✅ All modified and coverable lines are covered by tests. ✅ Project coverage is 87.15%. Comparing base ([`ba79ad9`](https://app.codecov.io/gh/canonical/multipass/commit/ba79ad9025b7c38e93c44ad519324645a5d47793?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)) to head ([`18a069b`](https://app.codecov.io/gh/canonical/multipass/commit/18a069b5d2d0c0318131027b4abf9e127d85b6e8?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)). ⚠️ Report is 23 commits behind head on main.
> 
> Additional details and impacted files
> ```diff
> @@            Coverage Diff             @@
> ##             main    #4526      +/-   ##
> ==========================================
> - Coverage   87.15%   87.15%   -0.00%     
> ==========================================
>   Files         247      246       -1     
>   Lines       14249    14137     -112     
> ==========================================
> - Hits        12418    12320      -98     
> + Misses       1831     1817      -14     
> ```
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/canonical/multipass/pull/4526?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical).
> 
> 🚀 New features to boost your workflow:

@ricab LCOV_EXCL_START/LCOV_EXCL_STOP seem to work fine:

image

Does that approach sound good to you?

PR #4530, opened 2025-11-28
Migrate QT to vcpkg

xmkg:
Move QT to `vcpkg`. A couple of dependencies in Qt's dependency chain (namely libsystemd, b2 and libxcrypt) needs a bunch of libraries to be pre-installed so this PR also takes care of that, too.

Only the necessary parts of Qt are enabled and installed (core, network, concurrent, openssl and dbus) to ensure minimum effect on the build time, and to avoid introducing more Qt stuff to codebase accidentally.

Significant changes:
- We're now linking against Qt statically (default for all vcpkg deps).
- Had to drop the Qt version from 6.10 to 6.9 since it's the latest available.
- Removed MacOS RPATH adjustment script for Qt
- Removed QT network information plugin code from CMake (the plugin seems to be no longer in use)
- Removed instructions regarding manual Qt installation from the README and build instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Please follow the platform-specific build instructions in the files below:
 ### Generic build tips
 
 **Qt version compatibility**
-Multipass is tested with **Qt 6.9.0**. Newer patch versions along the 6.9 track (e.g. 6.9.1) should be fine. Newer minor versions may work, but they may cause compatibility issues.
+Multipass is tested with **Qt 6.9.1**. Newer patch versions along the 6.9 track (e.g. 6.9.2) should be fine. Newer minor versions may work, but they may cause compatibility issues.
 
 You may use your preferred package manager to install Multipass.
 Note that only the official installers are supported.


tobe2098 (changes requested):
Looks good overall, and it is good to have this within vcpkg, but I cannot build this branch, I get a linking error (cannot find -lcap and -llz4, No such file or directory), which means that there are probably additional missing dependencies. I will try to test in a clean Ubuntu to see which ones, and testing in macOS and Windows is in progress.

tobe2098:
`libcap-dev` and `liblz4-dev` seem to be the only libraries that could be missing in some machines, I did not find others. Installing them with `apt install` solved the issue.

The disk memory cost of the static linking is ~10MB in debug, which is double Rust's. Unfortunately, in release this increases memory by 9MB per executable as well. ~~Is the statically linked library containing debuginfo?~~ From `strip -d`, the executables are already empty of debuginfo. Thoughts/opinions/options on the matter? A possibility would be to introduce #4537.

tobe2098:
macOS seems to compile fine, Windows 11, following the guide enters into an infinite loop when running `cmake --build . -j4`, where cmake keeps re-running the configure step.

EDIT: After restarting the VM and deleting the previous build, the current build seems to work. The failing build was also based on a clean build, but with the vcpkg flag from the guide, which seems to cause some sort of conflict.

README.md
```diff
@@ -180,10 +180,7 @@ Please follow the platform-specific build instructions in the files below:
 ### Generic build tips
 
 **Qt version compatibility**
-Multipass is tested with **Qt 6.10.0**. Newer patch versions along the 6.10 track (e.g. 6.10.4) should be fine. Newer minor versions may work, but they may cause compatibility issues.
-
-**ARM64 and cross-compilation**
-If you're using Apple Silicon (arm64) or cross-compiling, ensure that your `PATH` and `CMAKE_PREFIX_PATH` environment variables point to the correct Qt installation for your system architecture.
+Multipass is tested with **Qt 6.9.0**. Newer patch versions along the 6.9 track (e.g. 6.9.1) should be fine. Newer minor versions may work, but they may cause compatibility issues.
```
jimporter: Extremely minor nitpick. It looks like we're using [Qt 6.9.1](https://github.com/microsoft/vcpkg/blob/4334d8b4c8916018600212ab4dd4bbdc343065d1/ports/qt/vcpkg.json#L3).

```suggestion
Multipass is tested with **Qt 6.9.1**. Newer patch versions along the 6.9 track (e.g. 6.9.2) should be fine. Newer minor versions may work, but they may cause compatibility issues.
```

packaging/macos/fixup-qt6-libs-rpath.sh
jimporter, line 1: Nice that we can get rid of all this!

xmkg:
> Thoughts/opinions/options on the matter? A possibility would be to introduce #4537.

I think the magnitude of n times 10 megabytes is not that large a difference compared to what having the debug symbols offers. I'd say it's in the "acceptable" ballpark. AFAIK we don't have to much of a concern around the package or artifact size, unless the size difference is huge.

xmkg:
> Looks good overall, and it is good to have this within vcpkg, but I cannot build this branch, I get a linking error (cannot find -lcap and -llz4, No such file or directory), which means that there are probably additional missing dependencies. I will try to test in a clean Ubuntu to see which ones, and testing in macOS and Windows is in progress.

> EDIT: After restarting the VM and deleting the previous build, the current build seems to work. The failing build was also based on a clean build, but with the vcpkg flag from the guide, which seems to cause some sort of conflict.

I'm willing to address the documenting the dev deps part once it's settled in a separete PR. How does that sound to you?

tobe2098:
It sounds good to me. On the size aspect, I noticed that the real increase in snap size is only ~10MB (snapcraft must be doing some code de-duplication), so I think it is acceptable in order to have it integrated with vcpkg. Great work! I will accept changes after the conflicts are resolved :+1:

xmkg:
> It sounds good to me. On the size aspect, I noticed that the real increase in snap size is only ~10MB (snapcraft must be doing some code de-duplication), so I think it is acceptable in order to have it integrated with vcpkg. Great work! I will accept changes after the conflicts are resolved 👍

Rebased.

tobe2098 (approved):
LGTM!

xmkg:
Pulling the trigger based on 3 approvals (only a doc change + rebase happened since previous +2).

PR #4531, opened 2025-11-28
[test] Add std::abort in default block

tobe2098:
# Description

This PR adds `std::abort()` under the `default` label of a `switch` block. With this it is possible to build using the `-DCMAKE_BUILD_TYPE=Release` flag locally.

## Testing

- Unit tests
- Manual testing steps:

  1. Compile the project with `-DCMAKE_BUILD_TYPE=Release`
  2. The project compiles

## Checklist

## Additional Notes

Additionally, there is one failing test in this type of build (`ClientAlias.aliasCreatesAlias`). Due to lack of debug information I have not been able to pinpoint the failure, but it happens in `parse_args` inside `Client::run`.

### DIFF
--- a/tests/test_common_callbacks.cpp
+++ b/tests/test_common_callbacks.cpp
@@ -76,6 +76,7 @@ struct TestLoggingSpinnerCallbacks : public TestSpinnerCallbacks,
             break;
         default:
             assert(false && "shouldn't be here");
+            std::abort();
         }
     }
 };


ricab (approved):
Ouch, this went unnoticed for a while. Thanks for the fix!

tests/test_common_callbacks.cpp
```diff
@@ -76,6 +76,7 @@ struct TestLoggingSpinnerCallbacks : public TestSpinnerCallbacks,
             break;
         default:
             assert(false && "shouldn't be here");
+            std::abort();
```
xmkg, line 79: Not for this PR but there's an alternative we can use for such cases, which is the "-Wswitch" flag on gcc/clang, or C4062 on MSVC. Wrapping the switch statement with a pragma push/pop should prevent an unhandled switch case altogether:

https://gcc.godbolt.org/z/Thfs965dG

```c++
// clang-format off
#define EXHAUSTIVE_SWITCH_CASE_BEGIN                 \
    _Pragma("GCC diagnostic push")                   \
    _Pragma( "GCC diagnostic error \"-Wswitch\"")

#define EXHAUSTIVE_SWITCH_CASE_END _Pragma("GCC diagnostic pop")
// clang-format on
enum class e_foo : int { bar, baz, qux };

int main(void) {
    e_foo value{e_foo::bar};
    EXHAUSTIVE_SWITCH_CASE_BEGIN
    switch (value) {
        case e_foo::bar:
        case e_foo::baz:
            break;  // qux unhandled, should result in a diagnostic
                    // error
    }
    EXHAUSTIVE_SWITCH_CASE_END
}
```

PR #4537, opened 2025-12-01
Strip symbols in release

tobe2098:
# Description

- What does this PR do? This PR adds symbol stripping of unnecessary symbols from multipass targets. It also adds a cmake option to enable by default this feature. This feature does not exist in Windows because symbols are stored in a separate file.

- Why is this change needed? The whole multipass package requires higher than necessary amount of disk memory due to code duplication across targets. This PR temporarily addresses this issue by achieving disk memory gains through stripping unnecessary symbols in release builds only. The memory gains are 6MB in multipass, 6 MB in multipassd and 5MB in libdart_ffi.so; plus minor gains in other targets.

## Testing

**Done in release builds**

- Unit tests pass (the test failure of ClientAlias.aliasCreatesAlias is baseline in release builds).
- Manual testing steps:

  1. Launch multipassd
  2. Launch a VM with multipass
  3. Test launch (without name for petname), clone, find, get, set, info, mount, umount (classic for sshfs_server), delete, recover, snapshot, shell (install something to test snapshot feature) restore, exec, suspend, start, stop, transfer, wait-ready.  

## Checklist

## Additional considerations
This PR must be thoroughly tested both in macOS and Linux. Although we are only stripping unnecessary symbols it is a considerable change on the targets.


tobe2098:
The disadvantage of this PR will be that, on crash, the user will not be able to see the stack trace or any information related to the crash besides the daemon log. As per mustafa's suggestion, this could be remedied by having the debuginfo available to the snap via a similar mechanism to what gdb uses. This reason alone could be enough to discard the PR, since it would hinder non-graceful issue identification in the user side.

PR #4436, opened 2025-10-14
[clang-tidy] Introduce a simple .clang-tidy file

xmkg:
clang-tidy is a capable linter that would help us enforce the C++ code
style-related guideline rules we have.

Added macro and constexpr variable naming rules for starters.

MULTI-2382

Signed-off-by: Mustafa Kemal Gilor <mustafa.gilor@canonical.com>

PR #4080, opened 2025-05-08
Feature: New Windows Backend (Hyper-V Host Compute System)

xmkg:
This pull request implements the VM and VM factory code for the new Hyper-V Host Compute System backend, on top of the Host Compute System, Host Compute Networking, and virtdisk API wrappers needed for the Hyper-V API-based platform backend feature. The merge proposal includes the implementation for each wrapper and the required tests (unit, integration, module integration).

**IMPORTANT NOTE: The new HCS/HCN-based backend needs the `Virtual Machine Platform` feature to be enabled:**

``` ps
Enable-WindowsOptionalFeature -Online -FeatureName VirtualMachinePlatform
``` 

Test prefixes are as follows:

"test_ut": Denotes a unit test
"test_it": Denotes an integration test
"test_bb_cit"*: Denotes a component integration (a.k.a big bang) test

All wrappers follow the same design :

~~*API function table*
All API calls are dispatched through the API function table. The functions are bound to the real API by default. Tests can override the functions partially or completely as desired.~~

Examples: hyperv_hcn_api_table.h, hyperv_hcs_api_table.h, hyperv_virtdisk_api_table.h

*API Wrapper interface*
Defines the high-level interface for the API being wrapped. All API functions return "OperationResult" as the return type, an aggregate type of a status code, and a status message (wide string).

Examples: hyperv_hcn_wrapper_interface.h, hyperv_hcs_wrapper_interface.h, hyperv_virtdisk_wrapper_interface.h

*API wrapper implementation*
Implements the API wrapper interface.

Examples: hyperv_hcn_api_wrapper.h, hyperv_hcs_api_wrapper.h, hyperv_virtdisk_api_wrapper.h

*Parameter type definitions*
Encapsulates the parameters needed for a wrapper function. If an API function takes more than two parameters, a parameter type definition is used to encapsulate them.

Examples: hyperv_hcn_create_endpoint_params.h, virtdisk_create_virtual_disk_params.h

MULTI-1775
MULTI-1780
MULTI-1799
MULTI-1694

PR #4542, opened 2025-12-02
[applevz] AppleVZ boilerplate

sharder996:
Boilerplate for the basic virtual machine classes. Some of the expected functions have had their signatures added.

- `applevz_virtual_machine.cpp/h`: Self-explanatory, every type of virtual machine in Multipass needs one of these
- `applevz_virtual_machine_factory.cpp/h`: idem
- `applevz_bridge.mm/h: Where all the Objective-C++ code will live. This will include calls into the Virtualization Framework itself.
- `applevz_wrapper.cpp/h`: A mockable singleton wrapper around `apple_vz_bridge.h`. This will allow us to mock the Virtualization Framework for unit testing
- `test_applevz_virtual_machine.cpp`: Future home of unit tests for the `AppleVirtualMachine` class.

This PR also includes the necessary setup for compiling Objective-C++ code and adding the Virtualization Framework as a dependency.

This is the start of an effort to break up the development of the `applevz` Multipass backend into sizable pieces that make it easier to review. The compilation of all classes/files in this PR can effectively be turned off by turning the CMake `APPLEVZ_ENABLED` feature flag off.


src/platform/backends/apple/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+find_library(FOUNDATION_FRAMEWORK Foundation)
+find_library(VIRTUALIZATION_FRAMEWORK Virtualization)
+
+if(NOT FOUNDATION_FRAMEWORK)
+  message(FATAL_ERROR "Could not find Foundation framework")
+endif()
+
+if(NOT VIRTUALIZATION_FRAMEWORK)
+  message(FATAL_ERROR "Could not find Virtualization framework. Ensure you're building on a macOS SDK that provides it.")
+endif()
```
xmkg:
```suggestion
find_library(FOUNDATION_FRAMEWORK Foundation REQUIRED)
find_library(VIRTUALIZATION_FRAMEWORK Virtualization REQUIRED)
```

xmkg (changes requested):
LGTM overall, except for a small part.

src/platform/backends/apple/apple_virtual_machine.cpp
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+#include <apple/apple_virtual_machine.h>
+
+#include <multipass/exceptions/virtual_machine_state_exceptions.h>
+#include <multipass/top_catch_all.h>
+#include <multipass/vm_status_monitor.h>
+
+namespace multipass::apple
+{
```
jimporter: I have no strong opinion here (and if anything, prefer what you have), but in the interest of thoroughness, is this the style we prefer in our `.cpp` files? Most of the files I've seen do something along these lines:

```c++
namespace mpa = multipass::apple;

mpa::AppleVirtualMachine::AppleVirtualMachine(...);
```

Whatever style we use, I think consistency helps keep things easy to read.

CMakeLists.txt
```diff
@@ -189,6 +189,10 @@ if(UNIX)
       list(APPEND MULTIPASS_BACKENDS virtualbox)
   endif()
 
+  if (APPLE)
```
jimporter: For a very simple feature flag, how about something like this?

```suggestion
option(MULTIPASS_ENABLE_APPLEVZ "Enable AppleVZ backend on Apple platforms" ON)

# ...

if (APPLE AND MULTIPASS_ENABLE_APPLEVZ)
```
Later, we could have CI test with this both on and off.

src/platform/backends/apple/apple_virtual_machine.cpp
sharder996: I have been looking at #4080 at lot recently and this style was inspired by the code there. I don't have a strong opinion either way and will honestly just confirm to what the reviewer prefers 😄 

As for consistency, does something like `clang-format` have support for this. Otherwise, it would be hard to enforce consistency and the "correct" style would only live on as tribal knowledge.

CMakeLists.txt
sharder996: I'm open to doing something like this. Let's just reach a consensus with the team or the other reviewer on how we want a feature flag to look like and the how the inclusion/exclusion would work in CI.

jimporter: Sounds good to me. Should we wait on merging this then until we have a consensus on how to manage feature flags and/or partial implementations like this?

src/platform/backends/apple/apple_virtual_machine.cpp
jimporter: I think `clang-tidy` would do the trick here. So long as we make sure our linters are reasonably-flexible, I think adding some more linting would be a big help.

xmkg: That was kinda my point in today's e-mail as well. I've already submitted https://github.com/canonical/multipass/pull/4436 to get the ball rolling.

xmkg (dismissed):

CMakeLists.txt
sharder996: Addressed in #4552

sharder996:
@xmkg Failing because of clang-tidy. Also, clang-tidy seems pretty confused. I didn't get a chance to review the clang-tidy integration, but it would be nice to maybe have it live alongside the Lint workflow so it doesn't failing compilation.

xmkg:
> @xmkg Failing because of clang-tidy. Also, clang-tidy seems pretty confused. I didn't get a chance to review the clang-tidy integration, but it would be nice to maybe have it live alongside the Lint workflow so it doesn't failing compilation.

I've submitted a PR to disable it temporarily: https://github.com/canonical/multipass/pull/4563

The gist of it is it has to run after CMake configure because it needs the `compile_commands.json`, so we can't really move it to the lint step.

sharder996:
@jimporter Can I ask for your eyes on this again now that #4582 is merged? Only difference from your last review will be addition of a feature flag and a little bit of additional wiring to enable the backend in Multipass if the feature flag is enabled.

CMakeLists.txt
```diff
@@ -187,6 +187,10 @@ if(UNIX)
   endif()
 
   list(APPEND MULTIPASS_BACKENDS qemu)
+
+  if(APPLEVZ_ENABLED)
+    list(APPEND MULTIPASS_BACKENDS apple)
```
jimporter: I don't have any strong feelings on this, but would calling the backend `applevz` be clearer? When I saw `APPLE_ENABLED` in `platform_osx.cpp`, I was a bit confused at first until I realized it meant AppleVZ, and not just whether we were building for Apple platforms.

jimporter (dismissed):
This looks good to me overall. Approved with one tiny optional comment.

CMakeLists.txt
sharder996: Ok, I'm interpreting this as rename everything here that was `apple` as `applevz`; classes, directories, macros, etc.

sharder996: PR description also updated.

src/platform/backends/apple/apple_virtual_machine.cpp
sharder996: LMK if there is an actionable item on my end from this.

CMakeLists.txt
jimporter: Sounds good to me!

jimporter (dismissed):
LGTM!

xmkg (dismissed):

jimporter (dismissed):

tests/unit/applevz/test_applevz_virtual_machine.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <applevz/applevz_virtual_machine.h>
+
+#include "tests/unit/common.h"
+
+namespace multipass::test
+{
+struct AppleVZVirtualMachine_UnitTests : public testing::Test
+{
+};
+}; // namespace multipass::test
```
jimporter: Nit: extraneous semicolon.
```suggestion
} // namespace multipass::test
```

sharder996: I blame it on Copilot autocompletion 😄

xmkg (approved):

PR #4543, opened 2025-12-02
[objcpp] ObjCpp formatting rules

sharder996:
Formatting rules for Objective-C/C++ code. Basically just applying as many of the C++ rules as are supported onto Obj-C code.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{cpp,h}]
+[*.{mm,cpp,h}]
 indent_size = 4
 indent_style = space
 insert_final_newline = true


.clang-format
```diff
@@ -5,6 +5,42 @@ BasedOnStyle: LLVM
...
+BreakBeforeBraces: Custom
+ColumnLimit: 100
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+IndentExternBlock: NoIndent
+IndentWidth: 4
+InsertNewlineAtEOF: true
+PenaltyReturnTypeOnItsOwnLine: 120
+PointerAlignment: Left
```
xmkg, line 37: We can move these to a section where no language is set, like in libcxx's .clang-format file:

https://github.com/llvm/llvm-project/blob/main/libcxx/.clang-format

This way all langs would inherit from it. Then, we can define language-specific parts as needed.

sharder996: Done

xmkg:
@sharder996 I think you can target main directly with this one, then rebase you apple-vz branch once this is merged.

PR #4547, opened 2025-12-03
[macos] Set GUI build target

sharder996:
Accurately set the build target for the Flutter GUI when on macOS. Do the same for dependencies installed with `pod`.


tobe2098 (approved):
LGTM!

PR #4548, opened 2025-12-04
[s390x] Compile Multipass on s390x architecture

sharder996:
Gets Multipass compiling on s390x again. Changes are categorized into several areas:

#### CMake
- Enabled s390x compilation for vcpkg
- Moved check for `MULTIPASS_ENABLE_FLUTTER_GUI` in cmake up one level so we can avoid unnecessarily compiling the Dart FFI interface

#### Snapcraft / CI
- Included additional firmware for QEMU on s390x
- Fixed bug in Linux CI where we install `ccache` interactively

#### QEMU
- Adjusted QEMU process arguments when running on s390x
- Added s390x QEMU process arguments in unit test

#### Images
- Fixed bug in `CustomImageHost` where we incorrectly assume that the current CPU architecture will be in the manifest
- Made tests in `test_custom_image_host.cpp` more architecture agnostic so they continue to function currently even if the current CPU architecture is not in the list of supported architectures in the manifest
- Included s390x images from the Fedora scraper

~~From this PR, there are still a few outstanding unknowns:~~

~~Firstly, I've carefully based this branch to just before #4530 was merged into `main`. vcpkg fetches Qt from a domain that for whatever reason doesn't play nice with the proxy that my s390x development machine is behind. I'm still using the same version of Qt, just from a different source, so fingers-crossed this doesn't cause an issue when it runs in Launchpad. Also, the Canonical self-hosted s390x runners are incredibly flaky. After many re-runs I was only able to get to the packaging stage once when the runner subsequently failed due to a networking issue. The [documentation](https://canonical-self-hosted-github-runner-docs.readthedocs-hosted.com/en/latest/usage/available_runners/#available-runners) itself lists the runners as **experimental** so I gave up on them.~~ 

~~See code comments on adding a few configure options to the compilation of dbus and qtbase.~~ **FIXED**

~~Secondly, while omitting the GUI from snapcraft is simple enough, getting the same recipe to work for both GUI and non-GUI packages has proven to be quite complex. Turning off the compilation of the GUI on s390x can be easily done through CMake, but snapcraft still wants to pull in the `gnome` extension which depends on snaps that aren't available on s390x. We might be able to get around this requirement by painstakingly individually redefining what the `gnome` extension automates for us, but I foresee it being error prone and difficult to maintain in the future.~~

~~Instead, what I propose is that we maintain an additional `git` branch, say `gui-less-snap`, where we simply omit the `gui` app from `snapcraft.yaml` and limit the architectures under `build-on`. More components could theoretically be omitted to optimize the package size, but would come with the tradeoff of being more difficult to maintain. Then, we create an additional snap recipe in Launchpad that uses this branch and publishes Multipass for these special architecture. Maintenance of this branch could be automated with rebases onto `main` whenever merge queues runs with Mattermost/email alerts happening when intervention is required.~~

Snap packaging will be handled in Launchpad through the use of another snap recipe which will pull from a unique git branch. Will be addressed in a separate PR.

Closes #4549 

---

MULTI-2377

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -226,7 +226,8 @@ parts:
       VERSION=$( awk -F\" '/version_string/ { print $2 }' gen/multipass/version.h )
       craftctl set version=${VERSION}
       craftctl set grade=$( echo ${VERSION} | grep -qe '^[0-9]\+\.[0-9]\+\.[0-9]\+\($\|-rc\)' && echo stable || echo devel )
-      sed -i -e 's@^Icon=\(.\+\)$@Icon=/usr/share/icons/hicolor/scalable/apps/\1.svg@' ${CRAFT_PART_INSTALL}/usr/share/applications/multipass.gui.desktop
+      [ -f ${CRAFT_PART_INSTALL}/usr/share/applications/multipass.gui.desktop ] && \
+        sed -i -e 's@^Icon=\(.\+\)$@Icon=/usr/share/icons/hicolor/scalable/apps/\1.svg@' ${CRAFT_PART_INSTALL}/usr/share/applications/multipass.gui.desktop
 
       cd ${CRAFT_PART_SRC}/3rd-party
       find . -type f \( -name 'LICENSE*' -o -path '*/LICENSE*/*' \) -print0 | xargs -0 -I FILE install --mode=644 -D --no-target-directory FILE ${CRAFT_PART_INSTALL}/licenses/FILE


snap/snapcraft.yaml
```diff
@@ -295,12 +285,12 @@ parts:
       usr/lib/qemu/*: usr/bin/
       usr/share/qemu/*: qemu/
     override-pull: |-
-      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "s390x" ] && exit 0
```
sharder996, line 300: Not exactly sure why we have this line since I believe this will always exactly match up with `build-on`...

3rd-party/vcpkg-ports/dbus/portfile.cmake
```diff
@@ -0,0 +1,94 @@
...
+        -DDBUS_ENABLE_XML_DOCS=OFF
+        -DDBUS_INSTALL_SYSTEM_LIBS=OFF
+        #-DDBUS_SERVICE=ON
+        -DDBUS_WITH_GLIB=OFF
+        -DTHREADS_PREFER_PTHREAD_FLAG=ON
+        -DXSLTPROC_EXECUTABLE=FALSE
+        "-DCMAKE_INSTALL_SYSCONFDIR=${CURRENT_PACKAGES_DIR}/etc/${PORT}"
+        "-DWITH_SYSTEMD_SYSTEMUNITDIR=lib/systemd/system"
+        "-DWITH_SYSTEMD_USERUNITDIR=lib/systemd/user"
+        -DDBUS_SESSION_SOCKET_DIR=/tmp
```
sharder996: @xmkg, I had to add `-DDBUS_SESSION_SOCKET_DIR=/tmp` to the dbus build configuration to get dbus compiling on s390x.

Copilot/Claude Code's explanation on why this is needed: When vcpkg's triplet system sets `VCPKG_CMAKE_SYSTEM_NAME`, CMake detects this as cross-compilation even during native s390x builds. The dbus CMake configuration refuses to autodetect the session socket directory during cross-compilation and requires it to be explicitly set. `tmp` is the standard Linux location.

Is this perhaps an indication of a shortcoming of the base portfile as provided by vcpkg? Would this be better off fixed upstream?

Tagging you because you've done some work with vcpkg recently and thought you might have some ideas.

3rd-party/vcpkg-ports/qtbase/portfile.cmake
```diff
@@ -0,0 +1,608 @@
...
+        androiddeployqt6
+        syncqt
+        tracepointgen
+    )
+
+qt_install_submodule(PATCHES    ${${PORT}_PATCHES}
+                     TOOL_NAMES ${TOOL_NAMES}
+                     CONFIGURE_OPTIONS
+                        #--trace-expand
+                        -DCMAKE_CROSSCOMPILING:BOOL=OFF
```
sharder996: @xmkg idem had to add a configure option in order to get qtbase compiling on s390x.

Copilot/Claude code's explanation: Qt's build system also detected cross-compilation (due to vcpkg's triplet configuration) and required `QT_HOST_PATH` to be set. Since this is actually a native build on s390x, we explicitly tell CMake it's NOT cross-compiling, which allows Qt to build normally without needing a separate host toolchain.

sharder996:
@reviewers Please ignore the last two commits for the time being as they are experimental and will most likely change.

3rd-party/vcpkg-ports/dbus/portfile.cmake
xmkg: I think you should be able to avoid creating a port overlay with per-port customization in a custom triplet, something like this:

```cmake
# Base triplet settings, copied (or included) from the triplet file
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE static)

# Per-port customization
if(PORT STREQUAL "dbus")
    list(APPEND VCPKG_CMAKE_CONFIGURE_OPTIONS
        "-DDBUS_SESSION_SOCKET_DIR=/tmp"
    )
endif()
```

This way we wouldn't have to burden ourselves with maintaining the port overlay. We're already using custom triplets for Windows and macOS so it wouldn't hurt much to have a triplet for linux too. You can copy the default triplet that the build is using right now and customize it in the `vcpkg-triplets` folder.

I might need to opt-in to something similar for the gRPC port anyway.

xmkg: I'll also do it for the libssh -- so feel free to continue working with the port overlay. We can migrate it to the said approach after the libssh work is finished.

3rd-party/vcpkg-ports/qtbase/portfile.cmake
xmkg: CMAKE_CROSSCOMPILING is the perfect kind of thing to be set in the triplet file. Feel free to experiment with the triplet overlay approach to see if it works.

3rd-party/vcpkg-ports/dbus/portfile.cmake
xmkg: https://github.com/canonical/multipass/pull/4614

xmkg (approved):
Looks clean and neat! Approved (secondary).

PR #4550, opened 2025-12-04
[tests-constants] Mark stoi result as unused

xmkg:
...to get rid of this:

image

MULTI-2399

### DIFF
--- a/tests/test_constants.cpp
+++ b/tests/test_constants.cpp
@@ -25,8 +25,8 @@ using namespace testing;
 
 TEST(Constants, constantsConstraints)
 {
-    EXPECT_NO_THROW(std::stoi(mp::min_cpu_cores));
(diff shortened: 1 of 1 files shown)


tests/test_constants.cpp
```diff
@@ -25,8 +25,8 @@ using namespace testing;
 
 TEST(Constants, constantsConstraints)
 {
-    EXPECT_NO_THROW(std::stoi(mp::min_cpu_cores));
-    EXPECT_NO_THROW(std::stoi(mp::default_cpu_cores));
+    EXPECT_NO_THROW([[maybe_unused]] auto _ = std::stoi(mp::min_cpu_cores));
```
xmkg, line 28: This ends up in a if block so declaring a variable like this is fine.

PR #4561, opened 2025-12-09
[applevz] Basic VM functions

sharder996:
This PR implements some of the basic functions for manipulating the state of a virtual machine using the Apple Virtualization Framework. This includes:

- `start()`
- `shutdown(ShutdownPolicy shutdown_policy)`
- `suspend()`

All Objective-C++ code, other helper functions needed to perform these actions, and unit tests for these functions are also included.

On top of being able to manually manage memory, Objective-C uses Automatic Reference Counting (ARC) where the system reference counting to automatically insert appropriate memory management method calls for you at compile-time. However, in order for objects to persist across the Objective-C/C++ boundary, we must use some types from the `CoreFoundation` library instead of the `Foundation` library, i.e., `CFErrorRef` over `NSError`, also known as Toll-Free Bridged Types. This allows us to control who owns the object and avoids having ARC and the Objective-C memory management system release memory before we are done with it. There are several macros that are used to tell the compiler about the ownership semantics of an object, but we mainly use:

- `__bridge`; for objects such as the `VMHandle` which were declared on the C++ side and where we don't want ownership to transfer when casting the `std::shared_ptr<void>` to virtualization framework virtual machine type; `VZVirtualMachine`.

- `__bridge_retained`: for objects that were created on the Objective-C side and for which we want to persist even after the block in which they were created finishes. We are responsible for manually freeing this type of object, so care must be taken in order when using this macro to prevent memory leaks. The use of this macro here is limited to returning errors from the Objective-C side to the C++ side through the use of the custom C++ type; `CFError`, which automatically calls `CFRelease` when destructed.

The other unique memory management mechanism that is used is the `autoreleasepool` block. This is used in special circumstances when a large number of temporary autoreleased object. The `autoreleasepool` sends an `autorelease` message to all objects in that block which are immediately released instead of at the end of the current event-loop iteration. This prevents temporary objects from unnecessarily accumulating and causing excessive overhead.

Additional reading:
- [Toll-Free Bridged Types](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFDesignConcepts/Articles/tollFreeBridgedTypes.html)
- [Core Foundation Memory Management](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/CFMemoryMgmt.html)
- [About Memory Management](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html)

---

MULTI-2258
MULTI-2261

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -30,6 +30,8 @@
 #include <string>
 #include <vector>
 
+#include <fmt/format.h>
+
 namespace multipass
 {
 struct IPAddress;
@@ -75,7 +77,7 @@ public:
     virtual std::string ssh_hostname()
     {
         return ssh_hostname(std::chrono::minutes(2));
-    };
+    }
     virtual std::string ssh_hostname(std::chrono::milliseconds timeout) = 0;
     virtual std::string ssh_username() = 0;
     virtual std::optional<IPAddress> management_ipv4() = 0;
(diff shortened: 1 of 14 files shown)


sharder996:
Replaced by #4653

PR #4564, opened 2025-12-11
[contributing] Add guidelines for feature flags

jimporter:
# Description

This PR adds some initial guidelines for how contributors should work with feature flags. (Note that this only includes the parts relevant to contributors, and not our full specification, which is mainly of interest to maintainers.)

## Checklist

- [ ] I have tested the changes locally

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,14 +112,11 @@ branching for release.<br>
 target the `main` branch, even if it is part of a larger task. This should be the most common
 case.<br>
 **PR8.** If your PR relies on another one, target the other's branch.<br>
-**PR9.** When working on a larger set of changes with cohesive interdependence, consider using a
-feature branch.<br>
-**PR10.** Try to keep the number of concurrent feature branches small.<br>
-**PR11.** When PRs are stacked, prefer to merge them in order. The target branch will update
+**PR9.** When PRs are stacked, prefer to merge them in order. The target branch will update
 automatically upon merging.<br>
-**PR12.** PRs should include descriptions and/or point to appropriate context (within reason).<br>
-**PR13.** When authoring a PR, make sure to test it.<br>
-**PR14.** When authoring a PR, make sure to review its diff.<br>
+**PR10.** PRs should include descriptions and/or point to appropriate context (within reason).<br>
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -139,6 +136,18 @@ approval.<br>
 **RVW8.** Renovate PRs may be merged after a single primary approval.<br>
 **RVW9.** Review comments should be acknowledged by the author, but *resolved* by the reviewer.<br>
 
+### Feature flags (FF)
+
+**FF1.** When making an interdependent set of changes too large to review and merge as a single PR, consider adding
+a feature flag for it.<br>
+**FF2.** Where possible, feature flags should be used to disable any new code or behavior which is
+part of the given feature; however, "stub" code that has no effect on users need not be gated behind
+a feature flag.<br>
```
ricab: - FF3. Strive to encapsulate code that is dependent on a feature flag, preferably in its own files, so that those files can be entirely included or excluded from builds.
- FF4. Where separate files are not feasible or justified, use preprocessor directives to selectively enable or disable feature code. Keep this approach as the exception rather than the norm.
- FF5. Minimize unreachable code under a feature flag, i.e., code that can't be reached even though the feature is enabled.

CONTRIBUTING.md
```diff
@@ -139,6 +136,18 @@ approval.<br>
...
+**FF3.** PRs for code behind a feature flag should otherwise be treated as usual, with authors and
+reviewers maintaining the same standards of quality as for other PRs.<br>
+**FF4.** When a feature is fully-complete and suitable for release, the corresponding feature flag
+should be completely removed from Multipass.<br>
```
ricab, line 159: FF8? No unreachable code should be left behind after a feature flag is removed.

CONTRIBUTING.md
jimporter: Thanks for the suggestions. I've added these, and also split my old FF2 in two, to provide a bit more detail.

CONTRIBUTING.md
jimporter: Makes sense. Added this too.

xmkg (approved):

PR #4566, opened 2025-12-11
[utils] Use Boost.JSON for `AliasDict`

jimporter:
# Description

(Recreating this from #4558, since Github auto-closed that one after I merged the parent PR, and won't let me change the base branch to `main`...)

This changes `AliasDict` to use Boost.JSON. Functionally, everything should be the same as before, but there are a few notable differences now:

* (De)serialization is broken up into smaller bits, which should be easier to understand.
* The `AliasDict` constructor used to load the data from a file; now, the constructor just initializes the data structure, and to load a file, you call `AliasDict::load_file`. This makes it easier to unit test `AliasDict` without having to write temp files.
* Boost.JSON throws exceptions if a key is missing in a JSON object, whereas Qt's JSON library returns an empty/null object. I think the Boost.JSON way is better, but we'll need to be careful that we aren't relying on the Qt behavior anywhere.
* Boost.JSON preserves key order when writing JSON objects. This is nice, since it means we can adjust the ordering to be more human-readable.

## Testing

- Manual testing steps:

  1. Create an alias, e.g.: `./bin/multipass alias plucky-tortoise:whoami whoisplucky`
  2. Restart multipass
  3. Run the alias, e.g.: `~/.local/share/multipass/bin/whoisplucky`

## Checklist

### DIFF
--- a/tests/test_alias_dict.cpp
+++ b/tests/test_alias_dict.cpp
@@ -100,20 +100,8 @@ TEST_F(AliasDictionary, worksWithBrokenFile)
 
 TEST_F(AliasDictionary, skipsCorrectlyBrokenEntriesOldFormat)
 {
-    std::string file_contents{"{\n"
-                              "    \"alias1\": {\n"
-                              "        \"command\": \"first_command\",\n"
-                              "        \"instance\": \"first_instance\",\n"
-                              "        \"working-directory\": \"map\"\n"
-                              "    },\n"
-                              "    \"empty_entry\": {\n"
-                              "    },\n"
-                              "    \"alias2\": {\n"
-                              "        \"command\": \"second_command\",\n"
-                              "        \"instance\": \"second_instance\",\n"
-                              "        \"working-directory\": \"default\"\n"
-                              "    }\n"
-                              "}\n"};
+    auto file_contents =
+        mpt::load_test_file("alias_dict/broken_entries_old_format.json").toStdString();
 
     auto dict = value_to<mp::AliasDict>(boost::json::parse(file_contents), json_context);
 
@@ -136,25 +124,7 @@ TEST_F(AliasDictionary, skipsCorrectlyBrokenEntriesOldFormat)
 
 TEST_F(AliasDictionary, skipsCorrectlyBrokenEntries)
 {
-    std::string file_contents{"{\n"
-                              "    \"active-context\": \"default\",\n"
-                              "    \"contexts\": {\n"
-                              "        \"default\": {\n"
-                              "            \"alias1\": {\n"
-                              "                \"command\": \"first_command\",\n"
-                              "                \"instance\": \"first_instance\",\n"
-                              "                \"working-directory\": \"map\"\n"
-                              "            },\n"
-                              "            \"empty_entry\": {\n"
-                              "            },\n"
-                              "            \"alias2\": {\n"
-                              "                \"command\": \"second_command\",\n"
-                              "                \"instance\": \"second_instance\",\n"
-                              "                \"working-directory\": \"default\"\n"
-                              "            }\n"
-                              "        }\n"
-                              "    }\n"
-                              "}\n"};
+    auto file_contents = mpt::load_test_file("alias_dict/broken_entries.json").toStdString();
 
     auto dict = value_to<mp::AliasDict>(boost::json::parse(file_contents), json_context);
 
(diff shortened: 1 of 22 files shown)


tests/test_alias_dict.cpp
```diff
@@ -467,13 +459,13 @@ INSTANTIATE_TEST_SUITE_P(
                         "    \"contexts\": {\n"
                         "        \"default\": {\n"
                         "            \"llp\": {\n"
-                        "                \"command\": \"ls\",\n"
                         "                \"instance\": \"primary\",\n"
+                        "                \"command\": \"ls\",\n"
```
sharder996: I know its out of scope for this PR, but seeing as how we're here already, do you feel its worth it to move these string literals out into golden test files?

sharder996 (approved):
I don't see anything out of the ordinary here. I expect that merging in #4515 ironed out some of the details beforehand.

Otherwise, just one optionally improvement that could be done, but good for merging either way.

src/client/common/alias_dict.cpp
```diff
@@ -20,47 +20,39 @@
...
-} // namespace
 
-mp::AliasDict::AliasDict(mp::Terminal* term) : cout(term->cout()), cerr(term->cerr())
+mp::AliasDict::AliasDict(mp::Terminal* term,
+                         const std::string& active_context,
+                         const std::string& filename)
+    : active_context(active_context),
+      aliases{{active_context, AliasContext{}}},
+      aliases_file(filename),
+      cout(term->cout()),
```
tobe2098: I do not know if we can squeeze this in, but I think there is no code that still uses cout. Could be a TODO to remove it.

src/client/cli/formatter/csv_formatter.cpp
```diff
@@ -248,12 +249,12 @@ std::string mp::CSVFormatter::format(const mp::AliasDict& aliases) const
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "Alias,Instance,Command,Working directory,Context\n");
 
-    for (const auto& [context_name, context_contents] : sort_dict(aliases))
+    for (const auto& [context_name, context_contents] : sorted_map(aliases))
```
tobe2098: Maybe out of scope, but we could put a TODO if you agree:
- Constructing a map just to iterate over it is a bit wasteful. We could do a vector of key const refs, sort them and then iterate and index with the key. WDYT?

include/multipass/cli/alias_dict.h
```diff
@@ -74,6 +101,7 @@ class AliasDict
     {
         return aliases.cend();
     }
+
```
tobe2098, line 104: Was this blank line added for readability?

tobe2098 (changes requested):
Very good work Jim! LGTM mostly, just a couple nitpicks/questions.

include/multipass/cli/alias_dict.h
jimporter: Yeah, I thought it helped to keep the various `begin`/`end` pairs close together, but to add some space between each set. I don't have any strong feelings about it though.

src/client/cli/formatter/csv_formatter.cpp
jimporter: I'd considered that, but a container of references doesn't work unfortunately. Once a reference is set, you can't change it without some trickery, so `sort` wouldn't be able to swap the elements around. (The container probably won't compile anyway, since other functions like `insert` wouldn't work with references either.)

However, a `std::reference_wrapper` *does* work. After a few tries, I put something together that should be more efficient, and without too much more code.

src/client/common/alias_dict.cpp
jimporter: I removed this. If we ever need it again, we can just add it back.

tests/test_alias_dict.cpp
jimporter: Seems like a good idea to me. Done.

tests/test_data/alias_dict/.clang-format
```diff
@@ -0,0 +1 @@
+DisableFormat: true
```
tobe2098: I imagine this is here for the `formatter` data folder, would it be better to place it there instead? Or did you experience any issues with data formatting?

tobe2098 (dismissed):
LGTM, just one small thing. Great work Jim!

tests/test_data/alias_dict/.clang-format
jimporter: I put it in the folder above since the other test files in that directory weren't formatted to `clang-format`'s liking. We could fix the formatting to make `clang-format` succeed, but I opted to do it this way since the formatting in those files is what our serialization code writes out.

I don't mind either way though.

tests/test_data/alias_dict/formatter/.gitattributes
```diff
@@ -0,0 +1,4 @@
+*.csv -text
```
sharder996: We also have a `.gitattributes` in the top level directory. I don't have a preference whether we maintain one project level gitattributes file or scatter them where they are need in the test data. Just that we maintain an approach. So, I'd like to either see them combined or have the other one moved to `tests/test_data/formatters`

jimporter: Makes sense. I moved this to the top-level `.gitattributes` since that's the precedent we already established.

sharder996 (approved):
LGTM! Thanks @jimporter

jimporter:
(Trying a second merge since I'm pretty sure the first one failed due to a runner dying. Not sure if our builds are just brittle or if it's Microsoft's infra being flaky...)

PR #4567, opened 2025-12-12
Unify return code framework within the multipass CLI

### LINKED ISSUE #3054: [cli] When everything else goes well, `multipass shell` should return the exit code of the inner shell

### LINKED ISSUE #4428: Unify return codes from 3rd-party libraries with mp::ReturnCode
In the source, return codes from libssh, beyond SSH_OK (== or !=) are either ignored or improperly converted to mp::ReturnCode, which does not cover negative return codes and does not convert properly to the existing codes. Similarly, the same thing happens for the return codes obtained from the process run within the vm (obtained with ssh_get_exit_state). 

This may also be the case in other 3rd-party libraries we are using.

Extend the definition of ReturnCode and define a conversion from the different types of return code to mp::ReturnCode for later use (for example, SSH_AGAIN (-2) -> Retry (4)) or for logging purposes. The exit state return codes could be separate to mp::ReturnCode, and then they would have their own namespace for better for logging.

tobe2098:
# Description

- **What does this PR do?** This PR changes the way return codes are propagated up the stack up to the eventual return code of the CLI, and other minor adjustments. For shell and exec the return code for a multipass-side failure will be 255 from now on. 
- **Why is this change needed?** Apart from making explicit the fact that we use the same propagation path for VM return codes and multipass return codes, this changes solve an existing bug and allow the VM return code from multipass shell to be propagated to the CLI's return code. Additionally, ssh_channel_get_exit_state is moved to the existing framework for libssh calls and a unused function is removed.  

## Related Issue(s)

Closes #4428 
Closes #3054

## Testing

- Unit tests
- Manual testing steps:

  1. Launch daemon in another terminal (to avoid return code conflicts)
  2. multipass shell
  3. exit n (0-255)
  4. Return code == n
  5. If n=4 both in multipass shell and multipass exec -- exit n there is no infinite loop

## Checklist

MULTI-2412

### DIFF
--- a/include/multipass/cli/return_codes.h
+++ b/include/multipass/cli/return_codes.h
@@ -46,12 +46,12 @@ enum VMReturnCode
 // used, index-based get<> and holds_alternative<> are needed.
 using ReturnCodeVariant = std::variant<ReturnCode, VMReturnCode>;
 
-inline bool are_return_codes_equal(ReturnCodeVariant rc1, ReturnCode rc2)
+inline bool operator==(ReturnCodeVariant rc1, ReturnCode rc2)
 {
     // A VMReturnCode can only be obtained if everything in multipass works properly and we manage
     // to obtain that code. In that case, the mp::ReturnCode can only be ReturnCode::Ok.
-
-    return (std::holds_alternative<VMReturnCode>(rc1) && rc2 == ReturnCode::Ok) ||
-           (std::holds_alternative<ReturnCode>(rc1) && std::get<ReturnCode>(rc1) == rc2);
+    if (auto* rc = std::get_if<ReturnCode>(&rc1))
+        return *rc == rc2;
+    return rc2 == ReturnCode::Ok;
 }
 } // namespace multipass
--- a/src/client/cli/cmd/alias.cpp
+++ b/src/client/cli/cmd/alias.cpp
@@ -135,13 +135,13 @@ mp::ParseCode cmd::Alias::parse_args(mp::ArgParser* parser)
 
     auto ret = dispatch(&RpcMethod::info, info_request, on_success, on_failure);
 
-    if (are_return_codes_equal(ret, ReturnCode::DaemonFail))
+    if (ret == ReturnCode::DaemonFail)
     {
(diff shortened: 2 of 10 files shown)


include/multipass/cli/return_codes.h
```diff
@@ -32,6 +34,23 @@ enum ReturnCode
...
+    Retry = 4,
+    ShellExecFail = 255,
+};
+
+enum VMReturnCode
+{
 };
+// Only implicitly int-convertible types should be used here. If at least 2 of the same type are
+// used, index-based get<> and holds_alternative<> are needed.
+using ReturnCodeVariant = std::variant<ReturnCode, VMReturnCode>;
```
ricab, line 47: That's it, that's what we need! Thank you so much for taking care of it :+1:

ricab: PS: not an actual review, just had a light peek

sharder996 (dismissed):
LGTM! Thanks for tackling this @tobe2098

include/multipass/cli/return_codes.h
```diff
@@ -32,6 +34,24 @@ enum ReturnCode
...
+
+// The purpose of this enum is to have a compile-time tagged int for VM-internal return codes
+enum VMReturnCode
+{
 };
+// Only implicitly int-convertible types should be used here. If at least 2 of the same type are
+// used, index-based get<> and holds_alternative<> are needed.
+using ReturnCodeVariant = std::variant<ReturnCode, VMReturnCode>;
+
+inline bool are_return_codes_equal(ReturnCodeVariant rc1, ReturnCode rc2)
```
xmkg: Is there a reason why this function is not a free equal operator overload, like:

```c++
inline bool operator==(ReturnCodeVariant rc1, ReturnCode rc2) {
    if (auto* rc = std::get_if<ReturnCode>(&rc1))
        return *rc == rc2;
    return rc2 == ReturnCode::Ok;
}
```

Issue #4383, opened 2025-09-22
macOS 26 Tahoe: bridge subnet changed to 192.168.2.0/24, conflicts with local network


BeckerFrank:
**Describe the bug**
It is impossible to create a new multipass VM.

**To Reproduce**
How, and what happened?
1. start with an empty list of VM's (verify that /var/db/dhcpd_leases is empty)
1. in the Multipass UI select catalog and launch Ubuntu 25.10
1. after the download of the image you see the starting message but after some time you will get "timed out for waiting for response"
1. `multipass list shows
Name                    State             IPv4             Image
supple-alien            Unknown           --               Ubuntu 25.10
1. /var/db/dhcpd_leases is filled with
{
        name=supple-alien
        ip_address=192.168.2.2
        hw_address=1,52:54:0:12:aa:4b
        identifier=1,52:54:0:12:aa:4b
        lease=0x68d18bea
}

**Expected behavior**
What did you expect to happen?

**Logs**
[2025-09-22T18:48:12.615] [debug] [qemu-img] [20486] started: qemu-img info --output=json /var/root/Library/Caches/multipassd/qemu/vault/images/questing-20250903/questing-server-cloudimg-arm64.img
[2025-09-22T18:48:12.702] [debug] [qemu-img] [20487] started: qemu-img amend -o compat=1.1 /var/root/Library/Caches/multipassd/qemu/vault/images/questing-20250903/questing-server-cloudimg-arm64.img
[2025-09-22T18:48:12.770] [debug] [qemu-img] [20488] started: qemu-img info /var/root/Library/Caches/multipassd/qemu/vault/images/questing-20250903/questing-server-cloudimg-arm64.img
[2025-09-22T18:48:12.838] [debug] [qemu-img] [20489] started: qemu-img resize /var/root/Library/Application Support/multipassd/qemu/vault/instances/supple-alien/questing-server-cloudimg-arm64.img 5368709120
[2025-09-22T18:48:12.907] [debug] [qemu-img] [20490] started: qemu-img snapshot -l /var/root/Library/Application Support/multipassd/qemu/vault/instances/supple-alien/questing-server-cloudimg-arm64.img
[2025-09-22T18:48:12.915] [debug] [supple-alien] process working dir ''
[2025-09-22T18:48:12.915] [info] [supple-alien] process program 'qemu-system-aarch64'
[...]

PR #4569, opened 2025-12-12
[mac] Add some tracing for `arp` results

ricab:
Trying to debug #4383


ricab:
This has served its purpose.

PR #4571, opened 2025-12-14
[image-decoder] Convert to C++STL

sharder996:
Part 2 of removing Qt one piece at a time. This time in the `XzImageDecoder` class.

This PR replaces a few usages of `mp::Path` with `std::filesystem::path` and `QFile` with file streams.

Eliminates one more leaf node of the Qt network and pushes it up one more function call.


jimporter (changes requested):
The changes themselves all look good, though I'm a little worried since there's no test coverage for this code (at least not reported to Codecov). Would it be feasible to add some tests here?

PR #4575, opened 2025-12-15
[scraper] Check for `None` before validating

sharder996:
Latest run of the distro scraper [failed](https://github.com/canonical/multipass/actions/runs/20202613409/job/57996599670#step:6:83) because the Fedora scraper timed out and returned `None`. Added some addition error handling to account for this case.

### DIFF
--- a/tools/distro-scraper/scraper/cli.py
+++ b/tools/distro-scraper/scraper/cli.py
@@ -100,6 +100,7 @@ async def run_scraper(scraper_instance: BaseScraper) -> tuple[str, dict | None]:
         result = await scraper_instance.fetch()
     except Exception as e:
         logger.exception("Scraper '%s' failed: %s", name, e)
+        return name, None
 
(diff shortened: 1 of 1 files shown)


xmkg (dismissed):

tools/distro-scraper/scraper/cli.py
```diff
@@ -103,11 +103,16 @@ async def run_scraper(scraper_instance: BaseScraper) -> tuple[str, dict | None]:
 
     # Validate JSON structure
     try:
+        if result is None:
+            raise ValueError("No data returned from scraper")
+
```
jimporter: I think once we catch the exception on lines 101-102, we know for sure we won't be able to turn `result` into a `ScraperResult` (since it's guaranteed to be `None`). How about just returning early in that case?

```suggestion
        return name, None

    # Validate JSON structure
    try:
```

PR #4582, opened 2025-12-15
[cmake] Add support for defining feature flags

jimporter:
# Description

This adds a `feature_flag` macro in CMake that we can use to define feature flags, plus an option (`MP_DISABLE_ALL_FEATURES`) to disable them all in one go. We can then set this flag in the CI for our release branches (and nightly builds).

## Related Issue(s)

* https://github.com/canonical/multipass/pull/4570
* https://github.com/canonical/multipass/pull/4564

## Testing

Use the demo commit here to try out setting some feature flags, e.g. `cmake .. -DAVAILABILITY_ZONES_ENABLED=OFF` or `cmake .. -DMP_ALLOW_OPTIONAL_FEATURES=OFF`.

## Checklist

### DIFF
--- a/src/cmake/versioning.cmake
+++ b/src/cmake/versioning.cmake
@@ -46,22 +46,31 @@ function(determine_version OUTPUT_VARIABLE)
   endif()
 
   string(REGEX MATCH "^v.+-([0-9]+)-(g.+)$" GIT_VERSION_MATCH ${GIT_VERSION})
+
+  # Snapcraft expects versions to be at most 32 characters. Currently, with these feature flag
+  # suffixes, our longest version IDs are exactly at this limit.
   if(NOT MP_ALLOW_OPTIONAL_FEATURES)
     set(MULTIPASS_VERSION_FEATS "-noff")
   else()
     is_custom_feature_set(MULTIPASS_CUSTOM_FEATS)
     if(MULTIPASS_CUSTOM_FEATS)
-      set(MULTIPASS_VERSION_FEATS "-custom")
+      set(MULTIPASS_VERSION_FEATS "-cstm")
     endif()
   endif()
 
   if(GIT_RELEASE)
-    # Don't indicate presence/absence of optional features for release builds.
+    # Only use the exact release tag for release versions. Don't indicate presence/absence of
+    # optional features for release builds.
     set(NEW_VERSION ${GIT_RELEASE})
-  elseif(GIT_VERSION_MATCH AND MULTIPASS_BUILD_LABEL)
-    set(NEW_VERSION ${GIT_TAG}.${CMAKE_MATCH_1}.${MULTIPASS_BUILD_LABEL}+${CMAKE_MATCH_2}${MULTIPASS_VERSION_FEATS})
   elseif(GIT_VERSION_MATCH)
-    set(NEW_VERSION ${GIT_TAG}.${CMAKE_MATCH_1}+${CMAKE_MATCH_2}${MULTIPASS_VERSION_FEATS})
+    # If MULTIPASS_BUILD_LABEL is set, show add it to the version. Otherwise, use the number of
+    # commits ahead of the tag.
+    if(MULTIPASS_BUILD_LABEL)
+      set(NEW_VERSION ${GIT_TAG}.${MULTIPASS_BUILD_LABEL})
+    else()
+      set(NEW_VERSION ${GIT_TAG}.${CMAKE_MATCH_1})
+    endif()
+    string(APPEND NEW_VERSION "+${CMAKE_MATCH_2}${MULTIPASS_VERSION_FEATS}")
   else()
     message(FATAL_ERROR "Failed to parse version number: ${GIT_VERSION}")
   endif()


CMakeLists.txt
```diff
@@ -15,9 +15,19 @@
 cmake_minimum_required(VERSION 3.25)
 cmake_policy(SET CMP0079 NEW) # Allow target_link_libraries() in subdirs
 
+include(src/cmake/feature-flags.cmake)
 include(src/cmake/environment-utils.cmake)
 is_running_in_ci(IS_RUNNING_IN_CI)
 
+# A feature flag with a dependency.
+feature_flag(APPLEVZ_ENABLED "AppleVZ backend" APPLE)
```
sharder996: I'm not 100% sure, but the impression was that all the feature flags (an possibly feature flag specific cmake changes) would live in a separate file/module.

@xmkg would probably be able to provide a more knowledgeable opinion than I could.

CMakeLists.txt
```diff
@@ -15,9 +15,19 @@
...
+# A feature flag with no dependencies.
+feature_flag(AVAILABILITY_ZONES_ENABLED "Availability zone support")
+
+# Print the features' status to show this off.
+message(STATUS "APPLEVZ_ENABLED = ${APPLEVZ_ENABLED}")
```
sharder996: Is the intention to then litter:

```
#if(APPLEVZ_ENABLED)
  list(APPEND MULTIPASS_BACKENDS apple)
#endif
```

like changes when we need to make a change required by a feature flag?

jimporter: Yeah, something like that. It should be roughly equivalent to how we'd do things *without* the feature flag (for the AppleVZ case):

```cmake
if(APPLE)
 list(APPEND MULTIPASS_BACKENDS apple)
endif()
```

For feature flags with no dependencies, e.g. the AZ feature flag, we'd have to add some new `if`/`endif` blocks, but no way around that I suppose...

CMakeLists.txt
jimporter: Yeah, our spec says we'll put these in `feature_flags.cmake` at the root of the tree. I just wanted to throw together a quick example here that I'll remove before merging this (though I suppose I could have done that in the proper file...)

src/cmake/feature-flag.cmake
```diff
@@ -0,0 +1,33 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+include(CMakeDependentOption)
+
+# FIXME: How should we force all the various feature flags to change their
+# cached values if we change this setting after the first configure?
+option(MP_DISABLE_ALL_FEATURES "disable all optional features" OFF)
```
sharder996: iirc we were going to lay this out with the logic inverted to what it is here. i.e. `MP_ENABLE_ALL_FEATURES` with a default value of `ON`.

CMakeLists.txt
sharder996: And then turn `MP_ENABLE_ALL_FEATURES` off somewhere around here to turn off feature flags for release builds.

sharder996: And, apply the `-noff` suffix after the hash.

src/cmake/feature-flag.cmake
jimporter: Based on discussion in the standup, this is now `MP_ALLOW_OPTIONAL_FEATURES`.

CMakeLists.txt
jimporter: Should we do this automatically based on the state of `MP_ALLOW_OPTIONAL_FEATURES`, or maybe just add a new option like `MP_VERSION_SUFFIX` that we can set in CI? The former is simpler, but wouldn't account for a build with *some* feature flags enabled and others disabled. (I'm not sure we want that just yet, but it could come in handy if we want to publish Snaps for testing a given feature.)

src/cmake/feature-flag.cmake
```diff
@@ -0,0 +1,32 @@
...
+if(GIT_IS_RELEASE_BRANCH)
+  set(OPTIONAL_FEATURE_DEFAULT OFF)
+else()
+  set(OPTIONAL_FEATURE_DEFAULT ON)
+endif()
+option(MP_ALLOW_OPTIONAL_FEATURES "allow enabling optional features" ${OPTIONAL_FEATURE_DEFAULT})
+
+macro(feature_flag option doc)
+  if(${ARGC} GREATER 2)
+    cmake_dependent_option(${option} "${doc}" ON "MP_ALLOW_OPTIONAL_FEATURES;${ARGV2}" OFF)
```
sharder996, line 39: Is it possible to make this override cached variables? For example, if I configure a build directory while on a release branch, `MP_ALLOW_OPTIONAL_FEATURES=OFF`, as expected. However if I reconfigure that build directory with my own development branch, `MP_ALLOW_OPTIONAL_FEATURES` won't switch on. I have to do it manually with `-DMP_ALLOW_OPTIONAL_FEATURES=ON`.

sharder996 (changes requested):

CMakeLists.txt
sharder996: I think it should correspond with the state of `MP_ALLOW_OPTIONAL_FEATURES` so that a generated package is in line with how cmake configured the build. See [my other comment ](https://github.com/canonical/multipass/pull/4582#discussion_r2629211632) for more context.

ricab: Good point @jimporter! How about `multipass_feature` requiring a short suffix when declaring the feature, then append all suffixes? It could become long with multiple features... :/ Maybe just join all enabled feature names and generate a 4 char hash or something?

jimporter: @sharder996 Ok, I've shuffled things around so that `versioning.cmake` can automatically set the `-noff` suffix for us.

@ricab I didn't add this extra bit, since after thinking it over, it's (hopefully) just a theoretical problem for now. While we might do local builds with certain optional features on or off, I don't think we'll be distributing them anywhere.

jimporter:
> @jimporter Did something change in your most recent rebase? I'm noticing some things that I think were here and are now missing. For example, `MP_DISABLE_ALL_FEATURES` doesn't seem to exist in the code and:

From some prior discussions (I think in meetings last year), I changed this variable to be `MP_ALLOW_OPTIONAL_FEATURES` so that the variable name didn't contain a negative.

> Also, I think the `-noff` part of the version is missing.

This should work if you pass `-DMP_ALLOW_OPTIONAL_FEATURES=OFF`.

> And, I still see the case where cached variables aren't being overwritten (as per my previous comment).

I'm not sure how to fix this part though. I'll do some more digging to see if there's a workaround, but CMake's documentation on this doesn't address this case from what I can see.

sharder996:
Thanks for the reminder 😄 last year's discussions are still a little fuzzy. Regardless, I think the PR description is/was outdated. I think that part looks good.

As for cached variables, it would be really nice if we could figure out how to override them. Otherwise, it would make it confusing when you tell cmake to do one thing and it does another. Summoning the cmake guru, @xmkg, to see if maybe he has some ideas.

CMakeLists.txt
ricab: How about just adding some suffix to make it clear that there is a custom selection of FF. That would allow us to be sure that two packages with the same version would be identical, unless they had that suffix.

Also, to confirm: when we release, FF are disabled but the package version wouldn't include the `-noff` suffix, would it?

jimporter:
> As for cached variables, it would be really nice if we could figure out how to override them. Otherwise, it would make it confusing when you tell cmake to do one thing and it does another. Summoning the cmake guru, @xmkg, to see if maybe he has some ideas.

I have something that I think works for this now. It makes it harder to set this in the CMake GUI, but I'm not sure if we can get that *and* selectively-cache the value like this version is doing now.

sharder996:
> > As for cached variables, it would be really nice if we could figure out how to override them. Otherwise, it would make it confusing when you tell cmake to do one thing and it does another. Summoning the cmake guru, @xmkg, to see if maybe he has some ideas.
> 
> I have something that I think works for this now. It makes it harder to set this in the CMake GUI, but I'm not sure if we can get that _and_ selectively-cache the value like this version is doing now.

Ok, I think this is on the right track now. A default build turns on optional features:
```
% cmake -Bbuild -H. -GNinja -DMULTIPASS_ENABLE_FLUTTER_GUI=OFF 
-- MP_ALLOW_OPTIONAL_FEATURES = ON
-- APPLEVZ_ENABLED = ON
-- AVAILABILITY_ZONES_ENABLED = ON
```

A subsequent re-configure turns all features off:
```
% cmake -Bbuild -H. -GNinja -DMULTIPASS_ENABLE_FLUTTER_GUI=OFF -DMP_ALLOW_OPTIONAL_FEATURES=OFF
-- MP_ALLOW_OPTIONAL_FEATURES = OFF
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

But, going back to a default build does not turn the feature back on:
```
% cmake -Bbuild -H. -GNinja -DMULTIPASS_ENABLE_FLUTTER_GUI=OFF
-- MP_ALLOW_OPTIONAL_FEATURES = OFF
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

jimporter:
> But, going back to a default build does not turn the feature back on:
> 
> ```
> % cmake -Bbuild -H. -GNinja -DMULTIPASS_ENABLE_FLUTTER_GUI=OFF
> -- MP_ALLOW_OPTIONAL_FEATURES = OFF
> -- APPLEVZ_ENABLED = OFF
> -- AVAILABILITY_ZONES_ENABLED = OFF
> ```

As far as I can tell from [the CMake documentation](https://cmake.org/cmake/help/book/mastering-cmake/chapter/CMake%20Cache.html) and how `option` is specified, this is the expected behavior here. If you specify any CMake `option` on the command line, and then remove it in a subsequent call to CMake, it'll retain whatever value you set it to. If you want it to have some other value (e.g. whatever the default is), you have to override the previously-cached value.

I'm not exactly a fan of this behavior and find CMake's cache variables a bit confusing from a user PoV, but this does seem like what the CMake maintainers intend to happen.

xmkg:
> As for cached variables, it would be really nice if we could figure out how to override them. Otherwise, it would make it confusing when you tell cmake to do one thing and it does another. Summoning the cmake guru, @xmkg, to see if maybe he has some ideas.

So the behavior you've described is exactly how cached variables supposed to behave, but that's not ideal for our use case. AFAICT we want to be able to:

a) Determine a default `MP_ALLOW_OPTIONAL_FEATURES` value based on the branch when not explicitly given by user
b) Always respect user's explicit overrides, regardless of  `MP_ALLOW_OPTIONAL_FEATURES` is already cached or not

Part a is already addressed, but part b is not. There are two ways to solve this:

1-) At build time:
- Adding [--fresh](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-fresh) to configure command (e.g. `cmake --fresh -S . -B build /*...*/`) (requires CMake 3.24+)
  - **Pros:** it works, **cons:** nukes **all** cached variables, you have to remember setting it
- Explicitly [undefine](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-U) the `MP_ALLOW_OPTIONAL_FEATURES` variable so the CMake code would re-evaluate it. (`cmake -UMP_ALLOW_OPTIONAL_FEATURES -S . -B build`)
  - **Pros:** it works, it's isolated **cons:** you have to remember setting it

2-) Use the `two variables` trick

User-exposed option is a separate CMake variable so we can determine if and when we should invalidate the `internal` one:

```cmake
if(DEFINED MP_ALLOW_OPTIONAL_FEATURES)
  # User is explicitly overriding -- use that value.
  set(MP_ALLOW_OPTIONAL_FEATURES_INTERNAL "${MP_ALLOW_OPTIONAL_FEATURES}" CACHE INTERNAL
        "allow optional features" FORCE)
  # The variables passed via -D are automatically cached, so it's up to us to remove
  # the transient variable from the cache
  unset(MP_ALLOW_OPTIONAL_FEATURES)
  unset(MP_ALLOW_OPTIONAL_FEATURES CACHE)
else()
    # Unset the previously cached variable as a good measure so we can re-determine its value.
    unset(MP_ALLOW_OPTIONAL_FEATURES_INTERNAL CACHE)
    # Otherwise, just temporarily set its value based on whether we're on a release branch. This way,
    # if we change branches, the value will automatically update.
    is_release_branch(GIT_IS_RELEASE_BRANCH)
    if(GIT_IS_RELEASE_BRANCH)
      set(MP_ALLOW_OPTIONAL_FEATURES_INTERNAL OFF)
    else()
      message(STATUS "in feature branch")
      set(MP_ALLOW_OPTIONAL_FEATURES_INTERNAL ON)
    endif()
endif()

macro(feature_flag option doc)
  if(${ARGC} GREATER 2)
    cmake_dependent_option(${option} "${doc}" ON "MP_ALLOW_OPTIONAL_FEATURES_INTERNAL;${ARGV2}" OFF)
  else()
    cmake_dependent_option(${option} "${doc}" ON MP_ALLOW_OPTIONAL_FEATURES_INTERNAL OFF)
  endif()
endmacro()
```

Passing an argument via `-D...` always creates a cached variable so the code also explicitly unset the user-defined arg after consumption.

@sharder996 with the patch above your initial example becomes:

**Initial configure:**
```bash
❯ cmake -S . -B build
-- in feature branch
-- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = ON
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = ON
```

**User override:**
```bash
❯ cmake -S . -B build -DMP_ALLOW_OPTIONAL_FEATURES=OFF
-- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = OFF
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

**No-arg reconfigure:**

```bash
❯ cmake -S . -B build
-- in feature branch
-- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = ON
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = ON
```

Let me know if the behavior above is what you had in mind.

sharder996:
@xmkg Typo? Both `APPLEVZ_ENABLED` and `AVAILABILITY_ZONES_ENABLED` should be ON.
> No-arg reconfigure:
> ```
> ❯ cmake -S . -B build
> -- in feature branch
> -- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = ON
> -- APPLEVZ_ENABLED = OFF
> -- AVAILABILITY_ZONES_ENABLED = ON
> ```
> Let me know if the behavior above is what you had in mind.

xmkg:
> @xmkg Typo? Both `APPLEVZ_ENABLED` and `AVAILABILITY_ZONES_ENABLED` should be ON.
> 
> > No-arg reconfigure:
> > ```
> > ❯ cmake -S . -B build
> > -- in feature branch
> > -- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = ON
> > -- APPLEVZ_ENABLED = OFF
> > -- AVAILABILITY_ZONES_ENABLED = ON
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let me know if the behavior above is what you had in mind.

I'm in Linux so it can't be enabled:

```cmake
feature_flag(APPLEVZ_ENABLED "AppleVZ backend" APPLE)
```
(Note the APPLE prerequisite)

sharder996:
@xmkg Only odd behavior I found with your patch is:
```
% cmake -Bbuild -H. -GNinja -DMP_ALLOW_OPTIONAL_FEATURES=OFF -DAPPLEVZ_ENABLED=ON
-- MP_ALLOW_OPTIONAL_FEATURES = ON
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

xmkg:
> cmake -Bbuild -H. -GNinja -DMP_ALLOW_OPTIONAL_FEATURES=OFF -DAPPLEVZ_ENABLED=ON

I can't reproduce it on my end:

```bash
❯ cmake -Bbuild -H. -DMP_ALLOW_OPTIONAL_FEATURES=OFF -DAPPLEVZ_ENABLED=ON
-- MP_ALLOW_OPTIONAL_FEATURES (before processing) = OFF
-- MP_ALLOW_OPTIONAL_FEATURES = 
-- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = OFF
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

You might need to nuke the build folder if the variable is already cached (from the previous runs with the older behavior).

Below is with ON for completeness:

```bash
❯ cmake -Bbuild -H. -DMP_ALLOW_OPTIONAL_FEATURES=ON -DAPPLEVZ_ENABLED=ON
-- MP_ALLOW_OPTIONAL_FEATURES (before processing) = ON
-- MP_ALLOW_OPTIONAL_FEATURES = 
-- MP_ALLOW_OPTIONAL_FEATURES_INTERNAL = ON
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = ON
```

sharder996:
Alright, up to @jimporter and other reviewers what they think. I know its not unheard of to have other weird behavior when re-configuring build directories, so we don't have to get too fancy.

xmkg:
> Alright, up to @jimporter and other reviewers what they think. I know its not unheard of to have other weird behavior when re-configuring build directories, so we don't have to get too fancy.

I mean, it should be consistent in behavior once you've done a clean build with the patch. I recall observing the same and the issue disappeared after configuring it from scratch. If it's flaky, let me know because that's an issue :)

sharder996:
@xmkg I'll use a different flag to remove the platform ambiguity.

```
% cmake -Bbuild -H. -GNinja -DMP_ALLOW_OPTIONAL_FEATURES=OFF -DAVAILABILITY_ZONES_ENABLED=ON
-- MP_ALLOW_OPTIONAL_FEATURES = 
-- APPLEVZ_ENABLED = OFF
-- AVAILABILITY_ZONES_ENABLED = OFF
```

I guess the surprising part to me is that `MP_ALLOW_OPTIONAL_FEATURES` overrides setting individual feature flags. For the case where I only want to enable one specific flag.

xmkg:
> @xmkg I'll use a different flag to remove the platform ambiguity.
> 
> ```
> % cmake -Bbuild -H. -GNinja -DMP_ALLOW_OPTIONAL_FEATURES=OFF -DAVAILABILITY_ZONES_ENABLED=ON
> -- MP_ALLOW_OPTIONAL_FEATURES = 
> -- APPLEVZ_ENABLED = OFF
> -- AVAILABILITY_ZONES_ENABLED = OFF
> ```
> 
> I guess the surprising part to me is that `MP_ALLOW_OPTIONAL_FEATURES` overrides setting individual feature flags. For the case where I only want to enable one specific flag.

To me, the name sounds like it's a prerequisite for all the optional flags, but I'm not sure if that was the intent. Maybe @jimporter can clarify that part.

sharder996:
@xmkg I agree with your point on semantics. I just don't see an easy way of only enabling one feature flag.

jimporter:
@sharder996 I don't have a strong preference here, so feel free to overrule me, but for caching variables, I like option (1) from @xmkg's comment (using `--fresh` or `-UMP_ALLOW_OPTIONAL_FEATURES`). As far as I understand it, that's consistent with how cache variables normally work in CMake, so resetting `MP_ALLOW_OPTIONAL_FEATURES` would work the same way as for resetting any *specific* feature flag, as well as any other `option` settings we have in CMake. I find it a little less intuitive than option (2), but I think being consistent with the rest of CMake is worth it.

> I guess the surprising part to me is that MP_ALLOW_OPTIONAL_FEATURES overrides setting individual feature flags. For the case where I only want to enable one specific flag.

Right. This is a "disable all" flag, as opposed to setting the default value for feature flags. Since the number of active feature flags will probably be small, and in most cases we probably want to enable/disable all features (or maybe occasionally disable a single feature), I think this covers the practical cases we care about. Having some `MP_DEFAULT_FEATURE_FLAG_STATE` variable instead would be a bit more flexible, though it would be more work to get that to play nicely with CMake cache variables; the current code just relies on `cmake_dependent_option` to automatically update each feature flag if you change the overall flag.

I don't mind the idea of changing these semantics, though I'm not sure it'll come up in practice to make it worth the extra complexity. In practice, I imagine we'll mostly avoid setting feature flags (including `MP_ALLOW_OPTIONAL_FEATURES`) manually when doing builds, and just rely on the branch-detection logic.

sharder996:
Ok, so the current scenarios that work are:
- all on
- all off
- all on and manually turn specific flags off

The scenario that I thought might be missing was:
- all off and manually turn specific flags on

Regardless, my approval still stands and I'm fine either way.

ricab:
Hey @jimporter, I would love an answer on my questions/suggestion above regarding versions. Other than that, you have two approvals so I'd skip review if that is OK with you.

CMakeLists.txt
jimporter: @ricab Ok, I added a `-custom` suffix any time the feature flag set doesn't match the defaults. It checks `${FEATURE}_AVAILABLE`, and if any feature is available but not enabled, it counts as a custom build. (Note that when disabling `MP_ALLOW_OPTIONAL_FEATURES`, all the features are marked unavailable, so that doesn't count as "custom".)

ricab: That sounds good, thanks a lot!

 To confirm, that means:

- no suffix in a proper release → no feature flags enabled
- no suffix in outside release (dev) → all feature flags enabled
- `-noff` suffix → no feature flags, outside release
- `custom` suffix → any other option. 

Is this correct?

jimporter: @ricab Close: release versions *never* get a suffix (so technically, you could make a custom release build that doesn't have the `custom` suffix). There might be some benefit in adding a suffix in non-default cases, but I don't think we'd actually *do* that (and I'm not 100% sure adding a suffix is the right thing to do in this hypothetical situation anyway).

ricab: OK, that makes sense to me too. We retain the ability to tweak a release as we choose, but FF are disabled by default. 

Do you think it's worth sticking a corrected version of the topics above in https://github.com/canonical/mp-dev-wiki?

jimporter: Sure, I can try and update that. It looks like there's still some documentation about the open-source vs "full" variants of Multipass in there, so that's probably the first thing to update.

sharder996 (dismissed):
LGTM!

Not sure what, if anything has changed since @xmkg's last review, but feel free to hit the merge button if its just a rebase.

jimporter:
> Not sure what, if anything has changed since @xmkg's last review, but feel free to hit the merge button if its just a rebase.

@sharder996 This commit is the only new bit: https://github.com/canonical/multipass/pull/4582/commits/8af9c88df2e37b8a2c221a624d7c369dc463e4cb

sharder996:
@jimporter I noticed an issue with snap packages. Snapcraft only allow a version length of [32 characters](https://github.com/canonical/multipass/actions/runs/20865330281/job/59955099685?pr=4570#step:24:2616). The extra suffixes are pushing it over that limit. We might have to do something similar to what we do in Windows where the package version is slightly truncated, but the internal version as reported by multipassd is the full version.

jimporter:
@sharder996 Good catch! What do you think about this latest change? I adjusted how we get version info for Snapcraft by making a `version-metadata.env` file that we can use to retrieve the version when any other tools need it during the build process. That also means the `snapcraft.yaml` no longer relies on how we supply the version ID to our C++ code, which seems a bit cleaner to me.

sharder996:
Looks good! I like that you were able to combine it with the GUI.

One other thing, though 😄 When I define a feature flag, say `APPLEVZ_ENABLED`, it is merely defined as a CMake variable. I can't use it as a preprocessor definition in C++ code to enabled bits of code.

sharder996 (dismissed):
Just to leave a trail of what I tested. A snap from a PR was like this (this is the version string that is 32 characters long and is potentially causing problems):

```
1.17.0-dev.1372.pr4571+g84a3756b
```

and is now:

```
1.17.0-dev.1375.pr4582
```

I think this looks fine as `pr####` tells where the snap package came from. And a snap package from a particular commit was:

```
1.17.0-dev.1367+ge6c32397
```

and is now:

```
1.17.0-dev.1377
```

I usually don't like reducing the amount of information published, but since this is similar to what we do on Windows and that the full version string is still reported by the multipass daemon, I think I'm fine with this.

xmkg (dismissed):
Rubber-stamping again, this still LGTM.

ricab:
Hmm, reducing the version to `1.17.0-dev.1375.pr4582` means that there are multiple commits that would have the same version. I think we should maintain the invariant that any 2 packages with the same version are actually the same (on a given platform, and with the exception of the `-custom` suffix). 

Can't we instead just shave 2 or 3 characters off the hash?

ricab:
Otherwise, we're not just loosing information, we're loosing the property of a version string actually *identifying* a version.

sharder996:
> Hmm, reducing the version to `1.17.0-dev.1375.pr4582` means that there are multiple commits that would have the same version. I think we should maintain the invariant that any 2 packages with the same version are actually the same (on a given platform, and with the exception of the `-custom` suffix).
> 
> Can't we instead just shave 2 or 3 characters off the hash?

I think technically we don't need to change the version string at all at this time. We just need to make sure that any feature flag related suffixes aren't appended. We will however exceed the snap version character limit when we reach 10,000 PRs, I guess.

ricab:
How about dropping the number of commits since the tag, i.e., `1372` in `1.17.0-dev.1372.pr4571+g84a3756b`? That plus a reduction of 3 characters in the hash would save 8 characters: enough for `-noff` and even `-custom` + 5-digit PRs, right? For example: `1.17.0-dev.pr94571+g84a37-custom`. 

We could also remove the `g` instead of one of the hash characters. And `-custom` could become `-custm`...

jimporter:
@ricab With this latest change, custom builds are tagged with `-cstm` and the "number of commits since tag" is removed whenever `MULTIPASS_BUILD_LABEL` is set. That means that builds on the main branch have the same version ID as before (with 8-character GIT SHAs), and PR builds have a version like `1.17.0-dev.pr1234+g7d6db065-noff`.

I also added some more comments to the code to explain all the subtleties here, in case we need to revisit this in the future. What do you think?

edit: You can see the version in the CI run [here](https://github.com/canonical/multipass/actions/runs/21011750216/job/60408143371?pr=4582#step:13:1133).

xmkg:
> @ricab With this latest change, custom builds are tagged with `-cstm` and the "number of commits since tag" is removed whenever `MULTIPASS_BUILD_LABEL` is set. That means that builds on the main branch have the same version ID as before (with 8-character GIT SHAs), and PR builds have a version like `1.17.0-dev.pr1234+g7d6db065-noff`.
> 
> I also added some more comments to the code to explain all the subtleties here, in case we need to revisit this in the future. What do you think?
> 
> edit: You can see the version in the CI run [here](https://github.com/canonical/multipass/actions/runs/21011750216/job/60408143371?pr=4582#step:13:1133).

I think we can also ditch the -dev tag for pr-built snaps, pr already implies dev.

jimporter:
> I think we can also ditch the -dev tag for pr-built snaps, pr already implies dev.

@xmkg I'm open to others' thoughts here, but for what it's worth, I kept the `-dev` part because that part comes from the actual tag in our repo. I figured it would be clearest if the leading portion of the version ID always exactly matched the tag that the version is based on.

jimporter:
> I don't see the `-noff` suffix in the version that's printed in CI. I imagine that's expected?

Yep. I believe @sharder996's CI changes based on this PR will produce `-noff` builds though.

jimporter:
Merging now, since everyone approved and I got some extra thumbs-ups on the remaining discussions. Of course, we can always file a followup with more changes if we run into some issue in the future.

PR #4583, opened 2025-12-17
[docs] Remove `alias` and `host` from explanation

geoffreynyaga:
# Description

This PR removes two pages from the explanation section. The pages are `Alias` and `Host` and they lack a good threshold of content to warrant their own independent pages on the Multipass docs.

The PR also updates any internal linking to these removed pages to prevent broken links.

## Checklist

### DIFF
--- a/docs/how-to-guides/install-multipass.md
+++ b/docs/how-to-guides/install-multipass.md
@@ -17,7 +17,7 @@ Select the tab corresponding to your operating system (e.g. Linux) to display th
 
 ````{tab-item} Linux
 
-Multipass for Linux is published as a [snap package](https://snapcraft.io/docs/), available on the [Snap Store](https://snapcraft.io/multipass). Before you can use it, you need to [install `snapd`](https://docs.snapcraft.io/core/install). `snapd` is included in Ubuntu by default.
+Multipass for Linux is published as a [snap package](https://snapcraft.io/docs/), available on the [Snap Store](https://snapcraft.io/multipass). Before you can use it, you need to [install `snapd`](https://snapcraft.io/docs/tutorials/install-the-daemon/). `snapd` is included in Ubuntu by default.
 
 ````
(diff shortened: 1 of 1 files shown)


geoffreynyaga:
> Hey Geoffrey, I understand where you're coming from, but I am a little hesitant to just remove valid content. Have you considered a page for miscellaneous concepts that don't deserve their own pages, perhaps?


@ricab,
An approach we can use is utilize the [glossary](https://canonical-starter-pack.readthedocs-hosted.com/stable/reference/myst-syntax-reference/#glossary) MyST feature to list such definitions on the landing page (as glossary items). As their scope and documentation increases, then we can move them to their individual pages.
An example would be:

Screenshot 2025-12-17 at 16 52 22

ricab:
Ah perfect, that makes total sense to me. Do you think that should go in this or a separate PR?

geoffreynyaga:
> Ah perfect, that makes total sense to me. Do you think that should go in this or a separate PR?

It is OK to do this in the same PR since a separate PR will still follow the process of deleting the pages, unlinking old references and writing the glossary items. The only step we will have to re-do in this PR is to put back the internal refs that we had deleted and update them with the new reference.

ricab:
Oh, we've discussed this :sweat_smile:

Do you still mean to do that glossary thing @geoffreynyaga?

geoffreynyaga:
This is now the final landing page outline. Instead of deleting the `alias` and `host` explanation pages (since they were each effectively one sentence long), we extracted the content into a glossary section in the landing page. Since we were using MyST cross-reference syntax e.g.
```
(explanation-host)=
### Host
```
the links do not break and we are not required to set redirects.

Screenshot 2026-03-20 at 22 40 25

tobe2098:
I would go ahead with the merge if @ricab is ok with the changes

PR #4594, opened 2026-01-01
[ci] Replace ccache with sccache

sharder996:
This PR adds support for and replaces the use of `ccache` with `sccache` in Windows/macOS CI.

This provides several benefits such Windows support and the additional ability to also cache Rust builds.

The control of cache dumps in GitHub cache is also improved so that caches are only published when a PR has been pushed to `main`, usually by `github-merge-queue`. A cache for a PR is selected as the cache that was published from `main` where the current branch originated from, and if not found, the next cache matching against the `restore-keys` field.

Documentation for `sccache` can be found [here](https://github.com/mozilla/sccache).
Documentation for the `sccache` GitHub action can be found [here](https://github.com/marketplace/actions/sccache-action).

### DIFF
--- a/.github/workflows/windows-macos.yml
+++ b/.github/workflows/windows-macos.yml
@@ -240,6 +240,7 @@ jobs:
 
     - name: Run sccache-cache
       uses: mozilla-actions/sccache-action@v0.0.9
+      continue-on-error: true
 
     - name: Install Windows ADK
       if: ${{ runner.os == 'Windows' }}


CMakeLists.txt
```diff
@@ -51,9 +51,17 @@ else()
...
-# Use ccache if it's installed
+# Use caching option if it's installed
+find_program(SCCACHE_PROGRAM sccache)
 find_program(CCACHE_PROGRAM ccache)
-if(CCACHE_PROGRAM)
+if(SCCACHE_PROGRAM)
+  set(CMAKE_CXX_COMPILER_LAUNCHER "${SCCACHE_PROGRAM}")
+  set(CMAKE_C_COMPILER_LAUNCHER "${SCCACHE_PROGRAM}")
+  if("${HOST_OS_NAME}" STREQUAL "Windows")
+    set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>")
```
xmkg, line 61: Is there a reason why this is only needed when sccache is present?

xmkg (commented):
LGTM, except for a small question inline.

CMakeLists.txt
sharder996: From https://github.com/mozilla/sccache?tab=readme-ov-file#usage:

> For versions of cmake 3.25 and later, to compile with MSVC, you have to use the new CMAKE_MSVC_DEBUG_INFORMATION_FORMAT option, meant to configure the -Z7 flag. Additionally, you must set the cmake policy number 0141 to the NEW setting:
> 
> ```
> set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>")
> cmake_policy(SET CMP0141 NEW)
> ```
> 
> Example configuration where we automatically look for sccache in the PATH:
> 
> ```
> find_program(SCCACHE sccache REQUIRED)
> 
> set(CMAKE_C_COMPILER_LAUNCHER ${SCCACHE})
> set(CMAKE_CXX_COMPILER_LAUNCHER ${SCCACHE})
> set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>")
> cmake_policy(SET CMP0141 NEW)
> ```

tobe2098 (dismissed):
LGTM. Good work Scott!

tobe2098 (approved):

PR #4596, opened 2026-01-05
[ci] Use different action to avoid rate limiting

sharder996:
Switching out the GitHub action used for deleting old package version as cached by vcpkg. The previous action, `actions/delete-package-versions`, can only handle one package at a time, so a matrix was used to apply the action to all the packages that Multipass uses. GitHub matrices, however, only support up to 256 entries and with moving more dependencies to vcpkg, we are already hitting this limits.

This PR instead uses the `smartsquaregmbh/delete-old-packages` action which takes a list of packages and apply the operation to all of them at once.

This may hit GitHub API rate limiting, but running the workflow again will pick where the last run failed.

### DIFF
--- a/.github/workflows/prune-packages.yml
+++ b/.github/workflows/prune-packages.yml
@@ -58,7 +58,7 @@ jobs:
           organization: ${{ github.repository_owner }}
           type: nuget
           names: ${{ needs.list-packages.outputs.names }}
-          keep: 1
+          keep: 2
 
   notify:
     runs-on: ubuntu-latest


.github/workflows/prune-packages.yml
```diff
@@ -50,18 +51,14 @@ jobs:
...
+      - name: Delete old package versions
+        uses: smartsquaregmbh/delete-old-packages@v0.8.1
         with:
-          package-name: ${{ matrix.package-name }}
-          package-type: nuget
-          min-versions-to-keep: 1
+          organization: ${{ github.repository_owner }}
+          type: nuget
+          names: ${{ needs.list-packages.outputs.names }}
+          keep: 2
```
ricab, line 61: This would double the packages we cache, right (with respect to `min-versions-to-keep: 1`)? Is that intentional?

ricab (commented):
Nice! This makes sense to me and the code looks good. 

Although the action is not hugely popular (71 stars now), it looks pretty sane to me and the workflow already restricts permissions as much as possible. We could pin a specific sha instead of a version, to avoid tags moving from under us, but this is not special in that regard.

I have a question inline and I would ask for evidence that this is working properly. Could you provide a link to a successful run please? Did you try `dry-run` first?

.github/workflows/prune-packages.yml
sharder996: I was thinking of the situational where we upgrade a dependency. It usually takes a little bit for all current work to be rebased, so for a period of time two versions of a dependency are in use.

sharder996:
@ricab Yes, this action is not as relatively popular, but I really didn't want to have to manually call the GitHub API through a script.


I did a couple of test runs. The first failed due to GitHub rate limiting as it was catching up with deleting old packages. It is available [here](https://github.com/canonical/multipass/actions/runs/20732448374). I manually ran the workflow again which successfully cleaned up the remaining packages. It is available [here](https://github.com/canonical/multipass/actions/runs/20732491750).

.github/workflows/prune-packages.yml
ricab: Good point, +1

ricab:
> @ricab Yes, this action is not as relatively popular, but I really didn't want to have to manually call the GitHub API through a script.

Right, I didn't mean to suggest anything like that. 

> I did a couple of test runs. The first failed due to GitHub rate limiting as it was catching up with deleting old packages. It is available [here](https://github.com/canonical/multipass/actions/runs/20732448374). I manually ran the workflow again which successfully cleaned up the remaining packages. It is available [here](https://github.com/canonical/multipass/actions/runs/20732491750).

OK, thank you.

xmkg (approved):
LGTM, thanks Scott!

PR #4597, opened 2026-01-06
[snapshot] Store metadata using Boost.JSON

jimporter:
# Description

This updates the VM metadata objects in our code to use Boost.JSON instead of Qt's JSON. Since not everything has been converted to use Boost.JSON yet, this temporarily relies on converting back and forth between Boost.JSON and Qt JSON as needed.

## Testing

- All existing unit tests have been updated to account for these changes.

## Checklist

### DIFF
--- a/include/multipass/json_utils.h
+++ b/include/multipass/json_utils.h
@@ -48,16 +48,17 @@ public:
     virtual QJsonValue update_cloud_init_instance_id(const QJsonValue& id,
                                                      const std::string& src_vm_name,
                                                      const std::string& dest_vm_name) const;
-    virtual QJsonValue update_unique_identifiers_of_metadata(const QJsonValue& metadata,
-                                                             const multipass::VMSpecs& src_specs,
-                                                             const multipass::VMSpecs& dest_specs,
-                                                             const std::string& src_vm_name,
-                                                             const std::string& dest_vm_name) const;
     virtual QJsonArray extra_interfaces_to_json_array(
         const std::vector<NetworkInterface>& extra_interfaces) const;
     virtual std::optional<std::vector<NetworkInterface>> read_extra_interfaces(
         const QJsonObject& record) const;
 };
+
+boost::json::object update_unique_identifiers_of_metadata(const boost::json::object& metadata,
+                                                          const multipass::VMSpecs& src_specs,
+                                                          const multipass::VMSpecs& dest_specs,
+                                                          const std::string& src_vm_name,
+                                                          const std::string& dest_vm_name);
 
 namespace detail
 {
--- a/include/multipass/snapshot.h
+++ b/include/multipass/snapshot.h
@@ -24,8 +24,11 @@
 #include <string>
 #include <unordered_map>
 
-class QJsonObject;
 class QDateTime;
+namespace boost::json
+{
+class object;
+}
 
 namespace multipass
 {
@@ -50,7 +53,7 @@ public:
 
     // Note that these return references - careful not to delete the snapshot while they are in use
     virtual const std::unordered_map<std::string, VMMount>& get_mounts() const noexcept = 0;
-    virtual const QJsonObject& get_metadata() const noexcept = 0;
+    virtual const boost::json::object& get_metadata() const noexcept = 0;
 
     virtual std::shared_ptr<const Snapshot> get_parent() const = 0;
     virtual std::shared_ptr<Snapshot> get_parent() = 0;
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -305,7 +305,7 @@ std::unordered_map<std::string, mp::VMSpecs> load_db(const mp::Path& data_path,
             static_cast<mp::VirtualMachine::State>(state),
             mounts,
             deleted,
-            metadata,
+            mp::qjson_to_boost_json(metadata).as_object(),
             clone_count};
     }
     return reconstructed_records;
@@ -320,7 +320,7 @@ QJsonObject vm_spec_to_json(const mp::VMSpecs& specs)
     json.insert("ssh_username", QString::fromStdString(specs.ssh_username));
     json.insert("state", static_cast<int>(specs.state));
     json.insert("deleted", specs.deleted);
-    json.insert("metadata", specs.metadata);
+    json.insert("metadata", mp::boost_json_to_qjson(specs.metadata));
 
     // Write the networking information. Write first a field "mac_addr" containing the MAC address
     // of the default network interface. Then, write all the information about the rest of the
@@ -3084,14 +3084,14 @@ void mp::Daemon::persist_state_for(const std::string& name, const VirtualMachine
     persist_instances();
 }
 
-void mp::Daemon::update_metadata_for(const std::string& name, const QJsonObject& metadata)
+void mp::Daemon::update_metadata_for(const std::string& name, const boost::json::object& metadata)
 {
     vm_instance_specs[name].metadata = metadata;
 
     persist_instances();
 }
 
-QJsonObject mp::Daemon::retrieve_metadata_for(const std::string& name)
+boost::json::object mp::Daemon::retrieve_metadata_for(const std::string& name)
 {
     return vm_instance_specs[name].metadata;
 }
@@ -3204,7 +3204,7 @@ void mp::Daemon::create_vm(const CreateRequest* request,
                                            VirtualMachine::State::off,
                                            {},
                                            false,
-                                           QJsonObject()};
+                                           {}};
                 operative_instances[name] =
                     config->factory->create_virtual_machine(vm_desc,
                                                             *config->ssh_key_provider,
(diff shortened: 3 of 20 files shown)


src/utils/json_utils.cpp
```diff
@@ -181,6 +154,38 @@ std::optional<std::vector<mp::NetworkInterface>> mp::JsonUtils::read_extra_inter
...
+            if (!src_mac.empty())
+            {
+                const std::string& dest_mac = dest_specs.extra_interfaces[i].mac_address;
+                replace_substring(str, src_mac, dest_mac);
+            }
+        }
+        // string replacement is "instances/<src_name>"->"instances/<dest_name>" instead of
+        // "<src_name>"->"<dest_name>", because the second one might match other substrings of
+        // the metadata.
+        replace_substring(str, "instances/" + src_vm_name, "instances/" + dest_vm_name);
```
ricab: How about using [`boost::replace_all`](https://www.boost.org/doc/libs/latest/doc/html/doxygen/replace_8hpp_1a575fad69199b22609797adca04da4342.html) or [`boost::replace_first`](https://www.boost.org/doc/libs/latest/doc/html/doxygen/replace_8hpp_1a1e8bf05f74de4164e5d8c0d7b919cbe9.html)?

Or, just [`std::regex_replace`](https://en.cppreference.com/w/cpp/regex/regex_replace.html) if you'd prefer to avoid depending on boost for that.

include/multipass/json_utils.h
```diff
@@ -48,17 +48,18 @@ class JsonUtils : public Singleton<JsonUtils>
...
         const std::vector<NetworkInterface>& extra_interfaces) const;
     virtual std::optional<std::vector<NetworkInterface>> read_extra_interfaces(
         const QJsonObject& record) const;
 };
 
+boost::json::object update_unique_identifiers_of_metadata(const boost::json::object& metadata,
+                                                          const multipass::VMSpecs& src_specs,
+                                                          const multipass::VMSpecs& dest_specs,
+                                                          const std::string& src_vm_name,
+                                                          const std::string& dest_vm_name);
```
ricab, line 61: Why place this out of the `JsonUtils` class? Can't mock it that way.

include/multipass/snapshot.h
```diff
@@ -26,6 +26,10 @@
 
 class QJsonObject;
```
ricab: Can't this be dropped now?

src/daemon/daemon.cpp
```diff
@@ -305,7 +305,7 @@ std::unordered_map<std::string, mp::VMSpecs> load_db(const mp::Path& data_path,
             static_cast<mp::VirtualMachine::State>(state),
             mounts,
             deleted,
-            metadata,
+            mp::qjson_to_boost_json(metadata).as_object(),
```
ricab, line 308: This is just so you don't have to propagate boost-json everywhere for now, correct? It will be removed eventually, right? Just to confirm my understanding.

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -62,50 +63,43 @@ constexpr auto mount_arguments_key = "arguments";
...
-    if (metadata.contains(arguments_key) && metadata[arguments_key].type() == QJsonValue::Array)
-    {
-        auto array = metadata[arguments_key].toArray();
-        for (const QJsonValueRef val : array)
-        {
-            args.push_back(val.toString());
-        }
-    }
-    return args;
+    return mp::lookup_or<QStringList>(metadata, arguments_key, {});
```
ricab, line 73: Nice 🏅

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -152,12 +146,12 @@ auto hmc_to_qmp_json(const QString& command_line)
     return QJsonDocument(qmp).toJson();
 }
 
-auto get_qemu_machine_type(const QStringList& platform_args)
+std::string get_qemu_machine_type(const QStringList& platform_args)
```
ricab, line 149: 👍

src/platform/backends/qemu/qemu_virtual_machine.cpp
```diff
@@ -176,19 +170,17 @@ auto get_qemu_machine_type(const QStringList& platform_args)
 
     auto vmstate = QJsonDocument::fromJson(dump_file.readAll()).object();
 
-    auto machine_type = vmstate["vmschkmachine"].toObject()["Name"].toString();
-    return machine_type;
+    return vmstate["vmschkmachine"].toObject()["Name"].toString().toStdString();
```
ricab, line 173: yep, thanks

src/platform/backends/shared/base_snapshot.h
```diff
@@ -124,9 +126,9 @@ class BaseSnapshot : public Snapshot
         extra_interfaces;              // NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members)
     const VirtualMachine::State state; // NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members)
```
ricab: If you feel generous, these two could use a space for alignment.

src/utils/json_utils.cpp
```diff
@@ -77,6 +77,17 @@ void pretty_print_scalar(std::ostream& os, const boost::json::value& value)
...
+std::string& replace_substring(std::string& str, std::string_view from, std::string_view to)
+{
+    std::size_t pos = 0;
+    while ((pos = str.find(from, pos)) != std::string::npos)
+    {
+        str.replace(pos, from.size(), to);
+        pos += to.size();
+    }
+    return str;
+}
```
ricab: This must have been implemented millions of times. Let's just use an existing facility?

tests/qemu/test_qemu_backend.cpp
```diff
@@ -865,6 +866,7 @@ TEST_F(QemuBackend, verifyQemuArgumentsFromMetadataAreUsed)
     ASSERT_TRUE(qemu != processes.cend());
     EXPECT_TRUE(qemu->arguments.contains("-hi_there"));
     EXPECT_TRUE(qemu->arguments.contains("-hows_it_going"));
+    EXPECT_TRUE(qemu->arguments.contains("-mount_arg"));
```
ricab, line 869: 👍

src/utils/json_utils.cpp
jimporter: Good point. When I wrote that, I was pretty deep in the weeds, and didn't take any time to look for a Boost replacement. But part of the point of adopting Boost IMO is to avoid having to write so many little generic functions like this, so we should just use `boost::replace_all` here.

include/multipass/json_utils.h
jimporter: Maybe this is something to discuss with the team, but one of my goals for our mocks is to keep mocks around only for things that interact with some other part of the system, like I/O (file, network, etc), clocks, virtualization code, etc. This is just a simple pure function, so I don't think we actually need to mock it. By the end of my stack of changes for Boost.JSON, `JsonUtils` will go away entirely, since all the code will be pure functions or will have been moved elsewhere (e.g. `JsonUtils::write_json` is now `FileOps::write_transactionally`).

src/daemon/daemon.cpp
jimporter: Correct. This was really just to keep each PR for this project small and easy to review. This particular line will get removed in the next PR in my stack, when I convert `VMSpecs` over.

include/multipass/snapshot.h
jimporter: Good catch!

src/platform/backends/shared/base_snapshot.h
jimporter: I'll do one better and wrap all of these with `NOLINTBEGIN`/`NOLINTEND`. :)

src/utils/json_utils.cpp
ricab: Looks good, thanks!

src/daemon/daemon.cpp
ricab: Makes sense :+1:

src/platform/backends/shared/base_snapshot.h
ricab: Ah yes, better.

include/multipass/json_utils.h
ricab: Here I wasn't suggesting that you implement a mock, only that it be "mockable". But I am not going to block on that if it is going away. It's easy enough to add anyway.

I would love to hear you expand on and discuss those intentions in general, but that's for another time.

jimporter:
> Small process thing for future PRs: [RVW9](https://github.com/canonical/multipass/blob/main/CONTRIBUTING.md#reviews-rvw) recommends acknowledging review comments but leaving resolution to the reviewer. No worries here, just a kind reminder going forward 🙂

Got it. I'm used to it being the opposite way around from my last job, but I think this way makes more sense, really.

sharder996 (approved):
No additional comments from me. Looks good!

PR #4587, opened 2025-12-18
[gui] Use os field to determine icon

sharder996:
Noticed the terminal tab in the GUI wasn't displaying the correct icon. This fix correctly references the `os` field to determine the correct icon to display.

PR #4598, opened 2026-01-06
[ci] Make sure snapd in a proper state to use lxd

tobe2098:
Should fix the apparent issue found in #4587 CI runs for linux.


sharder996:
@tobe2098 Is there documentation on the environment variable you're using?

tobe2098:
@sharder996 no, I was trying a number of LLM suggestions after hitting a wall, and after ripgrepping the snapcraft codebase it seems like it was hallucinated. The proper running of the CI pipeline seems to have been a coincidence, it looks like a race condition (in some cases Debug fails while Release succeeds or vice versa).

tobe2098:
Re-running your PR's jobs seems to work now, so it probably was an issue elsewhere

PR #4600, opened 2026-01-07
[rust] Petname implementation and rust infrastructure

tobe2098:
# Description

- What does this PR do? This PR introduces all necessary CMake, snapcraft and CI infrastructure for Rust in Multipass. Additionally this PR also moves the petname functionality and some tests under Rust, and adds a small helper crate for macros. 
- Why is this change needed? This is a necessary first step for the development of block storage in Rust and a critical item in the introduction of Rust into the project.

## Checklist

MULTI-2452


rxx/Cargo.lock
ricab, line 1: I was taken aback when I saw the LoC stats on this PR, but relieved when I realized most of that is just this lock file. If we discount that and the one-off code for the fist Rust module, added and removed lines should be in the same ballpark, no?

tobe2098: If by the one-off code for the fist Rust module you mean the CMakeLists code and cargo in rxx/, it would be on the same ballpark, slightly higher because we needed to keep some of the boilerplate for mocking petname.

sharder996:
Hey @tobe, would you be able to do a rebase on this so I can test it with s390x/ppc64le?

tobe:
> Hey @tobe, would you be able to do a rebase on this so I can test it with s390x/ppc64le?

wrong tobe :^)

sharder996:
> > Hey @tobe, would you be able to do a rebase on this so I can test it with s390x/ppc64le?
> 
> wrong tobe :^)

Sorry!!! 😆 

Original message meant for @tobe2098!

sharder996:
```
error[E0308]: mismatched types
  --> petname/src/petname_generator.rs:78:56
   |
78 |         _ => return Err(PetnameError::InvalidSeparator(separator_i8)),
   |                         ------------------------------ ^^^^^^^^^^^^ expected `i8`, found `u8`
   |                         |
   |                         arguments to this enum variant are incorrect
   |
note: tuple variant defined here
  --> petname/src/petname_error.rs:25:5
   |
25 |     InvalidSeparator(i8),
   |     ^^^^^^^^^^^^^^^^
help: you can convert a `u8` to an `i8` and panic if the converted value doesn't fit
   |
78 |         _ => return Err(PetnameError::InvalidSeparator(separator_i8.try_into().unwrap())),
   |                                                                    ++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `petname` (lib) due to 1 previous error
```

tobe2098:
A very curious error... It does not occur in any build, is this s390x? The c_char is supposed to be i8, not u8. I will find a workaround.

Fixed

rxx/CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+  OUTPUT ${CARGO_GENERATED_FILES}
+  COMMAND ${CARGO_CMD} build
+  --manifest-path ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml
+  --workspace
+  ${rust_build_flag}
+  --lib
+  --features "${FEATURES_CSV}"
+  WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+  COMMENT "Building Rust crates and generating C++ glue..."
+)
```
jimporter: I believe that we can use [`CARGO_TARGET_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html) somewhere around here to set the Rust build directory to be somewhere inside the build dir that CMake uses. Then you wouldn't need the new `.gitignore` entry. Putting the Rust build artifacts in the CMake build dir will be important for feature flags, since we'll want the ability to have multiple build dirs locally, one for optional features enabled and one for them *disabled*.

BUILD.linux.md
```diff
@@ -10,6 +10,13 @@ sudo apt install devscripts equivs
 mk-build-deps -s sudo -i
 ```
 
+### Install the Rust compiler
+
+Go here for the most recent method. The current method is:
+```
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+```
```
jimporter: Ditto here about using APT/Snap. (Though we could recommend this as a fallback option, e.g. for other distros, preferably with a link to the Rust docs this comes from.)

tobe2098:
The idea behind the last two commits is to simplify the code and move the argument checks to compile-time instead of checking at runtime, since the arguments are always compile-time constants.

tobe2098:
Additionally, the change in naming from "petname" module to "namegen" module is to accommodate additional name generators like the one needed in the future block storage feature. Other name suggestions are welcome.

tobe2098:
Rebased for the CI split between macos and windows

.github/workflows/linux.yml
```diff
@@ -294,11 +295,15 @@ jobs:
...
         /snap/bin/lxc --project snapcraft file pull \
-          -p -r "$instance_name/root/parts/multipass/build/bin/multipass_tests" /tmp/coredump/multipass_tests
+          -p -r "$instance_name/root/parts/multipass/build/bin/multipass_cpp_tests" /tmp/coredump/multipass_cpp_tests
         echo "Pulled the executable."
         # Pull the coredump folder
         /snap/bin/lxc --project snapcraft file pull -p -r "$instance_name/coredump" /tmp/coredump
         echo "Pulled the coredumps folder."
+        mapfile -t BINS < <(cargo test --workspace --no-run --message-format=json | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable')
+        export TEST_BINARIES="${BINS[*]}"
+        for bin $TEST_BINARIES; do /snap/bin/lxc --project snapcraft file pull -p -r "$instance_name/$bin" /tmp/coredump/$bin
```
ricab: Doesn't this need an `in`? `for bin in $TEST_BINARIES`? And why the intermediate export? How about something like (untested):

```bash
mapfile -t test_bins < <(
  cargo test --workspace --no-run --message-format=json |
    jq -r 'select(.reason=="compiler-artifact" and .executable!=null) | .executable'
)

for bin in "${test_bins[@]}"; do
  /snap/bin/lxc --project snapcraft file pull -p -r "$instance_name/$bin" "/tmp/coredump/$bin"
done
```

rxx/namegen/src/petname_error.rs
```diff
@@ -0,0 +1,37 @@
...
+    InvalidWordNumber(i32),
+    InvalidSeparator(i8),
+}
+
+impl fmt::Display for PetnameError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PetnameError::InvalidWordNumber(num) => write!(f, "Invalid word number: {}", num),
+            PetnameError::InvalidSeparator(sep) => {
+                write!(f, "Invalid separator, ASCII code: {}", sep)
```
ricab, line 31: What happens if a u8 character with more bytes is passed? I imagine it would not compile due to the `c_char` param type in `make_petname`, correct?

rxx/namegen/src/petname_generator.rs
```diff
@@ -0,0 +1,66 @@
...
+sa::const_assert!(!NOUNS.is_empty());
+sa::const_assert!(!ADJECTIVES.is_empty());
+sa::const_assert!(!ADVERBS.is_empty());
+
+pub fn make_petname(num_words: NumWords, separator_8: c_char) -> Result<String, PetnameError> {
+    let separator_c = separator_8 as u8 as char;
+    let separator_c: char = match separator_c {
+        '-' | '_' => separator_c,
+        _ => return Err(PetnameError::InvalidSeparator(separator_8 as i8)),
+    };
```
ricab: Up to you, but I think an if would be cleaner in this case, since it would not need the reassign (untested):


```suggestion
    if !matches!(separator_c, '-' | '_') {
        return Err(PetnameError::InvalidSeparator(separator_8 as i8));
    }
```

rxx/namegen/src/petname_generator.rs
```diff
@@ -0,0 +1,66 @@
...
+        NumWords::Three => &[ADVERBS, ADJECTIVES, NOUNS],
+        num => return Err(PetnameError::InvalidWordNumber(num.repr)),
+    };
+
+    let mut rng_engine = rand::rng();
+
+    let words: Vec<_> = sources
+        .iter()
+        .map(|&arr| choose_from_str_array(arr, &mut rng_engine))
+        .collect();
```
ricab, line 53: Crafty 👍

rxx/namegen/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Antoni Bertolin Monferrer <antoni.monferrer@canonical.com>
```
ricab: We've long ago decided to remove this author lines. We all end up authoring the same files and it doesn't make sense to keep track of a bunch of authors in their contents. Git does that.

rxx/namegen/build.rs
```diff
@@ -13,16 +13,12 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Authored by: Alberto Aguirre <alberto.aguirre@canonical.com>
+ * Authored by: Antoni Bertolin Monferrer <antoni.monferrer@canonical.com>
```
ricab: Can we just remove this instead? I thought we'd removed all of these, but apparently not...

Your authorship is appreciated, preserved, and published via git 😉

rxx/namegen/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+# Authored by: Antoni Bertolin Monferrer <antoni.monferrer@canonical.com>
+
+[package]
+name = "namegen"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lib]
+crate-type = ["staticlib","rlib"]
```
ricab, line 22: Is the rlib needed today? Otherwise, is it worth declaring now (rather then when needed)? I am not saying it isn't, I would just like to hear your rationale.

rxx/namegen/build.rs
```diff
@@ -13,16 +13,12 @@
...
-
-namespace mp = multipass;
-
-mp::NameGenerator::UPtr mp::make_default_name_generator()
-{
-    return std::make_unique<mp::Petname>(mp::Petname::NumWords::TWO, "-");
+fn main() {
+    // Generate the .h and .cc files (_do not_ compile them)
+    let _ = cxx_build::bridge("src/lib.rs"); // drops the builder, just codegen
+    println!("cargo:rerun-if-changed=src/lib.rs");
```
ricab, line 21: Interesting. For my understanding, where is this run?

snap/snapcraft.yaml
```diff
@@ -211,6 +211,13 @@ parts:
     override-pull: |
       set -e
 
+      if [ ! -d "$HOME/.cargo" ]; then
+        echo "=== Installing Rust ==="
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+      fi
+      # Set PATH variable for cargo executable
+      . "$HOME/.cargo/env"
```
ricab: I don't suppose this could be achieved in a more snapcrafty way? Perhaps a separate part? 

https://documentation.ubuntu.com/snapcraft/stable/how-to/integrations/craft-a-rust-app/

Alternatively, can we at least install rust in a stable ubuntu-y way, i.e. apt/snap? I don't think we should just pipe an online script to sh in here. We don't control the contents and it would change from one version to the next, possibly in backward-incompatible ways... An Ubuntu package has a level of vetting and assurance.

BUILD.linux.md
```diff
@@ -10,6 +10,13 @@ sudo apt install devscripts equivs
 mk-build-deps -s sudo -i
 ```
 
+### Install the Rust compiler
+
+Go here for the most recent method. The current method is:
```
ricab: Did you mean to add a link in "go here"?

.gitignore
```diff
@@ -38,3 +38,6 @@ network-cache/
 
 # clangd code index cache
 .cache
+
+#Rust builds
+rxx/target
```
ricab: Can we instead move this into the build directory, please?

https://doc.rust-lang.org/cargo/reference/build-cache.html

ricab (changes requested):
Good stuff @tobe2098! 👏 

I have a few requests, but mostly questions for clarification inline.

A few general questions on the build:

- How do we run rust tests? Are they run when the module is built? Can we have a dedicated target that runs in the proper cmake phase?
- Does the build respect the chosen build directory fully? Does it keep from generating anything in the source repo or outside of the build directory? (apart from the one spotted below)

I'd invoke also a secondary review from @xmkg with a focus on the CMake and build, if that's alright. I know @sharder996 was already involved and @jimporter's review is pending, but this PR breaks new ground and more eyes won't hurt.

.github/workflows/linux.yml
tobe2098: Good catch, fix incoming soon.

rxx/namegen/src/petname_error.rs
tobe2098: I believe so, the function signature contains `char`, so the only accepted param is a single byte. The Rust error also only takes one single byte, so anything else will fail compilation.

rxx/namegen/src/petname_generator.rs
tobe2098: Yes, I believe that is cleaner

rxx/namegen/Cargo.toml
tobe2098: From my notes I believe this was a leftover of linking to call C++ from rust in logging. I left it to make it easier than keeping track of that, but I will test before removing to make sure. In the end only certain libraries may need the `rlib` type.

tobe2098: Alternatively it may be necessary to build with `rlib` when using other `rlib`s, but I did not write that down.

rxx/namegen/build.rs
tobe2098: This runs whenever `cargo` is run, so during the build process.

snap/snapcraft.yaml
tobe2098: That is a tricky question. Because we need to orchestrate `cargo` in all OSes inside CMake to support both CI and manual builds, cargo has to be present when CMake runs, so the snapcraft thing which works well for more isolated components does not work here. In that case the build files would also be built within `rxx/` and not `build/`.

In terms of Rust installation, I believe the snapcraft method gives you the same as the standard installation script in terms of versioning (latest stable from `rustup`). The Ubuntu package is 1) Not maintained by the Rust Foundation and 2) the installer version lags behind the `stable` from the Rust Foundation. I have tested `apt` and the compiler is the latest version, so we could go with this (there is also a snap `--classic`, higher version of `rustup`). We can go with the snap if that seems like a better option.

BUILD.linux.md
tobe2098: Good catch!

.gitignore
tobe2098: Oh yes, that was already done. This is not necessary anymore

tobe2098:
More eyes are definitely better. @sharder996's involvement was mainly to make sure compilation worked with the architecture support. The branch may need a re-visit after he does his gui-less-snap merge.

Tests are run with `ctest` (together with C++ tests), within the build/ folder (CI). The tests can also be run with `cargo test` in the `rxx/` folder, with more specific flags.

Tests are not run when the module is built, they have to be run on purpose.

The build should respect it, and it has as long as I have used it. The only issue is that running `cargo test` without `CARGO_TARGET_DIR=../build` from `rxx` will compile automatically to `rxx/target`. This is because without the env variable, `cargo` cannot see the build in `build/`. That is the only rough edge of moving all generated files to `build/`. `ctest` will still work as expected, using the build from `build/`

tobe2098:
On the build path issue, we could put it in `rxx/.cargo/config.toml`, but then there would be 2 sources of truth for that path, cmake and cargo, since cmake needs to know where the files are.

snap/snapcraft.yaml
tobe2098: The `apt` package was last updated in 2024, the snap has a version from 04/2025. The compiler is still the current version, just not the package manager `rustup`

.github/workflows/macos.yml
```diff
@@ -232,7 +235,7 @@ jobs:
     - name: Test
       working-directory: ${{ env.BUILD_DIR }}
       run: |
-        bin/multipass_tests
+        ctest --output-on-failure
```
xmkg: This only outputs when something fails. `ctest -V` would be the equivalent of the previous command

.github/workflows/windows.yml
```diff
@@ -176,7 +179,7 @@ jobs:
     - name: Test
       working-directory: ${{ env.BUILD_DIR }}
       run: |
-        bin/multipass_tests
+        ctest --output-on-failure
```
xmkg: idem

rxx/CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#Set build variables
+if (cmake_build_type_lower MATCHES "release")
```
xmkg: Can we use UPPER_CASE for the vars instead for consistency?

xmkg (commented):
Looks good overall, a few minor comments only.

.github/workflows/macos.yml
tobe2098: Done, also on Linux

rxx/CMakeLists.txt
tobe2098: Done

rxx/namegen/Cargo.toml
ricab: OK. My gut feeling is we should add the rlib only when we need it, but I will leave it up to you.

rxx/namegen/build.rs
ricab: Yeah, my question wasn't very well enunciated. I meant to ask about the mechanism that causes this to get run, but I figured it out now: cargo build scripts.

snap/snapcraft.yaml
ricab: Hmm I see. I would be more comfortable if we offloaded the responsibility of rust installation to an Ubuntu package, be that a deb or a snap published by Canonical. Even if it is not the most up-to-date version (arguably we should update the package ourselves if we need it) and even if they end up doing the same thing (today). Piping the online script should not be our responsibility.

ricab (dismissed):
Alright, I am good with this. I have tested a little and played with building, testing, and integrating in the IDE. Tried the snap as well. All good! Wonderful work @tobe2098!

I am leaving a couple of threads unresolved where I would prefer a slightly different approach, but leaving that up to others.

Note: So far I was only able to get CLion to run rust tests via CTest and report them as a single success/failure. The Rust plugin for CLion appears pretty limited. I'd probably use RustRover when working with Rust anyway.

snap/snapcraft.yaml
tobe2098: I understand. I think the snap has the closest to most recent version, and since the compiler is pulled by the package, it is always the latest stable version. I already changed it I think.

rxx/namegen/Cargo.toml
tobe2098: Yes, I will move the rlib to the logger branch once I create it, but removing from here. I forgot to remove it on the last round of changes.

tobe2098:
There are new changes to adapt to the macOS coredump upload

rxx/namegen/Cargo.toml
tobe2098: Actually, scratch that. `rlib` is needed for integration tests (`crate/tests`), so it is still necessary, or it won't compile.

.github/workflows/macos.yml
```diff
@@ -250,11 +253,33 @@ jobs:
...
       working-directory: ${{ env.BUILD_DIR }}
       run: |
-        # Catch and save the test executable's exit code
         trap 'echo "MULTIPASS_TESTS_EXIT_CODE=$?" >> $GITHUB_ENV' EXIT
         # Set soft limit for the core file size (512MiB)
         ulimit -c 1048576
-        bin/multipass_tests
+        ctest -V
+
+    - name: Gather coredump and executables
```
xmkg: Oh, I was thinking about this and how to adapt it when the tests are run by `ctest`. Glad to see that you sorted that out 👍

.github/workflows/macos.yml
```diff
@@ -250,11 +253,33 @@ jobs:
...
+      if: ${{ failure() && env.MULTIPASS_TESTS_EXIT_CODE != '0' }}
+      continue-on-error: true
+      run: |
+        set -o xtrace
+        echo "Test executable crashed. Gathering artifacts..."
```
xmkg: EXIT_CODE != 0 doesn't necessarily mean crashed. Crashed means either `MULTIPASS_TESTS_EXIT_CODE` == SEGV(139) (and a couple others like SIGILL), or non-zero exit code + core dumps under /cores. I think the ctest executable does not propagate the test executable's exit code, so it's a generic failure instead of specific, like 139. Hence, the way to detect a `crash` would be the latter option.

.github/workflows/macos.yml
tobe2098: It may require further work, it is hard to test locally. Do you think it is correct?

.github/workflows/macos.yml
```diff
@@ -250,11 +253,33 @@ jobs:
...
+
+        mkdir -p /tmp/coredump
+
+        if [ -d /cores ] && [ "$(ls -A /cores)" ]; then
+          cp /cores/* /tmp/coredump/
```
xmkg: Is there a reason to copy the files to tmp? I'm worried about this a bit because it would double the disk space consumption.

.github/workflows/macos.yml
```diff
@@ -250,11 +253,33 @@ jobs:
...
+        fi
+
+        cp "${{ env.BUILD_DIR }}/bin/multipass_tests" /tmp/coredump/
+
+        mapfile -t BINS < <(cargo test --workspace --no-run --message-format=json | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable')
```
xmkg: Instead of copying, you could do something like:

```yaml
- name: Collect rust binary paths
  if: failure() && env.MULTIPASS_TESTS_EXIT_CODE != '0'
  run: |
    mapfile -t BINS < <(cargo test --workspace --no-run --message-format=json \
      | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable')

    {
      echo 'EXTRA_UPLOAD_PATHS<<__EOF__'
      printf '%s\n' "${BINS[@]}"
      echo '__EOF__'
    } >> "$GITHUB_ENV"

```

Then, use EXTRA_UPLOAD_PATHS in the upload step:

```yaml
- name: Upload test coredump
  uses: actions/upload-artifact@v4
  if: ${{ failure() && env.MULTIPASS_TESTS_EXIT_CODE != '0' }}
  with:
    name: buildandtest-test-crash-${{ matrix.runs-on }}-${{ matrix.build-type }}
    retention-days: 7
    path: |
      /cores/**
      ${{ env.BUILD_DIR }}/bin/multipass_tests
      ${{ env.EXTRA_UPLOAD_PATHS }}
```

xmkg (changes requested):

ricab (dismissed):
Thanks again Antoni. I'll let you sort out the new CI stuff with Mustafa, I am fine with this otherwise. Adding my approval again.

.github/workflows/macos.yml
xmkg: I don't see any obvious problems. Let's see how the pipeline will go with your SIGILL commit.

tobe2098: I have fixed all the existing issues, now the coredump is uploaded and there are no issues gathering the Rust executables. I also corrected the linux one, which had the same issues.

rxx/CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+    #Static library path for the rust source.
+    set(RUST_OUT "${RUST_BIN_DIR}/${LIB_PREFIX}${CRATE_NAME}.${LIB_SUFFIX}")
+    #Target to depend on cargo build.
+    add_library(${CRATE_NAME}_rs STATIC IMPORTED)
+    set_target_properties(${CRATE_NAME}_rs PROPERTIES IMPORTED_LOCATION ${RUST_OUT})
+    set(${CRATE_NAME}_LINK_LIBS "")
+    if (WIN32)
+      list(APPEND ${CRATE_NAME}_LINK_LIBS "$<LINK_ONLY:ntdll>")
+    endif()
+    set(BRIDGE_SOUrce "${RUST_BUILD_DIR}/cxxbridge/${CRATE_NAME}/src")
```
xmkg: the last 3 letters are lowercase

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+[profile.dev]
+opt-level = 0
+debug = true
+strip = "none"
+debug-assertions = true
+overflow-checks = true
+lto = true
+panic = 'unwind'
+incremental = true
+codegen-units = 1 #Lower means slow compilation better optimization
```
xmkg: This is the debug profile, right? If so, why are we aiming to optimize the optimization level instead of the iteration speed?

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+namegen = { path = "namegen" }
+
+# This is debug
+[profile.dev]
+opt-level = 0
+debug = true
+strip = "none"
+debug-assertions = true
+overflow-checks = true
+lto = true
```
xmkg: This is an optimization flag, too. Shouldn't we remove it from the debug builds?

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+panic = 'unwind'
+incremental = true
```
xmkg: This is implicitly true for debug builds, no? https://doc.rust-lang.org/cargo/reference/profiles.html?highlight=incremental#dev

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+cxx-build = "^1.0.0"
+static_assertions = "^1.1.0"
+syn = "^2.0.0"
+quote = "^1.0.0"
+proc-macro2 = "^1.0.0"
+rand = "^0.9.0"
+macros = { path = "macros" }
+namegen = { path = "namegen" }
+
+# This is debug
```
xmkg, line 38: This seems to be redundant since we're only overriding optimization-related flags (lto, codegen-units). Is there a reason to keep this?

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+panic = 'unwind'
+incremental = true
+codegen-units = 1 #Lower means slow compilation better optimization
+rpath = false
+
+[profile.release]
+lto = true
+debug = false
+strip = "debuginfo"
+opt-level = "s"
```
xmkg: How much of a difference does this make w.r.t. the binary size? I'd prefer the default (3) if it's not too much.

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+lto = true
+panic = 'unwind'
+incremental = true
+codegen-units = 1 #Lower means slow compilation better optimization
+rpath = false
+
+[profile.release]
+lto = true
+debug = false
+strip = "debuginfo"
```
xmkg: I'd be inclined to keep this as "none" as well, to get more sensible stack traces if we're talking about a 10-20 mb overhead.

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+opt-level = "s"
+codegen-units = 1 #Lower codegen count improves compiler optimization
+panic = "abort"
```
xmkg: The default panic behavior for release is "unwind". Is there a reason to change that?

rxx/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+strip = "none"
+debug-assertions = true
+overflow-checks = true
+lto = true
+panic = 'unwind'
+incremental = true
+codegen-units = 1 #Lower means slow compilation better optimization
+rpath = false
+
+[profile.release]
```
xmkg, line 43: debug-assertions, overflow-checks, and debug overrides are redundant since they're on par with the default profile.  Also, it'd be nice if we could put a short comment right next to the override about why we decided to change the default.

xmkg (changes requested):
Thanks for sorting out the coredump handling in CI! I have a couple of comments regarding the cargo profiles, and a small typo.

rxx/CMakeLists.txt
tobe2098: Oh, good catch!

rxx/Cargo.toml
tobe2098: The logic was that the debug executable should look as much like the release one as possible for debugging purposes, WDYT?

rxx/Cargo.toml
tobe2098: Yes, I was just being explicit with them. I can remove the ones that are the same as the original profile

rxx/Cargo.toml
tobe2098: The codegen-units is also there to allow the Rust compilation to run in a single thread while other C++ units are being compiled.

xmkg: The debug executable is roughly the release executable minus optimizations. If we start adding optimizations back, we'd land somewhere in between, and we'd be troubleshooting the optimization artifacts, if any. The debug should be optimized for being "as faithful as possible to the source". No tricks, no magic. Just verbose machine code doing what's being told.

If the concern is about the tests being run in debug profile, we can opt in running them in the release build instead. Ideally the pipeline should be running both (tests w/ debug, tests w/ release) but if I were to choose one, I'd pick running UTs on the release build since it's what we're going to ship after all.

About the codegen-units, it makes sense since the c++ build process would be competing for the same resources. We can revisit it later if we find the build to be too slow after introducing more rust components.

rxx/Cargo.toml
tobe2098: Moving all size-related discussions here:
`lto` is useful in debug as well. It just removes dead code and moves around the position of code sections as far as I am aware, although we can do away with it.
The size flag was chosen due to concerns with the size of the added Rust std and crates. I have tested extensively and there seems to be no difference in the final executables in MiBs. The original testing was done under different conditions (logger crate, different petname implementation), but it seems here that unwind,opt-level and strip have no effect. A possible reason is the little amount of Rust code that is left after LTO linking to the main executable. But under the previous conditions, there were appreciable differences between using the optimizations and not.

I agree with removing the debug optimizations, but the release ones would require further discussion. As we increase the amount of code, parts of std and other crates would be pulled in, and could increase the difference in size increase. During discussions about the size impact there were concerns so I thought we would keep them unless they cause any issue. WDYT? We could discuss it more widely.

rxx/Cargo.toml
tobe2098: Apart from extra executable size, we cannot allow panics in our case, since it would instantly terminate the daemon in our case. Having the unwind sections is unnecessary when we cannot call panicking functions.

rxx/Cargo.toml
tobe2098: For now I will remove the optimizations

xmkg (dismissed):
LGTM!

xmkg (dismissed):
Rubber stamping again.

tobe2098:
After the last rebase, I would point to the coredump/multipass_integration_tests/ctest/cargo test integration as needing review

.github/workflows/macos.yml
```diff
@@ -229,16 +233,42 @@ jobs:
...
+    - name: Collect rust binary paths
+      working-directory: ${{ env.RUST_DIR }}
+      run: |
+        {
+          echo 'RUST_EXEC_PATHS<<__EOF__'
+          cargo test --workspace --no-run --message-format=json --target-dir ${{ env.BUILD_DIR }}/rxx \
+            | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable' \
+            | while IFS= read -r exe; do
+                [ -n "$exe" ] && echo "$exe"
+              done
```
jimporter: Just to make sure I understand. This is trying to get all the Rust executables' paths and put them into an env var, one per line, right? If so, what do you think about this?

```suggestion
            | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable'
```

And then below...

.github/workflows/macos.yml
```diff
@@ -229,16 +233,42 @@ jobs:
...
+
+        RUST_EXECUTABLES=()
+        while IFS= read -r line; do
+          [[ -n "$line" ]] || continue
+          RUST_EXECUTABLES+=("$line")
+        done <<< "$RUST_EXEC_PATHS"
+        for RUST_EXEC in "${RUST_EXECUTABLES[@]}";
+        do
+          codesign -s - -f --entitlements segv.entitlements "$RUST_EXEC"
+        done
```
jimporter: ... instead of using Bash, which is awkward at iterating over lines in a string, we use awk instead? I think that would be something like this:

```suggestion
        echo "$RUST_EXEC_PATHS" | awk $'/./ {
            system("codesign -s - -f --entitlements segv.entitlements \'" $0 "\'")
        }'
```

Then we can avoid Bash sorcery to create/split arrays, and also we don't need to worry about empty lines (the awk regex `/./` ensures we only look at non-empty lines).

.github/workflows/linux.yml
```diff
@@ -222,20 +222,26 @@ jobs:
...
       timeout-minutes: 2
       run: |
-
-        trap 'echo "MULTIPASS_TESTS_EXIT_CODE=$?" >> $GITHUB_ENV' EXIT
-        instance_name=`/snap/bin/lxc --project snapcraft --format=csv --columns=n list | grep multipass`
+        instance_name=$(/snap/bin/lxc --project snapcraft --format=csv --columns=n list | grep multipass)
+        trap '
+          EXIT_STATUS=$?
+          echo "MULTIPASS_TESTS_EXIT_CODE=$EXIT_STATUS" >> $GITHUB_ENV
+          # Check if any files were generated in /cores/
```
jimporter: Nit: Should this be `/coredump/` instead of `/cores/`? I believe the latter is just the location on macOS.

.github/workflows/macos.yml
jimporter: Just to make sure I understand. This is trying to get all the Rust executables' paths and put them into an env var, one per line, right? If so, what do you think about this?

```suggestion
            | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable'
```

And then below...

.github/workflows/macos.yml
jimporter: ... instead of using Bash, which is awkward at iterating over lines of text, how about using awk instead? I think that would look like this:

```suggestion
        echo "$RUST_EXEC_PATHS" | awk $'/./ {
            system("codesign -s - -f --entitlements segv.entitlements \'" $0 "\'")
        }'
```

That also means we don't need any tricks to account for empty lines, since the awk regex `/./` only select lines with at least one character.

include/multipass/name_generator.h
```diff
@@ -37,5 +37,37 @@ class NameGenerator : private DisabledCopyMove
...
+} // namespace detail
+
+// Templated functions to avoid exposing the PetnameProvider type in the interface header with
+// the default arguments. The templates ensure the argument checks are done at compile-time.
+template <NumWords NW = NumWords::Two, char S = '-'>
+    requires IsValidSeparator<S>
+NameGenerator::UPtr make_petname_provider()
+{
+    return detail::make_petname_provider_impl(NW, S);
+};
```
jimporter: Nit:

```suggestion
}
```

packaging/gui-less/snap/snapcraft.yaml
```diff
@@ -96,6 +96,7 @@ parts:
     plugin: cmake
     build-snaps:
     - cmake
+    - rustup
```
jimporter, line 98: 👍 I had a review comment about a previous iteration of this change (possibly in the main `snap/snapcraft.yaml`, but same difference...), but this is nice and simple now!

rxx/macros/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    let content = std::fs::read_to_string(&filename).expect("Could not read file");
+
+    // Split & reformat as string literals for quote!
+    let words: Vec<syn::LitStr> = content
+        .split_whitespace()
+        .map(|word| syn::LitStr::new(word, Span::call_site()))
+        .collect();
+
+    let output = quote! {
+        [ #( #words ),* ]
```
jimporter, line 37: So this is embedding the contents of a file as an array at compile time? Neat!

rxx/namegen/src/petname_generator.cpp
```diff
@@ -0,0 +1,16 @@
+
```
jimporter: Nit: Missing GPL header.

rxx/namegen/src/name_generator.cpp
```diff
@@ -0,0 +1,31 @@
+/*
```
jimporter, line 1: I wonder if this file makes more sense at `multipass/src/name_generator.cpp`, since it's the `.cpp` file for `multipass/include/multipass/name_generator.h`. Maybe that makes the include paths trickier though...

rxx/namegen/tests/test_petname_error.rs
```diff
@@ -0,0 +1,12 @@
+use namegen::petname_error::PetnameError;
```
jimporter, line 18: Nit: Missing GPL header.

rxx/namegen/tests/test_petname_generator.rs
```diff
@@ -0,0 +1,40 @@
+use namegen::{ffi::NumWords, petname_error::PetnameError, petname_generator::make_petname};
```
jimporter: Nit: Missing GPL header.

rxx/namegen/tests/test_petname_generator.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    for i in 0..TOTAL_NAMES {
+        let petname = make_petname(NumWords::Three, '-' as c_char).unwrap();
+        assert!(
+            hashset.insert(petname),
+            "Generated duplicate petname at iteration {}",
+            i
+        );
+    }
+}
```
jimporter, line 69: This test seems like it could cause some intermittent failures due to the randomness. Should we add the ability to set a random seed for our unit tests?

rxx/namegen/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+# Here the rust build depends on the C++ side of the rust_logger bridge because
+# the rust_logger crate is compiled within this crate's static library.
+# Because we did not build it, we specify the dependency like:
+#list(APPEND ${crate_name}_link_libs "$<LINK_ONLY:rust_logger>")
+#Add necessary sources for wrappers (it is advised that the source is named after
+# the header in include/multipass/)
+#target_sources(${crate_name} PRIVATE "src/name_generator_rs.cpp")
+#Link necessary libraries for the exceptions header
+#target_link_libraries(${crate_name} PRIVATE fmt::fmt-header-only Qt::Core)
```
jimporter: Is this comment still relevant? I don't see any non-comment references to `$<LINK_ONLY:rust_logger>`.

rxx/CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+endif()
+
+# Build output path
+set(RUST_BUILD_DIR "${CMAKE_BINARY_DIR}/rxx")
+set(RUST_BIN_DIR "${RUST_BUILD_DIR}/${RUST_BUILD}")
+
+#Find cargo, which will be a hard dependency
+find_program(CARGO_EXECUTABLE cargo)
+#Set the build directory via env variable
+set(CARGO_CMD ${CMAKE_COMMAND} -E env "CARGO_TARGET_DIR=${RUST_BUILD_DIR}" ${CARGO_EXECUTABLE})
```
jimporter, line 38: 👍 In a previous iteration of this review, I had a suggestion to do something like this. This will make our lives much easier with supporting multiple build dirs for a single source tree. (That's particularly useful for building with feature flags enabled/disabled.)

tests/unit/test_rust_integration.cpp
```diff
@@ -0,0 +1,63 @@
...
+using namespace testing;
+
+namespace
+{
+std::vector<std::string> split(const std::string& string, char delimiter)
+{
+    std::regex regex(std::string(1, delimiter));
+    return {std::sregex_token_iterator{string.begin(), string.end(), regex, -1},
+            std::sregex_token_iterator{}};
+}
```
jimporter, line 43: This is a pretty simple function, so I don't mind very much keeping it as-is, but Boost.StringAlgorithm gives us an implementation of `split` that we could use: https://www.boost.org/doc/libs/latest/doc/html/string_algo/usage.html#id-1.3.3.5.9

(We could also use ISO C++ ranges, but it would be a lot easier to do that in C++23 once [`std::ranges::to`](https://en.cppreference.com/cpp/ranges/to) is available.)

BUILD.linux.md
```diff
@@ -10,6 +10,13 @@ sudo apt install devscripts equivs
 mk-build-deps -s sudo -i
 ```
 
+### Install the Rust compiler
+
+Go [here](https://rust-lang.org/tools/install/) for the most recent method. The current method is:
+```
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
```
jimporter: In our `snapcraft.yaml`, we install `rustup` via Snap. Should we suggest doing the same here? That's more consistent, and then we're not suggesting a curl->sh pipeline. :)

jimporter (changes requested):
This looks really good overall. Thanks for all the time spent on this (and the patience in my review). Some comments below, but they're mostly minor.

BUILD.linux.md
tobe2098: Yes, good catch!

rxx/namegen/CMakeLists.txt
tobe2098: It was a carry over from the experimental branch, where the item is there. I will remove it for now, until I work on the Rust logger branch

rxx/namegen/tests/test_petname_generator.rs
tobe2098: I think it is reasonable. The combinatorics would fail at a rate of 0.024% or 1 test of every 4200, and it is avoidable.

tobe2098: Sadly, after attempting to do so, it would require a refactor in one of two directions: Either we bring back the petname struct in the Rust side or add a global RNG and a function to initialize it. I consider the first one to be the best approach, WDYT?

rxx/namegen/src/name_generator.cpp
tobe2098: Yes, and the CMake scripting would become riddled with path traversals. Technically, the "source module" is not in `src` to begin with, since all generated code comes from the rust side, but in `rxx`. The code would be separated and harder to map.
Although it is not ideal, I do not think the change would improve things.

rxx/macros/src/lib.rs
tobe2098: Yes :D

rxx/namegen/tests/test_petname_generator.rs
tobe2098: I went for the first approach, I think it is the best solution for now, unless we discard the test or modify it in another way.

The best approach may be to discard the test all-together, since the added code is significant just for seeding.

jimporter: Yeah, I would be ok with removing this test. We could also keep the old implementation around with this test, but make it warn on failure? On the other hand, we probably wouldn't pay close attention to a test that warns...

(Another way to do this would be to make sure that the various name lists have a coprime number of entries, and then we just generate a single random start index and increment that by one each time, using that to index into each list, modulo its size. That way, we'd have to cycle through the whole list before we get repeats. However, that's a bigger change that probably requires a bit more thought than this very sketchy idea. Maybe something to think about for the future.)

jimporter (dismissed):
LGTM!

tobe2098:
Re-requesting reviews after significant changes to the code

.github/workflows/linux.yml
```diff
@@ -222,19 +222,24 @@ jobs:
...
+            echo "CORE_DUMPS_FOUND=true" >> $GITHUB_ENV
+          fi
+        ' EXIT
         /snap/bin/lxc --project snapcraft start $instance_name
         # Let's print the core pattern so we can check if it's successfully propagated to the container.
         /snap/bin/lxc --project snapcraft exec $instance_name -- bash -c 'cat /proc/sys/kernel/core_pattern'
-        # Create the directory for the coredumps
-        /snap/bin/lxc --project snapcraft exec $instance_name -- bash -c 'mkdir -p /coredump'
+        # Create the directory for the coredumps and ensure the directory is empty
+        /snap/bin/lxc --project snapcraft exec $instance_name -- bash -c 'mkdir -p /coredump && rm -f /coredump/*'
```
ricab, line 238: This seems a little out of scope for this PR, no? Could it be moved to a separate PR? Or is there a special motivation that I am missing?

.github/workflows/linux.yml
```diff
@@ -320,16 +325,21 @@ jobs:
...
+        mapfile -t BINS < <(/snap/bin/lxc --project snapcraft exec "$instance_name" -- bash -c "
+          cd /root/parts/multipass/src/rxx
+          cargo test --workspace --no-run --message-format=json --target-dir /root/parts/multipass/build/rxx | \
+          jq -r 'select(.reason == \"compiler-artifact\" and .executable != null) | .executable'
+        ")
         for bin in "${BINS[@]}"; do
-          /snap/bin/lxc --project snapcraft file pull -p -r "$instance_name/$bin" "/tmp/coredump/$bin"
-        done"
+          /snap/bin/lxc --project snapcraft file pull -p -r "$instance_name/$bin" "/tmp/coredump/$(basename "$bin")"
+        done
```
ricab, line 336: Idem, pulling core dumps feels like something that should be done on top, no?

.github/workflows/macos.yml
```diff
@@ -232,16 +233,33 @@ jobs:
...
+    - name: Collect rust binary paths
+      working-directory: ${{ env.RUST_DIR }}
+      run: |
+        {
+          echo 'RUST_EXEC_PATHS<<__EOF__'
+          cargo test --workspace --no-run --message-format=json --target-dir ${{ env.BUILD_DIR }}/rxx --quiet \
+            | jq -r 'select(.reason == "compiler-artifact" and .executable != null) | .executable'
+          echo '__EOF__'
+        } >> "$GITHUB_ENV"
+
```
ricab, line 245: idem

rxx/Cargo.toml
```diff
@@ -37,24 +37,9 @@ namegen = { path = "namegen" }
...
-debug = false
-strip = "debuginfo"
-opt-level = "s"
-codegen-units = 1 #Lower codegen count improves compiler optimization
-panic = "abort"
-debug-assertions = false
-overflow-checks = false
-incremental = false #Removing incremental compilation allows lower codegen count
+lto = true #Removes unused code sections
+codegen-units = 1 #Less threads to compete with C++ compilation
```
ricab, line 45: We're sure that these inherit the remaining properties, yes?

ricab (commented):
Light re-review from my side. I wonder if the coredump pulling deserves a separate PR with dedicated reviewing. But I won't block on it. I did not review that part in detail, but if others have, that is probably enough.

rxx/Cargo.toml
tobe2098: You mean whether the values of the removed properties are the defaults of each profile? In that case, yes.

ricab: Exactly, thanks!

.github/workflows/linux.yml
tobe2098: I added the coredump work for completeness, and tested (works successfully in linux runners at least). I am open to separate on another PR though.

tobe2098: I think Mustafa reviewed it before Jim as well, but again, could be moved to a separate PR.

ricab (approved):
Alright, ship it!

tobe2098:
Merge seems to be blocked by a HyperV integration test failing in CI (not in a local VM).

.github/workflows/linux.yml
xmkg: I'm fine with both approaches since this is a non-critical troubleshooting improvement rather than a core CI step.

xmkg:
> Merge seems to be blocked by a HyperV integration test failing in CI (not in a local VM).

@tobe2098 Do you remember which test was failing? I can take a look at it if there's flakiness in the integration tests

xmkg:
+ @tobe2098  also, windows build seems to fail due to some errors in compiling the virtualbox backend

PR #4602, opened 2026-01-08
[gha-linux] Add retry to the Configure step

xmkg:
To alleviate the race condition in Snapcraft's LXD integration.

# Description

- What does this PR do?
It tries to fix a intermittent snapcraft pull failure in the configure step.
- Why is this change needed?
To have a better CI experience.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -169,10 +169,11 @@ jobs:
 
             exit $exit_code
           }
-
+          # The auto-refresh race would manifest itself in the first pull attempt.
+          # Hence wrapping only the first call.
           run_snapcraft /snap/bin/snapcraft pull --use-lxd inject-apt-mirrors
-          run_snapcraft /snap/bin/snapcraft pull --use-lxd configure-nuget-cache
-          run_snapcraft /snap/bin/snapcraft pull --use-lxd multipass
+          /snap/bin/snapcraft pull --use-lxd configure-nuget-cache
+          /snap/bin/snapcraft pull --use-lxd multipass
 
     - name: Run clang-tidy through the diff
       if: ${{ false && matrix.build-type == 'Debug' }}


.github/workflows/linux.yml
```diff
@@ -140,10 +140,39 @@ jobs:
...
-        /snap/bin/snapcraft pull --use-lxd multipass
+      # To alleviate the race condition in the Snapcraft's LXD
+      # integration where the snapd is auto-refreshing itself on
+      # container start.
+      #  https://github.com/canonical/craft-providers/issues/391
+      uses: nick-fields/retry@v3
+      with:
+        # We have to set a value for this as it's not optional
+        # https://github.com/nick-fields/retry/issues/51
+        timeout_minutes: 720 # 12 hours
```
sharder996, line 151: FWIW, I think GitHub workflows automatically time out after 6 hours.

xmkg:
Already in action:

image
image

.github/workflows/linux.yml
```diff
@@ -140,10 +140,39 @@ jobs:
...
+        max_attempts: 5
+        retry_wait_seconds: 20
+        retry_on_exit_code: 42
+        command: |
+          run_snapcraft() {
+            tmpfile=$(mktemp)
+
+            "$@" 2>&1 | tee "$tmpfile"
```
jimporter:
```suggestion
            "$@" |& tee "$tmpfile"
```

.github/workflows/linux.yml
```diff
@@ -140,10 +140,39 @@ jobs:
...
+            if grep -qE "daemon is stopping to wait for socket activation|Failed to wait for snap refreshes" "$tmpfile"; then
+              exit 42
+            fi
+
+            exit $exit_code
+          }
+
+          run_snapcraft /snap/bin/snapcraft pull --use-lxd inject-apt-mirrors
+          run_snapcraft /snap/bin/snapcraft pull --use-lxd configure-nuget-cache
+          run_snapcraft /snap/bin/snapcraft pull --use-lxd multipass
```
jimporter: From my understanding, if this fails, we'd always expect it to fail on the first call, right? I'm mainly wondering about what the right thing to do here is for retries: if this *could* fail on `configure-nuget-cache`, then retrying would also retry `inject-apt-mirrors`. If it can only fail on the first call, then we technically don't need the `run_snapcraft` wrapper for the 2nd and 3rd ones.

I don't think this is a serious enough issue to make any changes here (unless you really want to), though a comment might be helpful in case this code sticks around for a while.

xmkg: Yeah, it only should fail on the first attempt so run_snapcraft on the rest is redundant. I'll add a comment and also remove them from the subsequent calls. We can introduce them back later if needed .

PR #4608, opened 2026-01-13
[utils] Use Boost.JSON for `VMSpecs`

jimporter:
# Description

Yet another PR in the Boost.JSON migration. This time around, we're migrating `VMSpecs` to use Boost.JSON. As with the [migration of metadata](https://github.com/canonical/multipass/pull/4597), this includes some temporary conversion between Qt's JSON and Boost.JSON, though now this conversion is moved up to the snapshot code.

## Testing

All unit tests have been updated to handle the type migration.

## Checklist

### DIFF
--- a/src/utils/vm_specs.cpp
+++ b/src/utils/vm_specs.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <multipass/constants.h>
+#include <multipass/exceptions/ghost_instance_exception.h>
+#include <multipass/json_utils.h>
+#include <multipass/utils.h>
+#include <multipass/vm_specs.h>
+
+namespace mp = multipass;
+namespace mpu = multipass::utils;
+
+void mp::tag_invoke(const boost::json::value_from_tag&,
+                    boost::json::value& json,
+                    const mp::VMSpecs& specs)
+{
+    json = {{"num_cores", specs.num_cores},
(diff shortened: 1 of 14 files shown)


src/utils/vm_specs.cpp
```diff
@@ -0,0 +1,80 @@
...
+    auto mem_size = value_to<std::string>(json.at("mem_size"));
+    auto disk_space = value_to<std::string>(json.at("disk_space"));
+    auto mac_addr = value_to<std::string>(json.at("mac_addr"));
+    auto ssh_username = value_to<std::string>(json.at("ssh_username"));
+    auto deleted = value_to<bool>(json.at("deleted"));
+    auto metadata = json.at("metadata").as_object();
+
+    if (!num_cores && !deleted && ssh_username.empty() && metadata.empty() &&
+        !MemorySize{mem_size}.in_bytes() && !MemorySize{disk_space}.in_bytes())
+        throw GhostInstanceException();
```
sharder996, line 61: I find this procedure a little odd, but it does replicate the original behaviour, I guess...

sharder996 (dismissed):
LGTM

I figure at this point the foundation has already been laid for converting QJson and Boost.JSON and its just a matter of applying the operation in different places.

src/utils/vm_specs.cpp
jimporter: Yeah, I tried to be extra-careful here to do the same thing as before, even if it seemed like a strange idea. I'm all for improving the logic here too, but preferably as a separate step so that we can carefully evaluate all the behavioral changes.

tests/json_test_utils.cpp
```diff
@@ -115,88 +113,95 @@ std::unique_ptr<mpt::TempDir> plant_instance_json(const std::string& contents)
...
     ASSERT_EQ(default_mac, mac);
 
-    const auto extra = instance_object["extra_interfaces"].toArray();
-    ASSERT_EQ((unsigned)extra.size(), extra_ifaces.size());
+    const auto extra = instance_object.at("extra_interfaces").as_array();
+    ASSERT_EQ(extra.size(), extra_ifaces.size());
 
-    auto it = extra_ifaces.cbegin();
-    for (const auto extra_i : extra)
+    auto it = extra_ifaces.begin();
```
tobe2098: Is there a reason for getting a non-const iterator?

tobe2098 (dismissed):
LGTM! Good work Jim! Just a small inline question

tests/json_test_utils.cpp
jimporter: `extra_ifaces` is already const, so the iterator is const too. I generally prefer avoiding `cbegin` unless it's actually doing something different, but since that's more of a stylistic change that the team should agree on, I've reverted this part for now.

sharder996 (approved):

Issue #4607, opened 2026-01-13
[sshfs-mounts] sshfs mount creation fails


sharder996:
**Describe the bug**
Multipass fails to create sshfs mounts.

**To Reproduce**
1. `multipass mount ./ foo:test`

**Logs**
```
scott@Scotts-MacBook-Pro build-gui % multipass mount ./ foo:test -vvvv
[2026-01-13T13:42:31.479] [debug] [mount cmd] /Users/scott/Documents/dev/multipass/src/client/cli/cmd/mount.cpp:239 parse_args(): adding default uid mapping
[2026-01-13T13:42:31.481] [debug] [mount cmd] /Users/scott/Documents/dev/multipass/src/client/cli/cmd/mount.cpp:282 parse_args(): adding default gid mapping
[2026-01-13T13:42:31.497] [info] [sshfs-mount-handler] initializing mount /Users/scott/Documents/dev/multipass/build-gui => /home/ubuntu/test in 'foo'
[2026-01-13T13:42:31.533] [debug] [ssh session] Executing 'which snap'
[2026-01-13T13:42:31.571] [debug] [ssh session] Executing 'sudo snap list multipass-sshfs'
[2026-01-13T13:42:31.870] [debug] [ssh session] Executing '[ -e /snap ]'
[2026-01-13T13:42:31.873] [info] [sshfs-mount-handler] Installing the multipass-sshfs snap in 'foo'
[2026-01-13T13:42:31.873] [debug] [ssh session] Executing 'sudo snap install multipass-sshfs'
[2026-01-13T13:42:48.720] [info] [sshfs-mount-handler] process program '/Library/Application Support/com.canonical.multipass/bin/sshfs_server'
[2026-01-13T13:42:48.720] [info] [sshfs-mount-handler] process arguments '192.168.2.183, 22, ubuntu, /Users/scott/Documents/dev/multipass/build-gui, /home/ubuntu/test, 501:-1,, 20:-1,, 3'
[2026-01-13T13:42:48.724] [debug] [/Library/Application Support/com.canonical.multipass/bin/sshfs_server] [30293] started: /Library/Application Support/com.canonical.multipass/bin/sshfs_server 192.168.2.183 22 ubuntu /Users/scott/Documents/dev/multipass/build-gui /home/ubuntu/test 501:-1, 20:-1, 3
[...]

PR #4614, opened 2026-01-15
Migrate `libssh` to vcpkg

xmkg:
# Description

- What does this PR do?
  - Replaces `libssh` submodule with vcpkg
- Why is this change needed?
  - Customizing dependencies via forking is a terrible way of managing dependencies, and it leads to `dependency hell`
 

#Related issues

#4607

MULTI-2337


xmkg:
CI output for non-intrusive patching:

```text
 :: Successfully downloaded libssh-0.11.3.tar.xz
  :: -- Injecting Multipass patchset for vcpkg port `libssh`:
  :: --   0001-connector-stderr.patch
  :: --   0002-sftpserver-increase-handle-count.patch
  :: --   0003-fix-handle-realloc-failure.patch
  :: -- Extracting source /root/parts/multipass/src/3rd-party/vcpkg/downloads/libssh-0.11.3.tar.xz
  :: -- Applying patch /root/parts/multipass/src/3rd-party/vcpkg-triplets/../vcpkg-patches/libssh/0001-connector-stderr.patch
  :: -- Applying patch /root/parts/multipass/src/3rd-party/vcpkg-triplets/../vcpkg-patches/libssh/0002-sftpserver-increase-handle-count.patch
  :: -- Applying patch /root/parts/multipass/src/3rd-party/vcpkg-triplets/../vcpkg-patches/libssh/0003-fix-handle-realloc-failure.patch
  :: -- Applying patch 0001-export-pkgconfig-file.patch
  :: -- Applying patch 0003-no-source-write.patch
  :: -- Applying patch 0004-file-permissions-constants.patch
  :: -- Applying patch android-glob-tilde.diff
```

src/sshfs_mount/sftp_server.cpp
```diff
@@ -27,7 +27,10 @@
 
 #include <multipass/utils.h>
 
-#include <libssh/sftp_priv.h>
```
xmkg, line 30: This is a libssh-private header. We're only using one symbol from it so instead of messing with the include paths, I opted in declaring the function instead.

CMakeLists.txt
```diff
@@ -337,7 +342,6 @@ if(MSVC)
   add_definitions(-D_WIN32_WINNT=0xA00) # A gprc header requires this (specifies Windows 10)
   add_definitions(-DNOMINMAX) # otherwise windows defines macros which interfere with std::min, std::max
   add_definitions(-DMULTIPASS_PLATFORM_WINDOWS)
-  add_definitions(-DLIBSSH_STATIC) # otherwise adds declspec specifiers to libssh apis
```
xmkg, line 279: We don't have to worry about this anymore since it's all controlled by vcpkg.

CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ cmake_policy(SET CMP0079 NEW) # Allow target_link_libraries() in subdirs
 include(src/cmake/environment-utils.cmake)
 is_running_in_ci(IS_RUNNING_IN_CI)
 
-option(VCPKG_BUILD_DEFAULT "Enable or disable building the default vcpkg triplet, which includes both debug and release variants." OFF)
+option(MULTIPASS_VCPKG_BUILD_DEFAULT "Enable or disable building the dependencies in the default configuration, which includes both debug and release variants." OFF)
```
xmkg, line 22: Renamed this so it looks like more of a Multipass variable than a vcpkg one.

xmkg:
CLI-tests, macos 26.2:

```bash
========================================== FAILURES ==========================================
_________________________ TestShell.test_shell_exit_code_propagation _________________________

self = <cli_tests.cli_shell_test.TestShell object at 0x108b6f9d0>
instance = 'mithril-sting-legion-9906'

    @requires_multipass(">=1.17.0")
    def test_shell_exit_code_propagation(self, instance):
        """Launch an Ubuntu VM. Then, try to shell into it and execute some
        basic commands."""
    
        with shell(instance) as vm_shell:
            vm_shell.expect(r"ubuntu@.*:.*\$", timeout=30)
    
            # Send a command and expect output
            vm_shell.sendline('exit 42')
            vm_shell.expect(pexpect_eof)
            vm_shell.wait()
    
>           assert vm_shell.exitstatus == 42
E           assert 2 == 42
E            +  where 2 = <pexpect.pty_spawn.spawn object at 0x10930e0d0>.exitstatus

tools/cli_tests/cli_shell_test.py:87: AssertionError
================================== short test summary info ===================================
FAILED tools/cli_tests/cli_shell_test.py::TestShell::test_shell_exit_code_propagation - assert 2 == 42
============= 1 failed, 255 passed, 64 skipped, 4 xpassed in 4416.92s (1:13:36) ==============
```

`test_shell_exit_code_propagation` failure is to be expected -- as this branch wasn't rebased on main after https://github.com/canonical/multipass/pull/4567 got merged.

3rd-party/vcpkg-ports/libssh/multipass-patches/0001-connector-stderr.patch
```diff
@@ -0,0 +1,35 @@
+From 99bee08ef9a9aa495f757fec3c93286b0cf60db2 Mon Sep 17 00:00:00 2001
```
jimporter, line 1: This isn't blocking the review, but just for my knowledge, have we ever looked into upstreaming these patches?

3rd-party/vcpkg-triplets/per-package-overrides.cmake
```diff
@@ -0,0 +1,45 @@
+macro(multipass_inject_vcpkg_patches)
```
jimporter: I've been thinking this over, and while this is less boilerplate code than copying over the vcpkg port into our tree, I also wonder how stable this is. Could this break in subtle ways if we update our libssh dependency? (Or any other vcpkg dep that uses this new machinery.) I'm only somewhat familiar with vcpkg, so I don't have a good gut feeling for whether or not this is likely to require fixes in the future...

That's not to say that this is a bad idea, just that I don't know enough about this particular area to have a high level of confidence in approving the change.

jimporter (commented):
This looks good overall, and builds/runs correctly on my system with some basic manual tests. Just a few open-ended thoughts below.

3rd-party/vcpkg-ports/libssh/multipass-patches/0001-connector-stderr.patch
xmkg: AFAICT, there was an ex-colleague who worked closely with the upstream and landed most of the patches, but this has stopped being the case for a while now. Maybe that could be something we should look into.

3rd-party/vcpkg-triplets/per-package-overrides.cmake
xmkg: Yeah, this isn't ideal in my eyes either, but unfortunately, vcpkg does not provide any customization points for simply applying a couple of patches on top of an existing port -- you have to copy and maintain a port overlay. It is not possible to partially override the port either, so it's all or nothing. This approach provides a middle ground, but we have to override a vcpkg function to do that. The good thing is that it's possible to override the function only for the target port, so other ports are not affected at all. 

Vcpkg maintainers have suggested the triplet overrides instead of port overlays, but I don't think they meant "sure, override our functions". FWIW, vcpkg itself is overriding CMake built-ins like install(), add_executable|library, so it's a bit wild west.

> Could this break in subtle ways if we update our libssh dependency?

Yeah, for instance, it could stop applying our patches if the portfile decides to use  `vcpkg_from_github()` instead of `vcpkg_download_distfile()` + `vcpkg_extract_source_archive()`. That could be alleviated with a deferred patch check (more code) or a portfile review during update(more human effort). Other than that, I think it's a relatively safe override since the function itself has been there [since the dawn of time](https://github.com/microsoft/vcpkg/blame/734f8130ffe2f02cf855a3a42a2958f01b3fb005/scripts/cmake/vcpkg_extract_source_archive.cmake#L4), and updating the API would mean changing at least a couple of hundred portfiles and a major release, so I don't expect any significant changes.

There is another angle, which is the patches themselves. There might be subtle logic bugs due to incompatibility with newer versions. However, it's not different than the overlay port approach since it can happen there too.

sharder996:
Rebased and cleaned up a few small things I saw. @jimporter how do you propose moving forward on this PR?

xmkg:
@jimporter reverted to the overlay port approach. I think it's not worth messing around with the vcpkg's internals just to make things more convenient for us. We'd be better off with Conan if we're looking for something more flexible.

jimporter (approved):
This looks good to me, and is another big step on the road to consistent dependency management for Multipass!

I'd definitely be interested in seeing whether Conan would let us simplify some of this (especially if it can handle modifying the dep without having to copy all the package manager's portfiles), but that's a question for another day.

PR #4617, opened 2026-01-17
[snap] GUI-less snap packaging

sharder996:


sharder996:
Testing to see if a PR is required for a git branch to show up in Launchpad. Don't think it is, just takes a while for Launchpad to do a `git fetch`.

PR #4631, opened 2026-01-22
[az] Use Boost.JSON for availability zones

jimporter:
# Description

This is just a small PR to update the AZ code to use Boost.JSON where appropriate. As part of this, I also moved `MP_JSONUTILS.read_object_from_file` to `MP_FILEOPS.read_all`, to match the `MP_JSONUTILS.write_json` to `MP_FILEOPS.write_transactionally` migration.

## Testing

All unit tests updated to use Boost.JSON and `MP_FILEOPS.read_all`.

## Checklist

### DIFF
--- a/tests/unit/test_json_utils.cpp
+++ b/tests/unit/test_json_utils.cpp
@@ -36,45 +36,6 @@ using namespace testing;
 
 namespace
 {
-struct TestJsonUtils : public Test
-{
-    mpt::MockFileOps::GuardedMock guarded_mock_file_ops = mpt::MockFileOps::inject();
-    mpt::MockFileOps& mock_file_ops = *guarded_mock_file_ops.first;
-};
-
-TEST_F(TestJsonUtils, readObjectFromFileReadsFromFile)
-{
-    auto filedata = std::make_unique<std::stringstream>();
-    *filedata << "{ \"test\": 123 }";
-
-    EXPECT_CALL(mock_file_ops, open_read(_, _)).WillOnce(Return(std::move(filedata)));
-    const auto json = MP_JSONUTILS.read_object_from_file(":)");
-
-    ASSERT_NE(json.find("test"), json.end());
-    ASSERT_TRUE(json.find("test").value().isDouble());
-    EXPECT_EQ(json.find("test").value().toInt(), 123);
-}
-
-TEST_F(TestJsonUtils, readObjectFromFileThrowsOnFailbit)
-{
-    auto filedata = std::make_unique<std::stringstream>();
-    filedata->setstate(std::ios_base::failbit);
-
-    EXPECT_CALL(mock_file_ops, open_read(_, _)).WillOnce(Return(std::move(filedata)));
-
-    EXPECT_THROW(MP_JSONUTILS.read_object_from_file(":("), std::ios_base::failure);
-}
-
-TEST_F(TestJsonUtils, readObjectFromFileThrowsOnBadbit)
-{
-    auto filedata = std::make_unique<std::stringstream>();
-    filedata->setstate(std::ios_base::badbit);
-
-    EXPECT_CALL(mock_file_ops, open_read(_, _)).WillOnce(Return(std::move(filedata)));
-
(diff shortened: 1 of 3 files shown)


tests/unit/test_json_utils.cpp
```diff
@@ -114,15 +75,15 @@ INSTANTIATE_TEST_SUITE_P(
                                              {"eth2", "52:54:00:00:00:02", false}},
            std::vector<mp::NetworkInterface>{}));
 
-TEST_F(TestJsonUtils, givesNulloptOnEmptyExtraInterfaces)
+TEST(TestJsonUtils, givesNulloptOnEmptyExtraInterfaces)
```
jimporter, line 78: These changes just revert back to what the current state of `main` is. (I had to add `TEST_F` in a previous merge commit to the `availability_zones` branch.)

src/utils/file_ops.cpp
```diff
@@ -337,6 +337,13 @@ std::ifstream& mp::FileOps::read(std::ifstream& file, char* buffer, std::streams
     return file;
 }
 
+std::string mp::FileOps::read_all(const fs::path& filename) const
```
tobe2098: The function name is a bit puzzling. I understand it as "read all the file", but what do you think about other names like `read_into_string`, `read_all_file`, or similar?

jimporter: The name matches `QByteArray FileOps::read_all(QIODevice &file)`, but that'll go away one day (we hope), so we don't *need* to match it. `read_all_file` seems reasonable to me, or maybe `read_file_contents`?

jimporter:
I thought this over some more and after looking again at this code, I think it'd be better to refactor it more heavily than this. There's some surprising logic in here, and my changes so far were just a 1-to-1 translation, instead of taking advantage of `tag_invoke` and the like...

jimporter:
This should be ready for review again. Note: the first commit here is from https://github.com/canonical/multipass/pull/4654, which I cherry-picked into this branch since it'll merge to `main` first (and then I'll merge `main` into the `availability-zones` branch).

src/utils/file_ops.cpp
jimporter: I moved this change over to https://github.com/canonical/multipass/pull/4654, where I'm (for now at least) calling it `try_read_file`. Resolving this comment since it's not a part of this PR anymore.

include/multipass/base_availability_zone_manager.h
```diff
@@ -60,21 +58,14 @@ class BaseAvailabilityZoneManager final : public AvailabilityZoneManager
         mutable std::shared_mutex mutex{};
```
jimporter, line 58: Some commentary: I don't love this code in `BaseAvailabilityZoneManager`. There are two mutexes (or would that be "mutices"?), one here in `ZoneCollection` and one in `BaseAvailabilityZoneManager`. This inner one could probably be removed.

I'm also not sure the way we structure the files for AZs is ideal. Right now, we have one main file that just defines the last-active AZ for the manager, and then one file for each actual AZ. Maybe this has some benefits for reducing lock contention, but I'm not convinced that actually matters. We also save the AZ info to disk immediately upon change, which differs from how, e.g. `AliasDict` works (which saves in its destructor).

I think it would probably be simpler to have a single file defining our known AZs, which we save to disk in the destructor, matching `AliasDict`. That would simplify a lot of this code.

However, all of this is an issue for later, since it's only tangentially related to Boost.JSON. (I'll turn this into a GH issue once I've thought about the details a bit more.)

tobe2098: I cannot comment on the code architecture until I have reviewed it, but I think the matter of the mutex word is somewhat settled through [usage](https://english.stackexchange.com/questions/37455/mutexes-or-mutices) so you can vote by using the one you like the most ;)

tobe2098 (approved):

tobe2098:
I apologize @jimporter , instead of going into the merge queue and rebasing I did a merge commit instead. I restored the changes in the target branch, could you open a new PR to properly merge the changes?

PR #4644, opened 2026-01-27
[ppc64le] Multipass on ppc64le architecture

sharder996:
Multipass on ppc64le architecture. To do this, changes are made in a number of different areas:

- Similar to the port to s390x, we need to explicitly tell vcpkg that we are not cross compiling.
- gcc is very unreliable on ppc64le resulting in internal compiler error, segmentation faults, and infinite loops. We instead force vcpkg to use clang/clang++.

- We already set `CC` and `CXX` in snapcraft to clang, so the clang specific matrix job in the Linux workflow is redundant.

- Create a single source of truth in the distro scraper and only specify mappings for architectures that have different labels.
- Similarly, simplified simplestreams code to only specify mappings for unique architecture label mismatches.

- Adjusted QEMU arguments when running on ppc64le.

---

MULTI-2378


src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -105,10 +104,8 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(
     if (manifest_products_from_official.isEmpty())
         throw mp::GenericManifestException("No products found");
 
-    auto arch = arch_to_manifest.value(QSysInfo::currentCpuArchitecture());
-
-    if (arch.isEmpty())
-        throw mp::GenericManifestException("Unsupported cloud image architecture");
+    auto arch = QSysInfo::currentCpuArchitecture();
+    auto mapped_arch = arch_to_manifest.value(arch, arch);
```
ricab, line 108: 👍

CMakeLists.txt
```diff
@@ -102,7 +104,17 @@ else ()
...
-execute_process(COMMAND ${VCPKG_CMD} RESULT_VARIABLE VCPKG_BOOTSTRAP_RESULT)
+# GCC is unreliable on ppc64le and can cause internal compiler errors during vcpkg bootstrap
+set(VCPKG_BOOTSTRAP_ENV)
+if(${VCPKG_HOST_ARCH} STREQUAL "ppc64le")
+  set(VCPKG_BOOTSTRAP_ENV
+    "CC=clang"
+    "CXX=clang++"
+    "CMAKE_C_COMPILER=clang"
+    "CMAKE_CXX_COMPILER=clang++")
+endif()
```
jimporter, line 115: Could this go in `3rd-party/vcpkg-triplets/ppc64le-linux-release.cmake`, or does that run too late to bootstrap the compiler settings?

jimporter (approved):
This looks good to me. Just one small question below.

(Looking at the distro scraper again, I do wonder if we could make the Debian and Fedora code more similar, but that's a thought for another day...)

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
```diff
@@ -191,15 +191,25 @@ QStringList mp::QemuPlatformDetail::vm_platform_args(const VirtualMachineDescrip
 #elif defined Q_PROCESSOR_S390
         opts << "-machine"
              << "s390-ccw-virtio";
+#elif defined Q_PROCESSOR_POWER
+        opts << "-machine"
+             << "pseries,cap-large-decr=off";
```
tobe2098, line 196: The cap-large-decr flag:
As far as I understand, we are running multipass in a ppc64le machine that already should either support or not support the 64-bit decrement counter. The only possible conflict would be if the host or guest OS does not support it (Linux <4 or so) while the other one does support it, but that would be a Ubuntu version earlier than 18.04, which we do not support.
Did you add it because you found issues on the PowerPC you were using?

tests/unit/qemu/linux/test_qemu_platform_detail.cpp
```diff
@@ -190,8 +190,16 @@ TEST_F(QemuPlatformDetail, platformArgsGenerateNetResourcesRemovesWorksAsExpecte
         "-bios", "QEMU_EFI.fd", "-machine", "virt",
 #elif defined Q_PROCESSOR_S390
         "-machine", "s390-ccw-virtio",
+#elif defined Q_PROCESSOR_POWER
+        "-machine", "pseries,cap-large-decr=off",
```
tobe2098, line 194: Idem

tobe2098 (approved):
LGTM, great work Scott! Just a question on qemu emulation

src/platform/backends/qemu/linux/qemu_platform_detail_linux.cpp
sharder996: Yes, the flag was required in order to launch virtual machines. This was on a Noble runner launching Noble instances.

CMakeLists.txt
sharder996: Yes, we need this in order to compile vcpkg itself. There aren't any distributed binaries for this architecture so it gets compiled on demand.

tobe2098:
On compilation, my Vm fails with the following error:
```
/usr/bin/ld: libdaemon.a(daemon.cpp.o): in function `absl::lts_20250127::log_internal::LogMessage::operator<<(int)':
daemon.cpp:(.text._ZN4absl12lts_2025012712log_internal10LogMessagelsEi[_ZN4absl12lts_2025012712log_internal10LogMessagelsEi]+0x34): undefined reference to `absl::lts_20250127::log_internal::LogMessage& absl::lts_20250127::log_internal::LogMessage::operator<< <int, 0>(int const&)'
/usr/bin/ld: libdaemon.a(daemon.cpp.o): in function `absl::lts_20250127::log_internal::LogMessage::operator<<(unsigned long)':
daemon.cpp:(.text._ZN4absl12lts_2025012712log_internal10LogMessagelsEm[_ZN4absl12lts_2025012712log_internal10LogMessagelsEm]+0x30): undefined reference to `absl::lts_20250127::log_internal::LogMessage& absl::lts_20250127::log_internal::LogMessage::operator<< <unsigned long, 0>(unsigned long const&)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [src/daemon/CMakeFiles/multipassd.dir/build.make:279: bin/multipassd] Error 1
```

sharder996:
@tobe2098 You need to pass `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` when configuring. The snapcraft file will do this automatically BTW, but I didn't want force developer to use clang on development builds in case gcc support gets better in the future.

tobe2098:
Correct, now it compiles. When running tests, interestingly TestPlatformUnix.quitWatchdogQuitsOnSignal fails when running the whole suite, but succeeds when running it via `-gtest_filter`:
```
[ RUN      ] TestPlatformUnix.quitWatchdogQuitsOnSignal
unknown file: Failure
Uninteresting mock function call - taking default action specified at:
/home/ubuntu/multipass/tests/unit/unix/test_platform_unix.cpp:300:
    Function call: pthread_kill(133102238731584, 12)
          Returns: 0

[  FAILED  ] TestPlatformUnix.quitWatchdogQuitsOnSignal (16 ms)
```
This only happens inconsistently. When running the daemon it seems the CLI cannot communicate with the daemon:
```
ubuntu@ubuntu-ppc:~$ bg
[1]+ sudo multipass/build/bin/multipassd --logger stderr --verbosity trace &
ubuntu@ubuntu-ppc:~$ 
ubuntu@ubuntu-ppc:~$ ./multipass/build/bin/multipass ls
list failed: multipass socket access denied
```
Although the daemon runs fine.

sharder996:
@tobe2098 I can investigate further, but I've found that since we never had proper CI for s390x/ppc64le the unit tests were never written or executed with these platforms in mind. Canonical does offer some self-hosted runners in these architectures but the documentation lists them experimental and my testing of them has shown them to be very unreliable.

So, I can look into this unit test, but if the daemon is working properly the only benefit I see is that all unit tests pass when building/running manually.

tobe2098:
I see. I think manual testing is enough for now. What do you think about the CLI not being able to communicate with the daemon?

sharder996:
I'm not sure about that one either. I did see it happening occasionally but only when running the daemon as a "standalone" process. IIRC, it did not show up with the packaged snap.

tobe2098:
How did you build the snap? I tried with snapcraft but the platform seems to not be allowed in snapcraft.yaml, and the snap from this PR does not seem accessible.

sharder996:
You need to merge in these commits: https://github.com/canonical/multipass/compare/cf9c658dd497c5324407c9cd14b41d8608ce7bb2...gui-less-snap

tobe2098:
I have been able to compile most of the snap except for the qemu part: 
[logs.txt](https://github.com/user-attachments/files/25122070/logs.txt)

sharder996:
@tobe2098 Did you do a `snapcraft clean` before packaging? I just tried on a testflinger machine and it packaged successfully.

tobe2098:
Finally managed to package, the commits were lacking setting `clang` as the compiler for the qemu part, or else it does not compile

tobe2098:
The snap seems to work properly but I cannot test VMs because KVM does not run within a non-kvm VM. Approving.

PR #4645, opened 2026-01-29
[ci] Set build params on schedule event

sharder996:
Also set build label on schedule events.

Fixes https://github.com/canonical/multipass/actions/runs/21461340151/job/61814275221

### DIFF
--- a/.github/actions/build-params/action.yml
+++ b/.github/actions/build-params/action.yml
@@ -40,15 +40,8 @@ runs:
         || ${{ github.ref }} =~ ${{ inputs.release-tag-re }} ]]; then
         echo "channel=beta" >> $GITHUB_OUTPUT
 
-      # All other pushes annotate with the CI run number.
+      # All other events annotate with the CI run number.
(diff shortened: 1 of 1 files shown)


.github/actions/build-params/action.yml
```diff
@@ -42,7 +42,9 @@ runs:
 
       # All other pushes annotate with the CI run number.
       # This includes `main`, as `edge` will get published from automatic builds on snapcraft.io.
-      elif [[ ${{ github.event_name }} == "push" || ${{ github.event_name }} == "workflow_dispatch" ]]; then
+      elif [[ ${{ github.event_name }} == "push" \
+        || ${{ github.event_name }} == "workflow_dispatch" \
+        || ${{ github.event_name }} == "schedule" ]]; then
```
ricab: Do you think we should just have this as a catch-all fallback? This would have been useful for a while, to ensure we weren't missing special cases, but these days we could probably just have this as a plain `else` and remove the fall-through check.

sharder996: I don't see that being an issue.

PR #4650, opened 2026-01-29
[az] Fix qemu virtual switch creation and improve Subnet API

jimporter:
# Description

With the introduction of the `Subnet` class in the AZ branch, the qemu backend had a regression (possibly only in some cases) where creating the virtual switch failed. This happened because we incorrectly masked the host IP before we passed it to `ip address add`. To fix this, `Subnet` now retains the original IP address, but there's a new `canonical()` method that returns the canonical form with the IP address masked.

In addition, to help provide more-semantic usage of `Subnet` and move us closer to the `network_v4` class defined in Boost.ASIO and the ISO C++ Networking TS, I added a `broadcast_address()` method (this matches `broadcast()` in the specs). In the medium term, it would be nice to replace some of this common code with Boost.ASIO or (eventually) standard C++ (thanks @ricab for the suggestion).

I'm not sure if we should rename the `Subnet` class now. I think there's an argument that it should be called `CIDR` (classless inter-domain routing), since that's where the IP address + subnet mask concept comes from (and the "192.168.1.1/24" notation is called "CIDR notation"). The naming in our code is a little bit odd, since the stringification method for `Subnet` is `to_cidr`, but since a subnet is an IP address + subnet mask, it's *already* a CIDR block. That function just converts it to the standard *notation* for CIDR blocks.

## Testing

All unit tests updated to match the new API changes. In addition, I've improved the test that should have cause this regression, but which was tautological. The test had the same instructions as the code to be tested, so it would always pass for this issue. Now, we hardcode the expected command in the test, so future regressions here will cause a test failure.

- Manual testing steps:

  1. Start the daemon on Linux: `sudo ./bin/multipassd`
  2. Before this change, the daemon would exit immediately with a status of 1; now, it should run correctly
  3. (optional) Test `multipass launch`, `multipass shell`, etc

## Checklist

### DIFF
--- a/include/multipass/subnet.h
+++ b/include/multipass/subnet.h
@@ -34,6 +34,7 @@ public:
     {
         template <class T>
         explicit PrefixLengthOutOfRange(const T& value)
+            // Subnet masks of /31 or /32 require some special handling that we don't support.
             : FormattedExceptionBase{
                   "Subnet prefix length must be non-negative and less than 31: {}",
                   value}
@@ -63,21 +64,33 @@ public:
 
     Subnet(const std::string& cidr_string);
 
+    // Return the smallest IP address available in this subnet
     [[nodiscard]] IPAddress min_address() const;
+    // Return the largest IP address available in this subnet, excluding the broadcast address
     [[nodiscard]] IPAddress max_address() const;
+    // Return the number of usable IP addresses in this subnet
     [[nodiscard]] uint32_t usable_address_count() const;
 
+    // Return the original IP address
     [[nodiscard]] IPAddress address() const;
+    // Return the IP address with the subnet mask applied
     [[nodiscard]] IPAddress masked_address() const;
+    // Return the broadcast address for this subnet
     [[nodiscard]] IPAddress broadcast_address() const;
+    // Return the prefix length, e.g. the 24 in 192.168.1.0/24
     [[nodiscard]] PrefixLength prefix_length() const;
+    // Return the subnet mask as an IP, e.g. 255.255.255.0
     [[nodiscard]] IPAddress subnet_mask() const;
 
+    // Return this subnet with the subnet mask applied to the IP address
     [[nodiscard]] Subnet canonical() const;
 
+    // Return a string representing this subnet in CIDR notation
     [[nodiscard]] std::string to_cidr() const;
 
+    // Return the number of subnets of size `prefix_length` that fit in this subnet
     [[nodiscard]] size_t size(PrefixLength prefix_length) const;
+    // Get a child subnet that fits inside this one
     [[nodiscard]] Subnet get_specific_subnet(size_t block_idx, PrefixLength prefix_length) const;
 
     // Subnets are either disjoint or the smaller is a subset of the larger
--- a/include/multipass/subnet.h
+++ b/include/multipass/subnet.h
@@ -68,6 +68,7 @@ public:
     [[nodiscard]] uint32_t usable_address_count() const;
 
     [[nodiscard]] IPAddress network_address() const;
+    [[nodiscard]] IPAddress broadcast_address() const;
     [[nodiscard]] PrefixLength prefix_length() const;
     [[nodiscard]] IPAddress subnet_mask() const;
 
--- a/include/multipass/subnet.h
+++ b/include/multipass/subnet.h
@@ -68,7 +68,7 @@ public:
     [[nodiscard]] uint32_t usable_address_count() const;
 
     [[nodiscard]] IPAddress address() const;
-    [[nodiscard]] IPAddress network_address() const;
+    [[nodiscard]] IPAddress masked_address() const;
     [[nodiscard]] IPAddress broadcast_address() const;
     [[nodiscard]] PrefixLength prefix_length() const;
     [[nodiscard]] IPAddress subnet_mask() const;
--- a/include/multipass/subnet.h
+++ b/include/multipass/subnet.h
@@ -67,10 +67,13 @@ public:
     [[nodiscard]] IPAddress max_address() const;
     [[nodiscard]] uint32_t usable_address_count() const;
 
+    [[nodiscard]] IPAddress address() const;
     [[nodiscard]] IPAddress network_address() const;
     [[nodiscard]] IPAddress broadcast_address() const;
     [[nodiscard]] PrefixLength prefix_length() const;
     [[nodiscard]] IPAddress subnet_mask() const;
+
+    [[nodiscard]] Subnet canonical() const;
 
     [[nodiscard]] std::string to_cidr() const;
 
@@ -98,7 +101,7 @@ public:
     }
 
 private:
-    IPAddress address;
+    IPAddress ip_address;
     PrefixLength prefix;
 };
 } // namespace multipass
--- a/src/network/subnet.cpp
+++ b/src/network/subnet.cpp
@@ -78,7 +78,7 @@ mp::Subnet::Subnet(const std::string& cidr_string) : Subnet(parse(cidr_string))
 
 mp::IPAddress mp::Subnet::min_address() const
 {
-    return network_address() + 1;
+    return masked_address() + 1;
 }
 
 mp::IPAddress mp::Subnet::max_address() const
@@ -86,7 +86,7 @@ mp::IPAddress mp::Subnet::max_address() const
     // address + 2^(32 - prefix) - 1 - 1
     // address + 2^(32 - prefix) is the next subnet's network address for this prefix length
     // subtracting 1 to stay in this subnet and another 1 to exclude the broadcast address
-    return network_address() + ((1ull << (32ull - prefix)) - 2ull);
+    return masked_address() + ((1ull << (32ull - prefix)) - 2ull);
 }
 
 uint32_t mp::Subnet::usable_address_count() const
@@ -99,7 +99,7 @@ mp::IPAddress mp::Subnet::address() const
     return ip_address;
 }
 
-mp::IPAddress mp::Subnet::network_address() const
+mp::IPAddress mp::Subnet::masked_address() const
 {
     return apply_mask(ip_address, prefix);
 }
@@ -122,7 +122,7 @@ mp::IPAddress mp::Subnet::subnet_mask() const
 
 mp::Subnet mp::Subnet::canonical() const
 {
-    return Subnet{network_address(), prefix};
+    return Subnet{masked_address(), prefix};
 }
 
 // uses CIDR notation
@@ -161,7 +161,7 @@ mp::Subnet mp::Subnet::get_specific_subnet(size_t subnet_block_idx,
 
     // ex. 192.168.0.0 + (4 * 2^(32 - 24)) = 192.168.0.0 + 1024 = 192.168.4.0
     mp::IPAddress address =
-        network_address() + (subnet_block_idx * (std::size_t{1} << (32 - prefix_length)));
+        masked_address() + (subnet_block_idx * (std::size_t{1} << (32 - prefix_length)));
 
     return mp::Subnet{address, prefix_length};
 }
@@ -172,12 +172,12 @@ bool mp::Subnet::contains(Subnet other) const
     if (other.prefix_length() < prefix)
         return false;
 
-    return contains(other.network_address());
+    return contains(other.masked_address());
 }
 
 bool mp::Subnet::contains(IPAddress ip) const
 {
-    return network_address() <= ip && broadcast_address() >= ip;
+    return masked_address() <= ip && broadcast_address() >= ip;
 }
 
 std::strong_ordering mp::Subnet::operator<=>(const Subnet& other) const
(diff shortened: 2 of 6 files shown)


include/multipass/subnet.h
ricab, line 39: Off-topic, but shouldn't this say "less than 32"?

include/multipass/subnet.h
```diff
@@ -82,8 +86,8 @@ class Subnet
     [[nodiscard]] bool operator==(const Subnet& other) const = default;
 
 private:
-    IPAddress address;
-    PrefixLength prefix;
+    IPAddress address_;
+    PrefixLength prefix_;
```
ricab: New convention for fields? Private fields? I don't mind, but conventions like this aren't worth much unless they're meaning is understood and always followed. Otherwise, readers can't assume that variables without trailing underscores are not fields, for example. So I'd prefer if we discussed in the team and established a convention first. WDYT?

src/network/subnet.cpp
```diff
@@ -64,16 +64,11 @@ catch (const std::out_of_range& e)
 [[nodiscard]] mp::IPAddress apply_mask(mp::IPAddress ip, mp::Subnet::PrefixLength prefix_length)
 {
     const auto mask = get_subnet_mask(prefix_length);
-    for (size_t i = 0; i < ip.octets.size(); ++i)
-    {
-        ip.octets[i] &= mask.octets[i];
-    }
-    return ip;
+    return mp::IPAddress(ip.as_uint32() & mask.as_uint32());
```
ricab: We normally prefer brace-initialization, except when we want a different meaning (e.g. `vector(100)`). See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax as well as the first note in https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

src/network/subnet.cpp
```diff
@@ -64,16 +64,11 @@ catch (const std::out_of_range& e)
...
 }
 } // namespace
 
-mp::Subnet::Subnet(IPAddress ip, PrefixLength prefix_length)
-    : address(apply_mask(ip, prefix_length)), prefix(prefix_length)
+mp::Subnet::Subnet(IPAddress ip, PrefixLength prefix_length) : address_(ip), prefix_(prefix_length)
```
ricab: Idem: brace init

include/multipass/subnet.h
```diff
@@ -65,10 +65,14 @@ class Subnet
     [[nodiscard]] IPAddress max_address() const;
     [[nodiscard]] uint32_t usable_address_count() const;
 
+    [[nodiscard]] IPAddress address() const;
     [[nodiscard]] IPAddress network_address() const;
```
ricab: The difference between these two is not obvious from the names or signatures. Do we really need them both? If so, could you clarify with a comment please? But I see `address()` being used only in a test.

But really, shouldn't we sanitize the IP we get when constructing? Don't we expect 1.2.3.4/16 to be the same as 1.2.0.0/16 and 1.2.21.32/16? Why does the actual address matter here? I am thinking we should zero out the suffix at construction to have a single underlying object representation for each possible subnet. I.e. only ever create canonical objects (what `canonical()` returns today), to have a 1 to 1 mapping between objects and values in our problem domain. That would simplify the interface (fewer methods) and the implementation (e.g. ordering).

src/network/subnet.cpp
```diff
@@ -158,22 +169,21 @@ mp::Subnet mp::Subnet::get_specific_subnet(size_t subnet_block_idx,
...
-    return address <= ip && (max_address() + 1) >= ip;
+    return network_address() <= ip && broadcast_address() >= ip;
 }
 
 std::strong_ordering mp::Subnet::operator<=>(const Subnet& other) const
 {
-    const auto ip_res = address <=> other.address;
-
+    if (const auto ip_res = address_ <=> other.address_; ip_res != 0)
+        return ip_res;
```
ricab: Same concern regarding canonical vs spurious values. Why should 1.2.3.4/24 be ordered after 1.2.3.0/24?

If we used a canonical representation, we could just return the order of maked IPs (assuming 1.2.0.0/16 < 1.2.3.0/24)

ricab (commented):
Thanks @jimporter, great that you spotted, fixed, and properly tested this. I have some concerns about the underlying subnet representation that tie directly into the approach here. Let me know what you think.

include/multipass/subnet.h
jimporter: > The difference between these two is not obvious from the names or signatures. Do we really need them both? If so, could you clarify with a comment please? But I see address() being used only in a test.

I suppose we don't strictly need `address()` today, but for completeness of the `Subnet` interface, I think it's useful to be able to get the original IP address back out. `network_address()` is also only useful internally, so we could make that private (external users could still get the `network_address` via `.canonical().address()`). However, both of these correspond to members in [`boost::asio::ip::network_v4`](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/reference/ip__network_v4.html) (and also the proposed ISO standard [`std::net::ip::network_v4`](https://github.com/cplusplus/networking-ts/blob/2b5b5001b7b7d4b364bb25dae8c2e864165342d9/src/internetprotocol.tex#L2090)), and based on our prior discussion, I tried to nudge this class closer to the Boost/ISO interface.

> But really, shouldn't we sanitize the IP we get when constructing? Don't we expect 1.2.3.4/16 to be the same as 1.2.0.0/16 and 1.2.21.32/16? Why does the actual address matter here?

This behavior (somehow) is necessary, since we need the following line in `qemu_platform_detail_linux.cpp` to return something like `1.2.3.4/16`. It needs to be an IPv4 host plus subnet mask for the `ip(8)` command:

```c++
const auto cidr = mp::Subnet{subnet.min_address(), subnet.prefix_length()}.to_cidr();
```

This also matches the `network_v4` behavior:

```c++
jim@home ~
$ cat network_v4.cpp
#include <iostream>
#include <boost/asio.hpp>

int main() {
  auto net = boost::asio::ip::make_network_v4("1.2.3.4/24");
  std::cout << "net = " << net.to_string() << "\n"
            << "net.address() = " << net.address().to_string() << "\n";
}

jim@home ~
$ g++ network_v4.cpp -pthread -o network_v4

jim@home ~
$ ./network_v4
net = 1.2.3.4/24
net.address() = 1.2.3.4
```

I think there's a strong argument that `Subnet` isn't the right name for this class anymore, but I'm not sure what the right name is. `CIDR` is an option as mentioned above, but I don't know if that's quite right either.

I'd thought about a few other options for how to do this, for example, I could have retained the old `Subnet` behavior where we applied the mask to canonicalize the IP address, and then just constructed the CIDR string for `ip address add` manually. However, I settled on this one because it will make it easier to migrate to Boost.ASIO (and eventually to a future ISO standard networking API) later on.

include/multipass/subnet.h
jimporter: It's correct, though that's just because the `Subnet` class doesn't handle the special cases for subnet masks of `/31` or `/32`. From [Wikipedia](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing):

> [Normally,] a /31 network, with one binary digit in the host identifier, is unusable, as such a subnet would provide no available host addresses after this reduction. RFC 3021 creates an exception to the "host all ones" and "host all zeros" rules to make /31 networks usable for point-to-point links. /32 addresses (single-host network) must be accessed by explicit routing rules, as there is no address available for a gateway.

Since we don't implement any of that logic (it probably wouldn't be useful for us), we just don't support `/31` or `/32` at all.

include/multipass/subnet.h
jimporter: This is just fallout from adding an `address()` member function, and I went with the most obvious alternative at the time. I saw that this style was already used in a couple places, e.g. `src/client/gui/windows/runner/win32_window.h`. I'll just rename this to something more "normal" for now, and we can discuss with the team whether this would be a useful convention.

src/network/subnet.cpp
jimporter: This is also meant to match Boost.ASIO:

```c++
jim@home ~
$ cat network_v4.cpp
#include <iostream>
#include <boost/asio.hpp>

int main() {
  auto net1 = boost::asio::ip::make_network_v4("1.2.3.4/24");
  auto net2 = boost::asio::ip::make_network_v4("1.2.3.0/24");
  std::cout << "net1 == net2: " << (net1 == net2) << "\n";
}

jim@home ~
$ g++ network_v4.cpp -pthread -o network_v4

jim@home ~
$ ./network_v4
net1 == net2: 0
```

However, `boost::asio::ip::network_v4` doesn't support less/greater-than. I'm not sure we really *need* ordering though, so we could just support (in)equality operators here.

include/multipass/subnet.h
jimporter: I added some comments explaining that `/31` and `/32` are special cases that we don't handle.

ricab: Makes sense, thanks.

include/multipass/subnet.h
ricab: OK thanks. FWIW, `src/client/gui/windows/runner/win32_window.h` is not *really* our code - Flutter forces us to vendor it that way (which never felt right to me).

src/network/subnet.cpp
ricab: Thank you!

include/multipass/subnet.h
ricab: OK, I'll accept the argument about mimicking Boost.ASIO. Although a subset of the interface would be enough to achieve that, having 1.2.0.0/16 == 1.2.21.32/16 would break the symmetry, so I agree it makes sense to keep what you have.

I'd still love to see some docs distinguishing these two methods. Or clearer names: perhaps `original_address` and `masked_address`? It's also fine if you want to keep the same naming as boost, but then `network_address` would become just `network`, right?

src/network/subnet.cpp
ricab: Alright, fair enough.

ricab (commented):
Thanks for the changes and comments @jimporter. Only one remaining thread to settle on.

include/multipass/subnet.h
jimporter: @ricab I debated whether to match Boost.ASIO exactly or to keep things closer to how they are now on the AZ branch. Ultimately, I decided on the latter for now, since fully matching Boost.ASIO would have required more extensive changes, like replacing `min_address()` and `max_address()` with a new `address_range` class and replacing the construction from string with a `make_network_v4`-like function.

So I renamed `network_address` to `masked_address` and added some short comments explaining all the member functions. What do you think?

(The comment for `get_specific_subnet` isn't as clear as it could be, but I'm getting rid of that function soon anyway in #4675.)

ricab (commented):
Code looks good, thanks Jim!

I used PR snaps to reproduce the issue and test the fix, which worked 👍 

However, though I could launch new VMs, I couldn't start pre-existing VMs with this PR: Multipass always timed out trying to get into them, as though it couldn't find their IP. I am guessing that problem would have been there already, or is maybe something you'd want to deal with in a separate PR? That would make sense to me. If so, let me know and I can approve this one.

One other AZs remark, unrelated to this PR: I just tried to launch a VM on the CLI and the experience was a little awkward:

```
$ multipass launch --zone=2
launch failed: Invalid arguments supplied
Invalid zone name supplied: 2
$ multipass launch --zone 2
launch failed: Invalid arguments supplied
Invalid zone name supplied: 2
$ multipass launch --zone aiusdf
launch failed: Invalid arguments supplied
Invalid zone name supplied: aiusdf
$ multipass launch --help | grep -A1 zone
  --zone <zone>                The zone in which to launch the instance.
  --mount <source>:<target>    Mount a local directory inside the instance. If
```

I would have expected:

1. the help to explain what to pass to the option;
2. the error to say what I should do instead;
3. `--zone` to accept [1-3], not just zone[1-3].

jimporter:
> Code looks good, thanks Jim!
> 
> I used PR snaps to reproduce the issue and test the fix, which worked 👍

Thanks for testing it out!

> However, though I could launch new VMs, I couldn't start pre-existing VMs with this PR: Multipass always timed out trying to get into them, as though it couldn't find their IP. I am guessing that problem would have been there already, or is maybe something you'd want to deal with in a separate PR? That would make sense to me. If so, let me know and I can approve this one.

You mean pre-existing VMs from the main branch, right? That's on my radar to fix. In the feature flag PR, I've partially fixed it so that pre-existing VMs continue to work when the feature flag is disabled, but we'll need some migration process for when the feature flag is enabled.

> One other AZs remark, unrelated to this PR: I just tried to launch a VM on the CLI and the experience was a little awkward:
> 
> ```
> $ multipass launch --zone=2
> launch failed: Invalid arguments supplied
> Invalid zone name supplied: 2
> ```

Good point. We could look up zones by name or by index (starting at 1) to make this behave more nicely. I filed #4692 to track this.

ricab:
I am not sure what version I'd used to create the VMs to be honest. It might have been stable, but I am not certain.

ricab (approved):
Alright, LGTM!

PR #4651, opened 2026-01-29
[tests] Regenerate SSL certificates

sharder996:
Will expire in 2.9 million days.


jimporter (approved):
Looks good to me. Feel free to merge once this passes CI. (I think this is simple enough that it shouldn't require a secondary reviewer.)

PR #4652, opened 2026-01-29
[applevz] Add availability zones to AppleVZ

jimporter:
This is just a small fix to the AZ branch after merging in the AppleVZ stubs.


jimporter:
Small enough for a single review I think, especially given that it's code behind a feature flag.

sharder996 (approved):
Yup, LGTM

PR #4657, opened 2026-02-03
Improve ssh error logging

### LINKED ISSUE #4584: Improve logging when ssh_exec fails
While attempting to debug events related to failed ssh_execs in waiting for cloud-init, logs show an empty error message:
```
[utils] failed to run '[ -e /var/lib/cloud/instance/boot-finished ]', error message: ''
```

Track where is the exception originating from and improve how much information is made available in the exception message.

deepakshirkem:
# Description

## What does this PR do?
Improves error logging when SSH command execution fails with empty stderr. Previously, when SSH channels closed before stderr could be read, error messages were logged as empty strings (`error message: ''`), making debugging difficult. This PR ensures meaningful error information (exit code and context) is always provided.

## Why is this change needed?
During cloud-init waiting and other SSH operations, commands often fail with closed channels that produce no stderr output. The empty error messages made it impossible to diagnose issues. With this change, developers and users get actionable information even when stderr is unavailable.

## Related Issue(s)
Closes #4584

### Manual testing steps:
1. Build multipass from source: `cmake .. && make`
2. Start daemon with trace logging: `sudo ./bin/multipassd --verbosity trace 2>&1 | tee daemon.log &`
3. Launch a VM: `./bin/multipass launch --name test-vm`
4. Monitor logs during cloud-init: `sudo journalctl -f | grep "failed to run"`
5. Observe error messages now include exit codes instead of empty strings

### Root Cause Investigation
The empty error occurs in `src/ssh/ssh_process.cpp:184-189` where `read_std_error()` legitimately returns an empty string when the SSH channel is closed. This commonly happens during:
- Cloud-init boot-finished checks (`[ -e /var/lib/cloud/instance/boot-finished ]`)
- File existence checks on VMs that are still booting
- Commands that exit quickly before the channel can capture stderr

## Screenshots

**Before Fix:**
Before

**After Fix:**
Screenshot from 2026-02-02 21-28-48

## Checklist

### Changes Made:
- Modified `src/utils/utils.cpp` in the `run_in_ssh_session` function
- Added check for empty `error_msg` after reading stderr
- When empty, provides exit code and explanatory context
- Improved debug logging to distinguish between empty and non-empty stderr cases

### Example Output:
**Before:**
```
[utils] failed to run '[ -e /var/lib/cloud/instance/boot-finished ]', error message: ''
```

**After:**
```
[utils] failed to run '[ -e /var/lib/cloud/instance/boot-finished ]', exit_code 1 (no stderr output)
```

Exception message now contains `"Command exited with code 1"` providing actionable debugging information.

The fix is minimal and focused on the specific issue identified, ensuring all SSH execution failures have meaningful error messages for debugging.

### DIFF
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -288,10 +288,18 @@ std::string mp::Utils::run_in_ssh_session(mp::SSHSession& session,
 {
     auto proc = session.exec(cmd, whisper);
 
-    if (auto ec = proc.exit_code() != 0)
+    if (auto ec = proc.exit_code(); ec != 0)
     {
         auto error_msg = mp::utils::trim_end(proc.read_std_error());
-        mpl::debug(category, "failed to run '{}', error message: '{}'", cmd, error_msg);
+        if (error_msg.empty())
+        {
+            mpl::debug(category, "failed to run '{}', exit_code {} (no stderr output)", cmd, ec);
+        }
+        else
+        {
+            mpl::debug(category, "failed to run '{}', error message: '{}'", cmd, error_msg);
+        }
+
         throw mp::SSHExecFailure(error_msg, ec);
     }
 


tobe2098:
Hi @deepakshirkem , thank you. Could you remove the merge commits and rebase from main only if you need to? The merge commits are excessive clutter for multipass' commit history.

src/utils/utils.cpp
```diff
@@ -288,10 +288,21 @@ std::string mp::Utils::run_in_ssh_session(mp::SSHSession& session,
...
-    if (auto ec = proc.exit_code() != 0)
+    if (auto ec = proc.exit_code(); ec != 0)
     {
         auto error_msg = mp::utils::trim_end(proc.read_std_error());
-        mpl::debug(category, "failed to run '{}', error message: '{}'", cmd, error_msg);
+
+        if(error_msg.empty())
+        {
+            mpl::debug(category, "failed to run '{}', exit_code {} (no stderr output)", cmd, ec);
+            error_msg = fmt::format("Command exited with code {}", ec);
```
jimporter: I think we should do something different with this line (though there are a few options). This will affect the output of `SSHExecFailure::what` to include this message, but in `src/daemon/daemon.cpp` when catching this exception, we use it to construct a `grpc::Status` object that always includes the exit code anyway. The value of `what()` is passed into `grpc::Status` as the `error_details`, and right now those details are just the stderr output, which I think is what we want.

The simplest fix would be to remove this line, but one other options would instead be to adjust `SSHExecFailure` to have some new `stderr` method that we use in `daemon.cpp`, and to then have `SSHExecFailure::what` return a message like this.

jimporter (changes requested):
Thanks for taking a look at this. Most of this looks like it should improve the logging, but I see one issue below with how we construct the `SSHExecFailure` exception.

src/utils/utils.cpp
deepakshirkem: Hi @jimporter,

Thank you for the detailed feedback — that makes sense.

Since SSHExecFailure::what() is used as the grpc::Status error_details in daemon.cpp, and the exit code is already propagated separately, adding a synthetic message when stderr is empty would change the meaning of error_details.

To keep the behavior unchanged and avoid affecting gRPC consumers, my plan is to not modify error_msg and keep what() strictly tied to stderr output, while relying on the debug log to record the exit code when stderr is empty.

Does that sound reasonable, or would you prefer a different approach?

tobe2098:
Thank you for the corrections @deepakshirkem , but could you look at the Check Commits workflow? The formatting of commit messages does not follow our standard.

You would also need to sign the CLA: https://canonical.com/legal/contributors

src/utils/utils.cpp
jimporter: @deepakshirkem That sounds good to me, thanks!

deepakshirkem:
Hi @tobe2098 

I've fixed the commit message - it was just 1 character over the 50 limit. I've shortened it and force-pushed the update.

I've also signed the CLA now, so that should be resolved.

Apologies for the multiple iterations - this is my first time contributing to this repository and I'm still learning the contribution workflow. I appreciate your patience and guidance!

Thanks You!

tobe2098 (approved):

jimporter (approved):
This all looks good to me now. Thanks for your contribution to Multipass!

PR #4663, opened 2026-02-05
[cloud-init] Fix image resizing on Ubuntu Core images

### LINKED ISSUE #3229: Disk resizing not working for core20/22 images
**Describe the bug**
After attempting to grow the disk of a core20- or core22-based instance, no partition is grown to reflect the new size.

```
$ multipass exec -n a -- sudo fdisk -l /dev/sda
GPT PMBR size mismatch (10485759 != 14680063) will be corrected by write.
The backup GPT table is not on the end of the device.
Disk /dev/sda: 7 GiB, 7516192768 bytes, 14680064 sectors
Disk model: QEMU HARDDISK
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: A63BB26A-5487-41C0-AF40-19150D00E322

Device       Start      End Sectors  Size Type
/dev/sda1     2048     4095    2048    1M BIOS boot
/dev/sda2     4096  2461695 2457600  1.2G EFI System
/dev/sda3  2461696  3997695 1536000  750M Linux filesystem
/dev/sda4  3997696  4063231   65536   32M Linux filesystem
/dev/sda5  4063232 10485726 6422495  3.1G Linux filesystem
```

Note: tried with `core` as well, where the writable partition was properly resized, although disk reporting in `info` is wonky (13.7GiB).

**To Reproduce**
How, and what happened?
1. `multipass launch -n a core22`
1. `multipass stop a`
1. `multipass set local.a.disk=7G`
2. `multipass start a`
3. `multipass info a | grep Disk`
    - observe reported disk space is still around 4.8G 
1. `multipass exec -n a -- sudo fdisk -l /dev/sda`

**Expected behavior**
The writable partition (`/dev/sda5` above) would grow to use the available disk space.

**Logs**
```
set 15 22:24:13 ricab-laptop multipassd[173643]: Succeeded setting local.a.disk=7G
```

**Additional info**
 - OS: Ubuntu
- `multipass version`: 1.12.2
- `multipass info --all`
```
Name:           a
State:          Running
IPv4:           10.195.134.80
Release:        Ubuntu Core 22
Image hash:     1493fccb9d55 (Ubuntu Core 22)
CPU(s):         1
Load:           0.00 0.00 0.00
Disk usage:     1.4GiB out of 4.8GiB
Memory usage:   84.2MiB out of 951.9MiB
Mounts:         --

Name:           b
State:          Running
IPv4:           10.195.134.83
Release:        Ubuntu Core 16
Image hash:     db3a6130f46f (Ubuntu Core 16)
CPU(s):         1
Load:           0.00 0.00 0.00
Disk usage:     1.1GiB out of 13.7GiB
Memory usage:   32.2MiB out of 986.4MiB
Mounts:         --

Name:           c
State:          Running
IPv4:           10.195.134.175
Release:        Ubuntu Core 20
Image hash:     4692ea67d73e (Ubuntu Core 20)
CPU(s):         1
Load:           0.02 0.02 0.00
Disk usage:     1.0GiB out of 4.8GiB
Memory usage:   97.2MiB out of 971.2MiB
Mounts:         --
```

tobe2098:
# Description

- What does this PR do? Fixes an issue relating to ubuntu core image resizing. The change makes sure to notify the user when resizing a disk from an Ubuntu Core OS for the next necessary steps. Additionally, VirtualMachineDescription was moved from being held in the backend-specific VM classes to the BaseVirtualMachine class, and additional maintenance on the logging/messaging side of grpc replies and requests has been done.
- Why is this change needed? The disk resize cannot be done by us, as we would be altering how Ubuntu Core behaves, but the user may not be aware of what is happening and think it is a Multipass bug. The PR addresses these cases. We now also have a standardized way of returning messages back to the daemon with minimum disruption. 

## Related Issue(s)

Closes #3229

## Testing

- Unit tests
- Manual testing steps:

  1. Launch a core20/22/24 image `multipass launch -n a core:core22`
  2. Stop the instance `multipass stop a`
  3. Resize the instance with `multipass set local.a.disk=7G`
  4. The message appears on screen

## Checklist

MULTI-2227


sharder996 (changes requested):
```
scott@Scotts-MacBook-Pro build-gui % multipass stop core
scott@Scotts-MacBook-Pro build-gui % multipass set local.core.disk=10G
scott@Scotts-MacBook-Pro build-gui % multipass start core
scott@Scotts-MacBook-Pro build-gui % multipass info
Name:           core
State:          Running
Snapshots:      0
IPv4:           192.168.2.206
Release:        Ubuntu Core 24
Image hash:     29d2791cdce3 (Ubuntu Core 24)
CPU(s):         1
Load:           0.77 0.18 0.06
Disk usage:     957.9MiB out of 4.8GiB
Memory usage:   181.6MiB out of 953.2MiB
Mounts:         --
scott@Scotts-MacBook-Pro build-gui % multipass exec core -- df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda4       3.0G  442M  2.4G  16% /writable
tmpfs           191M  5.9M  185M   4% /run
tmpfs           477M     0  477M   0% /dev/shm
efivarfs        256K  2.4K  254K   1% /sys/firmware/efi/efivars
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda2       721M   84M  585M  13% /run/mnt/ubuntu-boot
/dev/sda1       1.2G  433M  749M  37% /boot/efi
/dev/sda3        26M   49K   23M   1% /var/lib/snapd/save
tmpfs           477M     0  477M   0% /mnt
tmpfs           477M     0  477M   0% /media
tmpfs           477M     0  477M   0% /tmp
/dev/sda4       3.0G  442M  2.4G  16% /var/log
tmpfs           477M     0  477M   0% /var/lib/sudo
tmpfs            96M  4.0K   96M   1% /run/user/1000
scott@Scotts-MacBook-Pro build-gui % multipass info core
Name:           core
State:          Running
Snapshots:      0
IPv4:           192.168.2.206
Release:        Ubuntu Core 24
Image hash:     29d2791cdce3 (Ubuntu Core 24)
CPU(s):         1
Load:           0.47 0.16 0.06
Disk usage:     957.9MiB out of 4.8GiB
Memory usage:   173.0MiB out of 953.2MiB
Mounts:         --
scott@Scotts-MacBook-Pro build-gui % multipass get local.core.disk
10.0GiB
```

Best of luck on your debugging journey 😄

tobe2098:
Now it should work. Apparently resize2fs only runs on the root partition and cannot be changed like growpart. Launching a new core image and resizing will work, although it is not instant, it relies on the script running on boot.

sharder996:
Still no luck after resizing to 10G:
```
ubuntu@foo:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           482M     0  482M   0% /dev/shm
tmpfs           193M  5.8M  187M   3% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda2       721M  145M  524M  22% /run/mnt/ubuntu-boot
/dev/sda1       1.2G  413M  769M  35% /boot/efi
/dev/sda4       3.0G  849M  2.0G  30% /writable
/dev/sda3        26M   48K   23M   1% /var/lib/snapd/save
tmpfs           482M     0  482M   0% /media
tmpfs           482M     0  482M   0% /mnt
tmpfs           482M     0  482M   0% /tmp
tmpfs           482M     0  482M   0% /var/lib/sudo
tmpfs            97M     0   97M   0% /run/user/1000
ubuntu@foo:~$ sudo fdisk -l /dev/sda
GPT PMBR size mismatch (10485759 != 20971519) will be corrected by write.
The backup GPT table is not on the end of the device.
Disk /dev/sda: 10 GiB, 10737418240 bytes, 20971520 sectors
Disk model: QEMU HARDDISK   
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 7CEC9013-A76C-400F-BD83-EBF89CCFC0C3

Device       Start      End Sectors  Size Type
/dev/sda1     2048  2459647 2457600  1.2G EFI System
/dev/sda2  2459648  3995647 1536000  750M Linux filesystem
/dev/sda3  3995648  4061183   65536   32M Linux filesystem
/dev/sda4  4061184 10485726 6424543  3.1G Linux filesystem
ubuntu@foo:~$ exit
logout
scott@Scotts-MacBook-Pro build-gui % multipass version
multipass   1.17.0-dev.1536+g76a09f32.mac
multipassd  1.17.0-dev.1536+g76a09f32.mac
```

tobe2098:
Did you create a new core VM?  The fix only works in  instances created with the new version.

tobe2098:
To see if it is another issue, create this file in your instance:
```
ubuntu@a:~$ cat /var/lib/cloud/scripts/per-boot/01-resize-data.sh 
#!/bin/bash
#multipass/version/1.17.0-dev.1536+g1832b5f4 # written by Multipass
#multipass/driver/qemu-8.2.2 # written by Multipass
PART_PATH=$(blkid -L ubuntu-data)
if [ -z "$PART_PATH" ]; then
    echo "Resize script: ubuntu-data partition not found."
    exit 0
fi
PARENT_DISK="/dev/$(lsblk -no pkname "$PART_PATH" | head -n 1)"
PART_NUM=$(lsblk -no partn "$PART_PATH")
growpart "$PARENT_DISK" "$PART_NUM" || true
resize2fs "$PART_PATH"
```
(with 0755 permissions)
The fix only injects the script in new instances, and it is a bit delayed after boot.

tobe2098:
Turns out `lsblk` does not have a feature from the script in `core22`, now it should work (in a new instance). For the fix in an existing instance just copy the script:
```
└─[λ] multipass exec a -- cat /var/lib/cloud/scripts/per-boot/01-resize-data.sh   
#!/bin/bash
#multipass/version/1.17.0-dev.1536+g0997a72b # written by Multipass
#multipass/driver/qemu-8.2.2 # written by Multipass
PART_PATH=$(blkid -L ubuntu-data)
if [ -z "$PART_PATH" ]; then
    echo "Resize script: ubuntu-data partition not found."
    exit 0
fi
PARENT_DISK="/dev/$(lsblk -no pkname "$PART_PATH" | head -n 1)"
PART_NUM=$(echo $PART_PATH | grep -o "[0-9]*$")
growpart "$PARENT_DISK" "$PART_NUM" || true
resize2fs "$PART_PATH"
```
Onto the per-boot location and restart or run it manually with `sudo`

Fix works in all `core` versions

src/daemon/daemon.cpp
```diff
@@ -3562,6 +3562,11 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
 
             vm->wait_for_cloud_init(timeout);
         }
+        else
```
sharder996: This should only be operating on Ubuntu core instances.

sharder996 (changes requested):
This does work properly and resizes the disk on Ubuntu Core images, but I'm unsure whether or not its the right approach.

Here's a couple thoughts I had which drive my reasoning:
1. This is only necessary for Ubuntu Core and in all other cases I think it would be better to rely on the tooling that drives `growpart`. Having the resize script/function be a part of the VirtualMachine class goes against this.
2. Proceeding the previous point, I feel like there is a better way or achitecturing this. Here's one idea I had:
- We could have the VM keep track when its disk was resized and it could apply the resize function itself. The daemon wouldn't have to get involved.

I'd love to here what @jimporter thinks, too.

tobe2098:
On the first point, I would add that there are cases outside of Ubuntu Core where this matters: whenever the user disables cloud init themselves (I tested this on 24.04). In those cases multipass can not resize the images properly either. I am unsure if this responsibility should fall on multipass though.

On the second point, I originally implemented it as per your suggestion. I later went back for two reasons: first, the operation, even when needed, takes negligible time on the start operation, and second, it would need an extra field in the persistent json to avoid inconsistencies across daemon restarts. An elegant solution would be to set the boolean to true on initialization, since a single resize is cheap, WDYT?

If it is the best option, we can go for said implementation, having to run that small script either way does feel like there should be a better option,  but the auto-disable of cloud-init does limit possible solutions.

jimporter:
> 2. Proceeding the previous point, I feel like there is a better way or achitecturing this. Here's one idea I had:
> 
> 
>     * We could have the VM keep track when its disk was resized and it could apply the resize function itself. The daemon wouldn't have to get involved.

I agree with @sharder996 that this seems like an architecturally-cleaner solution, and would make it easier to support a wider variety of guest images one day. However, I haven't thought this through far enough to have a clear idea of what exactly that other architecture would look like; for example, how do we tell the VM to do that, and does it work with custom images?

To help answer this question, here's a thought experiment: are there any other things that a user could change that require the guest to run some script to update things? Maybe changing the bridge subnet? (We might want to do this if the host OS's network configuration changes and we end up having a conflict, e.g. as in #3147.) If changing the bridge subnet, we might *need* for the guest OS to handle this itself, or maybe this isn't even a problem. I haven't started work on a fix for #3147, so I'm not sure yet.

Ultimately, my decision would come down to a couple things. First, whether there's a clear example of some change-synchronization action like this that *needs* to be done in either the host or the guest. If there's no example like that, then whether one or the other option is significantly simpler to implement. As much as clean architecture is important for making future changes easier, it's even more important not to overcomplicate an architecture for hypothetical changes we can't even clearly define.

I know none of that is really a definite answer one way or the other on this PR though. Would this be a useful topic for us to discuss at one of the team meetings?

tobe2098:
> Would this be a useful topic for us to discuss at one of the team meetings?

I put it in the discussion list.

tobe2098:
I have applied the change you suggested Scott. It still does not resolve the daemon vs vm encapsulation. The daemon seems to still be involved, since when starting a VM all functions are called from the daemon (like wait_for_cloud_init). We can discuss this further if you would prefer a different approach.

ricab:
I chanced upon the cloud-init while looking at other stuff. Did you try adding `/writable` to https://github.com/canonical/multipass/blob/eb05c4cc4fb9f583e863b05fb859564c31d62c29/src/daemon/base_cloud_init_config.h#L23?

tobe2098:
Yes, that was one of the first solutions (`/dev/disk/by-label/ubuntu-data`) but after the first boot/login cloud-init is disabled and does not run this anymore. It is also only a partial solution, since you still need to run `resize2fs` on the partition that was `grow-part`ed

tobe2098:
The new approach has been implemented. Tagging @amylily1011 for discussion about the returned message to the user to prompt them to take action (run `growpart` and `resize2fs` inside the instance).

include/multipass/exceptions/settings_exceptions.h
```diff
@@ -72,4 +72,10 @@ class UnsupportedSettingValueType : public SettingsException
     }
 };
 
+class ResizingUbuntuCoreInstance : public SettingsException
```
jimporter: At first glance, this name sounds like it's describing a successful resize operation that's in-progress. Since there might be other cases like this in the future (the user changes some setting, and it only partially succeeds), how about naming it something like this?

```suggestion
class SettingsChangeWarning : public SettingsException
```

Or any other variation on this, like `SettingsChangePartialSuccess`.

jimporter (changes requested):
This looks good to me, except for a minor naming comment I have about the new exception.

include/multipass/exceptions/settings_exceptions.h
tobe2098: Agreed, `SettingsChangePartialSuccess` is perfect I think

tobe2098:
Additional context: `snapcraft:corexx` behaves as an Ubuntu XX.04 LTS, not as a core image. Resizing behaves normally. 

Another possible approach is to have the daemon write on the VMDescription whether the image is core or not from the query, instead of depending on `desc.original_release`, but that would require the daemon to do additional work.

amylily1011:
@tobe2098 @jimporter What would be the best way for me to check what are the current messages right now and the current outputs? Are there any edge cases that I should look into or think about?

tobe2098:
When successfully setting a property we do not have an output. In the case the user resizes an Ubuntu Core image the message is:
```
Resizing an Ubuntu Core image requires additional steps be performed within the instance.
See `man growpart` and `man resize2fs`.
```

amylily1011:
@tobe2098 can multipass determine the device path and partition number at runtime (and substitute them in )?
(If this is possible, this will eliminate the biggest failure point. )

1. I think there should be a success message as confirmation unless the user is on quiet mode. In a normal mode, I would suggest the success message as below. 
```
>$ multipass set local.<instance>.disk.    //Resize instance

//Output on success
>$  Disk resized. To make the new space available on this Ubuntu Core instance, run the following commands inside the instance:
           multipass shell <instance name>
           
           sudo growpart /dev/sda 3
           sudo resize2fs /dev/sda3       

//In case we cannot fetch the success status, show this message instead.
     Check the resize status with the command below:
           multipass info <instance>
         
```
Optional implementation:
Then after the user runs `growpart` + `resize2fs` inside the instance

`resize2fs` can print the output when it finishes. (This should be enough to close the loop.)

```
//On the other terminal that the user runs `resize2fs`
resize2fs 1.46.5
The filesystem on /dev/sda3 is now 5242880 (4k) blocks long.

//On the multipass terminal
Disk resized. 
```
An alternative option, where their can be a repeat resize, we might want to show the output as below: ( I don't think this is necessary, but I want to point out that this can be the case.)

```
resize2fs 1.46.5
The filesystem on /dev/sda3 is now 5242880 (4k) blocks long.

Disk resized. To make the new space available on this Ubuntu Core instance, run the following commands inside the instance:
           multipass shell <instance name>
           
           sudo growpart /dev/sda 3
           sudo resize2fs /dev/sda3                   # prints confirmation when complete
         
```

2. For failed message, pointing the user to the man page seems like a inefficient output.  I think we should reduce the friction here.

```
//Output on failure.
>$ Disk resized at the partition level. Ubuntu Core images require two manual steps inside the instance to make the new space available (cloud-init is disabled after first boot).

Open a shell:
         multipass shell <instance name>

Then run in order:

         sudo growpart /dev/sda 3
         sudo resize2fs /dev/sda3

```

If device path can't be determined, add: ( to the bottom)
```
Run lsblk inside the instance to identify the disk partition.
```

3. On a repeated resize, can Multipass detect whether `resize2fs` was already run previously? I think this can help avoid confusion.

tobe2098:
Multipass could determine the partition number, but the partition is predictable (`/dev/sda5`), we could do:
```
Disk resized. To make the new space available on this Ubuntu Core instance, run the following commands inside the instance:
           multipass shell <instance name>
           
           sudo growpart /dev/sda 5              #Or partition with /writable
           sudo resize2fs /dev/sda5                  
```
The repeat resizes is not something we keep track of. The resize either fails completely (other error message) or succeeds partially (message above). WDYT?

On the resize2fs output, did you mean that multipass should output an additional message after running it?

amylily1011:
@tobe2098 Brilliant! We can disregard the repeatable one. 

As for the `resize2fs` output, yes we will separate this into 2 parts. 
The first part is immediately after running `multipass set local.<instance>.disk`
and the second part is after the user runs `growpart` + `resize2fs` inside the instance

tobe2098:
Getting output after running a shell command within multipass is outside the scope of what we are doing unless I am mistaken. Running commands within `shell` is a separate session, and `exec` is usually unfiltered although we could theoretically parse it. Is this what you were suggesting? I understand we want the user to know when they have succeeded, but a simpler way would be to check the size via `multipass info instance` or `sudo fdisk -l`.

amylily1011:
My understanding is that multipass can't detect the `exec` command but we can print output of `resize2fs` on completion. Or is that out of scope as well?

tobe2098:
After some discussion, and due to limitations of how we handle RPC status messaging (set failed is required for message printing), this is the current output message: 
```
set failed: Disk resized. To make the new space available on this Ubuntu Core instance, run the following commands

        multipass exec <instance> -- sudo growpart /dev/sda 5
        multipass exec <instance> -- sudo resize2fs /dev/sda5

Check the resize status with the command below:

         multipass info <instance>
```

tobe2098:
This PR is now ready for review

sharder996 (changes requested):
Not sure if it varies based on the release of Ubuntu Core, but I needed to grow the 4th partition, i.e. `sda5` doesn't exist:

```
(.venv) scott@Scotts-MacBook-Pro build-gui % multipass set local.bar.disk=15G
Disk resized. To make the new space available on this Ubuntu Core instance, run the following commands:

        multipass exec <instance> -- sudo growpart /dev/sda 5
        multipass exec <instance> -- sudo resize2fs /dev/sda5

Check the resize status with the command below:

         multipass info <instance>

(.venv) scott@Scotts-MacBook-Pro build-gui % multipass exec bar -- df -h     
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           482M     0  482M   0% /dev/shm
tmpfs           193M  3.4M  190M   2% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda2       721M  145M  524M  22% /run/mnt/ubuntu-boot
/dev/sda1       1.2G  413M  769M  35% /boot/efi
/dev/sda4       3.0G  807M  2.0G  29% /writable
/dev/sda3        26M   48K   23M   1% /var/lib/snapd/save
tmpfs           482M     0  482M   0% /media
tmpfs           482M     0  482M   0% /mnt
tmpfs           482M     0  482M   0% /tmp
tmpfs           482M     0  482M   0% /var/lib/sudo
tmpfs            97M     0   97M   0% /run/user/1000
```

Are we able to programatically tell the user to select the correct partition? Or, maybe edit the messaging to tell the user to select the correct partition? Also, I wonder if maybe it would be better to append the two commands together, i.e. `sudo growpart /dev/sda 5 && resize2fs /dev/sda5` so a user can quickly copy and paste the command. Summoning @geoffreynyaga to see if he has any insight.

tobe2098:
We can edit it to tell the user to find the partition number with `lsblk`, but programatically it cannot be done since the instance must be stopped while the resize happens.

tobe2098:
What do you think about:
```
Disk resized. To make the new space available on this Ubuntu Core instance, find the `/writable` partition using `lsblk` and run the following commands:

        multipass exec <instance> -- sudo growpart /dev/<part_name> <part_num>
        multipass exec <instance> -- sudo resize2fs /dev/<part_name+part_num>

Check the resize status with the command below:

         multipass info <instance>
```

sharder996:
What do you think of this?

```
growpart /dev/<disk> <partition_number>
resize2fs /dev/<disk><partition_number>
```

Another option I was contemplating for `<disk>` was `<block_device>`...

tobe2098:
I think that one is better, I will add it

include/multipass/qualified_return_value.h
```diff
@@ -0,0 +1,283 @@
...
+    // Disable copying to avoid string copies
+    MessageBag(const MessageBag&) = delete;
+    MessageBag& operator=(const MessageBag&) = delete;
+
+    MessageBag(MessageBag&&) = default;
+    MessageBag& operator=(MessageBag&&) = default;
+
+private:
+    // The message vector is wrapped to minimize the API surface
+    std::vector<std::string> _messages;
```
sharder996: Why is this a vector and not just created by appending strings together?

src/client/cli/cmd/common_callbacks.h
```diff
@@ -96,6 +103,11 @@ auto make_confirmation_callback(Terminal& term, QString key)
 
             client->Write(request);
         }
+        // If the reply does not contain the reply authorization, it contains a user message
+        else if (!reply.reply_message().empty())
+        {
+            term.cout() << reply.reply_message() << '\n';
```
sharder996: Why does this need an extra newline?

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -659,6 +659,10 @@ void HCSVirtualMachine::resize_disk(const MemorySize& new_size)
         throw ResizeDiskException{"Disk resize failed, details: {}", result};
     }
     description.disk_space = new_size;
+    if (is_core())
```
sharder996: Can this be moved to the base case by making use of the Template Method pattern?

src/platform/backends/shared/base_virtual_machine.h
```diff
@@ -40,10 +41,12 @@ class BaseVirtualMachine : public VirtualMachine
 public:
     BaseVirtualMachine(VirtualMachine::State state,
                        const std::string& vm_name,
```
sharder996, line 43: If we are already passing the `VirtualMachineDescription`, can we omit the `vm_name` since its already part of the description?

sharder996 (changes requested):
I have a couple comments, but I feel I don't have the knowledge of C++ to give valuable feedback on the new `Qualified` type. I want to invoke someone more experienced such as @xmkg, @ricab, or @jimporter.

src/platform/backends/shared/base_virtual_machine.h
```diff
@@ -234,3 +240,21 @@ inline void multipass::BaseVirtualMachine::refresh_start()
...
+
+inline std::string multipass::BaseVirtualMachine::core_image_disk_resize_message() const
+{
+    return std::string("Disk resized. To make the new space available on this Ubuntu Core "
+                       "instance, use lsblk to find the /writable partition and run the "
+                       "following commands:\n\n"
+                       "        multipass exec <instance> -- sudo growpart /dev/<block_device> "
+                       "<partition_number>\n"
+                       "        multipass exec <instance> -- sudo resize2fs "
+                       "/dev/<block_device><partition_number>\n\n"
```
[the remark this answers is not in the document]
tobe2098: This makes sense, block device is misleading in this case, disk device and partition number are correct in case the user wants to search more information about what is which of them.

include/multipass/qualified_return_value.h
tobe2098: Because that would make it much harder to process messages in the future. Although in some cases we just want to append them together, or print them one after the other, this approach is more flexible in case we want to filter out some messages or only print a particular one (first, last, ...). This also leaves the inter-message formatting to the callback function, instead of putting the responsibility in a class that is only supposed to carry the messages from A to B.

src/client/cli/cmd/common_callbacks.h
tobe2098: Sometimes it may, but only if there is no newline in the message already.

src/platform/backends/shared/base_virtual_machine.h
tobe2098: I had the same thought at first, but: 1) That would require a larger refactor that is out of scope for this PR, and 2) the VMDescription name field is non-const, while the vm_name field is const because we rely on the name for IDing the VM. They have different contracts, and until the renaming feature, I believe we should leave it as is. 
WDYT?

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
tobe2098: I agree, done.

include/multipass/qualified_return_value.h
```diff
@@ -0,0 +1,285 @@
...
+    // To allow usage of merge()
+    template <Qualifiable T>
+    friend class Qualified;
+};
+
+template <Qualifiable T>
+class [[nodiscard("You must collect() the value or bind it to retrieve the MessageBag")]] Qualified
+{
+public:
+    Qualified(T value, MessageBag messages)
```
xmkg: Just a drive-by comment: The name `Qualified` gave me the impression that this is a wrapper to somehow convey the qualification, as in keyword-sense, like const or volatile. I don't know if it's just me, though. I have an alternative name for this for your consideration: `Annotated`.

tobe2098: Yes, perhaps it is better. `Annotated` still does not feel like it completely fits, but it is better.

include/multipass/qualified_return_value.h
sharder996: Ok, thanks for the explanation!

src/platform/backends/shared/base_virtual_machine.h
sharder996: Sure, I'm fine with that. Just wanted to make sure it had been considered.

sharder996 (dismissed):
I don't have any more comments, but I'd still like to get another review on the new ~~`Qualified`~~`Annotated` type. Feel free to request your favourite reviewer for that 😄

tobe2098:
@jimporter was the original second reviewer, so I will leave it up to him :D

ricab:
FWIW, this just changed status from "Assessment" to "Incoming": https://warthogs.atlassian.net/browse/PF-5210

Perhaps there should be a comment in the code that this can be dropped once that is fixed, so that we check back in the future. WDYT?

ricab:
OTOH, I don't know if this will affect existing core images or only new ones...

tobe2098:
It could also not be enough for us if there are still manual steps required (the mentioned gadgets in the system). The only request was to remove flashing as a requirement.

ricab:
Yeah, I am not sure how much it will help, but my point is that we should reassess once that lands. It's easy to forget, so I suggest leaving a capsule to the future somewhere. Probably some automated Jira issue would be a better way to track than a comment in the code.

include/multipass/utils/grpc_utils.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace multipass
+{
+namespace utils
+{
+
+template <typename Reply, typename Request>
+    requires LogMsgReply<Reply>
+void send_messages(grpc::ServerReaderWriterInterface<Reply, Request>* server,
+                   MessageBag&& message_bag)
```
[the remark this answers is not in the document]
tobe2098: This is already required by the concept

include/multipass/annotated_value.h
```diff
@@ -0,0 +1,275 @@
...
+//  auto retval2 = f().collect(message_bag);
+// And in the end return {val,message_bag};
+// Later on the message bag can be iterated for messages.
+
+template <typename T>
+// The payload should never be a reference, and no move constructor will fail compilation
+concept CanBeAnnotated =
+    std::is_void_v<T> || (std::move_constructible<T> && !std::is_reference_v<T>);
+
+class MessageBag
```
jimporter: Naming nitpick: A "bag" is another name for a [multiset](https://en.wikipedia.org/wiki/Set_(abstract_data_type)#Multiset) (an associative container that can hold multiple elements that compare equal), but this is a data type that assumes the messages maintain insertion order, like a vector/list. How about calling this `MessageList` or `UserMessages` or something like that?

include/multipass/annotated_value.h
```diff
@@ -0,0 +1,275 @@
...
+
+    // To allow usage of merge()
+    template <typename T>
+        requires CanBeAnnotated<T>
+    friend class Annotated;
+};
+
+template <typename T>
+    requires CanBeAnnotated<T>
+class [[nodiscard("You must collect() the value or bind it to retrieve the MessageBag")]] Annotated
```
jimporter: I thought this over for a while, and I think this makes the API more complex. On the positive side, the `Annotated` type is reminiscent of `std::expected` in C++ or `std::result` in Rust, though those hold a value XOR error. In that case, both the value and the error can be said to be possible *results* of the function call (hence the Rust name), so they both fit as return values. In a more concrete sense, functions generally want to return errors to the user immediately when they're encountered.

On the other hand, messages here come along *with* values and we might produce messages in the middle of a function, proceeding on after that with more computation. Thus, we can get multiple messages. That produces some of the extra complexity seen in using the `Annotated` type: 1) we have to `collect` messages into a single `MessageBag; 2) to reduce copy-ctor calls, we have to sprinkle `std::move` in key places of the code; 3) in several cases below, we create an empty `MessageBag` first anyway, and then `collect` into it, which is pretty similar to just passing it as an argument.

Having `Annotated` return values also closes off a few other future possibilities for this code. First, it makes it harder to switch to `std::expected` if we wanted to start using that instead of exceptions (this could possibly be helpful in our Rustification efforts). Second, instead of using a `MessageBag`, a future version of this could pass in a hypothetical `MessageChannel` that immediately sends the message to the client via the stream.

One other area where the `Annotated` class is nice is that it makes it clear when we're ignoring the result (at least for `Annotated<void>`):

```c++
(void)doSomething();  // Nice, obvious
doSomething(/* messageBag = */ {});  // Less nice, non-obvious
```

If we go with passing in a `MessageBag` as an argument, one option to make this more explicit would be to add a sentinel value, something like so:

```c++
class MessageBag {
public:
  static struct ignore_tag {} ignore;

  MessageBag(ignore_tag) {}
};

// ...

doSomething(MessageBag::ignore);
```

src/client/cli/cmd/common_callbacks.h
```diff
@@ -21,13 +21,15 @@
...
 #include <multipass/constants.h>
+#include <multipass/reply_concepts.h>
 #include <multipass/terminal.h>
 
 #include <grpc++/grpc++.h>
 
 namespace multipass
 {
 template <typename Request, typename Reply>
+    requires LogReply<Reply>
```
jimporter: This could be simplified to the following, right?

```suggestion
template <typename Request, LogReply Reply>
```

(Ditto elsewhere in this file.)

jimporter (changes requested):
Thanks for all the work on this! I think this is overall a big improvement, both for the immediate problem we're solving as well as the new infrastructure to support passing messages along to the user.

I spent some time thinking over the new APIs and doing some little experiments with them to see how they'd work in other cases (e.g. `Annotated<std::string>`). It's hard to predict the future of what exactly we'll need, but I think I'd prefer holding off on introducing `Annotated`, at least until we have some more complex cases where we can be sure it's the better strategy. Some more-specific thoughts below.

Of course, I'm sympathetic to the fact that you already wrote this code and that it would feel a shame for it to go to waste, but we can always revisit it when we have more information. I mainly just want to avoid introducing more complexity too early, since it could easily become tech debt in the future that makes things harder on us.

include/multipass/annotated_value.h
tobe2098: Yes, `UserMessages` is definitely better. `MessageBag` was not very inspired.

include/multipass/annotated_value.h
tobe2098: You make very good arguments. I do not see why would it be harder to switch to `std::expected`, since we can just do `Annotated<std::expected>` and the result/exception is separate from just the messages. 

It is true that there is some awkwardness with `std::move` to avoid the string copies, and the message collection forced on one `MessageBag` with `.collect()` is quite similar to simply passing `MessageBag`, but something like the sentinel value is not forcing, which was the whole point of the design.

The idea is to constrain and limit the interface such that the chances of misuse are minimal, since misuse of messages would not be easily detectable. The solution has to be as forcing as possible to make sure that such messages are not lost. With `MessageBag`, you need to remember to pass it down (not `{}`) your calls to not "break the chain". With `Annotated`, at the callsite you are forced to deal with the `MessageBag` and remember to return it (it would be ideal if you were forced to return it).

Passing the `MessageChannel` would be less ideal. It would break encapsulation, since multiple parts of code would need to "know" about `MessageChannel`, its interface, to pass it down, and it would have to have a valid `null` state for callsites that do not receive a valid `MessageChannel` from the arguments. In that sense, the other options are better.

We can meet and discuss further, since I think it will be hard to reach consensus over text. It is also possible that there is a third possibility that would work better.

src/client/cli/cmd/common_callbacks.h
tobe2098: Yes, I thought about it because the requires is more explicit. But in retrospective it is a compile time check, so I do not think it is a better option, your suggestion is the way to go.

src/daemon/daemon.cpp
```diff
@@ -2533,7 +2536,8 @@ catch (const mp::NonAuthorizedBridgeSettingsException& e)
     {
         user_authorized_bridges.insert(get_bridged_interface_name());
 
-        MP_SETTINGS.set(QString::fromStdString(key), QString::fromStdString(val));
+        [[maybe_unused]] UserMessages messages{};
+        MP_SETTINGS.set(QString::fromStdString(key), QString::fromStdString(val), messages);
```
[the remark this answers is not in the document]
tobe2098: This is due to the fact that only the resize request populates the messages

include/multipass/utils/grpc_utils.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <multipass/rpc/multipass.grpc.pb.h>
+
+namespace multipass
+{
+namespace utils
+{
+template <LogMsgReply Reply, typename Request>
+void send_messages(grpc::ServerReaderWriterInterface<Reply, Request>* server,
+                   const UserMessages& message_bag)
```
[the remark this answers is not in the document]
tobe2098: This is not possible in the codebase since these types are implemented via the grpc templates, which will always contain both the getter and the setter

src/daemon/daemon.cpp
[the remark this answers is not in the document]
tobe2098: Already answered

include/multipass/utils/grpc_utils.h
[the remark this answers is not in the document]
tobe2098: Also already replied

src/client/cli/cmd/set.cpp
```diff
@@ -39,7 +40,11 @@ mp::ReturnCodeVariant cmd::Set::run(mp::ArgParser* parser)
         try
         {
             if (ret == ReturnCode::Ok)
-                MP_SETTINGS.set(key, val);
+            {
+                // We are in the client, messages cannot be sent
+                [[maybe_unused]] mp::UserMessages messages{};
+                MP_SETTINGS.set(key, val, messages);
```
jimporter, line 46: I think this works out fairly well as a way of making it explicit that we're ignoring messages. There might be a way to turn this into a one-liner, like by passing some hypothetical `UserMessages::ignore`, but I don't think that's necessary now. Maybe in the future if we find that we're repeating this pattern a lot.

jimporter (approved):
Looks good to me, thanks for all the work on this! One small comment below about some potential followup we could do in the future.

src/client/cli/cmd/set.cpp
tobe2098: Due to lvalue references this is not possible I believe, but it can be definitely revisited. Thanks!

jimporter: `UserMessages::ignore` could be a static member variable with a special initialization to mark it as "ignored". Still, no need to worry about this right now...

PR #4440, opened 2025-10-15
Refactor: Strip full file paths from SSH process logs

prachimaskar:
This PR replaces the previous PR #4196, which was closed automatically after the C++20 branch was deleted.

Changes:
Refactored SSH process logging to strip full file paths and display only the filename using std::source_location and consteval.

Why this PR is needed:
Improves log readability.
Enhances security by avoiding exposure of full file paths.

---

MULTI-2458

PR #4674, opened 2026-02-10
Fix file-path leaks in logs

ricab:
Re-issue of #4440.


jimporter (approved):

ricab:
Hey @jimporter, do you think this needs another review?

jimporter:
@ricab I'm ok with merging this with just my review (following along with our WIP updated review guidelines). The change is small, has automated tests, and only applies to a small number of logging calls.

There's some benefit to having more people on the team look at changes so that we're all familiar with what's in the codebase, but that shouldn't block merges. It's easy enough to briefly look over commits after they merge to `main` so that we can all keep up with what's new.

PR #4675, opened 2026-02-11
[az] Check whether subnets are in-use before allocating them for AZs

jimporter:
# Description

On the `main` branch, we have logic to make sure that when we pick out a subnet, it's not already in use. On the `availability-zones` branch however, we lost this functionality (this happened when merging in the new `Subnet` class). This PR adds that check back in. To do this, I added a new `SubnetAllocator` class that handles the bookkeeping of selecting the next available subnet, and then I was able to easily add in a check to the (already-existing) `subnet_used_locally` function.

## Related Issues

Hopefully, this will help resolve https://github.com/canonical/multipass/issues/4383 once the AZ branch merges into `main`.

## Testing

- Added new unit tests for the `SubnetAllocator` and applied mocks to the AZ unit tests
- Manual testing steps:

  1. Launch multipass daemon/GUI
  2. Launch a new VM
  3. Ensure you can connect to the shell

## Checklist

## Additional Notes

`subnet_used_locally` still isn't implemented on Windows, but we can do that in a followup.

### DIFF
--- a/src/network/subnet.cpp
+++ b/src/network/subnet.cpp
@@ -144,29 +144,6 @@ size_t mp::Subnet::size(mp::Subnet::PrefixLength prefix_length) const
     return std::size_t{1} << (prefix_length - this->prefix_length());
 }
 
-mp::Subnet mp::Subnet::get_specific_subnet(size_t subnet_block_idx,
-                                           mp::Subnet::PrefixLength prefix_length) const
-{
-    const size_t possible_subnets = size(prefix_length);
-    if (possible_subnets == 0)
-        throw std::logic_error(
-            fmt::format("A subnet with prefix length {} cannot be contained by {}",
-                        prefix_length,
-                        *this));
-
-    if (subnet_block_idx >= possible_subnets)
-        throw std::invalid_argument(
-            fmt::format("{} is greater than the largest subnet block index {}",
-                        subnet_block_idx,
-                        possible_subnets - 1));
-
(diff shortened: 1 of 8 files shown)


src/network/subnet.cpp
```diff
@@ -187,3 +165,36 @@ std::strong_ordering mp::Subnet::operator<=>(const Subnet& other) const
...
+            fmt::format("A subnet with prefix length {} cannot be contained by {}",
+                        prefix,
+                        base_subnet));
+}
+
+mp::Subnet mp::SubnetAllocator::next_available()
+{
+    const size_t possible_subnets = base_subnet.size(prefix);
+
+    while (true)
```
sharder996: Practically I don't see it happening, but this could theoretically result in an infinite loop. Should we cap this and throw an exception instead?

sharder996 (changes requested):
Looking good from a code review POV. Just a small question.

src/network/subnet.cpp
jimporter: It's technically capped because the `possible_subnets` count is always finite, though maybe there's a better way to annotate that in the code so we're not relying on second-order effects...

jimporter: Ok, I reworked this to pull the conditional from the inner `if` into the `while`, which hopefully makes it easier to follow.

sharder996: Looks better now, thanks!

sharder996 (approved):
LGTM

PR #4676, opened 2026-02-11
[applevz] ASIF image support

sharder996:
Add support for ASIF images on macOS 26 and later. Conversion from QCOW2 and existing VMs using RAW images are also handled.

As the `qemu-img` binary is required by more than just the `qemu` backend, it has been factored out of the backend code into the `utils` directory. It can now be used by the `applevz` and `hyperv` backends without having the `qemu` backend be compiled.

---

MULTI-2270
MULTI-2273


src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+Path AppleVZUtils::convert_to_supported_format(const Path& image_path) const
+{
+    if (macos_at_least(26, 0))
+    {
+        return convert_to_asif(image_path);
+    }
+    else
+    {
+        return backend::convert(image_path, "raw");
+    }
```
ricab: You could compact it a little:

```suggestion
    return macos_at_least(26, 0) ? convert_to_asif(image_path)
                                 : backend::convert(image_path, "raw");
```

src/platform/backends/applevz/applevz_virtual_machine.cpp
```diff
@@ -40,6 +45,7 @@ AppleVZVirtualMachine::AppleVZVirtualMachine(const VirtualMachineDescription& de
                                              const Path& instance_dir)
     : BaseVirtualMachine{desc.vm_name, key_provider, instance_dir}, desc{desc}, monitor{&monitor}
 {
+    amend_disk_image(this->desc);
```
ricab: Is this meant to deal with existing VMs when a user updates to macOS >= 26? If so, it might merit a comment.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+        return false;
+
+    char magic[4];
+    ssize_t bytes_read = read(fd, magic, 4);
+    close(fd);
+
+    if (bytes_read != 4)
+        return false;
+
+    return !std::memcmp(magic, "shdw", 4);
```
ricab: Why go back to C for this? C++ has higher-level constructs to deal with files.

(here and throughout the file)

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+              disk_space.human_readable());
+
+    auto process = MP_PROCFACTORY.create_process(
+        QStringLiteral("diskutil"),
+        QStringList() << "image" << "create" << "blank" << "--fs" << "none" << "--format" << "ASIF"
+                      << "--size" << QString::number(disk_space.in_bytes()) << image_path);
+
+    auto exit_state = process->execute();
+
+    if (!exit_state.completed_successfully())
```
ricab: How about using an if init statement? And you could make the variable `const`.


```suggestion
    if (const auto exit_state = process->execute(); !exit_state.completed_successfully())
```

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+namespace mp = multipass;
+namespace mpl = multipass::logging;
+
+namespace
+{
+constexpr auto category = "applevz-utils";
+
+void create_asif(const mp::Path& image_path, const mp::MemorySize& disk_space)
+{
+    mpl::info(category,
```
ricab: I would say the success log is the most important, no? Idk, it seems to me that write-ahead logging isn't super useful here. Failures would be reported as an exception. With a log at the end, that is enough to always have a log when this function executes, unless there is a crash. WDYT?

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+    ssize_t bytes_read = read(fd, magic, 4);
+    close(fd);
+
+    if (bytes_read != 4)
+        return false;
+
+    return !std::memcmp(magic, "shdw", 4);
+}
+
+void make_sparse(const mp::Path& path, const mp::MemorySize& disk_space)
```
ricab: Right now, it is not obvious what image type this applies to. You could fix that by renaming the `path` parameter to indicate what it is a path to.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+    }
+}
+
+void AppleVZUtils::resize_image(const MemorySize& disk_space, const Path& image_path) const
+{
+    mpl::trace(category, "Resizing image to: {}", disk_space.human_readable());
+
+    if (is_asif_image(image_path))
+    {
+        auto process = MP_PROCFACTORY.create_process(
```
ricab: Though this function is not very big, this part looks like an independent unit that could well be extracted (analogous to the else branch).

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+#include <vector>
+
+namespace mp = multipass;
+namespace mpl = multipass::logging;
+
+namespace
+{
+constexpr auto category = "applevz-utils";
+
+void create_asif(const mp::Path& image_path, const mp::MemorySize& disk_space)
```
ricab: The boilerplate to create and handle processes is a little repetitive in this file. How about extracting it into a generic function that the others call? 

It could receive the command and a string you used to compose messages. For example, "create ASIF image...", to which you could prepend "Trying to " for the initial log, "Failed to" for the exception, and "Managed to" for the final log. That's just an example, exact wording to be found.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+    if (ftruncate(fd, size_bytes) == -1)
+    {
+        close(fd);
+        throw std::runtime_error(fmt::format("Failed to resize file: {}", strerror(errno)));
+    }
+
+    close(fd);
+}
+
+mp::Path convert_to_asif(const mp::Path& source_path)
```
ricab: This function would benefit from some splitting. Leading comments mark good candidates.

Also, it looks a little AI-like. Did you inspect it closely? Are all paths covered?

Finally, as in others, using C++ should help reduce boilerplate.

src/utils/qemu_img_utils.cpp
```diff
@@ -113,19 +86,27 @@ void mp::backend::amend_to_qcow2_v3(const mp::Path& image_path)
                           "Failed to amend image to QCOW2 v3");
 }
 
-mp::Path mp::backend::convert_to_raw(const mp::Path& image_path)
+mp::Path mp::backend::convert(const mp::Path& source_path, const QString& target_format)
```
ricab: Nice 👍

include/multipass/json_utils.h
```diff
@@ -170,7 +170,9 @@ inline void tag_invoke(const boost::json::value_from_tag&,
 
 inline QString tag_invoke(const boost::json::value_to_tag<QString>&, const boost::json::value& json)
 {
-    return QString::fromStdString(value_to<std::string>(json));
+    if (json.is_string())
+        return QString::fromStdString(value_to<std::string>(json));
+    return QString::fromStdString(boost::json::serialize(json));
```
ricab, line 171: Why was this necessary? Is it really appropriate? 

Tag-🤪-invoking @jimporter for his wisdom.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -84,37 +85,7 @@ void delete_image_dir(const mp::Path& image_path)
...
-    {
-        image_size = mp::MemorySize(match.captured("size").toStdString());
-    }
-    else
-    {
-        throw std::runtime_error{"Could not obtain image's virtual size"};
-    }
-
-    return image_size;
+    return mp::MemorySize(mp::backend::get_image_info(image_path, "virtual-size").toStdString());
```
ricab, line 89: 😎

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,328 @@
...
+                    close(target_fd);
+                    throw std::runtime_error(fmt::format("Short write: read {} bytes but wrote {}",
+                                                         bytes_read,
+                                                         bytes_written));
+                }
+            }
+            else
+            {
+                // Skip zero blocks by seeking forward
+                if (lseek(target_fd, bytes_read, SEEK_CUR) == -1)
```
ricab: suggesting `else if`

ricab (changes requested):
Thanks for the atomic commits, they made reviewing a breeze.

I appreciate how you generalized existing utilities for reuse 👍 The new applevz_utils could use a little more love and I think we should have more thorough tests where files are being edited at the byte level.

src/platform/backends/applevz/applevz_virtual_machine.cpp
sharder996: Correct! This will convert previously existing VMs that were backed by a RAW disk image to an ASIF disk image if on macOS 26

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

src/platform/backends/applevz/applevz_utils.mm
sharder996: done

include/multipass/json_utils.h
sharder996: I found this to be necessary in order to parse strings and non-strings (i.e. integers) out of json values into QStrings.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,207 @@
...
+            std::vector<char> buffer(buffer_size);
+            while (source.read(buffer.data(), buffer_size) || source.gcount())
+            {
+                const auto bytes_read = source.gcount();
+                const bool is_zero = std::all_of(buffer.cbegin(),
+                                                 buffer.cbegin() + bytes_read,
+                                                 [](char c) { return c == '\0'; });
+
+                is_zero ? target.seekp(bytes_read, std::ios::cur)
+                        : target.write(buffer.data(), bytes_read);
```
ricab: So this just skips reads that happen to contain only zeros, right?

Consider a file with `0 0 0 0 0 0 0 0 0 0 0 42 0 0 0 0 0 0 0 0 0 0 0 42 0 0 0 0 0 0 0 0`. If the reads happened to be properly aligned, this would write `42 42`, but if they were chunked as `0 0 0 0 0 0 0 0 0 0 0 42 0 0 0` and `0 0 0 0 0 0 0 0 42 0 0 0 0 0 0 0 0` no zero's would be skipped, correct? I suppose that is not the intention.

Furthermore, could you please add some supporting documentation in the PR as to why all zeros can be skipped in the ASIF format? Can't a zero ever be valid data?

Finally, doesn't Apple provide a tool to perform this conversion instead?

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,207 @@
...
+    create_asif(asif_path, mp::MemorySize{std::to_string(source_size)});
+
+    mp::Path device_path = attach_asif(asif_path);
+    try
+    {
+        {
+            std::ifstream source(raw_path.toStdString(), std::ios::binary);
+            if (!source)
+                throw std::runtime_error(
+                    fmt::format("Failed to open source image: {}", raw_path.toStdString()));
```
ricab: This pattern is repeated multiple times in this function:

```
try-file-system-operation()
if (not success)
    throw something
```

You could extract it into separate small "checked_..." functions, like `checked_file_size` or `checked_open`.

Moreover, you often don't need to throw yourself, you can use [`basic_ios::exceptions`](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions.html) which covers the `badbit` too. With that, you should be able to create a single template to open both output and input streams.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,207 @@
...
+
+            std::ofstream target(device_path.toStdString(), std::ios::binary);
```
ricab: I guess this should use FileOps instead (and add anything needed there), right? The fact that these things aren't mocked here tells me failure cases are not being tested.

ricab (changes requested):
Better, thanks. `convert_to_asif` still has a few issues (and I still get the feeling of AI let loose...).

src/platform/backends/applevz/applevz_utils.mm
sharder996: Yup, that's correct. I could use a smaller buffer and eliminate some more zero'ed data from being copied, but it would require more `read` syscalls. Also, yes, zeros can be valid data, but because ASIF is a sparse image format, there is no difference between copying a zero or seeking past it.

As for documentation, there is no official documentation. All of the word here is the result of other implementations I've found and general reverse engineers of the file type. Nevertheless, I'll link some of information I've found online.

Yes, you can use `diskutil` to convert a RAW image to an ASIF image. I've found no way to go directly from qcow2 to ASIF. If you use `diskutil` to convert the image, it creates a one-to-one copy meaning you lose out on saving disk space with sparse images.

ricab: > I could use a smaller buffer and eliminate some more zero'ed data from being copied, but it would require more read syscalls.

No, I don't think that would be advisable. It would always have the same problem unless you read byte-by-byte. Instead, to remove zero's you would need to filter the buffer before writing (e.g. `std::copy_if`).

> Also, yes, zeros can be valid data, but because ASIF is a sparse image format, there is no difference between copying a zero or seeking past it.

Sparse formats need some compensating info to know where the zeros are. They drop the zeros but keep track of the ranges containing them. Otherwise, how would you recover the original information?

> If you use diskutil to convert the image, it creates a one-to-one copy meaning you lose out on saving disk space with sparse images.

Until we are very sure that we aren't loosing data, that sounds like a lesser evil to me.

sharder996:
Rebased on top of a bunch of QString to std::string conversions. Let me know if I missed anything.

include/multipass/json_utils.h
jimporter: I think `value_to<QString>` and `value_to<std::string>` should work exactly the same, aside from the actual return type. Since `value_to<std::string>(json_holding_an_int)` fails (right?), the same should apply for `value_to<QString>`.

The main reason I have for this is that eventually I hope we can remove the `QString` overload here entirely, replacing it with `std::string` usage. It'll be easier to do that if there's no special behavior in this implementation.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -15,6 +15,13 @@
  *
  */
 
+/*
+ * This file provides some support for creating and resizing ASIF disk images.
+ * There is no official public documentation for the ASIF format. The technical
+ * details used here are based on community reverse engineering. For reference,
+ * see: https://github.com/huven/asif-format
```
ricab, line 22: Ahh, thanks for the reference! 🙂 

> As usual, there are no technical details provided about the layout of the ASIF file.

😮‍💨 Another reason to label the feature as experimental for a while.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,165 @@
...
+void make_sparse(const std::filesystem::path& raw_image_path, const mp::MemorySize& disk_space)
+{
+    std::error_code ec;
+    std::filesystem::resize_file(raw_image_path, disk_space.in_bytes(), ec);
+
+    if (ec)
+        throw std::runtime_error(fmt::format("Failed to resize file: {}", ec.message()));
+}
+
+std::filesystem::path convert_to_asif(const std::filesystem::path& source_path, bool destructive)
```
ricab, line 110: I am trying to understand this new `destructive` parameter. My impression is that it should not be a parameter(?)

IIUC, all you want is to delete intermediate images that are created in *this* conversion call (not an earlier one), right? If so, this function already has that information: essentially `destructive <=> raw_path != source_path`. 

I am reading the comment on line 132 and thinking this unit should not have to care about where images belong. On the other hand, the destructive parameter opens up undesired cases that callers shouldn't need to worry about:

- raw input image + destructive → destroys the original image
- non-raw input image + non-destructive → leaves intermediate image

src/platform/backends/applevz/applevz_virtual_machine.cpp
```diff
@@ -15,21 +15,28 @@
...
+namespace mp = multipass;
+namespace mpl = mp::logging;
 
 namespace
 {
 constexpr static auto log_category = "applevz-vm";
+
+void amend_disk_image(mp::VirtualMachineDescription& desc)
+{
+    // Keep the source image intact in case anything goes wrong
```
ricab: This will take a toll on people's disk space. Since the logic always converts to ASIF on the newer macOS, if something did go wrong, people would still be unable to use the previous image to get into their instance, correct? Multipass would always try to convert again, it would not just start the VM with the existing image, right? They would need to mount the image to get data out. And because none of those images are sparse, any 20G VM would silently take 40G. The backup would remain around — without a supported way to access it — until people deleted the instance for good.

I guess the alternative would be to leave existing VMs alone and use ASIF only for new launches. If we include "clone" in those new launches, then people have:

- a way to trigger the conversion
- a way to access the old instance
- a way to delete the old instance

The main downsides I see is that conversion would not occur automatically, so people could miss the possibility. But that could be mitigated by providing a message prompting users to switch when they started the instance or at other important transitions. That would be an optional improvement; it could come later on and leverage Antoni's work for user messages.

Don't just jump into it though, please let me know your thoughts. This is just quick speculation, there may be things I am missing. If we did go with manual conversion, I would leave that to a separate PR.

src/platform/backends/shared/base_virtual_machine_factory.cpp
```diff
@@ -136,14 +137,17 @@ void mp::BaseVirtualMachineFactory::copy_instance_dir_with_essential_files(
 {
     assert(fs::exists(source_instance_dir_path) && fs::is_directory(source_instance_dir_path));
 
+    const bool has_asif =
+        std::any_of(fs::directory_iterator(source_instance_dir_path),
+                    fs::directory_iterator{},
+                    [](const auto& e) { return e.path().extension() == ".asif"; });
+
```
ricab: This is only here for converted instances with a raw "backup", correct?

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,165 @@
...
+        if (destructive)
+            QFile::remove(raw_path);
+        QFile::remove(asif_path);
+        throw;
+    }
+
+    // This is often an intermediate file so we remove it, unless it came from an existing VM
+    if (destructive)
+        QFile::remove(raw_path);
+    return asif_path;
```
ricab: You could streamline this stuff a bit by using scope guards, if you wanted. I'll leave that up to you.

ricab (commented):
Hey Scott, thanks for addressing my earlier reviews. The code looks neat and you're covering things much better with unit tests. 

I do have some concerns with new code you added, around conversion and backup images. Most importantly, I think the decision to covert existing VMs deserves some more consideration.

src/platform/backends/applevz/applevz_virtual_machine.cpp
sharder996: I kinda like your idea of leaving instances that were using a RAW disk image. Clone would convert to ASIF. Then, in the future, if we needed an ASIF image for a feature (as an example, snapshots), we could consider other approaches; prompting the user, etc.

WDYT

ricab: Yep, makes sense to me. Separate PR?

sharder996: Yeah, any fallback behavior should go in a separate PR IMO

sharder996:
@ricab WDYT of this implementation for what we discussed? Existing VMs that were using RAW disk images (macOS 15 and older) continue to use the same image even when the host upgrades to macOS 26. New VMs then start to use ASIF disk images. I will address cloning VMs later. Right now, the cloned VM will continue to use the RAW image.

The implication, though, is that the image vault is full of cached RAW disk images which it will try to convert to ASIF every time a VM is launched. In order to update the images in the vault, I introduced an 'image amender' that each backend/platform can potentially be passed to the image vault to update cached images. An existing use case of this functionality is the `amend_to_qcow2_v3`. Again, if a qcow2_v2 image had been cached in the vault, it would have been amended to v3 every time a VM would be launched. Instead we amend all the cached images at once.

src/platform/backends/applevz/applevz_utils.mm
```diff
@@ -0,0 +1,162 @@
...
+    mpl::info(category, "Converting {} to ASIF format", source_path);
+
+    // NO-OP if already RAW
+    auto raw_path = mp::backend::convert(source_path, "raw");
+
+    // This is often an intermediate file so we remove it, unless it came from an existing VM
+    auto raw_file_cleanup =
+        sg::make_scope_guard([&raw_path]() noexcept { QFile::remove(raw_path); });
+    if (raw_path == source_path)
+        raw_file_cleanup.dismiss();
```
ricab: I was a little confused by the comment now. There is no notion of VM here.

Also, creating a scope_guard only to dismiss it is a little awkward. How about something along these lines (untested):


```suggestion
    auto intermediate_cleanup = 
        sg::make_scope_guard([intermediate = (raw_path != source_path)]() noexcept { if (intermediate) QFile::remove(raw_path); });
```

ricab (commented):
I am not sure about the new image amender item. Those images are updated periodically so any old formats should be gone after a while anyway, no? Is it worth the extra commitment and complexity, which is not needed in most cases? I am not totally opposed either, I could be convinced, especially since you went to that trouble already. But can we avoid avoid augmenting this PR with that at this point. It is yet another piece of functionality that was not previously foreseen, right? Since you have the code, how about a separate PR with your arguments in the description? You could keep me as reviewer or assign someone else.

Apart from that and a small proposal inline, I am fine with the code. I have yet to test, but I will attempt that once the code settles, then we can get this merged.

sharder996:
@ricab Have separated out the commits for amending/updating images in the vault.

ricab (approved):
The code looks good, but I am afraid I can't test: just found my Mac is just below the cutoff for macOS 26...

@tobe2098, is yours eligible perhaps? See https://support.apple.com/en-us/122867

sharder996:
BTW, if you want to simulate the transition from macOS 15 to 26, you can create a VM using `main` and `applevz`, which uses RAW images, and then install the package from this PR, which uses ASIF images.

ricab:
Do you mean simulating newer hardware for the VM?

sharder996:
Simulate the transition from `applevz` using RAW images to using ASIF images.

sharder996:
@ricab Here's some logs from a VM launch done by @christophniederberger-canonical 

```
christoph@Christophs-Mac-mini ~ % multipass version
multipass   1.17.0-dev.pr4676+gc23f1405.mac
multipassd  1.17.0-dev.pr4676+gc23f1405.mac
christoph@Christophs-Mac-mini ~ % multipass set local.driver=applevz
christoph@Christophs-Mac-mini ~ % multipass launch -n foo -vvvv
[2026-04-22T16:58:38.525] [trace] [url downloader] Found https://cloud-images.ubuntu.com/releases/server/releases/noble/release-20260321/ubuntu-24.04-server-cloudimg-arm64.img in cache: false
[2026-04-22T16:58:38.526] [debug] [image vault] Verifying hash "1ea801e659d2f5035ac294e0faab0aac9b6ba66753df933ba5c7beab0c689bd0"
[2026-04-22T16:58:38.927] [info] [applevz-utils] Converting /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.img to ASIF format
[2026-04-22T16:58:38.933] [debug] [qemu-img] [33358] started: qemu-img info --output=json /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.img
[2026-04-22T16:58:38.954] [debug] [qemu-img] [33359] started: qemu-img convert -p -O raw /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.img /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.raw
[2026-04-22T16:58:41.578] [info] [applevz-utils] Trying to convert /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.raw to ASIF format at /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.asif
[2026-04-22T16:58:41.593] [debug] [diskutil] [33365] started: diskutil image create from /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.raw /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.asif -f asif
[1% completed] 58:41.876] [warning] [diskutil] 
[2% completed] 58:41.877] [warning] [diskutil] 
[3% completed] 58:41.877] [warning] [diskutil] 
[4% completed] 58:41.880] [warning] [diskutil] 
[5% completed] 58:41.889] [warning] [diskutil] 
[6% completed] 58:41.895] [warning] [diskutil] 
[7% completed] 58:41.904] [warning] [diskutil] 
[8% completed] 58:41.913] [warning] [diskutil] 
[9% completed] 58:41.925] [warning] [diskutil] 
[10% completed] 8:41.932] [warning] [diskutil] 
[11% completed] 8:41.940] [warning] [diskutil] 
[12% completed] 8:41.949] [warning] [diskutil] 
[13% completed] 8:41.961] [warning] [diskutil] 
[14% completed] 8:41.971] [warning] [diskutil] 
[15% completed] 8:41.978] [warning] [diskutil] 
[16% completed] 8:42.024] [warning] [diskutil] 
[17% completed] 8:42.053] [warning] [diskutil] 
[18% completed] 8:42.078] [warning] [diskutil] 
[19% completed] 8:42.085] [warning] [diskutil] 
[20% completed] 8:42.098] [warning] [diskutil] 
[21% completed] 8:42.124] [warning] [diskutil] 
[22% completed] 8:42.147] [warning] [diskutil] 
[23% completed] 8:42.173] [warning] [diskutil] 
[24% completed] 8:42.214] [warning] [diskutil] 
[25% completed] 8:42.247] [warning] [diskutil] 
[26% completed] 8:42.261] [warning] [diskutil] 
[27% completed] 8:42.270] [warning] [diskutil] 
[28% completed] 8:42.279] [warning] [diskutil] 
[29% completed] 8:42.294] [warning] [diskutil] 
[30% completed] 8:42.343] [warning] [diskutil] 
[31% completed] 8:42.377] [warning] [diskutil] 
[32% completed] 8:42.418] [warning] [diskutil] 
[33% completed] 8:42.460] [warning] [diskutil] 
[34% completed] 8:42.500] [warning] [diskutil] 
[35% completed] 8:42.531] [warning] [diskutil] 
[36% completed] 8:42.567] [warning] [diskutil] 
[37% completed] 8:42.607] [warning] [diskutil] 
[38% completed] 8:42.646] [warning] [diskutil] 
[39% completed] 8:42.654] [warning] [diskutil] 
[40% completed] 8:42.741] [warning] [diskutil] 
[41% completed] 8:42.763] [warning] [diskutil] 
[42% completed] 8:42.803] [warning] [diskutil] 
[43% completed] 8:42.882] [warning] [diskutil] 
[44% completed] 8:42.940] [warning] [diskutil] 
[45% completed] 8:42.966] [warning] [diskutil] 
[46% completed] 8:43.004] [warning] [diskutil] 
[47% completed] 8:43.075] [warning] [diskutil] 
[48% completed] 8:43.137] [warning] [diskutil] 
[49% completed] 8:43.153] [warning] [diskutil] 
[50% completed] 8:43.202] [warning] [diskutil] 
[51% completed] 8:43.227] [warning] [diskutil] 
[52% completed] 8:43.280] [warning] [diskutil] 
[53% completed] 8:43.310] [warning] [diskutil] 
[54% completed] 8:43.343] [warning] [diskutil] 
[55% completed] 8:43.374] [warning] [diskutil] 
[56% completed] 8:43.407] [warning] [diskutil] 
[57% completed] 8:43.453] [warning] [diskutil] 
[58% completed] 8:43.487] [warning] [diskutil] 
[59% completed] 8:43.556] [warning] [diskutil] 
[60% completed] 8:43.650] [warning] [diskutil] 
[61% completed] 8:43.718] [warning] [diskutil] 
[62% completed] 8:43.780] [warning] [diskutil] 
[63% completed] 8:43.858] [warning] [diskutil] 
[64% completed] 8:43.933] [warning] [diskutil] 
[65% completed] 8:43.989] [warning] [diskutil] 
[66% completed] 8:44.099] [warning] [diskutil] 
[68% completed] 8:44.377] [warning] [diskutil] 
[69% completed] 8:44.456] [warning] [diskutil] 
[70% completed] 8:44.556] [warning] [diskutil] 
[71% completed] 8:44.747] [warning] [diskutil] 
[72% completed] 8:44.820] [warning] [diskutil] 
[73% completed] 8:44.992] [warning] [diskutil] 
[74% completed] 8:45.061] [warning] [diskutil] 
[75% completed] 8:45.199] [warning] [diskutil] 
[76% completed] 8:45.321] [warning] [diskutil] 
[77% completed] 8:45.569] [warning] [diskutil] 
[78% completed] 8:45.729] [warning] [diskutil] 
[79% completed] 8:45.872] [warning] [diskutil] 
[80% completed] 8:46.058] [warning] [diskutil] 
[81% completed] 8:46.077] [warning] [diskutil] 
[82% completed] 8:46.303] [warning] [diskutil] 
[83% completed] 8:46.562] [warning] [diskutil] 
[84% completed] 8:46.566] [warning] [diskutil] 
[85% completed] 8:46.644] [warning] [diskutil] 
[86% completed] 8:46.728] [warning] [diskutil] 
[87% completed] 8:46.787] [warning] [diskutil] 
[88% completed] 8:46.872] [warning] [diskutil] 
[89% completed] 8:46.925] [warning] [diskutil] 
[90% completed] 8:46.982] [warning] [diskutil] 
[91% completed] 8:47.055] [warning] [diskutil] 
[92% completed] 8:47.080] [warning] [diskutil] 
[93% completed] 8:47.129] [warning] [diskutil] 
[94% completed] 8:47.177] [warning] [diskutil] 
[95% completed] 8:47.278] [warning] [diskutil] 
[96% completed] 8:47.358] [warning] [diskutil] 
[97% completed] 8:47.625] [warning] [diskutil] 
[98% completed] 8:47.714] [warning] [diskutil] 
[99% completed] 8:49.379] [warning] [diskutil] 
[100% completed] :49.390] [warning] [diskutil] 
[2026-04-22T16:59:03.640] [debug] [FileOps] Saved file `/var/root/Library/Application Support/multipassd/applevz/vault/multipassd-instance-image-records.json` successfully in attempt #1
[2026-04-22T16:59:03.654] [debug] [FileOps] Saved file `/var/root/Library/Caches/multipassd/applevz/vault/multipassd-image-records.json` successfully in attempt #1
[2026-04-22T16:59:03.664] [debug] [qemu-img] [33424] started: qemu-img info --output=json /var/root/Library/Caches/multipassd/applevz/vault/images/noble-20260321/ubuntu-24.04-server-cloudimg-arm64.asif
[2026-04-22T16:59:03.717] [trace] [applevz-utils] Resizing image to: 5.0GiB
[2026-04-22T16:59:03.717] [info] [applevz-utils] Trying to resize ASIF image: /var/root/Library/Application Support/multipassd/applevz/vault/instances/foo/ubuntu-24.04-server-cloudimg-arm64.asif, size: 5.0GiB
[2026-04-22T16:59:03.722] [debug] [diskutil] [33430] started: diskutil image resize --size 5368709120 /var/root/Library/Application Support/multipassd/applevz/vault/instances/foo/ubuntu-24.04-server-cloudimg-arm64.asif
[2026-04-22T16:59:04.741] [warning] [diskutil] 
[0% completed] Resizing disk image...
[2026-04-22T16:59:04.742] [warning] [diskutil] 
[2% completed] Resizing disk image...
[2026-04-22T16:59:04.845] [warning] [diskutil] 
[8% completed] Resizing disk image...
[2026-04-22T16:59:05.145] [warning] [diskutil] 
[10% completed] Resizing disk image...
[2026-04-22T16:59:05.147] [warning] [diskutil] 
[20% completed] Resizing disk image...
[2026-04-22T16:59:05.147] [warning] [diskutil] 
[20% completed] Resizing disk image...
[20% completed] Fitting partition table to new disk size...
[2026-04-22T16:59:05.260] [warning] [diskutil] 
[20% completed] Resizing disk image...
[30% completed] Fitting partition table to new disk size...
[2026-04-22T16:59:05.260] [warning] [diskutil] 
[20% completed] Resizing disk image...
[30% completed] Fitting partition table to new disk size...
100% completed

[2026-04-22T16:59:05.343] [trace] [applevz-utils] Successfully resized image: /var/root/Library/Application Support/multipassd/applevz/vault/instances/foo/ubuntu-24.04-server-cloudimg-arm64.asif
[2026-04-22T16:59:05.343] [trace] [applevz-vm] initialize_vm_handle() -> Creating VM handle for 'foo'
[2026-04-22T16:59:05.412] [debug] [applevz-vm] set_state() -> VM `foo` VZ state `stopped`
[2026-04-22T16:59:05.412] [info] [applevz-vm] set_state() > VM foo state changed from off to stopped
[2026-04-22T16:59:05.413] [debug] [FileOps] Saved file `/var/root/Library/Application Support/multipassd/applevz/multipassd-vm-instances.json` successfully in attempt #1
[2026-04-22T16:59:05.413] [trace] [applevz-vm] initialize_vm_handle() -> Created handle for VM 'foo'
[2026-04-22T16:59:05.414] [debug] [FileOps] Saved file `/var/root/Library/Application Support/multipassd/applevz/multipassd-vm-instances.json` successfully in attempt #1
[2026-04-22T16:59:05.414] [debug] [applevz-vm] start() -> Starting VM `foo`, current state stopped
[2026-04-22T16:59:05.415] [debug] [FileOps] Saved file `/var/root/Library/Application Support/multipassd/applevz/multipassd-vm-instances.json` successfully in attempt #1
[2026-04-22T16:59:05.416] [debug] [applevz-vm] start() -> starting VM `foo`
[2026-04-22T16:59:05.587] [debug] [applevz-vm] start() -> VM `foo` running
Launched: foo
```

chrisndbg:
@ricab and @sharder996 - I did additional tests on my macMini with macOS 26.3

Basic operations
```
christoph@Christophs-Mac-mini ~ % multipass info
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass start foo
christoph@Christophs-Mac-mini ~ % multipass info     
Name:           foo
State:          Running
Snapshots:      0
IPv4:           192.168.64.6
Release:        Ubuntu 24.04.4 LTS
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         1
Load:           0.44 0.10 0.03
Disk usage:     2.2GiB out of 4.8GiB
Memory usage:   242.2MiB out of 955.2MiB
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass stop foo
christoph@Christophs-Mac-mini ~ % multipass info    
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass clone foo
Cloned from foo to foo-clone1.
christoph@Christophs-Mac-mini ~ % multipass info     
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

Name:           foo-clone1
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass start foo-clone1
christoph@Christophs-Mac-mini ~ % multipass info            
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

Name:           foo-clone1
State:          Running
Snapshots:      0
IPv4:           192.168.64.7
Release:        Ubuntu 24.04.4 LTS
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         1
Load:           0.31 0.07 0.02
Disk usage:     2.2GiB out of 4.8GiB
Memory usage:   349.0MiB out of 955.2MiB
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass stop foo-clone1
christoph@Christophs-Mac-mini ~ % multipass info           
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

Name:           foo-clone1
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass delete foo-clone1
christoph@Christophs-Mac-mini ~ % multipass info             
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

Name:           foo-clone1
State:          Deleted
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass recover foo-clone1 
christoph@Christophs-Mac-mini ~ % multipass info
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

Name:           foo-clone1
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass delete foo-clone1  
christoph@Christophs-Mac-mini ~ % multipass purge
christoph@Christophs-Mac-mini ~ % multipass info             
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass get local.foo.disk
5.0GiB
christoph@Christophs-Mac-mini ~ % multipass set local.foo.disk=7.5G
christoph@Christophs-Mac-mini ~ % multipass get local.foo.disk     
7.5GiB
christoph@Christophs-Mac-mini ~ % multipass info                   
Name:           foo
State:          Stopped
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --
christoph@Christophs-Mac-mini ~ % multipass start foo
christoph@Christophs-Mac-mini ~ % multipass info     
Name:           foo
State:          Running
Snapshots:      0
IPv4:           192.168.64.6
Release:        Ubuntu 24.04.4 LTS
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         1
Load:           0.08 0.02 0.01
Disk usage:     2.2GiB out of 7.2GiB
Memory usage:   272.9MiB out of 955.2MiB
Mounts:         --
```

Updating from a previous version:
```
christoph@Christophs-Mac-mini ~ % multipass --version
multipass   1.17.0-dev.ci8+gcd14724d-noff.mac
multipassd  1.17.0-dev.ci8+gcd14724d-noff.mac

christoph@Christophs-Mac-mini ~ % multipass info
Name:           foo
State:          Running
Snapshots:      0
IPv4:           192.168.252.2
Release:        Ubuntu 24.04.4 LTS
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         1
Load:           0.11 0.04 0.01
Disk usage:     2.1GiB out of 4.8GiB
Memory usage:   202.9MiB out of 952.1MiB

christoph@Christophs-Mac-mini ~ % multipass exec foo -- ls -la 
total 28
drwxr-x--- 4 ubuntu ubuntu 4096 Apr 23 09:33 .
drwxr-xr-x 3 root   root   4096 Apr 23 09:32 ..
-rw-r--r-- 1 ubuntu ubuntu  220 Mar 31  2024 .bash_logout
-rw-r--r-- 1 ubuntu ubuntu 3771 Mar 31  2024 .bashrc
drwx------ 2 ubuntu ubuntu 4096 Apr 23 09:33 .cache
-rw-r--r-- 1 ubuntu ubuntu  807 Mar 31  2024 .profile
drwx------ 2 ubuntu ubuntu 4096 Apr 23 09:32 .ssh

// => Upgraded to new .pkg

christoph@Christophs-Mac-mini ~ % multipass --version
multipass   1.17.0-dev.pr4676+gc23f1405.mac
multipassd  1.17.0-dev.pr4676+gc23f1405.mac

christoph@Christophs-Mac-mini ~ % multipass info
Name:           foo
State:          Running
Snapshots:      0
IPv4:           192.168.252.2
Release:        Ubuntu 24.04.4 LTS
Image hash:     1ea801e659d2 (Ubuntu 24.04 LTS)
CPU(s):         1
Load:           0.04 0.05 0.01
Disk usage:     2.1GiB out of 4.8GiB
Memory usage:   228.9MiB out of 952.1MiB
Mounts:         --

christoph@Christophs-Mac-mini ~ % multipass exec foo -- ls -la
total 28
drwxr-x--- 4 ubuntu ubuntu 4096 Apr 23 09:37 .
drwxr-xr-x 3 root   root   4096 Apr 23 09:32 ..
-rw-r--r-- 1 ubuntu ubuntu  220 Mar 31  2024 .bash_logout
-rw-r--r-- 1 ubuntu ubuntu 3771 Mar 31  2024 .bashrc
drwx------ 2 ubuntu ubuntu 4096 Apr 23 09:33 .cache
-rw-r--r-- 1 ubuntu ubuntu  807 Mar 31  2024 .profile
drwx------ 2 ubuntu ubuntu 4096 Apr 23 09:32 .ssh
-rw-r--r-- 1 ubuntu ubuntu    0 Apr 23 09:37 .sudo_as_admin_successful
```

ricab:
Thanks @christophniederberger-canonical, that's encouraging!

ricab:
So is this good to merge now?

sharder996:
@ricab Are you asking me? I have nothing more to add to the PR.

ricab:
I don't either. I have given my approval for the code.

sharder996:
Alright, hitting the merge button!

PR #4679, opened 2026-02-14
  [daemon] Handle intentional VM shutdown during cloud-init 

### LINKED ISSUE #4456: Support shutdown as the final VM state in cloud-init
Run a cloud-init configuration that leaves the VM in a shutdown state at the end of the process.

**Current behavior**
When such a cloud-init is used in launching a VM (`multipass launch -n a --clout-init cloud-init.yaml`), the launching process will continue after the shutdown until it is interrupted or it timeouts (default 300 seconds or `--timeout value`). This happens because the function [wait_for_cloud_init](https://github.com/canonical/multipass/blob/d1709fa293ddfc19d7dc9178bc7f7cf5b37d0ba0/src/platform/backends/shared/base_virtual_machine.cpp#L291) waits until it can check via ssh the existence of the file `/var/lib/cloud/instance/boot-finished`, signalling the end of the cloud-init configuration. Since this is not possible with a VM in a Shutdown state, it waits until timeout.

This cloud-init functionality would require parsing the cloud-init file to get an intended final state and if it is not Running (including rebooting, since the machine ends up being running) or simply Shutdown, the wait for the cloud-init does not check for the presence of `/var/lib/cloud/instance/boot-finished` but just wait until the VM state is the expected one. 
Another solution is to just add the check for VM Shutoff and consider the cloud-init done if that is the case, since if that is the intended state, the VM will not perform more actions when it reaches that state (parsing in the previous option should be aware of this). The problem of this option is if something in cloud-init causes a premature shutdown due to a bug or some other problem.

**Additional context**
Related to #4199, was raised first by @CoenraadS:
> [Post](https://github.com/canonical/multipass/issues/4199#issuecomment-3310409351)
> I also encountered this using:
> 
> ```
> power_state:
>   mode: poweroff
>   message: "Shutting down VM after preparation"
>   timeout: 30
>   condition: True
> ```
> 
> Which seems the 'correct' way of shutting down after init. Windows + HyperV

deepakshirkem:
# Description

## What does this PR do?
This PR fixes the timeout issue when cloud-init uses `power_state: mode: poweroff` to shut down the VM. Instead of waiting for the full timeout period, Multipass now detects the intentional shutdown and treats the launch as successful.

## Why is this change needed?
Currently, when a user configures cloud-init to shut down the VM after initialization (using `power_state: poweroff`), `multipass launch` waits for SSH to become available and times out after the full timeout period (default 300 seconds), even though the VM has already completed initialization and shut down as intended.

This creates a poor user experience where:
- Launch appears to fail even though cloud-init completed successfully
- Users must wait several minutes for an inevitable timeout
- The VM ends up in the correct (Stopped) state, but the launch command reports failure

## Related Issue(s)
Closes #4456 

### Manual testing steps:
1. Create a cloud-init config with power_state shutdown:
```yaml
   #cloud-config
   runcmd:
     - echo "Cloud-init is running"
     - sleep 5
   
   power_state:
     mode: poweroff
     message: "Shutting down after cloud-init"
     timeout: 30
     condition: True
```

2. Launch VM with the config:
```bash
   multipass launch --name test-vm --cloud-init shutdown.yaml --timeout 120
```

3. Verify the launch completes quickly (~30-40 seconds instead of 120s timeout)

4. Check VM state:
```bash
   multipass list
   # Shows: test-vm   Stopped   --   Ubuntu 24.04 LTS
```

### Results:
**Before fix:**
- Launch times out after full timeout period (120s in test, 300s default)
- Error: "Timed out waiting for instance launch"
- VM ends up in Stopped state but launch reports failure

**After fix:**
- Launch completes in ~30-40 seconds
- Success message: "Launched: test-vm"
- VM in Stopped state as intended
- Logs show: "Instance powered off intentionally during initialization"

Screenshot from 2026-02-14 15-33-29

## Checklist

### Implementation Details:
- Added `IntentionalShutdownException` to signal cloud-init initiated shutdown
- Modified `BaseVirtualMachine::wait_until_ssh_up()` to detect stopped/off state during SSH wait
- Updated `Daemon::async_wait_for_ssh_and_start_mounts_for()` to catch the exception and treat as success
- Enhanced VirtualBox backend's `current_state()` to properly track state during startup

### Considerations:
- This change only affects VMs that intentionally shut down during cloud-init
- Normal launch behavior is unchanged
- VM remains in Stopped state, allowing users to start it later if needed
- The fix respects user intent when using cloud-init's power_state configuration

### Related:
- Originally reported by @CoenraadS
- Uses cloud-init `power_state` as documented in cloud-init documentation

### DIFF
--- a/src/platform/backends/qemu/qemu_img_utils.cpp
+++ b/src/platform/backends/qemu/qemu_img_utils.cpp
@@ -19,15 +19,16 @@
 
 #include <multipass/constants.h>
 #include <multipass/format.h>
+#include <multipass/json_utils.h>
 #include <multipass/memory_size.h>
 #include <multipass/platform.h>
 #include <multipass/process/qemuimg_process_spec.h>
 
-#include <QJsonDocument>
-#include <QJsonObject>
 #include <QRegularExpression>
 #include <QString>
 #include <QStringList>
+
+#include <boost/json.hpp>
 
 namespace mp = multipass;
 namespace mpp = multipass::platform;
@@ -74,9 +75,9 @@ mp::Path mp::backend::convert_to_qcow_if_necessary(const mp::Path& image_path)
         "Cannot read image format");
 
     auto image_info = qemuimg_info_process->read_all_standard_output();
-    auto image_record = QJsonDocument::fromJson(QString(image_info).toUtf8(), nullptr).object();
+    auto image_record = boost::json::parse(std::string_view(image_info));
 
-    if (image_record["format"].toString() == "raw")
+    if (mp::lookup_or<std::string>(image_record, "format", "") == "raw")
     {
         auto qemuimg_convert_spec = std::make_unique<mp::QemuImgProcessSpec>(
             QStringList{"convert", "-p", "-O", "qcow2", image_path, qcow2_path},
--- a/src/platform/backends/qemu/qemu_virtual_machine.cpp
+++ b/src/platform/backends/qemu/qemu_virtual_machine.cpp
@@ -36,9 +36,6 @@
 #include <multipass/vm_status_monitor.h>
 
 #include <QFile>
-#include <QJsonArray>
-#include <QJsonDocument>
-#include <QJsonObject>
 #include <QProcess>
 #include <QString>
 #include <QTemporaryFile>
@@ -127,23 +124,16 @@ auto make_qemu_process(const mp::VirtualMachineDescription& desc,
     return process;
 }
 
-auto qmp_execute_json(const QString& cmd)
+boost::json::object qmp_execute_json(const QString& cmd)
 {
-    QJsonObject qmp;
-    qmp.insert("execute", cmd);
-    return QJsonDocument(qmp).toJson();
+    return {{"execute", cmd.toStdString()}};
 }
 
-auto hmc_to_qmp_json(const QString& command_line)
+boost::json::object hmc_to_qmp_json(const QString& command_line)
 {
-    auto qmp = QJsonDocument::fromJson(qmp_execute_json("human-monitor-command")).object();
-
-    QJsonObject cmd_line;
-    cmd_line.insert("command-line", command_line);
-
-    qmp.insert("arguments", cmd_line);
-
-    return QJsonDocument(qmp).toJson();
+    auto qmp = qmp_execute_json("human-monitor-command");
+    qmp["arguments"] = {{"command-line", command_line.toStdString()}};
+    return qmp;
 }
 
 std::string get_qemu_machine_type(const QStringList& platform_args)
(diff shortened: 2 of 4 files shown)


deepakshirkem:
Hi @tobe2098,

I noticed the codecov report is showing missing coverage for the new code.

Could you please guide me on:
1. What test cases would be appropriate to add for this change?
2. Where should I add tests for `IntentionalShutdownException` and 
   the new `wait_until_ssh_up` behavior?

I want to make sure the tests follow the existing patterns in the codebase.

I'm happy to add the required tests before this gets merged.

Thank you!

tobe2098:
I am very happy to help you with that, but first we should tackle the solution itself. The issue you have chosen is particularly difficult, but if you want to continue working on the solution I will happily help. 

The issue we are trying to solve is being able to identify intentional shutdowns and stop trying to ssh in when/if that is the case. Assuming the VM is in the expected state if it shuts down during start up does not solve the problem. Do you think you could come up with an alternative solution?

Also feel free to work on another issue instead. I reiterate, this issue is not a simple one.

deepakshirkem:
Hi @tobe2098,

My current PR implements your **second proposed solution** from 
#4456- checking if the VM reaches a stopped/off state and 
considering cloud-init complete.

As you pointed out, the problem with this approach is it can't 
distinguish between:
- Intentional shutdown (cloud-init power_state) 
- Premature shutdown due to a bug 

I know this is a difficult issue but I really want to give it a 
try! Your feedback is the most valuable thing for me to get 
this right.

I can see two paths forward:

**Option 1: Parse cloud-init YAML (more robust)**
- Check for `power_state` in user-provided cloud-init config
- Only treat shutdown as intentional if explicitly configured
- Avoids false positives from crashes

**Option 2: Keep current approach + add warning**
- Keep VM shutoff check as-is
- Add a clear warning message to the user
- Document the known limitation

Do you have any other approach or option in mind that would 
be a better fit? I'm completely open to reworking the solution 
if you have a preferred direction!

Thank you !

tobe2098:
I think option 1 is best, but stop relying on LLM text to communicate with us please

deepakshirkem:
Hi @tobe2098, 
Apologies for the LLM text, I will communicate directly from now on.

I am going to implement option 1 (parsing cloud-init YAML). My plan is First i will find where the multipass currently reads the cloud-int YAML file then add the parsing logic there to check for power_state

deepakshirkem:
Hi @tobe2098 ,
I accidentally reset the latest commit  while addressing latest changes and force-pushed the branch, which caused the previous PR to be automatically closed.

I am creating another PR with latest changes . Sorry for the confusion.

Thank You!!

PR #4688, opened 2026-02-18
[qemu] Use AZ subnets on macOS

### LINKED ISSUE #3147: Subnet used for default bridge interface for qemu is "random"

jimporter:
# Description

In the `availability-zones` branch, we store subnet information in the files for each AZ. We should use this information when setting up qemu's networking for the VM on macOS.

I don't have access to a macOS device, so I'm speculatively making these changes. I'm also not able to manually trigger the CLI tests, since the [GitHub documentation](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow) says that manually-triggered workflows need to be defined on the default branch.

## Related Issue(s)

In addition to helping finish the AZ feature in general, this should also improve matters for #4383 by ensuring that we always use available subnets. However, note that this isn't robust enough yet to resolve #3147.

## Testing

- Unit tests updated to account for new flags
- Manual testing steps:

  1. Start Multipass daemon on macOS with qemu
  2. Launch a multipass VM and ensure connectivity

## Checklist

### DIFF
--- a/src/platform/backends/qemu/macos/qemu_platform_detail.h
+++ b/src/platform/backends/qemu/macos/qemu_platform_detail.h
@@ -19,7 +19,10 @@
 
 #include <qemu_platform.h>
 
+#include <multipass/availability_zone_manager.h>
 #include <multipass/path.h>
+
+#include <unordered_map>
 
 namespace multipass
 {
@@ -27,7 +30,7 @@ namespace multipass
 class QemuPlatformDetail : public QemuPlatform
 {
 public:
-    explicit QemuPlatformDetail();
+    explicit QemuPlatformDetail(const AvailabilityZoneManager::Zones& zones);
 
     std::optional<IPAddress> get_ip_for(const std::string& hw_addr) override;
     void remove_resources_for(const std::string& name) override;
@@ -41,7 +44,12 @@ public:
     void set_authorization(std::vector<NetworkInterfaceInfo>& networks) override;
 
 private:
+    using Bridges = std::unordered_map<std::string, Subnet>;
+
+    [[nodiscard]] static Bridges get_bridges(const AvailabilityZoneManager::Zones& zones);
+
     const QString host_arch{HOST_ARCH};
     const QStringList common_args;
+    const Bridges bridges;
 };
 } // namespace multipass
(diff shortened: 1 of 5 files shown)


sharder996 (approved):
Looking good to me on macOS:
```
scott@Scotts-MacBook-Pro build-gui % multipass ls
foo                     Running           10.97.0.4        Ubuntu 24.04 LTS    zone1(a)
foo2                    Running           10.97.1.2        Ubuntu 24.04 LTS    zone2(a)
foo3                    Running           10.97.2.2        Ubuntu 24.04 LTS    zone3(a)
```

One question for my own understanding, though. How come VMs in different zones live under `10.97.n` and not under the standard subnet `192.168.n`?

jimporter:
@sharder996 I believe the `97` is random, but looking through the Git history, we've been using IPs in `10.0.0.0/8` since the very beginning. Both `10.0.0.0/8` and `192.168.0.0/16` are standard subnets for local networks though: https://en.wikipedia.org/wiki/List_of_reserved_IP_addresses

I guess using something in `10.0.0.0/8` has a smaller chance of overlapping with default home networks though, which tend to use something in `192.168.0.0/16`.

PR #4695, opened 2026-02-21
[daemon] Handle cloud-init intentional shutdown

### LINKED ISSUE #4456: Support shutdown as the final VM state in cloud-init

deepakshirkem:
# Description

## What does this PR do?
This PR fixes the timeout issue when cloud-init uses  `power_state: mode: poweroff` to shutdown the VM. Insted of waiting for the full timeout period, Multipass now:

- Parse the cloud-init  YAML to detect if `power_state` is configured. 
- Distinguishes between intentional shutdown (configured in cloud-init) and unexpected crashes
- Treats intentional shutdown as successful launch completion

## Why is this change needed?
Currently, when a user configures cloud-init to shut down the VM after initialization, `multipass launch` waits for SSH and times out after 300 seconds, even though the VM has completed initialization successfully.

This creates a poor user experience:
- Launch appears to fail even though cloud-init completed successfully
- Users must wait several minutes for an inevitable timeout
- The VM ends up in the correct (Stopped) state, but launch reports failure

## Implementation Details

- Parse `power_state` field from user-provided cloud-init YAML during VM creation
- Store `expects_shutdown` flag in `VirtualMachineDescription` and pass to VM
- In `wait_until_ssh_up()`, If `expects_shutdown = true`: throw `IntentionalShutdownException` (success) and    If `expects_shutdown = false`: throw `StartException` (failure)
- Daemon catches `IntentionalShutdownException` and treats launch as successful

## Related Issue(s)
May be Closes #4456 

### Manual testing:

**Test 1: Intentional shutdown (with power_state)**
```yaml
#cloud-config
runcmd:
  - echo "Processing data"
  - sleep 5

power_state:
  mode: poweroff
  message: "Shutting down after init"
```
Result: Launch completes in ~30-40 seconds with success message, VM in Stopped state

**Test 2: Normal VM (no power_state)**
```bash
multipass launch --name normal-vm
```
Result:  Launch completes normally, VM in Running state with SSH available

**Test 3: Unexpected crash (VM stopped without power_state)**
Manually stop VM during launch:
```bash
VBoxManage controlvm test-vm poweroff
```
Result:  Launch fails with "VM stopped unexpectedly" error

N/A

## Checklist
- [ ] I have added necessary tests (need guidance on test structure)

## Additional Notes
N/A

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -29,6 +29,7 @@
 #include <multipass/exceptions/exitless_sshprocess_exceptions.h>
 #include <multipass/exceptions/ghost_instance_exception.h>
 #include <multipass/exceptions/image_vault_exceptions.h>
+#include <multipass/exceptions/intentional_shutdown_exception.h>
 #include <multipass/exceptions/invalid_memory_size_exception.h>
 #include <multipass/exceptions/not_implemented_on_this_backend_exception.h>
 #include <multipass/exceptions/snapshot_exceptions.h>
@@ -683,7 +684,7 @@ const std::string& get_instance_name(InstanceElem instance_element)
 }
 
 template <typename... Ts>
-auto add_fmt_to(fmt::memory_buffer& buffer, fmt::format_string<Ts...> fmt, Ts&&... fmt_params)
+auto add_fmt_to(fmt::memory_buffer& buffer,fmt::format_string<Ts...> fmt, Ts&&... fmt_params)
     -> std::back_insert_iterator<fmt::memory_buffer>
 {
     if (buffer.size())
@@ -3400,7 +3401,6 @@ void mp::Daemon::create_vm(const CreateRequest* request,
             vm_desc.meta_data_config = mpu::make_cloud_init_meta_config(name);
             vm_desc.user_data_config = YAML::Load(request->cloud_init_user_data());
             prepare_user_data(vm_desc.user_data_config, vm_desc.vendor_data_config);
-
             if (vm_desc.num_cores < std::stoi(mp::min_cpu_cores))
                 vm_desc.num_cores = std::stoi(mp::default_cpu_cores);
 
@@ -3670,18 +3670,35 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
             return fmt::to_string(errors);
         }
         const auto vm = it->second;
-        vm->wait_until_ssh_up(timeout);
 
-        if (std::is_same<Reply, LaunchReply>::value)
+        try
         {
-            if (server)
-            {
-                Reply reply;
-                reply.set_reply_message("Waiting for initialization to complete");
-                server->Write(reply);
-            }
+            vm->wait_until_ssh_up(timeout);
 
-            vm->wait_for_cloud_init(timeout);
+            if (std::is_same<Reply, LaunchReply>::value)
+            {
+                if (server)
+                {
+                    Reply reply;
+                    reply.set_reply_message("Waiting for initialization to complete");
+                    server->Write(reply);
+                }
+
+                vm->wait_for_cloud_init(timeout);
+            }
+        }
+        catch (const mp::IntentionalShutdownException& e)
+        {
+            if constexpr (std::is_same_v<Request, LaunchRequest>)
+            {
+                mpl::log(mpl::Level::info, name,
+                        "Instance powered off intentionally during initialization");
+                return {}; // Success - intentional shutdown
+            }
+            else
+            {
+                throw StartException(name, "Unexpected shutdown during start");
+            }
         }
 
         if (MP_SETTINGS.get_as<bool>(mp::mounts_key))
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -3712,17 +3712,7 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
                 {
                     if (!mount->is_mount_managed_by_backend())
                     {
-                        if (vm->current_state() == VirtualMachine::State::running)
-                        {
-                            mount->activate(server);
-                        }
-                        else
-                        {
-                            mpl::log(mpl::Level::info,
-                                     name,
-                                     "Mount '{}' saved but not activated (VM stopped)",
-                                                 target);
-                        }
+                        mount->activate(server);
                     }
                 }
                 catch (const mp::SSHFSMissingError&)
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -684,8 +684,9 @@ const std::string& get_instance_name(InstanceElem instance_element)
 }
 
 template <typename... Ts>
-auto add_fmt_to(fmt::memory_buffer& buffer,fmt::format_string<Ts...> fmt, Ts&&... fmt_params)
-    -> std::back_insert_iterator<fmt::memory_buffer>
+auto add_fmt_to(fmt::memory_buffer& buffer,
+                fmt::format_string<Ts...> fmt,
+                Ts&&... fmt_params) -> std::back_insert_iterator<fmt::memory_buffer>
 {
     if (buffer.size())
         buffer.push_back('\n');
@@ -969,8 +970,8 @@ mp::MemorySize compute_final_image_size(const mp::MemorySize image_size,
     if (!command_line_value)
     {
         auto default_disk_size_as_struct = mp::MemorySize(mp::default_disk_size);
-        disk_space =
-            image_size < default_disk_size_as_struct ? default_disk_size_as_struct : image_size;
+        disk_space = image_size < default_disk_size_as_struct ? default_disk_size_as_struct
+                                                              : image_size;
     }
     else if (*command_line_value < image_size)
     {
@@ -1655,8 +1656,8 @@ try
             auto remote_name = (!request->remote_name().empty() ||
                                 (request->remote_name().empty() && vm_images_info.size() > 1 &&
                                  remote != mp::release_remote))
-                                   ? remote
-                                   : "";
+                                 ? remote
+                                 : "";
 
             add_aliases(response.mutable_images_info(), remote_name, info);
         }
@@ -1741,8 +1742,8 @@ try
         const auto& name = vm.get_name();
 
         const auto& it = instance_snapshots_map.find(name);
-        const auto& snapshot_pick =
-            it == instance_snapshots_map.end() ? SnapshotPick{{}, true} : it->second;
+        const auto& snapshot_pick = it == instance_snapshots_map.end() ? SnapshotPick{{}, true}
+                                                                       : it->second;
 
         try
         {
@@ -2007,8 +2008,8 @@ try
         }
 
         const auto mount_type = request->mount_type() == MountRequest_MountType_CLASSIC
-                                    ? VMMount::MountType::Classic
-                                    : VMMount::MountType::Native;
+                                  ? VMMount::MountType::Classic
+                                  : VMMount::MountType::Native;
 
         VMMount vm_mount{request->source_path(), gid_mappings, uid_mappings, mount_type};
         vm_mounts[target_path] = make_mount(vm.get(), target_path, vm_mount);
@@ -2130,8 +2131,8 @@ try
                                                        *config->logger,
                                                        server};
 
-    auto timeout =
-        request->timeout() > 0 ? std::chrono::seconds(request->timeout()) : mp::default_timeout;
+    auto timeout = request->timeout() > 0 ? std::chrono::seconds(request->timeout())
+                                          : mp::default_timeout;
 
     if (!instances_running(operative_instances))
         config->factory->hypervisor_health_check();
@@ -2316,8 +2317,8 @@ try
         *config->logger,
         server};
 
-    auto timeout =
-        request->timeout() > 0 ? std::chrono::seconds(request->timeout()) : mp::default_timeout;
+    auto timeout = request->timeout() > 0 ? std::chrono::seconds(request->timeout())
+                                          : mp::default_timeout;
 
     auto [instance_selection, status] =
         select_instances_and_react(operative_instances,
@@ -2423,8 +2424,8 @@ try
 
                 auto snapshot_pick_it = instance_snapshots_map.find(instance_name);
                 const auto& [pick, all] = snapshot_pick_it == instance_snapshots_map.end()
-                                              ? SnapshotPick{{}, true}
-                                              : snapshot_pick_it->second;
+                                            ? SnapshotPick{{}, true}
+                                            : snapshot_pick_it->second;
 
                 if (!all || !purge) // if we're not purging the instance, we need to delete
                                     // specified snapshots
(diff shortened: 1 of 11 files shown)


deepakshirkem:
Hi @tobe2098 , I will required your feedback on this solution approach. 
One thing I am confuse about the CI is showing test failure in `test_base_virtual_machine_factory.cpp` due to missing `expected_shutdown`  field initializer. However when i ran `ctest -VV` locally , all test passed without any errors.

tobe2098:
Hi, it is a compilation error, unrelated to `ctest -VV`. You should attempt to recompile

deepakshirkem:
Hi @tobe2098 ,
I resolved those failing test can you please review them.

src/daemon/daemon.cpp
```diff
@@ -3280,6 +3284,16 @@ void mp::Daemon::create_vm(const CreateRequest* request,
...
+            if(vm_desc.user_data_config["power_state"])
+            {
+                auto ps = vm_desc.user_data_config["power_state"];
+                if(ps["mode"] && ps["mode"].as<std::string>() == "poweroff")
+                {
+                    mpl::log(mpl::Level::error, name, "DETECTED POWEROFF IN CONFIG");
+                    vm_desc.expects_shutdown = true;
+                }
+            }
+
```
tobe2098: Since vm_desc already has access to the data, would it make more sense to just parse the power_state value in-situ? Once you enter the function call that requires the information it could be parsed there for the value, instead of adding an additional field that does not really correspond to "VMDescription", since it is just an option of the contained cloud-init (if we were talking about a ParsedCloudInit object I would agree with the approach).

If the field were in BaseVirtualMachine it would be more convenient, since that virtual class does not have access to the VMDescription. The default could be false, and the field could be set in the constructor of the derived classes, where there is access to the VMDescription. What do you think?

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
```diff
@@ -448,8 +449,17 @@ mp::VirtualMachine::State mp::VirtualBoxVirtualMachine::current_state()
...
+        mpl::log(mpl::Level::info,
+                 name.toStdString(),
+                 "VM stopped during startup (cloud-init poweroff)");
+
+        state = present_state;
+        return state;
+    }
+
+    if (state == State::delayed_shutdown && present_state == State::running)
         return state;
```
tobe2098, line 456: Is this logic really necessary? Supporting the intentional shutdown should be backend-wide, and this provides no additional functionality. Let me know what your thinking process here was, there are other logic changes here (like what if starting && not stopped?), for which an explanation would make things easier.

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
```diff
@@ -448,8 +449,17 @@ mp::VirtualMachine::State mp::VirtualBoxVirtualMachine::current_state()
...
-    if ((state == State::delayed_shutdown && present_state == State::running) ||
-        state == State::starting)
+    if (state == State::starting && present_state == State::stopped)
+    {
+        mpl::log(mpl::Level::info,
+                 name.toStdString(),
+                 "VM stopped during startup (cloud-init poweroff)");
+
+        state = present_state;
+        return state;
```
tobe2098: This updates the state variable when starting when it used to not be the case. What was your reasoning for this change?

tobe2098 (changes requested):
The work shows promise, but I think we need to re-formulate the solution. Ideally, wait_* functions would return as if there were a success in the intentional shutdown case, since it is a success. Both wait_* functions were already re-formatted to accomodate restarts, so in the case where a restart would be detected within the try_to_ssh function due to state not being running, we could check if the state is shutdown and the intended state is shutdown as well and return TimeoutAction::Done. 

What do you think about this alternative solution?

src/daemon/daemon.cpp
deepakshirkem: Yes i think your approach make more sense.
My thinking was that parsing in the daemon would make the  implementation work across all backends  automatically.  I will try implementing it with the VirtualBox backed first and update here.

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
deepakshirkem: You are right, this does not support any logic. I added this only for debugging purpose while testing, to check the logs. I will remove this changes in the next commit. I will make sure that this type of mistake will not happen again.

deepakshirkem:
Hi @tobe2098,
I have updated the PR as per your feedback. Thank you for your input.

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -263,17 +264,47 @@ void mp::BaseVirtualMachine::wait_until_ssh_up(std::chrono::milliseconds timeout
...
+            if (expected_shutdown) {
+                mpl::log(mpl::Level::info, vm_name,
+                         "VM powered off as configured in cloud-init");
+                return utils::TimeoutAction::done;
+            } else {
+                mpl::log(mpl::Level::error, vm_name,
+                         "VM stopped unexpectedly");
+                throw StartException(vm_name, "VM stopped unexpectedly");
+            }
+        }
```
tobe2098, line 327: Maybe the exception could be thrown here instead of after the try_action_for.

tobe2098 (changes requested):
We are getting closer @deepakshirkem. It is not complete because it could be the case that the shutdown is detected on waiting for cloud-init, instead of on ssh up. That would be the case if the cloud-init takes long enough in longer cloud-init configurations.

We should also add the logic to the other backends. To avoid the repeated code, the yaml parsing code could be a function as well.

src/platform/backends/shared/base_virtual_machine.cpp
deepakshirkem: Hi @tobe2098, I tried throwing the exception as you suggested and removed the final state check, but after that I am not able to see the intentional exception or the start exception. I am not sure why this behavior changed. I will dig deeper into it and add logging to investigate.

tobe2098: Remember that if the shutdown happens during the cloud-init after the ssh_up you are not throwing in that function yet.

deepakshirkem: Yeah, you are right. But in my updated code, I am throwing the exception from that function. After some debugging, I think there may be a race condition issue—my vm_state is not updating inside the lambda.

So, can we do the same as in the old code where I added a final check? It introduces some delay, but it gives the expected results.

I wanted your feedback—does my observation about a race condition make sense?

deepakshirkem:
Hi @tobe2098, I implemented the changes as you suggested and added the helper function. I was confused during testing because, while doing the crash test manually, I stopped the VM before it reached the running state. That’s why it never stopped as expected, and I was not getting the expected behavior.
Now I have updated the logic and also increased the detection speed. I tested this on the QEMU backend.
Please review the changes. I am still very open to any new suggestions or improvements

tobe2098:
Hi @deepakshirkem, if there are conflicts do not merge main into the branch, do `git rebase main`. This way the commit history is kept cleaner. Thank you very much.

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
```diff
@@ -447,9 +448,7 @@ void mp::VirtualBoxVirtualMachine::suspend()
 mp::VirtualMachine::State mp::VirtualBoxVirtualMachine::current_state()
 {
     auto present_state = instance_state_for(name);
-
-    if ((state == State::delayed_shutdown && present_state == State::running) ||
-        state == State::starting)
+    if (state == State::delayed_shutdown && present_state == State::running)
```
tobe2098: Why is `starting` not checked now? Did you find something while testing?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -298,6 +330,7 @@ void mp::BaseVirtualMachine::wait_for_cloud_init(std::chrono::milliseconds timeo
     auto on_timeout = [] {
         throw std::runtime_error("timed out waiting for initialization to complete");
     };
+
```
tobe2098: Could this newline be removed?

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -284,12 +316,12 @@ void mp::BaseVirtualMachine::wait_for_cloud_init(std::chrono::milliseconds timeo
...
-        catch (const SSHExecFailure& e)
-        {
+        catch (const SSHExecFailure& e) // transitioning away from catching generic runtime errors                         
+        {                                // TODO remove once we're confident this is an anomaly
             return mpu::TimeoutAction::retry;
         }
-        catch (const std::exception& e) // transitioning away from catching generic runtime errors
-        {                               // TODO remove once we're confident this is an anomaly
+        catch (const std::exception& e)
+        {
```
tobe2098: This comment was moved accidentally

src/platform/backends/shared/base_virtual_machine.cpp
tobe2098: It makes sense, but if you check the vm_state=current_state(); the state is supposed to be properly updated. I can do more testing on that later on.

tobe2098 (changes requested):
I made some minor comments. You are doing great work @deepakshirkem!
Additionally, there is something that must be taken care of. In `daemon.cpp`, `wait_for_ssh_up` is called in all started/restarted instances, not only the launched ones. When dealing with the IntentionalShutdownException, we must treat it as a StartException whenever it is not a LaunchRequest.

src/platform/backends/shared/base_virtual_machine.cpp
deepakshirkem: Hi @tobe2098, You're right! I was testing the crash scenario incorrectly. I was stopping the VM before it reached the running state, which is why the state wasn't updating as expected in the lambda. That's why I thought there was a race condition.

deepakshirkem:
Hi @tobe2098,  I addressed all your review comments.

Thank You ((:

src/platform/backends/virtualbox/virtualbox_virtual_machine.cpp
deepakshirkem: Hi @tobe2098, I have addressed this  change.

src/daemon/daemon.cpp
```diff
@@ -3670,18 +3670,35 @@ error_string mp::Daemon::async_wait_for_ssh_and_start_mounts_for(
...
+            {
+                mpl::log(mpl::Level::info, name,
+                        "Instance powered off intentionally during initialization");
+                return {}; // Success - intentional shutdown
+            }
+            else
+            {
+                throw StartException(name, "Unexpected shutdown during start");
+            }
         }
```
tobe2098, line 3703: There is an additional topic: when launching a vm, mount options can be provided, which should still be added to the VM even if it is shut off. The current approach would discard the mounts by returning early. Of course, the classic mounts need to check that the vm is running, similar as is done in the Daemon::mount function. What do you think?

tobe2098 (changes requested):

src/daemon/daemon.cpp
deepakshirkem: Hi @tobe2098, Your right. I have to check existing code and provide solution that will handle this case. Thank you.

deepakshirkem:
Hi @tobe2098, Please go one more time on the changes and sorry for those test failure i tested locally now no test is failing. Please re-run the pipeline when you get a chance.  Thank you ::)

tobe2098 (changes requested):
Sorry to not have been clear with the mounts activation. The mounts are activated in `src/client/cli/launch.cpp` when launching. The way it works is that if the launch returns an OK code, the mounts will be attempted on the VM as a separate command. The changes you added can be undone, simply test that then launching a VM that starts and shuts down with the `cloud-init`, that the mounts via mount flags are stored.

tests/unit/test_base_virtual_machine.cpp
```diff
@@ -1350,6 +1350,7 @@ TEST_F(BaseVM, waitForCloudInitVMDownReconnects)
         .WillOnce(Return(mp::VirtualMachine::State::running))    // when 1st waiting for cloud-init
         .WillOnce(Return(mp::VirtualMachine::State::restarting)) // when retrying to ssh
         .WillOnce(Return(mp::VirtualMachine::State::running));   // back waiting for cloud-init
+        .WillOnce(Return(mp::VirtualMachine::State::running));   // back to cloud-init
```
deepakshirkem, line 1353: Added this call. Do you think this is the right approach? The test is passing locally, but I observed that it is failing in CI.

tobe2098:
Very well done @deepakshirkem. Thank you for your patience :D. I think the solution is achieved well, but the second commit's changes are not necessary as I stated in my previous message. The mounts for `multipass launch` are performed through an external `multipass mount` call, not through the `async_*` function. Otherwise it looks good to me :).

src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -314,8 +337,23 @@ void mp::BaseVirtualMachine::wait_until_ssh_up(std::chrono::milliseconds timeout
...
+            }
+            else
+            {
+                mpl::log(mpl::Level::error, vm_name, "VM stopped unexpectedly during cloud-init");
+                throw StartException(vm_name, "VM Stopped unexpectedly during cloud-init");
+            }
+        }
         try
         {
             ssh_exec("[ -e /var/lib/cloud/instance/boot-finished ]");
```
tobe2098, line 359: I just noticed an additional problem. expected_shutdown remains as `true` after launching and stopping which means that on a second start, a failed start would appear as a intentional shutdown.

tobe2098 (changes requested):
Hi again @deepakshirkem, I just noticed that the approach we went for has a small flaw. The issue itself is more complex than expected. Would you mind if I do some additional work on top of your commits? I believe the problem will require more extensive thought but I want to keep your contributions. What do you think?

deepakshirkem:
> Hi again @deepakshirkem, I just noticed that the approach we went for has a small flaw. The issue itself is more complex than expected. Would you mind if I do some additional work on top of your commits? I believe the problem will require more extensive thought but I want to keep your contributions. What do you think?

Hi @tobe2098, Of course! You can build on top of these commits. I'm always available if you need anything from my side

deepakshirkem:
Hi @tobe2098, I accidentally deleted my fork repository, which caused this PR to be closed. I still have all the changes locally. Would you like me to create a new PR with the same changes?

tobe2098:
Yes, if possible do so please. I will pull them into another branch

PR #4702, opened 2026-02-23
[image] Filter unsupported images at manifest parsing

### LINKED ISSUE #3934: Handle unsupported images better at launch
`multipass launch 16.04` behaves pretty much the same as `multipass launch lts`, except that the former fails to complete, because the image is no longer supported. 

We decided a while ago to scope such images out of what Multipass is supposed to support. See https://github.com/canonical/multipass/issues/3443 and https://github.com/canonical/multipass/issues/1068 for reference. However, the UX is not great today. People [keep](https://github.com/canonical/multipass/issues/3874) [trying](https://github.com/canonical/multipass/issues/3911#issuecomment-2638037239) this, seeing it fail in an ungraceful way, which must be disappointing.

I propose Multipass does one of 3 things:

1. Stop offering unsupported images entirely (people could still launch them from files or URLs directly).
2. Allow launching unsupported images only if the user passes a new `--allow-unsupported` flag. We should then probably require this flag for any file-/URL-based launches.
3. Warn the user that the image is unsupported. Warnings are not shown by default though, so we'd need to think about how to achieve this.

Since we want to allow file/URL based launches (allowing images that are not supported ubuntu releases) I think my preference would be option 2. 

**Additional Context**

```
$ multipass find xenial
No images or blueprints found.
$ multipass find --show-unsupported xenial
Image                       Aliases           Version          Description
xenial                                        20211001         Ubuntu 16.04 LTS
daily:xenial                                  20211001         Ubuntu 16.04 LTS
```

deepakshirkem:
# Description

### What does this PR do?
  Filters out unsupported (EOL) images at the simplestreams manifest  parsing level, so they never enter the system at all. 
Attempting to launch an unsupported image by name is handled just like launching an unknown image. URL-based launches remain available.

###  Why is this change needed?
Previously, Multipass was aware of unsupported images and had special code to deal with them (UnsupportedImageException, --show-unsupported flag). This change removes that complexity entirely by filtering at the root.

## Related Issue(s)

Closes #3934 

## Testing

- Manual testing steps:
  1. Run `multipass launch xenial` — returns unknown image error
  2. Run `multipass launch 22.04` — launches successfully
  3. Run `multipass find` — should only show supported images

Screenshot from 2026-02-25 02-31-31

## Checklist

## Additional Notes

NA

### DIFF
--- a/src/image_host/ubuntu_image_host.cpp
+++ b/src/image_host/ubuntu_image_host.cpp
@@ -19,7 +19,6 @@
 #include <multipass/exceptions/download_exception.h>
 #include <multipass/exceptions/image_not_found_exception.h>
 #include <multipass/exceptions/manifest_exceptions.h>
-#include <multipass/exceptions/unsupported_image_exception.h>
 #include <multipass/image_host/ubuntu_image_host.h>
 #include <multipass/platform.h>
 #include <multipass/query.h>
@@ -108,9 +107,6 @@ std::vector<std::pair<std::string, mp::VMImageInfo>> mp::UbuntuVMImageHost::all_
 
         if (const auto* info = match_alias(key, *manifest); info)
         {
-            if (!info->supported && !query.allow_unsupported)
-                throw mp::UnsupportedImageException(query.release);
-
             images.emplace_back(remote_name, *info);
         }
         else
@@ -119,7 +115,7 @@ std::vector<std::pair<std::string, mp::VMImageInfo>> mp::UbuntuVMImageHost::all_
 
             for (const auto& entry : manifest->products)
             {
-                if (entry.id.startsWith(key) && (entry.supported || query.allow_unsupported) &&
+                if (entry.id.startsWith(key) &&
                     found_hashes.find(entry.id.toStdString()) == found_hashes.end())
                 {
                     images.emplace_back(remote_name, entry);
@@ -148,19 +144,13 @@ mp::VMImageInfo mp::UbuntuVMImageHost::info_for_full_hash_impl(const std::string
     throw mp::ImageNotFoundException(full_hash);
 }
 
-std::vector<mp::VMImageInfo> mp::UbuntuVMImageHost::all_images_for(const std::string& remote_name,
-                                                                   const bool allow_unsupported)
+std::vector<mp::VMImageInfo> mp::UbuntuVMImageHost::all_images_for(const std::string& remote_name)
 {
     std::vector<mp::VMImageInfo> images;
     auto manifest = manifest_from(remote_name);
 
-    for (const auto& entry : manifest->products)
-    {
-        if (entry.supported || allow_unsupported)
-        {
-            images.push_back(entry);
-        }
-    }
+    for (const auto& entry : manifest->products)  
+        images.push_back(entry);
 
     if (images.empty())
         throw std::runtime_error(
--- a/src/simplestreams/simple_streams_manifest.cpp
+++ b/src/simplestreams/simple_streams_manifest.cpp
@@ -136,6 +136,8 @@ try
         const auto supported = lookup_or<bool>(product, "supported", false) ||
                                product_aliases.contains("devel") ||
                                (os == "ubuntu-core" && image_type == "stable");
+        if (!supported)
+            continue;
 
         const auto* versions = if_contains_object(product, "versions");
         if (!versions || versions->empty())
--- a/tests/unit/test_simple_streams_manifest.cpp
+++ b/tests/unit/test_simple_streams_manifest.cpp
@@ -204,4 +204,20 @@ TEST_F(TestSimpleStreamsManifest, correctlyMutatesCoreImages)
     EXPECT_EQ(info->release_codename, "Core 22");
 }
 
+TEST_F(TestSimpleStreamsManifest, filtersUnsupportedImages)
+{
+    auto json = mpt::load_test_file("simple_streams_manifest/unsupported_manifest.json");
+    auto manifest = mp::SimpleStreamsManifest::fromJson(json, std::nullopt, "");
+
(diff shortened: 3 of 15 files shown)


ricab (changes requested):
Hi @deepakshirkem, thank you for your interest. I am afraid this approach is too superficial. We'd be giving up an essential benefit in dropping unsupported images: not having to maintain any code to dealt with them at all. That would mean filtering them out at the root (when parsing simplestreams) and removing any code that dealt with them specifically. There would be no more `--show-unsupported` in `find` or `UnsupportedImageException`. Saying `multipass launch bionic` would effectively be handled just like `multipass launch unknown`.

deepakshirkem:
Hi @ricab, I should have though this through more carefully.  
I will start by looking into the simplestreams parsing logic and filter out unsupported images there before they enter the system. I will be making incremental commits to this PR as i progress. Is this right approach before i proceed with implementation?

ricab:
Hi @deepakshirkem, yes, that sounds reasonable. Note that the code is already aware of what images are considered unsupported. It should now skip them altogether.

Be sure to have a look at our contribution guidelines at https://github.com/canonical/multipass/blob/main/CONTRIBUTING.md and good luck!

deepakshirkem:
Hi @ricab, I have implemented requested changes please review.

deepakshirkem:
Hi @ricab, I just wanted to follow up on this PR when you get a chance. I understand you have a busy schedule. Just wanted to make sure this didn't get lost!

Thank You ::))

ricab:
Hi @deepakshirkem, sure, this is not lost, but things with higher priority keep getting in front and I haven't found the time yet.  Hope to get to it in the next couple of weeks.

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -135,7 +135,10 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(
...
             product["supported"].toBool() || product_aliases.contains("devel") ||
             (product["os"] == "ubuntu-core" && product["image_type"] == "stable");
-
+        
+        if (!supported)
+            continue;
+        
         const auto versions = product["versions"].toObject();
         if (versions.isEmpty())
             continue;
```
ricab: It's a detail, but if we are going to skip, better move the check as early as possible in the loop. For consistency, you can move the check on versions too, and pack the two checks together with the arch check above. You'll still need `product_aliases` though.

ricab (changes requested):
Hi @deepakshirkem, it looks like you did not fully build the project?? There are compilation errors. 

Can you please address copilot's complaints and rebase on `main`? Thank you.

src/simplestreams/simple_streams_manifest.cpp
deepakshirkem: Hi @ricab,  I have  moved the supported check earlier in the loop and packed it together with the versions check Both checks now happen as early as possible after the arch check,  keeping product_aliases available for the supported calculation.

deepakshirkem:
Hi @ricab ,  Apologies for the compilation errors in the previous commit. I had built locally without tests enabled. I have now enabled tests and fixed all the issues.

Thank You ::))

src/client/cli/cmd/find.cpp
```diff
@@ -115,11 +113,8 @@ mp::ParseCode cmd::Find::parse_args(mp::ArgParser* parser)
         {
             request.set_search_string(search_string.toStdString());
         }
-    }
-
-    request.set_allow_unsupported(parser->isSet(unsupportedOption));
+    }  
     request.set_force_manifest_network_download(parser->isSet(force_manifest_network_download));
-
```
ricab, line 122: This still has trailing whitespace. In fact, there is no reason to change these lines at all.

src/image_host/ubuntu_image_host.cpp
```diff
@@ -150,24 +146,18 @@ mp::VMImageInfo mp::UbuntuVMImageHost::info_for_full_hash_impl(const std::string
...
-
+        images.push_back(entry);       
+    
     if (images.empty())
         throw std::runtime_error(
             fmt::format("Unable to find images for remote \"{}\"", remote_name));
-
+            
     return images;
 }
```
ricab, line 160: idem: whitespace

ricab (changes requested):

deepakshirkem:
Hi @ricab, I address all suggestion and rebase with main. Please review once you get chance.

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -136,6 +136,8 @@ try
         const auto supported = lookup_or<bool>(product, "supported", false) ||
                                product_aliases.contains("devel") ||
                                (os == "ubuntu-core" && image_type == "stable");
+        if (!supported)
+            continue;
```
ricab, line 140: Now if we're going to skip these, no point having the supported flag in `VMImageInfo`, right?

tests/unit/test_simple_streams_manifest.cpp
```diff
@@ -204,4 +204,20 @@ TEST_F(TestSimpleStreamsManifest, correctlyMutatesCoreImages)
...
+    auto info = manifest->image_records.find("bionic");
+    EXPECT_NE(info, manifest->image_records.end());
+
+    info = manifest->image_records.find("xenial");
+    EXPECT_EQ(info, manifest->image_records.end());
+
+    info = manifest->image_records.find("16.04");
+    EXPECT_EQ(info, manifest->image_records.end());
+}
+} // namespace
```
ricab, line 223: Missing EOL.

ricab (changes requested):
Not quite there yet. If we're going to skip unsupported images entirely, we should not need to mark images as un/supported.

PR #4706, opened 2026-02-26
[cli] Disable spinner animation on non-live TTYs

### LINKED ISSUE #1541: Fix spinner for non-interactive use
Attempting to run non-interactive scripts that run multipass. Logs look like this:
```
Creating VMs...

                                                                                

                                                                                
Creating test-0  / 
                                                                                

                                                                                
Configuring test-0  /- 
                                                                                

                                                                                
Starting test-0  /-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\ 
                                                                                

                                                                                
Waiting for initialization to complete  /-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\| 
                                                                                
Launched: test-0

                                                                                
Created VM test-0.

```

This could be resolved by disabling spinners when not running on a live terminal, or via switch.

riken127:
Gating the spinner animation logic on the term->cout_is_live() flag. This prevents messy output (backspaces and animation frames) when Multipass is run in non-interactive environments.

# Description

This PR gates the CLI spinner's animation logic on the `term->cout_is_live()` flag.
- What does this PR do? Updates AnimatedSpinner's constructor to accept a TTY status flag and uses that flag to decide whether to spawn a background animation thread or simply output the message as static text.
- Why is this change needed? When Multipass is run in non-interactive environments, the animation frames create noisy logs. This change ensures a clean, static output in those cases.

## Related Issue(s)

Closes #1541 

## Testing

- Manual testing steps:
1. Built the project
2. Ran `./build/bin/multipass wait-ready` (Verified animation is visible)
3. Ran with a cat pipe `./build/bin/multipass wait-ready | cat`

Got:
```
riken@riken:~/programming/multipass$ ./build/bin/multipass wait-ready
Waiting for the Multipass daemon to be ready |^C
riken@riken:~/programming/multipass$ ./build/bin/multipass wait-ready | cat
Waiting for the Multipass daemon to be ready^C
```

N/A (Behavioural change for terminal output)

## Checklist

- [x] I have added necessary tests // only did manual testing, keeping checked
- [x] I have updated documentation (if needed) // no need to update docs, keeping checked

### DIFF
--- a/src/client/cli/cmd/animated_spinner.cpp
+++ b/src/client/cli/cmd/animated_spinner.cpp
@@ -16,6 +16,7 @@
 
 #include "animated_spinner.h"
 
+#include <cassert>
 #include <iostream>
 
 namespace mp = multipass;
@@ -29,8 +30,8 @@ void clear_line(std::ostream& out)
 }
 } // namespace
 
-mp::AnimatedSpinner::AnimatedSpinner(std::ostream& cout)
-    : spinner{'|', '/', '-', '\\'}, cout{cout}, running{false}
+mp::AnimatedSpinner::AnimatedSpinner(std::ostream& cout, bool is_live)
+    : spinner{'|', '/', '-', '\\'}, cout{cout}, running{false}, is_live(is_live)
 {
 }
 
@@ -46,15 +47,19 @@ void mp::AnimatedSpinner::start(const std::string& start_message)
     {
         current_message = start_message;
         running = true;
-        clear_line(cout);
-        cout << start_message << "  " << std::flush;
-        t = std::thread(&AnimatedSpinner::draw, this);
+
+        if (is_live)
+        {
+            clear_line(cout);
+            cout << start_message << "  " << std::flush;
+            t = std::thread(&AnimatedSpinner::draw, this);
+        }
     }
 }
 
 void mp::AnimatedSpinner::start()
 {
-    if (!current_message.empty())
+    if (is_live && !current_message.empty())
         start(current_message);
 }
 
@@ -69,7 +74,9 @@ void mp::AnimatedSpinner::stop()
         if (t.joinable())
             t.join();
     }
-    clear_line(cout);
+
+    if (is_live)
+        clear_line(cout);
 }
 
 void mp::AnimatedSpinner::print(std::ostream& stream, const std::string& message)
@@ -84,6 +91,9 @@ void mp::AnimatedSpinner::print(std::ostream& stream, const std::string& message
 void mp::AnimatedSpinner::draw()
 {
     std::unique_lock<decltype(mutex)> lock{mutex};
+
+    assert(is_live && "should only draw when running live");
+
     auto it = spinner.begin();
     while (running)
     {
@@ -92,6 +102,5 @@ void mp::AnimatedSpinner::draw()
         cout << "\b" << *it << std::flush;
         cv.wait_for(lock, std::chrono::milliseconds(100));
     }
-    cout << "\b"
-         << " " << std::flush;
+    cout << "\b" << " " << std::flush;
 }
--- a/src/client/cli/cmd/animated_spinner.cpp
+++ b/src/client/cli/cmd/animated_spinner.cpp
@@ -43,14 +43,17 @@ mp::AnimatedSpinner::~AnimatedSpinner()
 void mp::AnimatedSpinner::start(const std::string& start_message)
 {
     if (!is_live)
+    {
+        current_message = start_message;
+        cout << start_message << std::flush;
         return;
+    }
 
     std::unique_lock<decltype(mutex)> lock{mutex};
     if (!running)
     {
         current_message = start_message;
         running = true;
-
         clear_line(cout);
         cout << start_message << "  " << std::flush;
         t = std::thread(&AnimatedSpinner::draw, this);
@@ -65,7 +68,7 @@ void mp::AnimatedSpinner::start()
 
 void mp::AnimatedSpinner::stop()
 {
-    if(!is_live)
+    if (!is_live)
         return;
 
     std::unique_lock<decltype(mutex)> lock{mutex};
@@ -77,16 +80,11 @@ void mp::AnimatedSpinner::stop()
         if (t.joinable())
             t.join();
     }
-
-    
     clear_line(cout);
 }
 
 void mp::AnimatedSpinner::print(std::ostream& stream, const std::string& message)
 {
-    if (!is_live)
-        return;
-    
     stop();
 
     stream << message;
--- a/src/client/cli/cmd/animated_spinner.cpp
+++ b/src/client/cli/cmd/animated_spinner.cpp
@@ -56,7 +56,7 @@ void mp::AnimatedSpinner::start(const std::string& start_message)
             t = std::thread(&AnimatedSpinner::draw, this);
         }
         else
-            cout << start_message << std::flush;
+            cout << start_message << std::endl;
     }
 }
 
--- a/src/client/cli/cmd/animated_spinner.cpp
+++ b/src/client/cli/cmd/animated_spinner.cpp
@@ -42,18 +42,18 @@ mp::AnimatedSpinner::~AnimatedSpinner()
 
 void mp::AnimatedSpinner::start(const std::string& start_message)
 {
+    if (!is_live)
+        return;
+
     std::unique_lock<decltype(mutex)> lock{mutex};
     if (!running)
     {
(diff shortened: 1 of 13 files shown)


src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -29,8 +29,8 @@ void clear_line(std::ostream& out)
 }
 } // namespace
 
-mp::AnimatedSpinner::AnimatedSpinner(std::ostream& cout)
-    : spinner{'|', '/', '-', '\\'}, cout{cout}, running{false}
+mp::AnimatedSpinner::AnimatedSpinner(std::ostream& cout, bool is_tty)
```
ricab: Can we keep the `is_live` nomenclature throughout, please?

src/client/cli/cmd/animated_spinner.cpp
ricab, line 99: This should only be called when running live. Could we have an assert to signal that here please? Something like


```suggestion
    std::unique_lock<decltype(mutex)> lock{mutex};
    assert(is_tty && "should only draw when running live");
```

And `#include <cassert>` at the top.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -46,15 +46,23 @@ void mp::AnimatedSpinner::start(const std::string& start_message)
...
+
+        if (is_tty)
+        {
+            clear_line(cout);
+            cout << start_message << "  " << std::flush;
+            t = std::thread(&AnimatedSpinner::draw, this);
+        }
+        else
+        {
+            cout << start_message << std::flush;
```
ricab: I wonder if we should print any messages at all. Did you do it like this because of the way `start` is used in the [spinner callbacks](https://github.com/canonical/multipass/blob/eb05c4cc4fb9f583e863b05fb859564c31d62c29/src/client/cli/cmd/common_callbacks.h)?

Let's see what @amylily1011's says too.

ricab (changes requested):
Thanks for the PR! I have a couple of requests and questions for clarification.

I would also like to see manual tests results exercising this in the multiple commands that use the spinner, showing correct handling of tricky cases like errors from the daemon, update prompts, password in auhenticate...

src/client/cli/cmd/animated_spinner.cpp
riken127: Actually, I kept the messages because the original concern seemed to be about the clutter generated by the spinner itself.

From what I gather in the discussion, it seems safe to remove the else branch, which I’ve gone ahead and done.

Please correct me if I’ve misunderstood anything. Thanks again for your patience!

ricab: Indeed, the spinner clutter was the main issue. However (as per discussion in the issue) it also makes sense to skip status messages when running in the background. 

So yeah, I guess skipping the start message makes sense. I don't know if that will break things though, I believe start is used for things beyond mere status updates (see callbacks above).

ricab:
Hey @riken127, is this ready for another pass?

BTW, I know approaches vary a lot on different projects, but we favor separate atomic commits, without squashing, including review requests. See GIT5 and others in [CONTRIBUTING.md](https://github.com/canonical/multipass/blob/main/CONTRIBUTING.md#versioning-git). 

It makes it easier to see new changes. No worries about the one you already pushed, just something to keep in mind going forward.

riken127:
Hi @ricab,

Yes, i believe so!

I've modified the logic and performed additional manual tests. I had to push force on my last commit to fix a minor whitespace oversight and keep the history clean; I'm sorry about that!

I've verified the following cases in non-interactive mode using `| cat`:
1. `launch | cat`
```
$ ./multipass launch | cat
Launched: familiar-chaffinch
```
2. `./multipass authenticate | cat`
```
$ ./multipass authenticate | cat
Please enter passphrase: 
(Waited for input)
```
3. `./multipass launch charlie | cat`
```
$ ./multipass launch charlie | cat
launch failed: Unable to find an image matching "charlie" in remote "".
```

Regarding the callbacks and interaction angles you mentioned, I've made start() and stop() no-ops in non-live TTYs. I'm not sure if this covers all the edge cases you were thinking of, so if I've missed a specific interaction angle, please let me know. I'd be happy to take a closer look if you could provide more detail on what else might be affected.

Thank you for your patience, it took me a while to answer as I had some issues with my linux machine. Would love to hear some feedback!

ricab:
Cool, thanks again for your involvement @riken127! I am queuing this to review soon.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -69,11 +77,16 @@ void mp::AnimatedSpinner::stop()
         if (t.joinable())
             t.join();
     }
+
+
```
ricab:
```suggestion

```

ricab (changes requested):
Hi @riken127, I have circled back here and, after closer inspection, I don't think we can just skip messages altogether. That would prevent logs, for example, since they use `AnimatedSpinner::print` in the corresponding callback. And if we enable any message, better enable them all: it wouldn't make sense to eat some messages but not others depending on what exact method client code called.

Our CLI is undergoing a review right now and I don't think we should introduce radical changes until that process is concluded, so I think the cleanest approach for the time being would be to just print all messages but without the spinner clutter. IOW, when running non-interactively:

    1. `print` would do the exact same that it does today (no early return)
    2. `start` would print only the message (as you had it before)
    3. `stop` would return early, as you have it.

I believe that is the sane compromise at this point.

Additionally, it would be good to cover this case in unit tests.

Sorry about the moving target there. Hopefully you need only revert a couple of snippets. Let us know your thoughts and thanks again for your contribution!

riken127:
Hi @ricab,
I’ve added the unit tests you requested and aligned the behavior with your latest feedback.
Could you please take another look when you have a moment? Thank you.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -41,6 +42,13 @@ mp::AnimatedSpinner::~AnimatedSpinner()
 
 void mp::AnimatedSpinner::start(const std::string& start_message)
 {
+    if (!is_live)
+    {
+        current_message = start_message;
+        cout << start_message << std::flush;
+        return;
+    }
+
```
ricab: Although this doesn't create a thread in this case, it could still be called from different threads. Might as well keep it locked.

src/client/cli/cmd/animated_spinner.cpp
ricab, line 50: And you could keep the state machine the same. I don't see a reason to change the behavior in that regard.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -41,6 +42,13 @@ mp::AnimatedSpinner::~AnimatedSpinner()
 
 void mp::AnimatedSpinner::start(const std::string& start_message)
 {
+    if (!is_live)
+    {
+        current_message = start_message;
+        cout << start_message << std::flush;
```
ricab: You would need a newline to avoid gluing messages together, right? An endl would still flush and keep the log timely.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -54,12 +62,15 @@ void mp::AnimatedSpinner::start(const std::string& start_message)
 
 void mp::AnimatedSpinner::start()
 {
-    if (!current_message.empty())
+    if (is_live && !current_message.empty())
```
ricab, line 65: FWIW I realize this should be locked too (because of current_message), but it is not in the scope of this PR (and would need recursive locking or some other trick).

ricab (commented):
Hi again @riken127, I finally got back to this. I still have a couple of concerns. 

I played with it a little as I reviewed and I will push my reshuffling. Please let me know what you think.

riken127:
Hey @ricab,

Sorry for taking so long to get back to you on this.

Thanks for pushing the reshuffle. I went through the changes and they make sense to me. Keeping the existing lock/state-machine behaviour, while only limiting the live terminal effects, feels like the right approach here.

The newline in the non-live path also addresses the glued-message concern nicely.

I’m happy with this direction. Thanks again for taking the time to look into it.

PR #4711, opened 2026-03-04
[ci] Skip repo maintenance actions on forks

jimporter:
# Description

These changes stop Multipass forks from running unnecessary CI/CD workflows that never succeed on the fork, and don't really make sense to run in that context anyway. You can see a bunch of these on my fork, which I've been using as a way of ensuring that outside contributors have a decent experience: https://github.com/jimporter/multipass/actions

## Checklist

### DIFF
--- a/.github/workflows/prune-packages.yml
+++ b/.github/workflows/prune-packages.yml
@@ -11,6 +11,8 @@ permissions:
 
 jobs:
   list-packages:
+    if: ${{ github.repository == 'canonical/multipass' }}
+
     runs-on: ubuntu-latest
 
     outputs:
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -11,6 +11,8 @@ permissions:
 
 jobs:
   build:
+    if: ${{ github.repository == 'canonical/multipass' }}
+
     runs-on: self-hosted-linux-amd64-noble-large-tiobe
 
     env:


.github/workflows/close-inactive-issues.yml
```diff
@@ -8,6 +8,8 @@ permissions:
 
 jobs:
   close-issues:
+    if: ${{ github.repository == 'canonical/multipass' }}
```
sharder996: This one from the looks of it works fine on forks. Its entirely possible that a fork of a project can become active enough to have its own disjoint issues. I understand disabling things that are going to fail, but changing the behaviour of other workflows, I'm not so keen about.

.github/workflows/update-images.yaml
```diff
@@ -12,6 +12,8 @@ permissions:
 
 jobs:
   run-scraper:
+    if: ${{ github.repository == 'canonical/multipass' }}
```
sharder996: idem

If we're changing the behaviour of workflows, I feel like we could disable nightly builds as well.

.github/workflows/close-inactive-issues.yml
jimporter: I'm fine with reverting this part for now. My reasoning was to optimize for users forking the repo to file PRs, as opposed to forks that eventually diverge from us and produce their own releases. However, the main goal was really just to stop GitHub from sending emails about failed CI/CD workflows that we didn't specifically intend to work on a fork.

.github/workflows/update-images.yaml
jimporter: Disabling nightly workflows might also be a good idea, since it saves electricity, and for free users on GH, keeps Multipass from causing backlogs in CI job queues for the user's account (unless you pay, you only get 20 parallel jobs).

We can discuss this with the team as a followup though.

jimporter:
@sharder996 I'll add an entry into our meeting topics list to discuss how the team wants to deal with the other CI workflows.

Issue #2560, opened 2022-05-08
[Apple M1] IPv6 doesn't route to internet


Vayu:
**Describe the bug**
VM gets assigned IPv6 and can reach the host via it but not the Internet. IPv4 works fine connecting to internet. Host itself has IPv6 connectivity.

Cheeked on 2 different machines (m1 and m1 pro) with the latest macos

**To Reproduce**
How, and what happened?
1. `multipass shell
1. ping6 www.google.con

**Expected behavior**
Ipv6 routes to internet like ipv4

**Additional info**
- `multipass version` 1.8.1+mac
- `multipass info --all'
Name:           media
State:          Running
IPv4:           192.168.64.2
               10.9.10.103
Release:        Ubuntu 20.04.4 LTS
Image hash:     add2f33bf439 (Ubuntu 20.04 LTS)
Load:           0.13 0.05 0.01
Disk usage:     2.2G out of 4.7G
Memory usage:   146.0M out of 970.0M
Mounts:         --

Vayu:
Looks like traffic from inside VM gets NATed into ULA address of the host instead of GUA. E.g. when I send pings from the host, I see:
```
IP6 (flowlabel 0x90e00, hlim 64, next-header ICMPv6 (58) payload length: 16) 2603:....... > 2607:f8b0:4002:c08::63: [icmp6 sum ok] ICMP6, echo 
```
But when I do it from within the VM
```
IP6 (flowlabel 0xd41ac, hlim 63, next-header ICMPv6 (58) payload length: 64) fd16:....... > 2607:f8b0:4002:c08::63: [icmp6 sum ok] ICMP6, echo request, 
```
And the router doesn't route fd16 ULA address to the Internet.
Not sure where to see NAT settings that control this behavior

ricab:
There is also #3832 for this, which has a little more info. Let's continue tracking there.

PR #4715, opened 2026-03-06
[log] Print a more informative log message

ricab:
# Description

Print a more informative log message with a tip on how to revert bridge creation. This is still only a log, but at least it is better than nothing.

Also update docs with the same recipe.

## Related Issue(s)

Addresses #2560

## Testing

Setup:
  1. Install the snap
  2. `multipass set local.bridged-network=<appropriate-network>`

Test 1:
  1. `multipass launch bridged`, answer yes
  2. verify the bridge is created (e.g. with `ip link`)
  3. Observe the deletion command in the log, copy it.
  4. Execute the command from the log, verify the bridge is gone, e.g. with `ip l`
 
Test 2:
  1. `multipass launch -n test`
  2. `multipass stop test`
  3. `multipass set local.test.bridged=true` and answer yes
  5. Find the message in the log, copy the command
  6. Execute the command and verify the bridge is gone

Docs:
  - verify documentation [renders correctly](https://canonical-ubuntu-documentation-library--4715.com.readthedocs.build/multipass/4715/how-to-guides/manage-instances/create-an-instance/#bridging)

## Checklist

### DIFF
--- a/docs/how-to-guides/manage-instances/create-an-instance.md
+++ b/docs/how-to-guides/manage-instances/create-an-instance.md
@@ -213,7 +213,8 @@ use NetworkManager's command-line interface to remove the bridge:
 `nmcli connection delete br-eth0 br-eth0-child`.
 Use `nmcli connection show` to obtain the actual connection names.
 
-On installations that do not have `NetworkManager` installed (e.g. Ubuntu Server), the user can
+On installations that do not have `NetworkManager` installed (for example, Ubuntu Server), the user
+can
 still create a bridge by other means and pass that to Multipass. For instance, this configuration
 snippet achieves that with Netplan:
 


src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -220,6 +220,8 @@ std::string mp::Backend::create_bridge_with(const std::string& interface)
 {
     static constexpr auto log_category_create = "create bridge";
     static constexpr auto log_category_rollback = "rollback bridge";
+    static constexpr auto success_msg = "Created bridge: {}. "
+                                        "To revert, run: `nmcli connection delete {} {}`";
```
jimporter: What are your thoughts on doing this to avoid the repeated `format` arg?

```suggestion
    static constexpr auto success_msg = "Created bridge: {0}. "
                                        "To revert, run: `nmcli connection delete {0} {1}`";
```

Also, would it be better to put this definition closer to the `mpl::info` call that uses it? Since this is a placeholder string, it's useful to be able to see the placeholder and its arguments on the same screen.

src/platform/backends/shared/linux/backend_utils.cpp
```diff
@@ -265,7 +267,7 @@ std::string mp::Backend::create_bridge_with(const std::string& interface)
     rollback_guard.dismiss(); // we succeeded!
 
     auto ret = parent_name.toStdString();
-    mpl::info(log_category_create, "Created bridge: {}", ret);
+    mpl::info(log_category_create, success_msg, ret, ret, child_name);
```
jimporter: And then...

```suggestion
    mpl::info(log_category_create, success_msg, ret, child_name);
```

src/platform/backends/shared/linux/backend_utils.cpp
ricab: > What are your thoughts on doing this to avoid the repeated format arg?

Sure, good idea!

> Also, would it be better to put this definition closer to the mpl::info call that uses it?

OK. I was trying to tuck all those static constexprs at the beginning of the function, but I don't mind your proposal.

src/platform/backends/shared/linux/backend_utils.cpp
ricab: Yep

src/platform/backends/shared/linux/backend_utils.cpp
ricab: I ended up just inlining the string into the log call. WDYT?

jimporter (dismissed):
LGTM!

docs/how-to-guides/manage-instances/create-an-instance.md
```diff
@@ -209,7 +209,13 @@ This will temporarily disrupt connectivity on that interface.
 Do you want to continue (yes/no)?
 ```
 
-However, Multipass requires `NetworkManager` to achieve this. On installations that do not have `NetworkManager` installed (e.g. Ubuntu Server), the user can still create a bridge by other means and pass that to Multipass. For instance, this configuration snippet achieves that with Netplan:
+Multipass requires `NetworkManager` to achieve this, whose command-line interface can also be used
```
geoffreynyaga: A minor suggestion to break down the sentence into two. The word `whose` could refer to either Multipass or NetworkManager. It is also not clear if `NetworkManager` is a Multipass feature or an external tool (in which case we need to link to it).
``` diff
+ Multipass requires `[NetworkManager](https://)` to achieve this. You can also use `NetworkManager's` command-line interface to remove the bridge: `nmcli connection delete br-eth0 br-eth0-child`.
```
We can also reword the sentence to use "You" for consistency purposes (as done above). This is because we have mixed active and passive voice between the tool and the user. see https://docs.ubuntu.com/styleguide/en/index#active-and-passive-voice.

docs/how-to-guides/manage-instances/create-an-instance.md
```diff
@@ -209,7 +209,13 @@ This will temporarily disrupt connectivity on that interface.
...
+to remove the bridge: `nmcli connection delete br-eth0 br-eth0-child`. Use `nmcli connection show`
+to obtain the actual connection names.
+
+On installations that do not have `NetworkManager` installed (e.g. Ubuntu Server), the user can
```
geoffreynyaga: A small fix:
Use of latin abbreviations is discouraged as they are less approachable to international audiences (as per the style guide) https://docs.ubuntu.com/styleguide/en/index#latin-words-and-phrases
```suggestion
On installations that do not have `NetworkManager` installed (for example, Ubuntu Server), the user can
```

docs/how-to-guides/manage-instances/create-an-instance.md
ricab: Done.

docs/how-to-guides/manage-instances/create-an-instance.md
ricab: Done.

ricab:
Great review @geoffreynyaga, thanks! Fixes pushed

PR #4716, opened 2026-03-06
Fix memory leaks identified by ASan

jimporter:
# Description

This PR fixes several memory leaks identified by address sanitization when running the unit tests. Some of these were just in test code, but a few were in production code as well, especially with SFTP.

Build with address sanitization and run relevant tests:
```
CPPFLAGS="-fsanitize=address" LDFLAGS="-fsanitize=address -static-libasan" cmake .. -DCMAKE_BUILD_TYPE=Debug
cmake --build ..
./bin/multipass_tests"
```
Before this, the tests fail with ASan errors, but after, they pass.

## Checklist

### DIFF
--- a/include/multipass/ssh/sftp_utils.h
+++ b/include/multipass/ssh/sftp_utils.h
@@ -45,6 +45,7 @@ struct SFTPError : public std::runtime_error
 
 MP_SFTP_UNIQUE_PTR(sftp_new, sftp_free)
 MP_SFTP_UNIQUE_PTR(sftp_open, sftp_close)
+MP_SFTP_UNIQUE_PTR(sftp_limits, sftp_limits_free)
 MP_SFTP_UNIQUE_PTR(sftp_stat, sftp_attributes_free)
 MP_SFTP_UNIQUE_PTR(sftp_lstat, sftp_attributes_free)
 MP_SFTP_UNIQUE_PTR(sftp_opendir, sftp_closedir)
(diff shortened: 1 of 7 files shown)


jimporter:
@tobe2098 For review purposes, each commit has an explanation with details about what the actual issue was.

tobe2098:
Thank you Jim. It may be out of the scope for this PR, but if we do not want to mock new and free for sftp maybe we should remove the mocks altogether if possible.

jimporter:
@tobe2098 Hmm, another option here might be to mock `sftp_server_new` *and* `sftp_server_free`, since that's what we do with the client. I guess it depends on whether we prefer mocking everything SFTP-related, or if we'd rather mock only the subset of things that wouldn't work normally...

tobe2098:
That is a better idea. Since in some cases I have seen some mocks do `calloc` and `free` maybe the allocation functions should not be mocked as long as allocation and initialization is all they do. But that is definitely outside the scope of this (very necessary) PR and would require further discussion.

That said, is this PR still in progress?

jimporter:
@tobe2098 Ok, I've updated the test code to use the mock implementations more consistently, rather than just removing the mocks. This is review-ready again.

tobe2098 (approved):
LGTM, great work Jim!

tobe2098:
Briefly tested the SFTP + SSL code to see that all works as it should, adding to the merge queue

PR #4719, opened 2026-03-09
[lint] Upgrade to clang-format-19

xmkg:
clang-format-diff-18 seems to emit only the first file in the diff when there are multiple, which makes it really cumbersome to iterate over the lint failures.

~~This patch temporarily introduces clang-format-diff-19 as a workaround. The script still uses clang-format-18 for producing the diff, so the format output should be the same.~~

This patch upgrades the clang-format to 19 by reading the snapcraft.yml's CLANG_VERSION variable and installing the specific clang-format version.

# Description

- What does this PR do?
Fixes a CI lint step bug.
- Why is this change needed?
It's really hard to iterate over lint bugs as the tool emits one diff at a time.

### DIFF
--- a/.github/actions/read-snapcraft-yml-params/action.yml
+++ b/.github/actions/read-snapcraft-yml-params/action.yml
@@ -1,0 +1,24 @@
+name: Read snapcraft.yml parameters
+description: Extracts parameters from snap/snapcraft.yaml for CI comsumption
+
+outputs:
+  clang-version:
+    description: The CLANG_VERSION value
+    value: ${{ steps.extract.outputs.clang_version }}
+
+runs:
+  using: composite
+  steps:
+    - name: Install yq
+      shell: bash
+      run: go install github.com/mikefarah/yq/v4@latest
+
+    - name: Extract CLANG_VERSION
+      id: extract
+      shell: bash
+      run: |
+        # Extract CLANG_VERSION from snapcraft.yaml (single source of truth)
+        CLANG_VERSION=$(yq '.parts.multipass.build-environment[] | select(has("CLANG_VERSION")) | .CLANG_VERSION' snap/snapcraft.yaml | tr -d '"')
+        [ -z "$CLANG_VERSION" ] && { echo "::error::CLANG_VERSION not found in snapcraft.yaml"; exit 1; }
+        echo "clang_version=$CLANG_VERSION" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -61,7 +69,7 @@ jobs:
           # On pull requests, HEAD^1 will always be the merge base, so consider that diff for formatting.
           git diff -U0 --no-color HEAD^1 \
             | grep -vE '^\+\+\+ b/tests/unit/test_data/formatters/' \
```
jimporter: While we're here, I think we could also remove this line by adding this to `tests/unit/test_data/formatters/.clang-format`:

```yaml
DisableFormat: true
```

Though if you think that's out of scope for this PR, I can file a followup afterwards.

.github/actions/read-snapcraft-yml-params/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Install yq
+      shell: bash
+      run: go install github.com/mikefarah/yq/v4@latest
+
+    - name: Extract CLANG_VERSION
+      id: extract
+      shell: bash
+      run: |
+        # Extract CLANG_VERSION from snapcraft.yaml (single source of truth)
+        CLANG_VERSION=$(yq '.parts.multipass.build-environment[] | select(has("CLANG_VERSION")) | .CLANG_VERSION' snap/snapcraft.yaml | tr -d '"')
```
jimporter: `-r`/`--raw-output` would let you remove the extra `tr` step:
```suggestion
        CLANG_VERSION=$(yq -r '.parts.multipass.build-environment[] | select(has("CLANG_VERSION")) | .CLANG_VERSION' snap/snapcraft.yaml)
```
You could also add `-e`/`--exit-status` here to remove the line after this one, but then you wouldn't get a friendly error message. I'm ok with either option for that part.

jimporter (changes requested):
Good catch, this should fix things. Just a small suggestion to simplify the `yq` invocation, and an optional improvement to the pipeline where we actually run `clang-format`.

.github/workflows/lint.yml
xmkg: Good idea. Done.

.github/actions/read-snapcraft-yml-params/action.yml
xmkg: Removed `tr -d '"'` in favor of `yq -r`. I'd say let's keep the nice error message for now :)

jimporter (approved):

PR #4720, opened 2026-03-10
[fmt] Format macos CLI args following qemu-linux-detail

tobe2098:
Change the formatting in qemu_platform_detail_macos.cpp to be consistent with CLI arg formatting, which had been lost in a previous change of the file due to the change in clang version.

This formatting for CLI args is better for diffs and is worth maintaining manually.
It is also necessary because the new formatting fails the CI pipeline in the release branch due to clang version formatting mismatch.


ricab (approved):
LGTM 👍

PR #4722, opened 2026-03-11
Hyper-V API: Snapshots feature

xmkg:
Implement forward-merge on erase behavior for the virtdisk snapshot feature.

### DIFF
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -272,8 +272,8 @@ bool HCSVirtualMachine::has_saved_state_file() const
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
     const std::filesystem::path base_vhdx = description.image.image_path;
-    const std::filesystem::path head_avhdx =
-        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    const std::filesystem::path head_avhdx = base_vhdx.parent_path() /
+                                             virtdisk::VirtDiskSnapshot::head_disk_name();
     return MP_FILEOPS.exists(head_avhdx) ? head_avhdx : base_vhdx;
 }
 
@@ -725,7 +725,9 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdWString(), *this, description);
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdWString(),
+                                                        *this,
+                                                        description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -669,10 +669,7 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                                   "snapshots. To resize, delete the snapshots first."};
     }
 
-    // No snapshots remain, but a differencing "head" disk may still be layered on the
-    // base if the collapse at the last snapshot's deletion did not complete. Fold it
-    // back into the base first so we never resize a base that still has a child disk
-    // on top of it (which would corrupt the chain).
+    // A leftover head means collapse did not finish. Collapse it before resizing.
     if (const auto head_avhdx = get_snapshot_head_disk_path())
     {
         mpl::warn(get_name(),
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -725,7 +725,7 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdString(), *this, description);
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdWString(), *this, description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -271,10 +271,17 @@ bool HCSVirtualMachine::has_saved_state_file() const
 
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
+    return get_snapshot_head_disk_path().value_or(description.image.image_path);
+}
+
+std::optional<std::filesystem::path> HCSVirtualMachine::get_snapshot_head_disk_path() const
+{
     const std::filesystem::path base_vhdx = description.image.image_path;
-    const std::filesystem::path head_avhdx = base_vhdx.parent_path() /
-                                             virtdisk::VirtDiskSnapshot::head_disk_name();
-    return MP_FILEOPS.exists(head_avhdx) ? head_avhdx : base_vhdx;
+    std::filesystem::path head_avhdx =
+        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    if (MP_FILEOPS.exists(head_avhdx))
+        return head_avhdx;
+    return std::nullopt;
 }
 
 void HCSVirtualMachine::grant_access_to_scsi_device(const hcs::HcsScsiDevice& device) const
@@ -660,6 +667,19 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
     {
         throw ResizeDiskException{"Cannot resize the primary disk while there are "
                                   "snapshots. To resize, delete the snapshots first."};
+    }
+
+    // No snapshots remain, but a differencing "head" disk may still be layered on the
+    // base if the collapse at the last snapshot's deletion did not complete. Fold it
+    // back into the base first so we never resize a base that still has a child disk
+    // on top of it (which would corrupt the chain).
+    if (const auto head_avhdx = get_snapshot_head_disk_path())
+    {
+        mpl::warn(get_name(),
+                  "A snapshot head disk `{}` is still layered on the base even though no "
+                  "snapshots remain; collapsing it into the base before resizing.",
+                  *head_avhdx);
+        virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
     }
 
     if (const auto result =
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -679,7 +679,26 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                   "A snapshot head disk `{}` is still layered on the base even though no "
                   "snapshots remain; collapsing it into the base before resizing.",
                   *head_avhdx);
-        virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
+        try
+        {
+            virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
+        }
+        catch (const std::exception& e)
+        {
+            throw ResizeDiskException{
+                "Cannot resize the primary disk because the snapshot head disk `{}` could not be "
+                "collapsed into the base: {}",
+                *head_avhdx,
+                e.what()};
+        }
+
+        if (const auto remaining_head = get_snapshot_head_disk_path())
+        {
+            throw ResizeDiskException{
+                "Cannot resize the primary disk because snapshot head disk `{}` still exists after "
+                "collapse. Delete or repair the head disk before resizing.",
+                *remaining_head};
+        }
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -725,7 +725,7 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename, *this, description);
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdString(), *this, description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -23,6 +23,7 @@
 #include <hyperv_api/hcs/hyperv_hcs_event_type.h>
 #include <hyperv_api/hcs/hyperv_hcs_wrapper.h>
 #include <hyperv_api/hcs_virtual_machine_exceptions.h>
+#include <hyperv_api/virtdisk/virtdisk_snapshot.h>
 #include <hyperv_api/virtdisk/virtdisk_wrapper.h>
 
 #include <shared/windows/smb_mount_handler.h>
@@ -270,7 +271,10 @@ bool HCSVirtualMachine::has_saved_state_file() const
 
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
-    return description.image.image_path;
+    const std::filesystem::path base_vhdx = description.image.image_path;
+    const std::filesystem::path head_avhdx =
+        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    return MP_FILEOPS.exists(head_avhdx) ? head_avhdx : base_vhdx;
 }
 
 void HCSVirtualMachine::grant_access_to_scsi_device(const hcs::HcsScsiDevice& device) const
@@ -652,6 +656,12 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
 {
     mpl::debug(get_name(), "resize_disk() -> new_size `{}` MiB", new_size.in_megabytes());
 
+    if (get_num_snapshots() > 0)
+    {
+        throw ResizeDiskException{"Cannot resize the primary disk while there are "
+                                  "snapshots. To resize, delete the snapshots first."};
+    }
+
     if (const auto result =
             VirtDisk().resize_virtual_disk(description.image.image_path, new_size.in_bytes());
         !result)
@@ -704,12 +714,18 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
     const VMSpecs& specs,
     std::shared_ptr<Snapshot> parent)
 {
-    throw NotImplementedOnThisBackendException{"snapshot"};
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(snapshot_name,
+                                                        comment,
+                                                        instance_id,
+                                                        parent,
+                                                        specs,
+                                                        *this,
+                                                        description);
 }
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    throw NotImplementedOnThisBackendException{"snapshot"};
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename, *this, description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -678,7 +678,8 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                   *head_avhdx);
         try
         {
-            virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
+            virtdisk::VirtDiskSnapshot::collapse_head_into_base(get_name(),
+                                                                description.image.image_path);
         }
         catch (const std::exception& e)
         {
@@ -687,14 +688,6 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                 "collapsed into the base: {}",
                 *head_avhdx,
                 e.what()};
-        }
-
-        if (const auto remaining_head = get_snapshot_head_disk_path())
-        {
-            throw ResizeDiskException{
-                "Cannot resize the primary disk because snapshot head disk `{}` still exists after "
-                "collapse. Delete or repair the head disk before resizing.",
-                *remaining_head};
         }
     }
 
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -672,23 +672,10 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
     // A leftover head means collapse did not finish. Collapse it before resizing.
     if (const auto head_avhdx = get_snapshot_head_disk_path())
     {
-        mpl::warn(get_name(),
-                  "A snapshot head disk `{}` is still layered on the base even though no "
-                  "snapshots remain; collapsing it into the base before resizing.",
-                  *head_avhdx);
-        try
-        {
-            virtdisk::VirtDiskSnapshot::collapse_head_into_base(get_name(),
-                                                                description.image.image_path);
-        }
-        catch (const std::exception& e)
-        {
-            throw ResizeDiskException{
-                "Cannot resize the primary disk because the snapshot head disk `{}` could not be "
-                "collapsed into the base: {}",
-                *head_avhdx,
-                e.what()};
-        }
+        throw ResizeDiskException{
+            "Cannot resize the primary disk because the snapshot head disk `{}` is present even "
+            "though there are no snapshots.",
+            *head_avhdx};
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -669,7 +669,7 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                                   "snapshots. To resize, delete the snapshots first."};
     }
 
-    // A leftover head means collapse did not finish. Collapse it before resizing.
+    // A leftover head means collapse did not finish.
     if (const auto head_avhdx = get_snapshot_head_disk_path())
     {
         throw ResizeDiskException{
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -277,8 +277,8 @@ std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 std::optional<std::filesystem::path> HCSVirtualMachine::get_snapshot_head_disk_path() const
 {
     const std::filesystem::path base_vhdx = description.image.image_path;
-    std::filesystem::path head_avhdx =
-        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    std::filesystem::path head_avhdx = base_vhdx.parent_path() /
+                                       virtdisk::VirtDiskSnapshot::head_disk_name();
     if (MP_FILEOPS.exists(head_avhdx))
         return head_avhdx;
     return std::nullopt;
@@ -367,8 +367,8 @@ bool HCSVirtualMachine::maybe_create_compute_system()
 {
     // Always reset the handle and create a new one.
     hcs_system.reset();
-    auto attach_callback_handler =
-        sg::make_scope_guard([this]() noexcept { set_compute_system_callback_handler(); });
+    auto attach_callback_handler = sg::make_scope_guard(
+        [this]() noexcept { set_compute_system_callback_handler(); });
 
     if (const auto result = HCS().open_compute_system(get_name(), hcs_system))
     {
@@ -399,12 +399,12 @@ bool HCSVirtualMachine::maybe_create_compute_system()
                           .read_only = true}},
         .network_adapters =
             [&] {
-                const auto view =
-                    endpoints |
-                    std::views::transform([](const auto& endpoint) -> hcs::HcsNetworkAdapter {
-                        return {.endpoint_guid = endpoint.endpoint_guid,
-                                .mac_address = endpoint.mac_address.value()};
-                    });
+                const auto view = endpoints |
+                                  std::views::transform(
+                                      [](const auto& endpoint) -> hcs::HcsNetworkAdapter {
+                                          return {.endpoint_guid = endpoint.endpoint_guid,
+                                                  .mac_address = endpoint.mac_address.value()};
+                                      });
                 return std::vector(std::ranges::begin(view), std::ranges::end(view));
             }(),
         .guest_state = {.guest_state_file_path = get_guest_state_file_path(),
@@ -413,8 +413,8 @@ bool HCSVirtualMachine::maybe_create_compute_system()
                                                   ? std::optional(get_saved_state_file_path())
                                                   : std::nullopt}};
 
-    if (const auto create_result =
-            HCS().create_compute_system(create_compute_system_params, hcs_system);
+    if (const auto create_result = HCS().create_compute_system(create_compute_system_params,
+                                                               hcs_system);
         !create_result)
     {
         throw CreateComputeSystemException{"create_compute_system failed with {}",
@@ -682,8 +682,8 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
         virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
     }
 
-    if (const auto result =
-            VirtDisk().resize_virtual_disk(description.image.image_path, new_size.in_bytes());
+    if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
+                                                           new_size.in_bytes());
         !result)
     {
         throw ResizeDiskException{"Disk resize failed, details: {}", result};
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -271,17 +271,7 @@ bool HCSVirtualMachine::has_saved_state_file() const
 
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
-    return get_snapshot_head_disk_path().value_or(description.image.image_path);
-}
-
-std::optional<std::filesystem::path> HCSVirtualMachine::get_snapshot_head_disk_path() const
-{
-    const std::filesystem::path base_vhdx = description.image.image_path;
-    std::filesystem::path head_avhdx = base_vhdx.parent_path() /
-                                       virtdisk::VirtDiskSnapshot::head_disk_name();
-    if (MP_FILEOPS.exists(head_avhdx))
-        return head_avhdx;
-    return std::nullopt;
+    return description.image.image_path;
 }
 
 void HCSVirtualMachine::grant_access_to_scsi_device(const hcs::HcsScsiDevice& device) const
@@ -667,15 +657,6 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
     {
         throw ResizeDiskException{"Cannot resize the primary disk while there are "
                                   "snapshots. To resize, delete the snapshots first."};
-    }
-
-    // A leftover head means collapse did not finish.
-    if (const auto head_avhdx = get_snapshot_head_disk_path())
-    {
-        throw ResizeDiskException{
-            "Cannot resize the primary disk because the snapshot head disk `{}` is present even "
-            "though there are no snapshots.",
-            *head_avhdx};
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -678,23 +678,10 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
     // A leftover head means collapse did not finish. Collapse it before resizing.
     if (const auto head_avhdx = get_snapshot_head_disk_path())
     {
-        mpl::warn(get_name(),
-                  "A snapshot head disk `{}` is still layered on the base even though no "
-                  "snapshots remain; collapsing it into the base before resizing.",
-                  *head_avhdx);
-        try
-        {
-            virtdisk::VirtDiskSnapshot::collapse_head_into_base(get_name(),
-                                                                description.image.image_path);
-        }
-        catch (const std::exception& e)
-        {
-            throw ResizeDiskException{
-                "Cannot resize the primary disk because the snapshot head disk `{}` could not be "
-                "collapsed into the base: {}",
-                *head_avhdx,
-                e.what()};
-        }
+        throw ResizeDiskException{
+            "Cannot resize the primary disk because the snapshot head disk `{}` is present even "
+            "though there are no snapshots.",
+            *head_avhdx};
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -23,6 +23,7 @@
 #include <hyperv_api/hcs/hyperv_hcs_event_type.h>
 #include <hyperv_api/hcs/hyperv_hcs_wrapper.h>
 #include <hyperv_api/hcs_virtual_machine_exceptions.h>
+#include <hyperv_api/virtdisk/virtdisk_snapshot.h>
 #include <hyperv_api/virtdisk/virtdisk_wrapper.h>
 
 #include <shared/windows/smb_mount_handler.h>
@@ -270,7 +271,10 @@ bool HCSVirtualMachine::has_saved_state_file() const
 
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
-    return description.image.image_path;
+    const std::filesystem::path base_vhdx = description.image.image_path;
+    const std::filesystem::path head_avhdx =
+        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    return MP_FILEOPS.exists(head_avhdx) ? head_avhdx : base_vhdx;
 }
 
 void HCSVirtualMachine::grant_access_to_scsi_device(const hcs::HcsScsiDevice& device) const
@@ -658,6 +662,12 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
 {
     mpl::debug(get_name(), "resize_disk() -> new_size `{}` MiB", new_size.in_megabytes());
 
+    if (get_num_snapshots() > 0)
+    {
+        throw ResizeDiskException{"Cannot resize the primary disk while there are "
+                                  "snapshots. To resize, delete the snapshots first."};
+    }
+
     if (const auto result =
             VirtDisk().resize_virtual_disk(description.image.image_path, new_size.in_bytes());
         !result)
@@ -710,12 +720,18 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
     const VMSpecs& specs,
     std::shared_ptr<Snapshot> parent)
 {
-    throw NotImplementedOnThisBackendException{"snapshot"};
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(snapshot_name,
+                                                        comment,
+                                                        instance_id,
+                                                        parent,
+                                                        specs,
+                                                        *this,
+                                                        description);
 }
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    throw NotImplementedOnThisBackendException{"snapshot"};
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename, *this, description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -685,7 +685,26 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
                   "A snapshot head disk `{}` is still layered on the base even though no "
                   "snapshots remain; collapsing it into the base before resizing.",
                   *head_avhdx);
-        virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
+        try
+        {
+            virtdisk::VirtDiskSnapshot::collapse_head_into_base(description.image.image_path);
+        }
+        catch (const std::exception& e)
+        {
+            throw ResizeDiskException{
+                "Cannot resize the primary disk because the snapshot head disk `{}` could not be "
+                "collapsed into the base: {}",
+                *head_avhdx,
+                e.what()};
+        }
+
+        if (const auto remaining_head = get_snapshot_head_disk_path())
+        {
+            throw ResizeDiskException{
+                "Cannot resize the primary disk because snapshot head disk `{}` still exists after "
+                "collapse. Delete or repair the head disk before resizing.",
+                *remaining_head};
+        }
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -271,17 +271,7 @@ bool HCSVirtualMachine::has_saved_state_file() const
 
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
-    return get_snapshot_head_disk_path().value_or(description.image.image_path);
-}
-
-std::optional<std::filesystem::path> HCSVirtualMachine::get_snapshot_head_disk_path() const
-{
-    const std::filesystem::path base_vhdx = description.image.image_path;
-    std::filesystem::path head_avhdx = base_vhdx.parent_path() /
-                                       virtdisk::VirtDiskSnapshot::head_disk_name();
-    if (MP_FILEOPS.exists(head_avhdx))
-        return head_avhdx;
-    return std::nullopt;
+    return description.image.image_path;
 }
 
 void HCSVirtualMachine::grant_access_to_scsi_device(const hcs::HcsScsiDevice& device) const
@@ -673,15 +663,6 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
     {
         throw ResizeDiskException{"Cannot resize the primary disk while there are "
                                   "snapshots. To resize, delete the snapshots first."};
-    }
-
-    // A leftover head means collapse did not finish.
-    if (const auto head_avhdx = get_snapshot_head_disk_path())
-    {
-        throw ResizeDiskException{
-            "Cannot resize the primary disk because the snapshot head disk `{}` is present even "
-            "though there are no snapshots.",
-            *head_avhdx};
     }
 
     if (const auto result = VirtDisk().resize_virtual_disk(description.image.image_path,
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -272,8 +272,8 @@ bool HCSVirtualMachine::has_saved_state_file() const
 std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 {
     const std::filesystem::path base_vhdx = description.image.image_path;
-    const std::filesystem::path head_avhdx =
-        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    const std::filesystem::path head_avhdx = base_vhdx.parent_path() /
+                                             virtdisk::VirtDiskSnapshot::head_disk_name();
     return MP_FILEOPS.exists(head_avhdx) ? head_avhdx : base_vhdx;
 }
 
@@ -731,7 +731,9 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(
 
 std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QString& filename)
 {
-    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdWString(), *this, description);
+    return std::make_shared<virtdisk::VirtDiskSnapshot>(filename.toStdWString(),
+                                                        *this,
+                                                        description);
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -277,8 +277,8 @@ std::filesystem::path HCSVirtualMachine::get_primary_disk_path() const
 std::optional<std::filesystem::path> HCSVirtualMachine::get_snapshot_head_disk_path() const
 {
     const std::filesystem::path base_vhdx = description.image.image_path;
-    std::filesystem::path head_avhdx =
-        base_vhdx.parent_path() / virtdisk::VirtDiskSnapshot::head_disk_name();
+    std::filesystem::path head_avhdx = base_vhdx.parent_path() /
+                                       virtdisk::VirtDiskSnapshot::head_disk_name();
     if (MP_FILEOPS.exists(head_avhdx))
         return head_avhdx;
     return std::nullopt;
@@ -367,8 +367,8 @@ bool HCSVirtualMachine::maybe_create_compute_system()
 {
     // Always reset the handle and create a new one.
     hcs_system.reset();
-    auto attach_callback_handler =
-        sg::make_scope_guard([this]() noexcept { set_compute_system_callback_handler(); });
+    auto attach_callback_handler = sg::make_scope_guard(
+        [this]() noexcept { set_compute_system_callback_handler(); });
 
     if (const auto result = HCS().open_compute_system(get_name(), hcs_system))
     {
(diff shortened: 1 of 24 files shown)


src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -652,6 +656,12 @@ void HCSVirtualMachine::resize_disk(const MemorySize& new_size)
 {
     mpl::debug(get_name(), "resize_disk() -> new_size `{}` MiB", new_size.in_megabytes());
 
+    if (get_num_snapshots() > 0)
+    {
+        throw ResizeDiskException{"Cannot resize the primary disk while there are "
+                                  "snapshots. To resize, delete the snapshots first."};
+    }
```
ricab, line 668: Although this did not pose problems before, I understand the prudence of following Microsoft recommendations. Since this is a new restriction, should it be covered in our docs?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.h
```diff
@@ -0,0 +1,88 @@
...
+    VirtDiskSnapshot(const QString& filename,
+                     VirtualMachine& vm,
+                     const VirtualMachineDescription& desc);
+    [[nodiscard]] static std::string make_snapshot_filename(const Snapshot& ss);
+    [[nodiscard]] std::filesystem::path make_snapshot_path(const Snapshot& ss) const;
+    [[nodiscard]] static constexpr std::string_view head_disk_name() noexcept
+    {
+        return "head.avhdx";
+    }
+    [[nodiscard]] std::filesystem::path make_head_disk_path() const;
```
ricab: I understand `head_disk_name()`, but why are the other methods here public? IIANM, `make_snapshot_path` is only used internally, while the other two are used only internally and in tests. 

I am not a big fan of widening API surfaces only for testing. I'd recommend *testing behavior, not implementation*. Tests are all about redundancy, so it doesn't shock me if we need to compute paths and filenames independently.

ricab (changes requested):
Hi @xmkg, I am finally getting back to this 🙂 

Submitting a preliminary review with only a couple of comments for now. I still have to review the bulk of the code, but the structure looks good and I've mostly situated myself again. However, it looks like the build is failing now. Would you mind fixing before I dig deeper? Thanks!

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.h
xmkg: Yeah, right. The reason they're public is not intentional -- I'll make them private.

xmkg:
> Hi @xmkg, I am finally getting back to this 🙂
> 
> Submitting a preliminary review with only a couple of comments for now. I still have to review the bulk of the code, but the structure looks good and I've mostly situated myself again. However, it looks like the build is failing now. Would you mind fixing before I dig deeper? Thanks!

Sure thing! I've fixed the build.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,250 @@
...
+                                   VirtualMachine& vm,
+                                   const VirtualMachineDescription& desc)
+    : BaseSnapshot(filename, vm, desc), base_vhdx_path{desc.image.image_path}, vm{vm}
+{
+}
+
+std::string VirtDiskSnapshot::make_snapshot_filename(const Snapshot& ss)
+{
+    return fmt::format("{}.avhdx", ss.get_name());
+}
```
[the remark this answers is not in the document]
xmkg: Fixed.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+        // Restore the snapshot file
+        MP_FILEOPS.remove(self_path);
+        MP_FILEOPS.rename(self_temp_path, self_path);
+    };
+
+    for (const auto& [child_path, snapshot_ptr] : children)
+    {
+        MP_FILEOPS.copy(self_temp_path, self_path, {});
+
```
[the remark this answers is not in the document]
xmkg: False positive.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: I think so. Also, I can look into making this safe if it is a widely needed feature.

ricab: OK, thanks!

ricab:
Hey @xmkg, would you mind requesting me to review again once this is again ready? Thank you!

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,334 @@
...
+            }
+            MP_FILEOPS.remove(staged_path, ec); // drop the staged .new if still present
+        }
+        MP_FILEOPS.remove(self_path, ec);   // drop leftover merge scratch
+        try_rename(self_backup, self_path); // restore pristine self
+    });
+
+    // Pass 1: merge each child into a fresh copy of self and stash the result as
+    // <child>.new. No original files are modified.
+    for (const auto& [child_path, snapshot_ptr] : children)
```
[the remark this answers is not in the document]
xmkg: How come is this a high-priority issue... It's not even an "issue," let alone a high-priority one.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,433 @@
...
+    new_head_path.replace_extension(".new.avhdx"); // "head.avhdx" -> "head.new.avhdx"
+
+    std::error_code ec{};
+    MP_FILEOPS.remove(new_head_path, ec); // clear any leftover from a prior failed apply
+    create_new_child_disk(snapshot_path, new_head_path); // throws here => old head untouched
+
+    // Swap the freshly built head in for the old one. rename() replaces the
+    // destination atomically, so there is never a moment with no head.
+    MP_FILEOPS.rename(new_head_path, head_path);
+    mpl::debug(log_category, "apply_impl() -> new head from {} -> {}", snapshot_path, head_path);
```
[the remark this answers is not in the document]
xmkg: This is utterly invalid, and this bot won't stop posting it over and over again..

xmkg:
@ricab, ready for review again

tests/unit/hyperv_api/test_ut_hyperv_hcs_virtual_machine.cpp
```diff
@@ -661,6 +667,34 @@ TEST_F(HyperVHCSVirtualMachine_UnitTests, resize_disk)
...
 // ---------------------------------------------------------
 
+TEST_F(HyperVHCSVirtualMachine_UnitTests, resize_disk_fails_if_snapshot_head_remains)
+{
+    default_create_success();
+
+    EXPECT_CALL(mock_hcs, get_compute_system_state(Eq(mock_handle), _))
+        .WillRepeatedly(
+            DoAll([](const hcs_handle_t&,
+                     hcs_system_state_t& state) { state = hcs_system_state_t::stopped; },
```
ricab: How about SetArgReferee<N>(value)?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+#include <multipass/file_ops.h>
+#include <multipass/logging/log.h>
+#include <multipass/virtual_machine.h>
+#include <multipass/virtual_machine_description.h>
+
+#include <scope_guard.hpp>
+
+namespace
+{
+constexpr auto log_category = "virtdisk-snapshot";
```
ricab: In other backends, the name of the VM is used as the category for logging. Could this follow that approach too please? Unless there is a reason not too?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+static void try_rename(const std::filesystem::path& from, const std::filesystem::path& to) noexcept
+{
+    try
+    {
+        MP_FILEOPS.rename(from, to);
+    }
+    catch (const std::exception& e)
+    {
+        mpl::error(log_category,
+                   "Failed to restore `{}` -> `{}` during rollback: {}",
```
ricab: Did you mean "Failed to rename"?

And wouldn't a top-catch-all be enough here? You could then reuse it in capture_impl.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+namespace multipass::hyperv::virtdisk
+{
+
+struct CreateVirtdiskSnapshotError : FormattedExceptionBase<std::system_error>
+{
+    using FormattedExceptionBase::FormattedExceptionBase;
+};
+
+// Best-effort rename for noexcept rollback guards.
+static void try_rename(const std::filesystem::path& from, const std::filesystem::path& to) noexcept
```
ricab: Why static? Shouldn't this just go in the unnamed namespace like other things?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+                   "Failed to restore `{}` -> `{}` during rollback: {}",
+                   from,
+                   to,
+                   e.what());
+    }
+}
+
+namespace
+{
+// Keep backup/scratch files beside their disk.
```
ricab: Is there a reason for this to be nested in another namespace? The usual practice in the project is to have a single unnamed namespace first, at the top.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+namespace
+{
+constexpr auto log_category = "virtdisk-snapshot";
+namespace mpl = multipass::logging;
+} // namespace
+
+namespace multipass::hyperv::virtdisk
+{
+
+struct CreateVirtdiskSnapshotError : FormattedExceptionBase<std::system_error>
```
ricab: If this is meant to be used outside (for catching) I believe it should go in a header. If it is not meant to be used that way (for the time being), then better place it in the unnamed namespace.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+{
+    p += suffix;
+    return p;
+}
+
+bool is_direct_child_of(const std::filesystem::path& disk,
+                        const std::filesystem::path& parent_disk,
+                        bool required)
+{
+    if (!MP_FILEOPS.exists(disk))
```
ricab: Might as well check if the parent disk exists, while you're at it?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+        return false;
+
+    std::vector<std::filesystem::path> chain;
+    if (const auto r = VirtDisk().list_virtual_disk_chain(disk, chain, 2); !r)
+    {
+        if (required)
```
ricab: Do we really need this distinction? Couldn't we simplify things and throw every time? `collapse_head_into_base` would then fail in this case, which to me feels like the right thing to do.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+    catch (const std::exception& e)
+    {
+        mpl::warn(log_category,
+                  "Could not collapse head into base after erasing the last snapshot: {}",
+                  e.what());
+    }
+}
+
+void VirtDiskSnapshot::collapse_head_into_base(const std::filesystem::path& base_vhdx_path)
```
ricab: I am a little confused by the error handling in this function. No-op in some cases, throw in others, and sometimes only a log, despite the operation keeping the separate head we asked to collapse.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    std::vector<std::filesystem::path> chain;
+    if (const auto r = VirtDisk().list_virtual_disk_chain(disk, chain, 2); !r)
+    {
+        if (required)
+            throw CreateVirtdiskSnapshotError{r,
+                                              "Could not inspect virtual disk chain for `{}`",
+                                              disk};
+
+        mpl::warn(log_category, "Could not inspect optional disk chain for `{}`: {}", disk, r);
+        return false;
```
ricab: If we failed to read the chain for some reason, this could be wrong: the child could really be a direct child, but we would be reporting otherwise.

If the function returns a bool but also throws, better just separate the waters clearly and keep throwing for error reporting and the bool for the direct-child relationship.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+bool is_direct_child_of(const std::filesystem::path& disk,
+                        const std::filesystem::path& parent_disk,
+                        bool required)
+{
+    if (!MP_FILEOPS.exists(disk))
+        return false;
+
+    std::vector<std::filesystem::path> chain;
+    if (const auto r = VirtDisk().list_virtual_disk_chain(disk, chain, 2); !r)
```
ricab: BTW, I am a little concerned that there is no else to the if in virtdisk_wrapper.cpp:597...

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    if (MP_FILEOPS.exists(child))
+        throw CreateVirtdiskSnapshotError{std::make_error_code(std::errc::file_exists),
+                                          "Child disk `{}` already exists",
+                                          child};
+
+    const virtdisk::CreateVirtualDiskParameters params{
+        .path = child,
+        .predecessor = virtdisk::ParentPathParameters{parent}};
+
+    if (const auto result = VirtDisk().create_virtual_disk(params); !result)
```
ricab, line 128: There are a few functions in Virtdisk.h that refer specifically to snapshots. For example:

- https://learn.microsoft.com/en-us/windows/win32/api/virtdisk/nf-virtdisk-takesnapshotvhdset
- https://learn.microsoft.com/en-us/windows/win32/api/virtdisk/ns-virtdisk-apply_snapshot_vhdset_parameters

Are they unsuitable?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+    if (head_preexisted)
+    {
+        MP_FILEOPS.rename(head_path, snapshot_path);
+        head_became_snapshot = true;
+    }
+    else
+    {
+        const auto parent = get_parent();
+        const auto target = parent ? make_snapshot_path(*parent) : base_vhdx_path;
```
ricab: How could we get to a state with a parent but no head? Couldn't we implement this to prevent ever creating that situation and then just throw if it came about (e.g. because the user messed with internal files)?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+        create_new_child_disk(target, snapshot_path);
```
ricab: If we did have a parent but no head, wouldn't this just create an empty snapshot?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+}
+
+std::vector<std::filesystem::path> VirtDiskSnapshot::get_disk_children(
+    const std::filesystem::path& parent_disk) const
+{
+    std::vector<std::filesystem::path> result;
+
+    for (const auto& elem : vm.view_snapshots([this_index = get_index()](const Snapshot& ss) {
+             return ss.get_index() != this_index;
+         }))
```
ricab: Would you mind extracting this `view_snapshots` call to a named variable, for readability please? And you could just compare the pointer with this, no need to call get_index, right?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    {
+        auto path = make_snapshot_path(*elem);
+        if (is_direct_child_of(path, parent_disk, true))
+            result.push_back(std::move(path));
```
ricab: All snapshots know their parents, so you should be able to skip virtdisk API interrogation and the disk I/O:


```suggestion
        if (elem->get_parent() == this)
            result.push_back(make_snapshot_path(*elem));
```

But then you might as well just fold this into the view_snapshots predicate.

An assert for the disk-based `is_direct_child_of` would still be valuable, to be sure the two layers are in sync.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    try
+    {
+        collapse_head_into_base(base_vhdx_path);
+    }
+    catch (const std::exception& e)
+    {
+        mpl::warn(log_category,
+                  "Could not collapse head into base after erasing the last snapshot: {}",
+                  e.what());
+    }
```
ricab: Why catch this? The way I see it, failing in this step means failing the erase. Otherwise, we're leaving things in a half state.

Erasing a snapshot should leave us in the same state we would be in if we'd never taken it. In particular, erasing the last snapshot should leave us with a single disk.

You could dismiss the rollback in `erase_impl` only after this succeeds, although you'd probably need to rollback this collapse too. 

I believe the difficulty arises because your implementation has this particularity of the head existing only once you have 1 snapshot. You didn't have to do it that way. You could consider the case of 0 snapshots as having a single disk corresponding to the head. The head is just the layer that is being written at present. (If you want to look at it through the lens of layer diffs, with 0 snapshots the head would have the diff to a null layer, like a diff to an empty disk.) You would then have a single base case to deal with instead of two. Collapsing head into base would be just another child-parent merge.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+    const auto head_path = make_head_disk_path();
+    const auto snapshot_path = make_snapshot_path(*this);
+
+    // Build the replacement beside the live head so creation failure keeps the VM bootable.
+    auto new_head_path = head_path;
+    new_head_path.replace_extension(".new.avhdx"); // "head.avhdx" -> "head.new.avhdx"
+
+    std::error_code ec{};
+    MP_FILEOPS.remove(new_head_path, ec);
+    create_new_child_disk(snapshot_path, new_head_path);
```
ricab: This is another case where the code deals with possible leftovers of half-states. Why don't we just fail if we can't remove the file?

Then, if on the next occasion we found the file existed, that could only mean the user messed with internals or we had a bug (which we'd want to detect, not support/mask). 

Discarding current changes is probably the main reason to restore a snapshot. And those changes could be heavy.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -652,8 +663,43 @@ void HCSVirtualMachine::resize_disk_impl(const MemorySize& new_size)
...
-    if (const auto result =
-            VirtDisk().resize_virtual_disk(description.image.image_path, new_size.in_bytes());
+    if (get_num_snapshots() > 0)
+    {
+        throw ResizeDiskException{"Cannot resize the primary disk while there are "
+                                  "snapshots. To resize, delete the snapshots first."};
+    }
+
+    // A leftover head means collapse did not finish. Collapse it before resizing.
+    if (const auto head_avhdx = get_snapshot_head_disk_path())
```
ricab: IIUC, this complexity arises only because we adopted incomplete deletion as a valid situation. We could drop it if only we declared no leftovers as part of the post-conditions of erase_impl.

If the operation failed in the first place, how likely is it that it would succeed now? And if it was a transient thing that really might just work on retry, why not leave the onus of retrying to the user?

Robustness is about being able to handle errors, but that may just mean failing gracefully and early. It does not mean absorbing erroneous situations as part of valid program state.

ricab (changes requested):
Thanks for your efforts here @xmkg. This is the hardest backend to implement snapshots since you need to deal with the elementary disk layers that other backends just delegate (but see my question about the Virtdisk API). You did an impressive job here. This is much closer than the earlier attempt I saw.

That said, there are still a number of concerns I have. Details inline, but the mainly around:

- error handling
- half-states
- distinct no-head case

I haven't tried this manually and I have yet to review most of the tests.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Sadly, they are unsuitable because they're designed for VHDS disks. VHDS disks are typically used for shared data between Hyper-V VMs. For regular VHDX disks, the differencing-disk approach is still the way to go.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: The head disk is created lazily when the first snapshot is taken, so if there are no snapshots yet, there will be no head disk. (Also, the head disk collapses into the primary disk when there aren't any snapshots left)

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Technically, yes. The snapshot file wouldn't have any delta w.r.t. the parent.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: The check is against the parent disk path, though. get_disk_children is called for both this disk and its child disks. I've simplified it a bit; let me know what you think.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: I agree with not catching the exception. For the last part, I've done it this way to avoid additional I/O cost to the VMs that have no snapshots. Differencing disk layering is not free, and I thought it would be better to follow the "pay for what you use" principle here. I agree your approach would be simpler. What do you think?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Changed the remove call to the throwing variant instead.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: I guess we can just throw an exception here, but then if this scenario happens, there's no way to invoke collapse_head_into_base behavior externally, so the steps have to be taken manually, e.g., via PowerShell. I don't see this happening normally anyway, hence making a troubleshooting guide involving the recovery steps might be a better choice.

xmkg: Changed it to throw instead.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Good catch -- added an else statement returning the error code.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+            if (auto merge_r = VirtDisk().merge_virtual_disk_into_parent(child_path); !merge_r)
+                throw CreateVirtdiskSnapshotError{merge_r,
+                                                  "Could not merge child disk `{}` into `{}`",
+                                                  child_path,
+                                                  self_path};
+
+            MP_FILEOPS.rename(self_path, staged_path);
+            staged.emplace_back(child_path, staged_path);
+        }
```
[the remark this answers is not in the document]
xmkg: Known tradeoff.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,441 @@
...
+}
+
+void VirtDiskSnapshot::collapse_head_after_last_erase()
+{
+    // Self is still counted here, so the last snapshot means exactly one remains.
+    if (vm.get_num_snapshots() != 1)
+        return;
+
+    collapse_head_into_base(base_vhdx_path);
+}
```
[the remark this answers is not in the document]
xmkg: Intentional.

xmkg:
Hey @ricab, I've gone through your comments and addressed/answered them. LMK once you have the next batch :)

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,282 @@
...
+    const bool head_preexisted = MP_FILEOPS.exists(head_path);
+    bool head_became_snapshot = false;
+
+    // Leave no partial files after a failed capture.
+    auto rollback = sg::make_scope_guard([&]() noexcept {
+        std::error_code ec{};
+        if (head_became_snapshot)
+        {
+            MP_FILEOPS.remove(head_path, ec);
+            try
```
xmkg: use try_rename here

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,282 @@
...
+#include <multipass/virtual_machine_description.h>
+
+#include <fmt/format.h>
+#include <scope_guard.hpp>
+
+namespace
+{
+namespace mpl = multipass::logging;
+namespace mhv = multipass::hyperv;
+using PathPairs = std::vector<std::pair<std::filesystem::path, std::filesystem::path>>;
```
xmkg: use this alias from child_rebuild

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
ricab: OK, thanks for confirming.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
ricab: Ah understood. This is just the place where that head is created.

ricab: Ah no, but my question was about the parent. How can there be a parent if there are no snapshots?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
ricab: See above. Having a parent but no head still sounds like a contradiction to me.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,282 @@
...
+        {
+            MP_FILEOPS.remove(head_path, ec);
+            try
+            {
+                MP_FILEOPS.rename(snapshot_path, head_path);
+            }
+            catch (const std::exception& e)
+            {
+                mpl::error(vm.get_name(), "capture_impl() rollback failed: {}", e.what());
+            }
```
ricab: You could now replace any such direct renames with `try_rename`.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,282 @@
...
+void VirtDiskSnapshot::collapse_head_into_base(const std::string& vm_name,
+                                               const std::filesystem::path& base_vhdx_path)
+{
+    const auto head_path = base_vhdx_path.parent_path() / head_disk_name();
+
+    if (!MP_FILEOPS.exists(head_path))
+        return;
+
+    // Never merge the head into a snapshot that other snapshots still depend on.
+    if (!is_direct_child_of(head_path, base_vhdx_path))
```
ricab: The comment makes sense, but it is not what the if is checking, is it?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
ricab: It was probably not very clear, but the approach I had in mind would not come with any additional disks or IO. The "head" disk would just be the file that is named after the VM today. In the base case — VM with no snapshots — the head disk would be the VM's single disk (the file would probably still be named after the VM, not just "head"). To take the first snapshot, you would rename the VM's image (the head) to "snapshot1" or something. Same for the second snapshot and so on:

#### VM foo-bar with no snapshots

- foo-bar.vhdx <- head

#### Take a snapshot

- snapshot1.vhdx <- old head
- foo-bar.vhdx  <- new head

#### Take another snapshot

- snapshot1.vhdx
- snapshot2.vhdx <- old head
- foo-bar.vhdx  <- new head

This would avoid a special case for "do we have a head or not".

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
ricab: But there is no reason why removing the file would succeed here but not in a new erase attempt. If we just fail the erase, snapshot deletion will fail and the snapshot will still be around. The user can then just retry deletion explicitly, which will retry the same removal that is now done here.

ricab (commented):
Thanks for the changes and replies @xmkg! Here is a thinner review to follow up on open topics.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,282 @@
...
+VirtDiskSnapshot::VirtDiskSnapshot(const std::filesystem::path& filename,
+                                   VirtualMachine& vm,
+                                   const VirtualMachineDescription& desc)
+    : BaseSnapshot(filename, vm, desc), base_vhdx_path{desc.image.image_path}, vm{vm}
+{
+}
+
+std::string VirtDiskSnapshot::make_snapshot_filename(const Snapshot& ss)
+{
+    return fmt::format("{}.avhdx", ss.get_index());
```
ricab, line 68: BTW, good call using the index. Names change. 👍

xmkg: Credits go to @copilot on this one -- it spotted a good one! :D

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: I see what you mean now. IIUC, you are saying that head_preexisted already implies the parent's existence. That invariant is true in normal operation, unless someone goes and deletes the head manually. In that case, should we approach defensively? Otherwise, the code would create the snapshot from the base disk even if there's a parent.

ricab: Yep, pretty much. Here's how I see this:

> you are saying that head_preexisted already implies the parent's existence.

Yes, and the other way around. No head implies no parent. So head exists <=> parent exists, unless the user messes with internals, or there is a bug (on our end or the APIs we're using).

When users mess with internals, they break contract. The best they can hope for is an exception aborting the operation and leaving everything as it was (strong exception guarantee). That is what the code would attempt already if we didn't check for a parent here. We'd throw in line 137 and try to rollback. Technically a best-effort attempt rather than a guarantee, but we're already operating outside of the contract so anything goes anyway.

If we instead try to operate on the broken state, we're either

- adopting the situation as part of valid domain (so state now covered in the contract), or
- masking and prolonging the erroneous situation, where we'd better fail early.

Finally, if the situation came about because of a bug, we'd better detect it ASAP and consciously decide what to do about it. So we want to be noisy about it, not silent.

xmkg: OK, I've changed the code to throw instead. LMK what you think.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Removal is no longer attempted here -- am I missing something, or did you make that comment for the previous revision? I've noticed that the comment is stale -- I'll update that, too.

src/platform/backends/hyperv_api/virtdisk/virtdisk_wrapper.cpp
```diff
@@ -591,8 +595,10 @@ OperationResult VirtDiskWrapper::list_virtual_disk_chain(const std::filesystem::
             return OperationResult{E_FAIL, L"GetVirtualDiskInformation failed!"};
```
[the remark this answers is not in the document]
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Updated the comment

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Drop unreferenced backups.
+void ChildRebuild::finalize() noexcept
+{
+    std::error_code ec{};
+    for (const auto& [child_path, staged_path] : staged)
+        MP_FILEOPS.remove(with_suffix(child_path, ".old"), ec);
+    MP_FILEOPS.remove(self_backup, ec);
+}
```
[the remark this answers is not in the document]
xmkg: bogus

xmkg:
@ricab I think I've addressed all of your comments. I've also made the change to the head disk structure you've proposed. LMK what you think!

EDIT: All CLI tests are green on W11 Home.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    return result;
+}
+
+void VirtDiskSnapshot::erase_impl()
+{
+    const auto self_path = make_snapshot_path(*this);
+
+    auto children = get_children_of_disk(self_path);
+
```
ricab: If everything is correct, the disk "family tree" should match the snapshot "family tree", modulo the live disk, correct? If so, obtaining the disk of the current snapshot and then the children of that disk should have the same result as obtaining the children of the current snapshot and then get each of their disks. Is that so?

If that is the case, `get_children_of_disk` looks unnecessarily complex: it filters snapshots by their parenting relations, but deriving file paths repeatedly and using disk links to do so. Couldn't we instead just get the snapshots from the info we already have?

If the 1-to-1 mapping I have in mind is indeed expected, than we could still assert it, i.e. for each snapshot we found to be a direct child of the current snapshot we're erasing, assert that their disk is indeed a child of the current snapshot's disk. And similar asserts might be a good addition elsewhere in the code.

Note: `BaseVirtualMachine::head_snapshot` is not currently accessible, but that is something that I think we should change. The snapshot that is currently active (last taken/restored) should be public info actually reported to users when listing snapshots.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
ricab: Yes, I think I commented on an earlier version. Resolving.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.h
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Create a new differencing child disk.
+     *
+     * @param [in] parent Existing parent disk.
+     * @param [in] child New child disk path.
+     */
+    void create_new_child_disk(const std::filesystem::path& parent,
+                               const std::filesystem::path& child) const;
+    void try_rename(const std::filesystem::path& from, const std::filesystem::path& to) noexcept;
```
ricab: Should this be const?

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
```diff
@@ -97,6 +99,36 @@ struct OperationResult
     {
         return static_cast<std::error_code>(code);
     }
+
+    /**
+     * Make a OperationResult from WIN32 system error code
```
ricab:
```suggestion
     * Make an OperationResult from a WIN32 system error code
```

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
ricab: Added a more detailed comment with my thoughts and questions for you to in/validate. Closing this one to avoid old threads.

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
```diff
@@ -97,6 +99,36 @@ struct OperationResult
...
+     *
+     * @param win32_system_errc System error code
```
ricab: Param name out of sync

src/platform/backends/hyperv_api/virtdisk/virtdisk_exceptions.h
```diff
@@ -0,0 +1,29 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+#include <multipass/exceptions/formatted_exception_base.h>
```
ricab, line 20:
```suggestion
#include <multipass/exceptions/formatted_exception_base.h>

#include <system_error>
```

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
```diff
@@ -0,0 +1,140 @@
...
+      log_category{log_category}
+{
+    staged.reserve(this->children.size());
+}
+
+// Move self aside before arming rollback.
+void ChildRebuild::begin()
+{
+    MP_FILEOPS.rename(self_path, self_backup);
+}
```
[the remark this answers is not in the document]
ricab: Wrong: `std::filesystem::rename` overwrites.

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
```diff
@@ -0,0 +1,140 @@
...
+    : self_path{std::move(self_path)},
+      self_backup{with_suffix(this->self_path, ".tmp")},
+      children{std::move(children)},
+      grandchildren{std::move(grandchildren)},
+      log_category{log_category}
+{
+    staged.reserve(this->children.size());
+}
+
+// Move self aside before arming rollback.
```
ricab: Repeating the comments from the header invites divergence. I suggest removing them here.

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+// Drop unreferenced backups.
+void ChildRebuild::finalize() noexcept
+{
+    std::error_code ec{};
+    for (const auto& [child_path, staged_path] : staged)
+        MP_FILEOPS.remove(with_suffix(child_path, ".old"), ec);
+    MP_FILEOPS.remove(self_backup, ec);
```
ricab: Should any errors that arise at least be logged?

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
```diff
@@ -0,0 +1,140 @@
...
+        const auto backup_path = with_suffix(child_path, ".old");
+        if (MP_FILEOPS.exists(backup_path, ec))
+        {
+            MP_FILEOPS.remove(child_path, ec);
+            try_rename(log_category, backup_path, child_path);
+        }
+        MP_FILEOPS.remove(staged_path, ec);
+    }
+    MP_FILEOPS.remove(self_path, ec);
+    try_rename(log_category, self_backup, self_path);
```
ricab: Idem: logging errors? Maybe with a logged_remove helper?

tests/unit/hyperv_api/test_it_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,410 @@
...
+    static constexpr auto vhdx_size = 1024ULL * 1024 * 16;
+
+    NiceMock<MockVirtualMachine> vm;
+    VirtualMachineDescription desc{};
+    std::vector<Snap> snapshots;
+    int counter = 0;
+
+    void SetUp() override
+    {
+        auto bp = QDir(vm.tmp_dir->path()).filePath("base.vhdx");
```
ricab: `bp` is a little cryptic here, I wouldn't mind a clearer name. Also, should this move to "live" instead of "base"?

tests/unit/hyperv_api/test_it_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,410 @@
...
+    {
+        ss->erase();
+        std::erase(snapshots, ss);
+    }
+
+    Path base() const
+    {
+        return desc.image.image_path;
+    }
+    Path snapshot_path(const Snapshot& ss) const
```
ricab, line 120:
```suggestion

    Path snapshot_path(const Snapshot& ss) const
```

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
```diff
@@ -97,6 +99,36 @@ struct OperationResult
...
+     * @param status_msg Prefix message
+     */
+    static OperationResult from_win32(uint32_t win32_result_code, std::wstring status_msg)
```
ricab: How about taking a const ref to the string or a string view?

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
```diff
@@ -97,6 +99,36 @@ struct OperationResult
...
+    }
+
+    static OperationResult success()
+    {
+        return OperationResult{.code = NOERROR, .status_msg = L""};
+    }
+
+    static OperationResult failure(std::wstring reason)
+    {
+        return OperationResult{.code = E_FAIL, .status_msg = reason};
```
ricab: Might as well move the string:

```suggestion
        return OperationResult{.code = E_FAIL, .status_msg = std::move(reason)};
```

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,238 @@
...
+    PathPairs grandchildren;
+    for (const auto& child_path : children)
+        for (auto& grandchild : get_children_of_disk(child_path))
+            grandchildren.emplace_back(std::move(grandchild), child_path);
+
+    ChildRebuild rebuild{self_path, std::move(children), std::move(grandchildren), vm.get_name()};
```
ricab: You could move `self_path` too (after removing const).

src/platform/backends/hyperv_api/virtdisk/virtdisk_utils.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <vector>
+
+namespace multipass::hyperv::virtdisk
+{
+
+void try_rename(std::string_view log_category,
+                const std::filesystem::path& from,
+                const std::filesystem::path& to) noexcept
+{
+    top_catch_all(log_category, [from, to] { MP_FILEOPS.rename(from, to); });
```
ricab: Any reason to capture by value?

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+#include <multipass/exceptions/formatted_exception_base.h>
+#include <multipass/file_ops.h>
+#include <multipass/logging/log.h>
+#include <multipass/top_catch_all.h>
+#include <multipass/virtual_machine.h>
+#include <multipass/virtual_machine_description.h>
+
+#include <algorithm>
+#include <iterator>
```
ricab: I believe a few of these includes are now obsolete (e.g. top_catch_all).

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,667 @@
...
+using hyperv::virtdisk::VirtdiskSnapshotError;
+
+OperationResult op_ok()
+{
+    return OperationResult{S_OK, L""};
+}
+
+OperationResult op_fail()
+{
+    return OperationResult{E_FAIL, L"forced failure"};
```
ricab: Could you the new success/failure here.

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,667 @@
...
+        return std::filesystem::remove(p, ec);
+    });
+    ON_CALL(fops, copy(_, _, _))
+        .WillByDefault([](const fs::path& s, const fs::path& d, fs::copy_options o) {
+            std::filesystem::copy(s, d, o);
+        });
+    ON_CALL(fops, weakly_canonical(_)).WillByDefault([](const fs::path& p) {
+        return std::filesystem::weakly_canonical(p);
+    });
+
```
ricab: Different sets of ON_CALL delegations to the real std::filesystem are repeated in multiple tests. If you pulled the superset onto a separate helper, you could call it in tests and avoid the repetition.

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,667 @@
...
+// and no snapshot children (so view_snapshots is empty). The snapshot is first
+// captured (creating real self + head files on disk and marking it "captured"),
+// then erased. The VirtDisk wrapper is mocked so merges succeed/fail on demand.
+struct VirtDiskSnapshotErase : public VirtDiskSnapshotTest
+{
+    void SetUp() override
+    {
+        VirtDiskSnapshotTest::SetUp();
+
+        // No snapshot children: get_disk_children() iterates the snapshots and finds
```
ricab: It is now called `get_children_of_disk()`. LLMs like to generate this type of comment. It does not pull its weight IMO.

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
```diff
@@ -0,0 +1,667 @@
...
+        // nothing, so the live disk is the only child sitting on self.
+        ON_CALL(vm, view_snapshots(_)).WillByDefault(Return(VirtualMachine::SnapshotVista{}));
+
+        // Report every disk as parented on self, so get_disk_children() sees the live disk
```
ricab: idem: function name obsolete

ricab (commented):
Looking good, big improvements 🙂 I find this new approach much more elegant.

I have mostly small comments and I think this is getting close. Apart from the disk vs snapshot family trees, it should all be straightforward stuff. Perhaps we could go over it in a meeting. No real blockers, but I am putting manual testing off until we settle things.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.h
xmkg: Done

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Done

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
xmkg: Done

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
xmkg: Done

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
xmkg: done

tests/unit/hyperv_api/test_ut_hyperv_virtdisk_snapshot.cpp
xmkg: done

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_exceptions.h
xmkg: Done

tests/unit/hyperv_api/test_it_hyperv_virtdisk_snapshot.cpp
xmkg: Done

tests/unit/hyperv_api/test_it_hyperv_virtdisk_snapshot.cpp
xmkg: Done

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
xmkg: Done

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
xmkg: Done

src/platform/backends/hyperv_api/hyperv_api_operation_result.h
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_utils.cpp
xmkg: Done

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
xmkg: done

src/platform/backends/hyperv_api/virtdisk/virtdisk_child_rebuild.cpp
xmkg: done

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.cpp
xmkg: Okay, I've added validation code to erase_impl() as per our 1:1 live discussion. I've made it a runtime exception so both the virtdisk impl and the base snapshot would roll back the change instead of letting it go out of sync. Let me know what you think.

ricab: Yep, that's it. Thanks!

ricab (changes requested):
Alright, I just spent some 5h putting this through its paces and it works impressively well, apart from a few things, described below. They shouldn't be too hard to fix though, so this is very close to target! 🙂 

I also looked at your latest code changes and they all make sense to me.

#### Pass ✔️ 

- Disk contents with
    * addition and deletion of files, apt upgrades
    * many snapshots, complex inheritance trees
    * restores across different branches
    * deletions, including snapshots with multiple children, lineages of snapshots, leaves, and the root
    * interspersing all of the above
- multiple deletes in one go (e.g. `multipass delete instance.snapshot{1,2,3}`)
- deletion of all remaining snapshots - instance still fine
- parenting relationships
- avhdx disks named consistently with json files
- disks correctly removed on deletion (and merged, as verified above)
- reusing snapshot names (of deleted snapshots)
- editing snapshot names and comments
- simple usage on a couple of different Ubuntu images and Fedora
- expected error conditions correctly caught (except for error deletion while running: not hyper-v specific and already reported, so out of scope)

I did hit a couple of bugs:

#### Bug 1 - mounts

Note: I am not sure if this is strictly a snapshots issue. There cases where the mounts worked correctly when restoring snapshots, but I reproduced the problem multiple times using the recipe below. Still, I wouldn't be surprised if it could be reproduced without snapshots at all. Nor would I be too surprised if the problem was outside Hyper-V.

1. Prepare dirs for mounts
    1. `mkdir tmp tmp2`
    2. `echo bar > tmp/foo`
    3. `echo baz > tmp2/bar`
2. Launch with first mount and snapshot
    1. `multipass launch -n t1 --mount tmp`
    2. `multipass exec -n t1 -- ls tmp` # confirm contents
    3. `multipass stop t1`
    4. `multipass snapshot -n s1 -m 'first mount' t1`
    5. Optionally, inspect mounts in `multipass info [--snapshots]`
3. Replace mounts and snapshot
    1. `multipass umount t1`
    2. `multipass mount tmp2 t1`
    3. `multipass exec -n t1 -- ls tmp2` # confirm contents
    4. `multipass stop t1`
    5. `multipass snapshot -n s2 -m 'different mount' t1`
    6. Optionally, inspect mounts in `multipass info [--snapshots]`
4. Restore first snapshot and observe empty mount
    1. `multipass restore --destructive t1.s1`
    2. `multipass info t1` and observe the correct mount is listed
    3. However: `multipass exec -n t1 -- ls tmp` # Observe empty
    4. `multipass exec -n t1 -- pgrep sshfs` # Observe not running

#### Bug 2

Note: this one is in this PR, but by omission. It prevents changing cpus, memory, and networks from being restored properly. 

The VM keeps a copy of its description — something that deserves a global fix, but not here — but snapshots don't update it. You need something like this I believe: https://github.com/canonical/multipass/blob/90c46c643823c9a11b5e77c6ccac4326189dd7cc/src/platform/backends/qemu/qemu_snapshot.cpp#L126-L129

1. Launch an instance with default resources and snapshot
    1. `multipass launch -n t2`
    2. `multipass stop t2`
    3. `multipass snapshot -n s1 -m 'default resources' t2`
2. Modify instance properties and verify
    1. `multipass set local.t2.cpus=2`
    2. `multipass set local.t2.memory=3G`
    3. `multipass start t2`
    4. `multipass info t2` # observe new values
3. Restore snapshot and verify correct settings
    1. `multipass stop t2`
    2. `multipass restore --destructive t2.s1`
    3. `multipass get local.t2.cpus` # observe old value
    4. `multipass get local.t2.memory` # observe old value
4. Start the instance and observe wrong value
    1. `multipass start t2`
    2. `multipass info t2` # observe wrong values

#### Transient weirdness (stale IP)

Probably not related to snapshots, but I thought I'd share all the same. After a number of snapshots, deletes, and restores, I got some transient weirdness where Multipass appeared to have a stale IP in cache for the VM. Hosts.ics looked correct (although I only looked after the force stop and successful start).

```
PS C:\Users\me> multipass ls
t4                      Running           10.97.1.110      Ubuntu 26.04 LTS    zone2
PS C:\Users\me> multipass stop t4
PS C:\Users\me> multipass delete t4.s3
Snapshots can only be purged (after deletion, they cannot be recovered). Are you sure you want to continue? (Yes/no): yes
PS C:\Users\me> multipass ls --snapshots
Instance   Snapshot     Parent   Comment
t4         s1           --       Fresh
t4         s2           s1       with asdf
t4         s4           s2       Before restoring s1 (+foo)
t4         s2-1         s2       with bar
t4         s2-2         s2       with baz
t4         s2-2-1       s2-2     fdsa and upgrade
t4         snapshot8    s2-2-1   Before restoring s1
t4         s1-1         s1       adir/a
t4         snapshot10   s1-1     Before restoring s3
PS C:\Users\me> multipass restore t4.s2
Do you want to take a snapshot of t4 before discarding its current state? (Yes/no): no
Snapshot restored: t4.s2
PS C:\Users\me> multipass exec -n t4 -- ls -la
Starting t4 | # this would not complete, ended up ctrl+c'ing
PS C:\Users\me> multipass ls
t4                      Running           10.97.1.110      Ubuntu 26.04 LTS    zone2
PS C:\Users\me> multipass shell t4
shell failed: ssh connection failed: 'Timeout connecting to t4.mshome.net'
PS C:\Users\me> multipass stop t4
stop failed: ssh connection failed: 'Timeout connecting to t4.mshome.net'
PS C:\Users\me> multipass stop --force t4
PS C:\Users\me> multipass start t4
PS C:\Users\me> multipass ls
t4                      Running           10.97.1.45       Ubuntu 26.04 LTS    zone2
PS C:\Users\me> multipass exec -n t4 -- ls -la
total 28
drwxr-x--- 4 ubuntu ubuntu 4096 Aug 27 18:33 .
drwxr-xr-x 3 root   root   4096 Aug 27  2026 ..
-rw-r--r-- 1 ubuntu ubuntu  220 Feb 13  2026 .bash_logout
-rw-r--r-- 1 ubuntu ubuntu 3771 Feb 13  2026 .bashrc
drwx------ 2 ubuntu ubuntu 4096 Aug 27  2026 .cache
-rw-r--r-- 1 ubuntu ubuntu  807 Feb 13  2026 .profile
drwx------ 2 ubuntu ubuntu 4096 Aug 27  2026 .ssh
-rw-rw-r-- 1 ubuntu ubuntu    0 Aug 27 18:33 asdf
```

xmkg:
Hi @ricab, thanks for the extensive test! It's good to hear that there aren't many issues :)

> #### Bug 1 - mounts
> Note: I am not sure if this is strictly a snapshots issue. There cases where the mounts worked correctly when restoring snapshots, but I reproduced the problem multiple times using the recipe below. Still, I wouldn't be surprised if it could be reproduced without snapshots at all. Nor would I be too surprised if the problem was outside Hyper-V.

I looked into this, and it turned out to be a legitimate daemon bug, and I have a fix ready for it: #5180


> #### Bug 2
> Note: this one is in this PR, but by omission. It prevents changing cpus, memory, and networks from being restored properly.
> 
> The VM keeps a copy of its description — something that deserves a global fix, but not here — but snapshots don't update it. You need something like this I believe:

Makes sense, will do!

EDIT: Fixed

> #### Transient weirdness (stale IP)
> Probably not related to snapshots, but I thought I'd share all the same. After a number of snapshots, deletes, and restores, I got some transient weirdness where Multipass appeared to have a stale IP in cache for the VM. Hosts.ics looked correct (although I only looked after the force stop and successful start).

Weird... current implementation does not even cache the IP; it queries the hostname on each management_ipv4() call -- definitely something to look out for in testing.

src/platform/backends/hyperv_api/virtdisk/virtdisk_snapshot.h
```diff
@@ -84,6 +84,11 @@ class VirtDiskSnapshot : public BaseSnapshot
      * Owning VM.
      */
     const VirtualMachine& vm;
+
+    /**
+     * Owning VM description.
+     */
```
ricab, line 90: ~~"Owning" doesn't sound right. This is a reference to the real thing, which it must outlive but which lives and is owned elsewhere.~~

EDIT:

OK, I got it now. This would have helped dispel the ambiguity:

```suggestion
    /**
     * Owning VM's description.
     */
```

ricab (approved):
Yep, that's it! Thanks for all this work, it's working great.

PR #4725, opened 2026-03-11
[lxd] Change lxd endpoint

tobe2098:
# Description

- What does this PR do? This PR adapts the API used by Multipass to communicate with LXD to the newer post-LTS API. 
- Why is this change needed? Users with LXD >6 cannot use Multipass with LXD backend.

## Related Issue(s)

Closes reported issue in Mattermost.

## Testing

- Unit tests
- Manual testing steps:

  1. Install this PR's snap
  2. Install LXD snap --channel latest/stable
  3. Change Multipass backend to LXD `multipass set local.driver=lxd`
  4. `multipass launch -n a` creates a VM
  5. All other commands work as expected
  6. Stop the daemon, reinstall LXD from the --channel 5.21/stable or 5.0/stable
  7. Start the daemon
  8. `multipass launch -n a` creates a VM
  9. All other commands work as expected

## Checklist

## Additional Notes

Firewall issues: It seems that port 53 has to be allowed for mpbr0 or lxdbr0 manually in your firewall to have DNS resolution. It appears to happen only on authd based installations + standalone multipass, can be fixed with:
`sudo ufw allow in on mpbr0 to any port 53`
Suggested migration: Still untested how to move the VMs from one endpoint to the other one.


src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -118,7 +118,7 @@ QJsonObject generate_base_vm_config(const multipass::VirtualMachineDescription&
 {
     QJsonObject config{{"limits.cpu", QString::number(desc.num_cores)},
                        {"limits.memory", QString::number(desc.mem_size.in_bytes())},
-                       {"security.secureboot", "false"}};
+                       {"boot.mode", "uefi-nosecureboot"}};
```
ricab: Is this still compatible with older LXD versions?

tobe2098: Sadly it would only be compatible with >5.21, so no. We could remove this option and rely on the default if it works, but otherwise the APIs are not compatible.

tobe2098: This is the only line for which it is the case, at least according to their docs

ricab: Hahh, bummer. Do you know what the default is in versions 4, 5, and 6? 

Users can probably configure the default too, but that is not super concerning for a backend that is going away anyway. I mean, users can probably revert it if they want to buy a couple more months with the LXD backend in Multipass. If they are attached to their LXD default, better move on to Multipass with QEMU or just stick to LXD directly.

The alternative would be to query LXD for the version and use the appropriate JSON line. But that is new development, which is nagging. I guess we could try if it's a small change? WDYT?

tobe2098: I will look into it, should be feasible

tobe2098:
There is still a number of versions that are untested, in which the APIs could be not compatible with this patch. I tested with 5.21/stable, 6/stable, 5.0/stable, and it worked for all of them.

jimporter (commented):
I tested this out locally, and this works fine with `latest/stable` for LXD, but not `5.21/stable`. I get this error when trying `multipass launch` (even after setting up the firewall rule):

```
[2026-03-19T12:15:14.074] [error] [lxd request] Empty reply received for GET operation on unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0?project=multipass
launch failed: Empty reply received for GET operation on unix://multipass/var/snap/lxd/common/lxd/unix.socket@1.0?project=multipass
```

Am I just doing something wrong? This is with the Snaps for Multipass and LXD.

tobe2098:
The firewall rule is for instance networking, that log suggests that the lxd backend is not running at all. I would check the health of `lxd`. In my case I always purge the snap installation, since I think in this particular case, `snap refresh lxd --channel 5.21/stable` if you already had lxd >=6.7, would not work. It seems they changed the DB schema in that version, and downgrading is not possible without a clean install.

I just tested in my machine and it works as expected, although I only tested launch this time.

src/platform/backends/lxd/lxd_virtual_machine.cpp
```diff
@@ -114,11 +114,19 @@ std::optional<mp::IPAddress> get_ip_for(const QString& mac_addr,
...
-QJsonObject generate_base_vm_config(const multipass::VirtualMachineDescription& desc)
+QJsonObject generate_base_vm_config(const multipass::VirtualMachineDescription& desc,
+                                    mp::NetworkAccessManager* manager,
+                                    const QUrl& url)
 {
+    const auto lxd_metadata = mp::lxd_request(manager, "GET", url);
+    const auto version = lxd_metadata["metadata"]["environment"]["server_version"].toString();
+    bool isV6OrNewer = (version >= QString{"6.7"});
+    const auto boot_mode_key = (isV6OrNewer ? "boot.mode" : "security.secureboot");
+    const auto boot_mode_value = (isV6OrNewer ? "uefi-nosecureboot" : "false");
```
jimporter: It might be better to use `QVersionNumber` here so long as we're already using Qt. That should be more robust, e.g. if LXD gets updated to something like `6.10`.

Also, how about naming this variable `useBootMode`? On my first read, I was a bit confused about whether this was new in 6.7 or new in 6.0. (I checked the LXD release notes to be sure and can confirm that it's 6.7.)

```suggestion
    const auto version = QVersionNumber::fromString(lxd_metadata["metadata"]["environment"]["server_version"].toString());
    bool useBootMode = (version >= QVersionNumber(6, 7));
    const auto boot_mode_key = (useBootMode ? "boot.mode" : "security.secureboot");
    const auto boot_mode_value = (useBootMode ? "uefi-nosecureboot" : "false");
```

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
```diff
@@ -232,8 +232,10 @@ void mp::LXDVirtualMachineFactory::hypervisor_health_check()
     }
     catch (const LXDNotFoundException&)
     {
-        QJsonObject network{{"name", multipass_bridge_name},
-                            {"description", "Network bridge for Multipass"}};
+        QJsonObject network{
+            {"name", multipass_bridge_name},
+            //{"config", QJsonObject{{"ipv4.firewall", "false"}, {"ipv6.firewall", "false"}}},
```
jimporter: Should this line be removed?

jimporter (changes requested):
Ok, after a long misadventure in trying to repair my LXD setup locally, I've confirmed that this works on LXD 5.21 and 6.7. Just a couple small comments about the code changes, but otherwise it looks good.

src/platform/backends/lxd/lxd_virtual_machine.cpp
tobe2098: Thank you Jim. I think the suggestions are good. Good catch!

src/platform/backends/lxd/lxd_virtual_machine_factory.cpp
tobe2098: Oh sorry, yes. Some of the networking troubleshooting lines were left over.

jimporter (approved):
LGTM!

PR #4726, opened 2026-03-11
[vm-image-vault] Check for empty json before parse

xmkg:
boost::json::parse throws an exception if the input is empty so guard against that.


jimporter (approved):
Looks good to me! I think this is the obvious way to handle empty files, while still being noisy about invalid JSON.

PR #4730, opened 2026-03-13
[snapshot] Use Boost.JSON

jimporter:
# Description

This is a bit larger than the other Boost.JSON PRs. This time, we're updating the snapshot code. I've broken this into three parts:
1. Add a new `SnapshotDescription` class. This is important because `BaseSnapshot` is non-copyable, and so it's not possible to (de)serialize that class directly. This has another benefit though, since now it's more obvious which pieces from the snapshot actually get serialized.
2. Add all the usual Boost.JSON code. Most of this is similar to the previous PRs for Boost.JSON, though it's worth noting that I've added an `upgraded` flag to `SnapshotDescription`. That way, our deserializer can be responsible for detecting when we've upgraded the JSON schema, which `BaseSnapshot` can consult to decide whether to immediately persist the upgraded JSON to disk again.
3. Remove some now-unused JSON util code.

## Testing

Unit tests updated to match all the Boost.JSON-ification.

## Checklist

### DIFF
--- a/include/multipass/snapshot_description.h
+++ b/include/multipass/snapshot_description.h
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+#include "memory_size.h"
+#include "virtual_machine.h"
+#include "vm_mount.h"
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include <QDateTime>
+#include <QString>
+
+namespace multipass
+{
+struct SnapshotDescription
+{
+    SnapshotDescription(std::string name,
+                        std::string comment,
+                        int parent_index,
+                        std::string cloud_init_instance_id,
+                        int index,
+                        QDateTime creation_timestamp,
+                        int num_cores,
+                        MemorySize mem_size,
+                        MemorySize disk_space,
+                        std::vector<NetworkInterface> extra_interfaces,
+                        VirtualMachine::State state,
+                        std::unordered_map<std::string, VMMount> mounts,
+                        boost::json::object metadata);
+
+    std::string name;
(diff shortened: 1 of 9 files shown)


include/multipass/snapshot_description.h
```diff
@@ -0,0 +1,86 @@
...
+                        boost::json::object metadata,
+                        bool upgraded = false);
+
+    std::string name;
+    std::string comment;
+    int parent_index;
+
+    // Having these const simplifies thread safety (see `BaseSnapshot`).
+    // NOLINTBEGIN(cppcoreguidelines-avoid-const-or-ref-data-members)
+    const std::string cloud_init_instance_id;
```
tobe2098, line 64: Is there a reason to have this field duplicated here without removing it from BaseSnapshot?

src/platform/backends/shared/base_snapshot.cpp
```diff
@@ -52,222 +47,117 @@ QString derive_index_string(int index)
...
 
-// When it does not contain cloud_init_instance_id, it signifies that the legacy snapshot does not
-// have the item and it needs to fill cloud_init_instance_id with the current value. The current
-// value equals to the value at snapshot time because cloud_init_instance_id has been an immutable
-// variable up to this point.
-std::string choose_cloud_init_instance_id(const QJsonObject& json,
-                                          const std::filesystem::path& cloud_init_iso_path)
+mp::BaseSnapshot::BaseSnapshot(SnapshotDescription desc_,
+                               std::shared_ptr<Snapshot> parent_,
+                               VirtualMachine& vm_,
```
tobe2098: Is there a non-const function being called on vm_? I would make it const to avoid changing the other constructor.

src/platform/backends/shared/base_snapshot.h
```diff
@@ -43,7 +43,7 @@ class BaseSnapshot : public Snapshot
                  const std::string& cloud_init_instance_id,
                  std::shared_ptr<Snapshot> parent,
                  const VMSpecs& specs,
-                 const VirtualMachine& vm);
```
tobe2098, line 46: This one should be const as well if the other VirtualMachine can also be const.

tobe2098 (changes requested):
Good job, Jim! Just some small comments.

include/multipass/snapshot_description.h
jimporter: Oops, just a mistake. Fixed.

src/platform/backends/shared/base_snapshot.cpp
jimporter: Ok, done. I'd changed it to non-`const` so that all the constructors were the same (one of the overloads needed to be non-`const`), but I don't feel strongly about it, so I reverted that change.

src/platform/backends/shared/base_snapshot.h
jimporter: Same here.

tobe2098 (approved):
LGTM, great job Jim!

PR #4733, opened 2026-03-16
[qemu-proc-spec] Add `-uuid` to the arguments

### LINKED ISSUE #2586: Can't add qemu options when launch an instance

The vm launched by multipass does not have the file `/sys/class/dmi/id/product_uuid`.  I want add `-uuid $(uuidgen)` to "qemu-system-x86_64". 

Add a option to launch sub command, for example: 

```text
$ multipass launch -uuid xxx
```
**Additional context**

The kubernetes official website documentation says "Kubernetes uses these values to uniquely identify the nodes in the cluster. If these values are not unique to each node, the installation process may [fail](https://github.com/kubernetes/kubeadm/issues/31)" .

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#verify-mac-address

xmkg:
`-uuid` makes the /sys/class/dmi/id/product_uuid present in the VM which is required by some tools.

Closes: #2586

# Description

- Adds `-uuid` parameter generated from the VM name to the QEMU args.
- Some workflows require `/sys/class/dmi/id/product_uuid` to be present on the VM, which is possible via passing the `-uuid` flag to the QEMU process.

## Related Issue(s)

Closes #2586

## Testing

- Manual testing steps:

  1. Build Multipass from `main`, launch a VM in either Linux or macOS with the QEMU backend, and observe that the `/sys/class/dmi/id/product_uuid` is missing on the VMs
  2. Build Multipass from this branch, launch a VM in either Linux or macOS with the QEMU backend, and observe that the `/sys/class/dmi/id/product_uuid` is present.
  3. Reboot the VM from step 2, and observe that the `/sys/class/dmi/id/product_uuid` value does not change.
  4. Boot another VM,  observe that the `/sys/class/dmi/id/product_uuid` value is different from the launched VM in step 2.

## Checklist

## Additional Notes

MULTI-2507

### DIFF
--- a/src/platform/backends/qemu/qemu_vm_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vm_process_spec.cpp
@@ -38,7 +38,6 @@ mp::QemuVMProcessSpec::QemuVMProcessSpec(const mp::VirtualMachineDescription& de
 QStringList mp::QemuVMProcessSpec::arguments() const
 {
     QStringList args;
-
     if (resume_data)
     {
         args = resume_data->arguments;
@@ -63,6 +62,8 @@ QStringList mp::QemuVMProcessSpec::arguments() const
     }
     else
     {
+        // The UUID needs to be unique per VM and must be consistent across boots.
+        const auto vm_uuid = utils::make_uuid(desc.vm_name);
         auto mem_size =
             QString::number(desc.mem_size.in_megabytes()) + 'M'; /* flooring here; format documented
 in `man qemu-system`, under `-m` option; including suffix to avoid relying on default unit */
@@ -78,25 +79,24 @@ in `man qemu-system`, under `-m` option; including suffix to avoid relying on de
              << "-drive"
              << QString("file=%1,if=none,format=qcow2,discard=unmap,id=hda")
                     .arg(desc.image.image_path)
-             << "-device"
-             << "scsi-hd,drive=hda,bus=scsi0.0";
+             << "-device" << "scsi-hd,drive=hda,bus=scsi0.0";
         // Number of cpu cores
         args << "-smp" << QString::number(desc.num_cores);
         // Memory to use for VM
         args << "-m" << mem_size;
         // Control interface
-        args << "-qmp"
-             << "stdio";
+        args << "-qmp" << "stdio";
         // No console
         args << "-chardev"
              // TODO Read and log machine output when verbose
-             << "null,id=char0"
-             << "-serial"
+             << "null,id=char0" << "-serial"
              << "chardev:char0"
              // TODO Add a debugging mode with access to console
              << "-nographic";
         // Cloud-init disk
         args << "-cdrom" << desc.cloud_init_iso;
+        // To make `/sys/class/dmi/id/product_uuid` present
+        args << "-uuid" << vm_uuid;
     }
 
     for (const auto& [_, mount_data] : mount_args)
--- a/src/platform/backends/qemu/qemu_vm_process_spec.cpp
+++ b/src/platform/backends/qemu/qemu_vm_process_spec.cpp
@@ -67,7 +67,7 @@ QStringList mp::QemuVMProcessSpec::arguments() const
         auto mem_size =
             QString::number(desc.mem_size.in_megabytes()) + 'M'; /* flooring here; format documented
 in `man qemu-system`, under `-m` option; including suffix to avoid relying on default unit */
-
+        // clang-format off
         args << platform_args;
         // The VM image itself
         args << "-device"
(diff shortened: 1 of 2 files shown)


src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -78,25 +79,24 @@ in `man qemu-system`, under `-m` option; including suffix to avoid relying on de
              << "-drive"
              << QString("file=%1,if=none,format=qcow2,discard=unmap,id=hda")
                     .arg(desc.image.image_path)
-             << "-device"
-             << "scsi-hd,drive=hda,bus=scsi0.0";
+             << "-device" << "scsi-hd,drive=hda,bus=scsi0.0";
```
ricab: Could we keep the arguments on different lines, please? It makes it easier to add/remove without affecting surrounding things. This came up last week and Antoni actually just put up #4720 for it too.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -63,6 +62,8 @@ QStringList mp::QemuVMProcessSpec::arguments() const
     }
     else
     {
+        // The UUID needs to be unique per VM and must be consistent across boots.
+        const auto vm_uuid = utils::make_uuid(desc.vm_name);
```
ricab, line 66: A few questions regarding effects on clones:

- Since their name changes from the original, clones get a new UUID, correct?
- Is that compatible with pre-existing disk contents in typical scenarios where this is relevant?
- We cause cloud-init to rerun when cloning. Is that typically enough to cascade the UUID wherever relevant?

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -78,25 +79,24 @@ in `man qemu-system`, under `-m` option; including suffix to avoid relying on de
...
-             << "null,id=char0"
-             << "-serial"
+             << "null,id=char0" << "-serial"
              << "chardev:char0"
              // TODO Add a debugging mode with access to console
              << "-nographic";
         // Cloud-init disk
         args << "-cdrom" << desc.cloud_init_iso;
+        // To make `/sys/class/dmi/id/product_uuid` present
+        args << "-uuid" << vm_uuid;
```
ricab, line 102: To implement instance renaming, we'll probably want to keep the UUID stable. We'd probably save it along with other instance metadata, right? Do you foresee any impediment doing that? Just to be sure.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
ricab, line 1: This covers only the QEMU backend. What happens in other backends? Is a name-based UUID available in the new Hyper-V and AppleVZ today? If not, that's probably worth some wider team awareness and synchronizing.

ricab (commented):
Thanks for taking the initiative to fix this @xmkg. I have a few questions inline.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
xmkg: > We'd probably save it along with other instance metadata, right?

That should be sufficient if we update the UUID on clone. I can't think of any potential gotchas on that.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
xmkg: > Since their name changes from the original, clones get a new UUID, correct?

Yes. It should be treated as any other instance-unique field.

> Is that compatible with pre-existing disk contents in typical scenarios where this is relevant?

Yes, it is abstracted from the disk contents. In reality, it's a hardware serial-like info burned into the BIOS by the manufacturer. No disk involvement at all.

> We cause cloud-init to rerun when cloning. Is that typically enough to cascade the UUID wherever relevant?

The `/sys/class/dmi/id/product_uuid` file is immediately available to queries without cloud-init to be re-run.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
xmkg: It should be available for all off-the-shelf hypervisors already -- the existing hyper-v implementation already does that:

```
ubuntu@test-vm-3:~$ sudo cat /sys/class/dmi/id/product_uuid
e190e5e1-6261-40ca-a806-1a197a1d4ca0
```

I don't know the specifics for Apple VZ, but I think there should be a way to set it, as it's a pretty common thing for hypervisors to do. `https://developer.apple.com/documentation/virtualization/vzgenericmachineidentifier` looks promising, maybe @sharder996 could confirm that. I'll check the new Hyper-V implementation to make sure that it's still available.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
xmkg: Done

src/platform/backends/qemu/qemu_vm_process_spec.cpp
ricab: Sounds good.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
ricab: > The /sys/class/dmi/id/product_uuid file is immediately available to queries without cloud-init to be re-run.

What I meant was that consumers presumably cascade this into persistent configuration state, which would become obsolete in a clone and require manual cleanup. The UUID should typically be unique, even though there might be a use-case to keeping it unchanged in a clone, i.e. to fully replicate the original VM and keep anything depending on the specific UUID unchanged. It looks like some VM managers provide options for that. 

In our case, we change the MAC address as well, so this is just another thing that we wouldn't be copying.

ricab (approved):
Thank you Mustafa, looks good!

I tested and I confirm that two VMs with the same name get the same UUID, as expected. That can only occur at different points in space-time (different machine or different time), so that is probably OK. We can always revisit if we get complaints.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
sharder996:
```
ubuntu@foo:~$ sudo cat /sys/class/dmi/id/product_uuid
74933440-686b-7c4c-8257-02e2056220d4
ubuntu@foo:~$ exit
logout
(.venv) scott@Scotts-MacBook-Pro dev % multipass get local.driver
applevz
```

Looks like I got it already without using `VZGenericMachineIdentifier`.

PR #4734, opened 2026-03-16
Migrate xz-embedded to vcpkg

xmkg:
# Description

- What does this PR do?
Move xz-embedded to vcpkg by adding a minimal portfile for it. The portfile is required bc vcpkg has no official port for it.
- Why is this change needed?
It's needed for the "streamlining dependencies" initiative we have.

MULTI-2342


sharder996:
Any idea how hard it would be to upstream this? Just out of curiosity 😄

xmkg:
> Any idea how hard it would be to upstream this? Just out of curiosity 😄

That already crossed my mind, too. It would require a much more elaborate portfile, but we could try.

3rd-party/vcpkg-ports/xz-embedded/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.14)
```
jimporter, line 15: I have no strong opinion one way or the other, but should we put the GPL license header in here? It was present in the previous version of this file over in `3rd-party/xz-decoder/CMakeLists.txt`.

jimporter (dismissed):
This all looks good to me, and I verified that everything still works when building/running this branch locally. Just one optional comment below about license headers, but feel free to merge either way.

3rd-party/vcpkg-ports/xz-embedded/CMakeLists.txt
xmkg: Done

jimporter (approved):
LGTM!

PR #4735, opened 2026-03-16
[platform-win] Add null check to pOldDACL

xmkg:
The new_ACL function assumes that pOldDACL ptr would be non-empty after a successful `GetNamedSecurityInfo` call, but that premise does not exist

Excerpt from `GetNamedSecurityInfo` docs:

> A pointer to a variable that receives a pointer to the DACL in the
> returned security descriptor or NULL if the security descriptor has
> no DACL. The returned pointer is valid only if you set the DACL_SECURITY_INFORMATION
> flag. Also, this parameter can be NULL if you do not need the DACL.

NULL DACL is a valid return value, indicating that the queried security info has no DACL (i.e., no access restrictions).

The patch adds a null check to fix that.

MULTI-2519

### DIFF
--- a/src/platform/platform_win.cpp
+++ b/src/platform/platform_win.cpp
@@ -716,7 +716,7 @@ auto new_ACL(LPSTR path)
         return newACL;
     }
 
-    if(nullptr == pOldDACL)
+    if (nullptr == pOldDACL)
         return newACL;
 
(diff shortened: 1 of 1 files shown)


tobe2098 (approved):
LGTM, great catch Mustafa!

PR #4736, opened 2026-03-16
[new-release-monitor] Replace QString with std

xmkg:
Replace QT QString with std::string for new release monitor code.


tobe2098 (approved):
LGTM, nice job Mustafa!

PR #4739, opened 2026-03-17
[kaizen] CI tweaks

sharder996:
Some tweaks to things that are running on an automated basis:

1. Added a rule for merging conflicts when converting coverage reports for TICS
2. Added retry functionality for when we prune cached vcpkg packages. The growing list of cached packages sometimes causes rate limiting errors.
3. Stopped the distro-scraper from removing existing data when merging its output with an existing file.
4. Edited the fedora scraper to get architecture specific images from different mirrors (they are not all hosted in the same place). Factored out a bit of common code while I was at it.

### DIFF
--- a/tools/distro-scraper/scraper/scrapers/fedora.py
+++ b/tools/distro-scraper/scraper/scrapers/fedora.py
@@ -1,13 +1,15 @@
 import re
 import aiohttp
 import asyncio
+from urllib.parse import urljoin
+from bs4 import BeautifulSoup
 from dateutil import parser
 from ..base import BaseScraper, DEFAULT_TIMEOUT
 from ..models import SUPPORTED_ARCHITECTURES
 
 
-PRIMARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora/linux/releases"
-SECONDARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora-secondary/releases"
+PRIMARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora/linux/releases/"
+SECONDARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora-secondary/releases/"
 
 ARCH_MAP = {
     "arm64": "aarch64",
@@ -30,9 +32,8 @@ class FedoraScraper(BaseScraper):
         """
         Fetch the primary releases directory listing and return the highest numeric version.
         """
-        url = f"{PRIMARY_RELEASES_URL}/"
-        text = await self._fetch_text(session, url)
-        versions = re.findall(r'href="(\d+)/"', text)
+        entries = await self._fetch_dir_listing(session, PRIMARY_RELEASES_URL)
+        versions = [e.rstrip("/") for e in entries if re.fullmatch(r"\d+/", e)]
         if not versions:
             raise RuntimeError("No numeric release versions found in directory listing")
         latest = str(max(int(v) for v in versions))
@@ -46,8 +47,10 @@ class FedoraScraper(BaseScraper):
         Fetch an Apache-style HTML directory listing and return the linked filenames.
         """
         text = await self._fetch_text(session, url)
-        # Match href values that are plain filenames (no path separators or query strings)
-        return re.findall(r'href="([^"/?][^"/]*)"', text)
+        soup = BeautifulSoup(text)
+        entries = (soup.find("pre").find(string="Parent Directory").parent
+                      .find_next_siblings("a"))
+        return [i.text for i in entries]
 
     async def _fetch_checksum_file(
         self, session: aiohttp.ClientSession, url: str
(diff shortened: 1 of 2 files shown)


tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -1,13 +1,21 @@
...
-from scraper.models import SUPPORTED_ARCHITECTURES
-from ..base import BaseScraper
+from ..base import BaseScraper, DEFAULT_TIMEOUT
+from ..models import SUPPORTED_ARCHITECTURES
 
 
-RELEASES_URL = "https://fedoraproject.org/releases.json"
-DEFAULT_TIMEOUT = 10
+PRIMARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora/linux/releases"
+SECONDARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora-secondary/releases"
```
jimporter: If you added trailing slashes to these, then some of the remaining code would (arguably) be simpler:


```suggestion
PRIMARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora/linux/releases/"
SECONDARY_RELEASES_URL = "https://dl.fedoraproject.org/pub/fedora-secondary/releases/"
```

tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -18,151 +26,134 @@ def __init__(self):
...
-            mapped_arch = "power64le"
-
-        if mapped_arch in SUPPORTED_ARCHITECTURES:
-            return mapped_arch
-
-        raise ValueError(f"unsupported arch {arch}")
-
-    def _find_latest_version(self, images: list[dict]) -> str:
+        url = f"{PRIMARY_RELEASES_URL}/"
+        text = await self._fetch_text(session, url)
```
jimporter: ... then you could do this:


```suggestion
        text = await self._fetch_text(session, PRIMARY_RELEASES_URL)
```

This could also use `_fetch_dir_listing` to avoid needing to use a regex to parse the HTML.

tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -18,151 +26,134 @@ def __init__(self):
...
+        self, session: aiohttp.ClientSession, version: str, label: str
+    ) -> tuple[str, dict]:
         """
-        HEAD the URL and return a parsed Last-Modified header if present.
+        Locate, verify, and return image metadata for a single architecture.
         """
-        self.logger.info("Sending HEAD request to %s", url)
+        fedora_arch = ARCH_MAP.get(label, label)
+        base = SECONDARY_RELEASES_URL if label in SECONDARY_ARCHES else PRIMARY_RELEASES_URL
+        images_url = f"{base}/{version}/Cloud/{fedora_arch}/images/"
```
jimporter: ... and finally this (after adding `from urllib.parse import urljoin` near the top):


```suggestion
        images_url = urljoin(base, f"{version}/Cloud/{fedora_arch}/images/")
```

(`urljoin` is a bit picky and you *need* the trailing slash in the `base` value for this to work.)

tools/distro-scraper/scraper/scrapers/fedora.py
```diff
@@ -18,151 +26,134 @@ def __init__(self):
...
+    ) -> list[str]:
         """
-        Find the latest version string among a list of images.
+        Fetch an Apache-style HTML directory listing and return the linked filenames.
         """
-        if not images:
-            raise RuntimeError("No images to determine latest version")
+        text = await self._fetch_text(session, url)
+        # Match href values that are plain filenames (no path separators or query strings)
+        return re.findall(r'href="([^"/?][^"/]*)"', text)
```
jimporter: Rather than relying on regexes here, maybe use a real HTML parser? This would require adding `bs4` as a dependency and then `from bs4 import BeautifulSoup`.

```suggestion
        doc = BeautifulSoup(text)
        # Directory entries are links inside the main <pre> block immediately
        # following the "Parent Directory".
        entries = (doc.find("pre").find(string="Parent Directory").parent
                      .find_next_siblings("a"))
        return [i.text for i in entries]
```

jimporter: This probably then would need some extra filtering for directories vs non-directories (by the callers?). I'm not **100%** sure this is worth the effort, but it would help prevent potential future issues if something changes on the Fedora end.

sharder996:
@jimporter Agree with you on all points. I don't view this code and mission critical or production code so my resiliency standards are lower. FWIW the image source I decided to pull from comes from Fedora's release [tooling](https://forge.fedoraproject.org/releng/tooling/src/commit/f6a2e8e744e08ea7286c81dc4c7e33003197b444/quality-assurance/critpath/critpath.py#L66-L67) which I would expect to be a bit more stable.

PR #4741, opened 2026-03-18
Purge QT from vm image vault

xmkg:
# Description

- What does this PR do?
~~Partially replaces~~Purges QT types used in VM image vault with std types
- Why is this change needed?
The more we converge towards the std, the better.

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -78,9 +78,6 @@ std::string contents_of(const multipass::Path& file_path);
 std::filesystem::path normalize_path(const std::filesystem::path& path);
 std::string normalize_path(const std::string& path);
 QString normalize_path(const QString& path);
-// Converts QString to path, using appropriate string functions depending on the platform
-[[nodiscard]] std::filesystem::path qstr_to_path(const QString& qstr);
-[[nodiscard]] QString path_to_qstr(const std::filesystem::path& path);
 
 // filesystem mount helpers
 void make_target_dir(SSHSession& session,
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -78,6 +78,9 @@ std::string contents_of(const multipass::Path& file_path);
 std::filesystem::path normalize_path(const std::filesystem::path& path);
 std::string normalize_path(const std::string& path);
 QString normalize_path(const QString& path);
+// Converts QString to path, using appropriate string functions depending on the platform
+[[nodiscard]] std::filesystem::path qstr_to_path(const QString& qstr);
+[[nodiscard]] QString path_to_qstr(const std::filesystem::path& path);
 
 // filesystem mount helpers
 void make_target_dir(SSHSession& session,
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -327,16 +327,6 @@ inline bool multipass::utils::iequals(std::string_view lhs, std::string_view rhs
     });
 }
 
-inline bool multipass::utils::istarts_with(std::string_view str, std::string_view prefix)
-{
-    if (prefix.size() > str.size())
-        return false;
-    return std::ranges::equal(
-        str | std::views::take(prefix.size()),
-        prefix,
-        [](unsigned char a, unsigned char b) { return std::tolower(a) == std::tolower(b); });
-}
-
 template <typename OnTimeoutCallable, typename TryAction, typename... Args>
 void multipass::utils::try_action_for(OnTimeoutCallable&& on_timeout,
                                       std::chrono::milliseconds timeout,
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -32,6 +32,7 @@
 #include <filesystem>
 #include <functional>
 #include <future>
+#include <ranges>
 #include <string>
 #include <vector>
 
@@ -133,6 +134,7 @@ Str&& trim(Str&& s, Filter&& filter);
 template <typename Str>
 Str&& trim(Str&& s);
 bool iequals(std::string_view lhs, std::string_view rhs);
+bool istarts_with(std::string_view str, std::string_view prefix);
 std::string& trim_newline(std::string& s);
 std::string escape_for_shell(const std::string& s);
 std::vector<std::string> split(const std::string& string, const std::string& delimiter);
@@ -320,6 +322,16 @@ inline bool multipass::utils::iequals(std::string_view lhs, std::string_view rhs
     return std::ranges::equal(lhs, rhs, [](char c1, char c2) {
         return tolower(c1) == tolower(c2);
     });
+}
+
+inline bool multipass::utils::istarts_with(std::string_view str, std::string_view prefix)
+{
+    if (prefix.size() > str.size())
+        return false;
+    return std::ranges::equal(
+        str | std::views::take(prefix.size()),
+        prefix,
+        [](unsigned char a, unsigned char b) { return std::tolower(a) == std::tolower(b); });
 }
 
 template <typename OnTimeoutCallable, typename TryAction, typename... Args>
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -181,6 +181,25 @@ QString mp::utils::normalize_path(const QString& path)
     return QString::fromStdString(normalize_path(path.toStdString()));
 }
 
+std::filesystem::path mp::utils::qstr_to_path(const QString& qstr)
+{
+#ifdef _WIN32
+    return std::filesystem::path{qstr.toStdWString()};
+#else
+    // On Linux/macOS the native encoding is UTF-8
+    return std::filesystem::path{qstr.toStdString()};
+#endif
+}
+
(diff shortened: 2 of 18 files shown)


src/utils/vm_image_vault_utils.cpp
```diff
@@ -15,95 +15,142 @@
...
+    constexpr std::size_t buf_size = 8192;
+    char buf[buf_size] = {0};
+    do
+    {
+        stream.read(buf, buf_size);
+        if (stream.bad())
+            throw std::runtime_error("Failed to read data from device to hash");
+        if (auto count = stream.gcount(); count > 0)
+        {
+            if (EVP_DigestUpdate(ctx.get(), buf, static_cast<size_t>(count)) != 1)
```
xmkg, line 92: We don't have a mockable singleton for OpenSSL yet.

xmkg: It may be wise to make a centralized wrapper class, like cryptoutils.

src/utils/utils.cpp
```diff
@@ -181,6 +181,25 @@ QString mp::utils::normalize_path(const QString& path)
     return QString::fromStdString(normalize_path(path.toStdString()));
 }
 
+std::filesystem::path mp::utils::qstr_to_path(const QString& qstr)
+{
+#ifdef _WIN32
+    return std::filesystem::path{qstr.toStdWString()};
```
xmkg: To preserve Unicode characters in Windows.

ricab: Not a review, I just noticed this flying through. Shouldn't this be in the platform instead? See [CPP9](https://github.com/canonical/multipass/blob/51b93861bf46672f51ced8da4b6fa8dd41f244cc/CONTRIBUTING.md#c-cpp)

include/multipass/utils.h
```diff
@@ -322,6 +327,16 @@ inline bool multipass::utils::iequals(std::string_view lhs, std::string_view rhs
     });
 }
 
+inline bool multipass::utils::istarts_with(std::string_view str, std::string_view prefix)
```
jimporter: If I recall correctly, C++ ranges have a tendency to instantiate a lot of templates, so this might be better in `utils.cpp`.

include/multipass/utils.h
```diff
@@ -322,6 +327,16 @@ inline bool multipass::utils::iequals(std::string_view lhs, std::string_view rhs
...
+{
+    if (prefix.size() > str.size())
+        return false;
+    return std::ranges::equal(
+        str | std::views::take(prefix.size()),
```
jimporter: Too bad we're not on C++23 yet, since `std::ranges:starts_with` would simplify this a bit. We could add a comment about that here, but no big deal either way.

src/utils/utils.cpp
jimporter: I'm ok with either option here. This is small enough (and hopefully temporary), that I wouldn't mind a platform `#ifdef` in this case. I'll leave it up to you and @ricab though.

src/utils/vm_image_vault_utils.cpp
```diff
@@ -15,95 +15,144 @@
...
+    auto new_location = output_dir / file.filename();
 
-    if (!MP_FILEOPS.copy(file, new_location))
+    // Normalize the path to platform's preferred slashes
+    new_location.make_preferred();
+    std::error_code ec{};
+    std::filesystem::copy_options options{};
+    MP_FILEOPS.copy(file, new_location, options, ec);
+    if (ec)
         throw std::runtime_error(fmt::format("Failed to copy {} to {}", file, new_location));
```
jimporter: Two things:
1. You could remove the declaration of `options` and just pass `{}` as the third arg to `copy`.
2. If we're just checking `ec` and immediately throwing, could we use the throwing form of `MP_FILEOPS.copy` instead? The C++ standard implementation of `copy` throws a `std::filesystem::filesystem_error`, which inherits from `std::runtime_error`, so we're good as far as the exception type is concerned. (If the error message that the standard library gives us is bad, we could just throw a nicer one in `file_ops.cpp`.)

src/utils/vm_image_vault_utils.cpp
```diff
@@ -15,95 +15,144 @@
...
+            throw std::runtime_error("Failed to read data from device to hash");
+        if (auto count = stream.gcount(); count > 0)
+        {
+            if (EVP_DigestUpdate(ctx.get(), buf, static_cast<size_t>(count)) != 1)
+                throw std::runtime_error("Failed to update hash");
+        }
+
+    } while (stream);
+
+    unsigned char digest[EVP_MAX_MD_SIZE]{};
```
jimporter: This initialization is a different style from above (which used `= {0}`). Since most of the rest of this function is using copy-initialization, how about this?


```suggestion
    unsigned char digest[EVP_MAX_MD_SIZE] = {0};
```

tests/unit/test_image_vault_utils.cpp
```diff
@@ -54,84 +56,120 @@ TEST_F(TestImageVaultUtils, copyToDirHandlesEmptyFile)
...
-    EXPECT_CALL(mock_file_ops, exists(test_info)).WillOnce(Return(true));
-    EXPECT_CALL(mock_file_ops, copy(test_path, test_output)).WillOnce(Return(false));
+    EXPECT_CALL(mock_file_ops, exists(test_path)).WillOnce(Return(true));
+
+    EXPECT_CALL(mock_file_ops, copy(test_path, test_output, _, _))
+        .WillOnce([](const std::filesystem::path&,
+                     const std::filesystem::path&,
+                     std::filesystem::copy_options,
+                     std::error_code& ec) {
+            ec = std::make_error_code(std::errc::no_such_file_or_directory);
```
jimporter: I think this would work too:


```suggestion
        .WillOnce(SetArgReferee<4>(std::make_error_code(std::errc::no_such_file_or_directory)));
```

src/utils/vm_image_vault_utils.cpp
```diff
@@ -15,95 +15,144 @@
...
+        throw std::runtime_error("Failed to initialize hash context");
+
+    constexpr std::size_t buf_size = 8192;
+    char buf[buf_size] = {0};
+    do
+    {
+        stream.read(buf, buf_size);
+        if (stream.bad())
+            throw std::runtime_error("Failed to read data from device to hash");
+        if (auto count = stream.gcount(); count > 0)
```
jimporter, line 90: `gcount` technically returns a signed value, but I don't think it can ever *actually* return something less than 0. I guess this would be equivalent:

```suggestion
        if (auto count = stream.gcount())
```

But on the other hand, what you have already is stricter. I can't quite decide which one I prefer, so I think I'm fine with either.

jimporter (changes requested):
Nice, a bit less Qt in our codebase! This looks good overall to me; just a few minor comments.

(I think this PR must have subconsciously inspired me to purge Qt from `VMImage` [here](https://github.com/canonical/multipass/pull/4767)! Once I rebase it, that PR should simplify things even further, since we'll be able to remove some `qstr_to_path` calls. Hopefully we're not getting in each other's way too much here...)

include/multipass/utils.h
xmkg: Done

src/utils/vm_image_vault_utils.cpp
xmkg: Done

src/utils/vm_image_vault_utils.cpp
xmkg: Done

tests/unit/test_image_vault_utils.cpp
xmkg: Done

src/utils/utils.cpp
ricab: I am a little confused, I see a thumbs-up but this is still here and my review was requested... what am I missing?

To clarify my issue with this:

- it prevents reasoning about code as platform-independent (when not located in the platform)
- it is a slippery slope

Is there a strong reason why this should be located in utils rather than platform? IIUC, this really is a platform operation at heart?

I understand it could be temporary, but Qt is very ingrained in our codebase and that "temporary" could take a very long time. It's been pursued for a few months now, yet:

```
$ rg --no-heading '#include <Q' | wc -l
381
```

xmkg: > I understand it could be temporary, but Qt is very ingrained in our codebase and that "temporary" could take a very long time

I just wanted to get your response before making the change, that's all. No clarification is needed.

xmkg: done.

ricab: Thank you Mustafa. 

I wonder if we should introduce a `platform_all.cpp` to make people's lives easier by being able to implement this sort of brief function once — with ifdefs — rather than needing separate implementations. The problem is not ifdefs themselves, it's spreading out platform logic (same as if the daemon started doing "if this backend this, otherwise that": that's not daemon logic, it's backend logic.)

jimporter (approved):
LGTM!

PR #4743, opened 2026-03-19
cli-tests: Absorb CI test runner failure

xmkg:
Some of the tests fail for legitimate reasons intermittently, and right now, we don't want to make the calling workflow fail if CLI tests fail.


xmkg:
Here's the CI run:

https://github.com/canonical/multipass/actions/runs/23294493260/job/67738238522

PR #4312, opened 2025-08-19
Availability Zones

Sploder12:
This branch adds support for Availability Zones to Multipass, hidden behind a feature flag.

PR #4757, opened 2026-03-23
[applevz] Additional networks

sharder996:
This PR adds supports for using additional network interfaces when using the `applevz` backend driver.

This implementation of vmnet network interfaces was largely inspired by the following existing works:

- https://github.com/nirs/vmnet-helper
- https://github.com/lima-vm/socket_vmnet
- https://github.com/oven-sh/bun/blob/main/packages/bun-usockets/src/internal/networking/bsd.h

Necessary modifications to integrate this code into Multipass is outlined at the top of `applevz_vmnet.mm`, but all credit for unmodified code goes to the original author.

See the [vmnet documentation](https://developer.apple.com/documentation/vmnet?language=objc) for more reading on vmnet API usage.

Once #4312 is merged into `main` and the interactions with the `AvailabilityZoneManager` is more finalized, this implementation can be easily extended to support shared network interfaces.

---

MULTI-2259
MULTI-2263

### DIFF
--- a/src/platform/backends/applevz/applevz_bridge.mm
+++ b/src/platform/backends/applevz/applevz_bridge.mm
@@ -29,6 +29,8 @@
 #include <QString>
 #include <QUrl>
 
+#include <cassert>
+
 namespace multipass::applevz
 {
 struct VirtualMachineHandle
@@ -172,18 +174,52 @@ CFError init_with_configuration(const multipass::VirtualMachineDescription& desc
             [[VZVirtioEntropyDeviceConfiguration alloc] init];
         config.entropyDevices = @[ entropy ];
 
-        // Network device
+        // Network devices
+        // Primary NAT interface
+        NSMutableArray<VZNetworkDeviceConfiguration*>* networkDevices = [NSMutableArray array];
+
         VZVirtioNetworkDeviceConfiguration* netDevice =
             [[VZVirtioNetworkDeviceConfiguration alloc] init];
-
         VZNATNetworkDeviceAttachment* natAttachment = [[VZNATNetworkDeviceAttachment alloc] init];
         netDevice.attachment = natAttachment;
 
         VZMACAddress* mac =
             [[VZMACAddress alloc] initWithString:nsstring_from_stdstring(desc.default_mac_address)];
         [netDevice setMACAddress:mac];
+        [networkDevices addObject:netDevice];
 
-        config.networkDevices = @[ netDevice ];
+        // Extra bridged interfaces
+        NSArray<VZBridgedNetworkInterface*>* hostInterfaces =
+            [VZBridgedNetworkInterface networkInterfaces];
+        for (const auto& extra : desc.extra_interfaces)
+        {
+            NSString* targetId = [NSString stringWithCString:extra.id.c_str()
+                                                    encoding:NSUTF8StringEncoding];
+
+            VZBridgedNetworkInterface* found = nil;
+            for (VZBridgedNetworkInterface* hostIface in hostInterfaces)
+            {
+                if ([hostIface.identifier isEqualToString:targetId])
+                {
+                    found = hostIface;
+                    break;
+                }
+            }
+
+            assert(found && "Bridged interface not found - extra_interfaces must be validated "
+                            "before VM creation");
+
+            VZBridgedNetworkDeviceAttachment* bridgedAttachment =
+                [[VZBridgedNetworkDeviceAttachment alloc] initWithInterface:found];
+
+            VZVirtioNetworkDeviceConfiguration* bridgedDevice =
+                [[VZVirtioNetworkDeviceConfiguration alloc] init];
+            bridgedDevice.attachment = bridgedAttachment;
+
+            [networkDevices addObject:bridgedDevice];
+        }
+
+        config.networkDevices = networkDevices;
 
         // Memory balloon device
         VZVirtioTraditionalMemoryBalloonDeviceConfiguration* balloon =
--- a/src/platform/backends/applevz/applevz_bridge.mm
+++ b/src/platform/backends/applevz/applevz_bridge.mm
@@ -306,4 +306,23 @@ bool is_supported()
 {
     return [VZVirtualMachine isSupported];
 }
+
+std::vector<NetworkInterfaceInfo> bridged_network_interfaces()
+{
+    std::vector<NetworkInterfaceInfo> result;
+
+    for (VZBridgedNetworkInterface* iface in [VZBridgedNetworkInterface networkInterfaces])
+    {
+        NetworkInterfaceInfo info;
+        info.id = iface.identifier.UTF8String;
+        info.type = "N/A";
+        info.description = iface.localizedDisplayName
+                               ? std::string(iface.localizedDisplayName.UTF8String)
+                               : info.id;
+
+        result.push_back(std::move(info));
+    }
+
+    return result;
+}
 } // namespace multipass::applevz
--- a/src/platform/backends/applevz/applevz_bridge.mm
+++ b/src/platform/backends/applevz/applevz_bridge.mm
@@ -199,6 +199,10 @@ CFError init_with_configuration(const multipass::VirtualMachineDescription& desc
             VZVirtioNetworkDeviceConfiguration* bridgedDevice =
                 [[VZVirtioNetworkDeviceConfiguration alloc] init];
             bridgedDevice.attachment = bridge.attachment;
+            VZMACAddress* bridgedMac = [[VZMACAddress alloc]
+                initWithString:[NSString stringWithCString:extra.mac_address.c_str()
+                                                  encoding:NSUTF8StringEncoding]];
+            [bridgedDevice setMACAddress:bridgedMac];
 
             [networkDevices addObject:bridgedDevice];
 
--- a/src/platform/backends/applevz/applevz_bridge.mm
+++ b/src/platform/backends/applevz/applevz_bridge.mm
@@ -36,9 +36,10 @@ namespace multipass::applevz
 {
 struct VirtualMachineHandle
 {
-    std::shared_ptr<void> vm; // Ownership transfer of VZVirtualMachine*
-    dispatch_queue_t queue;   // Dispatch queue for VM operations
-    uint64_t id;              // Unique VM ID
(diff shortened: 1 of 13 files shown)


src/platform/backends/applevz/applevz_bridge.mm
```diff
@@ -172,18 +176,40 @@ CFError init_with_configuration(const multipass::VirtualMachineDescription& desc
...
         VZMACAddress* mac =
             [[VZMACAddress alloc] initWithString:nsstring_from_stdstring(desc.default_mac_address)];
         [netDevice setMACAddress:mac];
+        [networkDevices addObject:netDevice];
+
+        // Extra bridged interfaces
+        std::vector<VmnetRelayHandle> relays;
+        for (const auto& extra : desc.extra_interfaces)
+        {
+            VmnetBridge bridge = create_vmnet_bridge(extra.id);
```
jimporter, line 197: I don't feel super-strongly either way, but would this be clearer?

```suggestion
            auto& bridge = relays.emplace_back(create_vmnet_bridge(extra.id));
```

src/platform/backends/applevz/applevz_bridge.mm
```diff
@@ -172,18 +176,40 @@ CFError init_with_configuration(const multipass::VirtualMachineDescription& desc
...
+                [[VZVirtioNetworkDeviceConfiguration alloc] init];
+            bridgedDevice.attachment = bridge.attachment;
+            VZMACAddress* bridgedMac = [[VZMACAddress alloc]
+                initWithString:[NSString stringWithCString:extra.mac_address.c_str()
+                                                  encoding:NSUTF8StringEncoding]];
+            [bridgedDevice setMACAddress:bridgedMac];
+
+            [networkDevices addObject:bridgedDevice];
+
+            relays.push_back(std::move(bridge.relay));
```
jimporter, line 209: Followed by...

```suggestion
```

src/platform/backends/applevz/applevz_bridge.mm
```diff
@@ -306,4 +332,23 @@ bool is_supported()
...
+    for (VZBridgedNetworkInterface* iface in [VZBridgedNetworkInterface networkInterfaces])
+    {
+        NetworkInterfaceInfo info;
+        info.id = iface.identifier.UTF8String;
+        info.type = "N/A";
+        info.description = iface.localizedDisplayName
+                               ? std::string(iface.localizedDisplayName.UTF8String)
+                               : info.id;
+
+        result.push_back(std::move(info));
```
jimporter: How about something like this?

```suggestion
        result.emplace_back(/* id = */ iface.identifier.UTF8String;
                            /* type = */ "N/A",
                            /* description = */ iface.localizedDisplayName
                               ? std::string(iface.localizedDisplayName.UTF8String)
                               : info.id);
```

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+    std::vector<uint8_t> host_buffers;
+    std::vector<vmpktdesc> host_packets;
+    std::vector<iovec> host_iovs;
+    std::vector<msghdr_x> host_msgs;
+
+    // Pre-allocated buffers for vm->host (socket read / vmnet write).
+    std::vector<uint8_t> vm_buffers;
+    std::vector<vmpktdesc> vm_packets;
+    std::vector<iovec> vm_iovs;
+    std::vector<msghdr_x> vm_msgs;
```
jimporter: I think a few parts of this file could be simplified if you made a `VmnetBuffers` struct (feel free to name it something better; that's just the first thing that came to mind). For example here:


```suggestion
    struct VmnetBuffers
    {
        std::vector<uint8_t> buffers;
        std::vector<vmpktdesc> packets;
        std::vector<iovec> iovs;
        std::vector<msghdr_x> msgs;
    };
    // Pre-allocated buffers
    VmnetBuffers host; // host->vm (vmnet read / socket write).
    VmnetBuffers vm; // vm->host (socket read / vmnet write).
```

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+                iovs[i].iov_len = max_packet_bytes;
+                packets[i].vm_pkt_iov = &iovs[i];
+                packets[i].vm_pkt_iovcnt = 1;
+                msgs[i].msg_hdr.msg_iov = &iovs[i];
+                msgs[i].msg_hdr.msg_iovlen = 1;
+            }
+        };
+
+        bind_endpoint(host_buffers, host_packets, host_iovs, host_msgs);
+        bind_endpoint(vm_buffers, vm_packets, vm_iovs, vm_msgs);
```
jimporter: Then `bind_endpoint` could be a method on `VmnetBuffers`, so here you'd just call:

```c++
host.bind_endpoint();
vm.bind_enpoint();
```

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    mpl::debug(category, "vmnet bridged interface started on '{}'", physical_iface);
+}
+
+// Forward packets from the socket to the vmnet interface.
+// Runs as a blocking loop on a dedicated queue until the socket is closed.
+void forward_from_vm(VmnetRelay& relay, bool bulk)
+{
+    for (;;)
```
jimporter: This function runs a loop inside its body...

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+        vmnet_return_t status = vmnet_write(relay.iface, relay.vm_packets.data(), &write_count);
+        if (status != VMNET_SUCCESS)
+            mpl::trace(category, "vmnet_write() failed (status {})", static_cast<int>(status));
+    }
+}
+
+// Forward one batch of packets from the vmnet interface to the socket.
+// Returns false when vmnet has no more packets to deliver this callback.
+bool forward_from_host(VmnetRelay& relay, bool bulk)
+{
```
jimporter, line 235: ...but for this one, the main loop is defined in `start_forwarding_from_host`. While the loops themselves are different (the former is an infinite loop, and the latter loops until `forward_from_host` returns `false`), I think it would be clearer/more-symmetric if both had their main loops in the same spot. I don't really have a preference on which spot though...

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+
+namespace
+{
+constexpr auto category = "vmnet";
+constexpr uint32_t kSendBufferSize{65 * 1024};
+constexpr uint32_t kRecvBufferSize{4 * 1024 * 1024};
+constexpr int kMaxPacketCount{200};
+constexpr int kMaxPacketCountLarge{16};
+constexpr uint32_t kLargePacketThreshold{64 * 1024};
+constexpr uint64_t kSendRetryDelayNs{50 * 1000};
```
jimporter: All the uses of this below are for `sleep_for` calls, so how about this?

```suggestion
constexpr std::chrono::microseconds kSendRetryDelay{50};
```

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+            ssize_t sent;
+            do
+            {
+                sent = send(relay.fd, data, pkt_size, 0);
+                if (sent < 0 && errno == ENOBUFS)
+                    std::this_thread::sleep_for(std::chrono::nanoseconds{kSendRetryDelayNs});
+            } while (sent < 0 && errno == ENOBUFS);
+
+            if (sent < 0)
+                mpl::trace(category, "send() failed: {}", strerror(errno));
```
jimporter: This took me a couple tries to navigate through in my head. What do you think of the following? I *think* it should be equivalent.

```suggestion
            while ((sent = send(relay.fd, data, pkt_size, 0)) < 0)
            {
                if (errno == ENOBUFS)
                    std::this_thread::sleep_for(std::chrono::nanoseconds{kSendRetryDelayNs});
                else
                {
                    mpl::trace(category, "send() failed: {}", strerror(errno));
                    break;
                }
            }
```

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -0,0 +1,369 @@
...
+    std::array<int, 2> fds{};
+    if (socketpair(AF_UNIX, SOCK_DGRAM, 0, fds.data()) < 0)
+        throw std::runtime_error(fmt::format("socketpair() failed: {}", strerror(errno)));
+
+    setsockopt(fds[0], SOL_SOCKET, SO_RCVBUF, &kRecvBufferSize, sizeof(kRecvBufferSize));
+    setsockopt(fds[0], SOL_SOCKET, SO_SNDBUF, &kSendBufferSize, sizeof(kSendBufferSize));
+    setsockopt(fds[1], SOL_SOCKET, SO_RCVBUF, &kRecvBufferSize, sizeof(kRecvBufferSize));
+    setsockopt(fds[1], SOL_SOCKET, SO_SNDBUF, &kSendBufferSize, sizeof(kSendBufferSize));
+
+    return {fds[0], fds[1]};
```
jimporter: I think you can just return the `std::array` directly here (after adjusting the return type). The structured binding below should still work. (Another option would be to pass a `std::pair` to `socketpair`, which *should* be legal, but maybe it's a bit too sneaky...)

jimporter (changes requested):
I don't have a lot of knowledge about Apple's vmnet, so I just have some more general comments below.

src/platform/backends/applevz/applevz_bridge.mm
sharder996: Don't think this one works... We need the `attachment` from the bridge to give to the AppleVZ framework, but only need to keep the `relay` in the VM configuration.

jimporter: Ah ha, I misread this the first time around.

jimporter (dismissed):
LGTM, thanks!

sharder996:
@jimporter Rebased on `main` in order to get updated CI workflows. ~~Should be no new changes.~~ Aligned some VM factory tests with AZs.

jimporter (approved):
Reapproving after the rebase

PR #4758, opened 2026-03-23
[simplestreams] Use Boost.JSON

jimporter:
# Description

This PR migrates the simple streams code over to Boost.JSON. I've tried to match the old behavior as closely as possible, so most of the simple streams manifest fields are optional (unless they're required or their absence wouldn't make sense). Maybe we could be stricter about missing fields...

## Testing

- Covered by existing unit tests.
- Manual testing:
  1. Start the multipass daemon from an empty config
  2. Make sure you can launch an Ubuntu image

## Checklist

### DIFF
--- a/src/simplestreams/simple_streams_manifest.cpp
+++ b/src/simplestreams/simple_streams_manifest.cpp
@@ -19,9 +19,9 @@
 
 #include <QFileInfo>
 #include <QHash>
-#include <QJsonDocument>
-#include <QJsonObject>
 #include <QSysInfo>
+
+#include <boost/json.hpp>
 
 #include <multipass/constants.h>
 #include <multipass/exceptions/manifest_exceptions.h>
@@ -39,24 +39,22 @@ const QHash<QString, QString> arch_to_manifest{{"x86_64", "amd64"},
                                                {"power64", "ppc64el"},
                                                {"power64le", "ppc64el"}};
 
-QJsonObject parse_manifest(const QByteArray& json)
+const boost::json::object* if_contains_object(const boost::json::value& data, std::string_view key)
 {
-    QJsonParseError parse_error;
-    const auto doc = QJsonDocument::fromJson(json, &parse_error);
-    if (doc.isNull())
-        throw mp::GenericManifestException(parse_error.errorString().toStdString());
-
-    if (!doc.isObject())
-        throw mp::GenericManifestException("invalid manifest object");
-    return doc.object();
+    if (auto elem = data.as_object().if_contains(key))
+    {
+        if (auto obj = elem->if_object())
+            return obj;
+    }
+    return nullptr;
 }
 
-QString latest_version_in(const QJsonObject& versions)
+QString latest_version_in(const boost::json::object& versions)
 {
     QString max_version;
-    for (auto it = versions.constBegin(); it != versions.constEnd(); ++it)
+    for (const auto& [key, _] : versions)
     {
-        const auto version = it.key();
+        const auto version = QString::fromStdString(key);
         if (version < max_version)
             continue;
         max_version = version;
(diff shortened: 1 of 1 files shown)


src/platform/backends/shared/base_snapshot.cpp
```diff
@@ -93,7 +93,7 @@ std::shared_ptr<mp::Snapshot> find_parent(const mp::SnapshotDescription& desc,
 
 mp::BaseSnapshot::BaseSnapshot(SnapshotDescription desc_,
                                std::shared_ptr<Snapshot> parent_,
-                               const VirtualMachine& vm_,
+                               VirtualMachine& vm_,
                                bool captured_)
```
ricab: I don't mind args being named differently from the fields, even though the rules governing what is what should be clear enough as it is. However, I don't know about the trailing underscore, as that is often used in the opposite direction: to signal (usually private) fields. I was actually confused at first and went looking for the change in the fields.

include/multipass/snapshot_description.h
ricab, line 1: The compilation unit for this header is in `platforms/backends/shared`. That is also where the only thing depending on it is located, so it seems like a better place for this header too, no? I believe the include folder is meant for generic things that are used in different parts of the project.

(comment on the first line because GH didn't allow it on the file)

Image

src/platform/backends/shared/snapshot_description.cpp
ricab: Why move? `ctx` is `const` (and so is the `vm_desc` within).

tests/unit/test_base_snapshot.cpp
ricab, line 95: Ah, so much better, this was a real nuisance in Qt.

ricab, line 705: 👍

tests/unit/test_json_utils.cpp
ricab, line 99: Shouldn't these be replaced somehow?

src/platform/backends/shared/base_snapshot.cpp
ricab, line 190: Why remove the assert?

ricab, line 106: I am not sure this matches the original logic. This would end up being called when taking a new snapshot, right? If so, it could not have been captured yet and should not be persisted, correct? 

OTH, perhaps this is OK because it would never be true, and then it could probably be merged with the following constructor. Were you planning on doing that? Is that perhaps why you removed the `const` from the VM param (with the view of merging the two)?

src/platform/backends/shared/base_snapshot.h
ricab, line 96: I believe you forgot to remove this.

jimporter:
@ricab Weird, I'm not sure how the Snapshot code merged into this branch. Now fixed.

edit: Oh, I think I see what happened. I fixed a couple of things in the other PR this was built on top of. Then, when that PR merged, GitHub automatically changed the base here, which meant that the *old* version of the snapshot commits stuck around in this PR. I guess I should have just based this PR on `main` originally instead of making a chain of PRs.

src/platform/backends/shared/base_snapshot.cpp
jimporter: I'm ok with changing this, though the reason was because we use some of these variables in the body of the constructor, where same-named arguments would take priority over the member variables. `this->member` would resolve that, but that alternative is a bit verbose.

include/multipass/snapshot_description.h
jimporter: Makes sense.

src/platform/backends/shared/snapshot_description.cpp
jimporter: Fixed.

src/platform/backends/shared/base_snapshot.h
jimporter: I think Scott noticed this, and I already fixed it before merging.

src/platform/backends/shared/base_snapshot.cpp
ricab: Sure, I would be fine with different names, it's just ending with underscore is easy to read the other way around. We have discussed introducing a distinguishing naming convention for members, but that didn't go anywhere yet. Maybe that's something you would like to pick up? :wink: LLMs should help make the change, though I'd suggest one commit per class or type. I pity the reviewer though... and it is bound to generate a bunch of conflicts. But it's a one time thing.

src/simplestreams/simple_streams_index.cpp
```diff
@@ -20,45 +20,29 @@
...
-QJsonObject parse_index(const QByteArray& json)
+mp::SimpleStreamsIndex mp::SimpleStreamsIndex::fromJson(const QByteArray& json)
 {
-    QJsonParseError parse_error;
-    auto doc = QJsonDocument::fromJson(json, &parse_error);
-    if (doc.isNull())
-        throw std::runtime_error(parse_error.errorString().toStdString());
-
-    if (!doc.isObject())
+    auto doc = boost::json::parse(std::string_view(json));
```
ricab: Good use of the operator. I should make a mental note that Qt now provides these. 

How about using a `static_cast` rather than the functional cast?

```suggestion
    auto doc = boost::json::parse(static_cast<std::string_view>(json));
```

src/simplestreams/simple_streams_index.cpp
```diff
@@ -20,45 +20,29 @@
...
-        auto entry = value.toObject();
-        if (entry["datatype"] == "image-downloads")
+        auto entry = value.as_object();
+        if (entry.at("datatype") == "image-downloads")
         {
-            auto path_entry = entry["path"];
-            auto date_entry = entry["updated"];
-            return {path_entry.toString(), date_entry.toString()};
+            auto path_entry = value_to<std::string>(entry.at("path"));
+            auto date_entry = value_to<std::string>(entry.at("updated"));
```
ricab: This would now throw `boost::system::system_error` if these were absent, whereas empty objects would have been used before that. Was that intentional?

Note: when called from `mp::UbuntuVMImageHost::fetch_manifests`, this is wrapped in a try-catch that deals with a few specific exceptions, but not that boost one. I didn't track it down deep enough to find if empty objects here would result in one of the handled exceptions or something else, upon download.

src/platform/backends/shared/base_snapshot.cpp
jimporter: @ricab I'll send a message to the mailing list about naming for members. Luckily for us, if we adopt a new naming convention, we can fix our entire source tree deterministically with a simple `clang-tidy` rule. (I tried this out locally to see how it works, and it's pretty simple. We could even add that commit to a `blame.ignoreRevsFile` so that it doesn't mess with `git blame`!)

src/simplestreams/simple_streams_index.cpp
jimporter: How about brace-initialization, like `std::string_view{json}`? I think that's what you suggested previously.

src/simplestreams/simple_streams_index.cpp
jimporter: Fixed to use `lookup_or`. I think I missed doing this, since I originally wrote this commit before I created the `lookup_or` utility.

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -96,93 +94,95 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(
...
-            else if (product["os"] == "ubuntu-core" && items.contains("img.xz"))
+            else if (items->contains("img.xz") && os == "ubuntu-core")
                 image_key = "img.xz";
             // Last resort, use img
-            else
+            else if (items->contains("disk1.img"))
                 image_key = "disk1.img";
+            // Otherwise, give up
+            else
+                continue;
```
jimporter, line 175: @ricab I narrowed down the issue with the Maverick image to this hunk. It seems that Maverick is missing a compatible disk image. If you look at the [manifest](https://cloud-images.ubuntu.com/releases/streams/v1/com.ubuntu.cloud:released:download.json), the keys for the Maverick `items` object are only `manifest` and `tar.gz`, neither of which we support here.

I'd consider this a bugfix, since we can't even get a usable image file for Maverick. What do you think?

Addendum: here's the output when trying to launch a Maverick image on `main`:

```
$ ./bin/multipass launch maverick
launch failed: unable to write to file "/root/.cache/multipassd/vault/images/maverick-20120410"
```

Lucid fails like this:

```
± ./bin/multipass launch lucid
launch failed: Hash of /root/.cache/multipassd/vault/images/lucid-20150427/ubuntu-10.04-server-cloudimg-amd64-disk1.img does not match (expected 983b7be9dd1ed8c0b3118e309cdd2b2869b74ddcbb3e4c6c4c77e28e0a7b5e59 but got e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855)
```

For Maverick, the daemon doesn't log anything at all (presumably because it never downloads an image in the first place), but when trying to launch Lucid, the daemon logs this:

```
[2026-04-01T18:39:57.841] [warning] [url downloader] Failed to get https://cloud-images.ubuntu.com/releases/server/releases/lucid/release-20150427/ubuntu-10.04-server-cloudimg-amd64-disk1.img: Insecure redirect - trying cache.
```

src/simplestreams/simple_streams_index.cpp
ricab: Sure, sounds good.

src/simplestreams/simple_streams_manifest.cpp
ricab: Ah, alright, I agree. All clear here then.

src/simplestreams/simple_streams_index.cpp
```diff
@@ -20,45 +20,29 @@
...
 #include <multipass/simple_streams_index.h>
 
-#include <QJsonDocument>
-#include <QJsonObject>
+#include <boost/json.hpp>
 
 #include <stdexcept>
+#include <string_view>
+
+#include <multipass/json_utils.h>
```
ricab: Could you move this together with the other multipass include please?

src/simplestreams/simple_streams_manifest.cpp
```diff
@@ -169,11 +175,14 @@ std::unique_ptr<mp::SimpleStreamsManifest> mp::SimpleStreamsManifest::fromJson(
             if (items.contains("uefi1.img"))
                 image_key = "uefi1.img";
             // For Ubuntu Core images
-            else if (product["os"] == "ubuntu-core" && items.contains("img.xz"))
+            else if (items.contains("img.xz") && os == "ubuntu-core")
```
ricab, line 178: Tiny thing, but the second check is probably the fastest, so maybe better keep it first? Up to you.

PR #4761, opened 2026-03-24
[renovate] Improve review assignment and add cooldown

jimporter:
# Description

This makes two small improvements to our Renovate configs:
1. [Assign](https://docs.renovatebot.com/configuration-options/#reviewers) the Multipass team as reviewer so we can manage our reviewer list in a single place.
2. [Add a cooldown](https://docs.renovatebot.com/configuration-options/#minimumreleaseagebehaviour) to dependencies to make it easier to avoid supply-chain attacks.

(If one or the other of these is controversial, I could split it out. However, they're both small enough that I figured a single PR would be easier for review.)

## Checklist

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -8,4 +8,5 @@
     ],
     "schedule": "every 2 weeks on Monday",
     "commitMessagePrefix": "[deps] ",
+    "minimumReleaseAge": "5 days",
 }


renovate.json5
```diff
@@ -1,16 +1,16 @@
...
+    ],
+    "reviewers": [
+        "team:multipass",
+    ],
+    "schedule": "every 2 weeks on Monday",
+    "commitMessagePrefix": "[deps] ",
+
+    // Don't create PRs until the new dep is at least 5 days old
+    "minimumReleaseAge": "5 days",
+    "prCreation": "not-pending",
```
sharder996: How does this work if status checks are only run once a PR has been created?

I see this in the documentation which I assume inspired this change:

> It's recommended to also apply prCreation="not-pending" and internalChecksFilter="strict" to make sure that branches and PRs are only created after the minimumReleaseAge has passed.

but its under the [automerging](https://docs.renovatebot.com/configuration-options/#await-x-time-duration-before-automerging) heading which I don't believe we are doing.

jimporter: I looked through the docs a bit more to verify how things are supposed to work. Based on [this page](https://docs.renovatebot.com/key-concepts/minimum-release-age/), I think the problem is that the Renovate team didn't fully update the page I originally linked. This other page says:

> > Note: A previous version of the documentation (up until Renovate 42.19.9) recommended configuring [prCreation](https://docs.renovatebot.com/configuration-options/#prcreation). This is no longer the case.
>
> If no branch is created, Renovate will not raise a PR, regardless of [prCreation](https://docs.renovatebot.com/configuration-options/#prcreation)'s settings.

That page also says:

> If you have not configured [internalChecksFilter](https://docs.renovatebot.com/configuration-options/#internalchecksfilter), Renovate will use internalChecksFilter=strict as the default.

So I think we only really need `minimumReleaseAge`, and the others can be removed.

sharder996: Yeah, that makes more sense to me 😄

PR #4762, opened 2026-03-25
[utils] Use standard fs paths for AliasDict

jimporter:
# Description

This is just a small change to remove the last bits of Qt from `alias_dict.cpp`. It does two things:
1. Add a C++-standard overload of `write_transactionally` that we can use here
2. Use `std::filesystem::path` to generate the default filename

## Testing

Existing unit tests in `tests/unit/test_alias_dict.cpp` check that we can read in an `AliasDict` from the default filename.

## Checklist

### DIFF
--- a/include/multipass/file_ops.h
+++ b/include/multipass/file_ops.h
@@ -32,6 +32,7 @@
 
 #include <filesystem>
 #include <fstream>
+#include <string_view>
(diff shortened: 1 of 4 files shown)


sharder996 (approved):
LGTM

PR #4763, opened 2026-03-25
[gui] Mark latest LTS release as default image

sharder996:
`multipass launch` will create a VM using the image that is marked as the current LTS in simplestreams. However, when I go to launch an Ubuntu instance in the GUI, the most recently released image is the image that is prefilled in the dropdown box. This PR changed the behaviour of the dropdown box to be prefilled with the latest LTS release.

image


amylily1011:
LGTM :shipit: @sharder996

PR #4765, opened 2026-03-25
[vault] Search by image hash

sharder996:
This PR changes how the image vault searches for images in the cache by comparing the image hash/id instead of remote name + image name. This prevents some odd behaviours such as `multipass launch 26.04` and `multipass launch daily:26.04` which use the same image but causes it to be downloaded again.

Also cleaned up some dead code while I was at it.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -107,6 +107,7 @@ const std::unordered_set<std::string> no_bridging_release =
         "10.04",  "lucid", "11.10", "oneiric", "12.04",  "precise", "12.10",  "quantal", "13.04",
         "raring", "13.10", "saucy", "14.04",   "trusty", "14.10",   "utopic", "15.04",   "vivid",
         "15.10",  "wily",  "16.04", "xenial",  "16.10",  "yakkety", "17.04",  "zesty"};
+const std::unordered_set<std::string> no_bridging_core = {"core16"};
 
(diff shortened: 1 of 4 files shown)


src/daemon/daemon.cpp
```diff
@@ -107,10 +107,7 @@ const std::unordered_set<std::string> no_bridging_release =
         "10.04",  "lucid", "11.10", "oneiric", "12.04",  "precise", "12.10",  "quantal", "13.04",
         "raring", "13.10", "saucy", "14.04",   "trusty", "14.10",   "utopic", "15.04",   "vivid",
         "15.10",  "wily",  "16.04", "xenial",  "16.10",  "yakkety", "17.04",  "zesty"};
```
tobe2098, line 109: We do not support these images anymore, do we still check for custom images?

tobe2098 (approved):
LGTM, very good work Scott! Nice to see optimizations like these.

src/daemon/daemon.cpp
sharder996: They will show up when you use the `--show-unsupported` flag. We have talked about not showing these images anymore, but I'm not sure if a final decision was made.

sharder996:
@tobe This good for merging or did you have more comments?

tobe2098 (approved):

PR #4766, opened 2026-03-25
[applevz] Unsupported ops

sharder996:
This PR adds some polishing touches to the `applevz` backend in Multipass. Specifically:

- `multipass snapshot` throws `NotImplementedOnThisBackendException`
- `multipass suspend foo` throws `NotImplementedOnThisBackendException`
- VMs are returned to their original state after the daemon reboots
- AppleVZ will report a VM running as soon as the kernel has booted. Let `wait_until_ssh_up` transition state to `Running`

---

MULTI-2256
MULTI-2269

### DIFF
--- a/src/platform/backends/applevz/applevz_virtual_machine.h
+++ b/src/platform/backends/applevz/applevz_virtual_machine.h
@@ -56,6 +56,13 @@ public:
     void resize_memory(const MemorySize& new_size) override;
     void resize_disk(const MemorySize& new_size) override;
 
+    std::shared_ptr<const Snapshot> take_snapshot(const VMSpecs& /*specs*/,
+                                                  const std::string& /*snapshot_name*/,
+                                                  const std::string& /*comment*/) override
+    {
+        throw NotImplementedOnThisBackendException{"snapshots"};
+    }
+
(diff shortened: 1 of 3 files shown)


src/platform/backends/applevz/applevz_virtual_machine.h
```diff
@@ -65,5 +72,6 @@ class AppleVZVirtualMachine : public BaseVirtualMachine
     VirtualMachineDescription desc;
     VMStatusMonitor* monitor;
     VMHandle vm_handle{nullptr};
+    bool update_shutdown_status{true};
```
xmkg, line 75: I didn't quite get the purpose of this one -- would you elaborate a bit?

sharder996: This controls whether or not the daemon will record the new state. When the VM shuts down because the daemon is shutting down, we want the daemon to return the VM to that state when it starts up again. `update_shutdown_status` will stop `shutdown()` from calling `persist_state_for()` when the VM object is destructed while the VM is still running.

xmkg (approved):

PR #4767, opened 2026-03-26
[vmimage] Use `std::filesystem::path`

jimporter:
# Description

This converts `VMImage::image_path` from a `QString` to a `std::filesystem::path`. Since `VMImage` is used in quite a few places, a bunch of other code needed to be updated to accommodate this change.

## Related Issue(s)

This builds on the prior work from #4741.

## Testing

- Unit tests are all updated to account for this change. The underlying data and all the behaviors are identical to before.

## Checklist

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1332,7 +1332,7 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         }
 
         auto vm_image = fetch_image_for(name, *config->factory, *config->vault);
-        if (!vm_image.image_path.isEmpty() && !QFile::exists(vm_image.image_path))
+        if (!vm_image.image_path.empty() && !MP_FILEOPS.exists(vm_image.image_path))
         {
             mpl::warn(category,
                       "Could not find image for '{}'. Expected location: {}",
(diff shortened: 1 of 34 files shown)


src/daemon/daemon.cpp
```diff
@@ -1332,7 +1332,7 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
         }
 
         auto vm_image = fetch_image_for(name, *config->factory, *config->vault);
-        if (!vm_image.image_path.isEmpty() && !QFile::exists(vm_image.image_path))
+        if (!vm_image.image_path.empty() && !MP_FILEOPS.exists(vm_image.image_path))
```
jimporter, line 1335: Arguably, this could just be `std::filesystem::exists`  (though ADL should mean the `std::filesystem` is optional), but while I was here, it seemed like a good idea to use `MP_FILEOPS`. That way we're being more consistent about using mockable functions when checking the actual filesystem.

src/daemon/default_vm_image_vault.cpp
```diff
@@ -207,24 +208,24 @@ mp::VMImage mp::DefaultVMImageVault::fetch_image(const FetchType& fetch_type,
             throw std::runtime_error(
                 fmt::format("Invalid file URL `{}`; did you forget a slash?", query.release));
 
-        source_image.image_path = image_url.toLocalFile();
+        source_image.image_path = image_url.toLocalFile().toStdString();
 
-        if (!QFile::exists(source_image.image_path))
+        if (!MP_FILEOPS.exists(source_image.image_path))
```
jimporter, line 213: Same here.

src/daemon/daemon.cpp
xmkg: Good call!

xmkg (approved):
LGTM, thanks for this, Jim! It's a great follow-up to #4741, and I always enjoy seeing red QT lines replaced with the standard library.

PR #4770, opened 2026-03-27
Bugfix/remove footgun from trim functions

xmkg:
# Description

- What does this PR do?
Make trim return Str instead of Str&& to avoid returning dangling rvalue references.

- Why is this change needed?
To make it harder to misuse the trim_* functions.

MULTI-2515

### DIFF
--- a/tests/unit/test_utils.cpp
+++ b/tests/unit/test_utils.cpp
@@ -325,6 +325,21 @@ TEST(Utils, toCmdArgumentsWithDoubleQuotesAreEscaped)
     EXPECT_THAT(output, ::testing::StrEq("they said \\\"please\\\""));
 }
 
+// clang-format off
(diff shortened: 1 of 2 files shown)


tests/unit/test_utils.cpp
```diff
@@ -325,6 +325,28 @@ TEST(Utils, toCmdArgumentsWithDoubleQuotesAreEscaped)
     EXPECT_THAT(output, ::testing::StrEq("they said \\\"please\\\""));
 }
 
+// clang-format off
+template <typename Func, typename... ExtraArgs>
```
sharder996, line 329: Compile-time tests 😎

sharder996 (approved):
LGTM! Time to introduce concepts to Multipass?

PR #4773, opened 2026-03-30
[main] Add GOOGLE_PROTOBUF_VERIFY_VERSION to main

xmkg:
This macro verifies the linked protobuf lib version against headers.

# Description

- What does this PR do?
Add GOOGLE_PROTOBUF_VERIFY_VERSION macro call to daemon (unix, win), cli, and unit test main functions.
- Why is this change needed?
To ensure that the linked protobuf header and library versions match.

MULTI-2514

### DIFF
--- a/src/client/cli/main.cpp
+++ b/src/client/cli/main.cpp
@@ -49,5 +49,9 @@ int main_impl(int argc, char* argv[])
 
 int main(int argc, char* argv[])
 {
+    // Verify that the version of the library that we linked against is
+    // compatible with the version of the headers we compiled against.
+    GOOGLE_PROTOBUF_VERIFY_VERSION;
+
     return mp::top_catch_all("client", /* fallback_return = */ EXIT_FAILURE, main_impl, argc, argv);
 }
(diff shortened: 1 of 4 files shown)


sharder996 (commented):
I don't foresee this causing issues since we own both the client and server sides. However, the GUI also makes use of gRPC. Can we do the same check there?

xmkg:
> I don't foresee this causing issues since we own both the client and server sides.

Yeah, normally it shouldn't, and afaik different protobuf versions talking to each other is a supported use case. It's just to ensure that the binary is linking against the exact version of the protobuf that it used the headers of. It can happen with stale caches, buggy compilation caches or non-pristine environments that has a protobuf installed at the system level.

xmkg:
>However, the GUI also makes use of gRPC. Can we do the same check there?

I've taken a look at the GUI code, and it seems like we're using protobuf via Dart. This macro is only relevant for the C++ code that consumes the protobuf libs/headers.

sharder996 (approved):
LGTM! Thanks for the additional info!

PR #4539, opened 2025-12-01
[qemu] Set macos build target

sharder996:
Set the minimum macOS build target for QEMU and libssh.

QEMU still dynamically links against some brew dependencies which causes some warnings during the build process. Please let me know what the CLI tests reveal.

#4538 

---

MULTI-2396

PR #4220, opened 2025-07-09
[tools/cli_tests] introduce Multipass CLI tests

xmkg:
CLI tests aim to test the high-level, user-facing Multipass functionality
via the CLI commands. The tests are written in Python, and the suite uses
`pytest` as the test executor.

See `README.md` for more details.

Signed-off-by: Mustafa Kemal Gilor <mustafa.gilor@canonical.com>

PR #4774, opened 2026-03-30
[ci] Add CLI tests and modernized CI to the release branch

tobe2098:


tobe2098:
CLI tests are essentially untouched, so they do not require extensive review.

tobe2098:
I would outline that the main concerns are 1) the package versions are the same as on the 1.16.1 release 2) that the build system is not broken 3) that we publish the appropriate macos packages

sharder996 (changes requested):
For the backwards compatible builds, I believe you are still missing:
- #4539, and
- #4547, and possibly
- #4540

tobe2098:
Added. The last one is empty because I had to manually add the changes in an earlier cherry-pick.

tobe2098:
I re-did the cherry-picks, now it should be closer to actual commit history and conflict resolution was minimal.

sharder996 (approved):
CLI tests where entirely encapsulated in #4220 which I see in the commit history. As the title of the PR suggests, my approval encompasses the inclusion of the CLI tests and CI changes. Not necessarily approving the state of the release branch.

I will let @tobe2098 handle the merging of this into `release/1.16`.

xmkg (approved):

PR #4779, opened 2026-03-31
[qemu] Distinguish QemuPlatform implementations

ricab:
# Description

Distinguish QemuPlatform implementations by renaming them according to the appropriate platform. 
This helps with grep, IDEs, and makes things easier to reason about.

## Testing

- Unit tests.
- Launched instance on Linux.

## Checklist


xmkg (approved):
LGTM, this will indeed make things easier to search. I am going to ignore the coverage regression since this patch is a name refactor only.

ricab:
Thanks @xmkg! For the record, I don't think the coverage regressed. The places where those identifiers are mentioned are not tested above the threshold, but that remains exactly as it was.

PR #4783, opened 2026-04-02
[applevz] Support applevz with CLI tests

sharder996:
This PR adds some configuration to the CLI tests so that they can be run with the `applevz` backend driver in Multipass. Most notably, all `suspend` and `snapshot` tests are skipped. At this point in the git history several tests will fail, but are fixed by other AppleVZ PRs that are currently awaiting review. Yeah for TDD! 😄 

---

MULTI-2528

### DIFF
--- a/tests/cli/cli_vm_lifecycle_test.py
+++ b/tests/cli/cli_vm_lifecycle_test.py
@@ -58,6 +58,7 @@ class TestVmLifecycle:
         assert state(f"{instance}") == "Running"
         assert boot_id_before != get_boot_id(instance)
 
+    @pytest.mark.suspend
     def test_suspend_resume(self, instance):
         assert state(f"{instance}") == "Running"
         boot_id_before = get_boot_id(instance)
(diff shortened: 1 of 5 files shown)


tests/cli/cli_vm_lifecycle_test.py
```diff
@@ -116,6 +117,7 @@ def test_delete_purge(self):
                     assert not output
                     assert "does not exist" in output
 
+    @pytest.mark.suspend
```
xmkg: Can we move this to the assert suspend part? We can do sth like this:

```python
if cfg.driver != "applevz":
    assert multipass("suspend", name1, name2)
    assert state(name1) == "Suspended" and state(name2) == "Suspended"
```

xmkg (changes requested):
Looks good overall, I have a small comment for localizing skip suspend for a test case.

xmkg (approved):
LGTM👍

PR #4785, opened 2026-04-02
[utils] Only seed generator once per thread

### LINKED ISSUE #4796: Unit test `Daemon/LaunchWithBridges.createsNetworkCloudInitIso/` is flaky on CI
**Describe the bug**
```
[----------] 3 tests from Daemon/LaunchWithBridges
[ RUN      ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/0
[       OK ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/0 (7 ms)
[ RUN      ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/1
[       OK ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/1 (9 ms)
[ RUN      ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/2
/Users/runner/work/multipass/multipass/tests/unit/test_daemon.cpp:908: Failure
Actual function call count doesn't match EXPECT_CALL(*mock_factory, prepare_instance_image(_, _))...
         Expected: to be called once

[  FAILED  ] Daemon/LaunchWithBridges.createsNetworkCloudInitIso/2, where GetParam() = ({ ("name=eth0,mode=manual", "", ""), ("name=wlan0", "eth1", "52:54:00:") }, { "eth2" }) (10 ms)
[----------] 3 tests from Daemon/LaunchWithBridges (26 ms total)
```

https://github.com/canonical/multipass/actions/runs/24082041413/job/70250920851

sharder996:
This PR changes the way random generator is seeded for generating MAC addresses. This avoiding issues where generating multiple MAC addresses in quick succession can lack uniqueness. This has caused issues in CI when running unit tests.

https://github.com/canonical/multipass/actions/runs/23876122460/job/69619347352#step:24:5533

Stress testing the unit test in CI and linking a debugger reveals the underlying issue:
```
% lldb -- ./bin/multipass_tests --gtest_filter=Daemon/LaunchWithBridges.createsNetworkCloudInitIso\* --gtest_repeat=-1 --gtest_break_on_failure
...
(lldb) p e.what()
(const char *) 0x0000600002eed018 "Failed to generate an unique mac address after 5 attempts. Number of mac addresses in use: 2"
```

Fixes #4796

### DIFF
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -273,10 +273,9 @@ std::vector<std::string> mp::utils::split(const std::string& string, const std::
 
 std::string mp::utils::generate_mac_address()
 {
-    std::default_random_engine gen;
-    std::uniform_int_distribution<int> dist{0, 255};
+    thread_local std::mt19937 gen{std::random_device{}()};
+    thread_local std::uniform_int_distribution<int> dist{0, 255};
 
-    gen.seed(std::chrono::system_clock::now().time_since_epoch().count());
     std::array<int, 3> octets{{dist(gen), dist(gen), dist(gen)}};
     return fmt::format("52:54:00:{:02x}:{:02x}:{:02x}", octets[0], octets[1], octets[2]);
 }


src/utils/utils.cpp
```diff
@@ -273,10 +273,9 @@ std::vector<std::string> mp::utils::split(const std::string& string, const std::
 
 std::string mp::utils::generate_mac_address()
 {
-    std::default_random_engine gen;
-    std::uniform_int_distribution<int> dist{0, 255};
+    static thread_local std::mt19937 gen{std::random_device{}()};
```
xmkg: thread_local already implies static storage duration

xmkg (commented):
Drive-by minor comment

src/utils/utils.cpp
```diff
@@ -273,10 +273,9 @@ std::vector<std::string> mp::utils::split(const std::string& string, const std::
 
 std::string mp::utils::generate_mac_address()
 {
-    std::default_random_engine gen;
-    std::uniform_int_distribution<int> dist{0, 255};
+    thread_local std::mt19937 gen{std::random_device{}()};
```
ricab, line 276: For my understanding (not saying it's wrong, just want to understand), why did you choose to switch from the `default_random_engine` to `mt19937`?

src/utils/utils.cpp
```diff
@@ -273,10 +273,9 @@ std::vector<std::string> mp::utils::split(const std::string& string, const std::
...
+    static std::uniform_int_distribution<int> dist{0, 255};
```
ricab: Why is one of them local and the other one static?

- If this is used concurrently, there can be a race in the initialization of the distribution
- If it isn't used concurrently, `gen` could be static too.

I think it is a good idea to make this thread safe, but I believe the distribution would need to be `thread_local` too. Alternatively, you could lock around both and have them static.

ricab (commented):
A couple of questions. If you keep the thread-local approach, can you append " per thread" to the PR title and commit msg, please? Thanks.

src/utils/utils.cpp
sharder996: There are both going to work, but using `mt19937` is going to guarantee the algorithm used and the quality of randomness. `default_random_engine` is going to result in the compiler choosing an algorithm which might produce different results on different platforms.

src/utils/utils.cpp
sharder996: Yeah, I think you're correct here. I switched to making them both `thread_local` and avoiding the use of locks.

ricab (approved):
LGTM.

PR #4787, opened 2026-04-02
[docs] AppleVZ docs

sharder996:
This PR adds references in the docs to the new AppleVZ backend driver.

---

MULTI-2529

### DIFF
--- a/docs/explanation/driver.md
+++ b/docs/explanation/driver.md
@@ -14,7 +14,7 @@ On some platforms, it is possible to select a driver during installation. Until 
 Different sets of drivers are available on different platforms:
 
 - On Linux, Multipass can be configured to use QEMU. As of Multipass version 1.16, LXD and libvirt are no longer available.
-- On macOS, the options are QEMU and VirtualBox. As of Multipass version 1.13, Hyperkit is no longer available.
+- On macOS, the options are QEMU, the Apple Virtualization framework, and VirtualBox. As of Multipass version 1.13, Hyperkit is no longer available.
 - On Windows, Multipass uses Hyper-V (only available on Windows Pro) or VirtualBox.
 
 ## Default drivers
--- a/docs/explanation/driver.md
+++ b/docs/explanation/driver.md
@@ -46,9 +46,11 @@ There are two exceptions to the above:
 
 While we strive to offer a uniform interface across the board, not all features are available on all backends and there are some behaviour differences:
 
-| Feature | Only supported on... | Notes |
+| Feature | Not supported on... | Notes |
 |--- | --- | --- |
-| **Native mounts** | <ul><li>Hyper-V</li><li>QEMU</li></ul> | This affects the `--type` option in the [`mount`](/reference/command-line-interface/mount) command). |
+| **Native mounts** | <ul><li>AppleVZ</li><li>VirtualBox</li></ul> | This affects the `--type` option in the [`mount`](/reference/command-line-interface/mount) command). |
+| **Snapshots** | <ul><li>AppleVZ</li></ul> | This affects the [`snapshot`](/reference/command-line-interface/snapshot) command. |
+| **Suspend** | <ul><li>AppleVZ</li></ul> | This affects the [`suspend`](/reference/command-line-interface/suspend) command. |
 
 <!-- old formatting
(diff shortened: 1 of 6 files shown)


docs/how-to-guides/install-multipass.md
```diff
@@ -181,6 +181,12 @@ You've installed Multipass. Time to run your first commands! Use `multipass vers
 
 You've installed Multipass. Time to run your first commands! Use `multipass version` to check your version or `multipass launch` to create your first instance.
 
+Multipass defaults to using QEMU as its virtualisation provider. If you'd like to use AppleVZ, you can do so using the following command:
```
geoffreynyaga, line 184: Just a small fix.
We should use US english  for consistency. I'd recommend we swap `virtualisation` with `virtualization`. This also helps with consistency even when we reference AppleVZ.

Note: US english is now recommended internally https://docs.ubuntu.com/styleguide/en/#spelling and the word 'virtualization' is specifically listed under the "Other common terms" section. The functionality is already implemented in the Multipass repo, but we have not enabled the check on CI (it requires effort to update all pages). Hint: If you run `make run` locally, you should see a `docs/.sphinx/styles/Canonical/000-US-spellcheck.yml` file.

docs/explanation/driver.md
```diff
@@ -14,7 +14,7 @@ On some platforms, it is possible to select a driver during installation. Until
 Different sets of drivers are available on different platforms:
 
 - On Linux, Multipass can be configured to use QEMU. As of Multipass version 1.16, LXD and libvirt are no longer available.
-- On macOS, the options are QEMU and VirtualBox. As of Multipass version 1.13, Hyperkit is no longer available.
+- On macOS, the options are QEMU, the Apple Virtualization framework, and VirtualBox. As of Multipass version 1.13, Hyperkit is no longer available.
```
geoffreynyaga: Minor request. Since we use both the full name and the abbreviation interchangeably throughout the docs, I'd suggest introducing both together at first mention so readers always have the reference: 
```diff
+ the Apple Virtualization framework (AppleVZ), ...
```
Quick question: I know that Apple uses `VZ` as class prefix throughout its codebase. Are `apple-vz`/`AppleVZ` terms that Apple uses officially, or are they Multipass-specific shorthands derived from that prefix? Knowing this will help us write a more accurate explanation in the docs down the line.

docs/how-to-guides/manage-instances/modify-an-instance.md
```diff
@@ -58,7 +58,7 @@ On the other hand, it's always possible to fetch properties for all instances. U
 ```
 
 ```{note}
-Modifying instance settings is not supported when using the Hyperkit driver, which has been deprecated in favour of QEMU. The QEMU and VirtualBox drivers on Intel-based macOS hosts do support instance modification.
+Modifying instance settings is not supported when using the Hyperkit driver, which has been deprecated in favour of QEMU.
```
geoffreynyaga, line 61:
```suggestion
Modifying instance settings is not supported when using the Hyperkit driver, which has been deprecated in favour of QEMU.
> See more: [Migrate from Hyperkit to QEMU on macOS](how-to-guides-customise-multipass-migrate-from-hyperkit-to-qemu-on-macos)
```

docs/explanation/driver.md
sharder996: I have found no references to any other AppleVZ mutations in the Apple docs. Its always referred to by its full moniker.

docs/how-to-guides/install-multipass.md
sharder996: I used `virtualization` explicitly because its part of the name of the framework, but if you want I can find/replace all instances of `virtualisation` with the former.

docs/explanation/driver.md
ricab: Will the setting value be named as applevz? If so, better tie the two together somewhere in the docs (not sure where)

sharder996: It is mentioned in https://github.com/canonical/multipass/blob/ba40c5e91e1f5ca365615ffc12d63587ec971cbb/docs/how-to-guides/customise-multipass/set-up-the-driver.md?plain=1#L35-L48

geoffreynyaga (dismissed):

geoffreynyaga (approved):

PR #4593, opened 2026-01-01
[snap] Migrate snap base to core24

sharder996:
This PR updates the snap base of the Multipass snap from core22 to core24.

Necessary and relevant changes are grouped together in commits with the reasoning explained in the commit body.

MULTI-2431

Issue #3430, opened 2024-03-06
[qemu] UEFI firmware should be loaded via `-pflash` instead of `-bios`


townsend2010:
On Linux, we should be using `pflash` instead of `-bios` for loading the UEFI firmware.  See https://lists.katacontainers.io/pipermail/kata-dev/2021-January/001650.html for more info.

Thanks to @jnsgruk for pointing this out!

jimporter:
This is still valid. @xmkg You're assigned to this, but if you're busy, I could take a look.

xmkg:
> This is still valid. @xmkg You're assigned to this, but if you're busy, I could take a look.

I must have assigned this to myself at some point but it wasn't on my radar for a long time. Feel free to re-assign it as you see fit.

PR #4799, opened 2026-04-07
Move QEMU to vcpkg

xmkg:
# Description

### What does this PR do?
Move QEMU dependency to vcpkg:
  - Move multipass-specific patches to the multipass-patches/ folder
  -  Ensure static linkage of QEMU and QEMU tools by listing them as vcpkg dependencies
  - Prune unneeded QEMU features to reduce both build time and the amount of runtime dependencies:
    - "--disable-xkbcommon"
    -  "--disable-gtk"
    -  "--disable-opengl"
     - "--disable-libudev"
     -  "--disable-af-xdp"
     -  "--disable-slirp"
  - Remove system package reliance at all levels
  - Force Multipass to use vendored QEMU instead of the system one
  - Reduce build disparity between snap and non-snap builds
  - Remove redundant firmware files
    -  keymaps (VNC/GTK keyboard layouts)
    - bios-256k.bin -> We're using UEFI boot
  - Better snap & macOS build times since QEMU is now vcpkg-cached
  - Enable MSYS/MinGW build in Windows to build qemu-img
  - One QEMU recipe for all platforms!
  
### Why is this change needed?
Streamlining dependencies, consistency across platforms, and simplifying QEMU provisioning.

MULTI-2339


.github/workflows/windows-macos.yml
```diff
@@ -162,12 +162,6 @@ jobs:
       run: |
         python3 -m pip install --user --upgrade distlib
 
-    - name: Install specific QEMU from Choco
```
xmkg, line 165: No longer needed since we're bundling our own qemu-img

3rd-party/vcpkg-ports/qemu/configure-wrapper
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+args=""
+for arg in "$@"; do
```
xmkg, line 3: QEMU's configure is not like a traditional `configure` and it fails when some of the `conventional` configure flags are passed by `vcpkg_configure_make()`. This wrapper is here to strip those (we don't need them anyway).

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,269 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO qemu/QEMU
```
xmkg, line 16: Pulling upstream and applying our patches. Ditched the test option patch since we're now able to null out the test's meson file directly, achieving the same outcome.

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,269 @@
...
+    set(QEMU_ARCH ppc64)
+elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "s390x")
+    set(QEMU_ARCH s390x)
+else()
+    message(FATAL_ERROR "Unsupported architecture: ${VCPKG_TARGET_ARCHITECTURE}")
+endif()
+
+file(WRITE "${SOURCE_PATH}/tests/meson.build" "")
+
+if("system" IN_LIST FEATURES)
```
xmkg, line 41: `system` builds the emulator too. Non-system build is tool-only.

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,269 @@
...
+    set(QEMU_TARGET_LIST "${QEMU_ARCH}-softmmu")
+else()
+    set(QEMU_TARGET_LIST "")
+endif()
+
+set(QEMU_COMMON_OPTIONS
+    "--disable-bochs"
```
xmkg, line 48: Derived from the old QEMU CMake + snapcraft.yaml. Has a couple of extra ones at the end, listed in PR description.

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,269 @@
...
+    "--disable-gtk"
+    "--disable-opengl"
+    "--disable-libudev"
+    "--disable-af-xdp"
+    "--disable-slirp"
+    "--disable-curses"
+)
+
+if(VCPKG_TARGET_IS_WINDOWS)
+    ###########################################################################
```
xmkg, line 81: So building QEMU on Windows is not trivial since QEMU does not support MSVC builds. To overcome this, the portfile bootstraps an MSYS/MinGW environment. This requires a bit of setup, but it's a reasonable price to pay to achieve full platform parity.

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,269 @@
...
+    # injected by vcpkg_configure_make.
+    file(RENAME "${SOURCE_PATH}/configure" "${SOURCE_PATH}/configure.real")
+    file(COPY "${CMAKE_CURRENT_LIST_DIR}/configure-wrapper" DESTINATION "${SOURCE_PATH}")
+    file(RENAME "${SOURCE_PATH}/configure-wrapper" "${SOURCE_PATH}/configure")
+    file(CHMOD "${SOURCE_PATH}/configure" PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE)
+
+    set(ENV{PKG_CONFIG_PATH} "${CURRENT_INSTALLED_DIR}/lib/pkgconfig")
+    set(ENV{PKG_CONFIG_LIBDIR} "${CURRENT_INSTALLED_DIR}/lib/pkgconfig")
+
+    vcpkg_configure_make(
```
xmkg, line 194: We're utilizing `vcpkg_configure_make()` here to benefit from the environment isolation of the vcpkg. vcpkg_configure_make() expects a `traditional` configure file -- that's sorted out by the configure-wrapper file.

3rd-party/vcpkg-ports/qemu/vcpkg.json
```diff
@@ -0,0 +1,36 @@
...
+  "version": "10.0.3",
+  "port-version": 0,
+  "description": "QEMU with Canonical's patches for Multipass",
+  "homepage": "https://github.com/qemu/QEMU",
+  "license": "GPL-2.0-only",
+  "supports": "osx | linux | windows",
+  "dependencies": [
+    {
+      "name": "glib",
+      "platform": "!windows"
```
xmkg, line 12: The Windows part of the portfile manages its own dependencies.

snap/snapcraft.yaml
```diff
@@ -211,6 +216,13 @@ parts:
     override-pull: |
       set -e
 
+      if [ ! -d "$HOME/.cargo" ]; then
```
xmkg, line 219: rustc is needed for compiling QEMU's rust-rewritten modules.

snap/snapcraft.yaml
```diff
@@ -240,110 +252,6 @@ parts:
       cd ${CRAFT_PART_SRC}/3rd-party
       find . -type f \( -name 'LICENSE*' -o -path '*/LICENSE*/*' \) -print0 | xargs -0 -I FILE install --mode=644 -D --no-target-directory FILE ${CRAFT_PART_INSTALL}/licenses/FILE
 
-  qemu:
```
xmkg, line 243: We're able to ditch this completely since it's now vcpkg's responsibility to make QEMU available. The CMake install() part handles putting the compiled binaries in the right place.

snap/snapcraft.yaml
```diff
@@ -240,110 +252,6 @@ parts:
...
-    - qemu/keymaps*
-    - qemu/efi-virtio.rom*
-    - qemu/kvmvapic.bin*
-    - qemu/vgabios-*.bin*
-    - qemu/slof.bin*
-    - qemu/bios-256k.bin*
-    - qemu/s390-ccw.img*
-    - -usr/lib/*/pkgconfig
-
-  qemu-firmware:
```
xmkg, line 333: The firmware files are now copied from the QEMU's `pc-bios` folder. The only file lacking there is the OVMF.fd file, which is an older version of the EDK2 firmware. It could be fixed with a follow-up patch where the firmware is split into two: the code and the NVRAM files. Added the OVMF.fd file to the port folder for now, so as not to introduce any unwanted behavior changes.

src/cmake/populate-supported-backends-list.cmake
```diff
@@ -0,0 +1,31 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+if(UNIX)
```
xmkg, line 15: Extracted this for clarity (and also for making the backend list available before the `project()` call).

src/cmake/qemu-img-install-and-copy.cmake
```diff
@@ -0,0 +1,37 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# This module propagates QEMU binaries to wherever they're needed to exist.
+
+find_program(QEMU_IMG qemu-img
```
xmkg, line 17: Adds an install() target for qemu-img. Also copies it to the build/bin folder so local builds can use QEMU binaries without having to invoke the install procedure.

src/cmake/qemu-system-install-and-copy.cmake
```diff
@@ -0,0 +1,53 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# This module propagates QEMU binaries to wherever they're needed to exist.
+
+find_program(QEMU_SYSTEM qemu-system-${HOST_ARCH}
```
xmkg, line 17: The same as qemu-img, but for qemu-system.

src/platform/backends/qemu/qemu_base_process_spec.cpp
```diff
@@ -20,27 +20,29 @@
...
+#include <QCoreApplication>
+#include <QDir>
+
 namespace mp = multipass;
 namespace mpu = multipass::utils;
 
 QString mp::QemuBaseProcessSpec::program() const
 {
-    return QString("qemu-system-%1").arg(HOST_ARCH);
+    return QDir(QCoreApplication::applicationDirPath())
```
xmkg, line 31: So now qemu-system is always right next to the running process's directory, either let it be a snap, a install()ed one, or a local build. This path change reflects that.

src/platform/backends/qemu/qemu_base_process_spec.cpp
```diff
@@ -20,27 +20,29 @@
...
 {
-    try
-    {
-        return mpu::snap_dir().append("/qemu");
-    }
-    catch (const mp::SnapEnvironmentException&)
-    {
-        return QString();
-    }
+    return firmware_path();
```
xmkg, line 37: The firmware files are now in a known location relative to the bin dir, too. Introduced firmware_path() function to make it easier to access.

src/platform/backends/qemu/qemu_virtual_machine_factory.cpp
```diff
@@ -88,8 +89,14 @@ void mp::QemuVirtualMachineFactory::hypervisor_health_check()
 
 QString mp::QemuVirtualMachineFactory::get_backend_version_string() const
 {
-    auto process = mp::platform::make_process(
-        simple_process_spec(QString("qemu-system-%1").arg(HOST_ARCH), {"--version"}));
+    struct QemuVersionProcSpec : QemuBaseProcessSpec
```
xmkg, line 92: Small enough to be local.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -73,8 +74,10 @@ QStringList mp::QemuVMProcessSpec::arguments() const
             QString::number(desc.mem_size.in_megabytes()) + 'M'; /* flooring here; format documented
 in `man qemu-system`, under `-m` option; including suffix to avoid relying on default unit */
         // clang-format off
+        // Tell QEMU to where to look for the BIOS files
+        args << "-L"
```
xmkg, line 78: Now we're passing the expected path for the firmware files at runtime. We no longer hardcode it at build time.

src/platform/backends/qemu/qemu_vm_process_spec.cpp
```diff
@@ -192,7 +194,7 @@ profile %1 flags=(attach_disconnected) {
   /sys/module/vhost/parameters/max_mem_regions r,
 
   # binary and its libs
-  %4/usr/bin/%5 ixr,
+  %5 ixr,
```
xmkg, line 197: Now, the `program()` is the full path to the executable. We no longer need to prefix it with anything.

src/process/qemuimg_process_spec.cpp
```diff
@@ -33,7 +36,7 @@ mp::QemuImgProcessSpec::QemuImgProcessSpec(const QStringList& args,
 
 QString mp::QemuImgProcessSpec::program() const
 {
-    return "qemu-img";
+    return QDir(QCoreApplication::applicationDirPath()).filePath("qemu-img");
```
xmkg, line 39: Same as qemu-system.

tests/unit/qemu/CMakeLists.txt
```diff
@@ -8,17 +8,17 @@ target_sources(multipass_tests
     ${CMAKE_CURRENT_LIST_DIR}/test_qemu_vmstate_process_spec.cpp
 )
 
-add_executable(qemu-img
+add_executable(mock-qemu-img
```
xmkg, line 11: Name collision with qemu-img target declared before. This one is the mock, so the real one gets the "real" name.

BUILD.linux.md
```diff
@@ -58,32 +58,17 @@ First, install Multipass's runtime dependencies. On AMD64 architecture, you can
 ```
 sudo apt update
 sudo apt install libgl1 libpng16-16 libxml2 dnsmasq-base \
-    dnsmasq-utils qemu-utils libslang2 iproute2 iptables \
```
xmkg, line 61: We no longer need to install anything-QEMU now.

CMakeLists.txt
```diff
@@ -17,18 +17,22 @@ cmake_policy(SET CMP0079 NEW) # Allow target_link_libraries() in subdirs
 
 include(feature-flags.cmake)
 include(src/cmake/environment-utils.cmake)
-is_running_in_ci(IS_RUNNING_IN_CI)
+include(CMakeDependentOption)
 
 option(MULTIPASS_VCPKG_BUILD_DEFAULT "Enable or disable building the dependencies in the default configuration, which includes both debug and release variants." OFF)
 option(MULTIPASS_ENABLE_TESTS "Build tests" ON)
+cmake_dependent_option(FORCE_ENABLE_VIRTUALBOX "Enable VirtualBox" ON UNIX OFF)
```
xmkg, line 24: Moved this closer to the other options.

snap/snapcraft.yaml
xmkg: (stolen from one of the @tobe2098's rust PRs)

xmkg:
Unassigned @ricab to better distribute review load.

xmkg:
Let's try round-robin for another time... aaand we have a winner! @tobe2098

snap/snapcraft.yaml
xmkg: This is something that @jimporter intends to do in #3430, so mentioning here for completeness.

snap/snapcraft.yaml
tobe2098: XD nice one. It will not be necessary though, we settled for installing the `rustup` snap instead, and using it to install the default toolchain

tobe2098: But weren't we disabling the rust section? As per the description.

snap/snapcraft.yaml
```diff
@@ -144,7 +142,14 @@ parts:
     - libgtkmm-3.0-dev
     # https://github.com/microsoft/vcpkg/issues/47920
     - libltdl-dev # vcpkg-systemd: libxcrypt requires libltdl-dev from the system package manager
+    - autoconf # libxcrypt
+    - autoconf-archive # libxcrypt
+    - automake # libxcrypt
+    - libtool # libxcrypt
     - python3-pip
+    - bison
```
tobe2098: What package requires bison?

src/cmake/qemu-system-install-and-copy.cmake
tobe2098: This should not be present in Windows, correct?

src/platform/backends/qemu/qemu_base_process_spec.cpp
tobe2098: Is wd used exclusively to find the firmware?

tests/unit/test_qemuimg_process_spec.cpp
```diff
@@ -32,7 +32,7 @@ TEST(TestQemuImgProcessSpec, programCorrect)
 {
     mp::QemuImgProcessSpec spec({}, "");
 
-    EXPECT_EQ(spec.program(), "qemu-img");
+    EXPECT_TRUE(spec.program().endsWith("qemu-img"));
```
tobe2098: Was there an issue with EQ?

CMakeLists.txt
```diff
@@ -315,6 +312,12 @@ else()
   endif()
 endif()
 
+if("qemu" IN_LIST MULTIPASS_BACKENDS)
+  include(src/cmake/qemu-system-install-and-copy.cmake)
+endif()
+
+include(src/cmake/qemu-img-install-and-copy.cmake)
```
tobe2098, line 316: This one is not covered in the same conditionals as above?

vcpkg.json
```diff
@@ -46,6 +46,23 @@
...
                 "gtest"
             ]
+        },
+        "qemu": {
+            "description": "Enable vendored QEMU",
+            "$comment": "Windows only needs qemu-img, others need the qemu system emulator too.",
+            "dependencies": [
+                {
+                    "name": "qemu",
+                    "platform": "windows"
```
tobe2098, line 57: This means `qemu-img` only, correct?

3rd-party/vcpkg-ports/qemu/portfile.cmake
tobe2098: I can see... Is there a better approach? Maybe it is not feasible/practical, but I would suggest one of the following:
1. Use MinGW triplets via vcpkg (add dependencies, use the custom MinGW triplets to build QEMU or even the whole project)? 
2. We find a way to generate the pre-compiled binaries elsewhere (maybe a suggestion for the future for now, I do not think it is feasible in a single PR)
3. Would it be possible to explicitly download the exact .pkg.tar.zst files we need (from https://repo.msys2.org), and then feed them directly to pacman -U (install from local file)? This would give us more control if the versions are up for long enough in the mirror. May be a bit of an issue with dependencies (also have to be manually done).

tobe2098 (commented):
Great work Mustafa! It will be great to have this tied up. 

On testing: suspended instances in `edge` snap do not start up in the `edge/pr4799` snap:
```
Instance shutdown during start: qemu: could not load PC BIOS 'OVMF.fd'
```

CMakeLists.txt
tobe2098: I mean the conditionals in line: 135, where the qemu feature is added to the manifest in backends and WIN32, this one is unconditional

tobe2098:
I would also like @jimporter to get a second superficial look on the cmake code, if possible :)

vcpkg.json
xmkg: Yes, that's correct.

CMakeLists.txt
xmkg: Yeah, it is intentional. All backends (namely Hyper-V, QEMU, VirtualBox, and AppleVZ) use qemu-img for image conversion, hence no gating.

snap/snapcraft.yaml
xmkg: I think I forgot to update the description. I thought it wouldn't make any sense to disable Rust for our builds since we're going to have a Rust compiler in our environments anyway.

(EDIT: Updated the description)

src/cmake/qemu-system-install-and-copy.cmake
xmkg: Yes, that's correct. We're not building the system emulator on Windows right now.

CMakeLists.txt
tobe2098: Should it then not be the same in lines 135-137, since QEMU is used in all backends?

src/platform/backends/qemu/qemu_base_process_spec.cpp
xmkg: No -- we're explicitly passing "-L" flag to qemu-system-* to tell where to look for the firmwares. The working directory for QEMU is mostly irrelevant now.

tests/unit/test_qemuimg_process_spec.cpp
xmkg: Mostly laziness on my end. I'll add the full path :)

snap/snapcraft.yaml
```diff
@@ -144,7 +142,13 @@ parts:
     - libgtkmm-3.0-dev
     # https://github.com/microsoft/vcpkg/issues/47920
     - libltdl-dev # vcpkg-systemd: libxcrypt requires libltdl-dev from the system package manager
+    - autoconf # libxcrypt
+    - autoconf-archive # libxcrypt
+    - automake # libxcrypt
+    - libtool # libxcrypt
     - python3-pip
+    - bison # qtbase[dbus] -> glib2 -> gettext[tools] -> bison
```
xmkg, line 158: Added a comment to note who actually needs `bison`.

CMakeLists.txt
xmkg: I think so, yeah. As of now, all our backends depend on QEMU one way or the other. I'll update the code.

snap/snapcraft.yaml
xmkg: It turns out that we've already introduced an unintended behavior change with #4593; the core24 features a 4MB OVMF file, which has a different layout than the core22's 2MB file. Now I think it would be better to migrate to the split-fw file approach in this branch since the regression is already there.

xmkg: Committed https://github.com/canonical/multipass/commit/d5355cea465f4ae004444ef16b2cc42a5d40ef94 to fix this.

xmkg:
Now it passes the version upgrade tests when the other regression fixes are merged on top (https://github.com/canonical/multipass/pull/5017). Windows/macOS version upgrade tests are still to-do.

src/process/qemuimg_process_spec.cpp
```diff
@@ -33,7 +36,7 @@ mp::QemuImgProcessSpec::QemuImgProcessSpec(const QStringList& args,
...
 }
```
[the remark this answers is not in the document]
xmkg: This is bogus. QT resolves it just fine.

xmkg:
> Great work Mustafa! It will be great to have this tied up.
> 
> On testing: suspended instances in `edge` snap do not start up in the `edge/pr4799` snap:
> 
> ```
> start failed: The following errors occurred:
> Instance shutdown during start: qemu: could not load PC BIOS 'OVMF.fd'
> ```

@tobe2098 this is now fixed. It turned out this regression had already been introduced with the snap core24 migration.

3rd-party/vcpkg-ports/qemu/multipass-patches/0001-9p-uid-gid-mapping-support.patch
```diff
@@ -0,0 +1,302 @@
+From b2f6703ac694045d43793bda14f7e7e08768799e Mon Sep 17 00:00:00 2001
+From: Chris Townsend <christopher.townsend@canonical.com>
+Date: Mon, 8 Aug 2022 16:19:54 -0400
```
jimporter, line 3: Just so I understand what's happening here: where did these patches live prior to this PR? From the diff, they look new, but the header in this file shows they've been around for a long time.

3rd-party/vcpkg-ports/qemu/portfile.cmake
xmkg: Okay, now this has been fixed. The build is fully hermetic now. The solution is to introduce a lockfile for the MSYS packages. The code also now uses vcpkg_acquire_msys + DIRECT_PACKAGES, so vcpkg will be able to cache these.

3rd-party/vcpkg-ports/qemu/generate-mingw-lockfile.py
```diff
@@ -0,0 +1,185 @@
...
+LOCKFILE = Path(__file__).with_name("mingw64-toolchain.lock")
+
+
+def strip_constraint(dep: str) -> str:
+    """Drop any version constraint (`>=`, `=`, `<`, ...) from a dependency."""
+    for sep in (">=", "<=", "==", "=", ">", "<"):
+        idx = dep.find(sep)
+        if idx != -1:
+            return dep[:idx]
+    return dep
```
jimporter: Does Pacman support both `=` and `==` in its version requirements? If not, we should remove the unnecessary one. We could also simplify this with a regex, e.g.:

```suggestion
    return re.sub("(>=|<=|==|=|>|<).*", "", dep)
```

Or even:

```suggestion
    return re.sub("(>|<|=).*", "", dep)
```

Though the latter is a bit less explicit about the operators we look for.

3rd-party/vcpkg-ports/qemu/generate-mingw-lockfile.py
```diff
@@ -0,0 +1,185 @@
...
+def fetch(url: str) -> bytes:
+    with _urlopen(url) as resp:
+        return resp.read()
+
+
+def parse_db(db_bytes: bytes) -> dict[str, dict]:
+    """Parse a pacman sync database (a zstd-compressed tar of `desc` files)."""
+    raw = zstandard.ZstdDecompressor().decompress(db_bytes, max_output_size=512 * 1024 * 1024)
+    packages: dict[str, dict] = {}
+    with tarfile.open(fileobj=io.BytesIO(raw), mode="r:") as tar:
```
jimporter: Would something like this work? That would remove the extra `BytesIO` indirection and probably be a bit more memory-efficient as well (though I doubt the perf matters much here):

```suggestion
    packages: dict[str, dict] = {}
    with zstandard.ZstdDecompressor().open(db_bytes, max_output_size=512 * 1024 * 1024) as zstd, \
         tarfile.open(fileobj=zstd, mode="r:") as tar:
```

If we require Python 3.14, we could simplify even further:

```suggestion
    packages: dict[str, dict] = {}
    with tarfile.open(fileobj=io.BytesIO(raw), mode="r:zst") as tar:
```

3rd-party/vcpkg-ports/qemu/generate-mingw-lockfile.py
```diff
@@ -0,0 +1,185 @@
...
+            packages[name] = {
+                "filename": fields["FILENAME"][0],
+                "depends": [strip_constraint(d) for d in fields.get("DEPENDS", [])],
+                "provides": [strip_constraint(p) for p in fields.get("PROVIDES", [])],
+            }
+    return packages
+
+
+def resolve_closure(packages: dict[str, dict], roots: list[str]) -> list[str]:
+    provides = {}
```
jimporter: I imagine this would need a type annotation if we actually ran a type checker on our Python code...

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -0,0 +1,297 @@
...
+)
+
+if(VCPKG_TARGET_IS_WINDOWS)
+    ###########################################################################
+    # Windows: build only qemu-img (the sole QEMU tool Multipass uses here; no
+    # system emulator). The GCC/MinGW toolchain is acquired hermetically via
+    # vcpkg_acquire_msys() from packages pinned by URL + SHA512 -- vcpkg's own
+    # pinned msys shell layer plus the MinGW closure in mingw64-toolchain.lock
+    # (regenerate with generate-mingw-lockfile.py).
+    ###########################################################################
```
jimporter, line 89: As @tobe2098 commented previously, this seems pretty gnarly. I don't know how much flexibility we have here if we want to use vcpkg though. I know we've talked about migrating to Conan, so maybe some of this complexity would go away if/when we do that?

jimporter (changes requested):
I mostly looked over the CMake changes here as well as some of the MinGW support code. It would indeed be nice to get Qemu into vcpkg, but it seems that this comes with a lot of complexity (as least on Windows)...

3rd-party/vcpkg-ports/qemu/multipass-patches/0001-9p-uid-gid-mapping-support.patch
xmkg: They lived under this repository: https://github.com/canonical/qemu/commits/v10.0.3%2B9p-uid-gid-map/

3rd-party/vcpkg-ports/qemu/portfile.cmake
xmkg: Indeed, and unfortunately, we don't have much flexibility in this part. The problem is that QEMU's build is closely bound to UNIX tooling, and QEMU does not support MSVC builds. The only supported Windows compiler is GCC via MinGW-w64. Hence, the gnarly stuff needs to exist if we want to build anything from QEMU. A Conan recipe would have to do the same thing, just in Python code. A small advantage of Conan is that the lockfile generation code can be inlined into the recipe itself, but other than that, I don't see any reduction in complexity.

jimporter: I've used Conan a bit, and my hope would be that we could invoke Conan with the MinGW toolchain for qemu and then just have it return a pkg-config file (or something along those lines) that we can use in the right places in our build. I haven't looked into whether Conan supports per-dependency toolchain overrides like that, but if we could wire that up, then some of this code might go away. That's all hypothetical though and would require setting up Conan for the project before we could even try it out.

(Most of my Conan knowledge comes from hooking it into a meta-dependency-manager, and I did particularly like that it's very agnostic about the build system you or your dependencies use, as well as the way you feed those dependencies into your build. There are a lot of options on both sides that make it super-flexible. Hopefully something like that would work for toolchain overrides too.)

packaging/gui-less/snap/snapcraft.yaml
```diff
@@ -351,5 +257,24 @@ parts:
   glue:
     plugin: dump
     source: https://github.com/canonical/multipass.git
+
+# The Flutter GUI libraries depend on GUI components which are provided by the GNOME
+# extension. The extension is attached at runtime, so the linter incorrectly flags
+# the libraries as "missing dependencies".
     source-type: git
     source-subdir: snap-wrappers
+lint:
```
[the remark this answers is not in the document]
xmkg: bogus

3rd-party/vcpkg-ports/qemu/generate-mingw-lockfile.py
xmkg: Done

jimporter (approved):

jimporter:
Looks good to me on the CMake parts of this PR!

(edit: Oops, misclicked the comment button and temporarily closed this PR.)

src/cmake/populate-supported-backends-list.cmake
```diff
@@ -0,0 +1,36 @@
...
+    endif()
+
+    if(APPLEVZ_ENABLED)
+        list(APPEND MULTIPASS_BACKENDS applevz)
+    endif()
+elseif(WIN32)
+    set(MULTIPASS_BACKENDS hyperv virtualbox)
+
+    if(HYPERV_HCS_ENABLED)
+        add_definitions(-DHYPERV_HCS_ENABLED=1)
```
tobe2098, line 29: Is it possible this definition is redundant? Either that or APPLEVZ_ENABLED is missing a definition

tobe2098 (approved):
Great work Mustafa! This should simplify builds significantly. I have an unrelated question but we can merge 🎉

src/cmake/populate-supported-backends-list.cmake
xmkg: It is needed since we need code-level visibility for disabling listing the Hyper-V HCS backend in the platform code. 

https://github.com/canonical/multipass/blob/3ec6fe6242c8a27f2716016ced2e39c06d24caea/src/platform/platform_win.cpp#L31-L34

https://github.com/canonical/multipass/blob/3ec6fe6242c8a27f2716016ced2e39c06d24caea/src/platform/platform_win.cpp#L806-L817

AZs do the same thing, too:

```cmake
feature_flag(AVAILABILITY_ZONES_ENABLED "Availability zone support")
if(AVAILABILITY_ZONES_ENABLED)
  add_compile_definitions(AVAILABILITY_ZONES_FEATURE)
endif()

```

PR #4805, opened 2026-04-09
[cli-tests] Add missing import (vm_lifecycle)

xmkg:
There's a missing import in the cli_vm_lifecycle_test.py, which makes one of the tests fail. This patch fixes that.


sharder996 (approved):
In case anyone is looking for the offending test:

```
[1](https://github.com/canonical/multipass/actions/runs/24200102218/job/70647648880#step:16:393)
    ❌ test_lifecycle_multiple
  	NameError: name 'cfg' is not defined
```

PR #4810, opened 2026-04-10
[github-actions] Split Windows/MacOS workflows

xmkg:
The workflows are split for better organization, control, and clarity.

# Description

- What does this PR do?
Splits the Windows macOS workflow into macOS and Windows.
- Why is this change needed?
To remove the need for if this and that checks, and have more granular control over workflows

This PR splits the workflows. No additional code is being introduced.

MULTI-2534


tobe2098 (approved):
LGTM!

PR #4813, opened 2026-04-10
[vm-image-vault] Normalize image path before replace

xmkg:
The replace_all call assumes the path is single forward slashes, where Windows makes no such guarantees. Call generic_string() to grab a POSIX-like normalized path so it's always forward slashes.

MULTI-2563

### DIFF
--- a/src/daemon/default_vm_image_vault.cpp
+++ b/src/daemon/default_vm_image_vault.cpp
@@ -550,16 +550,24 @@ void mp::DefaultVMImageVault::clone(const std::string& source_instance_name,
                                  " already exists in the image records");
     }
 
+    if (source_instance_name == destination_instance_name)
+    {
+        throw std::runtime_error{
+            "Destination instance name could not be the same with the source."};
+    }
+
     auto& dest_vault_record = instance_image_records[destination_instance_name] =
         instance_image_records[source_instance_name];
 
     // string replacement is "instances/<src_name>"->"instances/<dest_name>" instead of
     // "<src_name>"->"<dest_name>", because the second one might match other substrings of the
     // metadata.
-    auto image_path = dest_vault_record.image.image_path.string();
+    // The path might have mixed slashes \\ / in Windows. Normalize it before replace.
+    auto image_path = dest_vault_record.image.image_path.generic_string();
     boost::replace_all(image_path,
                        "instances/" + source_instance_name,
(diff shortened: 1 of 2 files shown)


src/daemon/default_vm_image_vault.cpp
```diff
@@ -550,17 +550,28 @@ void mp::DefaultVMImageVault::clone(const std::string& source_instance_name,
...
+    {
+        throw std::runtime_error{"Path replace for the cloned image failed!"};
+    }
+
+    // Update the instance_image_records after path replace succeeds.
+    instance_image_records[destination_instance_name] = [dst = source_vault_record,
+                                                         &image_path]() mutable {
+        dst.image.image_path = image_path;
+        return dst;
+    }();
```
jimporter: I took another look at Boost.StringAlgo and found that there was a possibly-more-convenient function we could use here: `boost::replace_all_copy`. That makes a copy of the input, does the replacement, and returns the result. I tried using that and came up with the following:


```suggestion
    auto dest_vault_record = instance_image_records[source_instance_name];

    // string replacement is "instances/<src_name>"->"instances/<dest_name>" instead of
    // "<src_name>"->"<dest_name>", because the second one might match other substrings of the
    // metadata.
    auto image_path = dest_vault_record.image.image_path.generic_string();
    dest_vault_record.image.image_path =
        boost::replace_all_copy(image_path,
                                "instances/" + source_instance_name,
                                "instances/" + destination_instance_name);

    if (dest_vault_record.image.image_path.generic_string() == image_path)
        throw std::runtime_error{"Path replace for the cloned image failed!"};

    instance_image_records[destination_instance_name] = dest_vault_record;
```

Maybe this is simpler, since it avoids the need for a const-reference and a lambda. What do you think?

jimporter (changes requested):
Good catch! The difference between `.string()` and `.generic_string()` is definitely something to be careful about with the standard `path` type. One comment below; I don't feel *super* strongly about it, but using `boost::replace_all_copy` seems like it lets us simplify this code a bit.

src/daemon/default_vm_image_vault.cpp
xmkg: Sound good to me. I've made the changes.

jimporter (approved):
LGTM!

PR #4500, opened 2025-11-06
Add troubleshooting link for network launch errors.

chikle9090:
This pull request adds a helpful message when a Multipass instance launch fails due to network-related issues.

When the launch fails (e.g., firewall blocking the network), users will see a link to the official troubleshooting documentation:

https://multipass.run/docs/troubleshoot-networking

This improves the user experience by guiding users directly to the firewall troubleshooting instructions without requiring them to search the documentation.

Related issue: #533

PR #4518, opened 2025-11-20
[docs] Add troubleshooting link for launch errors

sharder996:
This pull request adds a helpful message when a Multipass instance launch fails due to network-related issues.

When the launch fails (e.g., firewall blocking the network), users will see a link to the official troubleshooting documentation:

https://multipass.run/docs/troubleshoot-networking

This improves the user experience by guiding users directly to the firewall troubleshooting instructions without requiring them to search the documentation.

Closes #4500

PR #4821, opened 2026-04-14
[cli] Hint at troubleshooting docs on timeout

### LINKED ISSUE #533: Suggest opening up firewall when launching times out
We've seen it a couple times already - when we're waiting too long for an instance to get an IP, we could print a message letting the user know that a firewall might be the issue and point them at some docs to help troubleshoot.

ShadiBahaa:
## Summary

When `multipass launch` times out waiting for an instance to become ready (e.g. due to a firewall blocking DHCP, or network misconfiguration), the error message provides no guidance on what to try next.

This adds a link to the troubleshooting documentation when the launch failure indicates a timeout, giving users a starting point to diagnose the issue.

**Before:**
```
test-vm: timed out waiting for response
```

**After:**
```
test-vm: timed out waiting for response
For troubleshooting, see https://documentation.ubuntu.com/multipass/stable/how-to-guides/troubleshoot/
```

Based on reviewer feedback from #4500:
- Does not mention any specific cause (multiple causes are possible)
- Points to the general troubleshooting page rather than a specific sub-page
- Uses a single string construction

## Test plan

- [x] Verify that launch timeouts now include the troubleshooting link in the error output
- [x] Verify that non-timeout launch failures (invalid disk size, invalid hostname, etc.) are unaffected
- [x] Verify that the `INVALID_NETWORK` error still shows its own troubleshooting link

Closes #533

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -594,6 +594,13 @@ mp::ReturnCodeVariant cmd::Launch::request_launch(const ArgParser* parser)
             }
         }
 
+        if (error_details.empty() && status.error_message().find("timed out") != std::string::npos)
+        {
+            error_details =
+                "For troubleshooting, see "
+                "https://documentation.ubuntu.com/multipass/stable/how-to-guides/troubleshoot/";
+        }
+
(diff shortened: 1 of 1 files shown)


xmkg:
The docs are updated in #4518.

PR #4822, opened 2026-04-14
[snap] GUI-less snap Launchpad integration

sharder996:
This PR makes the following changes in order to automate the building of a gui-less Multipass snap in Launchpad for the s390x and ppc64 architectures:

1. Additional supported architectures added to `snap/snapcraft.yaml`. They have been added here in order to minimum the number of edits made to the gui-less snapcraft file.
2. Mirrored version of `snap/snapcraft.yaml` without GUI located at `packaging/gui-less/snap/snapcraft.yaml`. Because pathing implications and assumptions made by snapcraft, the source of some parts are now the GitHub remote instead of your local development clone.
3. Python script to programmatically generate the mirrored snapcraft file. This is where edits to the gui-less snapcraft build should be defined.
4. CI step added to the Lint workflow which verifies that any edits made to the original snapcraft file are properly manifested in the gui-less version.
5. Optional pre-commit hook that regenerates the mirrored snapcraft file if `snap/snapcraft.yaml` has been edited.

The Launchpad side can be observed [here](https://launchpad.net/~multipass-team/multipass/+snap/multipass-gui-less). Once this is merged in, the snapcraft recipe in Launchpad will be modified to track the `main` branch, and additional recipes added to mirror the `candidate` and `beta` recipes, except without the GUI and on the s390x/ppc64 architectures.

---

MULTI-2379


snap/snapcraft.yaml
```diff
@@ -16,6 +16,8 @@ base: core24
 platforms:
   arm64:
   amd64:
+  s390x:
```
xmkg: s390x & ppc64el can't build the GUI, no? If so, would it make sense to move them to generate-snapcraft.py instead?

packaging/gui-less/generate-snapcraft.py
```diff
@@ -0,0 +1,100 @@
...
+OUTPUT = SCRIPT_DIR / "snap/snapcraft.yaml"
+
+MULTIPASS_SOURCE = "https://github.com/canonical/multipass.git"
+HEADER = (
+    "Generated by packaging/gui-less/generate-snapcraft.py — do not edit directly.\n"
+    "# Edit snap/snapcraft.yaml and re-run the script."
+)
+
+
+def add_flutter_gui_off(override_pull: str) -> str:
```
xmkg: This could be a cmake preset inheriting from "snap". That way, we can avoid the regex search altogether.

packaging/gui-less/generate-snapcraft.py
```diff
@@ -0,0 +1,100 @@
...
+    yaml.preserve_quotes = True
+    yaml.width = 4096
+
+    data = yaml.load(input_path)
+
+    # Remove the GUI app
+    del data["apps"]["gui"]
+
+    # Change multipass part to use remote git source
+    data["parts"]["multipass"]["source"] = MULTIPASS_SOURCE
```
xmkg, line 46: What's the reason for this? Would it mean a local build from this snapcraft file would fetch the Multipass from upstream, ignoring local changes?

packaging/gui-less/generate-snapcraft.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    # Change glue part to clone snap-wrappers from remote git source
+    glue = data["parts"]["glue"]
+    glue["source"] = MULTIPASS_SOURCE
+    glue["source-type"] = "git"
+    glue["source-subdir"] = "snap-wrappers"
+
+    with output_path.open("w") as f:
+        f.write(f"# {HEADER}\n")
```
xmkg: minor, but you can probably use this:

```
data.yaml_set_start_comment(HEADER)
```

xmkg (changes requested):

packaging/gui-less/generate-snapcraft.py
sharder996: Yeah, you understand correctly. I have it setup like this because when snapcraft executes from `packaging/gui-less/`, it mounts the CWD into the LXD instance meaning that it doesn't have access to any directories above `packaging/gui-less` (notably omitting `src` and `snap-wrappers`.

Symlinks to those directories don't work either, because LXD doesn't follow symlinks when mounting a directories (symlinks resolve to a path outside the mount causing errors) and we can't move the snapcraft file up because it conflicts with the original snapcraft file.

This is the compromise I came up with, but I would welcome other suggestions.

xmkg: So I locally tried a couple of approaches to tackle this. source: ../../ won't work because snapcraft apparently only copies the root directory to the snap/snapcraft.yaml. That's why they recommend putting snap/snapcraft.yaml at the root of the project.

There are workarounds but none is cleaner than what you have here already.

packaging/gui-less/generate-snapcraft.py
```diff
@@ -0,0 +1,92 @@
...
+    data["parts"]["multipass"]["source"] = MULTIPASS_SOURCE
+
+    # Use a CMake preset to turn off the GUI
+    data["parts"]["multipass"]["build-environment"].append(
+        {"CMAKE_PRESET": "snap-gui-less"}
+    )
+
+    # Change glue part to clone snap-wrappers from remote git source
+    glue = data["parts"]["glue"]
+    glue["source"] = MULTIPASS_SOURCE
```
xmkg, line 55: This would always pull `main` regardless of the branch we're building from. It wouldn't make much sense to set "source-tag" at the generation time so is it somehow possible to deduce this dynamically in the snapcraft.yaml file we're building?

xmkg (changes requested):

packaging/gui-less/generate-snapcraft.py
sharder996: I can see where having it always pull from `main` would cause a developer to make local changes for development purposes that would never end up getting committed. Or, produce a "gotcha" moment.

Can you give me an example of what you would achieve by dynamically setting the `source-tag`?

sharder996: I think I know what you're getting at, but in the end this is dev tooling. Not something that we are planning to run on an automated basis. So, if we were to dynamically set the branch, I assume that would mean that a developer would have different behaviour on their local dev branch versus on the `main` branch. We would have to be careful that these generated changes wouldn't be pushed to `main`.

xmkg: I was thinking more of what if we have to build an automated build of this from a non-main branch -- it wouldn't be immediately obvious that this branch is silently pulling the sources from main. 

I did a quick search, and it seems like snapcraft [does not support](https://forum.snapcraft.io/t/source-tag-or-source-commit-via-variable/35033) this use case. I guess this is something to keep in mind.

xmkg (approved):
Alright, this is as good as it gets. Thanks, @sharder996!

PR #4826, opened 2026-04-15
Cli rename instance

### LINKED ISSUE #255: Rename instance
It would be nice to be able to rename an existing instance.

ShadiBahaa:
# Description

- **What does this PR do?** Adds a `multipass rename <current_name> <new_name>` command that allows renaming an existing stopped instance across all layers: proto definition, image vault, daemon logic, CLI, and bash completions.
- **Why is this change needed?** Users currently have no way to rename an instance after creation, forcing them to delete and recreate it to change the name.

## Related Issue(s)

Closes #255

## Testing

- Unit tests
- Manual testing steps:

  1. Launch an instance: `multipass launch --name foo`
  2. Stop it: `multipass stop foo`
  3. Rename it: `multipass rename foo bar`
  4. Verify with `multipass list` that the instance now appears as `bar`
  5. Start it: `multipass start bar`

N/A — CLI-only change.

## Checklist

## Additional Notes

The rename operation requires the instance to be stopped, consistent with how `clone` and `snapshot` work. Renaming to the same name is a no-op. The CLI also updates any aliases that reference the old instance name.

### DIFF
--- a/completions/bash/multipass
+++ b/completions/bash/multipass
@@ -229,8 +229,8 @@ _multipass_complete()
     cmd="${COMP_WORDS[1]}"
     prev_opts=false
     multipass_cmds="authenticate transfer delete exec find help info launch list mount networks \
-                    purge recover restart restore shell snapshot start stop suspend umount version get set \
-                    clone alias aliases unalias"
+                    purge recover rename restart restore shell snapshot start stop suspend umount version \
+                    get set clone alias aliases unalias"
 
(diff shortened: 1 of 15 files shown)


ricab (changes requested):
Hi @ShadiBahaa, we have a spec written and approved internally for how to do this. The approach is very different, both at the interface (command, gui) and implementation (e.g. moving to ID-based vault) levels. 

There is a lot more to it and I am afraid this doesn't match our goals.

ricab:
BTW, you can achieve a similar effect today with clone:

```
$ multipass clone old new
$ multipass delete -p old
```

It takes a while longer to copy and needs the extra disk space, but the end result is the same.

PR #4343, opened 2025-09-05
[ci] Enable coredump upload for macOS

xmkg:
Mostly CI improvements as a result of 2-day troubleshooting macos-15 crash session:

- ~~Enable coredump generation for multipass_tests by codesigning it with a proper entitlement~~
- ~~Enable coredump generation at kernel level by setting `kern.coredump` and `kern.corefile`~~
- ~~Upload core dump + test executable to GitHub Artifacts on segv~~
- Enable tmate session when test execution fails with segv
- ~~Add protobuf version validation macro to the tests/main.cpp~~
- ~~Get rid of the generate_grpc_cpp, use protobuf_generate function~~
- ~~Pin macos-15 xcode version to 16.0~~

Signed-off-by: Mustafa Kemal Gilor <mustafa.gilor@canonical.com>

PR #4829, opened 2026-04-16
[wf-macos] Enable core dump upload on SEGV

xmkg:
Extracted from https://github.com/canonical/multipass/pull/4343


jimporter (approved):
LGTM! The `tmate` stuff from the other PR looks useful, but that's a little riskier. This change should be totally safe, though.

PR #4841, opened 2026-04-20
Fix daemon rpc client logger lifetime issue

xmkg:
# Description

Implementation for #4818.

### DIFF
--- a/src/daemon/daemon_rpc.cpp
+++ b/src/daemon/daemon_rpc.cpp
@@ -102,11 +102,13 @@ grpc::Status emit_signal_and_wait_for_result(OperationSignal operation_signal,
             return mpl::Level::info;
     }();
 
-    multipass::DaemonRpcContextImpl<T, U> ctx{server, level, mpx};
+    std::promise<grpc::Status> promise;
+    auto future = promise.get_future();
+    multipass::DaemonRpcContextImpl<T, U> ctx{promise, server, level, mpx};
     emit operation_signal(request,
                           static_cast<grpc::ServerReaderWriter<T, U>*>(server),
                           static_cast<multipass::DaemonRpcContext*>(&ctx));
-    return ctx.get_future().get();
+    return future.get();
 }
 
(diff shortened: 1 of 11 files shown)


src/daemon/daemon_rpc.cpp
```diff
@@ -85,14 +86,29 @@ auto server_socket_type_for(const std::string& server_address)
...
+                                             mpl::MultiplexingLogger& mpx)
 {
-    std::promise<grpc::Status> status_promise;
-    auto status_future = status_promise.get_future();
-    emit operation_signal(&status_promise);
+    auto level = [&request]() {
+        if constexpr (HasVerbosityLevel<U>)
+            return mpl::level_from(request->verbosity_level());
+        else
+            return mpl::Level::info;
```
tobe2098: Should the default not be level error?

tobe2098 (changes requested):
LGTM overall, just a small nitpick. Elegant solution, and it also fixes the missing logs problem from Client Loggers.

tobe2098:
You may want to rebase the branch, it still suffers of the YAML spec bug

src/daemon/daemon_rpc.cpp
xmkg: For logging, info sounds like a better default to me as it does not imply anything about the content of the message. I'd love to hear your reasoning, though since I'm not married to this by any means.

tobe2098: Maybe it could be argued that that is the case, but I would ask @amylily1011 for her opinion, the CLI guidelines may recommend a particular default. 

WDYT @amylily1011?

tobe2098 (commented):
LGTM, good work! The only thing remaining is the default logging level for the client

src/daemon/daemon_rpc.cpp
xmkg: Discussed out-of-band.

tobe2098 (approved):
Approved on my end 🎉

PR #4854, opened 2026-04-29
fix: warn when launch timeout skips mounts

### LINKED ISSUE #4831: [launch] Mount options silently dropped on timeout
**Describe the bug**
The mount options are silently dropped on timeout

**To Reproduce**
How, and what happened?
1. `multipass launch -n a --timeout 5 --mount ~/dir:/mnt/dir`
2. Error does not mention the missed mounts, even if the instance eventually reaches the running state.

**Expected behavior**
Multipass should clarify to the user that the mount flags are dropped if the mount initialization part is not reached.

**Additional info**
- OS: Ubuntu 24.04

NiShITa-code:
# Description

- Updates the `launch --timeout` message when `--mount` options are pending.
- If launch times out before mount setup is reached, the user is now told that the requested mounts were not attempted.
- This is needed because `--mount` routes are applied only after `request_launch()` completes successfully. When the client-side timeout fires first, the previous message only said launch timed out and did not mention the skipped mount setup.

## Related Issue(s)

Closes #4831

## Checklist

## Additional Notes

I noticed #4663 adds broader message handling for daemon/client replies. For this issue, I kept the change in the client launch timeout path because the timeout is triggered before the later mount phase is reached.

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -107,6 +107,37 @@ auto net_digest(const QString& options)
 
     return net;
 }
+
+std::string launch_timeout_message(const QString& instance_name,
+                                   const std::vector<std::pair<QString, QString>>& mount_routes)
+{
+    auto message = std::string{"Warning: Timeout while launching the instance."};
+
+    if (!mount_routes.empty())
+    {
+        const auto mount_request_count = mount_routes.size();
+        const auto plural_suffix = mount_request_count == 1 ? "" : "s";
+        message += fmt::format(" No mounts have been performed out of {} mount request{}.\n\n",
+                               mount_request_count,
+                               plural_suffix);
+        message += fmt::format("Run the command{} below to apply manually:\n", plural_suffix);
+
+        for (const auto& [source, target] : mount_routes)
+        {
+            message += "\n multipass mount ";
+            message += source.toStdString();
+            message += " ";
+            message += instance_name.toStdString();
+            if (!target.isEmpty())
+            {
+                message += ":";
+                message += target.toStdString();
+            }
+        }
+    }
+
+    return message;
+}
 } // namespace
 
 mp::ReturnCodeVariant cmd::Launch::run(mp::ArgParser* parser)
@@ -500,7 +531,7 @@ mp::ReturnCodeVariant cmd::Launch::request_launch(const ArgParser* parser)
         timer = cmd::make_timer(parser->value("timeout").toInt(),
                                 spinner.get(),
                                 cerr,
-                                "Timed out waiting for instance launch.");
+                                launch_timeout_message(instance_name, mount_routes));
         timer->start();
     }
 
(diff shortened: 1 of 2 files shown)


tobe2098 (changes requested):
Thank you for the contribution! I would like to point out that if there are multiple mounts, some mounts may have succeeded while others have not in the start-up process, the message should reflect that. What do you think @amylily1011 ?

NiShITa-code:
Thanks, that makes sense. I updated the warning so it no longer says that all requested mounts were not attempted. It now says that one or more requested mounts may not have completed before launch timed out, and includes the requested mount paths so users can see which mounts are affected by the timeout uncertainty.

amylily1011:
Thanks @NiShITa-code for your contribution. I think it's a great addition. @tobe2098 I'm wondering how this message might affect our plans, if we were to have exit codes for timeout as a solution (discussed in our CLI spec), which comes with mount warning from timeouts. For now, this seems like a good quick fix.

tobe2098:
Hi @NiShITa-code, there is some further work that we would like to add on top of this in the future, do you mind if we merge your contribution once that work is finished?

amylily1011:
HI @NiShITa-code, 
After discussing with the team, since we already have an initial error message drafted for this fix and we plan to introduce exit code for this warning. We have decided to separate those two tasks and would like you to fix this error/warning message. 

1. Can you show the count of unmounted file in the error message?
2. The message would look like below:
```
Warning: Timeout while launching the instance. No mounts have been performed out of [count] mount requests. 

Run the command below to apply manually: 
    
 multipass mount ~/projects <vm-name>:/projects 
```

NiShITa-code:
I added the count using the number of requested mounts, so the warning now says how many mount requests were affected by the timeout.

ricab:
Hi @amylily1011, I believe that, when any mounts were requested, we decided to limit the message to "Warning: Timeout while launching the instance. No mounts have been performed.". The user just issued the command, they know what they asked. I think the extra count phrase is too verbose for what it's worth.

amylily1011:
Hi @ricab, sorry for my late reply, I've went back to check our meeting notes for May 29 when we were discussing this and it looks like the action point says we should include the count in the message. I totally understood your point about verbosity and I think the count might add a small addition to the text but it might make the message clearer. When I compared the two messages, value of the count is that it can be a subtle confirmation of their mount counts. Which might not be as clear as listing out the whole list of files/folders, but it provides more hints than not showing, IMO. 

Both decisions would pose as minor correction, but I think we can unblock this PR for now and see if the response about count is more confusing or not.

ricab:
Hi @amylily1011, I was pretty sure in the end we said just "no mounts were performed", but I guess we did not update that detail in the notes. Maybe I misremember, I'd be curious to hear others' recollection.

On the substance: IMO that subtle confirmation risks being being perceived as a redundant annoyance. Perhaps a "subtle drag" in an interface that we want to be concise and slick. It's subjective, but the drive to simplify code and tests unties it for me.

In any case, nothing is blocked on my end here.

tobe2098:
Hi @NiShITa-code, could you implement the concluded changes?

tobe2098:
Hi @NiShITa-code, pinging again if you are still interested in completing this PR

PR #4857, opened 2026-04-30
[sshfs] Sshfs path normalization

tobe2098:
# Description

- What does this PR do? It resolves symlinks, `./` and `../` when validating paths in the sshfs server.

## Testing

- Unit tests
- See [here](https://github.com/canonical/multipass/security/advisories/GHSA-rhp2-23c4-r34w#advisory-comment-199930)

## Checklist

MULTI-2598

### DIFF
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -432,6 +432,7 @@ bool mp::SftpServer::validate_path(const fs::path& current_path, bool follows_sy
     }
     catch (const fs::filesystem_error& e)
     {
+        mpl::trace(category, "Could not resolve path {} : {}", current_path.string(), e.what());
         return false;
     }
 
@@ -470,7 +471,7 @@ std::string mp::SftpServer::host_to_guest_path(const fs::path& host_path) const
 {
     std::error_code ec;
     // Get the relative difference between the host path and the mount root
-    auto relative = fs::relative(host_path, source_path, ec);
+    auto relative = MP_FILEOPS.relative(host_path, source_path, ec);
 
     if (ec || relative.empty() || relative == "." || relative.begin()->string() == ".." ||
         relative.is_absolute() || relative.has_root_name())
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -405,7 +405,7 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
            has_gid_mapping_for(MP_FILEOPS.groupId(file_info));
 }
 
-bool mp::SftpServer::validate_path(const fs::path& current_path, bool follows_symlinks)
+bool mp::SftpServer::validate_path(const fs::path& current_path, bool follows_symlinks) const
 {
     if (source_path.empty() || current_path.empty())
         return false;
@@ -440,7 +440,7 @@ bool mp::SftpServer::validate_path(const fs::path& current_path, bool follows_sy
     return source_it == source_path.end();
 }
 
-fs::path mp::SftpServer::get_absolute_path(const char* path)
+fs::path mp::SftpServer::get_absolute_path(const char* path) const
 {
     fs::path raw = path != nullptr ? fs::path(path) : fs::path();
     if (raw.is_relative() && !raw.empty())
@@ -450,7 +450,7 @@ fs::path mp::SftpServer::get_absolute_path(const char* path)
     return raw;
 }
 
-std::optional<fs::path> mp::SftpServer::get_validated_path(sftp_client_message msg)
+std::optional<fs::path> mp::SftpServer::get_validated_path(sftp_client_message msg) const
 {
     bool follows{follows_symlinks(sftp_client_message_get_type(msg))};
     const auto path = get_absolute_path(sftp_client_message_get_filename(msg));
@@ -958,7 +958,7 @@ int mp::SftpServer::handle_readdir(sftp_client_message msg)
 
 int mp::SftpServer::handle_readlink(sftp_client_message msg)
 {
-    auto filename = get_validated_path(msg);
+    const auto filename = get_validated_path(msg);
     if (!filename.has_value())
         return reply_perm_denied(msg);
 
@@ -987,7 +987,7 @@ int mp::SftpServer::handle_readlink(sftp_client_message msg)
 
 int mp::SftpServer::handle_realpath(sftp_client_message msg)
 {
-    auto filename = get_validated_path(msg);
+    const auto filename = get_validated_path(msg);
     if (!filename.has_value())
         return reply_perm_denied(msg);
 
@@ -1002,7 +1002,7 @@ int mp::SftpServer::handle_realpath(sftp_client_message msg)
     }
 
     // Path is already absolute from get_validated_path
-    auto guest_path = host_to_guest_path(*filename);
+    const auto guest_path = host_to_guest_path(*filename);
     return sftp_reply_name(msg, guest_path.c_str(), nullptr);
 }
 
@@ -1207,7 +1207,7 @@ int mp::SftpServer::handle_setstat(sftp_client_message msg)
 
 int mp::SftpServer::handle_stat(sftp_client_message msg, const bool follow)
 {
-    auto filename = get_validated_path(msg);
+    const auto filename = get_validated_path(msg);
     if (!filename.has_value())
         return reply_perm_denied(msg);
 
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -477,7 +477,7 @@ std::string mp::SftpServer::host_to_guest_path(const fs::path& host_path) const
         relative.is_absolute() || relative.has_root_name())
     {
         // If the path is outside the mount or invalid, return the mount path
-        return target_path;
+        return target_path.generic_string();
     }
 
     // Return it as an absolute path from the perspective of the guest
@@ -597,7 +597,7 @@ void mp::SftpServer::run()
                 sshfs_process = create_sshfs_process(ssh_session,
                                                      sshfs_exec_line,
                                                      source_path.string(),
-                                                     target_path);
+                                                     target_path.generic_string());
                 sftp_server_session =
                     make_sftp_session(ssh_session, sshfs_process->release_channel());
 
--- a/src/sshfs_mount/sftp_server.cpp
+++ b/src/sshfs_mount/sftp_server.cpp
@@ -653,7 +653,7 @@ int mp::SftpServer::handle_fstat(sftp_client_message msg)
         return reply_perm_denied(msg);
     }
 
-    QFileInfo file_info(QString::fromStdString(path.string()));
+    QFileInfo file_info(path);
 
     if (file_info.isSymLink())
         file_info = QFileInfo(file_info.symLinkTarget());
@@ -668,8 +668,8 @@ int mp::SftpServer::handle_mkdir(sftp_client_message msg)
     if (!filename.has_value())
         return reply_perm_denied(msg);
 
-    QDir dir(filename->c_str());
-    QFileInfo current_dir(filename->c_str());
+    QDir dir(*filename);
+    QFileInfo current_dir(*filename);
     QFileInfo parent_dir(current_dir.path());
 
     if (!has_id_mappings_for(parent_dir))
@@ -683,14 +683,13 @@ int mp::SftpServer::handle_mkdir(sftp_client_message msg)
         return reply_perm_denied(msg);
     }
 
-    if (!dir.mkdir(filename->c_str()))
+    if (!dir.mkdir(QString::fromStdString(filename->string())))
     {
         mpl::trace(category, "{}: mkdir failed for '{}'", __FUNCTION__, filename);
         return reply_failure(msg);
     }
 
-    if (!MP_PLATFORM.set_permissions(filename->c_str(),
-                                     static_cast<fs::perms>(msg->attr->permissions)))
+    if (!MP_PLATFORM.set_permissions(*filename, static_cast<fs::perms>(msg->attr->permissions)))
     {
         mpl::trace(category, "{}: set permissions failed for '{}'", __FUNCTION__, filename);
         return reply_failure(msg);
@@ -699,7 +698,7 @@ int mp::SftpServer::handle_mkdir(sftp_client_message msg)
     int rev_uid = reverse_uid_for(parent_dir.ownerId(), parent_dir.ownerId());
     int rev_gid = reverse_gid_for(parent_dir.groupId(), parent_dir.groupId());
 
-    if (MP_PLATFORM.chown(filename->c_str(), rev_uid, rev_gid) < 0)
+    if (MP_PLATFORM.chown(filename->string().c_str(), rev_uid, rev_gid) < 0)
     {
         mpl::trace(category,
                    "failed to chown '{}' to owner:{} and group:{}",
@@ -718,7 +717,7 @@ int mp::SftpServer::handle_rmdir(sftp_client_message msg)
     if (!filename.has_value())
         return reply_perm_denied(msg);
 
-    QFileInfo current_dir(filename->c_str());
+    QFileInfo current_dir(*filename);
     if (MP_FILEOPS.exists(current_dir) && !has_id_mappings_for(current_dir))
     {
         mpl::trace(category,
@@ -760,7 +759,7 @@ int mp::SftpServer::handle_open(sftp_client_message msg)
     }
     const auto exists = fs::is_symlink(status) || fs::is_regular_file(status);
 
-    QFileInfo file_info(filename->c_str());
+    QFileInfo file_info(*filename);
     QFileInfo current_dir(file_info.path());
     if ((exists && !has_id_mappings_for(file_info)) ||
         (!exists && !has_id_mappings_for(current_dir)))
@@ -813,7 +812,7 @@ int mp::SftpServer::handle_open(sftp_client_message msg)
         auto new_uid = reverse_uid_for(current_dir.ownerId(), current_dir.ownerId());
         auto new_gid = reverse_gid_for(current_dir.groupId(), current_dir.groupId());
 
-        if (MP_PLATFORM.chown(filename->c_str(), new_uid, new_gid) < 0)
+        if (MP_PLATFORM.chown(filename->string().c_str(), new_uid, new_gid) < 0)
         {
             mpl::trace(category,
                        "failed to chown '{}' to owner:{} and group:{}",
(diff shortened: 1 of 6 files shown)


tobe2098:
Tagging @sharder996 for macOS testing

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +383,23 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
            has_gid_mapping_for(MP_FILEOPS.groupId(file_info));
 }
 
+bool mp::SftpServer::validate_path(const std::string& current_path)
+{
+    if (canonical_source.empty())
+        return false;
+
+    std::error_code ec;
+    auto normalized_current_path = fs::weakly_canonical(current_path, ec);
```
xmkg: Plain fs call -- can we do this through MP_FILEOPS?

tests/unit/test_sftpserver.cpp
```diff
@@ -2914,6 +2920,56 @@ INSTANTIATE_TEST_SUITE_P(SftpServer,
                                            MessageAndReply{SFTP_EXTENDED, SSH_FX_FAILURE}),
                          string_for_param);
 
+TEST_F(SftpServer, BlocksSiblingDirectoryBypass)
```
xmkg: Does the existing suite include positive tests, i.e., this path should be allowed? If not, it might be a good idea to add some to ensure that we're not blocking legitimate calls.

src/sshfs_mount/sftp_server.cpp
```diff
@@ -1200,7 +1211,7 @@ int mp::SftpServer::handle_symlink(sftp_client_message msg)
     const auto old_name = sftp_client_message_get_filename(msg);
```
xmkg: I'd say we should validate the old_name here too. What do you think?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -1292,7 +1303,7 @@ int mp::SftpServer::handle_extended(sftp_client_message msg)
         const auto old_name = sftp_client_message_get_filename(msg);
```
xmkg: idem (handle_symlink)

src/sshfs_mount/sftp_server.cpp
xmkg: Should we also validate the resolved realpath to ensure that it's still in the allowed dir?

xmkg: No validate_path call in this one.

xmkg: Should we also validate this?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -1159,7 +1170,7 @@ int mp::SftpServer::handle_setstat(sftp_client_message msg)
 int mp::SftpServer::handle_stat(sftp_client_message msg, const bool follow)
 {
     auto filename = sftp_client_message_get_filename(msg);
```
xmkg: It might be a good idea to wrap the  `sftp_client_message_get_filename` (or similar) call to centralize the validate_path checks, something like:

```c++
std::optional<const char*> get_filename_validated(sftp_client_message msg){
    const auto filename = sftp_client_message_get_filename(msg);
    if (!validate_path(filename))
    {
        mpl::trace(category,
                   "{}: cannot validate path '{}' against source '{}'",
                   __FUNCTION__,
                   filename,
                   source_path);
        return {};
    }
    return msg;
}

// define a macro to intercept raw calls to sftp_client_message_get_filename
#define sftp_client_message_get_filename(...) \
    static_assert(false, "Use get_filename_validated() instead of sftp_client_message_get_filename()")
```

xmkg (changes requested):

src/sshfs_mount/sftp_server.cpp
tobe2098: Good catch! Yes, I think we should probably check all paths sent from the client

src/sshfs_mount/sftp_server.cpp
tobe2098: This actually has more issues than that. absoluteFilePath is not resolving `../` `./` or symlinks at all, which means we are not resolving the path at all, just prepending the program's cwd if the path is relative! I think we should return an absolute path chrooted on the mount point, what do you think? 

According to the sftp spec we should be returning the resolved path in the server-side, but then we would be leaking information.

tobe2098: This function also has its issues. We should be querying the open fd instead of using the file path, which would also be vulnerable.

src/sshfs_mount/sftp_server.cpp
tobe2098: I agree, but see thread on symlinks for ubiquity. We could add a bool to this function as well

src/sshfs_mount/sftp_server.cpp
tobe2098: No, if we do, and a user creates a symlink in the shared folder that points outside of the shared folder (even if it is within the VM's file tree), the VM will never be able to delete it, stat it or anything, since all functions would fail on validation.

Additionally, this is a problem for path validation. `weakly_canonical` will resolve the symlink, which means that if the symlink is pointing outside of the folder, any symlink operations like deleting them will not be possible anymore, because the path fails validation.

tobe2098: The function will deal with both cases

src/sshfs_mount/sftp_server.cpp
tobe2098: Additionally, all "absolute paths" should be computed by adding the source_path, not by calling absolutePath() functions, since these should be adding sshfs_server's cwd

tobe2098: Otherwise, #define cannot be done because premock uses defines for mocking sftp functions, although it is useful to find all places where it is used.

src/sshfs_mount/sftp_server.cpp
tobe2098: I will pivot to fds in a separate PR, for now we can keep to fixing SFTP leaks

tobe2098: See: https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.11

tobe2098:
Ready for review

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
            has_gid_mapping_for(MP_FILEOPS.groupId(file_info));
 }
 
+bool mp::SftpServer::validate_path(const fs::path& current_path, bool follows_symlinks)
```
xmkg: This function can be marked as `const`

src/sshfs_mount/sftp_server.cpp
```diff
@@ -296,6 +312,7 @@ mp::SftpServer::SftpServer(SSHSession&& session,
       sshfs_process{create_sshfs_process(ssh_session, sshfs_exec_line, source, target)},
       sftp_server_session{make_sftp_session(ssh_session, sshfs_process->release_channel())},
       source_path{source},
+      canonical_source{source},
```
xmkg: This path is the verbatim copy of whatever is present in the source, right? If so, the name is a bit misleading since it's not a canonical path yet. Also, I didn't quite understand how it is different than the source_path itself. Could you elaborate?

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+    auto [source_it, current_it] = std::mismatch(canonical_source.begin(),
+                                                 canonical_source.end(),
+                                                 final_path.begin(),
+                                                 final_path.end());
+    return source_it == canonical_source.end();
+}
+
+fs::path mp::SftpServer::get_absolute_path(const char* path)
+{
+    fs::path raw(path);
```
xmkg: This is UB if path is nullptr, correct?

src/sshfs_mount/sftp_server.cpp
xmkg: Needs unit tests

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+    }
+
+    auto [source_it, current_it] = std::mismatch(canonical_source.begin(),
+                                                 canonical_source.end(),
+                                                 final_path.begin(),
+                                                 final_path.end());
+    return source_it == canonical_source.end();
+}
+
+fs::path mp::SftpServer::get_absolute_path(const char* path)
```
xmkg: Needs unit tests

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+{
+    fs::path raw(path);
+    if (raw.is_relative() && !raw.empty())
+    {
+        return canonical_source / raw;
+    }
+    return raw;
+}
+
+std::optional<fs::path> mp::SftpServer::get_validated_path(sftp_client_message msg)
```
xmkg: Needs unit tests

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+                   "{}: cannot validate path '{}' against source '{}'",
+                   __FUNCTION__,
+                   path,
+                   source_path);
+        return std::nullopt;
+    }
+    return path;
+}
+
+std::string mp::SftpServer::host_to_guest_path(const fs::path& host_path) const
```
xmkg, line 483: Needs unit tests

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+{
+    std::error_code ec;
+    // Get the relative difference between the host path and the mount root
+    auto relative = fs::relative(host_path, canonical_source, ec);
```
xmkg: Should be dispatched through MP_FILEOPS

src/sshfs_mount/sftp_server.cpp
```diff
@@ -935,24 +1002,18 @@ int mp::SftpServer::handle_realpath(sftp_client_message msg)
         return reply_perm_denied(msg);
     }
 
-    auto realpath = QFileInfo(filename).absoluteFilePath();
-    return sftp_reply_name(msg, realpath.toStdString().c_str(), nullptr);
+    // Path is already absolute from get_validated_path
+    auto guest_path = host_to_guest_path(*filename);
```
xmkg: can be marked as `const`

src/sshfs_mount/sftp_server.cpp
```diff
@@ -907,25 +980,19 @@ int mp::SftpServer::handle_readlink(sftp_client_message msg)
...
 {
-    auto filename = sftp_client_message_get_filename(msg);
-    if (!validate_path(source_path, filename))
-    {
-        mpl::trace(category,
-                   "{}: cannot validate path \'{}\' against source \'{}\'",
-                   __FUNCTION__,
-                   filename,
-                   source_path);
+    auto filename = get_validated_path(msg);
```
xmkg: This is const in some, non-const in others.

tests/unit/test_sftpserver.cpp
```diff
@@ -454,24 +455,29 @@ TEST_F(SftpServer, realpathFailsWhenIdsAreNotMapped)
...
     EXPECT_CALL(*file_ops, dir_iterator).WillOnce(Return(std::make_unique<mpt::MockDirIterator>()));
+    EXPECT_CALL(*file_ops, weakly_canonical).WillOnce([](const fs::path& path) {
+        return fs::weakly_canonical(path);
+    });
 
     REPLACE(sftp_reply_handle, [](auto...) { return SSH_OK; });
     REPLACE(sftp_get_client_message, make_msg_handler());
 
-    auto sftp = make_sftpserver(mpt::test_data_path().toStdString());
+    auto sftp = make_sftpserver(fs::weakly_canonical(mpt::test_data_path().toStdString()).string());
```
xmkg: Shouldn't we let the SFTP server canonicalize the given path?

tests/unit/test_sftpserver.cpp
```diff
@@ -481,21 +487,25 @@ TEST_F(SftpServer, opendirNotExistingFails)
...
+        return fs::weakly_canonical(path);
+    });
 
     REPLACE(sftp_get_client_message, make_msg_handler());
     int no_such_file_calls{0};
     REPLACE(sftp_reply_status,
             make_reply_status(msg.get(), SSH_FX_NO_SUCH_FILE, no_such_file_calls));
 
-    auto sftp = make_sftpserver(mpt::test_data_path().toStdString());
+    auto sftp = make_sftpserver(fs::weakly_canonical(mpt::test_data_path().toStdString()).string());
```
xmkg: idem

tests/unit/test_sftpserver.cpp
```diff
@@ -505,20 +515,24 @@ TEST_F(SftpServer, opendirNotReadableFails)
         err = std::make_error_code(std::errc::permission_denied);
         return std::make_unique<mpt::MockDirIterator>();
     });
+    EXPECT_CALL(*file_ops, weakly_canonical).WillOnce([](const fs::path& path) {
+        return fs::weakly_canonical(path);
+    });
 
     REPLACE(sftp_get_client_message, make_msg_handler());
-    auto sftp = make_sftpserver(mpt::test_data_path().toStdString());
+    auto sftp = make_sftpserver(fs::weakly_canonical(mpt::test_data_path().toStdString()).string());
```
xmkg: idem

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+            if (parent.empty())
+            {
+                parent = ".";
+            }
+            auto resolved_parent = MP_FILEOPS.weakly_canonical(parent);
+
+            final_path = (resolved_parent / current_path.filename()).lexically_normal();
+        }
+    }
+    catch (const fs::filesystem_error& e)
```
xmkg, line 446: Unused variable (e)

src/sshfs_mount/sftp_server.h
```diff
@@ -68,6 +68,10 @@ class SftpServer
     bool has_reverse_uid_mapping_for(const int uid);
     bool has_reverse_gid_mapping_for(const int gid);
     bool has_id_mappings_for(const QFileInfo& file_info);
+    bool validate_path(const fs::path& current_path, bool is_target);
```
xmkg: Parameter name mismatch with definition.

src/sshfs_mount/sftp_server.h
```diff
@@ -68,6 +68,10 @@ class SftpServer
...
+    std::string host_to_guest_path(const fs::path& host_path) const;
+    fs::path get_absolute_path(const char* path);
```
xmkg: can be marked `const`

src/sshfs_mount/sftp_server.cpp
```diff
@@ -389,6 +406,86 @@ bool mp::SftpServer::has_id_mappings_for(const QFileInfo& file_info)
...
+
+std::optional<fs::path> mp::SftpServer::get_validated_path(sftp_client_message msg)
+{
+    bool follows{follows_symlinks(sftp_client_message_get_type(msg))};
+    const auto path = get_absolute_path(sftp_client_message_get_filename(msg));
+    if (!validate_path(path, follows))
+    {
+        mpl::trace(category,
+                   "{}: cannot validate path '{}' against source '{}'",
+                   __FUNCTION__,
```
xmkg, line 475: We can probably use std::source_location for this

src/sshfs_mount/sftp_server.cpp
xmkg: can be `const`

xmkg (changes requested):
Another iteration for your consideration :)

src/sshfs_mount/sftp_server.cpp
tobe2098: Yes, so the path is canonical, because it is canonicalized in the instantiation of VMMount, higher in the call stack. 
I added the fs::path var to avoid allocating every time we do the path validation. Alternatively we could switch the type of the std::string source_path variable.

src/sshfs_mount/sftp_server.cpp
tobe2098: Good catch!

src/sshfs_mount/sftp_server.cpp
tobe2098: I did not change the constness of the variable, but I can review if making it const is possible

tests/unit/test_sftpserver.cpp
tobe2098: It is usually canonicalized at the VMMount level, I opted to not add another canonicalization, since it is a no-op in non-test code. On the other hand, the code would be more understandable. WDYT?

src/sshfs_mount/sftp_server.cpp
tobe2098: I will move this to a separate PR to avoid conflicts with the release branch when cherry-picking, if it is not an issue.

src/sshfs_mount/sftp_server.cpp
xmkg: Type switch sounds good to me if they contain the same thing.

tests/unit/test_sftpserver.cpp
xmkg: I think it would be better if:

a-) sftp_server would canonicalize regardless
b-) Throw on construction if the given path is not canonicalized

I'll leave it up to you.

src/sshfs_mount/sftp_server.cpp
xmkg: Sure thing!

xmkg (dismissed):
This LGTM, local mount tests work as expected. Also, cli tests are green. Needs to be verified on macOS, too.

xmkg:
@tobe2098, do you think it would make sense to add a CLI test for this?

tobe2098:
Maybe, which test case did you have in mind? The security advisory PoC?

xmkg:
> Maybe, which test case did you have in mind? The security advisory PoC?

Yeah, I think it's trivial enough to include it in the suite. If we do, we can at least have insurance against not introducing this bug again inadvertently.

tobe2098:
Let us not merge yet, there was another vulnerability that I just found.

tobe2098:
>> Maybe, which test case did you have in mind? The security advisory PoC?

>Yeah, I think it's trivial enough to include it in the suite. If we do, we can at least have insurance against not introducing this bug again inadvertently.

We can add it in the separate PR to avoid conflicts, although the unit tests should cover all existing attack surfaces.

tobe2098:
Tested on macOS and the jailbreak does not work, as intended.

tobe2098:
Ready for review again

xmkg (approved):

PR #4868, opened 2026-05-06
[gha-win-mac] Gate CLI test dispatch on pkg url

xmkg:
# Description

Add a condition to the CLI test dispatch jobs that verifies the package URL is set.


jimporter (approved):
LGTM!

PR #4869, opened 2026-05-06
[win] Use fragment files for Windows Terminal

### LINKED ISSUE #1585: Integration with Windows Terminal heavily modifies its settings
**Describe the bug**
Integration changes Windows Terminal settings file in ways its not supposed to:
- spaces converted to tabs
- added space between key and colon
- all keys got sorted by name
- blank lines got removed

**To Reproduce**
How, and what happened?
1. Make sure there's no multipass profile in settings.json
1. Run `multipass`
1. Open Windows Terminal `settings.json`

**Expected behavior**
Integration to just add multipass profile without modifying anything else.

**Additional info**
 - OS: Windows 10 Pro x64 Build 18363
- `multipass version`
multipass  1.3.0+win
multipassd 1.3.0+win
- `multipass info --all`
Name:           ubuntu-lts
State:          Running
IPv4:           172.18.48.150
Release:        Ubuntu 18.04.4 LTS
Image hash:     3b2e3aaaebf2 (Ubuntu 18.04 LTS)
Load:           0.08 0.02 0.01
Disk usage:     5.3G out of 9.5G
Memory usage:   3.0G out of 3.9G

**Additional context**
The problem is this behaviour is unlikable, pretty sure other users don't want their settings file setup got modified like that.

jimporter:
# Description

This PR changes how we add our Windows Terminal configs to use [fragment files](https://learn.microsoft.com/en-us/windows/terminal/json-fragment-extensions). This is simpler and safer, since we don't need to parse the user's settings file. It also means we can use Boost.JSON here instead of jsoncpp (which supports adding comments to JSON), so we can remove jsoncpp as a dependency.

## Testing

- Unit tests updated to check the new code
- Manual testing steps:

  1. From the `main` branch, start the Multipass daemon and run any command (e.g. `multipass list`)
  2. Close all Windows Terminals and open a new one
  3. Verify that the Multipass preset is available (click the down arrow by the new tab button)
  4. From this branch, start the Multipass daemon and run any command (e.g. `multipass list`)
  5. Close all Windows Terminals and open a new one
  6. Verify that the "Multipass (fragmentized)" preset is available, and the original "Multipass" preset is hidden

## Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "3rd-party/jsoncpp"]
-	path = 3rd-party/jsoncpp
-	url = https://github.com/open-source-parsers/jsoncpp.git
 [submodule "3rd-party/vcpkg"]
 	path = 3rd-party/vcpkg
 	url = https://github.com/microsoft/vcpkg
--- a/src/platform/platform_win.cpp
+++ b/src/platform/platform_win.cpp
@@ -681,8 +681,6 @@ void mp::platform::sync_winterm_profiles()
 {
     constexpr auto log_category = "winterm";
     const auto winterm_setting = MP_SETTINGS.get(mp::winterm_key);
-    if (winterm_setting == none)
-        return;
 
     const auto winterm_path = windows_terminal_config_path();
     if (winterm_path.isEmpty())
@@ -693,24 +691,38 @@ void mp::platform::sync_winterm_profiles()
         return;
     }
 
-    const auto mp_fragment_dir =
-        QDir{MP_UTILS.make_dir(QDir{winterm_path}.filePath("Fragments/Multipass"))};
-    const auto mp_fragment_file = QDir(mp_fragment_dir).filePath("multipass.json");
-    if (MP_FILEOPS.exists(mp_fragment_file))
-    {
-        mpl::debug(log_category,
-                   "Multipass fragment for Windows Terminal already exists: {}",
-                   mp_fragment_file.toStdString());
-        return;
-    }
+    const auto mp_fragment_dir = MP_PLATFORM.qstr_to_path(
+        MP_UTILS.make_dir(QDir{winterm_path}.filePath("Fragments/Multipass")));
+    const auto mp_fragment_file = mp_fragment_dir / "multipass.json";
 
-    try
+    if (winterm_setting == none)
     {
-        MP_FILEOPS.write_transactionally(mp_fragment_file, pretty_print(create_primary_profile()));
+        // If the user requested no Windows Terminal profile, delete the file. This is safe because
+        // any user modifications to the profile are in their main `settings.json`. It's necessary
+        // because their `settings.json` overrides the `hidden` property, meaning we can't use that
+        // property in our fragment.
+        if (MP_FILEOPS.exists(mp_fragment_file))
+            MP_FILEOPS.remove(mp_fragment_file);
     }
-    catch (const std::exception& e)
+    else
     {
-        mpl::error(log_category, "Failed to write Windows Terminal fragment: {}", e.what());
+        if (MP_FILEOPS.exists(mp_fragment_file))
+        {
+            mpl::debug(log_category,
+                       "Multipass fragment for Windows Terminal already exists: {}",
+                       mp_fragment_file);
+            return;
+        }
+
+        try
+        {
+            MP_FILEOPS.write_transactionally(mp_fragment_file,
+                                             pretty_print(create_primary_profile()));
+        }
+        catch (const std::exception& e)
+        {
+            mpl::error(log_category, "Failed to write Windows Terminal fragment: {}", e.what());
+        }
     }
 }
 
--- a/src/platform/platform_win.cpp
+++ b/src/platform/platform_win.cpp
@@ -19,6 +19,7 @@
 #include <multipass/exceptions/formatted_exception_base.h>
 #include <multipass/exceptions/settings_exceptions.h>
 #include <multipass/format.h>
+#include <multipass/json_utils.h>
 #include <multipass/logging/log.h>
 #include <multipass/platform.h>
 #include <multipass/settings/custom_setting_spec.h>
@@ -51,8 +52,6 @@
 #include <QtGlobal>
 
 #include <scope_guard.hpp>
-
-#include <json/json.h>
 
 #include <WS2tcpip.h>
 #include <WinSock2.h>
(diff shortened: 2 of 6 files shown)


src/platform/platform_win.cpp
```diff
@@ -827,34 +675,37 @@ QString mp::platform::interpret_setting(const QString& key, const QString& val)
 void mp::platform::sync_winterm_profiles()
 {
     constexpr auto log_category = "winterm";
-    const auto profiles_path = locate_profiles_path();
     const auto winterm_setting = MP_SETTINGS.get(mp::winterm_key);
+    if (winterm_setting == none)
+        return;
```
xmkg: The old implementation was hiding the profile if the setting was set to `none`. Is the behavior change intentional, or a regression?

src/platform/platform_win.cpp
```diff
@@ -827,34 +675,37 @@ QString mp::platform::interpret_setting(const QString& key, const QString& val)
...
+        QDir{MP_UTILS.make_dir(QDir{winterm_path}.filePath("Fragments/Multipass"))};
+    const auto mp_fragment_file = QDir(mp_fragment_dir).filePath("multipass.json");
+    if (MP_FILEOPS.exists(mp_fragment_file))
     {
-        const auto level = winterm_setting == none ? mpl::Level::debug : mpl::Level::warning;
-        mpl::log_message(level, log_category, e.what());
+        mpl::debug(log_category,
+                   "Multipass fragment for Windows Terminal already exists: {}",
+                   mp_fragment_file.toStdString());
+        return;
```
xmkg: What if we decided to update the parameters in the fragment? Wouldn't returning here cause the existing profiles to go stale?

src/platform/platform_win.cpp
```diff
@@ -140,192 +139,41 @@ QString interpret_winterm_setting(const QString& val)
...
+    return {{"profiles",
+             {{
+                  {"updates", mp::winterm_old_profile_guid},
+                  {"hidden", true},
+              },
+              {
+                  {"guid", mp::winterm_profile_guid},
+                  // FIXME: Before merging, rename this to "Multipass". The "fragmentized" suffix is
+                  // just to make manual testing easier.
+                  {"name", "Multipass (fragmentized)"},
```
xmkg, line 169: So, this file will hide the old profile and provide a new profile with the same name, correct? If so, is it okay for the profiles to have the same name?

.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "3rd-party/jsoncpp"]
```
xmkg, line 1: It's so nice to see this gone :)

tests/unit/windows/test_platform_win.cpp
```diff
@@ -181,428 +118,74 @@ TEST(PlatformWin, unsupportedWintermSettingValuesCauseException)
...
-    const auto [json_file_name, tmp_file_guard] = guarded_fake_json(json);
-    const auto timestamp = QFileInfo{json_file_name}.lastModified();
-
-    mp::platform::sync_winterm_profiles();
-
-    EXPECT_EQ(QFileInfo{json_file_name}.lastModified(), timestamp);
-    EXPECT_EQ(json, read_json(json_file_name));
-}
-
-TEST_P(TestWinTermSyncJson, wintermSyncDisablesVisibleProfileIfSettingNone)
```
xmkg, line 568: Relevant test cases to winterm=none ^^

src/platform/platform_win.cpp
jimporter: After reading the old implementation more carefully, I think there's a regression here. The behavior *should* be the same if the setting is `primary` or `none` the first time we run this hook. (If the setting is `none` and there's no Multipass profile, the old code did nothing.)

However, when changing the setting to `none` later, we should mark the profile as hidden. I'll fix that. We might be able to have the "hide this profile" flag exist as a *separate* JSON fragment, so hiding/unhiding is just a matter of creating/deleting the right file.

src/platform/platform_win.cpp
jimporter: Right. If we update the parameters, we'd need to update this code. I have a few ideas for how to do this; one could be to have a versioned filename (or include a hash of the contents in the filename), and then we know whether the profile is out of date without parsing the JSON.

src/platform/platform_win.cpp
jimporter: I'm open to any option here. We could give it a new name, or keep the old name.

One thing we *can't* do unfortunately is use JSON fragments to update the old profile's name. We'd have to modify it in the main `settings.json`, which I'd like to avoid. (Messing with the user's main settings file is a little riskier than messing with files Multipass owns.)

tests/unit/windows/test_platform_win.cpp
jimporter: Added a test case here.

src/platform/platform_win.cpp
jimporter: I've added a comment just above the definition of our profile fragment to remind any future authors to think about how to handle updating this config if we ever need to do so (not likely, in my estimation, but you never know).

src/platform/platform_win.cpp
jimporter: Ok, I've implemented this by removing the JSON fragment when `winterm_setting == none`. (See the second commit in this PR.)

xmkg:
@jimporter the CI seems to be stuck. The workflow run shows as running for the last 10 days, but with no progress. I tried force-cancelling the run, but it failed with error code 500. Can you try pushing a new commit so that a new workflow would be triggered?

src/platform/platform_win.cpp
xmkg: I'm fine with the same name approach -- I was just wondering if it would cause any conflicts, but I'd doubt that since the profiles seem to be identified by GUIDs.

jimporter:
@xmkg Ok, it looks like the CI succeeded this time. (Aside from `codecov/project` failing because I reduced the total amount of code in Multipass by enough for the coverage check to notice!)

xmkg:
@jimporter tested locally, the change works fine. I'd like to confirm some of the behavior to ensure they're "as intended":

-  The profile does not switch to the new one on version upgrade. It becomes active when client.apps.windows-terminal.profiles is explicitly set to primary + terminal restart.
- It appears on fresh install without having to set anything
- The multipass (fragmentized) profile does not get removed on uninstall, also persists after reboot

PR #4870, opened 2026-05-07
[tests] Generalize custom image matchers

sharder996:
Generalize the matching of 3rd-party images in the CLI tests so as not to depend on a certain release.

Another option was to parse `data/distributions/distribution-info.json`, but I wanted to avoid depending on files outside the cli test directory in case someone is doing a sparse checkout or if the CLI tests eventually move to a separate repository. We could also do a GET request on `https://raw.githubusercontent.com/canonical/multipass/refs/heads/main/data/distributions/distribution-info.json`, but that seemed a bit overkill.

### DIFF
--- a/tests/cli/cli_find_test.py
+++ b/tests/cli/cli_find_test.py
@@ -129,60 +129,34 @@ class TestFind:
             # Pull the keys present in expected and do a comparison
             assert {k: image[k] for k in expected_image} == expected_image
 
-    @pytest.mark.parametrize(
-        "param",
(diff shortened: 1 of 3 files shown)


xmkg:
@sharder996 I'm OK with the changes but the tests are failing:

  ❌ test_launch_debian[instance0]
	AssertionError
  ❌ test_launch_fedora[instance0]
	AssertionError

xmkg (approved):

PR #4880, opened 2026-05-15
[macOS] Update vcpkg to 2026.05.25 and qtbase to 6.11 for macOS 26.5

sharder996:
This PR updates the vcpkg submodule which uses a newer version of qtbase which includes a fix for [QTBUG-145239.patch](https://qt-project.atlassian.net/browse/QTBUG-145239). This patch is not yet available in a vcpkg release so the most recent commit ATOW is used.  

The latest macOS update for xcode manifests this issue the newer SDK ships a version of Clang that reports `__has_builtin(__yield)` as true on Apple Silicon, but `__yield()` still requires `<arm_acle.h>` for its declaration. Qt 6.9.1's `qyieldcpu.h` hits this path first, causing a fatal `-Werror,-Wimplicit-function-declaration` error during the bootstrap build of `qtbase`. On the latest version of macOS, Multipass fails to build.

The newer version of Qt has a few API changes which necessitate a few changes in the tests. I also update some regex strings to be more compliant with macOS.

Additional updates:
- gRPC is updated to 1.76.0, required as a result of other dependencies updating.
- Update control build dependencies list


sharder996:
Requesting a review from @xmkg since you have an M2 Mac that can verify the build failures with the latest version of macOS.

vcpkg.json
```diff
@@ -1,5 +1,5 @@
 {
-    "builtin-baseline": "74e6536215718009aae747d86d84b78376bf9e09",
+    "builtin-baseline": "c1ce926ddf75f166cbdd444bbaf463410ac99b17",
```
xmkg: The next vcpkg release is probably just around the corner -- it'd be wise to wait until early June to use a tagged release instead.

xmkg (commented):
The changes look good on the surface -- waiting until the CI passes for further testing. Also it might be a good idea to reflect the vcpkg upgrade in the PR title too.

vcpkg.json
sharder996: Yes, I'd like to wait as well. It's just a pain because all my builds are broken without this change.

sharder996: Updated to the latest release: https://github.com/microsoft/vcpkg/releases/tag/2026.05.25

xmkg (approved):

PR #4890, opened 2026-05-21
[arch] ppc64el maintenance

sharder996:
A couple fixes for Multipass on ppc64 and s390x:

1. Install cmake from apt as the snap package is no longer being updated. https://forum.snapcraft.io/t/announcement-dropping-architectures-from-the-cmake-snap/35137
2. Edit qemu system arg to use host cpu instead of hardcoded `POWER9`
3. Filter out non-JSON output from QMP. This was an issue as a ppc64 machine was logging:
```
May 21 19:38:48 baltar-kernel multipassd[119399]: QMP: Can't open directory /proc/device-tree/cpus/
                                                  Can't open directory /proc/device-tree/cpus/
                                                  Can't open directory /proc/device-tree/cpus/
                                                  {"QMP": {"version": {"qemu": {"micro": 3, "minor": 0, "major": 10}, "package": ""}, "capabilities": ["oob"]}}
```

---

MULTI-2631


tobe2098 (approved):
LGTM, good work Scott!

PR #4899, opened 2026-05-26
[git] Mark vcpkg port patches as generated

xmkg:
We're copying the upstream patches verbatim so mark them as generated to hide them by default in github PRs.

https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#generated-code


ricab (commented):
Hi Mustafa, I am a little confused by this. Here are a few questions I have:

- `vcpkg-ports/**/multipass-patches/*.patch` - these come from us, right?
- I believe some outside `multipass-patches` come from us as well, no? For example, the thread_local and more args in premock:
  * https://github.com/canonical/multipass/commit/65011fff641a7b8464ac008fe328bb3d2ccfd03a
  * https://github.com/canonical/multipass/commit/ebee1c456cbe07ed4b81872abe456f958de1c229
- For the ones that do come from upstream, isn't `linguist-vendored` a better match?
- Was there a specific motivation for this PR? Like do you have an example of the effect you are trying to obtain/avoid?

The leading `3rd-party` dir appears missing.

Thanks in advance for clarifying!

xmkg:
> Hi Mustafa, I am a little confused by this. Here are a few questions I have:

Sure!


> * `vcpkg-ports/**/multipass-patches/*.patch` - these come from us, right?

Yes. I've moved our patches to a specific folder so it would be easier to distinguish them from the upstream ones.

> * I believe some outside `multipass-patches` come from us as well, no? For example, the thread_local and more args in premock:

Good catch! There's no official vcpkg port for premock, but I think it would be best to stick with the pattern. I'll move them.

> * For the ones that do come from upstream, isn't `linguist-vendored` a better match?

Great idea! We're vendoring these parts, so I think that would suit better. 

> * Was there a specific motivation for this PR? Like do you have an example of the effect you are trying to obtain/avoid?

It came out from this PR https://github.com/canonical/multipass/pull/4880/changes. I thought it would be nice to cut the noise down by default.

> The leading `3rd-party` dir appears missing.

👍 I'll fix that too.

xmkg:
> Great idea! We're vendoring these parts, so I think that would suit better.

@ricab It seems like auto-collapse behavior is only available for `linguist-generated`.

ricab (commented):
This is a nice idea for improving diff readability, but I wonder if we can solve the underlying problem more directly.

The reason the upstream patches show up as noise in our diffs is that they're copies — and copies tend to drift, requiring manual maintenance to stay in sync. The .gitattributes approach also has a subtle downside: it marks those files as generated, which could hide the divergence.

Couldn't we instead symlink the patches to the originals in the submodule? That would avoid spurious diffs and ensure things stay in sync. We could also use CMake to verify that there were no broken links, before bootstrapping vcpkg. WDYT?

PR #4913, opened 2026-06-01
[platform-win] Initialize pSD to nullptr

xmkg:
pSD is being auto-released by the scope guard but it's not guaranteed to be initialized, and it's not null-initialized either meaning LocalFree would be called with a garbage value.

This patch fixes that.


ricab (approved):
Yep.

PR #4928, opened 2026-06-02
[cli-tests] QoL improvements (error handling, unit tests)

xmkg:
Two bugs in MultipassdGovernor caused raw CancelledError to propagate into pytest fixture setup instead of proper TestSessionFailure/TestCaseFailure exceptions:
1. _monitor() didn't catch CancelledError when awaiting a cancelled stdout_task.
2. start_async() self-cancelled its own task instead of raising a meaningful exception when the daemon died during startup.
Also adds __test__ = False to exception classes to suppress pytest collection warnings, and adds unit tests for the governor, event loop, and async subprocess utilities, including CI enablement.

The unit tests for the cli-test harness are implemented using AI-assisted, agentic workflow.

MULTI-2652

### DIFF
--- a/.github/workflows/cli-tests-ut.yml
+++ b/.github/workflows/cli-tests-ut.yml
@@ -1,4 +1,4 @@
-name: CLI Unit Tests
+name: Unit Tests for cli-test harness
 
 on:
   workflow_call:
--- a/.github/workflows/cli-tests-ut.yml
+++ b/.github/workflows/cli-tests-ut.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         with:
           sparse-checkout: tests/cli
 
--- a/tests/cli/unit/conftest.py
+++ b/tests/cli/unit/conftest.py
@@ -1,0 +1,19 @@
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="session")
+def ensure_sudo_auth():
+    """Override parent fixture - unit tests don't need sudo."""
+    yield
+
+
+@pytest.fixture(autouse=True, scope="session")
+def ensure_multipass_binaries_are_present():
+    """Override parent fixture - unit tests don't need multipass binaries."""
+    yield
+
+
+@pytest.fixture(autouse=True, scope="session")
+def environment_setup():
+    """Override parent fixture - unit tests don't need environment setup."""
+    yield
--- a/tests/cli/unit/test_async_subprocess.py
+++ b/tests/cli/unit/test_async_subprocess.py
@@ -19,7 +19,7 @@ class TestAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_normal_enter_exit(self):
         """Normal enter/exit should start and terminate process cleanly."""
-        async with AsyncSubprocess("sleep", "0.1") as proc:
+        async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(0.1)") as proc:
             assert proc is not None
             assert proc.returncode is None
 
@@ -28,7 +28,7 @@ class TestAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_process_already_exited_on_exit(self):
         """Process already exited on __aexit__ should not error."""
-        async with AsyncSubprocess("true") as proc:
+        async with AsyncSubprocess(sys.executable, "-c", "import sys; sys.exit(0)") as proc:
             await proc.wait()
             assert proc.returncode == 0
 
@@ -37,7 +37,7 @@ class TestAsyncSubprocess:
         """Cancellation during __aenter__ should spawn then terminate process."""
 
         async def cancellable_spawn():
-            async with AsyncSubprocess("sleep", "10") as proc:
+            async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(10)") as proc:
                 await asyncio.sleep(3600)
 
         task = asyncio.create_task(cancellable_spawn())
@@ -54,7 +54,7 @@ class TestAsyncSubprocess:
         """Cancellation during __aexit__ should still clean up process."""
 
         async def cancellable_exit():
-            async with AsyncSubprocess("sleep", "10") as proc:
+            async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(10)") as proc:
                 await asyncio.sleep(3600)
 
         task = asyncio.create_task(cancellable_exit())
@@ -69,7 +69,7 @@ class TestAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_long_running_process_terminated_on_exit(self):
         """Long-running process should be terminated on context exit."""
-        async with AsyncSubprocess("sleep", "3600") as proc:
+        async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(3600)") as proc:
             assert proc.returncode is None
 
         assert proc.returncode is not None
@@ -111,7 +111,7 @@ class TestStdoutAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_captures_stdout(self):
         """StdoutAsyncSubprocess should capture stdout."""
-        async with StdoutAsyncSubprocess("echo", "hello") as proc:
+        async with StdoutAsyncSubprocess(sys.executable, "-c", "print('hello')") as proc:
             stdout, _ = await proc.communicate()
             assert b"hello" in stdout
 
@@ -134,7 +134,7 @@ class TestSilentAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_suppresses_all_output(self):
         """SilentAsyncSubprocess should suppress all output."""
-        async with SilentAsyncSubprocess("echo", "hello") as proc:
+        async with SilentAsyncSubprocess(sys.executable, "-c", "print('hello')") as proc:
             await proc.communicate()
             assert proc.returncode == 0
 
--- a/tests/cli/unit/test_async_subprocess.py
+++ b/tests/cli/unit/test_async_subprocess.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+"""Unit tests for AsyncSubprocess context manager."""
+
+import asyncio
+import sys
+
+import pytest
+
+from cli.utilities.threadutils import AsyncSubprocess, StdoutAsyncSubprocess, SilentAsyncSubprocess
+
+
+class TestAsyncSubprocess:
+    """Test AsyncSubprocess context manager with cancellation handling."""
+
+    @pytest.mark.asyncio
+    async def test_normal_enter_exit(self):
+        """Normal enter/exit should start and terminate process cleanly."""
+        async with AsyncSubprocess("sleep", "0.1") as proc:
+            assert proc is not None
+            assert proc.returncode is None
+        
+        assert proc.returncode is not None
+
+    @pytest.mark.asyncio
+    async def test_process_already_exited_on_exit(self):
+        """Process already exited on __aexit__ should not error."""
+        async with AsyncSubprocess("true") as proc:
+            await proc.wait()
+            assert proc.returncode == 0
+
+    @pytest.mark.asyncio
+    async def test_cancellation_during_enter(self):
+        """Cancellation during __aenter__ should spawn then terminate process."""
+        async def cancellable_spawn():
+            async with AsyncSubprocess("sleep", "10") as proc:
+                await asyncio.sleep(3600)
+        
+        task = asyncio.create_task(cancellable_spawn())
+        await asyncio.sleep(0.1)
+        task.cancel()
+        
+        try:
+            await task
+        except asyncio.CancelledError:
+            pass
+
+    @pytest.mark.asyncio
+    async def test_cancellation_during_exit(self):
+        """Cancellation during __aexit__ should still clean up process."""
+        async def cancellable_exit():
+            async with AsyncSubprocess("sleep", "10") as proc:
+                await asyncio.sleep(3600)
+        
+        task = asyncio.create_task(cancellable_exit())
+        await asyncio.sleep(0.1)
+        task.cancel()
+        
+        try:
+            await task
+        except asyncio.CancelledError:
+            pass
+
+    @pytest.mark.asyncio
+    async def test_long_running_process_terminated_on_exit(self):
+        """Long-running process should be terminated on context exit."""
+        async with AsyncSubprocess("sleep", "3600") as proc:
+            assert proc.returncode is None
+        
+        assert proc.returncode is not None
+
+    @pytest.mark.asyncio
+    async def test_process_killed_after_terminate_timeout(self):
+        """Process should be killed if terminate times out."""
+        async with AsyncSubprocess(sys.executable, "-c", 
+                                   "import signal; signal.signal(signal.SIGTERM, signal.SIG_IGN); import time; time.sleep(3600)") as proc:
+            pass
+        
+        assert proc.returncode is not None
+
+
+class TestStdoutAsyncSubprocess:
+    """Test StdoutAsyncSubprocess captures stdout."""
+
+    @pytest.mark.asyncio
+    async def test_captures_stdout(self):
+        """StdoutAsyncSubprocess should capture stdout."""
+        async with StdoutAsyncSubprocess("echo", "hello") as proc:
+            stdout, _ = await proc.communicate()
+            assert b"hello" in stdout
+
+    @pytest.mark.asyncio
+    async def test_stderr_redirected_to_stdout(self):
+        """Stderr should be redirected to stdout."""
+        async with StdoutAsyncSubprocess(sys.executable, "-c", 
+                                         "import sys; print('out'); print('err', file=sys.stderr)") as proc:
+            stdout, _ = await proc.communicate()
+            assert b"out" in stdout
+            assert b"err" in stdout
+
+
+class TestSilentAsyncSubprocess:
+    """Test SilentAsyncSubprocess suppresses output."""
+
+    @pytest.mark.asyncio
+    async def test_suppresses_all_output(self):
+        """SilentAsyncSubprocess should suppress all output."""
+        async with SilentAsyncSubprocess("echo", "hello") as proc:
+            await proc.communicate()
+            assert proc.returncode == 0
+
+    @pytest.mark.asyncio
+    async def test_stderr_suppressed(self):
+        """Stderr should be suppressed."""
+        async with SilentAsyncSubprocess(sys.executable, "-c", 
+                                         "import sys; print('err', file=sys.stderr)") as proc:
+            await proc.communicate()
+            assert proc.returncode == 0
--- a/tests/cli/unit/test_async_subprocess.py
+++ b/tests/cli/unit/test_async_subprocess.py
@@ -6,7 +6,11 @@ import sys
 
 import pytest
 
-from cli.utilities.threadutils import AsyncSubprocess, StdoutAsyncSubprocess, SilentAsyncSubprocess
+from cli.utilities.threadutils import (
+    AsyncSubprocess,
+    StdoutAsyncSubprocess,
+    SilentAsyncSubprocess,
+)
 
 
 class TestAsyncSubprocess:
@@ -18,7 +22,7 @@ class TestAsyncSubprocess:
         async with AsyncSubprocess("sleep", "0.1") as proc:
             assert proc is not None
             assert proc.returncode is None
-        
+
         assert proc.returncode is not None
 
     @pytest.mark.asyncio
@@ -31,14 +35,15 @@ class TestAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_cancellation_during_enter(self):
         """Cancellation during __aenter__ should spawn then terminate process."""
+
         async def cancellable_spawn():
             async with AsyncSubprocess("sleep", "10") as proc:
                 await asyncio.sleep(3600)
-        
+
         task = asyncio.create_task(cancellable_spawn())
         await asyncio.sleep(0.1)
         task.cancel()
-        
+
         try:
             await task
         except asyncio.CancelledError:
@@ -47,14 +52,15 @@ class TestAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_cancellation_during_exit(self):
         """Cancellation during __aexit__ should still clean up process."""
+
         async def cancellable_exit():
             async with AsyncSubprocess("sleep", "10") as proc:
                 await asyncio.sleep(3600)
-        
+
         task = asyncio.create_task(cancellable_exit())
         await asyncio.sleep(0.1)
         task.cancel()
-        
+
         try:
             await task
         except asyncio.CancelledError:
@@ -65,16 +71,19 @@ class TestAsyncSubprocess:
         """Long-running process should be terminated on context exit."""
         async with AsyncSubprocess("sleep", "3600") as proc:
             assert proc.returncode is None
-        
+
         assert proc.returncode is not None
 
     @pytest.mark.asyncio
     async def test_process_killed_after_terminate_timeout(self):
         """Process should be killed if terminate times out."""
-        async with AsyncSubprocess(sys.executable, "-c", 
-                                   "import signal; signal.signal(signal.SIGTERM, signal.SIG_IGN); import time; time.sleep(3600)") as proc:
+        async with AsyncSubprocess(
+            sys.executable,
+            "-c",
+            "import signal; signal.signal(signal.SIGTERM, signal.SIG_IGN); import time; time.sleep(3600)",
+        ) as proc:
             pass
-        
+
         assert proc.returncode is not None
 
     @pytest.mark.asyncio
@@ -83,7 +92,7 @@ class TestAsyncSubprocess:
         async with AsyncSubprocess(sys.executable, "-c", "exit(0)") as proc:
             # Explicitly wait for the process to be reaped
             await proc.wait()
-        
+
         assert proc.returncode == 0
 
     @pytest.mark.asyncio
@@ -92,7 +101,7 @@ class TestAsyncSubprocess:
         async with AsyncSubprocess(sys.executable, "-c", "exit(42)") as proc:
             # Explicitly wait for the process to be reaped
             await proc.wait()
-        
+
         assert proc.returncode == 42
 
 
@@ -109,8 +118,11 @@ class TestStdoutAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_stderr_redirected_to_stdout(self):
         """Stderr should be redirected to stdout."""
-        async with StdoutAsyncSubprocess(sys.executable, "-c", 
-                                         "import sys; print('out'); print('err', file=sys.stderr)") as proc:
+        async with StdoutAsyncSubprocess(
+            sys.executable,
+            "-c",
+            "import sys; print('out'); print('err', file=sys.stderr)",
+        ) as proc:
             stdout, _ = await proc.communicate()
             assert b"out" in stdout
             assert b"err" in stdout
@@ -129,7 +141,8 @@ class TestSilentAsyncSubprocess:
     @pytest.mark.asyncio
     async def test_stderr_suppressed(self):
         """Stderr should be suppressed."""
-        async with SilentAsyncSubprocess(sys.executable, "-c", 
-                                         "import sys; print('err', file=sys.stderr)") as proc:
+        async with SilentAsyncSubprocess(
+            sys.executable, "-c", "import sys; print('err', file=sys.stderr)"
+        ) as proc:
             await proc.communicate()
             assert proc.returncode == 0
--- a/tests/cli/unit/test_async_subprocess.py
+++ b/tests/cli/unit/test_async_subprocess.py
@@ -77,6 +77,24 @@ class TestAsyncSubprocess:
         
         assert proc.returncode is not None
 
+    @pytest.mark.asyncio
+    async def test_process_exits_immediately_with_zero(self):
+        """Process that exits immediately with code 0 should be handled."""
+        async with AsyncSubprocess(sys.executable, "-c", "exit(0)") as proc:
+            # Explicitly wait for the process to be reaped
+            await proc.wait()
+        
+        assert proc.returncode == 0
+
+    @pytest.mark.asyncio
+    async def test_process_exits_immediately_with_error(self):
+        """Process that exits immediately with error code should be handled."""
+        async with AsyncSubprocess(sys.executable, "-c", "exit(42)") as proc:
+            # Explicitly wait for the process to be reaped
+            await proc.wait()
+        
+        assert proc.returncode == 42
+
 
 class TestStdoutAsyncSubprocess:
     """Test StdoutAsyncSubprocess captures stdout."""
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -37,12 +37,18 @@ class TestBackgroundEventLoop:
         """run_fn() should schedule a callable on the loop thread."""
         results = []
 
+        results = []
+        import threading
+
+        called = threading.Event()
+
         def callback():
             results.append("called")
+            called.set()
 
         with BackgroundEventLoop() as loop:
             loop.run_fn(callback)
-            time.sleep(0.1)
+            assert called.wait(timeout=1.0)
 
         assert results == ["called"]
 
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python3
+"""Unit tests for BackgroundEventLoop async task management."""
+
+import asyncio
+import time
+
+import pytest
+
+from cli.utilities.threadutils import BackgroundEventLoop
+
+
+class TestBackgroundEventLoop:
+    """Test BackgroundEventLoop lifecycle and task draining."""
+
+    def test_context_manager_starts_and_stops(self):
+        """Context manager should start and stop the loop."""
+        with BackgroundEventLoop() as loop:
+            assert loop.loop is not None
+            assert loop.thread.is_alive()
+        
+        assert not loop.thread.is_alive()
+
+    def test_run_returns_future_with_result(self):
+        """run() should return a Future that resolves the coroutine result."""
+        async def compute():
+            await asyncio.sleep(0.01)
+            return 42
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(compute())
+            result = future.result(timeout=1.0)
+        
+        assert result == 42
+
+    def test_run_fn_schedules_callable(self):
+        """run_fn() should schedule a callable on the loop thread."""
+        results = []
+        
+        def callback():
+            results.append("called")
+        
+        with BackgroundEventLoop() as loop:
+            loop.run_fn(callback)
+            time.sleep(0.1)
+        
+        assert results == ["called"]
+
+    def test_shutdown_is_idempotent(self):
+        """Calling shutdown() twice should not crash."""
+        loop = BackgroundEventLoop()
+        loop.start()
+        loop.shutdown()
+        loop.shutdown()
+
+    def test_drain_loop_until_no_tasks(self):
+        """drain_loop_until should exit immediately with no pending tasks."""
+        async def check_drain():
+            start = time.monotonic()
+            await loop.drain_loop_until(timeout=1.0)
+            return time.monotonic() - start
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(check_drain())
+            elapsed = future.result(timeout=2.0)
+        
+        assert elapsed < 0.1
+
+    def test_drain_loop_until_cancels_pending_tasks(self):
+        """drain_loop_until should cancel and await pending tasks."""
+        async def hanging_task():
+            await asyncio.sleep(3600)
+        
+        async def run_and_drain():
+            task = asyncio.create_task(hanging_task())
+            
+            start = time.monotonic()
+            await loop.drain_loop_until(timeout=1.0)
+            elapsed = time.monotonic() - start
+            
+            return task, elapsed
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task, elapsed = future.result(timeout=2.0)
+            
+            assert task.done()
+            assert task.cancelled()
+            assert elapsed < 0.5
+
+    def test_drain_loop_until_handles_tasks_spawning_tasks(self):
+        """drain_loop_until should handle tasks that spawn new tasks during cancellation."""
+        spawned = []
+        
+        async def spawner():
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                spawned.append(asyncio.create_task(asyncio.sleep(0.01)))
+                raise
+        
+        async def run_and_drain():
+            task = asyncio.create_task(spawner())
+            await loop.drain_loop_until(timeout=1.0)
+            return task
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task = future.result(timeout=2.0)
+            
+            assert task.done()
+            for t in spawned:
+                assert t.done()
+
+    def test_drain_loop_until_times_out_on_resistant_tasks(self):
+        """drain_loop_until should timeout on tasks that resist cancellation."""
+        async def resistant_task():
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                await asyncio.sleep(3600)
+        
+        async def run_and_drain():
+            task = asyncio.create_task(resistant_task())
+            await asyncio.sleep(0.01)  # Let task start and become visible
+            
+            start = time.monotonic()
+            await loop.drain_loop_until(timeout=0.2)
+            elapsed = time.monotonic() - start
+            
+            return task, elapsed
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task, elapsed = future.result(timeout=2.0)
+            
+            assert 0.2 <= elapsed < 0.5
+
+    def test_drain_loop_until_handles_multiple_tasks(self):
+        """drain_loop_until should cancel multiple pending tasks."""
+        async def task1():
+            await asyncio.sleep(3600)
+        
+        async def task2():
+            await asyncio.sleep(3600)
+        
+        async def run_and_drain():
+            t1 = asyncio.create_task(task1())
+            t2 = asyncio.create_task(task2())
+            
+            await loop.drain_loop_until(timeout=1.0)
+            
+            return t1, t2
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            t1, t2 = future.result(timeout=2.0)
+            
+            assert t1.done() and t1.cancelled()
+            assert t2.done() and t2.cancelled()
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -36,8 +36,6 @@ class TestBackgroundEventLoop:
     def test_run_fn_schedules_callable(self):
         """run_fn() should schedule a callable on the loop thread."""
         results = []
-
-        results = []
         import threading
 
         called = threading.Event()
@@ -71,7 +69,7 @@ class TestBackgroundEventLoop:
             future = loop.run(check_drain())
             elapsed = future.result(timeout=2.0)
 
-        assert elapsed < 0.1
+        assert elapsed < 1.0
 
     def test_drain_loop_until_cancels_pending_tasks(self):
         """drain_loop_until should cancel and await pending tasks."""
@@ -94,7 +92,7 @@ class TestBackgroundEventLoop:
 
             assert task.done()
             assert task.cancelled()
-            assert elapsed < 0.5
+            assert elapsed < 1.5
 
     def test_drain_loop_until_handles_tasks_spawning_tasks(self):
         """drain_loop_until should handle tasks that spawn new tasks during cancellation."""
@@ -143,7 +141,7 @@ class TestBackgroundEventLoop:
             future = loop.run(run_and_drain())
             task, elapsed = future.result(timeout=2.0)
 
-            assert 0.2 <= elapsed < 0.5
+            assert 0.2 <= elapsed < 2.0
 
     def test_drain_loop_until_handles_multiple_tasks(self):
         """drain_loop_until should cancel multiple pending tasks."""
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -17,32 +17,33 @@ class TestBackgroundEventLoop:
         with BackgroundEventLoop() as loop:
             assert loop.loop is not None
             assert loop.thread.is_alive()
-        
+
         assert not loop.thread.is_alive()
 
     def test_run_returns_future_with_result(self):
         """run() should return a Future that resolves the coroutine result."""
+
         async def compute():
             await asyncio.sleep(0.01)
             return 42
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(compute())
             result = future.result(timeout=1.0)
-        
+
         assert result == 42
 
     def test_run_fn_schedules_callable(self):
         """run_fn() should schedule a callable on the loop thread."""
         results = []
-        
+
         def callback():
             results.append("called")
-        
+
         with BackgroundEventLoop() as loop:
             loop.run_fn(callback)
             time.sleep(0.1)
-        
+
         assert results == ["called"]
 
     def test_shutdown_is_idempotent(self):
@@ -54,35 +55,37 @@ class TestBackgroundEventLoop:
 
     def test_drain_loop_until_no_tasks(self):
         """drain_loop_until should exit immediately with no pending tasks."""
+
         async def check_drain():
             start = time.monotonic()
             await loop.drain_loop_until(timeout=1.0)
             return time.monotonic() - start
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(check_drain())
             elapsed = future.result(timeout=2.0)
-        
+
         assert elapsed < 0.1
 
     def test_drain_loop_until_cancels_pending_tasks(self):
         """drain_loop_until should cancel and await pending tasks."""
+
         async def hanging_task():
             await asyncio.sleep(3600)
-        
+
         async def run_and_drain():
             task = asyncio.create_task(hanging_task())
-            
+
             start = time.monotonic()
             await loop.drain_loop_until(timeout=1.0)
             elapsed = time.monotonic() - start
-            
+
             return task, elapsed
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(run_and_drain())
             task, elapsed = future.result(timeout=2.0)
-            
+
             assert task.done()
             assert task.cancelled()
             assert elapsed < 0.5
@@ -90,91 +93,94 @@ class TestBackgroundEventLoop:
     def test_drain_loop_until_handles_tasks_spawning_tasks(self):
         """drain_loop_until should handle tasks that spawn new tasks during cancellation."""
         spawned = []
-        
+
         async def spawner():
             try:
                 await asyncio.sleep(3600)
             except asyncio.CancelledError:
                 spawned.append(asyncio.create_task(asyncio.sleep(0.01)))
                 raise
-        
+
         async def run_and_drain():
             task = asyncio.create_task(spawner())
             await loop.drain_loop_until(timeout=1.0)
             return task
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(run_and_drain())
             task = future.result(timeout=2.0)
-            
+
             assert task.done()
             for t in spawned:
                 assert t.done()
 
     def test_drain_loop_until_times_out_on_resistant_tasks(self):
         """drain_loop_until should timeout on tasks that resist cancellation."""
+
         async def resistant_task():
             try:
                 await asyncio.sleep(3600)
             except asyncio.CancelledError:
                 await asyncio.sleep(3600)
-        
+
         async def run_and_drain():
             task = asyncio.create_task(resistant_task())
             await asyncio.sleep(0.01)  # Let task start and become visible
-            
+
             start = time.monotonic()
             await loop.drain_loop_until(timeout=0.2)
             elapsed = time.monotonic() - start
-            
+
             return task, elapsed
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(run_and_drain())
             task, elapsed = future.result(timeout=2.0)
-            
+
             assert 0.2 <= elapsed < 0.5
 
     def test_drain_loop_until_handles_multiple_tasks(self):
         """drain_loop_until should cancel multiple pending tasks."""
+
         async def task1():
             await asyncio.sleep(3600)
-        
+
         async def task2():
             await asyncio.sleep(3600)
-        
+
         async def run_and_drain():
             t1 = asyncio.create_task(task1())
             t2 = asyncio.create_task(task2())
-            
+
             await loop.drain_loop_until(timeout=1.0)
-            
+
             return t1, t2
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(run_and_drain())
             t1, t2 = future.result(timeout=2.0)
-            
+
             assert t1.done() and t1.cancelled()
             assert t2.done() and t2.cancelled()
 
     def test_drain_loop_until_handles_failing_tasks(self):
         """drain_loop_until should handle tasks that raise exceptions."""
+
         async def failing_task():
             await asyncio.sleep(0.01)
             raise ValueError("task failed")
-        
+
         async def run_and_drain():
             task = asyncio.create_task(failing_task())
             await asyncio.sleep(0.05)
-            
+
             await loop.drain_loop_until(timeout=1.0)
-            
+
             return task
-        
+
         with BackgroundEventLoop() as loop:
             future = loop.run(run_and_drain())
             task = future.result(timeout=2.0)
-            
+
             assert task.done()
             assert isinstance(task.exception(), ValueError)
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -157,3 +157,24 @@ class TestBackgroundEventLoop:
             
             assert t1.done() and t1.cancelled()
             assert t2.done() and t2.cancelled()
+
+    def test_drain_loop_until_handles_failing_tasks(self):
+        """drain_loop_until should handle tasks that raise exceptions."""
+        async def failing_task():
+            await asyncio.sleep(0.01)
+            raise ValueError("task failed")
+        
+        async def run_and_drain():
+            task = asyncio.create_task(failing_task())
+            await asyncio.sleep(0.05)
+            
+            await loop.drain_loop_until(timeout=1.0)
+            
+            return task
+        
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task = future.result(timeout=2.0)
+            
+            assert task.done()
+            assert isinstance(task.exception(), ValueError)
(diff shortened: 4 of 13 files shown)


tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python3
```
jimporter, line 1: This shebang isn't necessary, for a few reasons:

1. This file is just test definitions, and so if you run the file on its own on the command line (assuming you set `PYTHONPATH` correctly), it doesn't run any tests.
2. The file isn't executable anyway.
3. If we *did* want to run the file directly, a better shebang would be `#!/usr/bin/env pytest`. However, we can (and should) just use `pytest` on the command line to drive the tests.

We're also missing the GPL header here. Ditto for the other files.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class TestAsyncSubprocess:
+    """Test AsyncSubprocess context manager with cancellation handling."""
+
+    @pytest.mark.asyncio
+    async def test_normal_enter_exit(self):
+        """Normal enter/exit should start and terminate process cleanly."""
+        async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(0.1)") as proc:
+            assert proc is not None
```
jimporter, line 23: It would arguably be better to check that the type of `proc` matches what we expect. Otherwise, it doesn't verify anything more than the next line (`assert proc.returncode is None`), aside from a possibly-friendlier error message.

```suggestion
            assert isinstance(proc, asyncio.Process)
```

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+                await asyncio.sleep(3600)
+
+        task = asyncio.create_task(cancellable_exit())
+        await asyncio.sleep(0.1)
+        task.cancel()
+
+        try:
+            await task
+        except asyncio.CancelledError:
+            pass
```
jimporter, line 67: The only differences between these two tests are their names and docstrings:

```diff
--- /tmp/enter	2026-07-08 16:50:36.276740153 -0700
+++ /tmp/exit	2026-07-08 16:50:50.252403381 -0700
@@ -1,12 +1,12 @@
     @pytest.mark.asyncio
-    async def test_cancellation_during_enter(self):
-        """Cancellation during __aenter__ should spawn then terminate process."""
+    async def test_cancellation_during_exit(self):
+        """Cancellation during __aexit__ should still clean up process."""
 
-        async def cancellable_spawn():
+        async def cancellable_exit():
             async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(10)") as proc:
                 await asyncio.sleep(3600)
 
-        task = asyncio.create_task(cancellable_spawn())
+        task = asyncio.create_task(cancellable_exit())
         await asyncio.sleep(0.1)
         task.cancel()
```

In fact, *neither* of these tests actually test what they claim. Both are just testing what happens when you cancel during the `await asyncio.sleep(3600)` in the local function terminates. You can verify this with the following diff. If the tests did what they said, we'd reach the `assert False` calls, but we don't:

```diff
diff --git a/tests/cli/utilities/threadutils.py b/tests/cli/utilities/threadutils.py
index 72ceb0f53..8415627e0 100644
--- a/tests/cli/utilities/threadutils.py
+++ b/tests/cli/utilities/threadutils.py
@@ -172,6 +172,7 @@ class AsyncSubprocess:
             self.proc = await fut
             return self.proc
         except asyncio.CancelledError:
+            assert False
             # If cancellation hit mid-spawn, the process may already exist.
             # Finish the spawn to obtain the handle, clean it up, then re-raise.
             try:
@@ -208,6 +209,7 @@ class AsyncSubprocess:
 
                 await p.communicate()
         except asyncio.CancelledError:
+            assert False
             # Ensure the subprocess is cleaned after.
             await p.communicate()
             # We did the cleanup; now propagate cancel.
```

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+    """Test AsyncSubprocess context manager with cancellation handling."""
+
+    @pytest.mark.asyncio
+    async def test_normal_enter_exit(self):
+        """Normal enter/exit should start and terminate process cleanly."""
+        async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(0.1)") as proc:
+            assert proc is not None
+            assert proc.returncode is None
+
+        assert proc.returncode is not None
```
jimporter, line 26: We should probably be checking that the return code is zero here:

```suggestion
        assert proc.returncode == 0
```

However, this test is essentially just relying on timing, since the 0.1s sleep here is shorter than the 5s sleep in `AsyncProcess.__aexit__`. Improving the commentary here (and possibly moving this down by `test_long_running_process_terminated_on_exit`, which is the other side of this timing check) might be wise though.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+        except asyncio.CancelledError:
+            pass
+
+    @pytest.mark.asyncio
+    async def test_long_running_process_terminated_on_exit(self):
+        """Long-running process should be terminated on context exit."""
+        async with AsyncSubprocess(sys.executable, "-c", "import time; time.sleep(3600)") as proc:
+            assert proc.returncode is None
+
+        assert proc.returncode is not None
```
jimporter, line 75: This test only checks that the process exits *eventually*, but doesn't check that it actually terminates immediately once the `async with` block exits. At minimum, we should check that the returncode is non-zero (or maybe even that it's `-SIGTERM`):

```suggestion
        assert proc.returncode != 0
```

Better would be to also check that the time spent on this test is <5s.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+    async def test_process_killed_after_terminate_timeout(self):
+        """Process should be killed if terminate times out."""
+        async with AsyncSubprocess(
+            sys.executable,
+            "-c",
+            "import signal; signal.signal(signal.SIGTERM, signal.SIG_IGN); import time; time.sleep(3600)",
+        ) as proc:
+            pass
+
+        assert proc.returncode is not None
```
jimporter, line 87: Similarly here. We should check that the returncode is nonzero, or possibly even `-SIGKILL`.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @pytest.mark.asyncio
+    async def test_process_exits_immediately_with_zero(self):
+        """Process that exits immediately with code 0 should be handled."""
+        async with AsyncSubprocess(sys.executable, "-c", "exit(0)") as proc:
+            # Explicitly wait for the process to be reaped
+            await proc.wait()
+
+        assert proc.returncode == 0
```
jimporter, line 96: This test is essentially identical to `test_process_exits_immediately_with_zero` above.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+class TestStdoutAsyncSubprocess:
+    """Test StdoutAsyncSubprocess captures stdout."""
+
+    @pytest.mark.asyncio
+    async def test_captures_stdout(self):
+        """StdoutAsyncSubprocess should capture stdout."""
+        async with StdoutAsyncSubprocess(sys.executable, "-c", "print('hello')") as proc:
+            stdout, _ = await proc.communicate()
+            assert b"hello" in stdout
```
jimporter, line 116: This should probably check the stderr output:

```suggestion
            stdout, stderr = await proc.communicate()
            assert b"hello" in stdout
            assert stderr is None
```

It might also be a good idea to make the stdout check stricter, like `assert stdout.replace(b"\r", b"") == b"hello\n"` or something. (I'm guessing that we'll see `\r\n` on Windows here.) Another option would be to add `end=""` to the `print` call so we can avoid the newline complexity.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+    async def test_stderr_redirected_to_stdout(self):
+        """Stderr should be redirected to stdout."""
+        async with StdoutAsyncSubprocess(
+            sys.executable,
+            "-c",
+            "import sys; print('out'); print('err', file=sys.stderr)",
+        ) as proc:
+            stdout, _ = await proc.communicate()
+            assert b"out" in stdout
+            assert b"err" in stdout
```
jimporter, line 128: Ditto here for checking stderr, and we could do something like `assert stdout.replace(b"\r", "") == b"out\nerr\n"` for stdout.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class TestSilentAsyncSubprocess:
+    """Test SilentAsyncSubprocess suppresses output."""
+
+    @pytest.mark.asyncio
+    async def test_suppresses_all_output(self):
+        """SilentAsyncSubprocess should suppress all output."""
+        async with SilentAsyncSubprocess(sys.executable, "-c", "print('hello')") as proc:
+            await proc.communicate()
```
jimporter, line 138: We should check that this returns the right values for stdout and stderr:

```suggestion
            stdout, stderr = await proc.communicate()
            assert stdout is None
            assert stderr is None
```

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+            assert proc.returncode == 0
```
jimporter, line 139: I notice that these `SilentAsyncSubprocess` tests check the return code, but the `StdoutAsyncSubprocess` ones don't. It'd be good to pick one strategy or the other. I don't think it matters which we do too much, since this is effectively just testing post-conditions of `asyncio.Process`.

tests/cli/unit/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @pytest.mark.asyncio
+    async def test_stderr_suppressed(self):
+        """Stderr should be suppressed."""
+        async with SilentAsyncSubprocess(
+            sys.executable, "-c", "import sys; print('err', file=sys.stderr)"
+        ) as proc:
+            await proc.communicate()
```
jimporter, line 147: Same here:

```suggestion
            stdout, stderr = await proc.communicate()
            assert stdout is None
            assert stderr is None
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+class TestBackgroundEventLoop:
+    """Test BackgroundEventLoop lifecycle and task draining."""
+
+    def test_context_manager_starts_and_stops(self):
+        """Context manager should start and stop the loop."""
+        with BackgroundEventLoop() as loop:
+            assert loop.loop is not None
+            assert loop.thread.is_alive()
+
```
jimporter, line 20: This test checks the inner `loop.loop` object exists during the loop, but not that we clean it up.

```suggestion

        assert loop.loop is None
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+        assert not loop.thread.is_alive()
+
+    def test_run_returns_future_with_result(self):
+        """run() should return a Future that resolves the coroutine result."""
+
+        async def compute():
+            await asyncio.sleep(0.01)
```
jimporter, line 27: If this is just trying to suspend the current task, a more idiomatic way to do that would be a [delay of 0](https://docs.python.org/3/library/asyncio-task.html#asyncio.sleep):

> Setting the delay to 0 provides an optimized path to allow other tasks to run.

```suggestion
            await asyncio.sleep(0)
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+        with BackgroundEventLoop() as loop:
+            future = loop.run(compute())
+            result = future.result(timeout=1.0)
+
+        assert result == 42
+
+    def test_run_fn_schedules_callable(self):
+        """run_fn() should schedule a callable on the loop thread."""
+        results = []
+        import threading
```
jimporter, line 39: This belongs at the top level of the file. (Assuming we keep using `threading.Event`; see below.) I'd only expect to see a function-level `import` if we're resolving some kind of circular dependency, which would never happen with a stdlib import. (Or lazy-loading, but there's no need to worry about that here.)

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+        called = threading.Event()
```
jimporter, line 41: I'm not totally clear on why we're testing the `BackgroundEventLoop` with a threading `Event` object. It seems mostly redundant with the `results` list, especially since my reading of the `BackgroundEventLoop` implementation suggests that it should allow a few seconds to drain the task queue, thus letting `callback` run to completion without any explicit thread-synchronization.

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+        def callback():
+            results.append("called")
+            called.set()
+
+        with BackgroundEventLoop() as loop:
+            loop.run_fn(callback)
+            assert called.wait(timeout=1.0)
+
+        assert results == ["called"]
```
jimporter, line 51: A simpler way to do all of this might be to use the `mock` library:

```suggestion
        from unittest import mock  # Put this at the top level

        callback = mock.MagicMock()

        with BackgroundEventLoop() as loop:
            loop.run_fn(callback)

        callback.assert_called_once()
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def test_shutdown_is_idempotent(self):
+        """Calling shutdown() twice should not crash."""
+        loop = BackgroundEventLoop()
+        loop.start()
+        loop.shutdown()
+        loop.shutdown()
```
jimporter, line 58: This test is fine (if a bit simplistic), though there's no corresponding test that `startup` is idempotent, which it is.

It might also be good to test that `startup`, `shutdown`, `startup`, `shutdown` doesn't crash (which I think it does today), though that might be getting excessively-robust. :)

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+        async def check_drain():
+            start = time.monotonic()
+            await loop.drain_loop_until(timeout=1.0)
+            return time.monotonic() - start
+
+        with BackgroundEventLoop() as loop:
+            future = loop.run(check_drain())
+            elapsed = future.result(timeout=2.0)
+
+        assert elapsed < 1.0
```
jimporter, line 72: Running `drain_loop_until` inside a task that runs in the background loop just adds an extra layer of indirection that's not present in how this function is used in the real code. I think this would be simpler, and work more like the real implementation:

```suggestion
        with BackgroundEventLoop() as loop:
            start = time.monotonic()
            loop.run(loop.drain_loop_until(timeout=1.0)).result()
            elapsed = time.monotonic() - start

        assert elapsed < 1.0
```

That said, it might be easier to test `drain_loop_until` if it returned something, e.g. a list of all the `TimeoutError`s it encounters. That way we could rely less on timing checks. (We could also use a timeout of 0 here.)

Since we do similar things a bunch of times below, this timing check could be abstracted out into a helper function too:

```python
    @staticmethod
    def _timed_drain_loop(loop, timeout = 5.0):
        start = time.monotonic()
        loop.run(loop.drain_loop_until(timeout).result()
        return time.monotonic() - start
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+            start = time.monotonic()
+            await loop.drain_loop_until(timeout=1.0)
+            elapsed = time.monotonic() - start
+
+            return task, elapsed
+
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task, elapsed = future.result(timeout=2.0)
```
jimporter, line 91: As above, I think the indirection here is unnecessary. Even though `BackgroundEventLoop.run` returns a `Future` instead of a `Task`, we can still check everything we need with it:

```suggestion
        with BackgroundEventLoop() as loop:
            future = loop.run(hanging_task())

            start = time.monotonic()
            loop.run(loop.drain_loop_until(timeout=1.0)).result()
            elapsed = time.monotonic() - start
```

(Also as above, this could be reduced further with a helper function for draining the loop.)

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                spawned.append(asyncio.create_task(asyncio.sleep(0.01)))
+                raise
+
+        async def run_and_drain():
+            task = asyncio.create_task(spawner())
+            await loop.drain_loop_until(timeout=1.0)
+            return task
```
jimporter, line 111: Same comment as above: this would probably be better to do directly in the `with` block.

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            task = future.result(timeout=2.0)
+
+            assert task.done()
+            for t in spawned:
+                assert t.done()
```
jimporter, line 119: The `spawned` list and the `for` loop here make this test more complex but also less robust. What actually happens here is that the `except` block in `spawner` never runs, so the `spawned` list is empty and we don't assert anything.

This might be a bug in the implementation, but I poked around a bit and didn't see an obvious root cause/fix.

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def test_drain_loop_until_times_out_on_resistant_tasks(self):
+        """drain_loop_until should timeout on tasks that resist cancellation."""
+
+        async def resistant_task():
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                await asyncio.sleep(3600)
```
jimporter, line 128: This block doesn't run either, similarly to the above test. You can verify this with the following diff:

```diff
diff --git a/tests/cli/unit/test_background_event_loop.py b/tests/cli/unit/test_background_event_loop.py
index 427324b3c..b89783f20 100644
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -121,10 +121,12 @@ class TestBackgroundEventLoop:
     def test_drain_loop_until_times_out_on_resistant_tasks(self):
         """drain_loop_until should timeout on tasks that resist cancellation."""
 
+        gothere = False
         async def resistant_task():
             try:
                 await asyncio.sleep(3600)
             except asyncio.CancelledError:
+                gothere = True
                 await asyncio.sleep(3600)
 
         async def run_and_drain():
@@ -142,6 +144,7 @@ class TestBackgroundEventLoop:
             task, elapsed = future.result(timeout=2.0)
 
             assert 0.2 <= elapsed < 2.0
+            assert gothere
 
     def test_drain_loop_until_handles_multiple_tasks(self):
         """drain_loop_until should cancel multiple pending tasks."""
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+            t1 = asyncio.create_task(task1())
+            t2 = asyncio.create_task(task2())
+
+            await loop.drain_loop_until(timeout=1.0)
+
+            return t1, t2
+
+        with BackgroundEventLoop() as loop:
+            future = loop.run(run_and_drain())
+            t1, t2 = future.result(timeout=2.0)
```
jimporter, line 165: As before, this could be simplified to:

```suggestion
        with BackgroundEventLoop() as loop:
            t1 = loop.run(task1())
            t2 = loop.run(task2())

            loop.run(loop.drain_loop_until(timeout=1.0)).result()
```

tests/cli/unit/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+    def test_drain_loop_until_handles_failing_tasks(self):
+        """drain_loop_until should handle tasks that raise exceptions."""
+
+        async def failing_task():
+            await asyncio.sleep(0.01)
+            raise ValueError("task failed")
+
+        async def run_and_drain():
+            task = asyncio.create_task(failing_task())
+            await asyncio.sleep(0.05)
```
jimporter, line 179: This test is at least described in a misleading way, or else is just testing the wrong thing. With the following diff, you can see that the `failing_task` is already complete before we start draining the loop:

```diff
diff --git a/tests/cli/unit/test_background_event_loop.py b/tests/cli/unit/test_background_event_loop.py
index 427324b3c..1d033ff84 100644
--- a/tests/cli/unit/test_background_event_loop.py
+++ b/tests/cli/unit/test_background_event_loop.py
@@ -178,6 +178,7 @@ class TestBackgroundEventLoop:
             task = asyncio.create_task(failing_task())
             await asyncio.sleep(0.05)
 
+            assert not task.done()
             await loop.drain_loop_until(timeout=1.0)
 
             return task
```

Based on the description of this test, I'd expect it to be testing that we properly handle an exception raised *during* `drain_loop_until`. As-is, `drain_loop_until` doesn't need to check the failing task, as it won't be in the result of `asyncio.all_tasks`.

tests/cli/unit/test_boolean_latch.py
```diff
@@ -0,0 +1,150 @@
...
+
+import threading
+import time
+
+import pytest
+
+from cli.utilities.threadutils import BooleanLatch
+
+
+class TestBooleanLatch:
```
jimporter, line 12: Looking through the source, it seems that the only difference between a `BooleanLatch` and a standard `threading.Event` is that we can wait for the `BooleanLatch` to be *false*. Could we just rearrange the CLI test code a bit to use a `threading.Event` object instead? Then we could get rid of these tests.

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@pytest.fixture
+def controller():
+    """Factory fixture for creating MockController instances."""
+
+    def _factory(exit_code=None, exit_delay=0.0):
+        return MockController(exit_code=exit_code, exit_delay=exit_delay)
+
+    return _factory
```
jimporter, line 69: Is this necessary? We could just call `MockController` directly instead of using a fixture that returns a wrapper factory (that just forwards the arguments directly on to the actual `MockController`). If we *do* want to keep the fixture, then this would probably be simpler:

```suggestion
@pytest.fixture
def controller():
    return MockController
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+
+
+@pytest.fixture
+def controller():
+    """Factory fixture for creating MockController instances."""
+
+    def _factory(exit_code=None, exit_delay=0.0):
+        return MockController(exit_code=exit_code, exit_delay=exit_delay)
+
+    return _factory
```
jimporter, line 86: Ditto here. I don't think this is necessary.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        return self.exit_code_value
+
+    async def exit_code(self):
+        return self.exit_code_value
+
+    def supports_self_autorestart(self):
+        return self.supports_autorestart
+
+    async def wait_for_self_autorestart(self):
+        pass
```
jimporter, line 76: This is identical to the `MockController` implementation in `test_governor.py`. It should probably go into a common file so both tests can use it.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+                "wait_for_multipassd_ready",
+                side_effect=ready_side_effect,
+            )
+        )
+        stack.enter_context(
+            patch.object(governor, "on_monitor_exit", side_effect=exit_fn or noop)
+        )
+        stack.enter_context(patch("cli.controller.multipassd_governor.Session"))
+        await governor.start_async()
+    return governor
```
jimporter, line 129: This is also identical to `run_governor` in `test_governor.py`, so it could go into a common file as well.

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+from contextlib import ExitStack
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from cli.controller.multipassd_governor import MultipassdGovernor
+from cli.multipass.exceptions import TestCaseFailure, TestSessionFailure
+
+
+class MockController:
```
jimporter, line 14: This should probably inherit from `MultipassdController`:

```suggestion
class MockController(MultipassdController):
```

(Plus the necessary `import`, of course.)

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+    async def is_active(self):
+        return self.exit_code_value is None and not self._stop_event.is_set()
+
+    async def wait_exit(self):
+        if self.exit_code_value is None:
+            await self._stop_event.wait()
+            return 0
+        if self.exit_delay > 0:
+            await asyncio.sleep(self.exit_delay)
+        return self.exit_code_value
```
jimporter, line 50: I find this logic a bit strange. Looking through the tests, it seems that the `MockController` is used in two different ways:

1. When `exit_code` is `None`, that represents a controller that pretends to do some stuff until `stop` gets called.
2. When `exit_code` is not `None`, that represents a controller that exits on its own after the specified timeout.

Maybe it would be better to split this class in two, one for each behavior? Or at least explain this in more detail...

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+
+class MockController:
+    """Mock controller that simulates daemon behavior."""
+
+    def __init__(self, exit_code=None, exit_delay=0.0):
+        self.exit_code_value = exit_code
+        self.exit_delay = exit_delay
+        self.start_called = False
+        self.stop_called = False
+        self.supports_autorestart = False
```
jimporter, line 22: In `test_governor_async.py`, this gets changed to `True` in a few tests manually, e.g.
```python
        ctrl = controller(exit_code=1)
        ctrl.supports_autorestart = False
```
I think it would be cleaner to just add this as an extra argument to `__init__`.

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+def controller():
+    """Factory fixture for creating MockController instances."""
+
+    def _factory(exit_code=None, exit_delay=0.0):
+        return MockController(exit_code=exit_code, exit_delay=exit_delay)
+
+    return _factory
+
+
+async def run_governor(controller, ready_fn=None, exit_fn=None):
```
jimporter, line 72: Nothing that I see ever sets `exit_fn`. I think we can remove it.
```suggestion
async def run_governor(controller, ready_fn=None):
```

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+
+        ready_side_effect = ready_fn or hang_forever
+        if not asyncio.iscoroutinefunction(ready_side_effect):
+            sync_ready = ready_side_effect
+
+            async def ready_side_effect(*args, **kwargs):
+                result = sync_ready(*args, **kwargs)
+                if asyncio.iscoroutine(result):
+                    return await result
+                return result
```
jimporter, line 98: The only `ready_fn`s we use are the default (this) or `lambda: True`. Instead of all this logic, which is only used to make that lambda `async` (and the `if asyncio.iscoroutine(result):` is thus always false), how about passing a `ready_delay=3600` argument to `run_governor` above? Then this can go away.

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+        pass
+
+    governor = MultipassdGovernor(controller, None, print_daemon_output=False)
+    with ExitStack() as stack:
+        stack.enter_context(
+            patch.object(
+                governor, "_ensure_client_certs_are_created", new_callable=AsyncMock
+            )
+        )
+        stack.enter_context(patch.object(governor, "_authenticate_client_cert"))
```
jimporter, line 88: The `ExitStack` isn't doing much for us that I can tell. How about this:

```suggestion
    with patch.object(governor, "_ensure_client_certs_are_created",
                      new_callable=AsyncMock), \
         patch.object(governor, "_authenticate_client_cert"):
```

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+                governor,
+                "wait_for_multipassd_ready",
+                side_effect=ready_side_effect,
+            )
+        )
+        stack.enter_context(
+            patch.object(governor, "on_monitor_exit", side_effect=exit_fn or noop)
+        )
+        stack.enter_context(patch("cli.controller.multipassd_governor.Session"))
+        await governor.start_async()
```
jimporter, line 111: And then...

```suggestion
        with patch.object(governor, "wait_for_multipassd_ready",
                          side_effect=ready_side_effect), \
             patch.object(governor, "on_monitor_exit",
                          side_effect=exit_fn or noop), \
             patch("cli.controller.multipassd_governor.Session"):
            await governor.start_async()
```

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+        (0, TestCaseFailure, "multipassd died with code 0"),
+        (1, TestSessionFailure, "multipassd died with code 1"),
+        (2, TestCaseFailure, "multipassd died with code 2"),
+    ],
+)
+async def test_startup_failures(controller, exit_code, exception, message):
+    """Test governor behavior when daemon exits during startup."""
+    with pytest.raises(exception) as exc_info:
+        await run_governor(controller(exit_code=exit_code))
+    assert message in str(exc_info.value)
```
jimporter, line 129:
```suggestion
    with pytest.raises(exception, match=re.escape(message)) as exc_info:
        await run_governor(controller(exit_code=exit_code))
```

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@pytest.mark.asyncio
+async def test_successful_startup(controller):
+    """Successful daemon startup should complete without errors."""
+    governor = await run_governor(controller(exit_code=None), ready_fn=lambda: True)
```
jimporter, line 135: As mentioned above, `ready_fn` could be something like `ready_delay=0`.

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+    assert governor.controller.start_called
+
+    await governor.stop_async()
+
+
+@pytest.mark.asyncio
+async def test_stop_after_successful_startup(controller):
+    """Stopping governor after successful startup should work cleanly."""
+    governor = await run_governor(controller(exit_code=None), ready_fn=lambda: True)
+    await governor.stop_async()
```
jimporter, line 148: It might be worth merging these two tests, since they both just test the governor's startup/shutdown lifecycle.

```suggestion
```

tests/cli/unit/test_governor.py
```diff
@@ -0,0 +1,156 @@
...
+    assert governor.controller.stop_called
+
+
+@pytest.mark.asyncio
+async def test_no_cancelled_error_on_exit_code_42(controller):
+    """Ensure CancelledError is not raised when daemon exits with code 42."""
+    with pytest.raises(TestCaseFailure):
+        await run_governor(controller(exit_code=42))
```
jimporter, line 156: This is just a more-limited version of the first parameterized case of `test_startup_failures` above (but without checking the error message).

```suggestion
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        governor = await run_governor(controller(exit_code=None), ready_fn=lambda: True)
+
+        async def slow_monitor():
+            await asyncio.sleep(3600)
+
+        governor.monitor_task = asyncio.create_task(slow_monitor())
+
+        with patch("asyncio.wait_for", side_effect=asyncio.TimeoutError):
+            await governor.stop_async()
+
```
jimporter, line 147: This should probably check whether the `stop` method was called on the controller, like in the next test:

```suggestion

        assert governor.controller.stop_called
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+    @pytest.mark.asyncio
+    async def test_on_monitor_exit_cancelled_task_returns_early(self, controller):
+        """on_monitor_exit should return early if task was cancelled."""
+        governor = MultipassdGovernor(
+            controller(exit_code=42), None, print_daemon_output=False
+        )
+
+        task = MagicMock()
+        task.cancelled.return_value = True
+
```
jimporter, line 173: Minor, but something like this would be more idiomatic Python in my opinion:


```suggestion
        task = Mock(**{'cancelled.return_value': True})
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        """stop_async should wait for monitor_task to complete normally."""
+        governor = await run_governor(controller(exit_code=None), ready_fn=lambda: True)
+
+        await governor.stop_async()
+
+        assert governor.controller.stop_called
+        assert governor.monitor_task.done()
+
+
+class TestGovernorOnMonitorExit:
```
jimporter, line 161: All these tests are a strange mix of how things are set up and what gets tested. For example:
* With the `asyncio_loop` arg for `MultipassdGovernor`...
    * `test_on_monitor_exit_cancelled_task_returns_early` passes `None`
    * `test_on_monitor_exit_exit_code_42_no_autorestart_triggers_restart` passes a `MagicMock` that we test the results of
    * `test_on_monitor_exit_exit_code_42_with_autorestart_no_restart` and `test_on_monitor_exit_other_exit_code_no_restart` pass a `MagicMock` that we ignore
* All the tests except `test_on_monitor_exit_exit_code_42_no_autorestart_triggers_restart` check that the controller's `start` was *not* called. Add a check to this test (either way)?

These could all probably be parameterized with a bit of work.

jimporter:
@xmkg So you're not left wondering, I'm currently reviewing this. It's just taking a while to go through all the tests...

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+
+        task = MagicMock()
+        task.cancelled.return_value = False
+
+        await governor.on_monitor_exit(task)
+
+        assert not ctrl.start_called
+
+
+class TestGovernorReadStream:
```
jimporter, line 230: All the tests in this class (with the possible exception of `test_read_stream_handles_cancelled_error`) could fairly easily be parameterized to reduce the code duplication here. Looking over the `MultipassdGoverner._read_stream` implementation, it might also be easier to test if that function raised an exception for each of the error conditions rather than just returning a plain string.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        await asyncio.sleep(0.1)
+
+        monitor_task.cancel()
+
+        try:
+            await monitor_task
+        except asyncio.CancelledError:
+            pass
+
+        assert monitor_task.cancelled()
```
jimporter, line 343: I don't think this is actually testing anything. On line 336 it cancels `monitor_task` and then the only assertion is that `monitor_task` is indeed cancelled. Unless it had completed in the 0.1s we allowed (unlikely due to the sleep in `slow_output`), then of course it would be cancelled.

From the comment, I think the test wants to check whether the local variable `stdout_task` inside `MultipassdGovernor._monitor` gets canceled, but that task is a call to `_read_stream`, which has the following block:

```python
        except asyncio.CancelledError:
            # Handle it gracefully.
            pass
```

That is, all cancellation errors are swallowed. The following patch is super hacky and I don't recommend applying it as-is (hence no suggestion), but it sketches out a way to make this test closer to what the docstring says:

```diff
diff --git a/tests/cli/unit/test_governor_async.py b/tests/cli/unit/test_governor_async.py
index ea4809063..84640f6ea 100644
--- a/tests/cli/unit/test_governor_async.py
+++ b/tests/cli/unit/test_governor_async.py
@@ -324,11 +324,15 @@ class TestGovernorMonitor:
             controller(exit_code=None), None, print_daemon_output=False
         )
 
+        canceled = False
         async def slow_output():
-            await asyncio.sleep(3600)
-            yield
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                nonlocal canceled
+                canceled = True
 
-        governor.controller.follow_output = slow_output
+        governor._read_stream = slow_output
 
         monitor_task = asyncio.create_task(governor._monitor())
         await asyncio.sleep(0.1)
@@ -341,6 +345,7 @@ class TestGovernorMonitor:
             pass
 
         assert monitor_task.cancelled()
+        assert canceled
 
     @pytest.mark.asyncio
     async def test_monitor_includes_error_reason_in_exception(self, controller):
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+            await asyncio.sleep(3600)
+            yield
+
+        governor.controller.follow_output = cancelling_output
+
+        with pytest.raises(TestCaseFailure) as exc_info:
+            await governor._monitor()
+
+        assert "multipassd died with code 2" in str(exc_info.value)
+        assert "Reason:" not in str(exc_info.value)
```
jimporter, line 413: Since this mocks `follow_output`, we're retesting parts of `_read_stream` already covered by `TestGovernorReadStream`. You could probably simplify the mocks here (just mock `_read_stream`?) and parameterize these tests so there's less repetition.

Also as noted in a comment above, you could simplify the assertions by passing the `match` argument to `pytest.raises`, e.g.:

```python
with pytest.raises(TestCaseFailure, match="some regex"):
    await governor._monitor()
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+
+class TestGovernorStartupFailures:
+    """Test governor startup failure scenarios."""
+
+    @pytest.mark.asyncio
+    async def test_daemon_never_ready_stops_and_exits(self, controller):
+        """wait_for_multipassd_ready returning False should stop daemon and exit."""
+        ctrl = controller(exit_code=None)
+
+        governor = MultipassdGovernor(ctrl, None, print_daemon_output=False)
```
jimporter, line 424: This setup code is the same across all the tests in this class. We could just set them up in `__init__`.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+                patch.object(
+                    governor, "wait_for_multipassd_ready", side_effect=never_ready
+                )
+            )
+            stack.enter_context(
+                patch(
+                    "cli.controller.multipassd_governor.pytest.exit",
+                    side_effect=SystemExit(12),
+                )
+            )
```
jimporter, line 446: `ExitStack` isn't necessary here (it's mainly useful for doing fancy conditional things with context managers). This would be simpler:

```suggestion
        with patch.object(governor, "_ensure_client_certs_are_created"), \
             patch.object(governor, "_authenticate_client_cert"), \
             patch.object(governor, "wait_for_multipassd_ready",
                          side_effect=never_ready), \
             patch("cli.controller.multipassd_governor.pytest.exit",
                   side_effect=SystemExit(12)):
```

Also note that we don't need to pass `new_callable` when mocking `_ensure_client_certs_are_created`. Python is smart enough to automatically use `AsyncMock` when mocking an async method.

This also replicates a lot of what's in `run_governor`, so with a bit of work, we could probably call that here and eliminate most of the `patch` calls.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+                patch.object(
+                    governor,
+                    "_ensure_client_certs_are_created",
+                    new_callable=AsyncMock,
+                    side_effect=RuntimeError("cert creation failed"),
+                )
+            )
+
+            with pytest.raises(RuntimeError, match="cert creation failed"):
+                await governor.start_async()
```
jimporter, line 473: This test is really just testing a single, simple line in `MultipassdGovernor.start_async`:

```python
await self._ensure_client_certs_are_created()
```

I don't know if we really need a test for such a simple bit of code.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+            stack.enter_context(
+                patch.object(
+                    governor,
+                    "_authenticate_client_cert",
+                    side_effect=RuntimeError("auth failed"),
+                )
+            )
+
+            with pytest.raises(RuntimeError, match="auth failed"):
+                await governor.start_async()
```
jimporter, line 497: Same with this except it's now testing this very-simple line:

```python
self._authenticate_client_cert()
```

I hope we can rely on Python's exception propagation working reliably. :)

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+            stack.enter_context(patch.object(governor, "_authenticate_client_cert"))
+            stack.enter_context(
+                patch.object(
+                    governor, "wait_for_multipassd_ready", side_effect=slow_ready
+                )
+            )
+            stack.enter_context(
+                patch.object(governor, "on_monitor_exit", new_callable=AsyncMock)
+            )
+            stack.enter_context(patch("cli.controller.multipassd_governor.Session"))
```
jimporter, line 529: As before, the `ExitStack` stuff isn't necessary. Also, it would probably be helpful to check that `wait_for_multipass_ready` is properly canceled in this case:

```suggestion
        with patch.object(governor, "_ensure_client_certs_are_created"), \
             patch.object(governor, "_authenticate_client_cert"), \
             patch.object(governor, "wait_for_multipassd_ready",
                          side_effect=slow_ready) as mock_wait_ready, \
             patch.object(governor, "on_monitor_exit"), \
             patch("cli.controller.multipassd_governor.Session"):
```

(Or as above, use something like `run_governor` here.)

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+
+            start_task = asyncio.create_task(governor.start_async())
+            await asyncio.sleep(0.1)
+
+            await governor.stop_async()
+
+            with pytest.raises(TestCaseFailure, match="daemon exited during startup"):
+                await start_task
```
jimporter, line 537: ... and then:

```suggestion
                await start_task
            assert mock_wait_ready.cancelled()
```

(This emits a warning that the mocked coroutine was never awaited, but otherwise everything works. It's probably worth investigating that warning since it might indicate a minor resource bug in the test harness.)

It might also be worth checking the status of `daemon_ready_event` here.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+                    governor, "wait_for_multipassd_ready", side_effect=slow_ready
+                )
+            )
+            stack.enter_context(
+                patch.object(governor, "on_monitor_exit", new_callable=AsyncMock)
+            )
+            stack.enter_context(patch("cli.controller.multipassd_governor.Session"))
+
+            start_task = asyncio.create_task(governor.start_async())
+            await asyncio.sleep(0.1)
```
jimporter, line 532: This could be a sleep of 0s since all we're doing is giving the async runner a chance to start the `slow_ready` coroutine:

```suggestion
            await asyncio.sleep(0)
```

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+
+            await governor.stop_async()
+
+            with pytest.raises(TestCaseFailure, match="daemon exited during startup"):
+                await start_task
+
+    @pytest.mark.asyncio
+    async def test_restart_cycle_resets_state(self, controller):
```
jimporter, line 540: Why is this test in the `TestGovernorConcurrentOperations` class? Is there some concurrency I'm not seeing? I'd expect `asyncio.create_task` or `asyncio.gather` somewhere in here, as in the above test...

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        await governor.stop_async()
+
+        assert governor.graceful_exit_initiated
+        assert ctrl.stop_called
+
+        governor2 = await run_governor(ctrl, ready_fn=lambda: True)
+
+        assert governor2.daemon_ready_event.is_set()
+        assert governor2.monitor_task is not None
+        assert not governor2.graceful_exit_initiated
```
jimporter, line 558: ... but the bigger problem is that this doesn't actually call `restart_async` as the docstring implies. It just creates a second, totally-independent governor and checks some basic properties on it.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        self, controller
+    ):
+        """Daemon exit with non-1 codes should raise TestCaseFailure."""
+        for exit_code in [2, 137, 139]:
+            ctrl = controller(exit_code=exit_code)
+
+            with pytest.raises(TestCaseFailure) as exc_info:
+                await run_governor(ctrl)
+
+            assert f"multipassd died with code {exit_code}" in str(exc_info.value)
```
jimporter, line 585: These are already covered in `test_governor.py`: see `test_startup_failures`.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        class TimeoutSubprocess(MockSubprocess):
+            async def communicate(self):
+                raise asyncio.TimeoutError()
+
+        with patch(
+            "cli.controller.multipassd_governor.StdoutAsyncSubprocess",
+            return_value=TimeoutSubprocess(),
+        ):
+            result = await MultipassdGovernor.wait_for_multipassd_ready(timeout=0.5)
+            assert result is False
```
jimporter, line 633: This is a comparatively minor issue, but this test will spin since we raise a `TimeoutError` immediately. A bit of instrumentation in my checkout shows that the mock `communicate` method got called 46,986 times when running this test.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+                if call_count < 2:
+                    raise asyncio.TimeoutError()
+                return MockSubprocess(
+                    stdout=b"multipass  1.12.0\nmultipassd  1.12.0\n", returncode=0
+                )
+
+        with patch(
+            "cli.controller.multipassd_governor.StdoutAsyncSubprocess",
+            return_value=FlakySubprocess(),
+        ):
```
jimporter, line 653: Two issues here:

1. This is triggering a timeout in a different spot than the above test. I think `communicate` is the more likely place that would see a timeout, instead of `__aenter__`
2. The indexing is wrong here, though we get lucky that it doesn't cause a test failure. On the first call, we timeout (fine). On the second call, we return the multipass version info: however, this is actually the 2nd attempt of the `multipass find` call!

It might be better to mock things so that we examine the command line args that we pass in; then we can be sure to return the correct result. Or at least just fix the indexing logic.

tests/cli/unit/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        """Returns False when version output has fewer than 2 lines."""
+        find_proc = MockSubprocess(stdout=b"", returncode=0)
+        version_proc = MockSubprocess(stdout=b"multipass  1.12.0\n", returncode=0)
+
+        with patch(
+            "cli.controller.multipassd_governor.StdoutAsyncSubprocess",
+            side_effect=[find_proc, version_proc],
+        ):
+            result = await MultipassdGovernor.wait_for_multipassd_ready(timeout=1)
+            assert result is False
```
jimporter, line 684: This test might be revealing a bug in the harness code. If the `multipass version` call returns nonsense, should we keep trying in a loop, or should we immediately hard-fail in `wait_for_multipassd_ready`?

If we *should* keep trying, this test isn't quite right. After the first two calls to `StdoutAsyncSubprocess`, our mock will start raising `StopIteration` because it's out of items. That's not a behavior we'd see with the real harness, so we'd want to fix this test.

tests/cli/unit/test_wait_for_future.py
```diff
@@ -0,0 +1,82 @@
...
+            wait_for_future(future, timeout=1.0)
+
+    def test_handles_already_completed_future(self):
+        """wait_for_future should handle already-completed future."""
+        future = Future()
+        future.set_result("done")
+
+        result = wait_for_future(future, timeout=1.0)
+
+        assert result == "done"
```
jimporter, line 46: This is equivalent to `test_returns_result_when_future_completes` above.

tests/cli/unit/test_wait_for_future.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_waits_for_delayed_completion(self):
+        """wait_for_future should wait for future that completes after delay."""
+        import threading
```
jimporter, line 50: This should go at the top level. (Function-level imports are really only useful for lazy-loading or circular dependency issues.)

tests/cli/unit/test_wait_for_future.py
```diff
@@ -0,0 +1,82 @@
...
+        assert result == "delayed"
+
+    def test_cancels_future_on_timeout(self):
+        """wait_for_future should cancel future on timeout."""
+        future = Future()
+
+        with pytest.raises(TimeoutError):
+            wait_for_future(future, timeout=0.1, poll_interval=0.01)
+
+        assert future.cancelled()
```
jimporter, line 73: This is almost the same as `test_raises_timeout_error_when_future_doesnt_complete`, except it checks that the future was cancelled. We could merge the two.

tests/cli/unit/test_wait_for_future.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_respects_custom_poll_interval(self):
+        """wait_for_future should respect custom poll interval."""
+        future = Future()
+        future.set_result(42)
+
+        result = wait_for_future(future, timeout=1.0, poll_interval=0.5)
+
+        assert result == 42
```
jimporter, line 82: This doesn't actually test anything about `poll_interval` (except that it's a valid kwarg).

.github/workflows/cli-tests-ut.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.14'
+
+      - name: Install dependencies
+        run: |
+          cd tests/cli
+          python -m venv .venv
+          source .venv/bin/activate
+          pip install -e .
```
jimporter, line 32: We're in a CI runner, so there's no need to install in editable mode.

```suggestion
          pip install .
```

tests/cli/unit/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.fixture(autouse=True, scope="session")
+def ensure_multipass_binaries_are_present():
+    """Override parent fixture - unit tests don't need multipass binaries."""
+    yield
+
+
+@pytest.fixture(autouse=True, scope="session")
+def environment_setup():
+    """Override parent fixture - unit tests don't need environment setup."""
+    yield
```
jimporter, line 19: It might be better to adjust the directory structure so we don't need to do this. Otherwise, if some new autouse fixture gets added to `tests/cli/unit/conftest.py`, unexpected things could happen to the unit tests.

jimporter (changes requested):
The changes to `multipassd_governor.py` look good to me and the reasoning for them makes sense. The AI-assisted tests are a mixed bag though. There are a few stylistic issues/inconsistencies that I see, but the bigger issue is that several of the tests are wrong or misleading (e.g. not testing what they claim to test).

Sorry for the time it took to review this, but once I saw a couple misleading test cases, I wanted to be sure I had a thorough understanding of every test so that I didn't miss anything subtle.

jimporter (changes requested):
The changes to `multipassd_governor.py` look good to me and the reasoning for them makes sense. The AI-assisted tests are a mixed bag though. There are a few stylistic issues/inconsistencies that I see, but the bigger issue is that several of the tests are wrong or misleading (e.g. not testing what they claim to test).

Sorry for the time it took to review this, but once I saw a couple misleading test cases, I wanted to be sure I had a thorough understanding of every test so that I didn't miss anything subtle.

jimporter:
Hopefully my review didn't double-post. Github errored out the first time I submitted...

tests/cli/unit/test_async_subprocess.py
xmkg: done

tests/cli/unit/test_async_subprocess.py
xmkg: Done

tests/cli-test-ut/test_async_subprocess.py
```diff
@@ -0,0 +1,148 @@
...
+
+class TestSilentAsyncSubprocess:
+    """Test SilentAsyncSubprocess suppresses output."""
+
+    @pytest.mark.asyncio
+    async def test_suppresses_all_output(self):
+        """SilentAsyncSubprocess should suppress all output."""
+        async with SilentAsyncSubprocess(sys.executable, "-c", "print('hello')") as proc:
+            await proc.communicate()
+            assert proc.returncode == 0
```
xmkg: Done

tests/cli-test-ut/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+
+class TestBackgroundEventLoop:
+    """Test BackgroundEventLoop lifecycle and task draining."""
+
+    def test_context_manager_starts_and_stops(self):
+        """Context manager should start and stop the loop."""
+        with BackgroundEventLoop() as loop:
+            assert loop.loop is not None
+            assert loop.thread.is_alive()
+
```
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli-test-ut/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        """stop_async should wait for monitor_task to complete normally."""
+        governor = await run_governor(controller(exit_code=None), ready_fn=lambda: True)
+
+        await governor.stop_async()
+
+        assert governor.controller.stop_called
+        assert governor.monitor_task.done()
+
+
+class TestGovernorOnMonitorExit:
```
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli-test-ut/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+        await asyncio.sleep(0.1)
+
+        monitor_task.cancel()
+
+        try:
+            await monitor_task
+        except asyncio.CancelledError:
+            pass
+
+        assert monitor_task.cancelled()
```
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli-test-ut/test_governor_async.py
```diff
@@ -0,0 +1,684 @@
...
+            )
+            stack.enter_context(patch("cli.controller.multipassd_governor.Session"))
+
+            start_task = asyncio.create_task(governor.start_async())
+            await asyncio.sleep(0.1)
+
+            await governor.stop_async()
+
+            with pytest.raises(TestCaseFailure, match="daemon exited during startup"):
+                await start_task
```
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done

tests/cli/unit/test_governor_async.py
xmkg: Done. I've made version stricter; it now requires cli version to be always present, and validates the daemon version is valid when the daemon version is present.

tests/cli/unit/test_governor.py
xmkg: Done

tests/cli/unit/test_governor.py
xmkg: Done

tests/cli-test-ut/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+            assert called.wait(timeout=1.0)
+
+        assert results == ["called"]
+
+    def test_shutdown_is_idempotent(self):
+        """Calling shutdown() twice should not crash."""
+        loop = BackgroundEventLoop()
+        loop.start()
+        loop.shutdown()
+        loop.shutdown()
```
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli-test-ut/test_background_event_loop.py
```diff
@@ -0,0 +1,190 @@
...
+                assert t.done()
+
+    def test_drain_loop_until_times_out_on_resistant_tasks(self):
+        """drain_loop_until should timeout on tasks that resist cancellation."""
+
+        async def resistant_task():
+            try:
+                await asyncio.sleep(3600)
+            except asyncio.CancelledError:
+                await asyncio.sleep(3600)
```
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_async_subprocess.py
xmkg: Done

tests/cli/unit/test_async_subprocess.py
xmkg: Noticed that the tests are not compatible with Windows. The tests now use mocks instead.

tests/cli/unit/test_async_subprocess.py
xmkg: Done

tests/cli/unit/test_async_subprocess.py
xmkg: Done

tests/cli/unit/test_async_subprocess.py
xmkg: Done

tests/cli/unit/test_background_event_loop.py
xmkg: Done

tests/cli/unit/test_governor.py
xmkg: Added a class comment explaining the behavior.

tests/cli/unit/test_governor.py
xmkg: Done

tests/cli/unit/test_governor.py
xmkg: Done

tests/cli/unit/test_governor.py
xmkg: Done

tests/cli/unit/test_governor.py
xmkg: Done

xmkg:
Hey @jimporter, I think I've addressed all the comments for this one. Let me know if I missed anything or want something done differently.

PR #4931, opened 2026-06-06
[applevz] AZ support

sharder996:
This PR adds AZ support for the `applevz` backend in Multipass powered by Vmnet and the Apple Virtualization framework. An 'AZ' is not much different from a network bridge so there isn't much new stuff here. The most interesting bits are found in `void start_vmnet_interface(VmnetRelay& relay, const multipass::Subnet& subnet)` of `applevz_vmnet.mm` with the rest of the diff being scaffolding, code cleanup, and restructuring.

One other thing. Rather than deal with the added complexity of having the `applevz` feature work with and without the `avability_zones` feature, I've decided to make the former feature depend on the latter. `applevz` is an unreleased feature that either be released alongside AZs or after so I don't see any reason to go to the trouble of supporting both scenarios as well as the subsequent migration of VMs from a non-AZs environment to an AZ-enabled one.

#### Server setup

```bash
# Host
iperf3 -s -D --logfile /tmp/iperf3-host.log

# foo VM
multipass exec foo -- sudo systemd-run --unit=iperf3-server /usr/bin/iperf3 -s

# bar VM
multipass exec bar -- sudo systemd-run --unit=iperf3-server /usr/bin/iperf3 -s
```

---

### Latency — VM -> host (ping, 20 packets, interval 0.2 s)

```bash
multipass exec foo -- ping -c 20 -i 0.2 192.168.252.1
```

| Metric (lower is better) | applevz | qemu |
|--------------------------|--------------|-----------|
| min (ms) | 0.16 | 0.16 |
| avg (ms) | 0.55 | 1.04 |
| max (ms) | 0.92 | 3.47 |

### Latency — host -> VM (ping, 20 packets, interval 0.2 s)

```bash
ping -c 20 -i 0.2 $FOO_IP
```

| Metric (lower is better) | applevz | qemu |
|--------------------------|--------------|-----------|
| min (ms) | 0.43 | 0.46 |
| avg (ms) | 0.88 | 0.75 |
| max (ms) | 1.71 | 1.95 |

### Latency — VM -> VM (ping, 20 packets, interval 0.2 s)

```bash
multipass exec foo -- ping -c 20 -i 0.2 $BAR_IP
```

| Metric (lower is better) | applevz | qemu |
|--------------------------|--------------|-----------|
| min (ms) | 0.33 | 0.35 |
| avg (ms) | 1.06 | 1.68 |
| max (ms) | 2.34 | 4.00 |

---

### TCP Throughput — VM -> host (4 parallel streams, 10 s)

```bash
# server: host (see setup above)
multipass exec foo -- iperf3 -c 192.168.252.1 -P 4 -t 10
```

| Metric | applevz | qemu |
|--------|--------------|-----------|
| sender (Gbits/s) (higher is better) | 8.63 | 3.33 |
| receiver (Gbits/s) (higher is better) | 8.62 | 3.32 |

### TCP Throughput — host -> VM (4 parallel streams, 10 s)

```bash
# server: foo VM (see setup above)
iperf3 -c $FOO_IP -P 4 -t 10
```

| Metric | applevz | qemu |
|--------|--------------|-----------|
| sender (Gbits/s) (higher is better) | 7.23 | 5.18 |
| receiver (Gbits/s) (higher is better) | 7.22 | 5.17 |
| retransmits (lower is better) | 1,381 | ~44,000 |

### TCP Throughput — VM -> VM (4 parallel streams, 10 s)

```bash
# server: bar VM (see setup above)
multipass exec foo -- iperf3 -c $BAR_IP -P 4 -t 10
```

| Metric | applevz | qemu |
|--------|--------------|-----------|
| sender (Gbits/s) (higher is better) | 7.62 | 3.16 |
| receiver (Gbits/s) (higher is better) | 7.61 | 3.16 |
| retransmits (lower is better) | 1,106 | ~11,700 |

---

MULTI-2262

### DIFF
--- a/src/platform/backends/applevz/applevz_vmnet.mm
+++ b/src/platform/backends/applevz/applevz_vmnet.mm
@@ -192,7 +192,7 @@ void start_vmnet_interface(VmnetRelay& relay, const multipass::Subnet& subnet)
 
     xpc_object_t opts = xpc_dictionary_create(nullptr, nullptr, 0);
     xpc_dictionary_set_uint64(opts, vmnet_operation_mode_key, VMNET_SHARED_MODE);
-    const auto start_addr = (subnet.min_address()).as_string();
+    const auto start_addr = subnet.min_address().as_string();
     const auto end_addr = subnet.max_address().as_string();
     const auto mask = subnet.subnet_mask().as_string();
     xpc_dictionary_set_string(opts, vmnet_start_address_key, start_addr.c_str());
--- a/src/platform/backends/applevz/applevz_vmnet.mm
+++ b/src/platform/backends/applevz/applevz_vmnet.mm
@@ -32,6 +32,7 @@
 
 #include <applevz/applevz_utils.h>
 #include <applevz/applevz_vmnet.h>
+#include <multipass/availability_zone.h>
 #include <multipass/logging/log.h>
 
 #include <fmt/format.h>
@@ -39,6 +40,7 @@
 #include <array>
 #include <chrono>
 #include <thread>
+#include <vector>
 
 #include <errno.h>
 #include <vmnet/vmnet.h>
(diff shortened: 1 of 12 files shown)


src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -144,41 +146,60 @@ void init_buffers()
...
+
+void start_vmnet_interface(VmnetRelay& relay, const multipass::Subnet& subnet)
+{
+    relay.queue = dispatch_queue_create(
+        fmt::format("com.canonical.multipassd.vmnet.zone.{}", subnet.to_cidr()).c_str(),
+        DISPATCH_QUEUE_SERIAL);
+
+    xpc_object_t opts = xpc_dictionary_create(nullptr, nullptr, 0);
+    xpc_dictionary_set_uint64(opts, vmnet_operation_mode_key, VMNET_SHARED_MODE);
+    const auto start_addr = (subnet.min_address() + 1).as_string();
```
jimporter: I don't have a Mac to test this on, but I notice that this is different from the corresponding macOS code for the Qemu backend, which uses `subnet.min_address()` as the starting address. (If the subnet is `192.168.3.0/24`, `min_address` should return an IP of `192.168.3.1`.)

jimporter (changes requested):
While I'm not able to try this out locally, this looks good to me, with the possible exception of one comment below.

src/platform/backends/applevz/applevz_vmnet.mm
sharder996: I think I followed:
https://github.com/canonical/multipass/blob/dac5019d70568ae764636fb0e8d4356ee9856968/src/platform/backends/qemu/linux/dnsmasq_process_spec.cpp#L34-L36

Is this a bug then? Or, am I misunderstanding this?

jimporter: There may well be a bug somewhere in here. The code I was looking at was here, which doesn't increment the min address: https://github.com/canonical/multipass/blob/dac5019d70568ae764636fb0e8d4356ee9856968/src/platform/backends/qemu/macos/qemu_platform_macos.cpp#L104-L109

I'll take a look at what's going on (at least on Linux) to verify that things are correct here. It may just be that the arguments are slightly different on Linux and macOS. (An IP like `196.168.3.1` is often just the gateway address, but I don't know off-hand whether these APIs expect the start address to be the gateway address or the first *non*-gateway address.)

jimporter: Doing a bit of archaeology, it's been this way for a while. The macOS code uses a start address of `x.y.z.1`, starting from when we added code to specify the address (`release/1.16` here): https://github.com/canonical/multipass/blob/c8ab690c06ee555de4f5d8659d606d160d29f91e/src/platform/backends/qemu/macos/qemu_platform_detail_macos.cpp#L31-L34

On Linux, we've passed `x.y.z.2` as the start address to `dnsmasq` since 2019 (`release/1.10` here): https://github.com/canonical/multipass/blob/5b05fb6f740558528199837a0c775316a2ace066/src/platform/backends/qemu/linux/dnsmasq_process_spec.cpp#L40-L42

On Linux, we also explicitly pass the gateway IP plus the subnet here (`main` branch here): https://github.com/canonical/multipass/blob/987dd25281e45f1e1c324fa853c95552abd24180/src/platform/backends/qemu/linux/qemu_platform_linux.cpp#L73

src/platform/backends/applevz/applevz_vmnet.mm
```diff
@@ -144,41 +146,60 @@ void init_buffers()
...
+
+void start_vmnet_interface(VmnetRelay& relay, const multipass::Subnet& subnet)
+{
+    relay.queue = dispatch_queue_create(
+        fmt::format("com.canonical.multipassd.vmnet.zone.{}", subnet.to_cidr()).c_str(),
+        DISPATCH_QUEUE_SERIAL);
+
+    xpc_object_t opts = xpc_dictionary_create(nullptr, nullptr, 0);
+    xpc_dictionary_set_uint64(opts, vmnet_operation_mode_key, VMNET_SHARED_MODE);
+    const auto start_addr = (subnet.min_address()).as_string();
```
jimporter: Tiny nitpick:

```suggestion
    const auto start_addr = subnet.min_address().as_string();
```

jimporter (dismissed):

jimporter (approved):

ricab:
@sharder996, @jimporter Is there anything specific that I should focus my review on for this?

sharder996:
I will let @jimporter override, but as per a Mattermost discussion:

> I'm ok with merging this without a secondary reviewer, since it's still experimental.

ricab:
OK, I am good with that too.

PR #4951, opened 2026-06-12
[ssh-process] Remove move operators and improve libssh error handling

tobe2098:
# Description

- What does this PR do? Make SSHProcess moved-from objects safe from UB.

## Checklist

MULTI-2657


ricab:
As discussed, please assign me back when it's ready.

ricab:
Hey @tobe2098, I had to move the base, so this needs a rebase. Let me know if you need a hand with that.

tobe2098:
Done @ricab

src/ssh/plain_ssh_process.cpp
```diff
@@ -97,6 +97,35 @@ mp::PlainSSHProcess::PlainSSHProcess(ssh_session session,
...
 
+mp::PlainSSHProcess::PlainSSHProcess(mp::PlainSSHProcess&& other)
+    : session_lock{std::move(other.session_lock)},
+      session{std::move(other.session)},
+      cmd{std::move(other.cmd)},
+      channel{std::move(other.channel)},
+      exit_result{std::move(other.exit_result)}
+{
+    // Moved-from object safety
+    other.session = nullptr; // libssh functions return SSH_ERROR
```
ricab: Although many libssh functions protect against nullptr and return SSH_ERROR, I am not sure we should just rely on that, for a few reasons:

- guarantees are often missing from their documentation, so the protection against nullptr may be called a robustness measure, but is it part of the contract?
- more importantly, we're not always checking for those errors to fail early :/ that's wrong regardless of move and I wouldn't mind if you fixed it. As it is, we get into weird states that the code is probably not meant to be in. For example, looking at just the first member function implemented in this file (`mp::PlainSSHProcess::exit_recognized`), it (indirectly) calls `ssh_add_channel_callbacks` with a null channel, which *should* return an error, but we happily proceed to `ssh_event_new`, `ssh_event_add_session`, `ssh_event_dopoll`... Only then do we handle the error.
- if there is no explicit protection against nullptrs in existing code, people are more likely to forget that case when introducing new code.

This should be relatively easy to fix with a few `if (!session)` in public methods. For example, `exit_recognized()` could begin with `if (!session) return true;`. `exit_code()` could `if (!session) return -42;`. Just need to keep things consistent (*valid but unspecified state*).

ricab (commented):
This probably works today, but it is not super easy to determine that with certainty and it would be easy to break things in the future. I am proposing additional checks below. Let me know what you think.

src/ssh/plain_ssh_process.cpp
tobe2098: All the functions called in SSHProcess do return SSH_ERROR on null argument from libssh, although I am not against checking the in the class itself.

Yes, an earlier iteration of the PR did fix the missing checks, but tests broke in the SSHFS department. I will see if the new changes have improved the situation.

I will fix those. My opinion is that a moved-from object should always "fail" when calling its methods (`false`, `throw`, etc...). WDYT?

ricab: Let me start by saying that this is a very tricky issue, one I have always struggled with in C++ (I think the language is just poorly designed in this respect).

See https://en.cppreference.com/cpp/utility/move:

> Unless otherwise specified, all standard library objects that have been moved from are placed in a "valid but unspecified state", meaning the object's class invariants hold (so functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from)

That is the golden standard in C++. In practice, you can use the object for anything that doesn't need it to be in a particular state (like assigning, deleting...). For other things, you define a state that is consistent with the invariants of the class and react accordingly.

This is slightly different from *a moved-from object should always fail* because that could break invariants. For example, suppose that the `SSHProcess` had the method `is_running()`. You'd probably want the invariant that  `is_running() != exit_recognized()`. Then you couldn't just return false from both. If you wanted to throw, you would need to extend that invariant to `is_running() != exit_recognized() || throws(is_running())`, and you could no longer declare `is_running` `noexcept`.

See the FAQ section in here https://herbsutter.com/2020/02/17/move-simply/

He argues that it is simple... IMO, it is anything but. He essentially recommends choosing one of:

- not supporting move
- having a default state (probably with a default constructor)
- support move as copy

IMO, the language is pretty broken in this respect. It forces people to choose between null and move: hence unique_ptrs to null. Herb says its unreasonable to ask people to always check the value before using it, but that is exactly what a null unique_ptr is. It just adds the precondition `not null` to anything that involves dereferencing. C++ is clearly behind Rust here.

In practice, we can get away by representing some no-op finished process (`:`):

- bool exit_recognized(timeout) -> immediate true (any state returning false would only do so after timeout)
- int exit_code(timeout) -> 0;
- read_std_output(), read_std_error() -> ""
- get_cmd() -> ":"

The alternative is the null approach, where callers can either assert it is non-null or need to check first. But then you need a way to check, i.e. an `is_valid()`.

The more I think about it, the more I lean toward just noncopyable.

ricab: Note that PlainSSHSession is much simpler because the interface is basically (once we remove that operator ssh_session):

- is_connected() - returns false if moved
- exec() - already has the precondition that it must be connected
- force_shutdown() - idempotent

The default state is already part of our problem domain (part of the set of possible SSHSession values) because we can't guarantee that a session is connected (depends on the other side), so we need to deal with it programmatically.

tests/unit/test_sftpserver.cpp
```diff
@@ -405,38 +440,56 @@ TEST_F(SftpServer, throwsOnSshFailureReadExit)
...
+        if (num_calls >= total_calls / 2) // The status has to only be retrieved (as a successful
+                                          // exit) once the loop is in its second recovery stage
+            // Otherwise we time out. Because of how we cache, we need to set up the return code on
+            // process creation after the first timeout
+            cb->channel_exit_status_function(nullptr,
+                                             nullptr,
+                                             exit_status_mock.success_status,
+                                             cb->userdata);
         return SSH_OK;
     };
```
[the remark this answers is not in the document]
tobe2098: This is not true due to a bug with `release_channel` and SSHProcess exit_code caching

tests/unit/test_sftpserver.cpp
```diff
@@ -405,38 +440,56 @@ TEST_F(SftpServer, throwsOnSshFailureReadExit)
...
-    auto sftp = make_sftpserver();
 
+    REPLACE(ssh_channel_new,
+            [](auto...) { return reinterpret_cast<ssh_channel>(0xdeadbeefdeadbeef); });
+    REPLACE(ssh_channel_free, [](auto...) { return; });
+    REPLACE(ssh_remove_channel_callbacks, [](auto...) { return SSH_OK; });
+    REPLACE(ssh_event_new, [](auto...) { return reinterpret_cast<ssh_event>(0xdeadbeefdeadbeef); });
+    REPLACE(ssh_event_free, [](auto...) { return; });
+    REPLACE(ssh_event_add_session, [](auto...) { return SSH_OK; });
+    REPLACE(ssh_event_dopoll, [](auto...) { return SSH_OK; });
```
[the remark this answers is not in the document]
tobe2098: This is exactly what the test is emulating, a timeout via underlying network activity in libssh

tobe2098:
Copilot review is over

src/ssh/plain_ssh_process.cpp
```diff
@@ -45,11 +45,16 @@ class ExitStatusCallback
         ssh_callbacks_init(&cb);
         cb.channel_exit_status_function = channel_exit_status_cb;
         cb.userdata = &result_holder;
-        ssh_add_channel_callbacks(channel, &cb);
+        registered = ssh_add_channel_callbacks(channel, &cb);
```
ricab, line 48: Shouldn't we throw if this fails?

ricab: Ties in with https://github.com/canonical/multipass/issues/4990. If we can't register the callback at construction, we haven't succeeded in properly creating an SSH process and should unwind, right?

src/ssh/plain_ssh_process.cpp
```diff
@@ -94,7 +100,7 @@ mp::PlainSSHProcess::PlainSSHProcess(ssh_session session,
       channel{make_channel(session, cmd)},
       exit_result{}
 {
-    assert(this->session_lock.owns_lock());
+    assert(this->session_lock.owns_lock() && session != nullptr);
```
ricab: If you add this to the assert, you are signaling that it is a precondition of the constructor that the session is not null. That should then be documented in the function signature. 

However, there is probably a better way to make sure this isn't null: receive a reference to the struct instead of a pointer. The caller (PlainSSHSession) already asserts non-null as part of the not-moved precondition to exec. You can always take the thing's address internally and keep asserting. That's then an internal invariant (not public anymore). Or you could keep a reference field and take the address to give to libssh.

src/ssh/plain_ssh_process.cpp
```diff
@@ -130,11 +136,39 @@ int mp::PlainSSHProcess::exit_code(std::chrono::milliseconds timeout)
 void mp::PlainSSHProcess::read_exit_code(std::chrono::milliseconds timeout, bool save_exception)
 {
     assert(std::holds_alternative<std::monostate>(exit_result));
-    ExitStatusCallback cb{channel.get(), exit_result};
+    std::exception_ptr eptr;
+    if (!channel)
```
ricab: The only way the channel could be null is if it was released. I am throwing `release_channel()` away entirely. `!!channel` will be an invariant of the class that we can assert then.

Would you mind marking this code with a TODO@sftp for me? Thanks.

src/ssh/plain_ssh_process.cpp
```diff
@@ -130,11 +136,39 @@ int mp::PlainSSHProcess::exit_code(std::chrono::milliseconds timeout)
...
+            const char* err = ssh_get_error(session);
+            cause = (err && *err) ? err : "could not add session to event";
+        }
+        eptr = std::make_exception_ptr(SSHProcessExitError{cmd, cause});
+
+        if (save_exception)
+            exit_result = eptr;
+
+        std::rethrow_exception(eptr);
+    }
```
ricab: We can avoid nesting ifs and checking conditions multiple times:


```suggestion
    std::optional<std::string> err = nullopt;
    if (!event)
        err = "could not allocate event";
    else if (!cb.is_registered())
        err = "could not register callback";
    else if ((ssh_event_add_session(event.get(), session) != SSH_OK))
    {
        const char* raw_err = ssh_get_error(session);
         err = (err && *err) ? err : "could not add session to event";
    }

    if (err)
    {
        eptr = std::make_exception_ptr(SSHProcessExitError{cmd, err});

        if (save_exception)
            exit_result = eptr;

        std::rethrow_exception(eptr);
    }
```

(code untested)

ricab: Probably also worth extracting into a helper (maybe a private method).

src/ssh/plain_ssh_process.cpp
```diff
@@ -147,18 +181,21 @@ void mp::PlainSSHProcess::read_exit_code(std::chrono::milliseconds timeout, bool
 
     if (!std::holds_alternative<int>(exit_result))
     {
-        std::exception_ptr eptr;
-        if (rc == SSH_ERROR) // we expect SSH_AGAIN or SSH_OK (unchanged) when there is a timeout
-            eptr = std::make_exception_ptr(SSHProcessExitError{cmd, std::strerror(errno)});
+        if (rc == SSH_ERROR)
+        { // we expect SSH_AGAIN or SSH_OK (unchanged) when there is a timeout
+            const char* err = ssh_get_error(session);
```
ricab: Hmm, the [documentation](https://api.libssh.org/stable/group__libssh__poll.html#gada90c0ca6919271708ba1ea0352632cb) says:

> Check errno to get more details about why it failed

src/ssh/plain_ssh_process.cpp
```diff
@@ -182,9 +219,9 @@ std::string mp::PlainSSHProcess::read_stream(StreamType type, int timeout)
     mpl::trace_location(category, "(type = {}, timeout = {})", static_cast<int>(type), timeout);
 
     // If the channel is closed there's no output to read
-    if (ssh_channel_is_closed(channel.get()))
+    if (!channel || ssh_channel_is_closed(channel.get()))
```
ricab: Could I get a TODO here too please?

ricab (commented):
Much improved, good job! 

A few questions and adjustment suggestions inline.

src/ssh/plain_ssh_process.cpp
tobe2098: We should throw, but only if this is done in the constructor of PlainSSHProcess. As of now the channel callback addition error could be caused by a number of things, since we add it relatively late and we do so multiple times. I think checking the registration on `read_exit_code` and throwing there is a good approach until we change how the class behaves

src/ssh/plain_ssh_process.cpp
tobe2098: The reference seems like a good approach

src/ssh/plain_ssh_process.cpp
tobe2098: No problem!

src/ssh/plain_ssh_process.cpp
tobe2098: I am aware of this, and for some time I went with `errno`, but during test fixing some of the errors were "No such file or directory", some of the errors are not very informative. ssh_get_error is straight up incorrect in this case though.

src/ssh/plain_ssh_process.cpp
tobe2098: :+1:

src/ssh/plain_ssh_process.cpp
tobe2098: I think the helper will be hard to understand, but the new structure is better

ricab: I don't really understand why, but seeing as this code will need revisiting and part will be removed, I'll accept it.

ricab (approved):
Yups! Good job!

PR #4960, opened 2026-06-16
[qemu] Disable GTK/SDL display backends

sharder996:
Disable the GTK/SDL display backends in QEMU. These display rely on X11 dependencies that are not present in the Multipass snap, causing VM launch failures. See below:

```
[2026-06-16T13:06:47.914] [error] [qemu factory] Qemu fail: 'Process returned exit code: 127' with outputs:

/snap/multipass/17453/usr/bin/qemu-system-x86_64: error while loading shared libraries: libX11.so.6: cannot open shared object file: No such file or directory

launch failed: Internal error: qemu-system-x86_64 failed getting vmstate (Process returned exit code: 127) with output:
/snap/multipass/17453/usr/bin/qemu-system-x86_64: error while loading shared libraries: libX11.so.6: cannot open shared object file: No such file or directory
```

### DIFF
--- a/packaging/gui-less/snap/snapcraft.yaml
+++ b/packaging/gui-less/snap/snapcraft.yaml
@@ -268,6 +268,9 @@ parts:
     - --disable-libvduse
     - --disable-vduse-blk-export
     - --disable-dbus-display
+    - --disable-gtk
+    - --disable-sdl
+    - --disable-sdl-image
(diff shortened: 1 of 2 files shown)


xmkg:
Self-reminder: Add these to #4799

xmkg (approved):

PR #4961, opened 2026-06-17
Bugfix/cli tests win fix event loop freeze

xmkg:
Win: Avoid blocking the event loop on close
    
EvtClose can wait for in-flight callbacks, stalling pytest daemon teardown. Close the subscription on a detached daemon thread so pytest daemon teardown is not stalled if it blocks.

### DIFF
--- a/tests/cli/controller/winsvc_multipassd_controller.py
+++ b/tests/cli/controller/winsvc_multipassd_controller.py
@@ -39,7 +39,7 @@ from cli.utilities import (
     StdoutAsyncSubprocess,
     SilentAsyncSubprocess,
     sudo,
-    run_detached_thread
+    run_detached_thread,
 )
 from cli.config import cfg
 from .controller_exceptions import ControllerPrerequisiteError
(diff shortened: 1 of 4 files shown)


ricab (commented):
I am OK with this modulo copilot flaggings.

tests/cli/controller/winsvc_multipassd_controller.py
```diff
@@ -147,22 +147,34 @@ def callback(action, _, handle):
...
+                SignalEvent=signal_event,
+                Query=query,
+                Callback=callback,
+                Flags=flags
+            )
+        except Exception:
+            # Avoid leaking the publisher metadata handle if the subscription
+            # creation fails (e.g., permissions/invalid query).
+            with suppress(Exception):
+                pubmeta.Close()
```
ricab, line 176: Couldn't pubmeta be opened in a with statement instead? Then the whole block would be covered.

ricab (commented):
Small suggestion/question inline.

ricab:
Also, the last commit is unverified. Would that be easy to fix @xmkg ?

tests/cli/controller/winsvc_multipassd_controller.py
xmkg: EvtOpenPublisherMetadata does not yield a generator, so it's not consumable with. There's another way: use the closing(expr), but it requires the object to implement the .close() method. Sadly, the object has `.Close()` instead.

xmkg:
> Also, the last commit is unverified. Would that be easy to fix @xmkg ?

Fixed

tests/cli/controller/winsvc_multipassd_controller.py
```diff
@@ -147,22 +147,34 @@ def callback(action, _, handle):
...
 
-                # Hand off to asyncio loop
-                loop.call_soon_threadsafe(
-                    q.put_nowait, msg.replace("\r\n", "\n"))
+                # Hand off to asyncio loop. In-flight callbacks can outlive the
+                # async generator teardown, so re-check `closing` and guard the
+                # scheduling in case the loop is already closing.
+                if closing.is_set():
+                    return
+                with suppress(RuntimeError):
```
ricab, line 155: I don't suppose there is a more specific type this could catch, to avoid sinking other problems?

tests/cli/controller/winsvc_multipassd_controller.py
ricab: OK, fine then.

ricab (commented):
Another detail I didn't notice before. Otherwise alright.

tests/cli/controller/winsvc_multipassd_controller.py
xmkg: Unfortunately, no. `call_soon_threadsafe` raises a RuntimeError when the loop is closed:

https://github.com/python/cpython/blob/9751e1d7df5f1ed0e1520cd24b16750b173c2fbf/Lib/asyncio/base_events.py#L876

https://github.com/python/cpython/blob/9751e1d7df5f1ed0e1520cd24b16750b173c2fbf/Lib/asyncio/base_events.py#L554

ricab: Alright, you did what you could then.

ricab (approved):
Thanks @xmkg, LGTM!

PR #4965, opened 2026-06-18
Add a couple of CPP rules to CONTRIBUTING.md

ricab:
Add a couple of rules for C++ development to our CONTRIBUTING.md.

I was doing some local branch cleanup and I noticed I had this around. Let me know your thoughts.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -293,9 +293,8 @@ not use platform `#ifdef`s (or other platform-conditional logic) outside of thos
 **CPP12.** Declare generic constants in a dedicated header.<br>
 **CPP13.** Avoid compilation warnings.<br>
 **CPP14.** To mock free functions and external APIs, wrap them with `MockableSingleton`.<br>
-**CPP15.** Avoid unconstrained `auto` return types in publicly visible methods and functions,
-outside tests, the anonymous namespace, unnamed lambdas, and templates with dependent return
-types.<br>
+**CPP15.** Avoid unconstrained `auto` return types in publicly visible methods and functions (except in
+tests, the anonymous namespace, unnamed lambdas, and templates with dependent return types).<br>
 **CPP16.** Prefer exceptions for error handling.<br>
 
 ### Text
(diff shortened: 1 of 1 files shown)


ricab:
Hey @jimporter, I was looking back at my branches and noticed this pending for a while now. Would you mind if I re-assigned review to you?

Feel free to bring others into the discussion!

CONTRIBUTING.md
```diff
@@ -293,6 +293,9 @@ not use platform `#ifdef`s (or other platform-conditional logic) outside of thos
 **CPP12.** Declare generic constants in a dedicated header.<br>
 **CPP13.** Avoid compilation warnings.<br>
 **CPP14.** To mock free functions and external APIs, wrap them with `MockableSingleton`.<br>
+**CPP15.** Avoid unconstrained `auto` return types in publicly visible methods and functions (except in
+tests, the anonymous namespace, unnamed lambdas, and templates with dependent return types).<br>
```
jimporter, line 297: I'm not necessarily opposed to this, but I do personally find `auto` return useful in a few cases, mainly if the actual return type is some extremely-long template type, and the function body is a) inline and b) simple enough that the return type is obvious to a human reader, e.g.:

```c++
template<typename ...T>
auto do_something() {
  some_type<some_really_complex_template<T...>> result;
  result.do_something();
  return result;
}
```

However, that's a pretty rare case, and I'm not sure it's worth crafting the perfect guideline to accommodate this.

ricab: Hey @jimporter, thanks for the review! 

I tend to think that a using declaration might still be preferable for something public, but in any case, wouldn't that fall in the "templates with dependent return types" exception? 

Maybe we can loosen the guideline a little without having to specify every possible exception. How about this?

```suggestion
**CPP15.** Avoid unconstrained `auto` return types in publicly visible methods and functions, unless 
there is a strong reason to have it (e.g. templates with complex dependent return types).<br>
```

In any case, note that META9 still applies:

> META9. Established guidelines are taken seriously, but with a grain of salt. They are guidelines after all, not absolute rules.

The "avoid" phrasing is meant to leave a margin for exceptions. It can all be argued during review.

jimporter: Fair enough, that wasn't the best example. I'm ok with this as-is then.

jimporter (approved):

CONTRIBUTING.md
ricab: Thank you for the review!

ricab:
Alright, I think this has simmered for long enough, so merging.

PR #4967, opened 2026-06-19
[hyperv-api] Add support for AZs

jimporter:
# Description

This adds the necessary plumbing to the Hyper-V API backend to support AZs. This entails adding a new virtual network with the appropriate subnet, plus DHCP for the instances to get an IP address.

## Testing

- Hyper-V API VM creation unit tests updated to check for the new virtual network
- Manual testing steps:

  1. Start the multipass daemon
  2. Run `multipass launch --name foo`
  3. Ensure you can connect over SSH via `multipass shell foo`
  4. (optional) Try again with `multipass launch --name bar --zone zone2 && multipass shell bar`

## Checklist

### DIFF
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
@@ -58,7 +58,6 @@ using hcs::HCS;
 using virtdisk::VirtDisk;
 
 constexpr auto log_category = "HyperV-Virtual-Machine-Factory";
-constexpr auto default_hyperv_switch_guid = "C08CB7B8-9B3C-408E-8E30-5E16A3AEB444";
 constexpr auto extra_interface_vswitch_name_fmtstr = "Multipass vSwitch ({})";
 /**
  * Regex pattern to extract the origin network name and GUID from an extra interface
@@ -72,7 +71,8 @@ HCSVirtualMachineFactory::HCSVirtualMachineFactory(const Path& data_dir,
           MP_UTILS.derive_instances_dir(data_dir,
                                         HCSVirtualMachineFactory::get_backend_directory_name(),
                                         instances_subdir),
-          az_manager)
+          az_manager),
+      az_network_guids{create_az_bridges(az_manager.get_zones())}
 {
 }
 
@@ -81,7 +81,7 @@ VirtualMachine::UPtr HCSVirtualMachineFactory::create_virtual_machine(
     const SSHKeyProvider& key_provider,
     VMStatusMonitor& monitor)
 {
-    return std::make_unique<HCSVirtualMachine>(default_hyperv_switch_guid,
+    return std::make_unique<HCSVirtualMachine>(az_network_guids.at(desc.zone),
                                                desc,
                                                monitor,
                                                key_provider,
@@ -341,6 +341,36 @@ void HCSVirtualMachineFactory::hypervisor_health_check()
                 state == WindowsFeatureState::Absent ? "Absent" : "Disabled"};
         }
     }
+}
+
+std::unordered_map<std::string, std::string> HCSVirtualMachineFactory::create_az_bridges(
+    const AvailabilityZoneManager::Zones& zones)
+{
+    std::unordered_map<std::string, std::string> az_mapping;
+    for (const auto& i : zones)
+    {
+        const auto& zone = i.get();
+        hcn::CreateNetworkParameters network_params{
+            .name = fmt::format("Multipass vNetwork ({})", zone.get_name()),
+            .type = hcn::HcnNetworkType::Ics(),
+            .flags = hcn::HcnNetworkFlags::enable_dhcp_server,
+            .guid = utils::make_uuid(network_params.name),
+            .ipams = {{.type = hcn::HcnIpamType::Static(),
+                       .subnets = {{.ip_address_prefix = {zone.get_subnet().to_cidr()}}}}}};
+
+        const auto create_network_result = HCN().create_network(network_params);
+        if (!create_network_result &&
+            static_cast<HRESULT>(create_network_result.code) != HCN_E_NETWORK_ALREADY_EXISTS)
+        {
+            throw CreateNetworkException{"Could not create network for {}, status: {}",
+                                         zone.get_name(),
+                                         create_network_result};
+        }
+
+        az_mapping.emplace(zone.get_name(), network_params.guid);
+    }
+
+    return az_mapping;
 }
 
 } // namespace multipass::hyperv
--- a/src/platform/backends/hyperv_api/hcn/hyperv_hcn_ipam_type.h
+++ b/src/platform/backends/hyperv_api/hcn/hyperv_hcn_ipam_type.h
@@ -43,6 +43,11 @@ struct HcnIpamType : FormatAsMixin<HcnIpamType>
         return {"static"};
     }
 
+    [[nodiscard]] bool operator==(const HcnIpamType& rhs) const
+    {
+        return value == rhs.value;
+    }
+
 private:
     HcnIpamType(std::string_view v) : value(v)
     {
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
@@ -425,6 +425,12 @@ void HCSVirtualMachine::set_state(hcs::ComputeSystemState compute_system_state)
 {
     mpl::debug(get_name(), "set_state() -> HCS state `{}`", compute_system_state);
 
+    if (state == State::unavailable)
+    {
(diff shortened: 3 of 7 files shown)


src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
```diff
@@ -81,7 +82,7 @@ VirtualMachine::UPtr HCSVirtualMachineFactory::create_virtual_machine(
     const SSHKeyProvider& key_provider,
     VMStatusMonitor& monitor)
 {
-    return std::make_unique<HCSVirtualMachine>(default_hyperv_switch_guid,
+    return std::make_unique<HCSVirtualMachine>(az_network_guids.at(desc.zone),
```
xmkg, line 84: The default_hyperv_switch_guid is probably unused now -- we can remove it.

src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
```diff
@@ -343,4 +344,34 @@ void HCSVirtualMachineFactory::hypervisor_health_check()
...
 
+std::unordered_map<std::string, std::string> HCSVirtualMachineFactory::create_az_bridges(
+    const AvailabilityZoneManager::Zones& zones)
+{
+    std::unordered_map<std::string, std::string> az_mapping;
+    for (const auto& i : zones)
+    {
+        const auto& zone = i.get();
+        hcn::CreateNetworkParameters network_params{
+            .name = fmt::format("Multipass vNetwork ({})", zone.get_name()),
```
xmkg, line 354: The default vSwitch had its DNS suffix set to .mshome.net, which allowed users to do this:

```
ssh test-vm.mshome.net
```

Also, this is how we discover the VMs assigned IP right now, and since the new AZ networks have no DNS suffix, we can't use it anymore, and VMs get stuck at start:

```
[2026-06-25T11:25:32.476] [debug] [HyperV-HCS-Wrapper] get_compute_system_state(...) > handle: (0x24db33f7eb0)
[2026-06-25T11:25:32.476] [trace] [HyperV-HCS-Wrapper] create_operation(...)
[2026-06-25T11:25:32.476] [debug] [HyperV-HCS-Wrapper] perform_hcs_operation(...) > result: true
[2026-06-25T11:25:32.476] [debug] [HyperV-HCS-Wrapper] wait_for_operation_result(...) > (0x24db32d8fd0), timeout: 240000 ms
[2026-06-25T11:25:32.476] [debug] [HyperV-HCS-Wrapper] wait_for_operation_result(...) > finished (0x24db32d8fd0), result_code: 0x0: The operation completed successfully.
[2026-06-25T11:25:32.476] [debug] [test-vm] set_state() -> HCS state `running`
[2026-06-25T11:25:32.476] [trace] [test-vm] ssh connection failed: 'Failed to resolve hostname test-vm.mshome.net
```

To support this, we need to add the DNS suffix parameter to the create_network call. It overlaps with the hostname resolution task I have. I'll update the HCN code in a separate PR to support this. It should be trivial -- I'll ping you once it's done.

xmkg (commented):
The code looks good. We have a regression due to missing DNS suffixes on the new AZ networks. It overlaps with the hostname resolution task on Windows, so I'll prioritize it today to unblock this.

src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
xmkg: Alternatively, we can use the HCN API to query the assigned IP address to the endpoint, but it would require more work. I'd say it would be a better choice -- something to consider in the future.

xmkg: I've opted to implement querying the assigned IP from HCN instead. resolve_hostname approach is fragile since the hosts.ics file is plagued with a 25-year-old corruption bug.  

https://github.com/canonical/multipass/pull/5097

xmkg:
@jimporter https://github.com/canonical/multipass/pull/5097 ~~needs to land before this, since the hostname resolution is broken in this branch.~~ #5097 removes the reliance on hostname resolution altogether for IP discovery.

ricab:
Adding this to the milestone, as I understand this is needed in the release. Please revert if my understanding is wrong.

src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
jimporter: Hmm, `<instance>.mshome.net` works fine for me (see below). Maybe you already fixed it, or there's just some configuration difference between our systems? My test was just on a mostly-stock Windows 11 VM. (I had previously enabled `HKML\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\IPEnableRouter` because I think we need that for cross-zone communication, but pinging by hostname works for if I reset that value and restart.)

```
C:\Users\Jim\Documents\multipass\build>bin\multipass ls
bar                     Running           10.97.1.223      Ubuntu 26.04 LTS    zone2
foo                     Running           10.97.0.53       Ubuntu 26.04 LTS    zone1

C:\Users\Jim\Documents\multipass\build>ping foo.mshome.net

Pinging foo.mshome.net [10.97.0.53] with 32 bytes of data:
Reply from 10.97.0.53: bytes=32 time=1ms TTL=64
Reply from 10.97.0.53: bytes=32 time=1ms TTL=64
Reply from 10.97.0.53: bytes=32 time=1ms TTL=64
Reply from 10.97.0.53: bytes=32 time=1ms TTL=64

Ping statistics for 10.97.0.53:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 1ms, Maximum = 1ms, Average = 1ms

C:\Users\Jim\Documents\multipass\build>ping bar.mshome.net

Pinging bar.mshome.net [10.97.1.223] with 32 bytes of data:
Reply from 10.97.1.223: bytes=32 time=1ms TTL=64
Reply from 10.97.1.223: bytes=32 time=1ms TTL=64
Reply from 10.97.1.223: bytes=32 time=1ms TTL=64
Reply from 10.97.1.223: bytes=32 time=1ms TTL=64

Ping statistics for 10.97.1.223:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 1ms, Maximum = 1ms, Average = 1ms
```

xmkg: @jimporter Weird... I don't think it should be possible by default since the vSwitches we create don't have the mshome.net suffix on this branch -- I'll test the build on my 4 Windows VMs to confirm.

sharder996 (changes requested):
#5088 is also present here. The QEMU virtual machine acts differently because it returns a cached state when `current_state()` is called whereas the HCS and AppleVZ virtual machine query and return a state from the hypervisor. This overwrites the previous state of `unavailable`. Maybe we can define a default behavior for `current_state()` in the Base virtual machine and fix this everywhere. LDYT

src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
xmkg: @jimporter I've tested the branch -- the launches were successful everywhere except my dev-w11-home VM. I recall changing some registry settings during the hostname resolution experiments, which may explain why it fails on dev VM. Logs are below for completeness.
[win11-hyperv-hcs-az-tests.txt](https://github.com/user-attachments/files/30317512/win11-hyperv-hcs-az-tests.txt)

xmkg:
> @jimporter #5097 ~needs to land before this, since the hostname resolution is broken in this branch.~ #5097 removes the reliance on hostname resolution altogether for IP discovery.

Further testing proved that the AZ hostname resolution is broken in my W11 dev VM, probably due to some registry flag changes I've made during the hostname resolution experiments. The attempts in W11 pro, W10 home and pristine W11 Home succeeded. #5097 is still a valid enhancement but not a blocker for this. I'll run CLI tests for further verification.

xmkg (dismissed):
LGTM, verified with CLI tests on Win11 Pro.

jimporter:
@sharder996 Ok, I've fixed the issue with the state reporting `Stopped` when disabling the instance's zone. Could you take a look? (This fix only applies to the Hyper-V API backend, since it's pretty simple and I didn't want to add a bunch of scaffolding to move more of this logic to the `BaseVirtualMachine`.)

sharder996 (approved):
I concur that this looks good.

@jimporter One thing that I did notice while testing is that the `hyperv_api` backend also suffers from subnet isolation. i.e. a VM in zone1 cannot reach a VM in zone2. However, since we have just been discussing scope creep in PRs, ~~I am going to merge this in and expect that it will get addressed in a subsequent PR.~~ created https://github.com/canonical/multipass/issues/5141 to track this

Thanks!

PR #4973, opened 2026-06-19
Feature: Version upgrade test harness

xmkg:
This PR introduces the initial revision of the version upgrade test harness.

The version upgrade suite checks that VMs and their state created on one Multipass version survive an upgrade to another. It lives at `tests/upgrade` (a sibling of `tests/cli`) and is its own entry point, run with `pytest tests/upgrade`. Because it is not under `tests/cli`, the standard `pytest tests/cli` run never collects it.

The set of tests is as follows:

 - **lifecycle:** sentinel + cpu/mem fingerprint survive; default(stopped), focal(suspended), debian(stopped, ≥1.17)
 - **suspend:** suspended VM returns Suspended, resumes, sentinel intact
 - **mount:** classic/native/suspended mount re-establishes; host+guest files intact
 - **mount mappings:** uid/gid maps persist in definition and stay applied (guest owner 0:0); classic + native(qemu)
 - **multi-mount:** classic+native on one VM both survive
 - **network:** extra-NIC MAC persists; --network stopped/suspended + --bridged
 - **snapshot:** BASE→{CHILD,FORK} tree; restore each, verify data isolation + cpu/mem/disk rollback
 - **clone:** clone after upgrade, sentinel intact in original + clone
 - **alias:** contexts + active-context + maps preserved, alias still runs
 - **delete/restore:** deleted (unpurged) VM recovers with data
 - **cloud-init:** marker persists, instance-id unchanged (didn't re-run)
 - **settings:** client.primary-name persists
 
 Also, made the following improvements to the CLI test framework:
 
  - **launch.py:** extra_args for custom launch flags (--cloud-init, --network, --bridged)
 - **conftest.py:** session-scoped daemon fixture for cross-test restarts, `caffeinate` for macOS runs
 - **helpers.py:** info() retries past post-launch sshd bounce
 - **multipass_cmd.py**: log signal number when client killed (no exit code)

Also adds a workflow to run the upgrade tests (Linux only), runs for stable -> PR build by default.


xmkg:
macOS and Windows runs tested locally -- both are green.

PR #4977, opened 2026-06-22
[clang-tidy] Add use after move check

tobe2098:
# Description

Use-after-move is bug-prone. Since C++ does not enforce/protect against this, using a static analyzer is our best bet.  

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,8 @@
 # Turn all warnings off but readability-identifier-naming
 Checks: >
   -*,
-  readability-identifier-naming
+  readability-identifier-naming,
+  bugprone-use-after-move
 CheckOptions:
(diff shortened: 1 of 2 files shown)


ricab (commented):
Nice, thanks @tobe2098!

I am OK with this, but I think we should try building before merging. I believe the CI filters need updating.

tobe2098:
Clang-tidy runs are currently disabled, so this change has no effect until re-enabled. The commit that disabled the runs states that there are some rough edges that need to be ironed out before re-enabling. WDYT @ricab?

ricab (approved):
OK, that's a separate thing IMO. Thanks for adding the file to filter. I have checked that this check was there as of version 19 (which we're using).

PR #4982, opened 2026-06-23
[cmake] Copy dart ffi binary to app bundle dir

sharder996:
This PR add a block in CMake to copy the C++/Dart ffi library into the GUI bundle directory. This mirrors what is being done on Linux:

https://github.com/canonical/multipass/blob/11d7f79ea0fc26b4e36ff73a6f5ab7184734af78/src/client/gui/CMakeLists.txt.linux#L59-L63

and allows one to execute `flutter build macos` from the `src/client/gui` directory and run the GUI without packaging the entire Multipass app.


xmkg:
@sharder996 the pipeline failed due to a codesign error

sharder996:
Decided to go with another approach.

PR #4989, opened 2026-06-25
[ci] Specify macOS version in all CI builds

jimporter:
# Description

This improves the reliability of our CI during periods (like now) when Github upgrades its `macos-latest` runner to a new version. The problem we had here is that we request XCode 16.2, which doesn't exist on macOS 26, and it's not easy to tell in the actions file the *real* macOS version, only the `runs-on` identifier we use.

## Related Issue(s)

See https://github.com/actions/runner-images/issues/14167.

## Checklist

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -39,11 +39,11 @@ jobs:
           const matrix = { include: [] };
 
           matrix.include.push({
-            "name": "macOS Latest (Apple Silicon)",
-            "runs-on": "macos-latest"
+            "name": "macOS 26 (Apple Silicon)",
+            "runs-on": "macos-26",
           },{
-            "name": "macOS 15 Sequoia (Intel Silicon)",
-            "runs-on": "macos-15-intel",
+            "name": "macOS 26 Sequoia (Intel Silicon)",
+            "runs-on": "macos-26-intel",
           })
 
           core.setOutput('matrix', JSON.stringify(matrix));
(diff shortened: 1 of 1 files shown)


jimporter:
@sharder996 I'd also be ok with upgrading our CI to use macOS 26, but I figured I'd just maintain the behavior we had prior to the start of GH's migration to macOS 26 as of June 15. One change at a time. :)

sharder996:
@jimporter How about https://github.com/canonical/multipass/pull/4991 instead?

jimporter:
@sharder996 If that PR fixes the build, then I think it's a good change to merge. Getting rid of unnecessary special-cases and overrides is always an improvement.

However, I also think it would be better for us to specify the expected version of our runners in general. Since our CI produces binaries that we distribute to users, it's safer and more-reproducible if we know for sure which OS/toolchain version was used to generate those binaries. Building on macOS 26 means using Xcode 26, and I wouldn't feel confident merely assuming that its builds have all the same compatibility guarantees as on Xcode 16.

sharder996:
I can see your argument for explicitly pinning the version, but in general I would like to see us moving forward in versioning and not falling back. IMO, having intermittent build failures due to upgrading environment/dependencies is good as it helps us identify things that need to be fixed or assumptions to resolve.

If we were to pin a macOS version, I would advocate for moving to the `macos-26` runner instead (also, there is a new `macos-26-intel` runner as well 🎉 ).

As far as compatibility goes, using macOS 15 or 26 shouldn't make a difference since we are still supporting back to macOS 13.3 and are setting the cmake build target.

If you have other thoughts, maybe this would be a good topic to bring up with the team? I'm willing to go with whatever consensus is made.

jimporter:
@sharder996 I'm fine with moving forward to 26 now as well. My main problem with using `-latest` in our CI is that during these migration periods, Github randomly chooses which OS to launch for each job.

sharder996:
I am in agreement. Let's try using the `macos-26-intel` runner straightaways then as well, no?

sharder996 (approved):

PR #4991, opened 2026-06-25
[ci] Remove specific version of xcode

sharder996:
Remove specific version of xcode as the underlying issue has seen been fixed.

https://github.com/protocolbuffers/protobuf/issues/21447


jimporter (approved):
Aside from the unrelated CI failures, I think this is a good change to make. I do still think we want #4989 for better reproducibility of our binaries though.

PR #5004, opened 2026-06-29
[ssh-process] Fix callback lifetime and remove error storage

### LINKED ISSUE #4990: [ssh] Bad exit status retrieval
**Describe the bug**
PlainSSHProcess attempts to retrieve the exit status by registering a callback and polling. The issue is that the callback is only registered for the duration of the call, so it will miss the return code if it does not come in that time window. The callbacks should ideally be registered for the lifetime of the channel.

tobe2098:
# Description

- What does this PR do? This PR first fixes the callback lifetime in PlainSSHProcess, and removes the previous convention of storing an exception pointer and re-throwing. Instead, we store the actual status of the channel and re-throw an error on every attempt.

- Why is this change needed? The previous implementation only registered callbacks during read_exit_code, which could cause a loss of the exit status of the `ssh_channel`.

## Related Issue(s)

Closes #4990

## Testing

- Unit tests
- Test classic mounts manually

## Checklist


src/ssh/plain_ssh_process.cpp
```diff
@@ -122,78 +187,46 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
 
     auto deadline = std::chrono::steady_clock::now() + timeout;
 
     int rc{SSH_OK};
     while ((std::chrono::steady_clock::now() < deadline) && rc == SSH_OK &&
-           !std::holds_alternative<int>(exit_result))
+           !exit_result.has_value() && !channel_closed && !channel_eof)
     {
         rc = ssh_event_dopoll(event.get(), timeout.count());
     }
```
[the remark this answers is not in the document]
tobe2098: `SSH_AGAIN` is returned after the full specified timeout has been completed. At that point we should consider a full timeout and stop polling. More correct implementation in #5016

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+mp::PlainSSHProcess::EventUPtr mp::PlainSSHProcess::get_event_in_session()
 {
-    if (auto eptrptr = std::get_if<std::exception_ptr>(&exit_result))
+
+    mp::PlainSSHProcess::EventUPtr event{MP_LIBSSH.ssh_event_new(),
+                                         [](ssh_event e) { MP_LIBSSH.ssh_event_free(e); }};
+
+    std::optional<std::string> err = std::nullopt;
+    if (!event)
+        err = "could not allocate event";
```
ricab: Libssh appears to use "event" as a shorthand for "event context", which makes much more sense conceptually and matches the documentation for many of the corresponding functions. For example, [docs for `ssh_event_new`](https://api.libssh.org/stable/group__libssh__poll.html#ga406895f99b6b486c4b65536e5399ff96) and below.

Since this would likely end up being communicated to the user (and in any case in the logs), I suggest expanding the message slightly:


```suggestion
        err = "could not allocate a libssh event context";
```

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+    else if (MP_LIBSSH.ssh_event_add_session(event.get(), session) != SSH_OK)
+    {
+        const auto raw_err = ssh_get_error(session);
+        err = fmt::format("could not add event to session: {}",
```
ricab: Similar:


```suggestion
        err = fmt::format("could not add a libssh event context to the SSH session: {}",
```

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+                          raw_err && *raw_err ? raw_err : "Empty error");
```
ricab: How about

```suggestion
                          raw_err && *raw_err ? raw_err : "no detail");
```

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+    }
+    if (err.has_value())
```
ricab, line 300: Suggesting a new line to visually separate from the previous if-else.

```suggestion

    if (err.has_value())
```

include/multipass/ssh/plain_ssh_process.h
```diff
@@ -63,17 +65,37 @@ class PlainSSHProcess : public SSHProcess
...
 
     std::unique_lock<std::mutex> session_lock; // do not attempt to re-lock, as this is moved from
     ssh_session session;
     std::string cmd;
+
+    ssh_channel_callbacks_struct cb;
     ChannelUPtr channel;
-    std::variant<std::monostate, int, std::exception_ptr> exit_result;
+
+    ExitResultType exit_result{};
```
ricab: I find `optional<int>` would be more informative here, especially given that the var is already named `exit_result`. But I'll leave it up to you.

src/ssh/plain_ssh_process.cpp
```diff
@@ -94,6 +68,37 @@ auto make_channel(ssh_session session, const std::string& cmd)
...
+    if (signal == "SEGV")
+        return base_signal_code + 11;
+    if (signal == "PIPE")
+        return base_signal_code + 13;
+    if (signal == "ALRM")
+        return base_signal_code + 14;
+    if (signal == "TERM")
+        return base_signal_code + 15;
+    return base_signal_code;
+}
```
ricab: After investigating this, I understand why you had to do it:

- The signal is sent from a Linux guest, but we need to interpret it on macOS and Windows too, where we can't just include Linux stuff.
- This subset of POSIX signals is part of the SSH protocol itself (https://www.rfc-editor.org/info/rfc4254/#section-6.9).

I also understand why you didn't go beyond to cover any [*Additional 'signal name' values [that] MAY be sent*](https://www.rfc-editor.org/info/rfc4254/#section-6.10): openssh [doesn't implement any](https://github.com/openssh/openssh-portable/blob/7e446d3f5917c2f2770981a89d0e54d5d064bf0c/session.c#L2264-L2283).

All of these seem to me like good choices. However, they are not immediately obvious, so could they be briefly covered in comments or even doc-comments in the code, please? You can also use commit messages to explain decisions in more detail for the future. And I do think this deserves a separate commit.

A few other points that I would like to clarify or request follow-up on:

- Why did you skip SIGUSR1 and SIGUSR2?
- Could this be tested please?
- Could the mapping be separated from the logic?
- I am also hardcoding a bunch of protocol constants for sftp in #5030: https://github.com/canonical/multipass/pull/5030/changes#diff-970596b902d9cd58bef93705bc4ddd39a3ee5964799a96b392c03b7c8239aef4. Do you think it would make sense to have a separate header for the protocol constants, mapping, and utility? Your case is much more specific, so the name would probably not be analogous to `sftp_protocol.h`, but perhaps `ssh_signals.h` or something? It might also make testing simpler. Up to you though.

src/ssh/plain_ssh_process.cpp
```diff
@@ -103,21 +108,81 @@ mp::PlainSSHProcess::PlainSSHProcess(ssh_session_struct& session,
...
+void mp::PlainSSHProcess::channel_exit_signal_cb(ssh_session,
+                                                 ssh_channel,
+                                                 const char* signal,
+                                                 int,
+                                                 const char*,
+                                                 const char*,
+                                                 void* userdata)
+{
+    auto* process = static_cast<mp::PlainSSHProcess*>(userdata);
+    mp::top_catch_all(category, [process, signal] {
```
ricab: Only the log could throw in here, so could the top catch be restricted to that operation, please? 

BTW, the  `top_catch_all` would probably fail and terminate when trying to log again. That's probably what we want when we can't even log. Although I get some goosebumps thinking of ClientLogger... anyway, something for another day.

src/ssh/plain_ssh_process.cpp
```diff
@@ -103,21 +108,81 @@ mp::PlainSSHProcess::PlainSSHProcess(ssh_session_struct& session,
...
+        mpl::error(category,
```
ricab: Consider the [docs for the logging levels](https://github.com/canonical/multipass/blob/eb05c4cc4fb9f583e863b05fb859564c31d62c29/include/multipass/logging/level.h#L32). I doubt that `error` is the right choice in here. There is no way for this code to know that an exit signal represents an error. For example, someone could launch a process that waits for something else to signal it of some condition. Exiting on a signal could be the success path.

Interpreting the exit result is a responsibility for the client of the process, not the process object itself. That's why we normally throw and let upper levels decide what to do about it. We'd best not throw in here, but we can record all the info that callers will need to decide how to handle: namely, the signal. Anyway, I would say that this should not go above warning.

src/ssh/plain_ssh_process.cpp
```diff
@@ -128,79 +193,46 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
+        return *exit_result;
 
     auto local_lock = std::move(session_lock); // unlock at the end
-    read_exit_code(timeout, /* save_exception = */ true);
+    read_exit_code(timeout);
 
-    assert(std::holds_alternative<int>(exit_result));
-    return std::get<int>(exit_result);
+    assert(exit_result.has_value()); // TODO: remove assert or exit_code must be called only if
+    // exit_recognized returned true
```
ricab: Can you explain the TODO and why you left it like this, please?

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+        err = fmt::format("could not add event to session: {}",
+                          raw_err && *raw_err ? raw_err : "Empty error");
+    }
+    if (err.has_value())
     {
-        assert(!session_lock.owns_lock());
-        std::rethrow_exception(*eptrptr);
+        throw SSHProcessExitError{cmd, err.value()};
     }
+    return event;
```
ricab, line 305:
```suggestion

    return event;
```

src/ssh/plain_ssh_process.cpp
```diff
@@ -267,14 +299,29 @@ ssh_channel mp::PlainSSHProcess::release_channel()
...
+        MP_LIBSSH.ssh_remove_channel_callbacks(channel.get(), &cb);
     return channel.release();
 }
 
-void multipass::PlainSSHProcess::rethrow_if_saved() const
+mp::PlainSSHProcess::EventUPtr mp::PlainSSHProcess::get_event_in_session()
 {
-    if (auto eptrptr = std::get_if<std::exception_ptr>(&exit_result))
+
+    mp::PlainSSHProcess::EventUPtr event{MP_LIBSSH.ssh_event_new(),
```
ricab, line 287: Why do you add and remove the event in every occasion? Couldn't this be a field in the process (or even the session)? Is it because we still want to use blocking calls like `ssh_channel_read_timeout`? But it looks like `ssh_handle_packets`
gets the poll context from the socket, which would resolve anyway, no?

src/ssh/plain_ssh_process.cpp
```diff
@@ -128,79 +193,46 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
     }
 
-    if (!std::holds_alternative<int>(exit_result))
+    if (!exit_result.has_value())
     {
-        if (rc == SSH_ERROR)
-        { // we expect SSH_AGAIN or SSH_OK (unchanged) when there is a timeout
+        if (channel_closed) // eof is ok
+            throw SSHProcessExitError{cmd, "channel is closed with no exit status"};
+        else if (rc == SSH_ERROR)
```
ricab: Nit: redundant else.

src/ssh/plain_ssh_process.cpp
```diff
@@ -128,79 +193,46 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
-        else
-            eptr = std::make_exception_ptr(SSHProcessTimeoutException{cmd, timeout});
-        // note that make_exception_ptr takes by value; we repeat the call with the concrete
-        // types to avoid slicing
-
-        if (save_exception)
-            exit_result = eptr;
-
-        std::rethrow_exception(eptr);
+        else // we expect SSH_AGAIN or SSH_OK (unchanged) when there is a timeout
```
ricab: Same: redundant else.

src/ssh/plain_ssh_process.cpp
```diff
@@ -128,79 +193,46 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
-    if (!std::holds_alternative<int>(exit_result))
+    if (!exit_result.has_value())
     {
-        if (rc == SSH_ERROR)
-        { // we expect SSH_AGAIN or SSH_OK (unchanged) when there is a timeout
+        if (channel_closed) // eof is ok
+            throw SSHProcessExitError{cmd, "channel is closed with no exit status"};
+        else if (rc == SSH_ERROR)
+        { // SSH_ERROR with no closed or eof means that poll returned <0. Further polling will fail
+          // (unless the error was EINTR). Not setting the closed or eof.
```
ricab, line 205: I am wondering if the the condition in the while above should cover this EINTR case. Reading the implementation, libssh already covers EINTR in an internal while (src/poll.c:728), so the only way we could get it is if it occurred exactly when getting the time, I think. That's very unlikely, but I guess possible. WDYT?

src/ssh/plain_ssh_process.cpp
```diff
@@ -103,21 +108,81 @@ mp::PlainSSHProcess::PlainSSHProcess(ssh_session_struct& session,
...
+{
+    auto* process = static_cast<mp::PlainSSHProcess*>(userdata);
+    mp::top_catch_all(category, [process, signal] {
+        mpl::error(category,
+                   "{}: Process terminated by remote signal: SIG{}",
+                   process->cmd,
+                   signal ? signal : "UNKNOWN");
+
+        auto sig_code{signal_to_exit_code(signal)};
+        process->exit_result = sig_code;
```
ricab: I am wondering if we should really collapse this with the exit status like shells do. It's a *lossy* approach, since exit statuses above 128 are possible. 

[POSIX establishes](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) the >128 value for shells (section 2.8.2 Exit Status for Commands). But we're not a shell. Certainly not at this layer. And in `multipass exec` we do treat the signal separately (not using this code).

The alternative would be to save it and provide it separately. Either back to a variant return or a throw... perhaps with a 3rd ExitlessSSHProcessException (which would then need to be renamed). We could then confidently drop the log above and let the caller decide what to do. But I guess we'd still need to revisit all call sites to see if they care about the distinction in any other way. Well, that should arguably be done in any case...

All that said, I am open to being convinced that this collapsing is good enough a checkpoint to settle for in this PR. I do realize that even our (local) ProcessFactory is remiss in this regard: it distinguishes crashes from normal exits, but it doesn't provide any signals. Just an `optional<int>`, so there is precedent. Looking a little deeper, it's a limitation that comes from [QProcess](https://doc.qt.io/qt-6/qprocess.html), which also discards the signal AFAICS. But there it is probably motivated by cross-platform generality, whereas we know our SSH processes execute on Linux and the signal is in any case established by the SSH protocol...

Please let me know how much of an effort you think that would take and whether it makes sense to you here, in another PR (e.g. to reduce changes going into main right now), as a indefinite TODO, or not at all.

ricab (changes requested):
Very nice Antoni, well spotted and fixed! 👏 

This one is dense with pulpy juice, which reflects in my own comment density in some parts. There is just a lot to it and I had to dig into a bunch of separate documentation to make an independent judgement, ranging from Qt to POSIX requirements on shells. With a whole bunch of SSH in between, of course 🙂 I had been mostly grounding myself on the assumption that the existing usage of libssh was correct and this was a good chance to dig a little further into it.

I have only skimmed the tests. As discussed, moving to MP_LIBSSH would make more sense to me. Other than that, I have a number of suggestions and requests, but also questions and various things that I would like to hear your opinion on.

src/ssh/plain_ssh_process.cpp
tobe2098: I forgot to add SIGUSR, it will be added. 
I will add tests for this in the new tests file. 
Yes. 
I think we can move this to a separate header, it does make sense. I would maybe set it as `ssh_utils.h/cpp` so it encompasses more functionality? WDYT?

src/ssh/plain_ssh_process.cpp
tobe2098: Hmm... that is an issue. Throwing inside the C-based callbacks does not work well, since C does not have the stack unwinding mechanisms for it within libssh code.

src/ssh/plain_ssh_process.cpp
tobe2098: The issue is closer to the multiple `ssh_event` issue and ownership of the underlying fds. In https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/poll.c#L999 You can see that in the while loop later in the function, the polling context is transferred out of the session and into the event. This will cause issues even in the current state, since I missed that the event should be removed from the session. I can say the following:

1. Ideally, the architecture we discusses (SSHSession owns the event and the polling responsibility) would be the ideal. This way there would be no potential conflicts with more than one event being registered.
2. In this PR, I can unregister the event when it is not used anymore, leaving it could mean make things hard for the transfer of responsibilities to the Sftp side. Transferring session internals under ownership of SSHProcess (permanently) also does not seem ideal.
3. I can make the event object a member, that is not an issue, it just needs to be registered and de-registered on ssh_poll calls.

WDYT?

src/ssh/plain_ssh_process.cpp
tobe2098: I missed that it is already implemented. Yes, I believe in this case the EINTR check is redundant here

src/ssh/plain_ssh_process.cpp
tobe2098: I understand the concern with shell, but this exit code _is_, in some way, a shell exit code. The `ssh_channel_request_exec` runs the command in a shell (https://unix.stackexchange.com/questions/414636/how-do-ssh-remote-command-line-arguments-get-parsed, I could not access OpenSSH due to network issues). That said, we could separate the signal storage.

The only drawback of this storage is that when checking exit status, we would have to check both variables and provide what is expected. Up until now, the interface was only assuming that we receive an exit code (which the libssh blocking implementation collapses into the same variable). We could imitate this by having the collapsed code and the signal name stored, or by having them separately and converting the signal to the exit code on request of an exit code. WDYT would be best? My decision to go with collapsed exit code was that we do not have a use for the signal name beyond logging, and the exit code has the same information (if >=128 the process was signal-terminated).  

PS: This callback will be triggered when the process exits due to a signal (not just receive a signal), so no ExitlessSSHProcess.

src/ssh/plain_ssh_process.cpp
tobe2098: Ignore the comment on issues in the current PR relating to de-registering the event. ssh_event_free seems to already do the de-registration properly: https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/poll.c#L1219

src/ssh/plain_ssh_process.cpp
tobe2098: Following the codepath the assert is redundant, so I will remove

src/ssh/plain_ssh_process.cpp
```diff
@@ -80,6 +48,13 @@ auto make_channel(ssh_session session, const std::string& cmd)
     mp::PlainSSHProcess::ChannelUPtr channel{
         MP_LIBSSH.ssh_channel_new(session),
         [](ssh_channel ch) { MP_LIBSSH.ssh_channel_free(ch); }};
+    // Add callbacks before traffic is open
+    mp::SSH::throw_on_error(
+        channel,
+        session,
+        fmt::format("[{}] failed to add channel callbacks", category).c_str(),
+        std::bind_front(&mp::Libssh::ssh_add_channel_callbacks, &mp::Libssh::instance()),
+        cb);
```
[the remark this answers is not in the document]
tobe2098: THis is not correct, since the throw_on_error function will already format this pointer into the string contained by the exception, copying the underlying string.

src/ssh/plain_ssh_process.cpp
```diff
@@ -80,20 +48,26 @@ auto make_channel(ssh_session session, const std::string& cmd)
...
-        "[ssh proc] failed to open session channel",
+        fmt::format("[{}] failed to open session channel", category).c_str(),
         std::bind_front(&mp::Libssh::ssh_channel_open_session, &mp::Libssh::instance()));
     mp::SSH::throw_on_error(
         channel,
         session,
-        "[ssh proc] exec request failed",
+        fmt::format("[{}] exec request failed", category).c_str(),
         std::bind_front(&mp::Libssh::ssh_channel_request_exec, &mp::Libssh::instance()),
         cmd.c_str());
```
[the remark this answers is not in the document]
tobe2098: The callbacks will be removed because they are associated to the channel and the channel is freed.

src/ssh/plain_ssh_process.cpp
```diff
@@ -128,79 +170,45 @@ bool mp::PlainSSHProcess::exit_recognized(std::chrono::milliseconds timeout)
...
+        return exit_result.value();
 
-    auto local_lock = std::move(session_lock); // unlock at the end
-    read_exit_code(timeout, /* save_exception = */ true);
+    read_exit_code(timeout);
 
-    assert(std::holds_alternative<int>(exit_result));
-    return std::get<int>(exit_result);
+    return exit_result.value();
 }
```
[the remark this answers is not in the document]
tobe2098: But the sshfs_process is always immediately `release_channel`'d, which releases the lock. Releasing the lock only on exit_code calls is very inconsistent behavior.

tests/unit/mock_ssh_callback_engine.h
```diff
@@ -0,0 +1,155 @@
...
+namespace multipass::test
+{
+struct CallbackChState
+{
+    int ssh_rc{SSH_OK};
+    bool eof{true};
+    bool closed{true};
+    std::optional<int> exit_code{0};
+    std::optional<std::string> signal;
+    std::chrono::milliseconds wait{std::chrono::milliseconds(0)};
```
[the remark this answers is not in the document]
tobe2098: @copilot, but it is not failing to compile. What would be the reason?

include/multipass/ssh/ssh_signal.h
```diff
@@ -0,0 +1,89 @@
...
+namespace SSH
+{
+
+struct SignalMapping
+{
+    std::string_view name;
+    int offset;
+};
+
+// The signal numbers vary in some architectures, but not in any we support
```
ricab, line 33:
```suggestion
// The signal numbers vary in some architectures, but not in any way we support
```

tests/unit/test_ssh_process.cpp
```diff
@@ -159,3 +157,32 @@ TEST_F(SSHProcess, getCmdReturnsCommandName)
...
+    callback_mock_engine.pop_state();
+
+    auto proc = session.exec("something");
+    EXPECT_THAT(proc->exit_code(), Eq(signal_code));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    SSHProcess,
+    SSHProcessSignals,
+    ::testing::Values("HUP", "INT", "QUIT", "USR1", "SEGV", "USR2", "TERM", "UNKNOWN"));
```
ricab: How about adding nullptr and "" to exercise ssh_utils.h:75?

include/multipass/ssh/ssh_signal.h
ricab, line 1: "Utils" seems very generic. I'd recommend ssh_signal.h or something like that. Up to you though.

include/multipass/ssh/ssh_signal.h
```diff
@@ -0,0 +1,89 @@
...
+    SignalMapping{"ABRT", 6},
+    SignalMapping{"FPE", 8},
+    SignalMapping{"KILL", 9},
+    SignalMapping{"USR1", 10},
+    SignalMapping{"SEGV", 11},
+    SignalMapping{"USR2", 12},
+    SignalMapping{"PIPE", 13},
+    SignalMapping{"ALRM", 14},
+    SignalMapping{"TERM", 15},
+};
```
ricab, line 48: Nice 😎

src/client/gui/pubspec.lock
ricab, line 1: Is this intentional?

src/ssh/plain_ssh_process.cpp
ricab: Looking at it again, it probably doesn't make sense to mix those two (your new header and the sftp_protocol.h). Dedicated header makes sense to me.

src/ssh/plain_ssh_process.cpp
ricab: But `top_catch_all` is `noexcept`, so nothing would escape. The program would terminate. We should probably make the logging functions best-effort and `noexcept`, but that's a separate matter.

src/ssh/plain_ssh_process.cpp
```diff
@@ -151,15 +120,23 @@ void mp::PlainSSHProcess::channel_exit_signal_cb(ssh_session,
                                                  void* userdata)
```
ricab: Should these be made no except then?

tests/unit/c_mock_defines.cmake
```diff
@@ -51,29 +51,12 @@ add_c_mocks(
...
-  sftp_reply_names_add
-  sftp_reply_handle
-  sftp_reply_version
-  sftp_get_client_message
-  sftp_client_message_free
-  sftp_client_message_get_data
-  sftp_client_message_get_filename
-  sftp_handle
-  sftp_handle_alloc
-  sftp_handle_remove
```
ricab, line 70: 🤤

tests/unit/mock_ssh_callback_engine.h
```diff
@@ -17,36 +17,44 @@
...
+#include <chrono>
 #include <optional>
 #include <queue>
+#include <string>
+#include <thread>
 
 namespace multipass::test
 {
-struct CallbackState
+struct CallbackChState
```
ricab: A couple of questions on this:

- Why did you rename? Avoid ambiguity?
- What is ch? Change? Check? 

Not saying you need to change it necessarily, brevity has its own value. But maybe there is another abbreviation that is more obvious. Or perhaps a comment...

src/ssh/plain_ssh_process.cpp
ricab: OK, I'll take it. We can distinguish if/when we need to.

src/ssh/plain_ssh_process.cpp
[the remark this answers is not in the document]
ricab: I still see the lock move in the latest diff. Should we resolve this?

ricab (commented):
Looking good 🙂 

On the coverage side, there are still a few untested changes in the process. Will perhaps #5172 take care of that?

Other than that, this is converging. I have mostly comment resolutions and small proposals. Also, could you please resolve any copilot comments that you don't think hold any water? 

Thanks a lot for your work on making this better!

tests/unit/mock_ssh_callback_engine.h
tobe2098: I can rename to CallbackChannelState. The idea is that eventually there will be non-channel callbacks, and this callback state does not represent said callbacks.

src/ssh/plain_ssh_process.cpp
tobe2098: Definitely, good catch

src/client/gui/pubspec.lock
tobe2098: No, that must have slipped through

include/multipass/ssh/ssh_signal.h
tobe2098: I can move it back to signal

include/multipass/ssh/ssh_signal.h
tobe2098: It would be `not in any architecture we support`

tests/unit/test_plain_ssh_session.cpp
```diff
@@ -110,13 +111,21 @@ TEST_F(TestPlainSSHSession, execThrowsWhenUnableToRequestChannelExec)
...
+        return SSH_AUTH_SUCCESS;
+    });
     mp::PlainSSHSession session = make_ssh_session();
 
-    REPLACE(ssh_is_connected, [](auto...) { return true; });
-    REPLACE(ssh_channel_open_session, [](auto...) { return SSH_OK; });
-    REPLACE(ssh_channel_request_exec, [](auto...) { return SSH_OK; });
+    EXPECT_CALL(*mock_libssh, ssh_is_connected).WillOnce([](auto...) { return true; });
+    EXPECT_CALL(*mock_libssh, ssh_channel_open_session).WillOnce([](auto...) { return SSH_OK; });
+    EXPECT_CALL(*mock_libssh, ssh_channel_request_exec).WillOnce([](auto...) { return SSH_OK; });
```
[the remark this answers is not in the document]
tobe2098: Correct, but that is to be addressed in another PR

tests/unit/test_ssh_process.cpp
```diff
@@ -34,35 +35,42 @@ namespace
...
+        ON_CALL(mock_libssh, ssh_event_add_session).WillByDefault([](auto...) { return SSH_OK; });
+    }
     const mpt::StubSSHKeyProvider key_provider;
     mpt::MockSSHTestFixture mock_ssh_test_fixture;
     mp::PlainSSHSession session{"theanswertoeverything", 42, "ubuntu", key_provider};
+    mpt::MockLibssh::GuardedMock libssh_guard{mpt::MockLibssh::inject()};
+    mpt::MockLibssh& mock_libssh = *libssh_guard.first;
+    mpt::CallbackEngineMock callback_mock_engine{mock_libssh,
+                                                 mpt::CallbackEngineMock::channel_exit_success};
 };
```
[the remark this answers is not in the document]
tobe2098: This is to be fixed in the sshprocess singleton tests PR.

tests/unit/mock_ssh_callback_engine.h
ricab: Ah I see. I didn't pick up on that from the "ch". I would be fine with `CallbackChannelState` or something generic if you prefer.

tobe2098: Done

PR #5005, opened 2026-06-29
[macOS] Instance hostname resolution

sharder996:
This PR implements a DNS resolver that allows the system's mDNSResponder to forward `.multipass` DNS requests to a thread being run by the Multipass daemon.

Requests come through a UDP socket and the requested hostname has its IP address queried by the Multipass daemon.

This is an MVP solution that provides most of the functionality of instance hostname resolution, but lacks a few potentially useful features such as IPv6 support and reverse lookups. If this direction was to be closen, some further development would be done along with proper unit testing.

### DIFF
--- a/src/platform/backends/shared/macos/dns_server.cpp
+++ b/src/platform/backends/shared/macos/dns_server.cpp
@@ -20,7 +20,9 @@
 #include <multipass/format.h>
 #include <multipass/logging/log.h>
 
+#include <fcntl.h>
 #include <netinet/in.h>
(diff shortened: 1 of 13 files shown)


src/platform/backends/shared/macos/dns_server.cpp
```diff
@@ -0,0 +1,323 @@
+/*
```
xmkg, line 1: The Windows side will need a DNS server too. How hard would it be to make it available for Windows as well?

PR #5017, opened 2026-07-03
[dnsmasq] Get_ip_for: Skip IPs from unserved range

xmkg:
The previous implementation returned the first matching entry for the given MAC address, which could be from an old bridge that's been served before. This patch ensures that get_ip_for skips the lease entries that are not in the served subnets.

### DIFF
--- a/src/platform/backends/qemu/linux/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/linux/dnsmasq_server.cpp
@@ -36,6 +36,7 @@ namespace mpl = multipass::logging;
 namespace
 {
 constexpr auto immediate_wait = 100; // period to wait for immediate dnsmasq failures, in ms
+constexpr auto log_category = "dnsmasq";
 
 auto make_dnsmasq_process(const mp::Path& data_dir,
                           const mp::BridgeSubnetList& subnets,
@@ -85,17 +86,17 @@ mp::DNSMasqServer::~DNSMasqServer()
     {
         QObject::disconnect(finish_connection);
 
-        mpl::debug("dnsmasq", "terminating");
+        mpl::debug(log_category, "terminating");
         dnsmasq_cmd->terminate();
 
         if (!dnsmasq_cmd->wait_for_finished(terminate_timeout))
         {
-            mpl::info("dnsmasq", "failed to terminate nicely, killing");
+            mpl::info(log_category, "failed to terminate nicely, killing");
             dnsmasq_cmd->kill();
 
             if (!dnsmasq_cmd->wait_for_finished(kill_timeout))
             {
-                mpl::warn("dnsmasq", "failed to kill (timed out)");
+                mpl::warn(log_category, "failed to kill (timed out)");
             }
         }
     }
@@ -124,6 +125,8 @@ std::optional<mp::IPAddress> mp::DNSMasqServer::get_ip_for(const std::string& hw
                 // address; skip it so we resolve the current, reachable one instead.
                 if (ip_in_served_subnets(subnets, ip))
                     return ip;
+                // TODO: Also check if the lease is still valid
+                // TODO: Aggregate all leases for the MAC before filtering
             }
             catch (const std::exception&) // unparseable address -> skip this line
             {
@@ -138,7 +141,7 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr, const QString& b
     auto ip = get_ip_for(hw_addr);
     if (!ip)
     {
-        mpl::warn("dnsmasq", "attempting to release non-existent addr: {}", hw_addr);
+        mpl::warn(log_category, "attempting to release non-existent addr: {}", hw_addr);
         return;
     }
 
@@ -146,7 +149,7 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr, const QString& b
     QObject::connect(&dhcp_release,
                      &QProcess::errorOccurred,
                      [&ip, &hw_addr](QProcess::ProcessError error) {
-                         mpl::warn("dnsmasq",
+                         mpl::warn(log_category,
                                    "failed to release ip addr {} with mac {}: {}",
                                    ip.value().as_string(),
                                    hw_addr,
@@ -157,7 +160,7 @@ void mp::DNSMasqServer::release_mac(const std::string& hw_addr, const QString& b
         if (exit_code == 0 && exit_status == QProcess::NormalExit)
             return;
 
-        mpl::warn("dnsmasq",
+        mpl::warn(log_category,
                   "failed to release ip addr {} with mac {}, exit_code: {}",
                   ip.value().as_string(),
                   hw_addr,
@@ -180,7 +183,7 @@ void mp::DNSMasqServer::check_dnsmasq_running()
 {
     if (!dnsmasq_cmd->running())
     {
-        mpl::warn("dnsmasq", "Not running");
+        mpl::warn(log_category, "Not running");
         start_dnsmasq();
     }
 }
@@ -207,7 +210,7 @@ std::string dnsmasq_failure_msg(const mp::ProcessState& state)
 
 void mp::DNSMasqServer::start_dnsmasq()
 {
-    mpl::debug("dnsmasq", "Starting dnsmasq");
+    mpl::debug(log_category, "Starting dnsmasq");
 
     finish_connection =
         QObject::connect(dnsmasq_cmd.get(), &mp::Process::finished, [](const ProcessState& state) {
--- a/src/platform/backends/qemu/linux/dnsmasq_server.cpp
+++ b/src/platform/backends/qemu/linux/dnsmasq_server.cpp
@@ -26,6 +26,7 @@
 
 #include <QDir>
 
+#include <algorithm>
 #include <chrono>
 #include <fstream>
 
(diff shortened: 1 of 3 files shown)


xmkg:
@ricab assigned you for this one too -- this is a follow-up for #4971. I've made another PR for clarity.

xmkg:
Tagging @jimporter too since this is related to AZ changes. Same for the base branch (qemu tap device).

xmkg:
Just verified that base + this branch fixes the post-upgrade network connectivity issues.

tests/unit/qemu/linux/test_dnsmasq_server.cpp
```diff
@@ -141,6 +141,108 @@ TEST_F(DNSMasqServer, returnsNullIpWhenLeasesFileDoesNotExist)
     EXPECT_FALSE(ip);
 }
 
+TEST_F(DNSMasqServer, getIpForSkipsStaleLeaseAndReturnsLiveOne)
+{
+    // Regression (MULTI-2701 follow-up): a version upgrade can change the served subnet, leaving a
```
jimporter, line 146: Just a quick comment here before I look at it in more detail: the goal is that we *don't* change the subnet upon upgrade from 1.16 to 1.17. That way, we avoid breaking things in case users are relying on this elsewhere.

Are you seeing the subnet change during upgrades? (This might still happen on Windows, but on macOS and Linux, the current code should use the same subnet for zone1 as we used in Multipass 1.16.)

xmkg: Yes, the subnet changes after the upgrade; at least that's the case in Linux. Here's a DHCP DORA from 1.16 suspended, 1.17 resumed VM:

```
2026-07-06T23:06:13+03:00 dnsmasq-dhcp[172776]: DHCPREQUEST(mpqemubrzone1) 10.195.180.46 52:54:00:a0:3c:cd
2026-07-06T23:06:13+03:00 dnsmasq-dhcp[172776]: DHCPNAK(mpqemubrzone1) 10.195.180.46 52:54:00:a0:3c:cd wrong network
|2026-07-06T23:06:16+03:00 dnsmasq-dhcp[172776]: DHCPDISCOVER(mpqemubrzone1) 10.195.180.46 52:54:00:a0:3c:cd
2026-07-06T23:06:16+03:00 dnsmasq-dhcp[172776]: DHCPOFFER(mpqemubrzone1) 10.97.0.46 52:54:00:a0:3c:cd
2026-07-06T23:06:16+03:00 dnsmasq-dhcp[172776]: DHCPREQUEST(mpqemubrzone1) 10.97.0.46 52:54:00:a0:3c:cd
2026-07-06T23:06:16+03:00 dnsmasq-dhcp[172776]: DHCPACK(mpqemubrzone1) 10.97.0.46 52:54:00:a0:3c:cd upg-suspend

```

It had 10.195.180.46 before the upgrade. The VM requested its existing lease again; dnsmasq rejected since it no longer serves that subnet, then the client initiated DORA and got 10.97.0.46 as the new IP.

jimporter: Ok, that's a bug then. We'd decided that the goal was to preserve the IP/subnet for existing instances if at all possible to minimize problems. I'd be ok revisiting this decision if we'd rather not use the random subnets from 1.16 in 1.17+ though.

I can take a look at fixing this to preserve the old subnet during upgrade. I'm not sure if that renders this PR obsolete though...

jimporter: @xmkg I filed https://github.com/canonical/multipass/pull/5031 to fix the bug. It might be worth keeping this PR around though; this behavior is more robust and could be useful if we add support for changing subnets. Changing subnets would be nice for avoiding conflicts with existing networks...

xmkg: Great to hear! I think we should keep this one regardless, since any subnet change would result in a network outage with the old code. I think we should also log it as a warning since the change is most likely due to an unexpected/unintended change. I'll also add an IP address check to the upgrade tests.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
```diff
@@ -105,7 +115,20 @@ std::optional<mp::IPAddress> mp::DNSMasqServer::get_ip_for(const std::string& hw
...
+            try
+            {
+                const IPAddress ip{fields[ipv4_idx]};
+                // Ignore leases outside the subnets we currently serve. A version upgrade that
+                // changes the bridge layout can leave an old lease pointing at an unreachable
+                // address; skip it so we resolve the current, reachable one instead.
+                if (ip_in_served_subnets(subnets, ip))
+                    return ip;
+            }
+            catch (const std::exception&) // unparseable address -> skip this line
```
ricab: Entries that could not be parsed and resulted in an exception would have thrown before. This now swallows any exception that could come about. So two questions:

- Why choose not to let parsing exceptions through now? Was there a specific problematic case that this prevents?
- If we indeed want to skip exceptions that signal a wrong IP format, wouldn't a `catch(std::invalid_argument)` be a better sieve? I am thinking that other exceptions that occur now or in the future might signal problems other than IP parsing. 

https://github.com/canonical/multipass/blob/61ce96166a8764dfeb8f8d866be3fa12b0a3132c/src/network/ip_address.cpp#L50

src/platform/backends/qemu/linux/dnsmasq_server.cpp
```diff
@@ -105,7 +115,20 @@ std::optional<mp::IPAddress> mp::DNSMasqServer::get_ip_for(const std::string& hw
...
         if (fields.size() > 2 && fields[hw_addr_idx] == hw_addr)
-            return IPAddress{fields[ipv4_idx]};
+        {
+            try
+            {
+                const IPAddress ip{fields[ipv4_idx]};
+                // Ignore leases outside the subnets we currently serve. A version upgrade that
+                // changes the bridge layout can leave an old lease pointing at an unreachable
+                // address; skip it so we resolve the current, reachable one instead.
+                if (ip_in_served_subnets(subnets, ip))
```
ricab, line 126: The check is good, but now I am slightly worried that we could still pick a stale IP that happens to be in the right subnet. Could that come about? Could dnsmasq ignore an existing lease and assign a new IP to the VM?

src/platform/backends/qemu/linux/dnsmasq_server.cpp
```diff
@@ -43,10 +44,19 @@ auto make_dnsmasq_process(const mp::Path& data_dir,
     auto process_spec = std::make_unique<mp::DNSMasqProcessSpec>(data_dir, subnets, conf_file_path);
     return MP_PROCFACTORY.create_process(std::move(process_spec));
 }
+
+bool ip_in_served_subnets(const mp::BridgeSubnetList& subnets, const mp::IPAddress& ip)
+{
+    return std::ranges::any_of(subnets, [&ip](const auto& bridge_subnet) {
+        return bridge_subnet.second.contains(ip);
+    });
```
ricab, line 53: nice

tests/unit/qemu/linux/test_dnsmasq_server.cpp
```diff
@@ -141,6 +141,108 @@ TEST_F(DNSMasqServer, returnsNullIpWhenLeasesFileDoesNotExist)
     EXPECT_FALSE(ip);
 }
 
+TEST_F(DNSMasqServer, getIpForSkipsStaleLeaseAndReturnsLiveOne)
```
ricab: The test would not pass if the stale lease was placed in the same subnet IIUC.

```suggestion
TEST_F(DNSMasqServer, getIpForSkipsStaleSubnetLeaseAndReturnsLiveOne)
```

tests/unit/qemu/linux/test_dnsmasq_server.cpp
```diff
@@ -141,6 +141,108 @@ TEST_F(DNSMasqServer, returnsNullIpWhenLeasesFileDoesNotExist)
...
+    auto dns_default = make_default_dnsmasq_server(); // no longer serves 192.168.65.0/24
+    EXPECT_FALSE(dns_default.get_ip_for(hw_addr));
+}
+
+TEST_F(DNSMasqServer, getIpForIgnoresNonIpv4AndMalformedLines)
+{
+    // A real leases file can start with a duid line and hold IPv6 entries or truncated/garbled
+    // lines. None of these should crash the lookup, and a valid in-subnet lease elsewhere in the
+    // file must still resolve. A matching MAC whose address field is unparseable is skipped, not
+    // fatal.
```
ricab, line 227: OK, I guess this answers the first of my questions before.

ricab (commented):
Yep, this fixes the issue for me. I have a questions for discussion, but the idea makes sense.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
xmkg: Assuming dnsmasq is the only authority on the lease file, I'd expect dnsmasq to lease that entry again, given if the lease is still not expired. If it's expired, in theory dnsmasq can allocate another IP address and keep the stale entry in the lease file. I'd say, in addition to the subnet check, it would be a good enhancement to check the lease expiry, too. Also, we should probably collect all of the entries matching the MAC instead of doing it greedily to see all the leases and pick the valid one.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
xmkg: This is to mimic dnsmasq's behavior -- it skips the unparseable lines on startup. If we do see an unparseable line here, we know that dnsmasq would've ignored it anyway (_*unless it's a legitimate lease from an address family we don't support, like IPv6_), meaning there's a good chance that we can still find the legitimate lease. Given that, I'd say catching it is better since letting it propagate is likely to make the VM unreachable. What do you think? Should we throw, or should I narrow the catch clause instead?

ricab: Makes sense. Let's go with the narrower catch then? Probably bad_alloc is the only other possibility today, but code could be added in the future.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
ricab: Yeah, I like both of those ideas. That's out of scope of this PR, but feel free to create an issue if you'd like.

xmkg: Great! I'll do a follow-up PR once this one is merged.

xmkg: Added TODOs to code.

src/platform/backends/qemu/linux/dnsmasq_server.cpp
xmkg: Done! Also added a debug log for skipped lease lines.

ricab (dismissed):
LGTM 👍

xmkg:
@ricab, can you approve again? GH removed the approval after the other branch got merged

ricab (approved):
Yep.

PR #5024, opened 2026-07-06
[gui] Avoid divide by 0 error caused by overflow

sharder996:
This PR changes how the number of columns of image cards is computed so that it's never 0. Having 0 columns causes a divide by 0 and the GUI goes into an infinite loop taking 100% CPU power and freezing the window. This bug can be replicated on `edge` by setting the GUI to its minimum wide.

---

MULTI-2758

### DIFF
--- a/src/client/gui/lib/catalogue/catalogue.dart
+++ b/src/client/gui/lib/catalogue/catalogue.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:math';
 
 import 'package:basics/basics.dart';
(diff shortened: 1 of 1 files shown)


ricab (approved):
Though I could not reproduce on Linux with edge, the fix makes sense to me. 

Both before and after the fix, I cannot make the window thinner than what's needed for a white-space surrounded card. No stdout or stderr messages either. I guess the Linux implementation has some protection for this... In any case, we should not divide by 0, so +1.

PR #5025, opened 2026-07-06
[gui] Make VM details header wrap w/ window width

sharder996:
This PR changes the behaviour of the header which is displayed in the GUI when the shell or instance details page is displayed. The header will now react to the width of the GUI and will wrap the window so that all header elements are always visible.

---

MULTI-2659

### DIFF
--- a/src/client/gui/lib/vm_details/vm_details_general.dart
+++ b/src/client/gui/lib/vm_details/vm_details_general.dart
@@ -182,7 +182,7 @@ class _RenderHeaderLayout extends RenderBox
           ? _placeNameRow(row, name, width, top)
           : _placeRow(row, actions, width, top);
     }
-    size = Size(width, top);
+    size = constraints.constrain(Size(width, top));
   }
 
   double _placeNameRow(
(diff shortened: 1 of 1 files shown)


src/client/gui/lib/vm_details/vm_details_general.dart
```diff
@@ -88,24 +91,156 @@ class VmDetailsHeader extends ConsumerWidget {
...
+
+    final usages = Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [cpu, memory, disk].gap(width: _headerGap).toList(),
+    );
 
     return Padding(
       padding: const EdgeInsets.all(20),
-      child: Row(children: list.gap(width: 40).toList()),
+      child: _HeaderLayout(
```
Laefy, line 106: Did you try using Wrap instead of Row?

sharder996: `Wrap` wouldn't work because I are specifying some custom behaviour when wrapping fields. Custom right alignment, truncating some fields with ellipsis, and the fact that some fields are wrapped together as one unit (i.e. system resource usages for VMs).

Laefy: Tested your branch to see the result, and indeed, it's gonna be hard to get something similar without special handling.
Approved :)

Laefy (approved):

PR #5030, opened 2026-07-06
Implement SftpSession API surface

### LINKED ISSUE #4952: SSHFS process channel leaks
**Describe the bug**
The SFTP session releases the ssh_channel on the SSHFS process, but it never frees it. Ownership is transferred but not properly adopted. Libssh's `sftp_server_new` doesn't transfer ownership and `sftp_server_free` does not free the channel.

https://github.com/canonical/multipass/blob/f4542c359969da40ec92dadb4a2592eea2f38337/src/sshfs_mount/sftp_server.cpp#L313
https://github.com/canonical/multipass/blob/f4542c359969da40ec92dadb4a2592eea2f38337/src/sshfs_mount/sftp_server.cpp#L65-L66

### LINKED ISSUE #4953: Bogus SSHFS health check
**Describe the bug**
The SftpServer calls `release_channel()` on the SSHFS `SSHProcess`, which nulls the channel field. But it later asks the process for the exit code, which internally registers an `ExitStatusCallback` with a null channel. An SSH_ERROR code is returned, but never handled. 

The process's exit code can never be observed and every health check times out silently. Recovery is driven by EOF alone.

https://github.com/canonical/multipass/blob/f4542c359969da40ec92dadb4a2592eea2f38337/src/sshfs_mount/sftp_server.cpp#L313
https://github.com/canonical/multipass/blob/b0ed7e3e077a2e1d596a2d57136a6dd4a8b1458c/src/ssh/ssh_process.cpp#L48

ricab:
# Description

Provide the interface that the SftpServer needs to replace SSHSession, SSHProcess, and raw libssh.

### Note: on the SftpMessage/PlainSftpMessage split and protocol vocabulary

This PR continues the trend of an interface/implementation split for the new Plain/SftpMessage types, abstracting libssh away from the interface level. Users of SftpSession can interact with the client side via SftpMessage objects and use an implementation-independent *vocabulary*, defined in sftp_protocol.h, instead of having to depend on libssh.

I debated this a lot. Although I don't fancy having to repeat constants that the program already has, and despite the few additional int copies per transaction (negligible next to encryption and network round-trips), I went with this design because:

- it gives tests a proper seam to mock messages and replies, without premock or real libssh
- it walks the path of encapsulating libssh, along with its C interface; SftpMessage offers a proper C++ interface instead (e.g. scoped enums, reply methods that reference the original object)
- the constants are fixed by the protocol, not specific to libssh; they are settled and not going to change, so there is no risk of skew
- even if there was, the values and underlying types are static_asserted against libssh's, so any divergence would fail to build (sort of like a build time test)
- there is precedent for this approach: `std::errc` does something similar with POSIX errno values, `std::filesystem::perms` hardcodes `owner_read = 0400`, `owner_write = 0200`, etc.

## Related Issue(s)

Closes #4952
Closes #4953

MULTI-2622
MULTI-2647
MULTI-2720

(Even though the bugs will only be gone once we hook this in, this is where the code changes to fix them are, so I am considering them addressed in here.)

## Testing

- Unit tests
- Manual testing steps:

  1.
  2.
  3.

## Checklist


include/multipass/ssh/ssh_key_provider.h
```diff
@@ -21,10 +21,11 @@
 
 #include <multipass/disabled_copy_move.h>
 
-#include <libssh/libssh.h>
-
 #include <string>
 
+struct ssh_key_struct;
+typedef ssh_key_struct* ssh_key;
+
```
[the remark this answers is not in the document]
ricab: > Any TU that includes this header and later includes libssh headers (directly or via libssh_wrapper.h) will fail to compile with a typedef redefinition. 

Wrong, unless perhaps there was a mismatch, in which case a compilation failure would be the desired outcome. However, to avoid any questions or ambiguity, I am moving the typedef to a using declaration inside the class that needs it.

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,205 @@
...
+        if (!raw_msg)
+            mpl::debug(category,
+                       "could not read the next client message: {}",
+                       MP_LIBSSH.ssh_get_error(plain_ssh_session.borrow_session(pass)));
+
+        break; // at this point, we either have a message or a connection drop/desync/malfunction
+    }
+
+    return raw_msg ? std::make_unique<PlainSftpMessage>(*raw_msg) : nullptr;
+}
```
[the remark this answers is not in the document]
ricab: Oh right, good catch! Fixed.

include/multipass/sshfs_mount/sftp_message.h
```diff
@@ -0,0 +1,183 @@
...
+
+namespace multipass
+{
+
+/**
+ * A unit of communication exchange between an SFTP client and an SFTP server, in either direction.
+ *
+ * Provides access to the client request along with methods to reply.
+ */
+class SftpMessage // TODO@sftp check if there is anythign unused at in the end
```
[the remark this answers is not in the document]
ricab: Right. Fixed.

include/multipass/ssh/ssh_session.h
```diff
@@ -19,14 +19,17 @@
...
 
-#include <libssh/libssh.h>
-
 #include <memory>
 #include <string>
 
+// TODO@sftp remove
+struct ssh_session_struct;
+typedef struct ssh_session_struct* ssh_session;
+
```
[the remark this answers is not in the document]
ricab: idem

include/multipass/ssh/plain_sftp_session.h
```diff
@@ -22,34 +22,100 @@
...
 #include <multipass/sshfs_mount/sftp_session.h>
 
-#include <libssh/sftp.h> // TODO@sftp avoid this include (need to go through MP_LIBSSH)
+#include <atomic>
+#include <chrono>
+#include <string>
+
+struct sftp_session_struct;
+struct ssh_channel_struct;
+typedef ssh_channel_struct* ssh_channel;
```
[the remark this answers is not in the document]
ricab: idem

src/sshfs_mount/sshfs_client_steward.cpp
```diff
@@ -0,0 +1,135 @@
...
+    catch (const std::exception& e)
+    {
+        mpl::debug(category, "'multipass-sshfs' snap package is not installed: {}", e.what());
+    }
+
+    // Fallback to default
+    try
+    {
+        return MP_UTILS.run_in_ssh_session(session, "sudo which sshfs");
+    }
```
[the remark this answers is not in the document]
ricab: Good point. Adding `-n` to the remaining `sudo` commands in this diff.

src/sshfs_mount/sshfs_client_steward.cpp
```diff
@@ -0,0 +1,135 @@
...
+                                                   const std::string& source) const
+{
+    const auto mount_path = [&session, &source] {
+        auto proc = session.exec(fmt::format("findmnt --source :{} -o TARGET -n", source));
+        return mp::utils::trim(proc->read_std_output());
+    }();
+
+    if (!mount_path.empty())
+        MP_UTILS.run_in_ssh_session(session, fmt::format("sudo umount {}", mount_path));
+}
```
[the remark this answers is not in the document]
ricab: Alright, moving paths to debug format on clean up too.

tests/unit/test_sshfsmount.cpp
```diff
@@ -354,21 +354,21 @@ INSTANTIATE_TEST_SUITE_P(SshfsMountThrowWhenError,
 // Commands to check that a version of FUSE smaller that 3 gives a correct answer.
 CommandVector old_fuse_cmds = {
     {"sudo env LD_LIBRARY_PATH=/foo/bar /baz/bin/sshfs -V", "FUSE library version: 2.9.0\n"},
```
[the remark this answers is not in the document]
ricab: These tests are outdated and will need to be replaced as we switch to the new SftpSession, so leaving it alone.

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,204 @@
...
     }
 
+    /* handles setting the sftp->client_version */
+    const sftp_client_message msg{MP_LIBSSH.sftp_get_client_message(raw_sftp_session.get())};
+    if (msg == nullptr)
+        throw SftpInitException{"Null client message"};
+
+    if (PlainSftpMessage wrapped{*msg}; wrapped.type() != mp::SftpMessageType::init)
+        throw SftpInitException{
+            fmt::format("FATAL: Packet read of type {} instead of SSH_FXP_INIT", msg->type)};
```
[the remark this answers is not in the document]
ricab: Oh indeed, duh...

include/multipass/sshfs_mount/sftp_message.h
```diff
@@ -0,0 +1,183 @@
...
+
+#pragma once
+
+#include <multipass/sshfs_mount/sftp_protocol.h>
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+
```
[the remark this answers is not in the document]
ricab: Done

tests/unit/test_plain_sftp_session.cpp
```diff
@@ -46,103 +57,535 @@ static_assert(!std::is_move_assignable_v<mp::PlainSftpSession>);
...
+        ON_CALL(mock_libssh, ssh_channel_read_timeout)
+            .WillByDefault([this](ssh_channel, void* dest, uint32_t count, int is_stderr, int) {
+                const auto& result = result_for(running_cmd);
+                const auto& stream = is_stderr ? result.std_err : result.std_out;
+                auto& read_so_far = bytes_read[{running_cmd, is_stderr != 0}];
+
+                // num_bytes will be zero once the stream is exhausted
+                const auto num_bytes = std::min<std::size_t>(count, stream.size() - read_so_far);
+
+                std::memcpy(dest, stream.data() + read_so_far, num_bytes);
```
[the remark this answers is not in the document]
ricab: Adding an expectation instead.

ricab:
Hey @tobe2098, sending this your way. It ended up becoming a lot larger than I would have liked, with all the tests and everything. If you prefer, I can try to split the tests out (though you'd still be the natural reviewer).

include/multipass/ssh/plain_sftp_message.h
```diff
@@ -0,0 +1,85 @@
...
+     * @param message A reference to the raw libssh SFTP message whose ownership is to be
+     * adopted by this PlainSftpMessage
+     */
+    explicit PlainSftpMessage(sftp_client_message_struct& message) noexcept;
+    PlainSftpMessage(const PlainSftpMessage&) = delete;
+    PlainSftpMessage& operator=(const PlainSftpMessage&) = delete;
+
+    // Make class final before enabling these
+    PlainSftpMessage(PlainSftpMessage&&) = delete;
+    PlainSftpMessage& operator=(PlainSftpMessage&&) = delete;
```
tobe2098, line 51: Class is already final, are the move operators still deleted intentionally?

include/multipass/sshfs_mount/sftp_message.h
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * @return The submessage name of an extended request (e.g. "posix-rename@openssh.com");
+     * nullopt if absent.
+     */
+    virtual std::optional<std::string_view> submessage() const noexcept = 0;
+
+    /**
+     * @return The access flags of an open request (a combination of SftpOpenFlags).
+     */
+    virtual uint32_t flags() const noexcept = 0;
```
tobe2098, line 72: This description is not correct. According to the spec, the flags are also used to store information. An example is which file attributes does the message contain (https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-5).

src/ssh/plain_sftp_message.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return {MP_LIBSSH.ssh_string_get_char(raw), MP_LIBSSH.ssh_string_len(raw)};
+
+    return {};
+}
+
+std::optional<std::string_view> mp::PlainSftpMessage::submessage() const noexcept
+{
+    const auto* raw = MP_LIBSSH.sftp_client_message_get_submessage(message.get());
+    return raw ? std::optional<std::string_view>{raw} : std::nullopt;
+}
```
tobe2098, line 77: `<optional>`. It does make sense to use it to represent `nullptr` returns, but then why not use it in `data`, `filename`, etc? Is there a reason for these decisions?

tobe2098: With the exception of already nullable types like ptrs

src/ssh/plain_sftp_message.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+bool mp::PlainSftpMessage::reply_data(const void* data, size_t len)
+{
+    if (len > static_cast<size_t>(std::numeric_limits<int>::max()))
+    {
+        assert(false && "data replies are bounded by packet size");
+        return false;
+    }
+
+    return MP_LIBSSH.sftp_reply_data(message.get(), data, static_cast<int>(len)) == SSH_OK;
```
tobe2098, line 141: This could be simplified with a `mp::saturate_cast<int>`, no?

include/multipass/sshfs_mount/sftp_session.h
```diff
@@ -17,10 +17,15 @@
...
+#include <memory>
+#include <optional>
+
 namespace multipass
 {
+class SftpMessage;
+
 /**
- * A server-side SFTP session.
+ * A local SFTP session, serving SFTP to a remote client.
```
tobe2098, line 28: Nitpick: the naming makes it a bit difficult to grasp that this is indeed an SFTP server, no? WDYT?

include/multipass/ssh/plain_ssh_process.h
```diff
@@ -64,23 +63,31 @@ class PlainSSHProcess : public SSHProcess
     // Obtain a non-owning libssh channel handle.
     // The caller adopts thread-safety responsibility for the channel with respect to this
     // SSHProcess and the SSHSession it belongs to.
-    ssh_channel borrow_channel(const PrivatePassProvider<PlainSftpSession>::PrivatePass&);
+    ssh_channel_struct* borrow_channel(const PrivatePassProvider<PlainSftpSession>::PrivatePass&);
```
tobe2098, line 66: This is probably just a nitpick, but I would alias the pointer type to `ssh_channel` like you did with `ssh_key`

include/multipass/ssh/plain_ssh_process.h
```diff
@@ -64,23 +63,31 @@ class PlainSSHProcess : public SSHProcess
...
+    };
+    using ChannelUPtr = std::unique_ptr<ssh_channel_struct, ChannelDeleter>;
+
     enum class StreamType
     {
         out,
         err
     };
 
+    static ChannelUPtr make_channel(ssh_session_struct* raw_session, const std::string& cmd);
```
tobe2098, line 81: idem on aliasing

src/ssh/sftp_wire_compat.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+static_assert(mp::SftpAttrFlags::size == SSH_FILEXFER_ATTR_SIZE);
+static_assert(mp::SftpAttrFlags::uidgid == SSH_FILEXFER_ATTR_UIDGID);
+static_assert(mp::SftpAttrFlags::permissions == SSH_FILEXFER_ATTR_PERMISSIONS);
+static_assert(mp::SftpAttrFlags::acmodtime == SSH_FILEXFER_ATTR_ACMODTIME);
+static_assert(mp::SftpAttrFlags::extended == SSH_FILEXFER_ATTR_EXTENDED);
+
+static_assert(mp::SftpFileMode::regular == SSH_S_IFREG);
+static_assert(mp::SftpFileMode::directory == SSH_S_IFDIR);
+static_assert(mp::SftpFileMode::symlink == SSH_S_IFLNK);
```
tobe2098, line 84: I could not find these in the ssh/sftp specs, did you find the source?

src/sshfs_mount/sshfs_mount.cpp
```diff
@@ -78,7 +78,7 @@ auto get_sshfs_exec_and_options(mp::SSHSession& session)
         }
     }
 
-    sshfs_exec = mp::utils::trim_end(sshfs_exec);
+    sshfs_exec = mp::utils::trim_end(sshfs_exec); // redundant
```
tobe2098, line 81: Should it have a `//TODO: remove` then?

tests/unit/test_plain_sftp_message.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+    EXPECT_THAT(msg->flags(), Eq(raw_flags));
+}
+
+TEST_F(TestPlainSftpMessage, offsetReadsRawField)
+{
+    raw_msg.offset = 123456789ull;
+    const auto msg = make_message();
+
+    EXPECT_THAT(msg->offset(), Eq(123456789ull));
```
tobe2098, line 148: Nitpick: would it be better to extract into a constexpr auto?

tests/unit/test_plain_sftp_message.cpp
```diff
@@ -0,0 +1,411 @@
...
+}
+
+TEST_F(TestPlainSftpMessage, lengthReadsRawField)
+{
+    raw_msg.len = 4096;
+    const auto msg = make_message();
+
+    EXPECT_THAT(msg->length(), Eq(4096u));
+}
```
tobe2098, line 157: idem for constexpr auto

src/sshfs_mount/sshfs_client_steward.cpp
```diff
@@ -0,0 +1,135 @@
+/*
```
tobe2098: I noticed that there is no logic for installing the sshfs snap or program in the steward. Where will it be placed?

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,205 @@
...
-            throw std::runtime_error(sshfs_process.read_std_error());
     }
+
+private:
+    constexpr static auto init_error_prefix = "[sftp] server init failed:"; // TODO square brackets?
+};
+
+void check_client_status(mp::SSHProcess& client_process)
+{
+    // should we have a way to wait for it to start running?
```
tobe2098, line 59: It is not possible without modifying the process. `ssh_channel_request_exec()` returns when the server confirms that the command was successfully received and dispatched by the SSH daemon, but not when execution starts. Once we go async, there is a callback that we can setup with a cv to wait for said confirmation.

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,205 @@
...
+    if (client_process.exit_recognized(client_exit_timeout))
+        throw mp::SSHException(client_process.read_std_error());
```
tobe2098, line 61: I understand that this is intended to throw with a 0 exit code, right? This slightly deviates from previous behavior.

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,205 @@
...
+      source{source},
+      client_cmd{client_steward.compose_client_command(plain_ssh_session, source, target)}
+{
+    spawn_client();
+}
+
+mp::PlainSftpSession::~PlainSftpSession()
+{
+    raw_sftp_session.reset(); // mind the order: this borrows the process's channel
+    client_process.reset();
```
tobe2098, line 157: Nitpick: This is pretty important and it is repeated twice in this file, should we maybe encapsulate in a function to ensure there is a single source of truth?

tobe2098: Also being a bit more explicit on the process variable name

src/ssh/plain_sftp_session.cpp
```diff
@@ -18,91 +18,205 @@
...
+        if (poll_result > 0) // bounded by session timeout
+            raw_msg = MP_LIBSSH.sftp_get_client_message(raw_sftp_session.get());
+        else if (poll_result == 0)
+            continue; // nothing to read yet
+
+        if (!raw_msg)
+            mpl::debug(category,
+                       "could not read the next client message: {}",
+                       MP_LIBSSH.ssh_get_error(plain_ssh_session.borrow_session(pass)));
+
```
tobe2098, line 210: A negative poll result could also mean EOF, in which case ssh_get_error will be empty (it was not an error). This can occur for example when the sshfs stops running. WDYT? Should we have a more exhaustive diagnostic of the result here?

tests/unit/test_plain_sftp_session.cpp
```diff
@@ -46,103 +57,536 @@ static_assert(!std::is_move_assignable_v<mp::PlainSftpSession>);
...
-                                                      fake_channel,
-                                                      sshfs_exit_code,
-                                                      channel_cbs->userdata);
+            if (const auto& exit_code = result_for(running_cmd).exit_code)
+                channel_cbs->channel_exit_status_function(fake_session,
+                                                          fake_channel,
+                                                          *exit_code,
+                                                          channel_cbs->userdata);
             return SSH_OK;
         });
```
tobe2098, line 106: I would recommend using the mechanism to manage callbacks that is in #5004 once it is merged and this stack is rebased.

tobe2098 (changes requested):
Good work Ricardo! Just some minor comments, it is a big PR XD

PR #5057, opened 2026-07-10
Unify VM resource minima

### LINKED ISSUE #3612: multipass launch with memory <1Gb fails on memtest
**Describe the bug**
`multipass -vvv launch 22.04 --cpus 1 --disk 10G --memory 128Mb --name memory-test` fails, unless I specify 1Gb of RAM.
Instance is then stuck - only server's hard reset helps.

**To Reproduce**
```bash
# multipass -vvv launch 22.04 --cpus 1 --disk 10G --memory 128Mb --name memory-test
[2024-07-30T17:51:50.400] [debug] [qemu-system-x86_64] [2939] started: qemu-system-x86_64 --version
[2024-07-30T17:51:51.179] [debug] [daemon] Applied AppArmor policy: multipass.qemu-img
[2024-07-30T17:51:51.184] [debug] [qemu-img] [2946] started: qemu-img info /var/snap/multipass/common/cache/multipassd/vault/images/jammy-20240726/ubuntu-22.04-server-cloudimg-amd64.img
[2024-07-30T17:51:51.277] [debug] [daemon] Applied AppArmor policy: multipass.qemu-img
[2024-07-30T17:51:51.281] [debug] [qemu-img] [2956] started: qemu-img resize /var/snap/multipass/common/data/multipassd/vault/instances/memory-test/ubuntu-22.04-server-cloudimg-amd64.img 10737418240
[2024-07-30T17:51:51.387] [debug] [daemon] Applied AppArmor policy: multipass.qemu-img
[2024-07-30T17:51:51.392] [debug] [qemu-img] [2965] started: qemu-img snapshot -l /var/snap/multipass/common/data/multipassd/vault/instances/memory-test/ubuntu-22.04-server-cloudimg-amd64.img
[2024-07-30T17:51:51.483] [debug] [daemon] Applied AppArmor policy: multipass.qemu-img
[2024-07-30T17:51:51.488] [debug] [qemu-img] [2974] started: qemu-img amend -o compat=1.1 /var/snap/multipass/common/data/multipassd/vault/instances/memory-test/ubuntu-22.04-server-cloudimg-amd64.img
[2024-07-30T17:51:51.713] [debug] [memory-test] process working dir '/snap/multipass/13001/qemu'
[2024-07-30T17:51:51.713] [info] [memory-test] process program 'qemu-system-x86_64'
[2024-07-30T17:51:51.713] [info] [memory-test] process arguments '-bios, OVMF.fd, --enable-kvm, -cpu, host, -nic, tap,ifname=tap-897722fc248,script=no,downscript=no,model=virtio-net-pci,mac=52:54:00:aa:47:93, -device, virtio-scsi-pci,id=scsi0, -drive, file=/var/snap/multipass/common/data/multipassd/vault/instances/memory-test/ubuntu-22.04-server-cloudimg-amd64.img,if=none,format=qcow2,discard=unmap,id=hda, -device, scsi-hd,drive=hda,bus=scsi0.0, -smp, 1, -m, 128M, -qmp, stdio, -chardev, null,id=char0, -serial, chardev:char0, -nographic, -cdrom, /var/snap/multipass/common/data/multipassd/vault/instances/memory-test/cloud-init-config.iso'
[2024-07-30T17:51:51.717] [debug] [qemu-system-x86_64] [2996] started: qemu-system-x86_64 -nographic -dump-vmstate /tmp/multipassd.YqFCtv
[2024-07-30T17:51:51.748] [debug] [daemon] Applied AppArmor policy: multipass.memory-test.qemu-system-x86_64
[2024-07-30T17:51:51.748] [info] [memory-test] process state changed to Starting
[2024-07-30T17:51:51.753] [info] [memory-test] process state changed to Running
[2024-07-30T17:51:51.753] [debug] [qemu-system-x86_64] [3001] started: qemu-system-x86_64 -bios OVMF.fd --enable-kvm -cpu host -nic tap,ifname=tap-897722fc248,script=no,downscript=no,model=virtio-net-pci,mac=52:54:00:aa:47:93 -device virtio-scsi-pci,id=scsi0 -drive file=/var/snap/multipass/common/data/multipassd/vault/instances/memory-test/ubuntu-22.04-server-cloudimg-amd64.img,if=none,format=qcow2,discard=unmap,id=hda -device scsi-hd,drive=hda,bus=scsi0.0 -smp 1 -m 128M -qmp stdio -chardev null,id=char0 -serial chardev:char0 -nographic -cdrom /var/snap/multipass/common/data/multipassd/vault/instances/memory-test/cloud-init-config.iso
[2024-07-30T17:51:51.753] [info] [memory-test] process started
memory-test: timed out waiting for response
```

**Expected behavior**
Instance to be created and run fine.

**Logs**
To be provided in the comments.

**Additional info**
 - OS:
```bash
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
```
- `multipass version`: 1.14.0
- `multipass info`:
```
Name:           memory-test
State:          Unknown
Snapshots:      0
IPv4:           --
Release:        --
Image hash:     82384f6c58ed (Ubuntu 22.04 LTS)
CPU(s):         --
Load:           --
Disk usage:     --
Memory usage:   --
Mounts:         --
```

**Additional context**

Instances creation works fine unless I specify memory size less than 1GB. Running on Ubuntu 22.04 with Ubuntu 22.04 - no changes.
Unless I get qemu's PID (with ps | grep mac_address) and `kill -9` it - multipass is stuck with that instance.

### LINKED ISSUE #3777: Unify VM resource minima in Daemon/CLI/GUI

0xSA7:
# Description

- **What does this PR do?**
  Updates the hardcoded minimums for memory (512M) and disk (1G) in the CLI and daemon so they match the GUI and [Ubuntu Core specs](https://documentation.ubuntu.com/core/reference/system-requirements/).
- **Why is this change needed?**
  The old limits (128M mem / 512M disk) were letting instances silently OOM crash during cloud-init, which caused the backend to hang indefinitely waiting for SSH. By enforcing these safer minimums early in the CLI/daemon, we catch the bad input before the gRPC payload is even sent, preventing the timeout loop.

## Related Issue(s)

Closes #3777
Closes #3612

## Testing

- **Unit tests:** Updated the hardcoded assertions in `test_instance_settings_handler.cpp`, `test_daemon.cpp`, and the formatter mocks to use the new boundaries. Ran the full CTest suite locally in Release mode (all passed).
- **Manual testing steps:**
  1. Built the `multipass` binaries locally.
  2. Ran `multipass launch --memory 128M --name test-the-patch`.
  3. Verified the CLI instantly rejects it with an invalid memory size error instead of hanging the hypervisor backend.

## Checklist

## Additional Notes

Hey @ricab and @sharder996,

Quick question about the test suite: while updating `TestInstanceModPersists` to make the CI happy, I noticed it fails if I set the mock memory to exactly `536870912` (512MB). I figured this is because the new default instance memory is now 512M, so passing 512M results in a no-op (no state mutation). This causes the `InstanceSettingsHandler` to skip the disk I/O write, which then naturally fails the `fake_persister_called` assertion.

I bumped the test value to `1073741824` (1GB) in this PR to force a state change and trigger the persister, which turned the tests green. Just wanted to double-check: is this No-Op I/O optimization the intended design for the persister, or is there a specific reason the old test was using the exact minimum boundary?

### DIFF
--- a/tests/unit/test_daemon.cpp
+++ b/tests/unit/test_daemon.cpp
@@ -1140,11 +1140,16 @@ INSTANTIATE_TEST_SUITE_P(Daemon, DaemonCreateLaunchTestSuite, Values("launch", "
 INSTANTIATE_TEST_SUITE_P(Daemon,
                          DaemonCreateLaunchPollinateDataTestSuite,
                          Combine(Values("launch", "test_create"), Values("foo", "")));
-INSTANTIATE_TEST_SUITE_P(Daemon,
+INSTANTIATE_TEST_SUITE_P(DaemonMemory,
                          MinSpaceRespectedSuite,
                          Combine(Values("test_create", "launch"),
-                                 Values("--memory", "--disk"),
-                                 Values("1024m", "2Gb", "987654321")));
+                                 Values("--memory"),
+                                 Values("512m", "1024m", "2Gb", "987654321")));
+INSTANTIATE_TEST_SUITE_P(DaemonDisk,
+                         MinSpaceRespectedSuite,
+                         Combine(Values("test_create", "launch"),
+                                 Values("--disk"),
+                                 Values("1024m", "2Gb", "1073741824", "9876543210")));
 INSTANTIATE_TEST_SUITE_P(Daemon,
                          MinSpaceViolatedSuite,
                          Combine(Values("test_create", "launch"),
--- a/tests/unit/test_instance_settings_handler.cpp
+++ b/tests/unit/test_instance_settings_handler.cpp
@@ -443,7 +443,7 @@ TEST_F(TestInstanceSettingsHandler, setMaintainsInstanceMemoryUntouchedIfSameBut
 TEST_F(TestInstanceSettingsHandler, setAllowsDecreaseInstanceMemory)
 {
     constexpr auto target_instance_name = "Dvořák";
-    constexpr auto less_mem_str = "256MiB";
+    const auto less_mem_str = mp::min_memory_size;
     const auto less_mem = mp::MemorySize{less_mem_str};
     const auto& actual_mem = specs[target_instance_name].mem_size = mp::MemorySize{"1024MiB"};
 
@@ -615,7 +615,7 @@ struct TestInstanceModOnStoppedInstance : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModOnStoppedInstance, setWorksOnOtherStates)
 {
     constexpr auto target_instance_name = "Beethoven";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    const auto val = "1500000000"; // exceed all minima
     const auto [property, state] = GetParam();
     const auto& target_specs = specs[target_instance_name];
 
@@ -642,7 +642,7 @@ struct TestInstanceModPersists : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModPersists, setPersistsInstances)
 {
     constexpr auto target_instance_name = "Tchaikovsky";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    constexpr auto val = "1500000000"; // used to force a state mutation
     const auto property = GetParam();
 
     specs[target_instance_name];
--- a/docs/reference/command-line-interface/launch.md
+++ b/docs/reference/command-line-interface/launch.md
@@ -51,11 +51,11 @@ Options:
   -d, --disk <disk>                     Disk space to allocate. Positive
                                         integers, in bytes, or decimals, with K,
                                         M, G suffix.
-                                        Minimum: 512M, default: 5G.
+                                        Minimum: 1G, default: 5G.
   -m, --memory <memory>                 Amount of memory to allocate. Positive
                                         integers, in bytes, or decimals, with K,
                                         M, G suffix.
-                                        Minimum: 128M, default: 1G.
+                                        Minimum: 512M, default: 1G.
   -n, --name <name>                     Name for the instance. If it is
                                         'primary' (the configured primary
                                         instance name), the user's home
--- a/include/multipass/constants.h
+++ b/include/multipass/constants.h
@@ -32,8 +32,8 @@ constexpr auto daily_remote = "daily";
 constexpr auto snapcraft_remote = "snapcraft";
 constexpr auto core_remote = "core";
 
-constexpr auto min_memory_size = "128M";
-constexpr auto min_disk_size = "512M";
+constexpr auto min_memory_size = "512M";
+constexpr auto min_disk_size = "1G";
 constexpr auto min_cpu_cores = "1";
 
 constexpr auto default_memory_size = "1G";


tobe2098:
Related #4183

tobe2098:
We might want to align with [Ubuntu Server - System Requirements](https://documentation.ubuntu.com/server/reference/installation/system-requirements/) and [Ubuntu Core - System Requirements](https://documentation.ubuntu.com/core/reference/system-requirements/) _both_.

tobe2098:
Do you think you could enforce the limits depending on the image? Server images have higher requirements than core.

0xSA7:
Hey @tobe2098 , exactly! I actually planned to implement that dynamic validation based on the image manifest, but I wanted to keep it in a separate PR. keeping this PR small and focused strictly on unifying the static baseline for Core workloads.

I'll open the follow-up PR for the dynamic Server image requirements shortly!

0xSA7:
I've force-pushed an update to resolve the recent CI blockers.
* Shortened the commit Message to pass the 50-character CheckCommits limit.
* Aligned my DCO `Signed-off-by` name with the Canonical CLA registry.

tests/unit/test_instance_settings_handler.cpp
```diff
@@ -642,7 +642,7 @@ struct TestInstanceModPersists : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModPersists, setPersistsInstances)
 {
     constexpr auto target_instance_name = "Tchaikovsky";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    constexpr auto val = "1073741824"; // To work with the 1G minimum allowed for memory
     const auto property = GetParam();
```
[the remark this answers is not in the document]
tobe2098: This comment does indeed need correction

tobe2098 (changes requested):
LGTM @0xSA7 ! Just change the comment and this is ready to merge!

tests/unit/test_instance_settings_handler.cpp
```diff
@@ -642,7 +642,7 @@ struct TestInstanceModPersists : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModPersists, setPersistsInstances)
 {
     constexpr auto target_instance_name = "Tchaikovsky";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    constexpr auto val = "1073741824"; // 1GB used to force a state mutation (actual minimum is 512M)
```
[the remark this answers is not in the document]
0xSA7: cool, makes more sense!

0xSA7:
@tobe2098 comments are updated! Just pinging u to clear the requested changes block so we can get the CI rolling again

0xSA7:
@xmkg I've fixed the Linter CI pipeline

tobe2098:
@0xSA7 , there is still a lint error

0xSA7:
@tobe2098 Lint is fixed! 
I have a feeling the only bug left is the one I originally came here to fix 😂

0xSA7:
Hey @tobe2098 , just a gentle ping All CI checks are green and the linting issue is fully resolved. Whenever you have a moment, could you dismiss the old Changes Requested block so we can get this merged? Thanks!

ricab (commented):
Hi @0xSA7, a couple of questions on this PR:

- Does snapcraft work well with these minima, when using multipass as a backend? Worth trying out.
- Doesn't the documentation matching updates? Or is it already sourcing these constants?

0xSA7:
Hi @ricab, thanks for reviewing!

- I'm actually not entirely familiar with how to properly test Multipass specifically as a backend for it in this context. Could you elaborate a bit on what exactly needs to be tested or point me to reproduce the scenario you have in mind? I want to make sure I validate exactly what you're concerned about before we merge.
- The documentation strings are hardcoded and don't source the constants directly. I will force-pushed an update to include the corrections in `launch.md` to reflect the new 512M/1G ninimum standard.

ricab:
@0xSA7 

> * I'm actually not entirely familiar with how to properly test Multipass specifically as a backend for it in this context. Could you elaborate a bit on what exactly needs to be tested or point me to reproduce the scenario you have in mind? I want to make sure I validate exactly what you're concerned about before we merge.

Snapcraft can use Multipass as a [build provider](https://ubuntu.com/docs/snapcraft/9/how-to/select-a-build-provider/). It uses some special images for that, which you can find with `multipass find snapcraft:`. So, for testing, I guess we'd need to verify that:

- that those images all still launch fine (this should be fine)
- Snapcraft still works with Multipass on all supported core bases (20-26). For this, you would need to `snap set snapcraft provider=multipass` and then snapcraft projects using the 4 supported bases.

> * The documentation strings are hardcoded and don't source the constants directly. I will force-pushed an update to include the corrections in `launch.md` to reflect the new 512M/1G ninimum standard.

:+1:

0xSA7:
Hi @ricab,

I spent some time validating the Snapcraft integration. Here is what I found:

- The specialized builder images (`multipass launch snapcraft:core22/core26`) can be launched properly by the modified daemon. Our `512M/1G` floor doesn't block their initial provisioning at all.

- For Snapcraft  I hit local gRPC certificate validation barriers when trying to force a confined snapcraft to talk to my unconfined, locally built socket

     In Snapcraft the `craft-providers` ([multipass_provider.py -> launched_environment](https://github.com/canonical/craft-providers/blob/dd5094a918bad26e48bfd32f6c78c5c5ae2751a2/craft_providers/multipass/multipass_provider.py#L302))  explicitly hardcodes generous baseline requests: `mem_gb=2` and `disk_gb=64`, So our new daemon will never reject its requests, Since Snapcraft never requests resources below our new `512M` absolute floor.

- I've pushed a new commit to update the hardcoded minimums in `docs/launch.md`. This seems to be the main place where the CLI help strings are documented. Please let me know if there are any other documentation files that need to be updated.

Let me know what you think!

tests/unit/test_instance_settings_handler.cpp
```diff
@@ -641,14 +615,13 @@ struct TestInstanceModOnStoppedInstance : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModOnStoppedInstance, setWorksOnOtherStates)
 {
     constexpr auto target_instance_name = "Beethoven";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    constexpr auto val = "536870912"; // 512MiB minimum memory
```
sergio-costas: Maybe getting this value directly from the configuration, to avoid having to update it in all places every time it is changed...?

ricab (commented):
Thanks for verifying @0xSA7. Changes look promising. Given what you found about snapcraft, I would be happy with a single test against it. 

CI will produce a snap for this PR. You can install it and let snapcraft use it, bypassing the barriers you mentioned earlier.

tests/unit/test_daemon.cpp
```diff
@@ -1144,7 +1144,7 @@ INSTANTIATE_TEST_SUITE_P(Daemon,
                          MinSpaceRespectedSuite,
                          Combine(Values("test_create", "launch"),
                                  Values("--memory", "--disk"),
-                                 Values("1024m", "2Gb", "987654321")));
+                                 Values("1024m", "2Gb", "1073741824")));
```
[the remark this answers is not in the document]
0xSA7: > Consider splitting the instantiation so --memory and --disk each get an appropriate set of passing values.

Should I split memory and disk tests? @sergio-costas @tobe2098

tests/unit/test_instance_settings_handler.cpp
0xSA7: Hi @sergio-costas, completely agree.

I've updated the tests to dynamically pull the minimum value

0xSA7:
> CI will produce a snap for this PR

Thanks for pointing that out, @ricab

I’ve downloaded the CI artifact ([multipass_1.17.0-dev.pr5057+gee5e1b32_amd64.snap](https://github.com/canonical/multipass/actions/runs/29738712346)), installed it locally, and tested it end-to-end with snapcraft.

Everything is running smoothly. The resource allocations (--cpus 2 --memory 2G --disk 64G) in the logs below were automatically generated by snapcraft itself when spinning up the managed core20 instance, not hardcoded by me:

```
2026-07-23 20:23:44.636 Using provider 'multipass' from snap config.
2026-07-23 20:23:44.636 Executing on host: multipass version
2026-07-23 20:23:45.308 Launching managed ubuntu 26.04 instance...
2026-07-23 20:23:45.310 Converted name 'snapcraft-test-core20-amd64-5024389' to instance name 'snapcraft-test-core20-amd64-5024389'
2026-07-23 20:23:45.310 Executing on host: multipass list --format json
2026-07-23 20:23:45.905 Executing on host: multipass launch snapcraft:resolute --name snapcraft-test-core20-amd64-5024389 --cpus 2 --memory 2G --disk 64G
```

Please let me know if there's anything else needed!

tests/unit/test_instance_settings_handler.cpp
```diff
@@ -669,14 +642,14 @@ struct TestInstanceModPersists : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModPersists, setPersistsInstances)
 {
     constexpr auto target_instance_name = "Tchaikovsky";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    // 1GiB used to force a state mutation (actual minimum is 512MiB)
+    constexpr auto val = "1073741824";
```
ricab: I don't think this is right yet. This is a parameterized test that is instantiated for disk, cpu, and memory. Disk is the one whose minimum is 1G. The value should be above that, right? 

Might as well just make this a sufficiently large number. Perhaps 50'000'000 or something. No actual allocation should take place.

tests/unit/test_instance_settings_handler.cpp
```diff
@@ -641,14 +615,13 @@ struct TestInstanceModOnStoppedInstance : public TestInstanceSettingsHandler,
 TEST_P(TestInstanceModOnStoppedInstance, setWorksOnOtherStates)
 {
     constexpr auto target_instance_name = "Beethoven";
-    constexpr auto val = "134217728"; // To work with the 128M minimum allowed for memory
+    const auto val = QString::number(mp::MemorySize{mp::min_memory_size}.in_bytes());
```
ricab: If we were going for constants, the highest minimum would be disk, not memory. Don't we need to go above the current value for it to stick, though?

IMO, we should just go with a sufficiently large value. Tests are all about redundancy. Yes, we could conceivably at some point increase a minimum above whatever we choose in these tests. But we might also be doing so by mistake, which these tests would highlight only if we hardcoded a big number independently. 

OTOH, if we were going to be strict with using defined constants we'd need to take the max minimum for the 3 numeric properties that these tests exercise.

ricab (changes requested):
Hey @0xSA7, requesting a couple of small changes in tests.

tests/unit/test_instance_settings_handler.cpp
0xSA7: the current value `1073741824` is exactly 1GiB. Since the new minimum for the disk is also 1G, does it fail to trigger the state mutation simply because it strictly matches the boundary rather than exceeding it?

I'm completely fine with changing it to an arbitrary large number like 50000000 to avoid these edge cases, but I just wanted to understand the exact mechanics you had in mind.

tests/unit/test_instance_settings_handler.cpp
0xSA7: would it be more suitable to use the `min_disk_size` here, since it represents the highest minimum boundary among the three, or should we just apply the same "sufficiently large hardcoded value" approach for simplicity?

I'm flexible to go with whichever approach aligns best with the project's testing philosophy

tests/unit/test_instance_settings_handler.cpp
ricab: The real settings refuse shrinking the disk and can implement the "setting the same value" as a simple no-op. I did not dig back into the tests enough to know whether this is also the case with the mock setup, but your comment on forcing state mutation seemed to imply so. It would make the test more aligned with expected behavior.

tests/unit/test_instance_settings_handler.cpp
ricab: See my reply above. You could also pull the minimum and add something. I am not sure it's worth the trouble...

0xSA7:
I hardcoded  `1500000000` (1.5 billion) for both tests.

This value was chosen because it safely exceeds the 1GiB disk minimum to guarantee a state mutation, while remaining within the 32-bit integer limit. This prevents the cpus parser from throwing an `InvalidSettingException` during the parameterized test iterations. (encountered when testing larger numbers)

Tell me what you think @ricab

ricab:
It makes sense to me @0xSA7. Thanks!

ricab (dismissed):
Giving my approval for the topics I reviewed. Not a full review though, I am leaving that to @tobe2098.

tests/unit/test_daemon.cpp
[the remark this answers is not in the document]
tobe2098: Yes, I think that would be an improvement

tobe2098 (changes requested):
Great work @0xSA7 ! It is really close, just a small thing

tests/unit/test_daemon.cpp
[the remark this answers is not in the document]
0xSA7: @tobe2098 Done, Waiting for CI

tobe2098 (approved):

tobe2098:
Good work @0xSA7 !

ahmedhosnypro:
Awesome 👍

PR #5063, opened 2026-07-10
[macOS/az] Override subnet isolation

sharder996:
When using vmnet-shared subnets on macOS, the OS will automatically install pf anchors in `com.apple.internet-sharing/network_isolation` with `block drop quick` between distinct subnets resulting in VMs in different zones not being able to communicate with each other. This PR adds a couple system commands to one, ensure that IP forwarding is enabled (it was already enabled on my machine, buts we set it just to make sure), and two, add additional pf anchors that get evaluated before the ones that are auto generated by the OS.

There are some disadvantages to this approach:
1. Using `pfctl` requires elevated privileges
2. PF rules are a system resource that can be modified by other services or even the OS. For that reason, we reassert the added pf anchors on various VM operations.

The alternative approach that I explored is to create a single shared `vmnet`-shared interface plus a userspace network fabric. One in-process L2 switch per AZ and an in-process L3 router that forwards between zones and out the uplink. This would be simpler to integrate with `applevz`, but more difficult with QEMU. Because of the complexity of this approach and the fact that this is blocking the AZ feature, the first approach was preferred. In the future, when we look at removing QEMU from use on macOS and reduce the privileges of the daemon process, this or a similar solution can be explores more.

---

MULTI-2761

### DIFF
--- a/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
+++ b/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
@@ -78,6 +78,9 @@ void AppleVZVirtualMachineFactory::hypervisor_health_check()
     {
         throw std::runtime_error("Virtualization is not supported on this system.");
     }
+
+    // Re-assert cross-zone routing in case macOS reloaded its pf anchors
+    enable_zone_routing(az_manager);
 }
 
 void AppleVZVirtualMachineFactory::remove_resources_for_impl(const std::string& name)
--- a/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
+++ b/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
@@ -19,8 +19,21 @@
 #include <applevz/applevz_virtual_machine.h>
 #include <applevz/applevz_virtual_machine_factory.h>
 #include <multipass/utils/qemu_img_utils.h>
+#include <shared/macos/backend_utils.h>
 
 namespace mp = multipass;
+
+namespace
+{
+void enable_zone_routing(mp::AvailabilityZoneManager& az_manager)
+{
+    std::vector<mp::Subnet> subnets;
+    for (const auto& zone : az_manager.get_zones())
+        subnets.push_back(zone.get().get_subnet());
+
+    mp::backend::enable_cross_zone_routing(subnets);
+}
+} // namespace
 
 namespace multipass::applevz
 {
@@ -30,6 +43,7 @@ AppleVZVirtualMachineFactory::AppleVZVirtualMachineFactory(const Path& data_dir,
           MP_UTILS.derive_instances_dir(data_dir, get_backend_directory_name(), instances_subdir),
           az_manager)
 {
+    enable_zone_routing(az_manager);
 }
 
 VirtualMachine::UPtr AppleVZVirtualMachineFactory::create_virtual_machine(
--- a/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
+++ b/src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
@@ -23,18 +23,6 @@
 
 namespace mp = multipass;
 
-namespace
-{
-void enable_zone_routing(mp::AvailabilityZoneManager& az_manager)
-{
-    std::vector<mp::Subnet> subnets;
-    for (const auto& zone : az_manager.get_zones())
-        subnets.push_back(zone.get().get_subnet());
-
-    mp::backend::enable_cross_zone_routing(subnets);
-}
-} // namespace
-
 namespace multipass::applevz
 {
 AppleVZVirtualMachineFactory::AppleVZVirtualMachineFactory(const Path& data_dir,
@@ -43,7 +31,7 @@ AppleVZVirtualMachineFactory::AppleVZVirtualMachineFactory(const Path& data_dir,
           MP_UTILS.derive_instances_dir(data_dir, get_backend_directory_name(), instances_subdir),
           az_manager)
 {
-    enable_zone_routing(az_manager);
+    mp::backend::enable_cross_zone_routing(az_manager);
 }
 
 VirtualMachine::UPtr AppleVZVirtualMachineFactory::create_virtual_machine(
@@ -80,7 +68,7 @@ void AppleVZVirtualMachineFactory::hypervisor_health_check()
     }
 
     // Re-assert cross-zone routing in case macOS reloaded its pf anchors
-    enable_zone_routing(az_manager);
+    mp::backend::enable_cross_zone_routing(az_manager);
 }
 
 void AppleVZVirtualMachineFactory::remove_resources_for_impl(const std::string& name)
--- a/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
+++ b/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
@@ -42,23 +42,13 @@ auto get_common_args(const QString& host_arch)
 
     return qemu_args;
 }
-
-void enable_zone_routing(const std::unordered_map<std::string, mp::Subnet>& bridges)
-{
-    std::vector<mp::Subnet> subnets;
-    subnets.reserve(bridges.size());
-    for (const auto& [_, subnet] : bridges)
-        subnets.push_back(subnet);
-
-    mp::backend::enable_cross_zone_routing(subnets);
-}
 } // namespace
 
 mp::QemuPlatformMacOS::QemuPlatformMacOS(const AvailabilityZoneManager& az_manager)
     : common_args{get_common_args(host_arch)}, az_manager{az_manager}
 
 {
-    enable_zone_routing(bridges);
+    mp::backend::enable_cross_zone_routing(az_manager);
 }
 
 std::optional<mp::IPAddress> mp::QemuPlatformMacOS::get_ip_for(const std::string& hw_addr)
@@ -75,7 +65,7 @@ void mp::QemuPlatformMacOS::platform_health_check()
     // TODO: Add appropriate health checks to ensure the QEMU backend will work as expected
 
     // Re-assert cross-zone routing in case macOS reloaded its pf anchors
-    enable_zone_routing(bridges);
+    mp::backend::enable_cross_zone_routing(az_manager);
 }
 
 QStringList mp::QemuPlatformMacOS::vmstate_platform_args()
--- a/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
+++ b/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
@@ -42,12 +42,23 @@ auto get_common_args(const QString& host_arch)
 
     return qemu_args;
 }
+
+void enable_zone_routing(const std::unordered_map<std::string, mp::Subnet>& bridges)
+{
+    std::vector<mp::Subnet> subnets;
+    subnets.reserve(bridges.size());
+    for (const auto& [_, subnet] : bridges)
+        subnets.push_back(subnet);
+
+    mp::backend::enable_cross_zone_routing(subnets);
+}
 } // namespace
 
 mp::QemuPlatformMacOS::QemuPlatformMacOS(const AvailabilityZoneManager& az_manager)
     : common_args{get_common_args(host_arch)}, az_manager{az_manager}
 
 {
+    enable_zone_routing(bridges);
 }
 
 std::optional<mp::IPAddress> mp::QemuPlatformMacOS::get_ip_for(const std::string& hw_addr)
--- a/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
+++ b/src/platform/backends/qemu/macos/qemu_platform_macos.cpp
@@ -73,6 +73,9 @@ void mp::QemuPlatformMacOS::remove_resources_for(const std::string& name)
 void mp::QemuPlatformMacOS::platform_health_check()
 {
     // TODO: Add appropriate health checks to ensure the QEMU backend will work as expected
+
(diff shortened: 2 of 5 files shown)


src/platform/backends/applevz/applevz_virtual_machine_factory.cpp
```diff
@@ -19,9 +19,22 @@
...
 
 namespace mp = multipass;
 
+namespace
+{
+void enable_zone_routing(mp::AvailabilityZoneManager& az_manager)
+{
+    std::vector<mp::Subnet> subnets;
+    for (const auto& zone : az_manager.get_zones())
+        subnets.push_back(zone.get().get_subnet());
```
jimporter: We do something like this three times in this patch (though the others also call `std::vector::reserve`). Since the backend function is "enable_cross_**zone**_routing", we could just pass the `AvailabilityZoneManager` into that directly, eliminating the need for these extra wrappers.

src/platform/backends/qemu/macos/qemu_platform_macos.cpp
```diff
@@ -42,6 +42,16 @@ auto get_common_args(const QString& host_arch)
...
+
+void enable_zone_routing(const std::unordered_map<std::string, mp::Subnet>& bridges)
+{
+    std::vector<mp::Subnet> subnets;
+    subnets.reserve(bridges.size());
+    for (const auto& [_, subnet] : bridges)
+        subnets.push_back(subnet);
+
+    mp::backend::enable_cross_zone_routing(subnets);
+}
```
jimporter: This is similar to above, except that we pull the subnets from a `std::unordered_map`. That mapping exists in the code because I was patterning the implementation after the Linux+qemu version. In light of the changes in this PR, I think that was the wrong way to go, and it would be better to have `QemuPlatformMacOS` behave more like the AppleVZ backend, especially since AppleVZ is the direction we're headed.

Instead of the unordered map of zone names to subnets, it would probably be nicer to keep around a reference to the AZ manager in `QemuPlatformMacOS`. I'll put up a small PR to do this shortly.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,41 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
 }
+
+void mp::backend::enable_cross_zone_routing(const std::vector<mp::Subnet>& zone_subnets)
+{
+    set_ip_forward();
+
+    std::vector<std::string> cidrs;
+    cidrs.reserve(zone_subnets.size());
+    for (const auto& subnet : zone_subnets)
+        cidrs.push_back(subnet.canonical().to_cidr());
```
jimporter: Rather than this, we could probably just get the canonical CIDRs in-line in the nested loop below. This might be ever-so-slightly more runtime efficient (at the cost of a bit more memory), but since N=3 (at least for now), I think we might as well aim for code simplicity.

jimporter (changes requested):
Overall, I like the reasoning of what you went with here. Just a couple small comments below. I filed a PR for the qemu platform changes I mentioned below: #5078.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,37 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+
+    // macOS automatically enforces isolation between subnets via a pf "block drop quick"
+    // rule. Override the rule by adding a "pass quick" rule in an earlier-evaluated anchor.
+    std::string rules;
+    for (const auto& src : zones)
+        for (const auto& dst : zones)
+            if (&src.get() != &dst.get())
+                rules += fmt::format("pass quick inet from {} to {} flags any keep state\n",
+                                     src.get().get_subnet().canonical().to_cidr(),
+                                     dst.get().get_subnet().canonical().to_cidr());
```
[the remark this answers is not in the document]
sharder996: Testing says otherwise.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,40 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+        return;
+
+    set_ip_forward();
+
+    // macOS automatically enforces isolation between subnets via a pf "block drop quick"
+    // rule. Override the rule by adding a "pass quick" rule in an earlier-evaluated anchor.
+    std::string rules;
+    for (const auto& src : zones)
+        for (const auto& dst : zones)
+            if (&src.get() != &dst.get())
```
jimporter: How about this? In practice it's equivalent (since each zone should have a non-overlapping subnet), but this lets us avoid pointer comparisons. The performance should be roughly the same, since `.get_subnet()` is just a dereference under the hood:

```suggestion
            if (src.get().get_subnet() != dst.get().get_subnet())
```

If you have strong(ish) feelings about the original way being better though, I don't think this is a critical issue.

tests/unit/macos/test_backend_utils.cpp
```diff
@@ -114,3 +117,91 @@ TEST_P(GetNeighbourIPInvalidInputsTests, inValidInputCases)
...
+    const auto mock_process_factory = mpt::MockProcessFactory::Inject();
+    auto [mock_utils, utils_guard] = mpt::MockUtils::inject();
+
+    const mp::Subnet subnet{"192.168.64.0/24"};
+    NiceMock<mpt::MockAvailabilityZone> zone;
+    ON_CALL(zone, get_subnet()).WillByDefault(ReturnRef(subnet));
+
+    NiceMock<mpt::MockAvailabilityZoneManager> mock_manager;
+    const std::vector<std::reference_wrapper<const mp::AvailabilityZone>> zones{zone};
+    EXPECT_CALL(mock_manager, get_zones()).WillRepeatedly(Return(zones));
```
jimporter: You could also use the `StubAvailabilityZoneManager` here:

```suggestion
    const mpt::StubAvailabilityZoneManager az_manager{"192.168.64.0/24"};
```

The constructor arg is optional really, since by default we'll create a zone manager with one zone:

```suggestion
    const mpt::StubAvailabilityZoneManager az_manager;
```

tests/unit/macos/test_backend_utils.cpp
```diff
@@ -114,3 +117,91 @@ TEST_P(GetNeighbourIPInvalidInputsTests, inValidInputCases)
...
+    NiceMock<mpt::MockAvailabilityZone> zone_a, zone_b, zone_c;
+    ON_CALL(zone_a, get_subnet()).WillByDefault(ReturnRef(subnet_a));
+    ON_CALL(zone_b, get_subnet()).WillByDefault(ReturnRef(subnet_b));
+    ON_CALL(zone_c, get_subnet()).WillByDefault(ReturnRef(subnet_c));
+
+    NiceMock<mpt::MockAvailabilityZoneManager> mock_manager;
+    const std::vector<std::reference_wrapper<const mp::AvailabilityZone>> zones{zone_a,
+                                                                                zone_b,
+                                                                                zone_c};
+    EXPECT_CALL(mock_manager, get_zones()).WillRepeatedly(Return(zones));
```
jimporter: Likewise here:

```suggestion
    const mpt::StubAvailabilityZoneManager az_manager{"192.168.64.0/24", "192.168.65.0/24", "192.168.66.0/24"};
```

tests/unit/macos/test_backend_utils.cpp
```diff
@@ -114,3 +117,91 @@ TEST_P(GetNeighbourIPInvalidInputsTests, inValidInputCases)
...
+            else
+            {
+                EXPECT_THAT(
+                    captured_rules,
+                    Not(HasSubstr(fmt::format("from {} to {}",
+                                              src.get().get_subnet().canonical().to_cidr(),
+                                              dst.get().get_subnet().canonical().to_cidr()))));
+            }
+        }
+    }
```
jimporter: This test is effectively just redoing the computations in `backend_utils.cpp`, so the test is mostly tautological. Since there are only 3*2 = 6 `pfctl` directives, it might be easier to write out the full string we expect.

```suggestion
    EXPECT_EQ(captured_rules,
              "pass quick inet from 192.168.64.0/24 to 192.168.65.0/24 flags any keep state\n"
              "pass quick inet from 192.168.64.0/24 to 192.168.66.0/24 flags any keep state\n"
              "pass quick inet from 192.168.65.0/24 to 192.168.64.0/24 flags any keep state\n"
              "pass quick inet from 192.168.65.0/24 to 192.168.66.0/24 flags any keep state\n"
              "pass quick inet from 192.168.66.0/24 to 192.168.64.0/24 flags any keep state\n"
              "pass quick inet from 192.168.66.0/24 to 192.168.65.0/24 flags any keep state\n");
```

Or if that makes the subnets too hard to read, how about something like this?

```suggestion
    EXPECT_EQ(captured_rules,
              fmt::format(
                  ("pass quick inet from {0} to {1} flags any keep state\n"
                   "pass quick inet from {0} to {2} flags any keep state\n"
                   "pass quick inet from {1} to {0} flags any keep state\n"
                   "pass quick inet from {1} to {2} flags any keep state\n"
                   "pass quick inet from {2} to {0} flags any keep state\n"
                   "pass quick inet from {2} to {1} flags any keep state\n"),
                  "192.168.64.0/24", "192.168.65.0/24", "192.168.66.0/24"));
```

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,40 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+                rules += fmt::format("pass quick inet from {} to {} flags any keep state\n",
+                                     src.get().get_subnet().canonical().to_cidr(),
+                                     dst.get().get_subnet().canonical().to_cidr());
```
jimporter, line 142: After reading past this a few times, I wonder: would it be better to write to the child process immediately here? Then you don't need the extra `rules` variable for accumulating these declarations:

```suggestion
                pfctl->write(QByteArray::fromStdString(
                    fmt::format("pass quick inet from {} to {} flags any keep state\n",
                                src.get().get_subnet().canonical().to_cidr(),
                                dst.get().get_subnet().canonical().to_cidr()));
```

(Not sure if that's how `clang-format` wants it indented, but you get the idea I hope.)

jimporter (changes requested):
This looks good overall, with just a few small stylistic things I noticed. Thanks for the unit tests as well.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,40 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+
+    const auto pfctl = mp::platform::make_process(
+        mp::simple_process_spec("pfctl", {"-a", pf_anchor, "-f", "-"}));
+    pfctl->start();
+    if (!pfctl->wait_for_started())
+    {
+        mpl::warn(category, "Failed to run pfctl");
+        return;
+    }
```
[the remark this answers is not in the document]
tobe2098: I think this suggestion has merit (if the method exists). When doing the future permissions work this will be useful

tobe2098 (commented):
Looks good Scott! A small addition to the copilot comment, but nothing else to add to Jim's and Copilot's reviews

src/platform/backends/shared/macos/backend_utils.cpp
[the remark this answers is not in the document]
jimporter: (Replying to the LLM comment)

If we do this (I have no opinion, and I'm skeptical that this will come up in practice), we'd want to fix the format string, `Process::error_string` would return something like: `program: pfctl; error: <some error message>`, so we don't need the `pfctl` in the format string here.

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,40 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+
+    pfctl->write(QByteArray::fromStdString(rules));
+    pfctl->close_write_channel();
```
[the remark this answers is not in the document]
jimporter: (Replying to the LLM comment)

A warning here if we have an incomplete write seems suspicious to me. First, I'm not sure this can happen unless things are catastrophically broken. Second, if they *are* that broken, the safest thing to do might be to terminate. (This really just seems like a typical Copilot-ism of wanting more error checks regardless of how helpful they really are.)

src/platform/backends/shared/macos/backend_utils.cpp
```diff
@@ -115,3 +122,40 @@ std::optional<mp::IPAddress> mp::backend::get_neighbour_ip(const std::string& ma
...
+
+    if (!pfctl->wait_for_finished() || !pfctl->process_state().completed_successfully())
+        mpl::warn(category,
+                  "Failed to install pf rules for cross-zone routing: {}\n{}",
+                  pfctl->process_state().failure_message(),
+                  pfctl->read_all_standard_error());
```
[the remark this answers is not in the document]
jimporter: (Replying to the LLM comment)

This is a `QProcess` under the hood. `QProcess` kills the process in its destructor if it's still alive. I don't think this is necessary.

tests/unit/macos/test_backend_utils.cpp
jimporter: I still think something like my previous comment (writing out the exact set of pfctl commands) would be helpful so that we don't have the same `fmt::format` call in both our source code and our unit tests.

jimporter (changes requested):

jimporter (approved):
Thanks, this all looks good now!

PR #5067, opened 2026-07-10
Applevz polish

sharder996:
This PR ties up a loose end with the applevz feature:
- Add the `applevz` driver option to the GUI using the string "Apple Virtualization framework"
- Realign documentation with new user-facing string for the `applevz` driver

### DIFF
--- a/src/client/gui/lib/platform/macos.dart
+++ b/src/client/gui/lib/platform/macos.dart
@@ -14,6 +14,7 @@ class MacOSPlatform extends MpPlatform {
   @override
   Map<String, String> get drivers => const {
         'qemu': 'QEMU',
+        'applevz': 'AppleVZ',
         'virtualbox': 'VirtualBox',
       };
 
--- a/feature-flags.cmake
+++ b/feature-flags.cmake
@@ -19,8 +19,5 @@ if(AVAILABILITY_ZONES_ENABLED)
   add_compile_definitions(AVAILABILITY_ZONES_FEATURE)
 endif()
 
-# Multipass backend integrating with Apple Virtualization framework
-feature_flag(APPLEVZ_ENABLED "AppleVZ backend" "APPLE;AVAILABILITY_ZONES_ENABLED")
-
 # The new Windows backend based on Hyper-V Host Compute System / Host Compute Networking APIs
 feature_flag(HYPERV_HCS_ENABLED "Hyper-V HCS backend" WIN32)
--- a/src/client/cli/cmd/set.cpp
+++ b/src/client/cli/cmd/set.cpp
@@ -44,6 +44,12 @@ mp::ReturnCodeVariant cmd::Set::run(mp::ArgParser* parser)
                 // We are in the client, messages cannot be sent
                 [[maybe_unused]] mp::UserMessages messages{};
                 MP_SETTINGS.set(key, val, messages);
+
+                // TODO(applevz): remove when qemu is replaced by applevz on macOS
+                if (key == driver_key && val == "qemu" &&
+                    MP_PLATFORM.is_backend_supported("applevz"))
+                    cerr << "Warning: the 'qemu' driver is deprecated on macOS and will be "
+                            "removed in a future release.\n";
             }
         }
         catch (const SettingsException& e)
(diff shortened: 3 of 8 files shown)


ricab:
Did we decide to deprecate QEMU on macOS?

sharder996:
@ricab I believe that is ultimately the goal. I understood from Christoph that this is what we are doing.

ricab:
I am not aware of that decision. Is it documented somewhere? AFAIK, QEMU remains the default and AppleVZ is still sort of beta and in any case untested in the wild. IMO, it would be more prudent not to commit to deprecation just yet (for 1.17). We can clarify with Christoph in a few weeks.

ricab (changes requested):
☝️

jimporter (changes requested):
Clearing out review for now. My understanding was similar to @ricab, that we were releasing AppleVZ in 1.17 as a preview, with deprecation of qemu on macOS to happen in 1.18-ish.

sharder996:
@chrisndbg Requesting your review to clear up the decision of adding deprecation warnings to the use of QEMU.

chrisndbg:
@ricab & @sharder996 - as discussed offline, we will not (yet) mark QEMU as deprecated in the product. This will happen once AppleVZ is feature complete and out of "preview".
Will will share our intentions with the community and users in the release notes and/or release announcements.

sharder996:
@chrisndbg @ricab Dropped the QEMU deprecation warnings for now and have the commits stored in another branch for when we do.

src/client/gui/lib/platform/macos.dart
```diff
@@ -14,6 +14,7 @@ class MacOSPlatform extends MpPlatform {
   @override
   Map<String, String> get drivers => const {
         'qemu': 'QEMU',
+        'applevz': 'AppleVZ',
```
ricab, line 17: Should the long string say "Virtualization Framework"?

For reference, the new Hyper-V said "Hyper-V API" and I switched to "Host Compute System".

See screenshot at https://github.com/canonical/multipass/pull/5226.

ricab (commented):
I am wondering about the user facing string in the GUI.

src/client/gui/lib/platform/macos.dart
sharder996: @chrisndbg

chrisndbg: As per today's decision, we go for “Apple Virtualization framework” & `applevz`

chrisndbg (changes requested):
This changes only the GUI but there are many occurrences of "AppleVZ" throughout the documentation, too.

Since we changed this to "Apple Virtualization framework", the according term must be used in the documentation, too.

chrisndbg (approved):
lgtm. Thanks!

ricab (commented):
The GUI isn't rendering the new name very well:

Image

It fits well when I click the drop down:

Image

I guess we need just a little more space.

src/client/gui/lib/settings/settings.dart
```diff
@@ -6,6 +6,8 @@ import 'usage_settings.dart';
 import 'virtualization_settings.dart';
 import 'about_section.dart';
 
+const settingFieldWidth = 270.0;
```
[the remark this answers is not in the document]
ricab: I don't think this causes any build/analysis issue, but it does increase coupling. Every settings child now depends on the parent and therefore they end up depending on each other. That can affect incremental rebuilds and tests cannot check one thing in isolation without pulling the whole settings page. 

I would say we'd better place the constant in a `settings/constants.dart` file and import that instead.

ricab (commented):
@sharder996, could you please address [this comment](https://github.com/canonical/multipass/pull/5067#discussion_r4086298662)?

ricab:
Also, it would be a good idea to create an issue for this, so that we can track it along with the rest in the Kanban board for the RC.

ricab (commented):
Thank you. The code looks good to me, but there are a couple of conflicts now and no CI package I can test. Once that's sorted, I'll approve if everything renders correctly on my end.

sharder996:
@ricab Sorry, I didn't notice. I have rebased on `main`.

ricab (approved):
Yep, that's it.

PR #5069, opened 2026-07-11
[daemon] Reject more vCPUs than the host has

### LINKED ISSUE #5061: Instance with no IP causes the daemon to fail `list` and `info`
**Describe the bug**
Launching an unsupported image under certain conditions can result in an instance without an IP. This causes both `info` and `list` to respond with `x failed: IP not available`. 

**Reproduced with**
Have a couple of ssh instances into the machine (the cpus argument is not necessary after the first failure to cause further failures even after purging the instances, may be unrelated to the issue).
```
antoni@MacBookPro ~ % multipass launch -n a a --cpus 12
a: timed out waiting for response
antoni@MacBookPro ~ % multipass ls
list failed: IP not available
antoni@MacBookPro ~ % multipass ls
list failed: IP not available
antoni@MacBookPro ~ % multipass delete a --purge
```

**Expected behavior**
An instance with no IP should be handled by the `Unknown` state, and in such state `list` and `info` should not query the IP. 

**Additional info**
- OS: macos 15
- CPU architecture or model:  x86
- driver: applevz (also observed by @geoffreynyaga on qemu)

ecgang:
### Problem

The daemon bounds a launch request's **memory** and **disk** against the host,
but there is no equivalent ceiling on the **CPU count**. Requesting more vCPUs
than the host has is accepted, and the resulting VM never gets an IP — it hangs
at boot and wedges the daemon, which then reports "IP not available" until the
instance is removed with `delete --purge`.

Reproducer (on an 8-core host):

```
multipass launch --cpus 12
```

The same over-provisioning is also reachable at runtime via
`multipass set local.<instance>.cpus=<n>` on a stopped instance.

### Fix

Reject a request for more vCPUs than the host reports, at both seams:

- **Launch** — `validate_create_arguments` now adds a new `INVALID_NUM_CORES`
  launch error when `num_cores > host`, and the client prints an actionable
  message.
- **`set cpus`** — `update_cpus` throws `InvalidSettingException` for the same
  condition.

Details:

- The bound is strict (`>`, not `>=`), so a VM may still claim every core.
- The `set cpus` check applies only to an **actual** change, so re-setting the
  current value stays a no-op even if that value already exceeds the host (e.g.
  a spec that outlived a move to a smaller host).
- Both seams skip the check when the host count cannot be determined
  (`get_cpus()` returning a non-positive value), so a failed probe never
  rejects an otherwise valid request.

### Testing

- New daemon tests: refuses a launch above the host count, allows exactly the
  host count, and still accepts a request when the host count is unavailable.
- New settings-handler tests: refuses `set cpus` above the host (leaving the
  spec untouched) and keeps a same-value set a no-op even when above the host.
- Full unit test suite passes.

Fixes #5061

### DIFF
--- a/src/client/cli/cmd/launch.cpp
+++ b/src/client/cli/cmd/launch.cpp
@@ -589,6 +589,13 @@ mp::ReturnCodeVariant cmd::Launch::request_launch(const ArgParser* parser)
                 error_details =
                     fmt::format("Invalid memory size value supplied: {}.", request.mem_size());
             }
+            else if (error == LaunchError::INVALID_NUM_CORES)
+            {
+                error_details = fmt::format(
+                    "Invalid number of CPUs requested: {}. It must not exceed the number "
+                    "available on the host.",
+                    request.num_cores());
+            }
             else if (error == LaunchError::INVALID_HOSTNAME)
             {
                 error_details =
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -468,6 +468,19 @@ auto validate_create_arguments(const mp::LaunchRequest* request, const mp::Daemo
         }
     }
 
+    // CPUs are the one launch resource with no host-side upper bound: unlike memory and
+    // disk (checked above / when reserving space), an over-provisioned core count is
+    // accepted and then wedges the instance at boot -- it times out, leaving the daemon
+    // reporting "IP not available" until the user runs `delete --purge`. Reject a request
+    // for more vCPUs than the host reports up front instead. See issue #5061.
+    // (A stricter "leave one core for the host" policy would be `>= host_cpus`; `>` is
+    // used here so a VM may still use every core.) The `host_cpus > 0` guard skips the
+    // check when the host core count can't be determined (get_cpus() returns -1), so a
+    // failed probe never rejects an otherwise valid request.
+    if (const auto num_cores = request->num_cores(), host_cpus = MP_PLATFORM.get_cpus();
+        num_cores > 0 && host_cpus > 0 && num_cores > host_cpus)
+        option_errors.add_error_codes(mp::LaunchError::INVALID_NUM_CORES);
+
     if (!instance_name.empty() && !mp::utils::valid_hostname(instance_name))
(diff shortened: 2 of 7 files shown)


tobe2098:
Thank you @ecgang for the contribution, although we would prefer if you could request assignment in the issue before you start working.

ecgang:
Thanks for the review and the pointer, @tobe2098 — and apologies for going straight to a PR here. Understood: request assignment on the issue before starting work. I'll follow that going forward rather than opening a PR cold. Happy to make any changes you'd like on this one in the meantime.

xmkg:
@ecgang I have doubts on whether we should cap the vCPU count of VMs on host's CPU core count, see: https://github.com/canonical/multipass/issues/5061#issuecomment-4955598556

ricab:
> I have doubts on whether we should cap the vCPU count of VMs on host's CPU core count, see: https://github.com/canonical/multipass/issues/5061#issuecomment-4955598556

FWIW, we discussed this a lot with design for the GUI. I thought the outcome was to allow higher allocations and place the sliders in a warning state, for example, when changing the allocation for an existing VM that was created with many CPUs from the CLI.

ecgang:
Thanks @xmkg, @ricab — fair correction, and you're right. vCPU count and host cores are orthogonal; oversubscription is legitimate, so a hard cap at the host core count is the wrong call and would block valid configs. Withdrawing that approach.

Reading the thread, the real problem is the daemon/`list` behavior: an instance that can't report an IP lands in a non-`Unknown` state, and `list` then blocks on IP retrieval — that's what wedges things.

@xmkg — I'd like to take the `list`-resilience piece you sketched: bound the per-instance IP fetch with a short timeout (~1s), fall back to the cached IP (or render it as pending) and dispatch an async refresh, then move on to the next VM, so one unreachable instance can't stall the whole listing. I'm also happy to take the early-start state fix (instance → `Unknown` when IP retrieval fails) if you'd rather start at the root cause.

Should this land on #5061 now that it's re-scoped to daemon state, or a fresh issue? I'll request assignment there before writing any code. I'll close this PR in the meantime.

PR #5082, opened 2026-07-16
[bugfix] Remove OpenSSL's global atexit cleanup

### LINKED ISSUE #4742: `multipass info` command intermittently crashes with double free or corruption
**Describe the bug**
As the title says:

```
ERROR tests/cli/cli_clone_test.py::TestClone::test_clone_to_self - RuntimeError: Failed parsing JSON content: {
    "errors": [
    ],
    "info": {
        "fangorn-galadriel-fellowship-6888": {
            "cpu_count": "2",
            "disks": {
                "sda1": {
                    "total": "6156561920",
                    "used": "2096821760"
                }
            },
            "image_hash": "7aa6d9f5e8a3a55c7445b138d31a73d1187871211b2b7da9da2e1a6cbf169b21",
            "image_release": "24.04 LTS",
            "ipv4": [
                "10.67.127.242"
            ],
            "load": [
                0.91,
                0.22,
                0.08
            ],
            "memory": {
                "total": 1000497152,
                "used": 361598976
            },
            "mounts": {
            },
            "release": "Ubuntu 24.04.4 LTS",
            "snapshot_count": "0",
            "state": "Running"
        }
    }
}
double free or corruption (fasttop)
```

**To Reproduce**
There are no reproducer at this time. It's an intermittent behavior observed during the CLI test execution in CI.

**Expected behavior**
No crashes.

### LINKED ISSUE #5034: `multipass exec` on a disabled VM occassionally causes a double free or corruption in the CLI (fasttop)
**Describe the bug**
```
ubuntu@copilot-sandbox:~/multipass/tests/manual/availability_zones$ multipass exec a -- true || echo "$?"
Cannot start the instance 'a' while unavailable.
double free or corruption (fasttop)
Aborted (core dumped)
134
```
This occurs with a frequency of ~3/400
**To Reproduce**
How, and what happened?
1. Launch a VM
2. Disable its zone
3. Run `multipass exec vm -- true || echo "$?"` a lot of times (~200) (sequentially within a shell or script).

**Expected behavior**
What did you expect to happen? No memory corruption

**Logs**
Nothing relevant in the logs, issue seems contained to the CLI

**Additional info**
- OS: Ubuntu 26.04

**Additional context**
Potentially related to #4742. It occurs with higher likelihood if `exec` requests are run concurrently, but concurrency is not yet supported. 

This also occurs with other commands (e.g., `list`), which means #4742 could be affecting all CLI commands.

**Potential causes**
Running exec/shell makes the client attempt to auto-start the instance: ssh_info returns ABORTED, so the exec client re-invokes `multipass start` in-process (see cmd::Exec::run -> run_cmd_and_retry in src/client/cli/cmd/exec.cpp). The start is rejected ("Cannot start the instance '<x>' while unavailable") and the client-side error/retry handling could be corrupting the heap.

### LINKED ISSUE #5081: [cli] Intermittent memory issues when exiting CLI
Every once in a while the CLI finishes its execution by printing out some memory related errors/warning. For example:

```
$ multipass ls
No instances found.
multipass(76281,0x16df6f000) malloc: *** error for object 0x7c9154180: pointer being freed was not allocated
multipass(76281,0x16df6f000) malloc: *** set a breakpoint in malloc_error_break to debug
```

I can usually trigger this within about 1K iterations of:

```
i=1; while multipass ls; do echo "Iteration: $((i++))"; done
```

Been noticing this on macOS 26.5, but have not tested on other platforms yet.

tobe2098:
# Description

- What does this PR do? The double free issue seems to be caused by a race condition between global SSL registered cleanup (by some non-Qt consumer) and Qt SSL cleanup (changing the SSL backend also fixes the issue). By disabling the atexit registered OpenSSL cleanup, the issue does not occur anymore.

## Related Issue(s)

Closes #5034
May close #4742
Closes #5081

## Testing

- Manual testing steps:

  1. Running 
`
for i in {1..5000}; do
        build/bin/multipass ls 1&>1 2&>1 || echo $?
done
` should not give any print, with any command.
  2. Running the same command in another branch will print a number of error codes (`134-139`).

```
┌─(main)*(1)
└─[λ] for i in {1..5000}; do
        build/bin/multipass ls 1&>1 2&>1 || echo $?
done

139
134
```

## Checklist

## Additional Notes

MULTI-2779

### DIFF
--- a/src/client/cli/main.cpp
+++ b/src/client/cli/main.cpp
@@ -23,6 +23,8 @@
 #include <multipass/ssh/libssh_scope_guard.h>
 #include <multipass/top_catch_all.h>
 
+#include <openssl/crypto.h>
+
 #include <QCoreApplication>
 
 namespace mp = multipass;
@@ -52,6 +54,11 @@ int main(int argc, char* argv[])
     // Verify that the version of the library that we linked against is
     // compatible with the version of the headers we compiled against.
     GOOGLE_PROTOBUF_VERIFY_VERSION;
+
+    // Prevent OpenSSL from registering an atexit() cleanup handler; this avoids shutdown races with
+    // gRPC cleanup.
+    if (OPENSSL_init_crypto(OPENSSL_INIT_NO_ATEXIT, nullptr) != 1)
+        return EXIT_FAILURE;
 
     multipass::LibsshScopeGuard libssh_guard;
 


src/client/cli/main.cpp
```diff
@@ -53,6 +55,11 @@ int main(int argc, char* argv[])
     // compatible with the version of the headers we compiled against.
     GOOGLE_PROTOBUF_VERIFY_VERSION;
 
+    // Prevent OpenSSL from registering an atexit() cleanup handler; this avoids shutdown races with
+    // gRPC cleanup.
+    if (OPENSSL_init_crypto(OPENSSL_INIT_NO_ATEXIT, nullptr) != 1)
+        return EXIT_FAILURE;
```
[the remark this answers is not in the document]
tobe2098: Logging is not functional at this point of time in runtime

src/client/cli/main.cpp
```diff
@@ -53,6 +55,11 @@ int main(int argc, char* argv[])
     // compatible with the version of the headers we compiled against.
     GOOGLE_PROTOBUF_VERIFY_VERSION;
 
+    // Prevent OpenSSL from registering an atexit() cleanup handler; this avoids shutdown races with
+    // gRPC cleanup.
+    if (OPENSSL_init_crypto(OPENSSL_INIT_NO_ATEXIT, nullptr) != 1)
```
xmkg, line 60: The same race would exist in the daemon too, no? Shouldn't we do the same in the daemon's main?

xmkg (changes requested):

src/client/cli/main.cpp
xmkg, line 61: Also, should we make the grpc_init/shutdown explicit while we're at it?

tobe2098: Grpc initialization is done lazily and it works well, this issue seems more related to the bundling of multiple openssl consumers with one statically linked openssl). I attempted doing that to resolve the issue, but it did not have an effect on the double frees, and as such it is out of scope for this PR. Feel free to submit a PR for that though :D

src/client/cli/main.cpp
tobe2098: Potentially, but the issue has not been observed there, and I would not add changes that are not necessary for now. 

Although one of the causes is Qt, it is not clear to me that grpc is the other cause, or how exactly is the race condition being set up, since the trace only shows the end state. Unless this was also experienced in the daemon/gui I would not introduce changes on those yet. WDYT?

xmkg: Sounds good.

xmkg (approved):
LGTM

PR #5084, opened 2026-07-16
[qemu/arm] Migrate suspend snapshots from QEMU v8.2.4 to v10.0.3

### LINKED ISSUE #5053: Suspended VM resume fails on arm mac after 1.16.3 -> main upgrade
```
-[2026-07-09T20:54:55.074] [debug] [upg-suspend] QMP: {"QMP": {"version": {"qemu": {"micro": 3, "minor": 0, "major": 10}, "package": ""}, "capabilities": ["oob"]}}

\[2026-07-09T20:54:55.161] [warning] [upg-suspend] qemu-system-aarch64: Size mismatch: 0000:00:01.0/virtio-net-pci.rom: 0x40000 != 0x80000: Invalid argument
qemu-system-aarch64: error while loading state for instance 0x0 of device 'ram'

[2026-07-09T20:54:55.161] [warning] [/Library/Application Support/com.canonical.multipass/bin/qemu-system-aarch64] 
[2026-07-09T20:54:55.161] [warning] [upg-suspend] qemu-system-aarch64: Error -22 while loading VM state

```

We should vendor the `virtio-net-pci.rom` like we did it for the other firmware files.

sharder996:
This PR addresses an issue where a VM suspended with Multipass v1.16.3 (QEMU v8.2.4) and resumed with an edge build of Multipass (QEMU v10.0.3) fails.

One of the problems with suspend snapshots was already addressed by reverting the offending upstream commit. This approach worked, but it was not optimal as we would have to continue carrying the patch forward indefinitely until such a time that we decided to introduce a version compatibility break. That patch has been replaced with a different approach that migrates the offending pieces of data loaded from the suspend snapshot. This [patch](https://github.com/canonical/multipass/compare/bugfix/qemu-suspend-state?expand=1#diff-57b85cf4f0e177cc1d252c75af587e75db9c1f2d79d8b538c6c817b1ad5861d4) can then be dropped in a subsequent release of Multipass (assuming that users have since resumed their VMs in this release of Multipass).

The second problem addresses a mismatch in CPU register data. Both these patches introduce code that is selectively activated for VMs that were suspended from QEMU 8.2.4. This [patch](https://github.com/canonical/multipass/compare/bugfix/qemu-suspend-state?expand=1#diff-c7183b6b795473ee980ad6d8922c4c923b08abad318fb356a91df6f1ce5b3935) follow the same criteria as the first and can be removed in an identical fashion.

The third issue was introduced when we changed the ROM file that is included with QEMU. On macOS, we need a different ROM file from Linux hosts.

Fixes #5053 

---

MULTI-2773


sharder996:
@tobe2098 Can you test on an Intel Mac please?

tobe2098:
Tested in intel mac, works as intended (suspended instance starts properly in new version)

3rd-party/vcpkg-ports/qemu/multipass-patches/0004-migration-remap-legacy-id-aa64pfr1-el1-key.patch
```diff
@@ -0,0 +1,67 @@
...
++     * sysreg encoding (op2=2 instead of op2=1), so an instance suspended by
++     * such a release records that register's value under the op2=2 key. Rewrite
++     * the legacy key to the correct one on load so it matches this QEMU's
++     * register list. The rewrite is order-preserving (the two keys have no
++     * other list entry between them) and a no-op for native suspends, which
++     * never contain the legacy key. Drop once instances suspended by the
++     * affected releases are no longer supported.
++     */
++    {
++        const uint64_t legacy_pfr1 = cpreg_to_kvm_id(
```
xmkg, line 62: Since this compatibility issue is specific to old ARM HVF snapshots, could we guard the remap with hvf_enabled() and only rewrite the legacy key when the correct PFR1 key is absent?

3rd-party/vcpkg-ports/qemu/multipass-patches/0006-migration-accept-subset-feature-id-registers.patch
```diff
@@ -0,0 +1,83 @@
...
++             * state load in cpu_post_load(). For feature-ID-space registers,
++             * tolerate the mismatch when the incoming value is field-wise not
++             * greater than the host's -- i.e. the guest ran with a subset of
++             * the host's features. The guest keeps the host's (superset) value,
++             * which is safe under HVF since the hardware supports those
++             * features. This is host- and feature-agnostic: no per-CPU list of
++             * features to mask, and a no-op for natively-suspended instances
++             * whose values already match.
++             */
++            if (!(arm_cpreg_in_idspace(ri) &&
```
xmkg: Could we scope this to hvf_enabled() and only allow the known 0 -> 1 BTI (ID_AA64PFR1_EL1) and I8MM (ID_AA64ISAR1_EL1) transitions, requiring all other bits to match? The current field-wise comparison is broader than needed.

xmkg (changes requested):

3rd-party/vcpkg-ports/qemu/portfile.cmake
```diff
@@ -283,10 +284,18 @@ if("system" IN_LIST FEATURES)
...
+    # mismatch on resume aborts the load. Older releases sourced this ROM
+    # differently per platform, so we must match each platform's historical size:
+    #   - Linux staged Ubuntu's `ipxe-qemu` ROM (512 KB) -> ship the vendored copy.
+    #   - macOS bundled QEMU's in-tree ROM (256 KB)       -> ship the in-tree copy.
+    # See vendor/README.md for details.
+    if(VCPKG_TARGET_IS_LINUX)
+        file(COPY "${CMAKE_CURRENT_LIST_DIR}/vendor/efi-virtio.rom" DESTINATION "${CURRENT_PACKAGES_DIR}/Resources/qemu")
+    else()
+        file(COPY "${SOURCE_PATH}/pc-bios/efi-virtio.rom" DESTINATION "${CURRENT_PACKAGES_DIR}/Resources/qemu")
+    endif()
```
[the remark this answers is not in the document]
xmkg: @sharder996 could we include this hunk as well ^

xmkg:
Verified with version upgrade tests; all pass.

xmkg (approved):
LGTM, thanks Scott!

PR #5097, opened 2026-07-22
Hyper-V API: Use HcnQueryEndpointProperties for retrieving VM's IP address

xmkg:
The existing hostname resolution approach for getting the VM IP is unreliable since the hosts.ics file tends to break on its own due to a longstanding Windows bug. Using the HCN API for querying the info will make the IP resolution much more robust than what we already have. Also introduced a `GetIpNetTable2` based fallback lookup when the HCN API yields no IP addresses. This is to overcome HCN API's shortcomings for querying IP addresses for the ICS networks.

MULTI-2897


xmkg:
I've found an issue. I'll mark this PR as ready once I fix it.

src/network/ip_address.cpp
```diff
@@ -38,18 +39,18 @@ bool is_valid_octet(int value)
...
-    s >> a >> ch >> b >> ch >> c >> ch >> d;
+    s >> octets[0] >> sep[0] >> octets[1] >> sep[1] >> octets[2] >> sep[2] >> octets[3];
 
-    if (!is_valid_octet(a) || !is_valid_octet(b) || !is_valid_octet(c) || !is_valid_octet(d))
+    if (!std::ranges::all_of(octets, is_valid_octet) ||
+        !std::ranges::all_of(sep, [](char c) { return c == '.'; }))
         throw std::invalid_argument(fmt::format("invalid IP address {}", ip));
 
-    return {{as_octet(a), as_octet(b), as_octet(c), as_octet(d)}};
+    return {{as_octet(octets[0]), as_octet(octets[1]), as_octet(octets[2]), as_octet(octets[3])}};
```
jimporter, line 50: What do you think about using `sscanf` here? `scanf` and friends are pretty inflexible, and not the best choice for *user* input. However, as a temporary solution until we migrate to Boost.ASIO, maybe it would be a clean way to do this?

Something like this would probably work:

```c++
    int a, b, c, d;
    if (std::sscanf(ip.c_str(), "%d.%d.%d.%d", &a, &b, &c, &d) != 4)
        throw std::invalid_argument(fmt::format("invalid IP address {}", ip));
```

xmkg:
I've also ported the fix to the old hyper-v backend since it's easy to do.

ricab:
Not a review, I just wanted to say that it is great that we won't have to depend on that file any more! :clap:

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
```diff
@@ -186,6 +187,52 @@ std::pair<OperationResult, UniqueHcnNetwork> open_network(const std::string& net
...
+
+    UniqueHcnEndpoint endpoint{};
+    const auto result = perform_hcn_operation([&](auto&& rmsgbuf) {
+        return API().HcnOpenEndpoint(guid_from_string(endpoint_guid), out_ptr(endpoint), rmsgbuf);
+    });
+
+    return std::make_pair(result, std::move(endpoint));
+}
+
+std::optional<std::vector<std::string>> endpoint_ip_addresses(const boost::json::object& endpoint)
```
jimporter, line 202: For deserializing JSON data, I've mostly tried to use [`tag_invoke`](https://www.boost.org/doc/libs/latest/libs/json/doc/html/conversion/custom_conversions.html). This case is a little different from usual, since we only want to deserialize some of the data. However, the rest of the code might be clearer if we had a `tag_invoke` overload like so:

```c++
HcnEndpointInfo tag_invoke(const value_to_tag<HcnEndpointInfo>&, const value& json)
{ /* ... */ }
```

(This function would then extract the IP addresses and MAC address for `HcnEndpointInfo`.)

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
```diff
@@ -186,6 +187,52 @@ std::pair<OperationResult, UniqueHcnNetwork> open_network(const std::string& net
...
+
+            if (const auto* address = configuration.as_object().if_contains("IpAddress");
+                address && !append_address(*address))
+                return std::nullopt;
+        }
+    }
+
+    if (const auto* address = endpoint.if_contains("IPAddress");
+        address && !append_address(*address))
+        return std::nullopt;
```
jimporter, line 231: I don't see this field in [`HostComputeEndpoint`](https://learn.microsoft.com/en-us/virtualization/api/hcn/hns_schema#HostComputeEndpoint). Is it from an older version of the API? If so, maybe we should check the `SchemaVersion` or at least add a comment here since Microsoft's documentation of these APIs is very sparse.

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
```diff
@@ -263,6 +310,59 @@ OperationResult HCNWrapper::delete_endpoint(const std::string& endpoint_guid) co
...
+        return {E_UNEXPECTED, L"Failed to process JSON returned from the API"};
+
+    std::optional<std::string> mac_address;
+    if (const auto* value = endpoint_properties.if_contains("MacAddress"))
+    {
+        if (!value->is_string())
+            return {E_UNEXPECTED, L"Failed to process JSON returned from the API"};
+
+        mac_address = value->as_string();
+    }
```
jimporter, line 357: Would it be simpler here to use exceptions? Then you could have something like the following (assuming suitable changes to `endpoint_ip_address` or something similar):

```c++
    try
    {
        const auto parsed = boost::json::parse(properties_as_str);
        out_info = {
            .mac_address = parsed.at("MacAddress").as_string(),
            .ip_addresses = endpoint_ip_addresses(endpoint_properties),
        };
        // Alternately, set `out_info = value_to<HcnEndpointInfo>(parsed);` if there's a `tag_invoke` overload.
    }
    catch(const std::exception& e)
    {
        mpl::error(log_category, "query_endpoint(...): failed to parse JSON: {}", e.what());
        return {E_UNEXPECTED, L"Failed to process JSON returned from the API"};
    }
```

Another option if you want to keep using error codes would be to use the `tag_invoke` overload as described above and do something like this:

```c++
    std::error_code ec;
    if (const auto parsed = boost::json::parse(properties_as_str, ec); !ec)
    {
        if (auto res = boost::json::try_value_to<HcnEndpointInfo>(parsed))
        {
            out_info = std::move(*res);
            return;
        }
    }
    return {E_UNEXPECTED, L"Failed to process JSON returned from the API"};
```

src/platform/backends/shared/windows/network_utils.cpp
jimporter, line 1: I added a file for some of the netioapi functions at `src/platform/backends/hyperv_api/net_io_api.cpp` as part of #5169. We should probably merge these, though I don't have any preference on which file we keep.

src/platform/backends/shared/windows/network_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+} // namespace
+
+WindowsNetworkUtils::WindowsNetworkUtils(
+    const Singleton<WindowsNetworkUtils>::PrivatePass& pass) noexcept
+    : Singleton<WindowsNetworkUtils>{pass}
+{
+}
+
+std::optional<std::string> WindowsNetworkUtils::permanent_ipv4_neighbor(
```
jimporter, line 70: Unlike a lot of the other Hyper-V API `Singleton`s, this is a higher-level wrapper. There's a reasonable amount of logic in here that we have to mock out in our unit tests. While I don't think we need to unit test this logic in its own, would it be a good idea to have the mockable part just call `GetIpNetTable2` and return a `std::unique_ptr` of the table? Then the rest of the logic below could be in an ordinary free function that gets tested indirectly when we run the Hyper-V unit tests.

src/platform/backends/shared/windows/network_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+    auto canonical_mac = mac_address;
+    std::ranges::replace(canonical_mac, '-', ':');
+    if (!utils::valid_mac_address(canonical_mac))
+    {
+        logging::error(log_category, "Invalid MAC address `{}`", mac_address);
+        return std::nullopt;
+    }
+    std::ranges::transform(canonical_mac, canonical_mac.begin(), [](unsigned char character) {
+        return static_cast<char>(std::tolower(character));
+    });
```
jimporter, line 82: Since we already need to perform some transformations to canonicalize the input MAC address, maybe we should format it the same way `GetIpNetTable2` does? Then the call below...

src/platform/backends/shared/windows/network_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+    if (!table)
+    {
+        logging::error(log_category, "GetIpNetTable2 returned no neighbor table");
+        return std::nullopt;
+    }
+
+    const auto matches = [&canonical_mac](const MIB_IPNET_ROW2& row) {
+        return row.Address.si_family == AF_INET && row.State == NlnsPermanent &&
+               row.PhysicalAddressLength == ethernet_address_length &&
+               canonical_mac_address(row.PhysicalAddress) == canonical_mac;
```
jimporter, line 102: ...  could just be a `memcmp`.

(I'm not worried about perf here. I doubt a few string conversions will be noticeable, but doing this would let us remove `canonical_mac_address` above.)

src/platform/backends/shared/windows/network_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+        return std::nullopt;
+    }
+
+    const std::unique_ptr<MIB_IPNET_TABLE2, decltype(&FreeMibTable)> table{raw_table,
+                                                                           &FreeMibTable};
+    if (!table)
+    {
+        logging::error(log_category, "GetIpNetTable2 returned no neighbor table");
+        return std::nullopt;
+    }
```
jimporter, line 97: Is it possible for this to happen? I'd hope that if `GetIpNetTable2` allocates a valid table if if returns `NO_ERROR`. (If it doesn't, something has gone seriously wrong and crashing might be the best response.)

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+        return network_parameters;
+    }();
+
+    const auto endpoint_parameters = [&network_parameters]() {
+        hyperv::hcn::CreateEndpointParameters endpoint_parameters{};
+        endpoint_parameters.network_guid = network_parameters.guid;
+        endpoint_parameters.endpoint_guid = "aee79cf9-54d1-4653-81fb-8110db97029f";
+        endpoint_parameters.mac_address = "52-54-00-E9-36-7E";
+        return endpoint_parameters;
+    }();
```
jimporter, line 80: We could use C++20 designated initializers here:

```suggestion
    const hyperv::hcn::CreateNetworkParameters network_parameters = {
        .name = "multipass-hyperv-cit",
        .flags = hyperv::hcn::HcnNetworkFlags::enable_dhcp_server,
        .guid = "b4d77a0e-2507-45f0-99aa-c638f3e47486",
        .ipams = {{.type = hyperv::hcn::HcnIpamType::Static(),
                   .subnets = {hyperv::hcn::HcnSubnet{"10.99.99.0/24"}}}}};

    const hyperv::hcn::CreateEndpointParameters endpoint_parameters = {
        .network_guid = network_parameters.guid,
        .endpoint_guid = "aee79cf9-54d1-4653-81fb-8110db97029f",
        .mac_address = "52-54-00-E9-36-7E"};
```

I think these could also go into the `HyperV_ComponentIntegrationTests` class above so they can be reused across the other test cases. Most/all of the tests in this file use these structs.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+            output << input.rdbuf();
+        }
+    }
+
+    const auto network_adapter = [&endpoint_parameters]() {
+        hyperv::hcs::HcsNetworkAdapter network_adapter{};
+        network_adapter.endpoint_guid = endpoint_parameters.endpoint_guid;
+        network_adapter.mac_address = *endpoint_parameters.mac_address;
+        return network_adapter;
+    }();
```
jimporter, line 114:
```suggestion
    const hyperv::hcs::HcsNetworkAdapter network_adapter = {
        .endpoint_guid = endpoint_parameters.endpoint_guid,
        .mac_address = *endpoint_parameters.mac_address};
```

I think this could also go into `HyperV_ComponentIntegrationTests`.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+        return vm_parameters;
+    }();
+
+    if (HCS().open_compute_system(create_vm_parameters.name, handle))
+    {
+        (void)HCS().terminate_compute_system(handle);
+        handle.reset();
+    }
+    (void)HCN().delete_endpoint(endpoint_parameters.endpoint_guid);
+    (void)HCN().delete_network(network_parameters.guid);
```
jimporter, line 139: All the test cases here do this (or something very much like it). Can we move this logic to `HyperV_ComponentIntegrationTests` as well?

For the last two lines, another test has this comment:
```
// Remove remnants from previous tests, if any.
```
It'd probably be worth mentioning this here (or in the common location in `HyperV_ComponentIntegrationTests`).

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+    {
+        const auto& [status, status_msg] = HCN().create_network(network_parameters);
+        ASSERT_TRUE(status.success());
+    }
+
+    // Create the test endpoint
+    {
+        const auto& [status, status_msg] = HCN().create_endpoint(endpoint_parameters);
+        ASSERT_TRUE(status.success());
+    }
```
jimporter, line 151: I believe these can also go into `HyperV_ComponentIntegrationTests`.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+    ASSERT_TRUE(endpoint_info.mac_address);
+
+    std::optional<std::string> neighbor_address;
+    for (auto attempts = 0; attempts < 120 && !neighbor_address; ++attempts)
+    {
+        neighbor_address = windows_network_utils().permanent_ipv4_neighbor(
+            *endpoint_info.mac_address);
+        if (!neighbor_address)
+            std::this_thread::sleep_for(500ms);
+    }
```
jimporter, line 182: Is the problem here just that the Win32 APIs take a bit until they reflect the state we want to test? It'd be nice if there were some event we could wait for, but if not, a comment mentioning what's going on (to the best of your knowledge) would be plenty for me.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+{
+    hyperv::hcs::HcsSystemHandle handle{nullptr};
+    const auto network_parameters = []() {
+        hyperv::hcn::CreateNetworkParameters parameters{};
+        parameters.name = "multipass-hyperv-hcn-ip-cit";
+        parameters.guid = "b4d77a0e-2507-45f0-99aa-c638f3e47487";
+        parameters.ipams = {hyperv::hcn::HcnIpam{hyperv::hcn::HcnIpamType::Static(),
+                                                 {hyperv::hcn::HcnSubnet{"10.99.100.0/24"}}}};
+        return parameters;
+    }();
```
jimporter, line 202: As above, this could use designated initializers and/or move to a common location in `HyperV_ComponentIntegrationTests`.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,29 +16,271 @@
...
+        const auto& [status, status_msg] = HCN().create_network(network_parameters);
+        ASSERT_TRUE(status.success());
+        ASSERT_TRUE(status_msg.empty());
+    }
+
+    {
+        const auto& [status, status_msg] = HCN().create_endpoint(endpoint_parameters);
+        ASSERT_TRUE(status.success());
+        ASSERT_TRUE(status_msg.empty());
+    }
```
jimporter, line 239: I think all of this could go into `HyperV_ComponentIntegrationTests` to be shared among all the test cases.

tests/unit/hyperv_api/test_it_hyperv_hcn_api.cpp
```diff
@@ -166,6 +171,45 @@ TEST_F(HyperVHCNAPI_IntegrationTests, create_delete_endpoint)
...
+TEST_F(HyperVHCNAPI_IntegrationTests, query_endpoint_returns_host_assigned_ipv4)
+{
+    CreateNetworkParameters network_params{};
+    network_params.name = "multipass-hyperv-api-hcn-query-endpoint-test";
+    network_params.guid = "b70c479d-f808-4053-aafa-705bc15b6d68";
+    network_params.ipams = {HcnIpam{HcnIpamType::Static(), {HcnSubnet{"172.50.224.0/20"}}}};
+
+    CreateEndpointParameters endpoint_params{};
+    endpoint_params.network_guid = network_params.guid;
+    endpoint_params.endpoint_guid = "b70c479d-f808-4053-aafa-705bc15b6d70";
```
jimporter, line 183: These could use designated initializers:

```suggestion
    const CreateNetworkParameters network_params{
        .name = "multipass-hyperv-api-hcn-query-endpoint-test",
        .guid = "b70c479d-f808-4053-aafa-705bc15b6d68",
        .ipams = {{.type = HcnIpamType::Static(),
                   .subnets = {HcnSubnet{"172.50.224.0/20"}}}}};

    const CreateEndpointParameters endpoint_params{
        .network_guid = network_params.guid,
        .endpoint_guid = "b70c479d-f808-4053-aafa-705bc15b6d70"};
```

They could also probably go into the `HyperVHCNAPI_IntegrationTests` fixture, but that might be a bit too much extra work for this PR. Maybe as a followup if you'd rather not do it right now?

tests/unit/hyperv_api/test_ut_hyperv_hcn_api.cpp
```diff
@@ -62,6 +63,45 @@ struct HyperVHCNAPI_UnitTests : public ::testing::Test
...
+        logger_scope.mock_logger->expect_log(mpl::Level::trace,
+                                             "HCNWrapper::query_endpoint(...) > endpoint_guid: "
+                                             "af3fb745-2f23-463c-8ded-443f876d9e81");
+        logger_scope.mock_logger->expect_log(
+            mpl::Level::trace,
+            "open_endpoint(...) > endpoint_guid: af3fb745-2f23-463c-8ded-443f876d9e81");
+        logger_scope.mock_logger->expect_log(mpl::Level::trace,
+                                             "perform_hcn_operation(...) > result: true",
+                                             testing::Exactly(2));
+        logger_scope.mock_logger->expect_log(mpl::Level::trace, "query_endpoint result:");
```
jimporter, line 103: I don't have strong feelings on this, so if you prefer it this way I don't mind. However, I worry that checking the trace-level logs would make it harder to modify the real code later; if we adjust logging for other reasons (e.g. to ease debugging from users), we'll also have to update the tests. If these checks add value, then I don't see an issue with them, but if not, then dropping them might be better.

I suppose one benefit of checking the logs is that we can make sure that each block of code with a log statement actually gets run as expected, though on the other hand I think that's checking internal implementation details that a unit test doesn't need to worry about.

In any case, it's up to you. I just wanted to call attention to this block.

(The same applies below to other `expect_log` calls, though I didn't comment on each block separately.)

tests/unit/hyperv_api/test_ut_hyperv_hcn_api.cpp
```diff
@@ -756,12 +796,148 @@ TEST_F(HyperVHCNAPI_UnitTests, delete_endpoint_failure)
...
+    hcn::HcnEndpointInfo endpoint_info;
+    const auto result = HCN().query_endpoint("af3fb745-2f23-463c-8ded-443f876d9e81", endpoint_info);
+
+    ASSERT_TRUE(result);
+    ASSERT_EQ(endpoint_info.ip_addresses.size(), 2);
+    EXPECT_EQ(endpoint_info.ip_addresses[0], "fe80::1");
+    EXPECT_EQ(endpoint_info.ip_addresses[1], "172.20.1.2");
+}
+
+TEST_F(HyperVHCNAPI_UnitTests, query_endpoint_rejects_malformed_properties)
```
jimporter, line 915: If you add a `tag_invoke` overload for deserializing the `HcnEndpointInfo`, then this test (and the one after it) could just test `boost::json::value_to<HcnEndpointInfo>`.

jimporter (changes requested):
Thanks, based on what I've heard from Christoph and Geoffrey, this will be a major improvement to reliability on Windows. (The old code works for me in my test VM, so I wouldn't notice a difference.)

Some comments below, but they're mostly minor things. Overall, this is looking good.

src/network/ip_address.cpp
xmkg: Done

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: Done

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: Done

tests/unit/hyperv_api/test_ut_hyperv_hcn_api.cpp
xmkg: Done

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: Yeah, unfortunately, the MS docs are pretty thin. This field comes from the real downstream consumers like hcsshim and WSL. Here are a couple of references:

https://github.com/microsoft/dotnet-computevirtualization/blob/e1cbbeeea5c9063c96c9a567c5879adb89d2f1ca/src/Microsoft.Windows.ComputeVirtualization/Schema/Networking.cs#L94-L173

https://github.com/microsoft/hcsshim/blob/4712998fa57874aa7c28963702ba0c2213e3ebbf/hcn/hcnendpoint.go#L15-L43

https://github.com/microsoft/WSL/blob/74c6ac9abb77f336bd4c796146c23cdb90877ed7/src/windows/common/WslCoreNetworkEndpointSettings.cpp#L10-L22

I've updated the code to request schema version 2 explicitly and added a comment about the IPAddress fallback.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Yeah, it takes a bit for the internal tables to be updated, and there are no events to listen to. I've added a comment to clarify that.

tests/unit/hyperv_api/test_ut_hyperv_hcn_api.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

tests/unit/hyperv_api/test_it_hyperv_hcn_api.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

xmkg:
@jimporter, ready for another round of review.

src/platform/backends/hyperv_api/hcn/hyperv_hcn_endpoint_info.cpp
```diff
@@ -0,0 +1,51 @@
...
+    {
+        for (const auto& configuration : configurations->as_array())
+        {
+            if (const auto* address = configuration.as_object().if_contains("IpAddress"))
+                append_address(*address);
+        }
+    }
+
+    // HCN schema v2 uses IpConfigurations[].IpAddress. Top-level IPAddress belongs to
+    // the legacy HNS schema-v1 endpoint projection, which some Windows versions may return.
```
jimporter, line 45: 👍 Once we get Ubuntu Insights up and running, we can check what versions of HCS are in use out in the wild so that we have better info about what versions we should support.

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
```diff
@@ -263,6 +277,49 @@ OperationResult HCNWrapper::delete_endpoint(const std::string& endpoint_guid) co
...
+    mpl::trace(log_category, "HCNWrapper::query_endpoint(...) > endpoint_guid: {}", endpoint_guid);
+
+    out_info = {};
+
+    const auto& [open_result, endpoint] = open_endpoint(endpoint_guid);
+    if (!open_result)
+        return open_result;
+
+    UniqueCotaskmemString properties{};
+    constexpr auto query = LR"({"SchemaVersion":{"Major":2,"Minor":0}})";
```
jimporter, line 292: I see this is new. Just checking: this doesn't cause any problems with Windows systems using schema v1.0 (see the comment you added above in `hyperv_hcn_endpoint_info.cpp`), right?

src/platform/backends/shared/windows/network_utils.cpp
```diff
@@ -0,0 +1,108 @@
...
+        return std::nullopt;
+
+    std::array<unsigned char, ethernet_address_length> address;
+    for (std::size_t index = 0; index < address.size(); ++index)
+    {
+        unsigned int octet;
+        const auto* begin = mac_address.data() + index * 3;
+        const auto [end, error] = std::from_chars(begin, begin + 2, octet, 16);
+        if (error != std::errc{} || end != begin + 2)
+            return std::nullopt;
```
jimporter: Can this happen? If we know the MAC address is valid, then I think any failures here would be a sign that something has gone seriously wrong, and it might be best to abort.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,91 +16,311 @@
...
+        remove_resources();
+        vm_name.clear();
+        endpoint_guids.clear();
+        network_guid.clear();
+    }
+
+    auto cleanup_guard()
+    {
+        return sg::make_scope_guard([this]() noexcept { cleanup_resources(); });
+    }
```
jimporter: Is this necessary? From looking at the tests below, I think it's sufficient to rely on `TearDown` to call `cleanup_resources`. Both the scope guard and `TearDown` should run at nearly the same time in all cases if I'm understanding the tests.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
```diff
@@ -16,91 +16,311 @@
...
+        .guid = "b4d77a0e-2507-45f0-99aa-c638f3e47487",
+        .ipams = {{.type = hyperv::hcn::HcnIpamType::Static(),
+                   .subnets = {hyperv::hcn::HcnSubnet{"10.99.100.0/24"}}}}};
+
+    const std::string vm_name{"multipass-hyperv-hcn-ip-cit-vm"};
+    const std::string mac_address{"00:15:5d:9d:cf:69"};
+    const hyperv::hcn::CreateEndpointParameters endpoint_parameters{
+        .network_guid = network_parameters.guid,
+        .endpoint_guid = "db4bdbf0-dc14-407f-9780-00155d9dcf69",
+        .mac_address = "00-15-5D-9D-CF-69"};
```
jimporter, line 257: Can these use `make_network_parameters` and `make_endpoint_parameters`? No worries if not though; it just wasn't clear to me if the values here need to be different, or if we can use those utility functions.

jimporter (dismissed):
Thanks for all the changes to the tests. The big bang integration tests especially are a lot easier to read to see what's happening now. Just a few small questions below. I've approved this in case the answer to all the questions is that there's nothing left to do for them. :)

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: Yes, this is just a preference request. If the system supports the 2.0 schema, it'll respond with it. If not, we also support the old HNS schema as a fallback.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: I think I forgot to remove this after introducing the teardown. I'll remove it.

src/platform/backends/shared/windows/network_utils.cpp
xmkg: Done

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: The endpoint GUID has to be set to this specific value so the VM can pick it up. I'll refactor later to move the VM-class tests out of this suite, and then these params will be redundant.

tests/unit/hyperv_api/test_bb_cit_hyperv.cpp
xmkg: Done

jimporter (approved):
Everything looks good to me. Let's merge this and publish our first test candidate for 1.17!

Issue #4904, opened 2026-05-27
[az] Pre-validate zone names before performing operations


jimporter:
**What are you trying to do?**

Currently, when disabling an AZ, we always prompt the user for confirmation first. This has two issues:
1. If the AZ name is invalid, we still prompt the user, even though the command can never succeed.
2. If there are no instances in the AZ, it's probably unnecessary to prompt at all.

Similarly, if enabling/disabling multiple zones, this causes us to stop updating the zones' states after the first invalid name, so we end up with a partially-successful operation. It would be better to check the names first, and then start operating on them to reduce the problems here. (You could still end up with partial success if the actual enabling/disabling failed for some reason, but that can't really be avoided.)

PR #5103, opened 2026-07-24
[az] Pre-validate zone names in the CLI

### LINKED ISSUE #4692: [az] Allow specifying zones by index on the CLI
Currently, our availability zones (on the `availability-zones` branch) are just named "zone1", "zone2", and "zone3". When launching a VM, it would be nice if we didn't have to specify the "zone" part of the name, as in `--zone=zone1`.

To solve this, we could allow referencing zones by index (starting at 1), so `--zone=1` would work.

jimporter:
# Description

This PR ensures that we first validate zone names when specified on the CLI, and additionally support specifying the zone as a index, starting at 1.

## Related Issue(s)

Closes #4692, #4904

## Testing

- Unit tests updated to account for these changes
- Manual testing steps:

  1. Check `enable-zones` with valid/invalid zone names and zone index (inc invalid indices)
  2. Check `disable-zones` as above, verify that invalid message happens immediately (not after the confirmation prompt)
  3. Check that `launch` handles valid zones, errors immediately on invalid

## Checklist

### DIFF
--- a/include/multipass/cli/dispatch_rpc.h
+++ b/include/multipass/cli/dispatch_rpc.h
@@ -1,0 +1,181 @@
+/*
+ * Copyright (C) Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+#include <multipass/callable_traits.h>
+#include <multipass/cli/return_codes.h>
+#include <multipass/format.h>
+#include <multipass/reply_concepts.h>
+#include <multipass/utils.h>
+
+#include <QLocalSocket>
+#include <QString>
+
+#include <grpc++/grpc++.h>
+
+namespace multipass::cmd
+{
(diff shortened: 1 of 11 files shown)


src/client/cli/cmd/common_cli.h
sharder996, line 1: Was there a specific reason why we are putting logic in the client and removing it from the daemon? Just thinking that we might need to replicate the same logic again for the GUI.

include/multipass/cli/dispatch_rpc.h
sharder996, line 1: I just want to double check. I don't notice any change in actual code functionality in this diff. Was this just to improve code organization?

sharder996 (commented):
Overall, this looks good. My main concern is that we are moving logic from the daemon to the client and that we might need to replicate (and keep in sync) this for the GUI as well.

include/multipass/cli/dispatch_rpc.h
jimporter: Yeah, all this is doing is pulling the relevant functions out of the class so they can be used as free functions.

src/client/cli/cmd/common_cli.h
jimporter: I'm open to moving this to the daemon side if we ever change how our GUI works, but my logic here was that in the CLI, we want to be able to convert a string from the user into the canonical zone name (so, e.g. "1" becomes "zone1"). In the GUI, the user would probably just select the zone they want out of a list, e.g. by clicking. Today at least, we don't need to support a GUI user typing in "1" and converting that to "zone1". The common data we need in both cases is just the full list of zones, which the GUI uses (in the `az-add-gui` branch).

This PR is really just adding the necessary scaffolding for the client to fetch the same data and do the string-based normalization.

sharder996: Ok, I can see how the pre-validation might have different implementations based on the client, so I'm okay with this being in the client.

sharder996 (approved):
LGTM! Thanks for answering my questions!

PR #5114, opened 2026-07-27
fix/firewall-detection-logic

### LINKED ISSUE #5113: Firewall backend selection logic
## Description

The firewall detection logic incorrectly tries to infer which iptables backend is active by checking which backend contains rules. This approach is fundamentally flawed because **iptables-nft and iptables-legacy share the same kernel netfilter ruleset**.

## Steps to Reproduce

1. System configured with `update-alternatives` to use iptables-nft as default
2. System has rules from multiple sources (e.g., Kubernetes using iptables-nft, Multipass using iptables-legacy)
3. Observe that Multipass uses iptables-legacy despite system preference being iptables-nft

## Expected Behavior

Multipass should respect the system's default iptables configuration and use the backend that `/usr/sbin/iptables` points to.

## Root Cause

The current `detect_firewall()` function tries to determine the active backend by checking if rules exist in each backend. This fails on hybrid systems because both backends share the same ruleset - rules added via one backend appear in both.

## The Fix

The fix implements a two-level detection strategy:
1. **Primary**: Check `/usr/sbin/iptables` symlink to determine system preference
2. **Fallback**: Only use rule presence detection when system config is unavailable

This properly handles systems where:
- Multiple tools use different iptables backends
- Both backends have rules present
- System configuration indicates a preference

## Related Files

- `src/platform/backends/qemu/linux/firewall_config.cpp`

See PR #5114 for the implementation.

Franck-Sorel:
## Summary

The firewall detection logic incorrectly tries to infer which iptables backend is active by checking which backend contains rules. This approach is fundamentally flawed because **iptables-nft and iptables-legacy share the same kernel netfilter ruleset**.

## Root Cause

The current detection logic in `detect_firewall()` attempts to determine the active backend by checking if rules exist in each backend:

```cpp
firewall_exec = kernel_supports_nftables() &&
                (is_firewall_in_use(nftables) || !is_firewall_in_use(iptables))
                    ? nftables
                    : iptables;
```

This fails on hybrid systems where multiple tools use different backends (e.g., Multipass using legacy, Kubernetes using nftables), leading to incorrect backend selection.

## The Fix

The fix implements a two-level detection strategy:

1. **Primary**: Check `/usr/sbin/iptables` symlink to determine system preference
2. **Fallback**: Only use rule presence detection when system config is unavailable

This properly handles systems where:
- Multiple tools use different iptables backends
- Both backends have rules present
- System configuration indicates a preference

## Changes

- Added `detect_firewall_from_system_config()` function to read `/usr/sbin/iptables` symlink
- Modified `detect_firewall()` to use system configuration as the primary detection method
- Falls back to rule-based detection only when system config is unavailable

## Testing

The fix should be tested on:
- Systems with iptables-nft as default (Debian 10+, Ubuntu 20.10+)
- Systems with iptables-legacy as default (older systems)
- Hybrid systems where multiple iptables backends have rules

## Related Issues

- Fixes #5113

### DIFF
--- a/src/platform/backends/qemu/linux/firewall_config.cpp
+++ b/src/platform/backends/qemu/linux/firewall_config.cpp
@@ -326,22 +326,6 @@ void clear_firewall_rules_for(const QString& firewall,
     }
 }
 
-bool is_firewall_in_use(const QString& firewall)
-{
-
-    return std::any_of(
-        firewall_tables.cbegin(),
-        firewall_tables.cend(),
-        [&firewall](const QString& table) {
-            QRegularExpression re{"^-[ARIN]"};
-            auto rule_lines = get_firewall_rules(firewall, table).split('\n');
-
-            return std::any_of(rule_lines.cbegin(), rule_lines.cend(), [&re](const QString& line) {
-                return re.match(line).hasMatch();
-            });
-        });
-}
-
 // We require a >= 5.2 kernel to avoid weird conflicts with xtables and support for inet table NAT
 // rules. Taken from LXD :)
 bool kernel_supports_nftables()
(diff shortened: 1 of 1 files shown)


ricab (changes requested):
There are two wrong premises underlying this PR:

- that iptables-legacy and iptables-nft share tables.
- that you can determine the system preference by looking at `/usr/sbin/iptables`

So, I think this is misguided. I have provided a [longer explanation on the issue](https://github.com/canonical/multipass/issues/5113#issuecomment-5104867869).

PR #5120, opened 2026-07-31
[bugfix] Disable Valgrind compilation on s390x

sharder996:
This PR disables the compilation of Valgrind on s390x. Valgrind is getting pulled in as a transitive dependency of glib and QEMU now that QEMU is being managed through vcpkg. Valgrind will not compile with newer versions of clang and its strictly enforcement of compiler warnings.

A successful compilation on an s390x runner can be found [here](https://github.com/canonical/multipass/actions/runs/32302539839/job/96228218459).

Closes #5154 

MULTI-2820

### DIFF
--- a/3rd-party/vcpkg-triplets/s390x-linux-release.cmake
+++ b/3rd-party/vcpkg-triplets/s390x-linux-release.cmake
@@ -10,5 +10,7 @@ set(VCPKG_CMAKE_CONFIGURE_OPTIONS "-DCMAKE_CROSSCOMPILING:BOOL=OFF")
 
 # Disable Valgrind which glib pulls in and compiles. Valgrind will not compile on s390x
 # with newer versions of clang.
-set(VCPKG_C_FLAGS "-DNVALGRIND")
-set(VCPKG_CXX_FLAGS "-DNVALGRIND")
+if(PORT STREQUAL "glib")
+    set(VCPKG_C_FLAGS "-DNVALGRIND")
+    set(VCPKG_CXX_FLAGS "-DNVALGRIND")
+endif()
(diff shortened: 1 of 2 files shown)


3rd-party/vcpkg-triplets/s390x-linux-release.cmake
```diff
@@ -7,3 +7,8 @@ set(VCPKG_BUILD_TYPE release)
 # Workaround for dbus and qtbase: when vcpkg sets VCPKG_CMAKE_SYSTEM_NAME, CMake may treat
 # native s390x builds as cross-compiling.
 set(VCPKG_CMAKE_CONFIGURE_OPTIONS "-DCMAKE_CROSSCOMPILING:BOOL=OFF")
+
+# Disable Valgrind which glib pulls in and compiles. Valgrind will not compile on s390x
+# with newer versions of clang.
+set(VCPKG_C_FLAGS "-DNVALGRIND")
+set(VCPKG_CXX_FLAGS "-DNVALGRIND")
```
xmkg: Can we gate this to glib port only?

```suggestion
if(PORT STREQUAL "glib")
    set(VCPKG_C_FLAGS "-DNVALGRIND")
    set(VCPKG_CXX_FLAGS "-DNVALGRIND")
endif()
```

xmkg (changes requested):
I have a small change request, looking good otherwise.

3rd-party/vcpkg-triplets/s390x-linux-release.cmake
sharder996: What if another dependency starts pulling in Valgrind. We would have to turn it off again for that dependency...

xmkg: But that dependency would have its own disable flag, no? AFAICT, the Valgrind def is a glib thing

sharder996: Oh right, I think I remember it being a unique flag that glib exclusively uses.

xmkg (approved):
LGTM, feel free to merge it after the s390x run succeeds :)

sharder996:
@xmkg Here's a successful build on s390x with vcpkg binary caching disabled just to make sure everything builds correctly.
https://github.com/canonical/multipass/actions/runs/32517659938/job/96882916632?pr=5154

PR #5128, opened 2026-08-07
[az] Persist was_running across daemon restarts

### LINKED ISSUE #5087: [az] VM state is not remembered across zone disable/enable
As expected, when one disables a zone the VMs in that zone and turned off. Enabling the zone again will return the VMs to the running state. However, if one restarts the Multipass daemon in between these two operations, the daemon will "forget" the original state of the VMs in the disabled zone:

```
scott@Scotts-MacBook-Pro Downloads % multipass disable-zones zone2
This operation will forcefully stop the VMs in zone2. Are you sure you want to continue? (Yes/no): y
Zone disabled: zone2
foo2                    Stopped           --               Ubuntu 26.04 LTS    zone2(n/a)
scott@Scotts-MacBook-Pro Downloads % sudo launchctl kickstart -k system/com.canonical.multipassd
foo2                    Stopped           --               Ubuntu 26.04 LTS    zone2(n/a)
scott@Scotts-MacBook-Pro Downloads % multipass enable-zones zone2                               
Zone enabled: zone2
foo2                    Stopped           --               Ubuntu 26.04 LTS    zone2
```

joelle-a-dev:
## Summary

`was_running` (whether a VM should be restarted when its zone is re-enabled) lived only as an in-memory flag on `BaseVirtualMachine`, never part of `VMSpecs`/the instance-db JSON. Separately, backend VM constructors decide their initial in-memory `state` from local heuristics (e.g. QEMU checks for a suspend snapshot on disk) with no awareness of zones at all, so a VM whose zone was disabled always came back as `off`/`suspended` on reconstruction instead of `unavailable`.

Together this is why `enable-zones` stops restarting instances if `multipassd` restarts while their zone is disabled — exactly the repro in #5087.

- Promote `was_running` to a public field on `VirtualMachine`, mirroring how `state` already works, so the daemon can read/write it directly without new virtual interfaces or backend-specific plumbing
- Add `was_running` to `VMSpecs` and its JSON (de)serialization (defaults to `false` for existing instance records, so old on-disk data stays loadable)
- `Daemon::persist_state_for` now also syncs the live `was_running` value into the persisted spec whenever a VM's state changes
- At daemon startup, when a persisted spec has `state == unavailable`, restore both `state` and `was_running` directly on the reconstructed VM (parallel to the existing "was running" reconciliation branch)

## Test plan

- [x] Added `Daemon.restoresUnavailableStateAndWasRunningOnConstruction` in `tests/unit/test_daemon.cpp`, which reproduces the reported scenario end-to-end: plants a persisted instance record with `state: unavailable, was_running: true`, constructs a fresh `Daemon` over it (simulating a restart), and asserts the reconstructed VM has both fields restored
- [x] Extended `fake_vm_properties`/`fake_json_contents` in the shared daemon test fixture with a `was_running` field to support the above
- [x] Traced every changed call site by hand for compile correctness (field visibility, `lookup_or<bool>` usage matching existing fields, other `VMSpecs{...}` aggregate-init call sites in `tests/unit/qemu/test_qemu_snapshot.cpp` — confirmed the new trailing field with a default member initializer doesn't break them under `-Werror -Wall -pedantic`)
- [ ] **Not compiled or run** — this environment doesn't have the Qt/gRPC/protobuf/Boost toolchain this project pulls via vcpkg, and building it from scratch here would take hours with no cached artifacts. Would appreciate a maintainer/CI build check before merge.

Fixes #5087

### DIFF
--- a/include/multipass/virtual_machine.h
+++ b/include/multipass/virtual_machine.h
@@ -136,6 +136,11 @@ public:
     std::condition_variable state_wait;
     std::mutex state_mutex;
 
+    // Whether the VM was running when its zone was last disabled; used to decide whether to
+    // start it back up when the zone becomes available again. Public (like `state`) so it can be
+    // persisted and restored by the daemon across restarts.
+    bool was_running = false;
+
 protected:
     explicit VirtualMachine(State state = State::off) : state{state}
     {
--- a/include/multipass/vm_specs.h
+++ b/include/multipass/vm_specs.h
@@ -46,6 +46,7 @@ struct VMSpecs
     int clone_count =
         0; // tracks the number of cloned vm from this source vm (regardless of deletes)
     std::string zone;
+    bool was_running = false; // whether the VM was running when its zone was last disabled
 
     friend inline bool operator==(const VMSpecs& a, const VMSpecs& b) = default;
 };
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1426,6 +1426,16 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
             break;
             }
         }
+        else if (spec.state == e_state::unavailable)
+        {
+            // Backends decide the VM's initial in-memory state from local heuristics (e.g. QEMU
+            // looks for a suspend snapshot on disk) and know nothing about availability zones, so
+            // a VM whose zone was disabled always comes back as `off`/`suspended` here. Restore
+            // the persisted `unavailable` state and whether it was running, so that a later
+            // enable-zones still knows to start it back up.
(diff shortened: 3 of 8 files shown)


jimporter:
Thanks for your interest in working on Multipass. However, I was already assigned to work on this issue and have some work in progress to do this in a different way. (In particular, I'm a little concerned that in this PR, we're adding a new field to the instance state that's only valid in some states.)

We'd be happy to help get future contributions of yours merged/reviewed into Multipass, though it would be best to find an unassigned issue so that we're not duplicating effort. I also recommend building/testing locally, since after a brief look at this PR, I don't think it actually does anything with this new `was_running` field.

joelle-a-dev:
Thank you @jimporter, very well noted.

PR #5135, opened 2026-08-15
[GUI] Bring window to front on tray Quit even when only obscured, not minimized

### LINKED ISSUE #4210: [GUI] Quit appears to fail when called from the tray and set to ask
**Describe the bug**
On macOS, when Multipass is configured to ask about running instances on quit (the default), clicking "Quit" on the tray icon appears to have no effect. The question is asked in the Multipass window, but won't be visible unless the Window is unobstructed or minimized.

**To Reproduce**
1. Open the GUI
2. Go to settings
3. Set "When closing Multipass" to "Ask about running instances"
4. Launch or start an instance and wait for it to get to running state
5. Bring a larger window into focus (on the same screen) such that the GUI is no longer visible
6. Click the Multipass tray icon and select "Quit"
7. Observe no apparent reaction

Interestingly, this does not occur if the GUI is minimized: it is restored and brought into focus then.

**Expected behavior**
I expected the GUI to always come into focus when a question is asked.

**Logs**
Please provide logs from the daemon, see [accessing logs](https://canonical.com/multipass/docs/accessing-logs) on where to find them on your platform.

**Additional info**
- OS: macOS 15.3.2
- `multipass version` 1.16.0-rc5.3814+ga3a71f27.mac

xhon-pelushi:
## What this does

With "Ask about running instances" enabled, clicking tray Quit while the Multipass GUI window is obscured by another window (visible, not minimized, but not focused) never brought the confirmation dialog to front — so it looked like nothing happened.

`onWindowClose()` only called `windowManager.showAndRestore()` when the window was invisible or minimized, missing the "obscured but visible" case. This adds `!await windowManager.isFocused()` to that condition, so the window is restored/focused whenever it isn't already, not just when hidden or minimized.

## Testing

- `flutter analyze lib/main.dart`: same 5 pre-existing issues before and after (all pre-existing generated-protobuf-type issues, confirmed via `git stash`/`git stash pop`) — zero new issues introduced.
- Could not manually verify on macOS in this environment (no macOS device available); the fix relies on `window_manager`'s documented `isFocused()`/`show()` API and directly extends the existing, already-merged restore pattern from #4068/#4138.

Fixes #4210

### DIFF
--- a/src/client/gui/lib/main.dart
+++ b/src/client/gui/lib/main.dart
@@ -184,7 +184,9 @@ class _AppState extends ConsumerState<App> with WindowListener {
     }
 
     // checking the need to restore the window
-    if (!await windowManager.isVisible() || await windowManager.isMinimized()) {
+    if (!await windowManager.isVisible() ||
+        await windowManager.isMinimized() ||
+        !await windowManager.isFocused()) {
       windowManager.showAndRestore();
     }
 


sharder996 (approved):
Tested and verified on macOS. Thanks for the fix, @xhon-pelushi!

PR #5140, opened 2026-08-19
[bugfix] AZ state transitions

### LINKED ISSUE #5089: [az] `enable-zones` returns before VMs are running/reachable
After enabling a zone, the VMs in that zone will be brought up. However, the command will return before those VMs are actually reachable via SSH which will result in other commands breaking in the meantime.

```
scott@Scotts-MacBook-Pro Downloads % multipass enable-zones zone1
Zone enabled: zone1
list failed: IP not available

... # After some time

bar                     Running           192.168.252.4    Ubuntu 26.04 LTS    zone1
foo                     Running           192.168.252.3    Ubuntu 26.04 LTS    zone1
```

sharder996:
This PR fixes several issues with AZs that were caused by state transition being done by the AZ manager and then not being reflected in the daemon.

The fix proposed here is to have the AZ manager return a list of VMs that are being returned to their original `running` state after the AZ has been re-enabled. The daemon can then track these VMs and update their state as SSH comes up.

So, on top of the linked GitHub issue, this PR also fixes several others issues with AZ state transitions.

1. SSHFS mounts were not being stopped when a VM was forced off due to an AZ being disabled. This resulted in orphaned SSHFS processes. We now explicitly stop the mounts.
2. SSHFS mounts were not being restarted for VMs that are in an AZ that was enabled. Having the daemon track the starting VMs now also automatically starts the mounts.

In conclusion, this is another difficulty born out of the monolithic daemon and not having VM objects be responsible for their own resources, namely VM mounts. Due to the time sensitive nature of releasing AZs, the presented diff is favoured over a larger refactor.

Fixes #5089 

---

MULTI-2790

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2916,19 +2916,34 @@ void mp::Daemon::zones_state(const ZonesStateRequest* request,
 try // clang-format on
 {
     auto& az_manager = *config->az_manager;
+
+    std::vector<std::string> transitioned;
+    auto apply_to = [&](const std::string& zone_name) {
+        auto names = az_manager.get_zone(zone_name).set_available(request->available());
+        transitioned.insert(transitioned.end(),
+                            std::make_move_iterator(names.begin()),
+                            std::make_move_iterator(names.end()));
+    };
+
     if (request->zones().empty())
     {
         for (auto&& zone : az_manager.get_zones())
         {
-            az_manager.get_zone(zone.get().get_name()).set_available(request->available());
+            apply_to(zone.get().get_name());
         }
     }
     else
     {
         for (const auto& zone_name : request->zones())
         {
-            az_manager.get_zone(zone_name).set_available(request->available());
+            apply_to(zone_name);
         }
+    }
+
+    for (const auto& name : transitioned)
+    {
+        if (request->available())
+            on_restart(name);
     }
 
     context->set_value(grpc::Status{});
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2974,7 +2974,7 @@ void mp::Daemon::on_suspend()
 void mp::Daemon::on_restart(const std::string& name)
 {
     stop_mounts(name);
-    auto future_watcher = create_future_watcher([this, &name]() {
+    auto future_watcher = create_future_watcher([this, name]() {
         try
         {
             auto virtual_machine = operative_instances.at(name);
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2944,6 +2944,8 @@ try // clang-format on
     {
         if (request->available())
             on_restart(name);
+        else
+            stop_mounts(name);
     }
 
     context->set_value(grpc::Status{});
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -2917,12 +2917,13 @@ try // clang-format on
 {
     auto& az_manager = *config->az_manager;
 
-    std::vector<std::string> transitioned;
-    auto apply_to = [&](const std::string& zone_name) {
-        auto names = az_manager.get_zone(zone_name).set_available(request->available());
-        transitioned.insert(transitioned.end(),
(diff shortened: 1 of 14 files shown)


src/daemon/daemon.cpp
```diff
@@ -2916,21 +2916,38 @@ void mp::Daemon::zones_state(const ZonesStateRequest* request,
...
         {
-            az_manager.get_zone(zone_name).set_available(request->available());
+            apply_to(zone_name);
         }
     }
 
+    for (const auto& name : transitioned)
+    {
+        if (request->available())
+            on_restart(name);
```
tobe2098: `on_restart` here is redundant, it calls `stop_mounts`, which for an unavailable instance should have already been called. I would propose moving the parts of `on_restart` that enable the mounts to a `start_mounts` function, and then call it here as well. WDYT?

src/daemon/daemon.cpp
```diff
@@ -2916,21 +2916,38 @@ void mp::Daemon::zones_state(const ZonesStateRequest* request,
 try // clang-format on
 {
     auto& az_manager = *config->az_manager;
+
+    std::vector<std::string> transitioned;
+    auto apply_to = [&](const std::string& zone_name) {
```
tobe2098: Could we avoid capturing all variables? `&transitioned, &az_manager, available=request->available()` should be enough, WDYT?

src/daemon/daemon.cpp
```diff
@@ -2916,21 +2916,38 @@ void mp::Daemon::zones_state(const ZonesStateRequest* request,
 try // clang-format on
 {
     auto& az_manager = *config->az_manager;
+
+    std::vector<std::string> transitioned;
```
tobe2098: From the code the intent of the `set_available` return bool seems to be "requires mount activation/deactivation" rather than "transitioned", since if `!was_running`, the function returns false for that VM. Would you mind changing the naming? I think it can be a bit misleading when reading the abstractions.

tests/unit/test_base_availability_zone.cpp
```diff
@@ -128,17 +128,17 @@ TEST_F(BaseAvailabilityZoneTest, AvailabilityStateManagement)
...
     zone.add_vm(mock_vm1);
     zone.add_vm(mock_vm2);
 
     // Setting to current state (true) shouldn't trigger VM updates
-    zone.set_available(true);
+    EXPECT_THAT(zone.set_available(true), IsEmpty());
 
-    // Setting to new state should notify all VMs
-    zone.set_available(false);
+    // Setting to new state should notify all VMs, reporting only those that transitioned
```
tobe2098: idem with the abstraction comment

tobe2098 (changes requested):
Great work Scott! Just some nitpicks, but the overall logic solves the issues.

src/daemon/daemon.cpp
sharder996: Sure, I'm fine with this for now. In the future, I'd like to do a refactor of how we handle mounts to hopefully make this a lot cleaner.

src/daemon/daemon.cpp
```diff
@@ -2916,21 +2916,39 @@ void mp::Daemon::zones_state(const ZonesStateRequest* request,
...
 {
     auto& az_manager = *config->az_manager;
+
+    std::vector<std::string> restart_mounts;
+    auto apply_to = [&restart_mounts, &az_manager, available = request->available()](
+                        const std::string& zone_name) {
+        auto names = az_manager.get_zone(zone_name).set_available(available);
+        restart_mounts.insert(restart_mounts.end(),
+                              std::make_move_iterator(names.begin()),
+                              std::make_move_iterator(names.end()));
```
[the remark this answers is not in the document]
tobe2098: This is a valid point. Ideally the mounts should be addressed for transitioned instances. When `BaseAvailabilityZone::set_available` throws, the transitioned instances are transitioned back to available status. Those transitioned instances would have unhandled mounts in this case.

Additionally, in this function, when an exception is thrown `restart_mounts` is discarded, leaving further mounts unhandled.

tobe2098 (changes requested):

PR #5145, opened 2026-08-19
[packaging] Remove old GUI app before installing

sharder996:
This PR adds a preinstall script that makes sure the old GUI app is gone before installing the new one. Some systems (this causes issue on macOS) don't want to remove the old symlink when the app version isn't strictly increasing. This can cause issues when trying to install CI packages.

### Testing steps:

Install a package from a PR that has a higher PR number (results in a higher version number) from this PR. Install a package from another PR that has a lower number. Open spotlight and try to launch the GUI. You should get an error from the system:

> You can’t open the application “Multipass” because it may be damaged or incomplete.

Install a package from a PR that has a higher PR number (results in a higher version number) from this PR. Install the package from this PR. Open spotlight and GUI launches as expected.


tobe2098 (approved):
LGTM Scott! Good catch

tobe2098:
Tested and reproduced

PR #5147, opened 2026-08-20
[applevz] Short circuit state check when zone unavailable

### LINKED ISSUE #5088: [az] Inconsistent VM state after `disable-zones`
If a backend driver does not have the suspend VM operation implemented, VMs are turned off and displayed as being `Stopped` instead of `Unavailable`.

This only really affects the `applevz` driver AFAIK, so the priority should be reflected as such.

```
bar                     Stopped           --               Ubuntu 26.04 LTS    zone1(n/a)
foo                     Stopped           --               Ubuntu 26.04 LTS    zone1(n/a)
scott@Scotts-MacBook-Pro Downloads % multipass get local.driver
applevz
...
bar                     Unavailable       --               Ubuntu 26.04 LTS    zone1(n/a)
foo                     Unavailable       --               Ubuntu 26.04 LTS    zone1(n/a)
scott@Scotts-MacBook-Pro Downloads % multipass get local.driver
qemu
```

sharder996:
This PR makes `current_state` short circuit when the zone is disabled. This prevents the daemon from reporting VMs in a disabled zones from reporting a `Stopped` state in the terminal output.

Fixes #5088 

---

MULTI-2791

### DIFF
--- a/src/platform/backends/applevz/applevz_virtual_machine.cpp
+++ b/src/platform/backends/applevz/applevz_virtual_machine.cpp
@@ -248,9 +248,17 @@ void AppleVZVirtualMachine::suspend()
 
 VirtualMachine::State AppleVZVirtualMachine::current_state()
 {
-    // Get state from AppleVZ, translate it to our state enum, and notify the monitor
+    if (state == State::unavailable)
+    {
+        mpl::debug(log_category, "current_state() -> VM `{}` zone is unavailable", vm_name);
+        assert(!vm_handle);
+        return state;
+    }
+
     if (!vm_handle)
         return State::stopped;
+
+    // Get state from AppleVZ, translate it to our state enum, and notify the monitor
     set_state(MP_APPLEVZ.get_state(vm_handle));
     return state;
 }
--- a/src/platform/backends/applevz/applevz_virtual_machine.cpp
+++ b/src/platform/backends/applevz/applevz_virtual_machine.cpp
@@ -248,16 +248,6 @@ void AppleVZVirtualMachine::suspend()
 
 VirtualMachine::State AppleVZVirtualMachine::current_state()
 {
-    if (state == State::unavailable)
-    {
-        mpl::debug(log_category, "current_state() -> VM `{}` zone is unavailable", vm_name);
-        assert(!vm_handle);
-        return state;
-    }
-
-    if (!vm_handle)
-        return State::stopped;
-
     // Get state from AppleVZ, translate it to our state enum, and notify the monitor
     set_state(MP_APPLEVZ.get_state(vm_handle));
     return state;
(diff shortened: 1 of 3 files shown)


src/platform/backends/applevz/applevz_virtual_machine.cpp
```diff
@@ -248,9 +248,17 @@ void AppleVZVirtualMachine::suspend()
 
 VirtualMachine::State AppleVZVirtualMachine::current_state()
 {
-    // Get state from AppleVZ, translate it to our state enum, and notify the monitor
+    if (state == State::unavailable)
+    {
+        mpl::debug(log_category, "current_state() -> VM `{}` zone is unavailable", vm_name);
+        assert(!vm_handle);
+        return state;
+    }
```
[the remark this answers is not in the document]
sharder996: Because setting the state to `unavailable` will kill the VM process and drop the handle.

jimporter: I did this a bit differently in the Hyper-V backend; I put this check in `set_state` instead of `current_state`. I'm not sure it will make a behavioral difference either way (given the current code), but I think it's worth being consistent here, since AppleVZ and the Hyper-V API are fairly similar to each other in my opinion.

My reasoning for putting this in `set_state` was that whenever `state == unavailable`, we want to ensure that we don't consult the platform APIs to determine the state; those APIs don't know what "unavailable" is. Instead, our code needs to set `state` to a different value directly (e.g. `state = starting;`) before we start pulling the state from the platform APIs.

On the other hand, this is a bit less obvious than having the check in `current_state` like you've done. The risk I see is that future changes to this code could call `set_state` when the instance is unavailable, thus overwriting `state`.

It's possible I wrote the Hyper-V in an overly-defensive way, or it's just defensive enough. :) What do you think?

jimporter (changes requested):
I don't have a macOS system to test this on, but I believe it should work given the surrounding code. I just have one comment below about where we do the `state == unavailable` check.

(Note that I marked this as "request changes", but depending on what we decide, those changes might happen in the Hyper-V backend, possibly in a different PR.)

src/platform/backends/applevz/applevz_virtual_machine.cpp
sharder996: Ok, I made it mirror the HCSVirtualMachine class more closely. The reason I had done it this way was because of the null handle check in `current_state()` setting the state to stopped. I had made an effort to check logic out of the ObjC++ side, but since this is a relatively small piece of logic I'm fine with moving it.

jimporter (approved):
This looks good to me.

One area of potential future improvement for both this and the Hyper-V API backend is that we could change the `set_state` methods in each backend to be something like `sync_state`, which would take zero arguments and fetch the state from the platform APIs when appropriate. That might make the logic clearer, but we don't have to do it right now. Maybe I'll look into something like that when I have a bit of downtime.

PR #5150, opened 2026-08-21
[utils] Runtime-time unreachability check

tobe2098:
# Description

- What does this PR do? Adds a compile-time guarantee for certain sections of the code that should not be reached
- Why is this change needed? We are currently relying on `assert(false)`, which is necessary for run-time checks, but having a compile-time option will also allow us to catch changes that would violate the assumption. It also applies to release builds, which is another advantage.

## Testing

- Compile the project

## Checklist

### DIFF
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -24,7 +24,6 @@
 #include <multipass/virtual_machine.h>
 
 #include <fmt/base.h>
-#include <fmt/format.h>
 
 #include <yaml-cpp/yaml.h>
 
@@ -33,8 +32,10 @@
 #include <filesystem>
 #include <functional>
 #include <future>
+#include <ranges>
 #include <source_location>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include <QDir>
@@ -56,11 +57,13 @@ namespace utils
 {
 
 // Marks code that should never be reached: logs the location and aborts.
-[[noreturn]] inline void UNREACHABLE(std::source_location loc = std::source_location::current())
+[[noreturn]] inline void UNREACHABLE(std::string_view message,
+                                     std::source_location loc = std::source_location::current())
 {
     logging::log_location(logging::Level::error,
                           logging::detail::with_source_location<std::string_view>("FATAL", loc),
-                          "Reached code marked as unreachable");
+                          "Reached unreachable code: {}",
+                          message);
     std::abort();
 }
 
--- a/include/multipass/utils.h
+++ b/include/multipass/utils.h
@@ -17,21 +17,23 @@
 
 #pragma once
 
-#include <multipass/logging/level.h>
+#include <multipass/logging/log_location.h>
 #include <multipass/network_interface_info.h>
 #include <multipass/path.h>
 #include <multipass/singleton.h>
 #include <multipass/virtual_machine.h>
 
 #include <fmt/base.h>
+#include <fmt/format.h>
 
 #include <yaml-cpp/yaml.h>
 
 #include <chrono>
+#include <cstdlib>
 #include <filesystem>
 #include <functional>
 #include <future>
-#include <ranges>
+#include <source_location>
 #include <string>
 #include <vector>
 
@@ -52,6 +54,15 @@ class SSHSession;
 
 namespace utils
 {
+
+// Marks code that should never be reached: logs the location and aborts.
+[[noreturn]] inline void UNREACHABLE(std::source_location loc = std::source_location::current())
+{
+    logging::log_location(logging::Level::error,
+                          logging::detail::with_source_location<std::string_view>("FATAL", loc),
+                          "Reached code marked as unreachable");
+    std::abort();
+}
 
 // enum types
 enum class QuoteType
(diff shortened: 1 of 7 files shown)


tobe2098:
Requesting to @sharder996 for testing compilation in ppc64 and s390x

ricab (changes requested):
Hi Antoni, I contemplated the same thing when implementing this code and ended up deciding against it. Here's why:

- Formally, this breaks ODR: `UNREACHABLE` is odr-used at the call side, despite never being defined. The standard requires exactly one definition in the program for every odr‑used function and violating it makes the program ill‑formed (no diagnostic required). From [basic.def.odr]/10:

> Every program shall contain exactly one definition of every non-inline function or variable that is odr-used in that program outside of a discarded statement (8.5.1); no diagnostic required [...]

- In practice, I am also not (yet) convinced that we could rely on this to break builds *if and only if* the code was reachable. I don't think the compiler can always prove whether an instruction is reachable, which is probably why the language does not offer any such facility. `std::unreachable` in C++23 instead invokes undefined behavior). If this was possible, you would have solved the halting problem.

Claude tells me:

> Reachability is undecidable in general; DCE is a heuristic. As soon as the "impossible" condition depends on something the optimizer can't fold (a value from another TU, a virtual call, a loop the compiler can't reason about), the call stays even though it can never execute at runtime. In the diff, the fall‑off‑the‑end after a loop/switch is exactly the kind of thing a compiler often will not prove.

Furthermore, this would rely on specific compiler/linker configurations, making it fragile.

I am of course open to being convinced that this works but, given the halting problem and the standard's own std::unreachable, I would call that an extraordinary claim and so request extraordinary evidence. 

If you are uneasy with the assert, I wouldn't mind throwing a logic_error or calling terminate directly.

tobe2098:
@ricab, that is correct. `UNREACHABLE` cannot be used as a drop-in replace in some places like switch statements, exactly because in some cases the compiler cannot know if that place is every reached. The idea was to use it only in the places where the compiler knows it cannot be reached (e.g., a loop without a `break`). 

Aside from that, breaking ODR does make it undesirable. The best option seems to me in this case to be throwing a logic error.

include/multipass/utils.h
```diff
@@ -53,6 +55,15 @@ class SSHSession;
 namespace utils
 {
 
+// Marks code that should never be reached: logs the location and aborts.
+[[noreturn]] inline void UNREACHABLE(std::source_location loc = std::source_location::current())
```
ricab: I am wondering if we should call this `panic()`... up to you, I am fine with it either way.

src/client/cli/formatter/json_formatter.cpp
```diff
@@ -241,7 +241,8 @@ std::string mp::JsonFormatter::format(const InfoReply& reply) const
         }
         else
         {
-            assert(false && "either one of instance or snapshot details should be populated");
+            // either one of instance or snapshot details should be populated
+            mp::utils::UNREACHABLE();
```
ricab: This loses the information in the assert message. Why don't you add a param in UNREACHABLE and write it in the log?

ricab (commented):
Ah much better, this I am fine with :)

A couple of suggestions inline. The name I am not sure about, but the message I think would be a nice addition.

include/multipass/utils.h
tobe2098: Although `panic()` captures the concept of the abort, I think unreachable() represents the intention better. I would agree with a `panic` function that just aborts, and can thus be used anytime you want to induce an abort.

include/multipass/utils.h
```diff
@@ -53,6 +56,17 @@ class SSHSession;
...
 {
 
+// Marks code that should never be reached: logs the location and aborts.
+[[noreturn]] inline void UNREACHABLE(const std::string& message,
+                                     std::source_location loc = std::source_location::current())
+{
+    logging::log_location(logging::Level::error,
+                          logging::detail::with_source_location<std::string_view>("FATAL", loc),
+                          "Reached unreachable code: {}",
+                          message);
```
ricab, line 66: Hmm, this is no longer using the new error_location, yet it is still introduced in this PR. Could you please clarify?

ricab (commented):
Question inline.

include/multipass/utils.h
tobe2098: That was my bad. The reason to not use it is because it would capture the location in UNREACHABLE(), which is not what we want. I can remove the error_location implementation if you prefer, since it is not currently used.

sharder996 (commented):
@tobe2098 Running the other arches in #5159. Feel free to rebase as needed. Let me know if there's anything else you require of me 😄 

P.S. I should look into seeing if we can run the ppc64el/s390x runners on demand...

tobe2098:
Thanks @sharder996

include/multipass/utils.h
ricab: OK, got it.

I think it would be better to remove it since it is obvious to add once it is needed and that way we'd avoid it somehow ageing poorly without being exercised. Not a blocking preference though. If you choose to keep it, I would request just a unit test covering it please.

ricab (commented):
I am good with this overall, just the matter of error_location to resolve one way or another.

include/multipass/utils.h
tobe2098: Removed :+1:

ricab (approved):
LGTM, I'll start using it!

PR #5160, opened 2026-08-25
[hyperv] Hyper-V to Host Compute System Migration

xmkg:
The PR implements the necessary code to perform migration from the existing Hyper-V implementation to the Host Compute System.

The migration approach is by explicit request, triggered by `multipass set local.driver=hyperv_api` for existing installations. This triggers the migration procedure. Only stopped VMs will be eligible for the migration.

The migrated VMs will be kept in the old backend. The user needs to explicitly delete them after confirming that the migrated VMs are working as intended. The user is able to switch back and forth between the hyperv and hyperv_api drivers. Each hyperv -> hyperv_api switch will trigger a migration attempt for the eligible VMs, while skipping the already migrated ones naturally due to name collision.

The installer now offers the Host Compute System as a backend option, and it's the default for new installations. Users can still select the other available backends.

[Flowchart](https://github.com/user-attachments/files/32064183/Flowchart.pdf)

MULTI-2903


packaging/windows/wix/ui/CustomWixUI_InstallDir.wxs
```diff
@@ -59,10 +59,10 @@
                 Value="CustomWelcomeDlg" />
             <Publish Dialog="LicenseAgreementDlg" Control="Next" Event="NewDialog"
                 Value="ChooseHypervisorDlg"
-                Condition="LicenseAccepted = &quot;1&quot; AND (NOT WINDOWS_EDITION &gt;&lt; &quot;Core&quot; AND NOT WIX_UPGRADE_DETECTED)" />
```
xmkg, line 62: To show backend options in the home edition.

packaging/windows/wix/Package.wxs
```diff
@@ -3,7 +3,7 @@
     <Package Name="Multipass" Manufacturer="Canonical Ltd" Version="$(Version)"
         UpgradeCode="faebc626-1491-49bb-8162-22b1684013d9">
         <MajorUpgrade AllowDowngrades="yes" />
-        <Launch Condition="Installed OR WIN_VERSION >= 17134" Message="!(loc.FeatureUpdateReqd)" />
+        <Launch Condition="Installed OR WIN_VERSION >= 17763" Message="!(loc.FeatureUpdateReqd)" />
```
xmkg, line 6: Build 17763 is Windows 10 Version 1809, which is the minimum version that supports HCS.

packaging/windows/wix/ui/CustomWixUI_InstallDir.wxs
sharder996: This is for showing the License Agreement dialog. Why does showing that depend on the edition of Windows?

sharder996: Nevermind. I understand now.

packaging/windows/wix/ui/CustomWixUI_InstallDir.wxs
```diff
@@ -71,10 +71,10 @@
 
             <Publish Dialog="ChoosePathInstallDlg" Control="Back" Event="NewDialog"
                 Value="LicenseAgreementDlg"
-                Condition="WINDOWS_EDITION &gt;&lt; &quot;Core&quot; OR WIX_UPGRADE_DETECTED" />
+                Condition="WIX_UPGRADE_DETECTED" />
```
sharder996, line 74: I guess the custom path selection only makes a difference when using HCS and not VirtualBox.

packaging/windows/wix/Package.wxs
```diff
@@ -89,8 +91,17 @@
...
+        <!--
+          The new Hyper-V/HCS driver (hyperv_api, backed by VirtualMachinePlatform) is the default
+          on every supported edition. The legacy full Hyper-V driver (hyperv, backed by
+          Microsoft-Hyper-V) is only offered on editions that support it: Home ("Core") editions do
+          not, so the legacy option is hidden there. VirtualBox remains selectable everywhere.
+        -->
+        <Property Id="LEGACY_HYPERV_AVAILABLE" Secure="yes" />
+        <SetProperty Id="LEGACY_HYPERV_AVAILABLE" Value="1"
+            Condition="WINDOWS_EDITION AND NOT (WINDOWS_EDITION &gt;&lt; &quot;Core&quot;)"
+            After="AppSearch" Sequence="both" />
```
sharder996, line 104: `LEGACY_HYPERV_AVAILABLE` is only used in the installer UI, so `Sequence="ui"` is sufficient, right?

packaging/windows/wix/Package.wxs
```diff
@@ -89,8 +91,17 @@
...
-        <SetProperty Id="DRIVER" Value="virtualbox" Condition="WINDOWS_EDITION &gt;&lt; &quot;Core&quot;"
-            After="AppSearch" />
+
+        <!--
+          The new Hyper-V/HCS driver (hyperv_api, backed by VirtualMachinePlatform) is the default
+          on every supported edition. The legacy full Hyper-V driver (hyperv, backed by
+          Microsoft-Hyper-V) is only offered on editions that support it: Home ("Core") editions do
+          not, so the legacy option is hidden there. VirtualBox remains selectable everywhere.
+        -->
+        <Property Id="LEGACY_HYPERV_AVAILABLE" Secure="yes" />
```
sharder996, line 101: `LEGACY_HYPERV_AVAILABLE` doesn't get used on the server side of the installation, so we can set `Secure` to `no`, right?

packaging/windows/wix/Package.wxs
```diff
@@ -17,7 +17,9 @@
                 Key='SOFTWARE\Microsoft\Windows NT\CurrentVersion' Name='CurrentBuildNumber' />
         </Property>
 
-        <Property Id="DRIVER" Value="hyperv" Secure="yes" />
+        <Property Id="DRIVER" Value="hyperv_api" Secure="yes" />
+        <Property Id="DRIVER_FULL" Value="hyperv_api" Secure="yes" />
+        <Property Id="DRIVER_LIMITED" Value="hyperv_api" Secure="yes" />
```
sharder996, line 22: These properties are only used in the UI phase of the installation process, so they don't need to be set as secure.

src/client/cli/cmd/animated_spinner.cpp
```diff
@@ -61,14 +61,14 @@ void mp::AnimatedSpinner::start()
...
-    if (running)
-    {
-        running = false;
-        cv.notify_one();
-        lock.unlock();
-        if (t.joinable())
-            t.join();
-    }
+    if (!running)
+        return;
```
sharder996, line 65: `clear_line` no longer gets called when `!running`. Does this make any difference in terminal output?

src/rpc/multipass.proto
```diff
@@ -489,6 +489,8 @@ message SetReply {
     string log_line = 1;
     string reply_message = 2;
     bool needs_authorization = 3;
+    string summary = 4;
+    string migration_phase = 5;
```
sharder996, line 493: Can we make a new message type to nest in here so that these new fields are tied together? Also, then the new message type can be named according to its purpose, i.e `HCSMigrationReport`.

src/client/cli/cmd/remote_settings_handler.cpp
```diff
@@ -127,14 +130,50 @@ class RemoteSet : public RemoteSettingsCmd
...
+                grpc::ClientReaderWriterInterface<mp::SetRequest, mp::SetReply>* client) {
+                if (!reply.log_line().empty())
+                    spinner.print(term.cerr(), reply.log_line());
+
+                if (key.startsWith(mp::daemon_settings_root) &&
+                    key.endsWith(mp::bridged_network_name) && reply.needs_authorization())
+                {
+                    spinner.stop();
+                    mp::BridgePrompter prompter{&term};
+                    const std::vector<std::string> networks{reply.reply_message()};
```
sharder996:
```suggestion
                    const std::vector<std::string> networks{1, reply.reply_message()};
```

To match the original declaration.

src/daemon/daemon.cpp
```diff
@@ -1272,8 +1243,82 @@ void populate_snapshot_info(mp::VirtualMachine& vm,
...
+                                 grpc::ServerReaderWriterInterface<Reply, Request>*,
+                                 DaemonRpcContext*)) {
+            QObject::connect(&rpc,
+                             signal,
+                             this,
+                             [this, slot](const Request* request,
+                                          grpc::ServerReaderWriter<Reply, Request>* server,
+                                          DaemonRpcContext* context) {
+                                 if (!allowed_during_migration<Request> &&
+                                     reject_if_migrating("perform this operation", context))
```
sharder996: Do the rest connect afterwards?

src/daemon/daemon.cpp
```diff
@@ -2526,7 +2618,11 @@ try
     mpu::send_messages(server, messages);
     mpl::debug(category, "Succeeded setting {}={}", key, val);
 
+#if defined(HYPERV_HCS_ENABLED)
+    context->set_value(transition.complete(server));
```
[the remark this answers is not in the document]
xmkg: This is the case for any kind of blocking daemon operation.

src/daemon/daemon.cpp
```diff
@@ -2505,8 +2581,24 @@ try
 {
     auto key = request->key();
     auto val = request->val();
+    if (key == mp::driver_key)
+        val = mp::daemon::interpret_driver(QString::fromStdString(val)).toStdString();
```
sharder996, line 2606: `hyperv_api` does not have any aliases, so I'm not sure why this is necessary. `DriverTransition` then doesn't need it, and the `interpret_driver` function is already registered here: https://github.com/canonical/multipass/blob/ec646a10e7ca68ce3e9c61eca2c93bace74c3ed3/src/daemon/daemon_init_settings.cpp#L115-L117

Laefy:
I ran some tests (CLI only for now): https://docs.google.com/document/d/1TwdEhBravZcvyUCPIbZwyjl05ldy60PnpzEutNboRCs/edit?tab=t.0

For the summary:
- On the old backend, I encountered again the bug where the daemon is waiting on SSH forever. I had to kill the daemon and restarted it again
- I had an issue where the Running machines ended up in a Suspended state (happened after the upgrade, but also when I switched the driver from hyperv_api back to hyperv)
- The first go from hyperv -> hyperv_api went well
- However, after going back to the old driver to make some changes to the non migrated VMs, I started to notice weird behavior when switching again to the new driver:
  - Unable to really start the VMs (list says that they are running, but then I get ssh timeout errors when attempting to do real work)
  - One of the VM got assigned an 172.x.x.x IP address, eventhough it seems they should end up with a 10.x.x.x one
  - Killing and restarting the daemon did not work
  - Only rebooting Windows allowed me to get out of that state.

src/platform/platform_win.cpp
```diff
@@ -18,6 +18,7 @@
 #include <multipass/constants.h>
 #include <multipass/exceptions/formatted_exception_base.h>
 #include <multipass/exceptions/settings_exceptions.h>
+#include <multipass/file_ops.h>
```
sharder996, line 21: Extra leftover include?

tests/unit/hyperv_api/test_ut_hyperv_hcs_virtual_machine_factory.cpp
```diff
@@ -173,6 +183,57 @@ TEST_F(HyperVHCSVirtualMachineFactory_UnitTests,
     uut->remove_resources_for(vm_name);
 }
 
+TEST_F(HyperVHCSVirtualMachineFactory_UnitTests,
+       release_resources_removes_named_endpoint_without_compute_system)
```
sharder996, line 187: Inconsistent case for test naming. I think the project convention is to camel case which is observed in places, but in others it resorts to using snake case.

src/daemon/daemon.cpp
```diff
@@ -2505,8 +2581,24 @@ try
...
     std::string bridge_name;
 
+#if defined(HYPERV_HCS_ENABLED)
```
sharder996: Should we instead be using the backend specific macro `HYPERV_API_ENABLED` so that this doesn't start pulling in Windows code on other platforms once the feature flag is removed?

tests/unit/CMakeLists.txt
sharder996, line 1: idem Use backend specific macro instead of feature flag.

sharder996 (changes requested):
Releasing current review comments so we can being iterating.

src/daemon/daemon.cpp
xmkg: good idea.

src/platform/backends/hyperv_api/hcs_virtual_machine_factory.h
```diff
@@ -46,6 +47,11 @@ struct HCSVirtualMachineFactory final : public BaseVirtualMachineFactory
         return "hyperv_api";
     };
 
+    [[nodiscard]] QString get_backend_directory_name() const override
+    {
+        return "hyperv_api";
+    }
```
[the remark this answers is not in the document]
xmkg: There's no installation and no old HCS records.

src/daemon/daemon.cpp
xmkg: done

src/client/cli/cmd/remote_settings_handler.cpp
```diff
@@ -127,14 +130,50 @@ class RemoteSet : public RemoteSettingsCmd
         set_request.set_val(val.toStdString());
         set_request.set_authorized(user_authorized);
 
-        auto streaming_confirmation_callback =
-            mp::make_confirmation_callback<mp::SetRequest, mp::SetReply>(*term, key);
```
sharder996, line 131: The deletion of `make_confirmation_callback` here now leaves zero reference to the function and is essentially dead code.

src/daemon/hyperv_driver_transition.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+grpc::Status mhv::DriverTransition::prepare(const std::string& key, const std::string& value)
+{
+    if (key != mp::driver_key)
+        return grpc::Status::OK;
+
+    const auto current_driver = MP_SETTINGS.get(mp::driver_key).toStdString();
+    const auto migrate_hyperv = current_driver == "hyperv" && value == "hyperv_api";
+    if (!migrate_hyperv && !(current_driver == "hyperv_api" && value != "hyperv_api"))
```
sharder996: For readability sake, I think it would be helpful to put `!(current_driver == "hyperv_api" && value != "hyperv_api")` into a separate variable similar to `migrate_hyperv`. If I'm understanding correctly, the second clause is to prevent someone from switching away from `hyperv_api` while a migration is in progress.

src/daemon/hyperv_driver_transition.cpp
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    for (const auto& [name, spec] : context.specs)
+    {
+        if (!release_hcs_resources(name))
+            throw std::runtime_error{
+                fmt::format("Could not release hyperv_api resources for '{}'", name)};
+    }
+}
```
sharder996: For my own understanding. This function is called when switching away from `hyperv_api`, right? Checking that VMs are stopped is just a safeguard and they should have been stopped/suspended by another mechanism by this point, correct?

src/daemon/hyperv_migration.h
```diff
@@ -0,0 +1,153 @@
...
+    DaemonHyperVInstanceMigrator(const std::unordered_map<std::string, VMSpecs>& specs,
+                                 const InstanceTable& operative_instances,
+                                 const InstanceTable& deleted_instances,
+                                 VirtualMachineFactory& source_factory,
+                                 AvailabilityZoneManager& az_manager,
+                                 const Path& data_dir,
+                                 HyperVMigrationTargetRecords& target_records);
+    ~DaemonHyperVInstanceMigrator();
+
+    [[nodiscard]] InstanceMigrationResult migrate(const std::string& name,
```
sharder996, line 131: I think this one can be private.

src/platform/backends/hyperv_migration/hyperv_target_transaction.cpp
```diff
@@ -0,0 +1,397 @@
...
+    std::vector<fs::path> chain;
+    if (const auto result = VirtDisk().list_virtual_disk_chain(disk, chain, 2); !result)
+        throw std::runtime_error{
+            fmt::format("Could not inspect virtual disk chain for '{}': {}", disk, result)};
+    if (chain.empty())
+        throw std::runtime_error{fmt::format("Virtual disk chain for '{}' is empty", disk)};
+    if (chain.size() < 2)
+        return {};
+    return chain[1];
+}
```
sharder996: Thinking maybe these should these go in `utils.cpp`?

src/daemon/hyperv_migration.cpp
```diff
@@ -0,0 +1,569 @@
...
+        return "instance is unavailable";
+
+    try
+    {
+        {
+            const std::lock_guard lock{vm_it->second->state_mutex};
+            const auto state = vm_it->second->state;
+            if (state == VirtualMachine::State::starting ||
+                state == VirtualMachine::State::restarting)
+                return fmt::format("instance is {} and needs to be stopped",
```
sharder996: The below powershell command doesn't account for VMs that are preforming these operations?

src/platform/backends/hyperv_migration/hyperv_disk_layout.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (chain.empty())
+        throw std::runtime_error{fmt::format("Virtual disk chain for '{}' is empty", disk)};
+
+    return chain;
+}
+
+bool same_path(const fs::path& lhs, const fs::path& rhs)
+{
+    return MP_FILEOPS.weakly_canonical(lhs) == MP_FILEOPS.weakly_canonical(rhs);
+}
```
sharder996: Should this go in `utils.cpp`?

src/platform/backends/hyperv_migration/hyperv_target_transaction.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+std::string unique_target_name(const fs::path& source, std::unordered_set<std::string>& taken)
+{
+    auto base = source.filename().string();
+    if (taken.insert(base).second)
+        return base;
+
+    for (int suffix = 1;; ++suffix)
+    {
+        auto candidate = fmt::format("{}-{}{}",
```
sharder996, line 71: Should we maintain the same format as what snapshots are named in Multipass? i.e. `<name/"snapshot"><n>`

src/daemon/daemon.cpp
```diff
@@ -2547,7 +2639,11 @@ try
     mpu::send_messages(server, messages);
     mpl::debug(category, "Succeeded setting {}={}", key, val);
 
+#if defined(HYPERV_API_ENABLED)
+    context->set_value(transition.complete(server));
```
sharder996, line 2631: I think there is a discrepancy in expectations here. We write to the log that we successfully changed the setting regardless of whether or not the migration succeeded. But, then we set the rpc context to the result of the migration. So, should we be returning the result of changing the setting or the result of the migration?

packaging/windows/wix/Package.wxs
xmkg: Yes, changed to UI.

packaging/windows/wix/Package.wxs
xmkg: Done

packaging/windows/wix/Package.wxs
xmkg: Done

src/rpc/multipass.proto
xmkg: Done

src/client/cli/cmd/remote_settings_handler.cpp
xmkg: Done

src/platform/platform_win.cpp
xmkg: Done

tests/unit/hyperv_api/test_ut_hyperv_hcs_virtual_machine_factory.cpp
xmkg: Done

src/daemon/daemon.cpp
xmkg: DriverTransition::prepare() runs before the set() function, so it needs the driver normalized explicitly.

src/daemon/daemon.cpp
sharder996, line 2623: I think the existence of `migration_records` is how `complete()` determines if we're performing a migration or not. For non-driver setting changes it isn't immediately clear that a migration will be skipped. I would recommend either a comment saying that.

src/daemon/daemon.cpp
```diff
@@ -3133,7 +3229,10 @@ void mp::Daemon::create_vm(const CreateRequest* request,
 
     auto timeout = timeout_for(request->timeout());
 
-    preparing_instances.insert(name);
+    if (!begin_instance_preparation(name,
```
sharder996: Shouldn't this be one of the first, if not the first thing to check in this function? For example, why bother checking the zone name or permission for bridging if I can't even create the VM?

src/daemon/daemon.cpp
```diff
@@ -2822,15 +2918,16 @@ try
         if (auto dest_vm_status = validate_dest_name(destination_name); !dest_vm_status.ok())
             return context->set_value(std::move(dest_vm_status));
 
+        if (!begin_instance_preparation(destination_name, "clone an instance", context))
```
sharder996: Same idea as lower comment. Check earlier?

sharder996 (changes requested):
Was a little of churn as I was finishing up this review, so hopefully everything is still applicable. From a code point of view I can't see anything else outstanding and I'm relying on @Laefy's functional testing to make sure everything works.

src/platform/backends/hyperv_migration/hyperv_target_transaction.cpp
xmkg: Since this is throwaway code and the only consumer is this, I'd say YAGNI.

src/client/cli/cmd/animated_spinner.cpp
xmkg: It doesn't change what's visible on a terminal. If the spinner was running, stop() still clears the line. If it has already stopped, the line was cleared at that point, so clearing it again did nothing.

src/client/cli/cmd/remote_settings_handler.cpp
xmkg: Good catch. We need to restore make_confirmation_callback after we're done with the migration code. I've changed the code so it's now obvious what to do in that case.

src/daemon/hyperv_driver_transition.cpp
xmkg: Done

src/daemon/hyperv_migration.h
xmkg: The test code exercises this function, making it private would mean routing all the requests through migrate_all. Would it be okay to keep it like this?

src/daemon/hyperv_migration.cpp
xmkg: Good point. Removed it.

src/platform/backends/hyperv_migration/hyperv_disk_layout.cpp
xmkg: Same as the other util suggestion.

xmkg: I've extracted them to a migration-specific util file. Do let me know if it looks good.

src/platform/backends/hyperv_migration/hyperv_target_transaction.cpp
xmkg: `<name/"snapshot">` is the display name. This one is the snapshot's name on the disk. HCS expects snapshots to be named as index.avhdx, so the migration code uses that format.

src/daemon/daemon.cpp
xmkg: The old hyperkit migration returned the result of the migration: https://github.com/canonical/multipass/blob/2b4cade727c201050c5e70ae6e82f2dc57916c71/src/daemon/daemon.cpp#L3184-L3189

I'll make the messages indicate that the driver change has succeeded.

xmkg: Done, lmk what you think.

src/daemon/daemon.cpp
xmkg: I've removed these functions. LMK what you think.

src/daemon/daemon.cpp
xmkg: See above

src/daemon/daemon.cpp
xmkg: Done

xmkg:
@sharder996, I think I've addressed all of your comments. Please let me know what you think.

src/daemon/daemon.cpp
sharder996: Ok, I'm fine with it then.

src/daemon/hyperv_migration.h
sharder996: Yeah, its fine. This is kinda temp code anyways.

src/daemon/daemon.cpp
sharder996: Ok, with fresh eyes I understand now. `hyper-v` is the string that has aliases and needs to be normalized.

packaging/windows/wix/Package.wxs
```diff
@@ -17,7 +17,9 @@
                 Key='SOFTWARE\Microsoft\Windows NT\CurrentVersion' Name='CurrentBuildNumber' />
         </Property>
 
-        <Property Id="DRIVER" Value="hyperv" Secure="yes" />
+        <Property Id="DRIVER" Value="hyperv_api" Secure="yes" />
```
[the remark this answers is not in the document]
sharder996: I guess the feature flag is still in and has not been removed. In the meantime there will be a mismatch since feature flags don't work in WiX code.

sharder996 (dismissed):
From a code POV I have no further comments. I expect there still to be a bit of churn from Copilot reviews and possibly from @Laefy's testing as well. Ping me if you need a re-review or anything else.

packaging/windows/wix/Package.wxs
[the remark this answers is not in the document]
xmkg: Yeah, non-issue since we have to merge the feature flag removal PR before release anyway.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -585,6 +612,22 @@ std::optional<IPAddress> HCSVirtualMachine::management_ipv4()
...
+    hcs_system.reset();
+    if (!resources_released)
+    {
+        throw ComputeSystemStateException{
+            "Could not release HCS resources before restoring snapshot '{}.{}'",
+            get_name(),
+            name};
+    }
+
+    BaseVirtualMachine::restore_snapshot(name, specs);
```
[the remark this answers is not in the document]
xmkg: Incorrect. VirtdiskSnapshot::apply_impl() already restores those.

src/daemon/hyperv_migration.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+    recover();
+}
+
+void multipass::hyperv::HyperVMigrationTargetRecords::recover()
+{
+    std::error_code error;
+    auto iterator = MP_FILEOPS.dir_iterator(instances_root, error);
+    if (error || !iterator)
+        return;
```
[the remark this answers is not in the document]
xmkg: Fair enough. Done.

src/platform/backends/shared/windows/net_io_api.cpp
```diff
@@ -54,4 +54,9 @@ IpNetTableResult NetIOAPI::GetIpNetTable2(ADDRESS_FAMILY Family) const
     const auto error = ::GetIpNetTable2(Family, &table);
     return {error, IpNetTable{table, &free_mib_table}};
 }
+
+DWORD NetIOAPI::DeleteIpNetEntry2(const MIB_IPNET_ROW2* Row) const
+{
+    return ::DeleteIpNetEntry2(Row);
```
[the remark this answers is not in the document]
xmkg: Oh, chill. It compiles just fine.

src/platform/backends/hyperv_migration/hyperv_disk_layout.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    for (const auto& value : object.at("Snapshots").as_array())
+    {
+        const auto& snapshot = value.as_object();
+        const auto [_, inserted] = discovered.snapshots.emplace(
+            boost::json::value_to<std::string>(snapshot.at("Name")),
+            boost::json::value_to<std::string>(snapshot.at("Path")));
+        if (!inserted)
+            throw std::runtime_error{"Hyper-V contains duplicate checkpoint names"};
+    }
```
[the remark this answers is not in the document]
xmkg: False positive.

tests/unit/test_hyperv_migration_records.cpp
```diff
@@ -0,0 +1,857 @@
...
+        failing_file_ops.emplace(mpt::MockFileOps::inject<StrictMock>());
+        EXPECT_CALL(*failing_file_ops->first, try_read_file(manifest_path))
+            .WillOnce(Return(manifest));
+        EXPECT_CALL(
+            *failing_file_ops->first,
+            write_transactionally(
+                QDir::fromNativeSeparators(QString::fromStdWString(
+                    (target_root / "vault" / "multipassd-instance-image-records.json").wstring())),
+                _))
+            .WillOnce(Throw(std::runtime_error{"target store write failed"}));
```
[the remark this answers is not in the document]
xmkg: This path runs after transaction.commit(mapping) has already set committed = true (the mock is injected on the "Committing target records" phase, which follows the transaction commit). rollback() returns immediately on committed, so neither exists nor a second manifest read happens under the StrictMock. The single try_read_file expectation covers target_records.commit()'s read. No change needed.

xmkg:
Done addressing Copilot's comments. @Laefy, ready for testing.

Laefy:
@xmkg I can't ^^' The build was red and the branch has conflicts. Can you rebase please? :slightly_smiling_face:

xmkg:
@Laefy JFYI, I've rebased and fixed the build.

PR #5169, opened 2026-08-27
[hyper-v] Enable IP forwarding for AZ subnets

### LINKED ISSUE #5141: [hyperv_api] AZ subnets are isolated
Subnets created while using the `hyperv_api` backend are isolated resulting in a VM in `zone` not being able to ping a VM in `zone2`.

jimporter:
# Description

This PR enables IP forwarding for AZ subnets, which allows instances in different subnets to communicate with each other.
Resolves #5141.

## Testing

- Unit tests updated to mock out the new Win32 API calls (I just used `NiceMock` here since I don't see these calls as particularly interesting for our tests, but I could be convinced otherwise).
- Manual testing steps:

  1. Start instances in zones 1 and 2:
        ```
        multipass launch --name foo --zone zone1
        multipass launch --name bar --zone zone2
        ```
  3. Get the IPs for the instances: `multipass list`
  4. Shell into `foo`: `multipass shell foo`
  5. Make sure you can ping `bar`: `ping <ip-for-bar>`
  6. (optional) Do the same to make sure `bar` can ping `foo`

    Note: since the VMs are Linux, you can't use ICS hostnames to ping from inside them, so `ping bar.mshome.net` won't work.

## Checklist

### DIFF
--- a/src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
+++ b/src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
@@ -22,6 +22,8 @@
 #include <hyperv_api/hcs/hyperv_hcs_wrapper.h>
 #include <hyperv_api/hcs_virtual_machine.h>
 #include <hyperv_api/hcs_virtual_machine_exceptions.h>
+#include <hyperv_api/hyperv_api_string_conversion.h>
+#include <hyperv_api/net_io_api.h>
 #include <hyperv_api/virtdisk/virtdisk_wrapper.h>
 
 #include <multipass/constants.h>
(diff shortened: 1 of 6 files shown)


src/platform/backends/hyperv_api/hcs_virtual_machine_factory.cpp
```diff
@@ -350,17 +369,32 @@ std::unordered_map<std::string, std::string> HCSVirtualMachineFactory::create_az
...
+            MP_NETIOAPI.InitializeIpInterfaceEntry(&row);
+            row.Family = AF_INET;
+            row.InterfaceLuid = luid_for_hcn_network(name);
+            row.ForwardingEnabled = true;
+
+            if (const auto set_iface_result = MP_NETIOAPI.SetIpInterfaceEntry(&row);
+                set_iface_result != NO_ERROR)
+            {
+                mpl::error(log_category, "Could not set IP forwarding for {}", name);
+            }
```
jimporter, line 395: This part is worth calling out. I thought about when exactly we should enable forwarding, and decided that we should only do it when newly creating the AZ networks (usually first run of the daemon). That way, users can (if they really want to) change these settings, e.g. to simulate different network configurations.

However, this means that if you've run the nightlies on Windows, you'll need to clear out the AZ networks:

```powershell
Get-HNSNetwork | ? Name -Like "Multipass vNetwork" | Remove-HNSNetwork
```

xmkg: Makes sense to me. We're not removing and recreating HCN networks either, so any changes would persist until the next reboot, and enabling forwarding only on creation is consistent in that sense.

xmkg (approved):
LGTM, works as expected. It might be a good idea to add a CLI test for it, I'll submit that as a separate PR.

PR #5171, opened 2026-08-27
[docs] Pin Vale to 3.13 and ignore failing link

geoffreynyaga:
# Description
This PR fixes recent issues with spellcheck due to unpinned vale version in upstream. We are therefore pinning Vale to version 3.13.0.0 in the documentation requirements.txt file.

 ## Related Issue(s)
 Related to canonical/sphinx-stack#625

## Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -227,7 +227,8 @@ linkcheck_ignore = [
     "https://unix.stackexchange.com",  # it seems stackexchange is now blocking bots
     "https://developer.hashicorp.com/packer",
     "https://www.freedesktop.org/*",
-    "https://asciinema.org/*"
+    "https://asciinema.org/*",
+    r"https://askubuntu\.com/.*",
 ]
 
 linkcheck_retries = 3
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -29,4 +29,4 @@ sphinx-sitemap
 
 # Vale dependencies
 rst2html
-vale
+vale==3.13.0.0


docs/conf.py
```diff
@@ -227,7 +227,8 @@
     "https://unix.stackexchange.com",  # it seems stackexchange is now blocking bots
     "https://developer.hashicorp.com/packer",
     "https://www.freedesktop.org/*",
-    "https://asciinema.org/*"
+    "https://asciinema.org/*",
+    r"https://askubuntu\.com/.*",
```
ricab, line 231: A few questions on this:

- Why do you need to escape the first dot on this one but not the others?
- Why combine the escaped dot with the raw string?
- Why use regex syntax here (`.*`) where others use globbing (plain `*`)?

ricab (commented):
Thanks @geoffreynyaga, I have questions about the new link exclusion. Also, would you mind updating the PR title and description to reflect that other change as well?

docs/conf.py
geoffreynyaga: @ricab 
I was working on the robotics [repo](https://github.com/canonical/robotics_documentation/blob/67d9cb7331f4e44716946bd8e0657563dd0edfd5/docs/conf.py#L251) and I learn't that Sphinx recommends we use regular expressions in the ignore list (and other lists as well) https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_ignore 

The interesting question now becomes why the other url pattern still works e.g. `https://asciinema.org/*`. A bit of research and claude tells me that Sphinx uses some internal regex matching; so in the url above, the unescaped dots still match literal dots. I would recommend we update the list to use regex, perhaps in another PR

ricab: OK, makes sense to me. 

So just to be entirely sure, you really want a string with the '\' character followed by the '.' character, correct? If so, I'd say this is ready to go.

geoffreynyaga: This is one of the highly debated decisions in python. 
tldr; Both single and double strings are accepted, the decision is usually a style guide/preference one. In [PEP8](https://peps.python.org/pep-0008/#string-quotes), the standard recommends you choose one and stick with it. However, I have used both black https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings  and uv's Ruff formatter https://docs.astral.sh/ruff/settings/#format_quote-style and they both recommend double quotes. VScode also enforces this in some configurations.
So I believe are good to go.

geoffreynyaga: I will then create another PR to update the other regex entries instead if doing now and rebasing later

ricab: Hi @geoffreynyaga, ok for the regex PR. 

As to the string, I wasn't referring to single or double strings at all. Rather that you are using a raw string (with the leading `r` before the opening quotes), but you still have a `\` before the `.`. I guess this is intentional now, so fine.

ricab (approved):
Approving with the understanding that the remaining links will be separately morphed into regular expressions.

PR #5180, opened 2026-08-28
[daemon] Restore mount handlers from snapshots

xmkg:
Fix snapshot mount restore bug due to short-circuit boolean evaluation.

MULTI-2900


ricab (approved):
Yup, thanks for fixing!

PR #5192, opened 2026-09-01
[tests] Fix ASSERT_DEBUG_DEATH in Windows

xmkg:
The [Aa] pattern is not supported in Windows' gtest regex implementation, whereas \\w is not supported on Mac. Trim the pattern to "ssert" to make it work on both platforms.

### DIFF
--- a/tests/unit/test_cli_prompters.cpp
+++ b/tests/unit/test_cli_prompters.cpp
@@ -190,7 +190,7 @@ TEST_F(CLIPrompters, failsIfNoNetworks)
     EXPECT_CALL(mock_terminal, cin_is_live()).WillRepeatedly(Return(false));
     EXPECT_CALL(mock_terminal, cout_is_live()).WillRepeatedly(Return(false));
     // https://google.github.io/googletest/advanced.html#regular-expression-syntax
-    ASSERT_DEBUG_DEATH(prompter.bridge_prompt(nets), "[Aa]ssert");
+    ASSERT_DEBUG_DEATH(prompter.bridge_prompt(nets), "ssert");
 }
 
 TEST_P(BridgePrompterTests, correctlyReturns)
--- a/tests/unit/test_utils.cpp
+++ b/tests/unit/test_utils.cpp
@@ -394,7 +394,7 @@ TEST(Utils, trimNewlineAssertionWorks)
 {
     std::string s{"wrong"};
     // https://google.github.io/googletest/advanced.html#regular-expression-syntax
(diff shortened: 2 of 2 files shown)


ricab (commented):
Hi @xmkg, I am fine with this, but could you please share where the issue manifested? I am sure you saw it somewhere, but I am surprised that `\w` doesn't work on macOS (the docs seem to indicate otherwise).

xmkg:
> Hi @xmkg, I am fine with this, but could you please share where the issue manifested? I am sure you saw it somewhere, but I am surprised that `\w` doesn't work on macOS (the docs seem to indicate otherwise).

Sure thing! Here you go: https://github.com/canonical/multipass/commit/e5e2cfed83ee99c6f14fbb137c15059d0219ea20

xmkg:
> > Hi @xmkg, I am fine with this, but could you please share where the issue manifested? I am sure you saw it somewhere, but I am surprised that `\w` doesn't work on macOS (the docs seem to indicate otherwise).
> 
> Sure thing! Here you go: [e5e2cfe](https://github.com/canonical/multipass/commit/e5e2cfed83ee99c6f14fbb137c15059d0219ea20)

This is the previous commit that I `fixed` Windows :D https://github.com/canonical/multipass/commit/da7262ea8287132d8d6c1c015cfa9d3d8b63d6ad

ricab (commented):
But didn't `\\w` use to work work on macOS? I didn't find any such failure in CI. Please let me know if I am missing something, but it looks to me like this was just an oversight (maybe a harness running on a macOS environment and forgetting that we have windows). I wonder if we could just revert... 

That said, this is a very small thing. If you're confident this is what we need, let me know and I will approve. But I will also approve the `\\w` if it passes CI.

xmkg:
> But didn't `\\w` use to work work on macOS? I didn't find any such failure in CI. Please let me know if I am missing something, but it looks to me like this was just an oversight (maybe a harness running on a macOS environment and forgetting that we have windows). I wonder if we could just revert...
> 
> That said, this is a very small thing. If you're confident this is what we need, let me know and I will approve. But I will also approve the `\\w` if it passes CI.

ASSERT_DEBUG_DEATH is debug-only, so it does not generate any failures in our CI configuration matrix. It might be a good idea to run debug builds on CI, maybe on a schedule, to detect debug-only failures like this.

ricab (approved):
Hmm, good point. OK I'll approve and leave it up to you if you want to follow up.

PR #5216, opened 2026-09-07
[fix] Fix ordering of struct conversion from json

tobe2098:
# Description

- What does this PR do? Fixes the ordering of fields in the VMImageInfo struct when converting from a json to the struct. It also adds designated initialization as a preventative measure for this bug.

## Testing

- Unit tests

## Checklist


ricab:
Hey @tobe2098, thanks for submitting. Before I review, would you mind clarifying what exactly is the bug that this fixes and how can I reproduce it? Thanks in advance.

tobe2098:
Hi @ricab, it does not fix any apparent bug, but the incorrectness of the field ordering affects the #5175 PR, which could potentially cause some issues.

tests/unit/test_vm_image_info.cpp
```diff
@@ -0,0 +1,70 @@
...
+    "release_title": "12",
+    "items": {
+        "x86_64": {
+            "image_location": "https://example.com/debian-12-amd64.qcow2",
+            "id": "debian-12-hash",
+            "version": "20250804",
+            "size": 444896256
+        }
+    }
+})";
```
ricab: To avoid repeating magic values (here and in expectations), it would be preferable to have vars being used on both ends. You could achieve that by turning this into a format string and feeding it those variables. You could inline or keep it here, up to you.

tests/unit/test_vm_image_info.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace mp = multipass;
+
+using namespace testing;
+
+namespace
+{
+// A single distro entry, matching the structure of a custom image host manifest value. The
+// `release_title` and `release_codename` fields are given deliberately distinct values so that a
+// mix-up between them (e.g. from positional aggregate initialization) is detected.
```
ricab: Personally, I would drop this comment. It's just one more thing to read, not very useful IMO. Perhaps just `// keep fields distinct`?

ricab (commented):
Hey @tobe2098, thanks for the fix. +1 for the designated initializers. Just a couple of proposals inline, nothing critical.

tests/unit/test_vm_image_info.cpp
```diff
@@ -0,0 +1,86 @@
...
+    "release_title": "{}",
+    "items": {{
+        "x86_64": {{
+            "image_location": "{}",
+            "id": "{}",
+            "version": "{}",
+            "size": {}
+        }}
+    }}
+}})",
```
ricab, line 56: You might extract this to a separate var to ease the formatting, which I found a little awkward as it is. But I'll leave that up to you.

ricab (approved):
Alright, looks good. And well spotted!

PR #5222, opened 2026-09-08
[hcs] Remove COM port device from serial

xmkg:
Attaching a COM port device to the serial output adversely affects the boot time (adds around ~50s), as the Linux kernel defaults to 9600 baud and there is no clean way to change that.

Removing the COM port device and the named pipe for the time being, as it was a troubleshooting escape hatch anyway.

MULTI-2895


sharder996 (approved):
Tested and verified. Working as described.

PR #5223, opened 2026-09-09
[hcs] Ensure VM endpoints are deleted

### LINKED ISSUE #5220: HCS: Purging an instance leaves network endpoints behind
**Describe the bug**
The Windows HCS backend leaks network endpoints when purging an instance. The daemon terminates the VM before factory cleanup, which then fails to reopen it and skips endpoint deletion.

**To Reproduce**
Run in elevated PowerShell with the HCS backend:

$before = @(Get-HnsEndpoint | Select-Object -ExpandProperty Id)
multipass launch --name endpoint-leak
$created = @(Get-HnsEndpoint | Where-Object Id -NotIn $before |
    Select-Object -ExpandProperty Id)

multipass delete --purge endpoint-leak

Get-HnsEndpoint | Where-Object Id -In $created |
    Format-Table Id, Name, MacAddress

Endpoints created for the instance remain after purge. Avoid creating other VMs or containers during these steps.

**Expected behavior**
Purging an instance removes its endpoints, even if the VM has already terminated.

**Logs**
The isolated API reproduction confirmed:

Termination: 0x00000000
Reopen:      0x8037010E (HCS_E_SYSTEM_NOT_FOUND)
Endpoint remains after skipped cleanup.

**Additional info**

• OS: Windows
• Backend: HCS

**Additional context**
Cleanup currently depends on retrieving the VM’s  RuntimeId . Tagging endpoints with a deterministic instance-based  Name would allow discovery and deletion without reopening the VM.

xmkg:
The previous approach was relying on opening the VM to
obtain the RuntimeId of the VM which means VM has to be in
running state.

This patch eliminates the need by tagging the endpoints with
the VM name.

Closes: #5220 

MULTI-2898

### DIFF
--- a/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
+++ b/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
@@ -319,8 +319,9 @@ OperationResult HCNWrapper::find_endpoints_by_name(const std::string& name,
 
     // No filter -- enumerate every endpoint known to HCN, then inspect each one's `Name`
     // individually. This does not require the owning compute system to be open/alive.
-    const auto result =
-        API().HcnEnumerateEndpoints(L"{}", out_ptr(json_output), out_ptr(result_msgbuf));
+    const auto result = API().HcnEnumerateEndpoints(L"{}",
+                                                    out_ptr(json_output),
+                                                    out_ptr(result_msgbuf));
 
     if (FAILED(result) || !json_output)
     {
@@ -351,9 +352,9 @@ OperationResult HCNWrapper::find_endpoints_by_name(const std::string& name,
 
         UniqueCotaskmemString query_result{}, query_error{};
         const auto query_status = API().HcnQueryEndpointProperties(endpoint.get(),
-                                                                    L"{}",
-                                                                    out_ptr(query_result),
-                                                                    out_ptr(query_error));
+                                                                   L"{}",
+                                                                   out_ptr(query_result),
+                                                                   out_ptr(query_error));
         if (FAILED(query_status) || !query_result)
         {
             mpl::warn(log_category,
@@ -445,8 +446,9 @@ OperationResult HCNWrapper::enumerate_networks(std::vector<std::string>& out_net
     UniqueCotaskmemString enumerate_result{}, result_msgbuf{};
 
     // List all HCN network GUIDs
-    const auto result =
-        API().HcnEnumerateNetworks(L"{}", out_ptr(enumerate_result), out_ptr(result_msgbuf));
+    const auto result = API().HcnEnumerateNetworks(L"{}",
+                                                   out_ptr(enumerate_result),
+                                                   out_ptr(result_msgbuf));
     if (enumerate_result)
     {
         // json_output would contain the network GUIDs.
--- a/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
+++ b/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
@@ -322,7 +322,7 @@ OperationResult HCNWrapper::find_endpoints_by_name(const std::string& name,
     const auto result =
         API().HcnEnumerateEndpoints(L"{}", out_ptr(json_output), out_ptr(result_msgbuf));
 
-    if (!result || !json_output)
+    if (FAILED(result) || !json_output)
     {
         return {result, {result_msgbuf ? result_msgbuf.get() : L""}};
     }
--- a/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
+++ b/src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
@@ -186,6 +186,18 @@ std::pair<OperationResult, UniqueHcnNetwork> open_network(const std::string& net
     return std::make_pair(result, std::move(network));
 }
 
+std::pair<OperationResult, UniqueHcnEndpoint> open_endpoint(const std::string& endpoint_guid)
+{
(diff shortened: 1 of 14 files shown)


src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
```diff
@@ -298,6 +310,80 @@ OperationResult HCNWrapper::enumerate_attached_endpoints(
...
+    if (ec)
+    {
+        return {E_FAIL, L"Json parse error"};
+    }
+
+    for (const auto& elem : as_json.as_array())
+    {
+        const std::string endpoint_guid{elem.as_string()};
+
+        const auto& [open_result, endpoint] = open_endpoint(endpoint_guid);
```
Laefy, line 343: HcnCloseEndpoint never called? (cf wil::unique_any for resource clean up)

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -337,19 +338,26 @@ void HCSVirtualMachine::set_compute_system_callback_handler()
...
     // Additional endpoints, a.k.a. extra interfaces.
     std::ranges::transform(description.extra_interfaces,
                            std::back_inserter(params),
-                           [](const auto& v) -> hcn::CreateEndpointParameters {
+                           [&endpoint_name](const auto& v) -> hcn::CreateEndpointParameters {
                                return {.network_guid = multipass::utils::make_uuid(v.id),
                                        .endpoint_guid = mac2uuid(v.mac_address),
-                                       .mac_address = replace_colon_with_dash(v.mac_address)};
+                                       .mac_address = replace_colon_with_dash(v.mac_address),
+                                       .name = endpoint_name};
```
Laefy, line 360: The GUID of each endpoint looks to be deterministic too (depends on the `description`, which I suppose is fixed once a VM is created). Why not just recreating all the endpoints GUID from that description, instead of introducing a Name property?

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: open_endpoint returns result + UniqueHcnEndpoint, which handles the cleanup automatically:

```c++
struct HcnEndpointCloser
{
    void operator()(HCN_ENDPOINT p) const noexcept
    {
        (void)API().HcnCloseEndpoint(p);
    }
};

using UniqueHcnEndpoint = std::unique_ptr<std::remove_pointer_t<HCN_ENDPOINT>, HcnEndpointCloser>;
```

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Yes, the endpoint GUIDs are deterministic.

We only have the name of the VM in the `void remove_resources_for_impl(const std::string& name) override;` function. We might either provide the VM spec as an argument or pull the VM spec from the daemon inside the function and reconstruct the original endpoint GUID list. I opted to use the name field instead to help with troubleshooting later -- we'd have an easier time determining which HCN endpoints belong to multipass this way, especially on systems that use WSL and/or Docker. Plus, it's easier to query which HCN endpoints belong to which VM without requiring the VM to be up. What do you think? Does this sound reasonable, or should we try reconstructing the GUID list?

Laefy: I agree on the informative part. It could be useful to have it somewhere in the endpoint properties if needed.
However, I think it's easier to maintain a system if when making change to an entity, we use its ID to do it. Imo, more generally, I think multipass should maintain a clear list of the ids of the external entities it "owns", or at least be able to rebuild it if need be, which is the case for the endpoints here. This would indeed mean that VMSpec should be passed to `remove_resources_for` and subsequent calls.
But, I'm not against your implementation, it's perfectly reasonable, I just wanted to challenge a bit the choice ^^

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
Laefy: Ah yes, I'm blind, my bad!

Laefy (approved):

Laefy:
Small question: you don't use the PR template?

src/platform/backends/hyperv_api/hcn/hyperv_hcn_wrapper.cpp
xmkg: No worries :) It's not immediately obvious since it's hidden behind a couple of indirections.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Sounds reasonable to me. What about keeping the name as an informative tag but rebuilding the GUID list for removal?

xmkg:
> Small question: you don't use the PR template?

I skip it unless I have relevant information to fill in

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
Laefy: Yes, that'd be perfect

xmkg: Hmm, I looked into it a bit, but it's not trivial to pull the spec into remove_resources_for_impl() or pass it as an argument. In some cases, the spec might not be present, so it would either have to be an optional arg or there would need to be two remove_resources_for_impl overloads. Given that only the HCS backend would benefit from it, and we already have a working way and the endpoints are ephemeral, I'd say let's defer it to a later refactor.

Laefy: Sounds good!

Issue #5235, opened 2026-09-11
multipassd intermittently becomes stuck on shutdown due to mount teardown in Linux


xmkg:
**Describe the bug**
`multipassd` intermittently becomes stuck on shutdown if there's a VM with a mount (QEMU). 

```
Eyl 11 20:59:06 mustafa-ubuntu-workstation dnsmasq[882486]: exiting on receipt of SIGTERM
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: Goodbye!
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: QMP: {"timestamp": {"seconds": 1789149546, "microseconds": 724393}, "event": "SHUTDOWN", "data": {"guest": false, "reason": "host-signal"}}
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: VM shut down
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: qemu-system-x86_64: terminating on signal 15 from pid 1 (/sbin/init)
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: 
Eyl 11 20:59:06 mustafa-ubuntu-workstation multipassd[882217]: Executing 'findmnt --type 9p | grep '/home/ubuntu/tmpkt6cir65 mdeea958bf7e7385692198cabdaae23''
```

Here we can see that the daemon is being shut down; snap tries to stop the systemd service, and systemd sends SIGTERM to multipassd and all its child processes to signal them to exit gracefully (control-group).  The daemon receives the signal and prints `Goodbye!`, then destructors run while the qemu child process terminates itself due to the SIGTERM it received. Then the mount teardown code triggers; it tries to execute `findmnt --type 9p` through `ssh_exec_process,` but it becomes wedged due to the following:

a-) The VM is no longer alive
b-) The established timeout of the SSH connection is disabled to fix an SFTP bug; the dead connection hangs forever
c-) ssh_exec_process does not check the VM state

Hence, the daemon becomes stuck while shutting itself down.

**Expected behavior**
The daemon should shut down cleanly without getting stuck.

**Logs**
Please provide logs from the daemon, see [Access logs](https://canonical.com/multipass/docs/stable/how-to-guides/troubleshoot/access-logs/) on where to find them on your platform.

**Additional info**
- OS: Ubuntu 24.04
- CPU architecture or model: x86
[...]

PR #5237, opened 2026-09-11
[base-vm] Check VM state in ssh_exec_process

xmkg:
The VM might have been shut down while the cached session is alive, and given that the SSH session's established state does not have a timeout, it might wedge the caller forever.

Check if the VM is running before proceeding with the exec.

## Related Issue(s)

#5235 

MULTI-2902

### DIFF
--- a/src/platform/backends/shared/base_virtual_machine.cpp
+++ b/src/platform/backends/shared/base_virtual_machine.cpp
@@ -247,6 +247,14 @@ std::unique_ptr<mp::SSHProcess> mp::BaseVirtualMachine::ssh_exec_process(const s
 
     std::optional<std::string> log_details = std::nullopt;
     bool reconnect = true;
+
+    // Fail fast: a stale cached session would block forever (established timeout is LONG_MAX).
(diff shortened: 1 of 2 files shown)


src/platform/backends/shared/base_virtual_machine.cpp
```diff
@@ -787,10 +795,10 @@ void mp::BaseVirtualMachine::persist_generic_snapshot_info() const
     auto count_path = instance_dir.filePath(count_filename);
 
     QFile head_file{head_path};
-    auto head_file_rollback =
-        make_common_file_rollback(head_path,
-                                  head_file,
-                                  std::to_string(head_snapshot->get_parents_index()) + "\n");
+    auto head_file_rollback = make_common_file_rollback(
```
xmkg, line 798: stray formatting.

tobe2098 (approved):
LGTM Mustafa! Ideally, once we use the intended libssh interface, this will not be necessary anymore, but for now it is a good safeguard.

Issue #5080, opened 2026-07-15
Daemon blocks indefinitely on spawning new SSHProcess when the underlying VM has shut down


tobe2098:
**Describe the bug**
The daemon occasionally hangs (permanently) when `multipass launch` results in an aborted start instance. No timeouts bail out the daemon. 

From [gdb traces](https://github.com/user-attachments/files/30047776/logs.txt), the culpable seems to be the `state_mut` being locked in exec_process in BaseVM. While this is locked, the ssh session hangs because the VM is down, and there is no timeout. As a consequence, the mutex stays locked and later deadlocks the main thread upon `on_shutdown` being called.

The deadlock is caused by an indefinite block on `poll` due to a lack of timeouts in the `ssh_session`.

**To Reproduce**
Use the testing steps in #5079.

**Expected behavior**
No deadlocks.

**Logs**
```
Jul 15 13:55:41 thinkpad multipassd[1855082]: plain_ssh_process.cpp:222, `std::string multipass::PlainSSHProcess::read_stream(StreamType, int)`: channel closed
Jul 15 13:55:41 thinkpad multipassd[1855082]: failed to run '[ -e /var/lib/cloud/instance/boot-finished ]', exit_code 1 (no stderr output)
Jul 15 13:55:42 thinkpad multipassd[1855082]: Executing '[ -e /var/lib/cloud/instance/boot-finished ]'
Jul 15 13:55:42 thinkpad multipassd[1855082]: plain_ssh_process.cpp:217, `std::string multipass::PlainSSHProcess::read_stream(StreamType, int)`: (type = 1, timeout = -1)
Jul 15 13:55:42 thinkpad multipassd[1855082]: plain_ssh_process.cpp:222, `std::string multipass::PlainSSHProcess::read_stream(StreamType, int)`: channel closed
Jul 15 13:55:42 thinkpad multipassd[1855082]: failed to run '[ -e /var/lib/cloud/instance/boot-finished ]', exit_code 1 (no stderr output)
Jul 15 13:55:42 thinkpad multipassd[1855082]: QMP: {"timestamp": {"seconds": 1784116542, "microseconds": 408511}, "event": "RTC_CHANGE", "data": {"offset": -1, "qom-path": "/machine/unattached/device[4]/rtc"}}
Jul 15 13:55:43 thinkpad multipassd[1855082]: Executing '[ -e /var/lib/cloud/instance/boot-finished ]'
[...]

PR #5238, opened 2026-09-11
[ssh] Implement SSH keepalive instead of infinite timeout

xmkg:
# Description

The SSH session timeout (`SSH_OPTIONS_TIMEOUT`) has been infinite since 67bb8885a2, so that the SFTP server backing sshfs mounts never gives up its blocking `sftp_get_client_message()` read while a mount sits idle. The side effect is that every other blocking libssh call (auth, channel open/exec, sftp round trips, writes) can hang forever on an unresponsive peer.

This PR makes the SFTP server independent of the session timeout and then bounds the timeout.

## Related Issue(s)

Relates to #5080

## Testing

- Manual testing steps (Linux, QEMU backend):

  1. Run the `sshfs_server` from this branch against a fresh instance (with `multipass-sshfs` installed) and mount a directory; write and read files from both sides.
  2. Leave the mount idle for several minutes. The mount stays up and works afterwards. With sshd at `LogLevel DEBUG1` in the instance, `journalctl _PID=<sshd user child>` shows `server_input_global_request: rtype keepalive@openssh.com` every 30s; TCP counters on the connection change only at those points.
  3. Send SIGTERM to `sshfs_server`: it stops within a second with exit code 0 and the instance unmounts.
  4. `multipass stop` the instance while mounted: `sshfs_server` exits with failure through the existing "SFTP server thread stopped unexpectedly" path.

N/A

## Checklist

## Additional Notes

- The keepalive interval (30s) and the session timeout (10s) are independent: the former is how long the server waits for client activity before probing, the latter how long any single blocking call waits for a reply. The poll interval is not subject to the session timeout.
- Missed-reply counting (OpenSSH's `ServerAliveCountMax`) is not implemented; a silently vanished peer is detected once TCP retransmissions of the unacked keepalive give up. Can be a follow-up if wanted.

MULTI-2901

### DIFF
--- a/include/multipass/ssh/libssh_wrapper.h
+++ b/include/multipass/ssh/libssh_wrapper.h
@@ -68,6 +68,7 @@ public:
                                              void* dest,
                                              uint32_t count,
                                              int is_stderr) const;
+    virtual int ssh_channel_poll_timeout(ssh_channel channel, int timeout, int is_stderr) const;
     virtual int ssh_channel_request_pty_size(ssh_channel channel,
                                              const char* term,
                                              int cols,
--- a/include/multipass/ssh/libssh_wrapper.h
+++ b/include/multipass/ssh/libssh_wrapper.h
@@ -50,6 +50,9 @@ public:
     virtual int ssh_userauth_publickey(ssh_session session,
                                        const char* username,
                                        const ssh_key privkey) const;
+    // Declared in libssh/server.h, but it is a plain client-side global request
+    // ("keepalive@openssh.com") that waits for the peer's reply. Always returns SSH_OK.
+    virtual int ssh_send_keepalive(ssh_session session) const;
 
     // --- channel -------------------------------------------------------------
     virtual ssh_channel ssh_channel_new(ssh_session session) const;
--- a/src/ssh/libssh_wrapper.cpp
+++ b/src/ssh/libssh_wrapper.cpp
@@ -143,6 +143,11 @@ int mp::Libssh::ssh_channel_read_nonblocking(ssh_channel channel,
     return ::ssh_channel_read_nonblocking(channel, dest, count, is_stderr);
 }
 
+int mp::Libssh::ssh_channel_poll_timeout(ssh_channel channel, int timeout, int is_stderr) const
+{
+    return ::ssh_channel_poll_timeout(channel, timeout, is_stderr);
+}
+
 int mp::Libssh::ssh_channel_request_pty_size(ssh_channel channel,
                                              const char* term,
                                              int cols,
--- a/src/ssh/libssh_wrapper.cpp
+++ b/src/ssh/libssh_wrapper.cpp
@@ -17,6 +17,8 @@
 
 #include <multipass/ssh/libssh_wrapper.h>
 
+#include <libssh/server.h>
+
 extern "C"
 {
 int sftp_reply_version(sftp_client_message msg);
@@ -88,6 +90,11 @@ int mp::Libssh::ssh_userauth_publickey(ssh_session session,
                                        const ssh_key privkey) const
 {
     return ::ssh_userauth_publickey(session, username, privkey);
+}
+
+int mp::Libssh::ssh_send_keepalive(ssh_session session) const
+{
+    return ::ssh_send_keepalive(session);
 }
 
 // --- channel ----------------------------------------------------------------
--- a/src/ssh/plain_ssh_session.cpp
+++ b/src/ssh/plain_ssh_session.cpp
@@ -50,16 +50,22 @@ mp::PlainSSHSession::PlainSSHSession(const std::string& host,
         throw mp::SSHException("could not allocate ssh session");
 
     /**
-     * Sometimes ssh_connect blocks while a VM is booting up where
-     * it won't return to the caller at all. Since the timeout is
-     * set to ~infinity the ssh_connect would block forever.
+     * Two timeouts govern this session (libssh reads SSH_OPTIONS_TIMEOUT as long seconds):
      *
-     * The attempt timeout is set as timeout before the ssh_connect
-     * in order to prevent that from happening. The established timeout
-     * is used afterward to prevent timing out in connected state.
+     *  - connect_timeout_secs bounds ssh_connect. A booting VM may accept the TCP connection
+     *    and then stall, so this needs to be short.
+     *  - established_timeout_secs is installed once connected and bounds every subsequent
+     *    blocking libssh call that uses SSH_TIMEOUT_DEFAULT: authentication, channel open/exec
+     *    requests, sftp request/response round trips, channel writes waiting for window space,
+     *    global requests (keepalives), etc. It is intentionally finite so that a hung peer
+     *    cannot block a thread forever.
+     *
+     * Consequently, components that keep a session idle for long periods (e.g. an sshfs mount
+     * whose client may send nothing for hours) must not rely on blocking reads. They should
+     * poll with an explicit timeout and send keepalives instead (see SftpServer::run).
      */
-    constexpr long connect_timeout_secs = 5; // < how long to wait for ssh_connect
-    constexpr long established_timeout_secs = std::numeric_limits<long>::max();
+    constexpr long connect_timeout_secs = 5;      // < how long to wait for ssh_connect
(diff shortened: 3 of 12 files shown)


xmkg:
@jimporter, the auto-assignment picked you -- feel free to reassign if you feel like there are too many reviews pending on you already.

xmkg:
I noticed that @tobe2098 already filed an issue for this. I think it'd make sense to ask him for a review :)  So I stumbled upon this issue today, and I know the history behind the infinite SSH established timeout. It was basically there to prevent inactive SFTP sessions from being dropped, but on the other hand, it causes other bugs. I've implemented a keep-alive system and switched to bounded poll reads. Let me know what your impressions are.

xmkg:
Also, #5237 is complimentary.

tobe2098:
It looks good @xmkg, but it is a big change to merge onto the release instead of having it in the `vsock` branch with related changes. I have one question: in which context did you reproduce the issue? In my case the deadlock only occurred once I enabled cloud-init shutdowns. If there is another reasonable action that triggers this, then we would have to merge for the release, otherwise, I would advocate to get it in `sftp-session-api` instead. 

The reason that I am hesitant to include in `main` for the release is because other sections of the ssh code have been developed under the assumption of infinite session timeout. This change would require a lot of extra scrutiny if it were to be included for this release.

Otherwise, good thinking! It is close to the existing implementation in `sftp-session-api`, but the keepalive is something we will need there as well.

xmkg:
Hey @tobe2098, thanks for the review!

> I have one question: in which context did you reproduce the issue?

I've encountered this when I'm running the CLI tests on both Linux and Windows, in non-cloud-init cases. In my case, it happens most often in the mount tests. I've provided more detail in this issue: https://github.com/canonical/multipass/issues/5235. A repro would roughly look like this:

```bash
1. multipass launch -n vm
2. multipass mount --type native <src> vm # (9p; a classic sshfs mount works too).                                 
3. sudo systemctl stop snap.multipass.multipassd.service # (equivalently snap stop multipass, or any SIGTERM of the daemon).
```

It's a race, so it may take a while to reproduce. I'll attach a reproducer script too. 

The systemd unit is the default KillMode=control-group, so systemd SIGTERMs the whole cgroup at once -- qemu and dnsmasq included, meaning that they're being finalized out-of-band. 
 
In the scenario above, qemu dies immediately, then the daemon's own teardown runs the mount handler's deactivate, which SSHes back into the VM to findmnt/umount. The cached session still reports connected, so `ssh_channel_open_session` blocks on the now-dead peer -- and at the infinite `established_timeout_secs` it never returns. The service sits "stopping" until systemd's timeout, resulting in a wedged daemon. 

In principle, the problem can manifest itself whenever the following condition is true:

- The VM is not in the Running state for whatever reason, e.g., an out-of-band change
- An exec is exercised through the cached SSH session

> The reason that I am hesitant to include in main for the release is because other sections of the ssh code have been developed under the assumption of infinite session timeout.

The only beneficiary of the infinite timeout ATM is the SFTP session. The infinite timeout change was introduced just because of it (https://github.com/canonical/multipass/pull/3735/changes/67bb8885a2a2cf2823e2304ecd758d3c746fed04), so it's safe to say that the rest of the SSH-reliant code is not designed around this assumption. Hence, a potential regression would only affect the classic mounts. I've already had to fix another bug due to this change where the connections were getting stuck in the connecting state forever: https://github.com/canonical/multipass/commit/e4101b1cb7793967d99815e8bf067226f2e06cf2 . I'd say it's causing much more trouble than it's fixing.

Having an infinite timeout on the SSH session is extremely troublesome and causes a myriad of cascading failures and stuck states. If you ask me, I'd rather have a broken mount than a broken daemon.

As for testing, we'd need to create a long-running, idle VM, preferably with mounts, and check SSH session behavior. We'd expect the SSH sessions to be kept alive even when there's no activity. Another aspect to verify would be the sleep behavior. I'd expect the connections to persist.

xmkg:
The reproducer script, as promised.
[repro.sh](https://github.com/user-attachments/files/32191298/repro.sh)

Also, got a backtrace confirming that the main daemon thread is stuck on polling SSH while it's unresponsive:

```
[Switching to thread 1 (Thread 0x745338f4f300 (LWP 468156))]
#0  0x00007453392f84fd in __GI___poll (fds=0x7452d8009530, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
29    in ../sysdeps/unix/sysv/linux/poll.c
#0  0x00007453392f84fd in __GI___poll (fds=0x7452d8009530, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00005782d5d91919 in ssh_poll_ctx_dopoll ()
#2  0x00005782d5d92ebe in ssh_handle_packets ()
#3  0x00005782d5d92cda in ssh_handle_packets_termination ()
#4  0x00005782d5d76f5d in channel_open ()
#5  0x00005782d534f8ed in void multipass::SSH::throw_on_error<std::unique_ptr<ssh_channel_struct, void (*)(ssh_channel_struct*)>&, std::_Bind_front<int (multipass::Libssh::*)(ssh_channel_struct*) const, multipass::Libssh*>>(std::unique_ptr<ssh_channel_struct, void (*)(ssh_channel_struct*)>&, ssh_session_struct*, char const*, std::_Bind_front<int (multipass::Libssh::*)(ssh_channel_struct*) const, multipass::Libssh*>&&) ()
#6  0x00005782d534d80e in multipass::PlainSSHProcess::PlainSSHProcess(ssh_session_struct&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::unique_lock<std::mutex>) ()
#7  0x00005782d534cb03 in multipass::PlainSSHSession::exec(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#8  0x00005782d52ff721 in multipass::BaseVirtualMachine::make_ssh_process(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#9  0x00005782d52ff446 in multipass::BaseVirtualMachine::ssh_exec_process(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#10 0x00005782d526420b in multipass::QemuMountHandler::is_active() ()
#11 0x00005782d5264e95 in multipass::QemuMountHandler::~QemuMountHandler() ()
#12 0x00005782d5264f99 in multipass::QemuMountHandler::~QemuMountHandler() ()
#13 0x00005782d51d3827 in std::__detail::_Hashtable_alloc<std::allocator<std::__detail::_Hash_node<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::unique_ptr<multipass::MountHandler, std::default_delete<multipass::MountHandler> >, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::unique_ptr<multipass::MountHandler, std::default_delete<multipass::MountHandler> > > > > >, true> > >::_M_deallocate_node(std::__detail::_Hash_node<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::unique_ptr<multipass::MountHandler, std::default_delete<multipass::MountHandler> >, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::unique_ptr<multipass::MountHandler, std::default_delete<multipass::MountHandler> > > > > >, true>*) ()
#14 0x00005782d51af11e in multipass::Daemon::~Daemon() ()
#15 0x00005782d519348a in main ()
```

tobe2098 (approved):
LGTM Mustafa! I tested and it seems like the SSH sessions are kept alive properly

PR #5239, opened 2026-09-11
[daemon] List/Info: Skip runtime info if no IP

### LINKED ISSUE #5061: Instance with no IP causes the daemon to fail `list` and `info`

xmkg:
Skip attempting to obtain runtime info when no management IP is present
for a VM. This allows commands like list/info to succeed even when
there's a VM with no mgmt IP.

Closes #5061

MULTI-2896

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -1826,7 +1826,7 @@ try
         entry->set_current_release(current_release);
         entry->set_os(os);
 
-        if (request->request_ipv4() && MP_UTILS.is_running(present_state))
+        if (request->request_ipv4() && MP_UTILS.is_running(present_state) && vm.management_ipv4())
         {
             auto management_ip = vm.management_ipv4();
             auto all_ipv4 = vm.get_all_ipv4();
@@ -3784,7 +3784,7 @@ void mp::Daemon::populate_instance_info(VirtualMachine& vm,
     timestamp->set_seconds(created_time.toSecsSinceEpoch());
     timestamp->set_nanos(created_time.time().msec() * 1'000'000);
 
-    if (!no_runtime_info && MP_UTILS.is_running(present_state))
+    if (!no_runtime_info && MP_UTILS.is_running(present_state) && vm.management_ipv4())
         RuntimeInstanceInfoHelper::populate_runtime_info(vm,
                                                          info,
                                                          instance_info,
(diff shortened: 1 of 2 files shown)


src/daemon/daemon.cpp
```diff
@@ -1826,7 +1826,7 @@ try
         entry->set_current_release(current_release);
         entry->set_os(os);
 
-        if (request->request_ipv4() && MP_UTILS.is_running(present_state))
+        if (request->request_ipv4() && MP_UTILS.is_running(present_state) && vm.management_ipv4())
         {
             auto management_ip = vm.management_ipv4();
             auto all_ipv4 = vm.get_all_ipv4();
```
Laefy: Why not just putting this line and the for under the if below?

src/daemon/daemon.cpp
```diff
@@ -3784,7 +3784,7 @@ void mp::Daemon::populate_instance_info(VirtualMachine& vm,
     timestamp->set_seconds(created_time.toSecsSinceEpoch());
     timestamp->set_nanos(created_time.time().msec() * 1'000'000);
 
-    if (!no_runtime_info && MP_UTILS.is_running(present_state))
+    if (!no_runtime_info && MP_UTILS.is_running(present_state) && vm.management_ipv4())
         RuntimeInstanceInfoHelper::populate_runtime_info(vm,
```
Laefy, line 3789: Is it not possible to do the ssh connection without the IP?

xmkg: No, not until the VSOCK feature lands.

Laefy: Ok, maybe just put a comment on top of that then ^^

src/daemon/daemon.cpp
xmkg: done

src/daemon/daemon.cpp
xmkg: done

Laefy (approved):

PR #5243, opened 2026-09-14
[snap] Set multipassd stop-mode to sigterm

xmkg:
The default value, sigterm-all, sends sigterm to all processes spawned by multipassd, including dnsmasq and the QEMU processes. It causes the processes to be ended out-of-band, opening a window for race conditions.

Set stop-mode to `sigterm` to let multipassd handle the termination of its child processes as it should.

Relates-to: #5080, #5235

MULTI-2904


Laefy (approved):
Do you need help testing the snap

xmkg:
> Do you need help testing the

I did test it locally, but it'd be nice if you could also confirm that it's the daemon that now initiates the shutdown of the child processes (dnsmasq, qemu). The old behavior was systemd sending SIGTERM to the whole process group at once. Now only the daemon should receive the SIGTERM.

xmkg:
I've found a bug in the daemon that prevents us from merging this. If the daemon receives SIGTERM during a VM launch, the launch request won't be cancelled, so the daemon gets blocked until the RPC completes. This diverges from the old behavior where systemd would send SIGTERM to all, including the QEMU VM being launched, and the daemon would shut itself down.

The old behavior isn't right either, but fixing this properly will take much more effort than changing the KillMode, so parking this for now.

xmkg:
I'm going to close this one and open an issue instead.

xmkg:
@ricab, I need your opinion on this one. 

This branch fixes the problem stated in the description, and #5228 explains why I closed this branch without merging it. I've rebased and tested this branch now, and it works, but there's a caveat. 

With this branch, the daemon shuts QEMU down gracefully,  but not until the launch completes (https://github.com/canonical/multipass/pull/5243#issuecomment-5681301711), meaning the daemon will wait until the launch in flight finalizes properly. I feel like this is the right thing to do, but it will make the shutdown considerably longer should the case described above happen. What do you think?

The shorter version of the question is: `What should we do about in-flight launch requests when the daemon receives a graceful shutdown request (SIGTERM)?"`

On a side note, this bug also prevents QEMU VMs from getting suspended on daemon shutdown since the VM process will exit before the VM destructor runs, so we have more than one reason to fix this:

```c++
mp::QemuVirtualMachine::~QemuVirtualMachine()
{
    if (vm_process)
    {

        mp::top_catch_all(vm_name, [this]() {
            if (state == State::running)
            {
                suspend();
            }
            else
            {
                shutdown();
            }
        });
    }
}
```

ricab:
Hmm, I think waiting for it to complete before shutting down is the right thing to do. Anything else risks corrupting instances (which would be even worse on starts of existing instance that could have user data).

A couple of questions:

- Can a user who doesn't mind corrupting or loosing the instance still run a `multipass stop --force` or `multipass delete --purge` at that stage? If we already asked the gRPC shutdown, I suppose that is no longer possible. Any ideas to circumvent that? Perhaps waiting on start futures before asking for gRPC to shutdown?
- Is the default timeout for daemon shutdown (with the snap) enough for a full launch timeout + a full suspend timeout?

We have had multiple complaints about instances getting corrupted on daemon shutdown. This just might fix some of them.

xmkg:
- Can a user who doesn't mind corrupting or loosing the instance still run a multipass stop --force or multipass delete --purge at that stage? If we already asked the gRPC shutdown, I suppose that is no longer possible. Any ideas to circumvent that? Perhaps waiting on start futures before asking for gRPC to shutdown?

Right now, no, they can't since we're already calling shutdown_and_wait() for gRPC. To support that, we'd need to introduce a "shutting down" flag and only accept admission of a select set of gRPC requests during this period (e.g., only accept stop and delete).

- Is the default timeout for daemon shutdown (with the snap) enough for a full launch timeout + a full suspend timeout?

No. The snapcraft.yml does not have the stop-timeout set for multipassd. It's set to 5 minutes for my edge installation. We can change it, though.

ricab:
Thinking about it a little further, I think the ideal behavior for launches would be to just abort them. Users could re-launch when the daemon came back up. Starts are a different matter though and it is not easy to tell them apart once the preparation phase has finished and VM::start is called. We'd need to tag the futures. Worse, we'd need to find a way to implement cancellation... Probably more than you are going for here, so the easy answer is just wait for all.

On the hard shutdown timeout, maybe it is still better to have that overall 5 minute cap. If an instance hasn't finished starting and suspending in 5 min, it is unlikely it ever would. And 5 minutes for shutdown is already pretty high IMO. It doesn't shock me that we apply a stricter boundary in that situation.

xmkg:
> Worse, we'd need to find a way to implement cancellation

Agreed, handling it gracefully, or even forcefully, in a clean way will take a good amount of effort.

> If an instance hasn't finished starting and suspending in 5 min, it is unlikely it ever would.

This depends on what the user is launching (imagine a cloud-init with dozens of dependencies), but we can't set a timeout that covers everything either. Cancelling looks like the best option on the table.

ricab:
> This depends on what the user is launching (imagine a cloud-init with dozens of dependencies), but we can't set a timeout that covers everything either. Cancelling looks like the best option on the table.

Indeed. I was counting on heavy cloud-inits being the odd case. Not impossible for sure, but also not the most likely I would say. Importantly, heavy cloud-init work runs at the first boot, which means we'd be dealing with a launch and a hard termination would not loose anything unrecoverably.

PR #5259, opened 2026-09-15
[hcs] Fix issue with powering off after suspension

### LINKED ISSUE #5263: Multipass fails to force shutdown a suspended VM on Windows using the HCS backend
**Describe the bug**
If you suspend a VM, then try to forcefully power it off, the multipass daemon gets stuck.

**To Reproduce**
```
> multipass.exe launch --name suspend-shutdown-test
> multipass.exe suspend suspend-shutdown-test

> multipass.exe delete suspend-shutdown-test --purge # Gets stuck
> multipass.exe disable-zones 1 # Gets stuck
> multipass.exe stop suspend-shutdown-test --force # Gets stuck
```

**Expected behavior**
Multipass should not get stuck

**Additional info**
- OS: Windows 11 Pro

theartful:
# Description

If you suspend a VM, then try to forcefully power it off, the multipass daemon gets stuck. 
```
> multipass.exe launch --name suspend-shutdown-test
> multipass.exe suspend suspend-shutdown-test

> multipass.exe delete suspend-shutdown-test --purge # Gets stuck
> multipass.exe disable-zones 1 # Gets stuck
> multipass.exe stop suspend-shutdown-test --force # Gets stuck
```

The issue is that the VM stays in the suspended state, and an infinite loop happens here:
https://github.com/canonical/multipass/blob/33f26feaf45470011f9e07c3b22569ef5cb9b391/src/platform/backends/hyperv_api/hcs_virtual_machine.cpp#L476-L485
because we never go to either of the `stopped` state or `off` state.

The fix is to try to remove the suspension state file if we're trying to forcefully shutdown the VM (matching the behavior of the qemu backend for example).

Closes #5263

## Testing

  1. Launch an instance using the HCS backend and suspend it.
  2. Forcefully stop the suspended instance with `multipass stop --force <instance>`.
  3. Confirm that the instance is powered off and can be started again.

## Checklist


src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -462,8 +455,22 @@ void HCSVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
...
+        // FIXME: If the VM is suspended, this will produce an error message, as
+        // suspended VMs are already terminated.
+        //
+        // We also can't test if it's suspended using state == State::suspended,
+        // because "suspend" might fail to persist on disk, in which case the VM
+        // will be paused, and should be terminated.
+        // The proper solution is to fetch the api state and check whether
+        // the VM is in running/unknown/paused state, but that would cause a number
+        // of tests to fail, so let's just ignore the error log message for now.
         const auto r = HCS().terminate_compute_system(hcs_system);
```
xmkg, line 468: We shouldn't proceed with the wait lambda if terminate fails, since it's what triggers the state change via the compute system callback. We should also probably reset the hcs_system handle in compute_system_event_callback when a system exits.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -479,6 +486,15 @@ void HCSVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
         case VirtualMachine::State::stopped:
         case VirtualMachine::State::off:
             return multipass::utils::TimeoutAction::done;
+        case VirtualMachine::State::suspended:
```
xmkg, line 489: This does not need special handling -- suspended VMs are, by definition, terminated. We need to ensure that if a suspended VM gets a forceful shutdown call, we remove the suspend state file and set the VM state to OFF.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -462,8 +455,22 @@ void HCSVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
...
+        // because "suspend" might fail to persist on disk, in which case the VM
+        // will be paused, and should be terminated.
+        // The proper solution is to fetch the api state and check whether
+        // the VM is in running/unknown/paused state, but that would cause a number
+        // of tests to fail, so let's just ignore the error log message for now.
         const auto r = HCS().terminate_compute_system(hcs_system);
         mpl::debug(get_name(), "shutdown -> terminate_compute_system result: {}", r.code);
+
+        if (shutdown_policy == ShutdownPolicy::Poweroff)
+            remove_saved_state_file_if_exists();
```
xmkg, line 472: We can change this if block to

```
if suspended:
     if not remove_saved_state_file_if_exists:
           ... log a warning msg
     set state to off
```

xmkg (changes requested):

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -121,16 +121,20 @@ HCSVirtualMachine::HCSVirtualMachine(const std::string& network_guid,
...
-               state);
+               compute_state);
 
     // Reflect compute system's state
-    set_state(state);
-    HCSVirtualMachine::handle_state_update();
+    auto prev_state = state;
+    set_state(compute_state);
+
+    // Persist initial state even if unchanged
```
xmkg, line 135: This comment is a bit confusing -- is the state change persisted for prev_state != this->state anywhere else?

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -364,31 +366,36 @@ void HCSVirtualMachine::set_state(hcs::ComputeSystemState compute_system_state)
...
-    if (state == prev_state)
+void HCSVirtualMachine::set_state(VirtualMachine::State new_state)
+{
+    if (state == new_state)
         return;
 
-    mpl::info(get_name(), "set_state() -> State changed from {} to {}", prev_state, state);
+    mpl::info(get_name(), "set_state() -> State changed from {} to {}", state, new_state);
+    state = new_state;
+    handle_state_update();
```
xmkg, line 398: Ah okay, nevermind. Found my answer.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -468,23 +478,36 @@ void HCSVirtualMachine::shutdown(ShutdownPolicy shutdown_policy)
...
     case ShutdownPolicy::Poweroff:
         mpl::debug(get_name(),
                    "shutdown() -> Requested halt/poweroff, initiating forceful shutdown");
+
+        // FIXME: There is a rare case where suspend fails, and fails to terminate
+        // the VM as well. In this case the VM will be "paused". This is not handled
+        // for now.
+        if (state == State::suspended)
+        {
+            if (const auto ec = remove_saved_state_file_if_exists(); ec)
```
xmkg, line 493: The suspend code already tries to remove the saved state after a failed save call. Is there a reason to retry again here? I'd just let it proceed to terminate -- if it fails, it fails.

theartful: This is where the bug happened. If we don't remove the saved state file while shutting down, the VM will be convinced it's still suspended.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -502,34 +525,74 @@ void HCSVirtualMachine::suspend()
...
-    else
+
+    // If saved failed, we try to resume
+    if (const auto ec = remove_saved_state_file_if_exists(); ec)
     {
-        throw SaveComputeSystemException{"Could not pause VM for suspend: {}", pause_result};
+        // FIXME: This leaves the class in an undefined state.
+        // Although the suspend file is corrupt, terminating the VM will still
+        // transition it to the suspended state.
+        mpl::warn(get_name(),
```
xmkg: This will require manual intervention by the user. If we can't remove the file, we probably can't override it either. Right now, the existence of the suspend state file means a VM is suspended, so it's not wrong to say that the VM is suspended. If the save_compute_file produces a corrupt save file even when it reports an error, what we could do is save it to a temporary file before renaming it to the real one. This way, the VM won't be treated as suspended due to a corrupt state.

xmkg: We could use sth like `get_saved_state_file_path().tmp` for the name.

theartful: Good suggestion

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -502,34 +525,74 @@ void HCSVirtualMachine::suspend()
...
-            // Save succeeded. Now, it's safe to terminate the system.
-            shutdown(ShutdownPolicy::Poweroff);
-        }
-        else
-            throw SaveComputeSystemException{"Could not save the virtual machine state for VM `{}` "
-                                             "to the disk for suspend. Error details: {}",
-                                             get_name(),
-                                             r};
+        // Save succeeded. Now, it's safe to terminate the system.
+        return shutdown(ShutdownPolicy::Poweroff);
```
xmkg: Isn't this a void function? I guess this is an MSVC-specific extension

theartful: AFAIK this is standard C++

theartful: I can split it into two if you prefer

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Also, remove_saved_state_file_if_exists() already prints a warning, no? I think it would be better to avoid duplication

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Okay, it's part of the standard. Still, it's a bit confusing to see, though.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg, line 524: This is a bit lengthy as-is. What about something like:

```c++
void HCSVirtualMachine::suspend()
{
    mpl::debug(get_name(), "suspend() -> Suspending, current state {}", state);

    if (const auto r = HCS().pause_compute_system(hcs_system); !r)
        throw SaveComputeSystemException{"Could not pause VM `{}` for suspend: {}", get_name(), r};

    const auto save_result = HCS().save_compute_system(hcs_system, get_saved_state_file_path());
    if (save_result)
    {
        shutdown(ShutdownPolicy::Poweroff);
        return;
    }

    if (const auto ec = remove_saved_state_file_if_exists())
        mpl::warn(get_name(), "Could not remove partial state file '{}': {}", get_saved_state_file_path(), ec);

    const auto resume_result = HCS().resume_compute_system(hcs_system);
    if (resume_result)
    {
        update_current_state();
        throw SaveComputeSystemException{
            "Could not save the state of VM `{}` for suspend; the VM was resumed. Save error: {}",
            get_name(),
            save_result};
    }

    mpl::error(get_name(), "Suspend failed (save: {}; resume: {}), powering off", save_result, resume_result);
    shutdown(ShutdownPolicy::Poweroff);

    throw SaveComputeSystemException{
        "Could not save the state of VM `{}` for suspend or resume it; the VM was terminated. "
        "Save error: {}; resume error: {}",
        get_name(),
        save_result,
        resume_result};
}
``` 

We can even extract the failed save recovery path to its own function for clarity:

```c++
void HCSVirtualMachine::recover_from_failed_save() noexcept
{
    if (const auto ec = remove_saved_state_file_if_exists())
        mpl::warn(get_name(), "Could not remove partial state file '{}': {}", get_saved_state_file_path(), ec);

    if (const auto r = HCS().resume_compute_system(hcs_system))
        return update_current_state();  // or split into two lines, per the review thread

    mpl::error(get_name(), "Could not resume after failed suspend ({}); powering off", r);
    try
    {
        shutdown(ShutdownPolicy::Poweroff);
    }
    catch (const std::exception& e)
    {
        mpl::error(get_name(), "Power off after failed suspend also failed: {}", e.what());
    }
}
```

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
xmkg: Then, attempting a temporary save would solve this, too

theartful: Not sure how. If the machine is suspended, then this file has to exist, and we have to try to remove it.

xmkg: Discussed out-of-band.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -687,4 +708,46 @@ std::shared_ptr<Snapshot> HCSVirtualMachine::make_specific_snapshot(const QStrin
...
+            return ec;
+        }
+    }
+
+    return {};
+}
+
+void HCSVirtualMachine::recover_from_failed_save()
+{
+    remove_saved_state_file_if_exists();
```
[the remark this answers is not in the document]
theartful: The strategy here as discussed with Mustafa is to simply warn in `recover_from_failed_save`, and `remove_saved_state_file_if_exists` already does that. After trying to recover, we throw an exception in `suspend`.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
theartful: Discussed on MM

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
theartful: Addressed in the latest commit

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
theartful: Addressed in the latest commit

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
theartful: Addressed in the latest commit

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -121,16 +121,20 @@ HCSVirtualMachine::HCSVirtualMachine(const std::string& network_guid,
...
     mpl::debug(get_name(),
                "HCSVirtualMachine() > created_from_scratch: {}, state: {}",
                created_from_scratch,
-               state);
+               compute_state);
 
     // Reflect compute system's state
-    set_state(state);
-    HCSVirtualMachine::handle_state_update();
+    auto prev_state = state;
```
tobe2098: This could be const

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -121,16 +121,20 @@ HCSVirtualMachine::HCSVirtualMachine(const std::string& network_guid,
...
+    set_state(compute_state);
+
+    // Persist initial state even if unchanged
+    if (prev_state == this->state)
+        HCSVirtualMachine::handle_state_update();
```
tobe2098: Is there a reason to persist the state in this case?

tobe2098 (changes requested):
Great work Essam! Just a couple of nitpicks/questions

I also want to ask, in the PR description you describe the issue as being partially caused by the try_action loop. That loop is not present anymore in main or in this branch. Have you tested if the issue still occurs?

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
theartful: Just preserving the old behavior because copilot complained :")
I centralized setting the state to `set_state` function to ensure that `handle_state_update` is always called  when the state changes.

theartful:
The infinite loop happened before in the `try_action_loop`, but the root cause is that the state file was not removed. The central change in this PR is
```C++
        if (state == State::suspended)
        {
            if (const auto ec = remove_saved_state_file_if_exists(); ec)
                throw ShutdownComputeSystemException("Could not remove state file '{}': {}",
                                                     get_saved_state_file_path(),
                                                     ec);
            update_current_state();
            return;
        }
```
in the shutdown function.

tobe2098 (approved):
On view of the previous discussion, testing and my review, it LGTM Essam!

PR #5270, opened 2026-09-17
[daemon] Check persisted spec for initial state

xmkg:
The VM can (and will) change the spec once it's constructed. Keep an
immutable copy of the initial spec so the code can use it to observe
the persisted state instead of the runtime state.

MULTI-2910


src/daemon/daemon.cpp
```diff
@@ -1385,23 +1388,32 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
...
+                mpl::warn(
+                    category,
+                    "{} is deleted but has incompatible state {}, resetting state to {} (stopped)",
+                    name,
+                    static_cast<int>(spec.state),
+                    static_cast<int>(e_state::stopped));
+                auto& mutable_spec = entry.second;
+                mutable_spec.state = e_state::stopped;
+            }
+            continue;
```
Laefy, line 1403: Not fond of the pattern where we're continuing/breaking out of a loop and there's still so much code afterward. It's easy to make mistakes later by not realizing there's a branch that got out of the loop or function earlier than the other. Unless the code is short, or it's an early-exit pattern, I think it's best to avoid it.
It's only an opinion, so I let you decide if you prefer to keep it or not

src/daemon/daemon.cpp
```diff
@@ -1322,8 +1322,10 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
 
     for (auto& entry : vm_instance_specs)
     {
-        const auto& name = entry.first;
-        auto& spec = entry.second;
+        // Immutable copy of the persisted spec and the name -- VMs can update the
+        // spec once they're created through persist calls and the code needs to know the initial
+        // state for several things, like auto-resume.
+        const auto [name, spec] = entry;
```
Laefy: For clarity, can you keep the code the same here (halfway through the review, I had already forgot that the spec was a copy 😅​), and instead have an additional: `const auto original_state = spec.state;`
\+ it makes it self explanatory and you don't need the comment anymore (at least not at this spot)

src/daemon/daemon.cpp
```diff
@@ -1385,23 +1388,32 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
...
+                mutable_spec.state = e_state::stopped;
+            }
+            continue;
         }
 
-        if (!spec.deleted)
-            init_mounts(name);
+        // No deleted spec must cross this boundary.
+        assert(!spec.deleted);
+        init_mounts(name);
```
Laefy, line 1408: Not related to your PR really, but why are we initializing mounts here (+ when recovering and cloning) and not when an instance is started? Is there a reason to manage them on stopped instances?

Laefy (changes requested):
Just one little thing that could be improved I think.
Also, can you update the PR description to explain the intention behind the change? Can be something like: running VM get auto suspended when shutting down (?) the daemon, and so their state when restarted is wrongfully put back to suspended instead of running.
Thanks! :)

src/daemon/daemon.cpp
xmkg: Would it be clearer if we made the loop iterate by value instead? I'm not really fond of exposing a mutable reference to the downstream code.

src/daemon/daemon.cpp
xmkg: Hmm... the alternative is to have an else block here, which also has the following downsides:

- one level of additional nesting
- you can still write code in the wrong scope

What do you think? Do you think it would be better to have an else block here, or did you have something else in mind? I'm at 60/40 here, so I don't have a strong preference.

Laefy: I don't have any quick fix in mind when letting the rest of the code the same way.
I'm not sure what are the responsibilities of everybody here: like, why are the VMs handling their own restoration, if the daemon still has work to do on them afterward? Maybe the VM could transmit to the daemon what is the next step to do, since it look like the VM are not managing their transitions themselves (at least not here).
Anyway, I think it'd be quite a minefield to change, so probably just leave it like that ^^

src/daemon/daemon.cpp
xmkg: Discussed out-of-band.

src/daemon/daemon.cpp
xmkg: The first question should be why the mount handlers are owned by the daemon in the first place. The short answer to your question is: wrong ownership semantics.

src/daemon/daemon.cpp
```diff
@@ -1385,23 +1388,32 @@ mp::Daemon::Daemon(std::unique_ptr<const DaemonConfig> the_config)
...
+            {
+                mpl::warn(
+                    category,
+                    "{} is deleted but has incompatible state {}, resetting state to {} (stopped)",
+                    name,
+                    static_cast<int>(spec.state),
+                    static_cast<int>(e_state::stopped));
+                auto& mutable_spec = entry.second;
+                mutable_spec.state = e_state::stopped;
+            }
```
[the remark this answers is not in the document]
xmkg: I hope not. A deleted and running VM would be a fundamental bug.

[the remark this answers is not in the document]
Laefy: However, there is still one issue with that one: the state of the VM is now put to stopped, but its mounts have not been initialized

src/daemon/daemon.cpp
xmkg: Done

Laefy (approved):
Just PR desc to update

PR #5271, opened 2026-09-17
[hcs] Implement suspend on destruct

### LINKED ISSUE #5269: HCS backend does not suspend running VMs on destruction
**Describe the bug**
Host compute system VMs are not suspending running VMs on destruction. This is the case for the other backends so HCS should mimic the behavior.

xmkg:
# Description
Implement suspend-on-destruct behavior for HCS virtual machines.

Closes #5269


xmkg:
#5270 needs to land first to test this end-to-end.

tests/unit/hyperv_api/test_ut_hyperv_hcs_virtual_machine.cpp
```diff
@@ -512,20 +499,68 @@ TEST_F(HyperVHCSVirtualMachine_UnitTests, vm_suspend_success)
...
+    InSequence sequence;
+    EXPECT_CALL(monitor, persist_state_for(dummy_vm_name, mp::VirtualMachine::State::running));
+    EXPECT_CALL(mock_hcs, pause_compute_system(Eq(mock_handle))).Times(1);
+    EXPECT_CALL(
+        mock_hcs,
+        save_compute_system(Eq(mock_handle), Property(&mhv::hcs::HcsPath::get, saved_state_file)))
+        .Times(1);
+    EXPECT_CALL(mock_hcs, terminate_compute_system(Eq(mock_handle))).Times(1);
+    EXPECT_CALL(monitor, persist_state_for(dummy_vm_name, mp::VirtualMachine::State::off));
+    EXPECT_CALL(monitor, persist_state_for(dummy_vm_name, mp::VirtualMachine::State::suspended));
```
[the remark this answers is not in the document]
xmkg: The test passes.

src/platform/backends/hyperv_api/hcs_virtual_machine.cpp
```diff
@@ -133,6 +133,20 @@ HCSVirtualMachine::HCSVirtualMachine(const std::string& network_guid,
...
+HCSVirtualMachine::~HCSVirtualMachine()
+{
+    top_catch_all(vm_name, [this]() {
+        if (current_state() != State::running)
+            return;
+
+        // Auto-suspend if running
+        suspend();
+        // Persist previous VM state
+        state = VirtualMachine::State::running;
```
theartful: Why do we persist the state as "running">

theartful, line 145: Why do we persist the state as "running"?

xmkg: (discussed live)

theartful (approved):
Ideally I think we should have a type similar to `Watchable<T>` or something, and the state becomes `Watchable<VirtualMachine::State>`. Setting the value of this type would have `notify_all` semantics, and it would have something like `wait` and `wait_for` to wait for a specific value.

This is beyond the scope of the PR though :")

PR #5272, opened 2026-09-17
[tests-cli] Collapse help tests to 2

xmkg:
Help tests are parameterized and outnumber the VM tests by a lot. This creates a false feeling of progress when the help tests are being executed since they're quite fast.

Reduce them to just 2 while keeping the exercised test path as-is.

```
BEFORE —  f8f06620a  (fully parametrized): 249 items

cli_help_test.py::TestHelp::test_help[C]
cli_help_test.py::TestHelp::test_help[C.UTF-8]
cli_help_test.py::TestHelp::test_help[en_US.UTF-8]
cli_help_test.py::TestHelp::test_help[tr_TR.UTF-8]
cli_help_test.py::TestHelp::test_help[de_DE.UTF-8]
cli_help_test.py::TestHelp::test_help[fr_FR.UTF-8]
   ... (8 locales × ~30 commands for test_per_command_help) ...
cli_help_test.py::TestHelp::test_per_command_help[wait-ready-en-US]
cli_help_test.py::TestHelp::test_per_command_help[wait-ready-tr-TR]

(8  test_help  + 8×30  test_per_command_help  = 249)

AFTER — current: 2 items

cli_help_test.py::TestHelp::test_help
cli_help_test.py::TestHelp::test_per_command_help

```


Laefy (approved):
Depends on how you're running tests, but it's kind of bothering when you have a breakage in a specific test, and you're actually unable to re-execute just that one. Here it's ok because these tests run fast, but generally, I'm quite against regrouping tests. The more tests can be independent, the better (can assert more locally, and interrupt quicker, instead of to batch failures and giving them back to the user, which is already the purpose of the testing framework). If you really want to make that change, I'm ok with it, but I really don't think that the progress bar of the tests should be a concerned compared to guaranteeing tests isolation

